PR #408, opened 2019-08-14
Improve equals() on config objects

narendly:
### Issues

Fixes $407

### Description

ClusterConfig, InstanceConfig, ResourceConfig's equals() methods were only comparing the IDs, which is not desirable because it is possible to have two instances of each with the same IDs but different contents/fields. This diff improves the equals() methods for these configs so that the actual contents are compared as well.

Changelist:
1. Change equals() on the said config classes so that the actual content is compared

### Tests

Simple equals() method. No tests needed.

### Commits

### Code Quality


narendly:
@lei-xia authored the original version of equals() for these config classes. Need his input.

i3wangyi:
It'd be better to write some unit tests to justify 
- cases where original equals won't work but it should
- cases with your change, the equals method works

Issue #409, opened 2019-08-14
TASK: Execution delay is not respected for the jobs


alirezazamani:
Current implementation of Task Framework checks if a Job is ready for the scheduling or not. If not, the job is inserted in the inflight jobs without any delay time restrictions which is not desirable.

alirezazamani:
It is fixed in PR #410.

PR #410, opened 2019-08-15
Fix the execution delay for the jobs (409)

alirezazamani:
**Issues**
Fix the execution delay for the jobs #409 .

**Description**
In the Task Framework part of helix, execution delay for the jobs is not respected.
In this commit, when the job is extracted from the inflighJobs queue, the timeline has been checked before scheduling.

**Tests**
The following tests are written for this issue:
TestExecutionDelay.java

Test Result 1: mvn test
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 843, Failures: 1, Errors: 0, Skipped: 1
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  54:26 min
[INFO] Finished at: 2019-08-21T10:00:54-07:00
[ERROR] 
[ERROR] Please refer to /home/anajari/my_repos/helix/helix-core/target/surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] 

Test Result 2: mvn test -Dtest="TestAlertingRebalancerFailure"
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 7.05 s - in org.apache.helix.integration.TestAlertingRebalancerFailure
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  11.735 s
[INFO] Finished at: 2019-08-21T10:26:21-07:00

**Commits**
My commits all reference appropriate Apache Helix GitHub issues in their subject lines, and I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines:
  - Subject is separated from body by a blank line
  - Subject is limited to 50 characters 
  - Subject does not end with a period
  - Subject uses the imperative mood ("add", not "adding")
  - Body wraps at 72 characters
  - Body explains "what" and "why", not "how"

**Code Quality**
My diff has been formatted using helix-style.xml

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
+++ b/helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
@@ -165,7 +165,9 @@ public class WorkflowDispatcher extends AbstractTaskDispatcher {
     RuntimeJobDag runtimeJobDag = _clusterDataCache.getTaskDataCache().getRuntimeJobDag(workflow);
     if (runtimeJobDag != null) {
       for (String inflightJob : runtimeJobDag.getInflightJobList()) {
-        processJob(inflightJob, currentStateOutput, bestPossibleOutput, workflowCtx);
+        if (System.currentTimeMillis() >= workflowCtx.getJobStartTime(inflightJob)) {
+          processJob(inflightJob, currentStateOutput, bestPossibleOutput, workflowCtx);
+        }
       }
     } else {
       LOG.warn(String.format(
@@ -269,29 +271,16 @@ public class WorkflowDispatcher extends AbstractTaskDispatcher {
           clusterDataCache, clusterDataCache.getAssignableInstanceManager())) {
         JobConfig jobConfig = jobConfigMap.get(job);
 
-        // Since the start time is calculated base on the time of completion of parent jobs for this
-        // job, the calculated start time should only be calculate once. Persist the calculated time
-        // in WorkflowContext znode.
-        long calculatedStartTime = workflowCtx.getJobStartTime(job);
-        if (calculatedStartTime < 0) {
-          // Calculate the start time if it is not already calculated
-          calculatedStartTime = System.currentTimeMillis();
-          // If the start time is not calculated before, do the math.
-          if (jobConfig.getExecutionDelay() >= 0) {
-            calculatedStartTime += jobConfig.getExecutionDelay();
-          }
-          calculatedStartTime = Math.max(calculatedStartTime, jobConfig.getExecutionStart());
-          workflowCtx.setJobStartTime(job, calculatedStartTime);
-        }
-
+        long calculatedStartTime = computeStartTimeForJob(workflowCtx, job, jobConfig);
         // Time is not ready. Set a trigger and update the start time.
-        if (System.currentTimeMillis() < calculatedStartTime) {
-          timeToSchedule = Math.min(timeToSchedule, calculatedStartTime);
-        } else {
+        // Check if the job is ready to be executed.
+        if (System.currentTimeMillis() >= workflowCtx.getJobStartTime(job)) {
           scheduleSingleJob(job, jobConfig);
           workflowCtx.setJobState(job, TaskState.NOT_STARTED);
           processJob(job, currentStateOutput, bestPossibleOutput, workflowCtx);
           scheduledJobs++;
+        } else {
+          timeToSchedule = Math.min(timeToSchedule, calculatedStartTime);
         }
       }
       nextJob = jobDag.getNextJob();
@@ -602,4 +591,24 @@ public class WorkflowDispatcher extends AbstractTaskDispatcher {
     }
     cache.removeContext(workflow);
   }
+
+  private long computeStartTimeForJob(WorkflowContext workflowCtx, String job,
+      JobConfig jobConfig) {
+    // Since the start time is calculated base on the time of completion of parent jobs for this
+    // job, the calculated start time should only be calculated once. Persist the calculated time
+    // in WorkflowContext znode.
+    long calculatedStartTime = workflowCtx.getJobStartTime(job);
+    if (calculatedStartTime < 0) {
+      // Calculate the start time if it is not already calculated
+      calculatedStartTime = System.currentTimeMillis();
+      // If the start time is not calculated before, do the math.
+      if (jobConfig.getExecutionDelay() >= 0) {
+        calculatedStartTime += jobConfig.getExecutionDelay();
+      }
+      calculatedStartTime = Math.max(calculatedStartTime, jobConfig.getExecutionStart());
+      workflowCtx.setJobStartTime(job, calculatedStartTime);
+    }
+    return calculatedStartTime;
+  }
+
 }
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
@@ -1,0 +1,132 @@
+package org.apache.helix.integration.task;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
+import java.util.List;
+import org.apache.helix.HelixAdmin;
+import org.apache.helix.HelixDataAccessor;
+import org.apache.helix.HelixException;
+import org.apache.helix.PropertyKey;
+import org.apache.helix.PropertyPathBuilder;
+import org.apache.helix.TestHelper;
+import org.apache.helix.task.JobConfig;
+import org.apache.helix.task.JobContext;
+import org.apache.helix.task.JobQueue;
+import org.apache.helix.task.TaskConstants;
+import org.apache.helix.task.TaskState;
+import org.apache.helix.task.TaskUtil;
+import org.apache.helix.task.Workflow;
+import org.apache.helix.task.WorkflowContext;
+import org.apache.helix.tools.ClusterVerifiers.BestPossibleExternalViewVerifier;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/**
+ * Test To check if the Execution Delay is respected.
+ */
+public class TestExecutionDelay extends TaskTestBase {
+
+  @BeforeClass
+  public void beforeClass() throws Exception {
+    super.beforeClass();
+  }
+
+  @Test
+  public void testExecutionDelay() throws Exception {
+    // Execution Delay is set to be a large number. Hence, the workflow should not be completed
+    // soon.
+    final long executionDelay = 100000000000L;
+    String workflowName = TestHelper.getTestMethodName();
+    Workflow.Builder builder = new Workflow.Builder(workflowName);
+
+    // Workflow DAG Schematic:
+    //          JOB0
+    //           /\
+    //          /  \
+    //         /    \
+    //       JOB1   JOB2
+
+    JobConfig.Builder jobBuilder = JobConfig.Builder.fromMap(WorkflowGenerator.DEFAULT_JOB_CONFIG)
+        .setMaxAttemptsPerTask(1).setWorkflow(workflowName)
+        .setJobCommandConfigMap(ImmutableMap.of(MockTask.JOB_DELAY, "1000"));
+
+    JobConfig.Builder jobBuilder2 = JobConfig.Builder.fromMap(WorkflowGenerator.DEFAULT_JOB_CONFIG)
+        .setMaxAttemptsPerTask(1).setWorkflow(workflowName)
+        .setJobCommandConfigMap(ImmutableMap.of(MockTask.JOB_DELAY, "1000"));
+
+    JobConfig.Builder jobBuilder3 = JobConfig.Builder.fromMap(WorkflowGenerator.DEFAULT_JOB_CONFIG)
+        .setMaxAttemptsPerTask(1).setWorkflow(workflowName)
+        .setJobCommandConfigMap(ImmutableMap.of(MockTask.JOB_DELAY, "1000"));
+
+    builder.addParentChildDependency("JOB0", "JOB1");
+    builder.addParentChildDependency("JOB0", "JOB2");
+    builder.addJob("JOB0", jobBuilder.setExecutionDelay(executionDelay));
+    builder.addJob("JOB1", jobBuilder2);
+    builder.addJob("JOB2", jobBuilder3);
+
+    // Since the execution delay is set as a long time, it is expected to reach time out.
+    // If the workflow completed without any exception, it means that ExecutionDelay has not been
+    // respected.
+    _driver.start(builder.build());
+
+    // Verify workflow Context is created.
+    // Wait until Workflow Context is created.
+    boolean workflowCreated = TestHelper.verify(() -> {
+      WorkflowContext wCtx1 = _driver.getWorkflowContext(workflowName);
+      return wCtx1 != null;
+    }, 30 * 1000);
+    Assert.assertTrue(workflowCreated);
+
+    // Verify StartTime is set for the Job0.
+    // Wait until startTime is set.
+    boolean startTimeSet = TestHelper.verify(() -> {
+      WorkflowContext wCtx1 = _driver.getWorkflowContext(workflowName);
+      long startTime = -1;
+      try {
+        startTime = wCtx1.getJobStartTime(TaskUtil.getNamespacedJobName(workflowName, "JOB0"));
+      } catch (Exception e) {
+        // pass
+      }
+      return startTime != -1;
+    }, 30 * 1000);
+    Assert.assertTrue(startTimeSet);
+
+    // Verify JobContext is not created. Creation of the JobContext means that the job is ran or
+    // in-progress. The absence of JobContext implies that the job did not run.
+    boolean isJobStarted = TestHelper.verify(() -> {
+      JobContext jobCtx0 =
+          _driver.getJobContext(TaskUtil.getNamespacedJobName(workflowName, "JOB0"));
+      JobContext jobCtx1 =
+          _driver.getJobContext(TaskUtil.getNamespacedJobName(workflowName, "JOB1"));
+      JobContext jobCtx2 =
+          _driver.getJobContext(TaskUtil.getNamespacedJobName(workflowName, "JOB2"));
+      return (jobCtx0 != null || jobCtx1 != null || jobCtx2 != null);
+    }, 30 * 1000);
+    Assert.assertFalse(isJobStarted);
+
+    WorkflowContext wCtx1 = _driver.getWorkflowContext(workflowName);
+    Assert.assertNull(wCtx1.getJobState(TaskUtil.getNamespacedJobName(workflowName, "JOB0")));
+    Assert.assertNull(wCtx1.getJobState(TaskUtil.getNamespacedJobName(workflowName, "JOB1")));
+    Assert.assertNull(wCtx1.getJobState(TaskUtil.getNamespacedJobName(workflowName, "JOB2")));
+  }
+}


helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.helix.task.TaskUtil;
+import org.apache.helix.task.Workflow;
+import org.apache.helix.task.WorkflowContext;
+import org.apache.helix.tools.ClusterVerifiers.BestPossibleExternalViewVerifier;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+public class TestExecutionDelay extends TaskTestBase {
+  private HelixAdmin admin;
```
narendly: Let's use an underscore for private variables.
`admin -> _admin`

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,79 @@
...
+
+  @BeforeClass public void beforeClass() throws Exception {
```
narendly: Did you format the diff with helix-style? @BeforeClass should take up a line to itself.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,79 @@
...
+    admin = _gSetupTool.getClusterManagementTool();
+    super.beforeClass();
```
narendly: Minor: usually superclass's methods are called first.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Test(expectedExceptions = HelixException.class) public void testExecutionDelay()
```
narendly: Style - annotation should take up a line on its own

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,79 @@
...
+    builder.addParentChildDependency("JOB0", "JOB1");
+    builder.addParentChildDependency("JOB0", "JOB2");
+    builder.addJob("JOB0", jobBuilder.setExecutionDelay(100000000L));
+    builder.addJob("JOB1", jobBuilder2);
+    builder.addJob("JOB2", jobBuilder3);
+
+    _driver.start(builder.build());
+
+    _driver.pollForWorkflowState(workflowName, 15000, TaskState.COMPLETED);
+
```
narendly: Could we remove this empty line? :)

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
```
narendly: Could we remove this empty line? :)

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,79 @@
...
+    builder.addParentChildDependency("JOB0", "JOB1");
+    builder.addParentChildDependency("JOB0", "JOB2");
+    builder.addJob("JOB0", jobBuilder.setExecutionDelay(100000000L));
+    builder.addJob("JOB1", jobBuilder2);
+    builder.addJob("JOB2", jobBuilder3);
+
+    _driver.start(builder.build());
+
+    _driver.pollForWorkflowState(workflowName, 15000, TaskState.COMPLETED);
```
narendly: For longs, please append "L" at the end.
`15000 -> 15000L`

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
narendly: Where is this HelixAdmin used?

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,79 @@
...
+            .setJobCommandConfigMap(ImmutableMap.of(MockTask.JOB_DELAY, "1000"));
+
+    JobConfig.Builder jobBuilder3 =
+        JobConfig.Builder.fromMap(WorkflowGenerator.DEFAULT_JOB_CONFIG).setMaxAttemptsPerTask(1)
+            .setWorkflow(workflowName)
+            .setJobCommandConfigMap(ImmutableMap.of(MockTask.JOB_DELAY, "1000"));
+
+    builder.addParentChildDependency("JOB0", "JOB1");
+    builder.addParentChildDependency("JOB0", "JOB2");
+    builder.addJob("JOB0", jobBuilder.setExecutionDelay(100000000L));
```
narendly: Magic number -> make it a constant?

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
narendly: Magic number - make it a constant?

helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
```diff
@@ -165,7 +165,9 @@ public void updateWorkflowStatus(String workflow, WorkflowConfig workflowCfg,
     RuntimeJobDag runtimeJobDag = _clusterDataCache.getTaskDataCache().getRuntimeJobDag(workflow);
     if (runtimeJobDag != null) {
       for (String inflightJob : runtimeJobDag.getInflightJobList()) {
-        processJob(inflightJob, currentStateOutput, bestPossibleOutput, workflowCtx);
+        if (System.currentTimeMillis() >= workflowCtx.getJobStartTime(inflightJob)) {
```
narendly: For the execution delay itself, adding this check may do the trick. However, two more things to consider:
1. Would this be the right place to place the check? Say, should we include this check inside processJob()? Or should we create a separate method that checks whether a job is eligible to be processed?
2. Could there be more checks we need to perform other than execution delay? It would be great if we could make this a comprehensive fix :)

narendly:
A comment on the test you added. I had a hard time understanding what you're trying to test in your test. I think we could make the test more robust by adding a more intuitive test case, and also add some description in your test so that anyone could see what you're doing clearly.

helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
alirezazamani: Thank you for your comments.
1- Based on same logic that is added few line below (which calculates the start time and store these information and process the job if possible), this is the place that this check should happen. (same as the scheduleJobs method in WorkflowDistpatcher).
2- About the comprehensive change, I agree with you. This is just a quick fix. We will need to revisit several concepts such as inflight jobs and retry delay. I am working on the retry delay and will create new issue for that as well. After that we can think on a comprehensive changes to the Task Framework.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
alirezazamani: I fixed that. Thanks.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
alirezazamani: Fixed.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
alirezazamani: Not used. I deleted it.

helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
```diff
@@ -602,4 +590,28 @@ private void removeContextsAndPreviousAssignment(String workflow, Set<String> jo
...
+        calculatedStartTime += jobConfig.getExecutionDelay();
+      }
+      calculatedStartTime = Math.max(calculatedStartTime, jobConfig.getExecutionStart());
+      workflowCtx.setJobStartTime(job, calculatedStartTime);
+    }
+    return calculatedStartTime;
+  }
+
+  private boolean isJobReadyToExecute(WorkflowContext workflowCtx, String job) {
+    if (System.currentTimeMillis() < workflowCtx.getJobStartTime(job)) {
```
narendly: Simplify the boolean condition?

helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
```diff
@@ -269,29 +271,15 @@ private void scheduleJobs(String workflow, WorkflowConfig workflowCfg,
           clusterDataCache, clusterDataCache.getAssignableInstanceManager())) {
         JobConfig jobConfig = jobConfigMap.get(job);
 
-        // Since the start time is calculated base on the time of completion of parent jobs for this
```
narendly: Could you preserve this comment? I think it is valuable.

helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
```diff
@@ -602,4 +590,28 @@ private void removeContextsAndPreviousAssignment(String workflow, Set<String> jo
     }
     cache.removeContext(workflow);
   }
+
+  private long calculateStartTime(WorkflowContext workflowCtx, String job, JobConfig jobConfig) {
```
narendly: I think computeStartTimeForJob or something like that is more intuitive and specific. What do you think?

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.helix.task.Workflow;
+import org.apache.helix.task.WorkflowContext;
+import org.apache.helix.tools.ClusterVerifiers.BestPossibleExternalViewVerifier;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/**
+ * Test To check if the Execution Delay is respected.
+ * If Workflow is completed regardless of execution delay, it means this task should be failed.
```
narendly: Clarify? This line is confusing..

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @BeforeClass
+  public void beforeClass() throws Exception {
+    super.beforeClass();
+  }
+
+  @Test(expectedExceptions = HelixException.class)
+  public void testExecutionDelay() throws InterruptedException {
+    final long timeOutTime = 15000L;
+    final long executionDelay = 100000000L;
```
narendly: Is this supposed to be some arbitrarily large number? In that case, Long.MAX_VALUE and add a simple inline comment on how it will be used.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,82 @@
...
+    String workflowName = TestHelper.getTestMethodName();
+    Workflow.Builder builder = new Workflow.Builder(workflowName);
+
+    JobConfig.Builder jobBuilder = JobConfig.Builder.fromMap(WorkflowGenerator.DEFAULT_JOB_CONFIG)
```
narendly: Add a block comment on what DAG you're creating. Something like

```
//       1
//     /    \
//   2      3
```

You get the idea? Helps others understand what you're doing.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,82 @@
...
+public class TestExecutionDelay extends TaskTestBase {
+
+  @BeforeClass
+  public void beforeClass() throws Exception {
+    super.beforeClass();
+  }
+
+  @Test(expectedExceptions = HelixException.class)
+  public void testExecutionDelay() throws InterruptedException {
+    final long timeOutTime = 15000L;
```
narendly: One might wonder, why 15 sec?

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,82 @@
...
+    builder.addParentChildDependency("JOB0", "JOB2");
+    builder.addJob("JOB0", jobBuilder.setExecutionDelay(executionDelay));
+    builder.addJob("JOB1", jobBuilder2);
+    builder.addJob("JOB2", jobBuilder3);
+
+    // Since the execution delay is set as a long time, it is expected to reach time out.
+    // If the workflow completed without any exception, it means that ExecutionDelay has not been
+    // respected.
+    _driver.start(builder.build());
+    _driver.pollForWorkflowState(workflowName, timeOutTime, TaskState.COMPLETED);
```
narendly: I think there's a better way to test this. Instead of relying on pollForWorkflowState timing out, could you check the status of the workflow **and** the individual jobs? What if the controller is very slow and takes 15 seconds to schedule jobs?

One way of not relying on timing is
1. Start workflow
2. Wait until it is in progress and job 0 has been looked at (controller tried to schedule)
3. Read the context and check start time is set with the right amount
4. Check that job 0 and its children jobs have not run by checking their states

This way, your test is robust against timing issues, correct? No matter how long controller takes to start and schedule that first job instead of relying on 15 sec, an arbitrarily long enough number? :)

helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
```diff
@@ -602,4 +590,28 @@ private void removeContextsAndPreviousAssignment(String workflow, Set<String> jo
...
+      calculatedStartTime = Math.max(calculatedStartTime, jobConfig.getExecutionStart());
+      workflowCtx.setJobStartTime(job, calculatedStartTime);
+    }
+    return calculatedStartTime;
+  }
+
+  private boolean isJobReadyToExecute(WorkflowContext workflowCtx, String job) {
+    if (System.currentTimeMillis() < workflowCtx.getJobStartTime(job)) {
+      return false;
+    } else {
```
dasahcc: else clause is not needed

helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
```diff
@@ -165,7 +165,9 @@ public void updateWorkflowStatus(String workflow, WorkflowConfig workflowCfg,
     RuntimeJobDag runtimeJobDag = _clusterDataCache.getTaskDataCache().getRuntimeJobDag(workflow);
     if (runtimeJobDag != null) {
       for (String inflightJob : runtimeJobDag.getInflightJobList()) {
-        processJob(inflightJob, currentStateOutput, bestPossibleOutput, workflowCtx);
+        if (isJobReadyToExecute(workflowCtx, inflightJob)) {
```
dasahcc: I am OK to fix it right now. But please put a note or TODO here. Adding non-scheduled job to inflight job list make people conceptually confused.

narendly: @dasahcc I discussed with @alirezazamani offline. I feel that we just neglected all the checking required before calling `processJob` on inflight jobs. "Inflight" in this sense doesn't mean that a job is necessarily running; rather, it means that the job has been passed over onto the Controller and now out of RuntimeJobDAG's responsibility. Otherwise we end up doing a lot of "passing jobs back and forth" for every failure/delay situation that will complicate the code/logic even further.

So in short, it was a simple bug that we didn't check before doing processJob here.

helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
```diff
@@ -602,4 +590,33 @@ private void removeContextsAndPreviousAssignment(String workflow, Set<String> jo
...
+      calculatedStartTime = Math.max(calculatedStartTime, jobConfig.getExecutionStart());
+      workflowCtx.setJobStartTime(job, calculatedStartTime);
+    }
+    return calculatedStartTime;
+  }
+
+  private boolean isJobReadyToExecute(WorkflowContext workflowCtx, String job) {
+    long calculatedStartTime = workflowCtx.getJobStartTime(job);
+    long timeNow = System.currentTimeMillis();
+    if (timeNow < calculatedStartTime) {
```
narendly: This method could be simplified to:

`return System.currentTimeMillis() >= workflowCtx.getJobStartTime(job);`

Also, are we sure that there's no other checks required in this function? If it's just going to be a one-liner (in other words, there are no other checks to perform than checking the calculatedStartTime against current time to determine if a job is ready to be processed/updated), and if the usage is only in this place, perhaps it may not be worth it to make it a separate function. If that's the case, let's just keep what we had inline.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,124 @@
...
+    builder.addJob("JOB2", jobBuilder3);
+
+    // Since the execution delay is set as a long time, it is expected to reach time out.
+    // If the workflow completed without any exception, it means that ExecutionDelay has not been
+    // respected.
+    _driver.start(builder.build());
+
+    // Verify workflow Context is created.
+    // Wait until Workflow Context is created.
+    boolean resultWorkflow = TestHelper.verify(() -> {
```
narendly: Minor: 
Boolean naming convention
https://stackoverflow.com/questions/40339847/variable-and-method-naming-conventions-for-boolean-verbs-in-java

workflowCreated might be a better option here?

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,124 @@
...
+    // Wait until startTime is set.
+    boolean resultStartTime = TestHelper.verify(() -> {
+      WorkflowContext wCtx1 = _driver.getWorkflowContext(workflowName);
+      long startTime = -1;
+      try {
+        startTime = wCtx1.getJobStartTime(TaskUtil.getNamespacedJobName(workflowName, "JOB0"));
+      } catch (Exception e) {
+        // pass
+      }
+      return (startTime != -1);
```
narendly: parentheses not needed.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,124 @@
...
+    }, 60 * 1000);
+    Assert.assertTrue(resultStartTime);
+
+    // Verify Job Context is not created. If Job Context is created, it means that the execution
```
narendly: Explain that job context creation implies that the job ran.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,124 @@
...
+    // delay is not respected. if not created, it means that the execution delay is respected.
```
narendly: is respected because the absence of JobContext implies that the job did not run.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,124 @@
...
+    boolean resultJob = TestHelper.verify(() -> {
```
narendly: Workflow context at this point exists and the start time has been set. Is there a reason you're using `TestHelper.verify()?` You don't need it here

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,124 @@
...
+      JobContext jobCtx =
```
narendly: Verifying the absence of JobContexts for JOB0 makes sense, but what about JOB1 and JOB2?

Alternatively, you should also do a check for the status using getJobState() or getJobStates()

helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
alirezazamani: There is no other checks it needed. I agree. I moved it inline to the caller function.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
alirezazamani: @narendly In this case we need this verify. The reason is to avoid using Thread.sleep and then check the jobstate using getJobState. It takes some time for the controller to write the job context and job state to the zookeeper. There are cases where startTime and workflow context is set but the job context is empty and getJobState returns null.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
alirezazamani: Fixed. In new commit I am checking if there is no JobContext for all the jobs.

helix-core/src/test/java/org/apache/helix/integration/task/TestExecutionDelay.java
```diff
@@ -0,0 +1,132 @@
...
+      JobContext jobCtx1 =
+          _driver.getJobContext(TaskUtil.getNamespacedJobName(workflowName, "JOB1"));
+      JobContext jobCtx2 =
+          _driver.getJobContext(TaskUtil.getNamespacedJobName(workflowName, "JOB2"));
+      return (jobCtx0 != null || jobCtx1 != null || jobCtx2 != null);
+    }, 30 * 1000);
+    Assert.assertFalse(isJobStarted);
+
+
+    Assert.assertNull(_driver.getWorkflowContext(workflowName).getJobState(TaskUtil.getNamespacedJobName(workflowName, "JOB0")));
```
narendly: We seem to be making a ZK read call for each check. Is it possible to just do one read and do the checks on one read?

alirezazamani:
@narendly mvn test results have been updated in PR.

alirezazamani:

Issue #412, opened 2019-08-15
Add a cluster level default instance capacity config for the WAGED rebalancer


jiajunwang:
During the discussion with our potential customers, we noticed that a default instance capacity could be very helpful.

The default capacity will be applied to the instance when the instance level capacity map is empty or not configured.

PR #413, opened 2019-08-15
Add cluster level default instance config.

jiajunwang:
### Issues

#412

### Description

Add cluster level default instance config.
This config will be applied to the instance when there is no (or empty) capacity configuration in the Instance Config.
Also add unit tests.

### Tests

TestAssignableNode
TestClusterConfig

NA since the WAGED rebalancer has not been integrated yet.

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
@@ -88,15 +88,15 @@ public class AssignableNode {
       Collection<AssignableReplica> existingAssignment) {
     reset();
 
-    Map<String, Integer> instanceCapaity = instanceConfig.getInstanceCapacityMap();
-    if (instanceCapaity.isEmpty()) {
-      instanceCapaity = clusterConfig.getDefaultInstanceCapacityMap();
+    Map<String, Integer> instanceCapacity = instanceConfig.getInstanceCapacityMap();
+    if (instanceCapacity.isEmpty()) {
+      instanceCapacity = clusterConfig.getDefaultInstanceCapacityMap();
     }
-    _currentCapacity.putAll(instanceCapaity);
+    _currentCapacity.putAll(instanceCapacity);
     _faultZone = computeFaultZone(clusterConfig, instanceConfig);
     _instanceTags = new HashSet<>(instanceConfig.getTags());
     _disabledPartitionsMap = instanceConfig.getDisabledPartitionsMap();
-    _maxCapacity = instanceCapaity;
+    _maxCapacity = instanceCapacity;
     _maxPartition = clusterConfig.getMaxPartitionsPerInstance();
 
     assignNewBatch(existingAssignment);
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
@@ -88,11 +88,15 @@ public class AssignableNode {
       Collection<AssignableReplica> existingAssignment) {
     reset();
 
-    _currentCapacity.putAll(instanceConfig.getInstanceCapacityMap());
+    Map<String, Integer> instanceCapaity = instanceConfig.getInstanceCapacityMap();
+    if (instanceCapaity.isEmpty()) {
+      instanceCapaity = clusterConfig.getDefaultInstanceCapacityMap();
+    }
+    _currentCapacity.putAll(instanceCapaity);
     _faultZone = computeFaultZone(clusterConfig, instanceConfig);
     _instanceTags = new HashSet<>(instanceConfig.getTags());
     _disabledPartitionsMap = instanceConfig.getDisabledPartitionsMap();
-    _maxCapacity = instanceConfig.getInstanceCapacityMap();
+    _maxCapacity = instanceCapaity;
     _maxPartition = clusterConfig.getMaxPartitionsPerInstance();
 
     assignNewBatch(existingAssignment);
--- a/helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
+++ b/helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
@@ -720,9 +720,16 @@ public class ClusterConfig extends HelixProperty {
   }
 
   /**
-   * Set the default instance capacity information with an Integer mapping
+   * Set the default instance capacity information with an Integer mapping.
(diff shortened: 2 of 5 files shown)


helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -700,6 +703,48 @@ public void setInstanceCapacityKeys(List<String> capacityKeys) {
...
+   */
+  public void setDefaultInstanceCapacityMap(Map<String, Integer> capacityDataMap)
+      throws IllegalArgumentException {
+    if (capacityDataMap == null || capacityDataMap.size() == 0) {
+      throw new IllegalArgumentException("Default Instance Capacity Data is empty");
+    }
+
+    Map<String, String> capacityData = new HashMap<>();
+
+    capacityDataMap.entrySet().stream().forEach(entry -> {
```
narendly: Is it better to stream this vs doing Collection.forEach?

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/model/TestAssignableNode.java
```diff
@@ -200,4 +200,17 @@ public void testParseFaultZone() throws IOException {
...
+  public void testDefaultInstanceCapacity() {
+    ClusterConfig testClusterConfig = new ClusterConfig("testClusterConfigId");
+    testClusterConfig.setDefaultInstanceCapacityMap(_capacityDataMap);
+
+    InstanceConfig testInstanceConfig = new InstanceConfig("testInstanceConfigId");
+
+    AssignableNode assignableNode =
+        new AssignableNode(testClusterConfig, testInstanceConfig, _testInstanceId,
+            Collections.emptyList());
+    Assert.assertEquals(assignableNode.getMaxCapacity(), _capacityDataMap);
```
narendly: Why getMaxCapacity? shouldn't you get default capacity instead?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -88,11 +88,15 @@ private void refresh(ClusterConfig clusterConfig, InstanceConfig instanceConfig,
...
+    Map<String, Integer> instanceCapaity = instanceConfig.getInstanceCapacityMap();
+    if (instanceCapaity.isEmpty()) {
+      instanceCapaity = clusterConfig.getDefaultInstanceCapacityMap();
+    }
+    _currentCapacity.putAll(instanceCapaity);
     _faultZone = computeFaultZone(clusterConfig, instanceConfig);
     _instanceTags = new HashSet<>(instanceConfig.getTags());
     _disabledPartitionsMap = instanceConfig.getDisabledPartitionsMap();
-    _maxCapacity = instanceConfig.getInstanceCapacityMap();
+    _maxCapacity = instanceCapaity;
```
narendly: typo!

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -88,11 +88,15 @@ private void refresh(ClusterConfig clusterConfig, InstanceConfig instanceConfig,
       Collection<AssignableReplica> existingAssignment) {
     reset();
 
-    _currentCapacity.putAll(instanceConfig.getInstanceCapacityMap());
+    Map<String, Integer> instanceCapaity = instanceConfig.getInstanceCapacityMap();
```
narendly: Do:
```
_maxCapacity = instanceConfig.getInstanceCapacityMap();
if (_maxCapacity != null && _maxCapacity.isEmpty()) {
  _maxCapacity = clusterConfig.getDefaultInstanceCapacityMap();
}
..
_currentCapacity.putAll(_maxCapacity);
```
Also, do we want to recompute currentCapacity from scratch every rebalance? We certainly could for an MVP but just making sure (this optimization could come later).

jiajunwang: getInstanceCapacityMap won't return null. This is guaranteed and the unit test covers it.

You are right. That one of the reasons why I tried to decouple the model classes as far as I can. In this case, if one instance is not changed, we can just keep it. That can be done in the Cluster Model Provder.
It is in my plan, but will be future.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
jiajunwang: Good catch!

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -700,6 +703,48 @@ public void setInstanceCapacityKeys(List<String> capacityKeys) {
     return capacityKeys;
   }
 
+  /**
+   * Get the default instance capacity information from the map fields.
+   *
```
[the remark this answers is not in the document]
jiajunwang: Yes...
I guess you don't like the extra line? I personally like it. But if you modify our format file, I can follow that style. No strong preference.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: It's not making much difference here.

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/model/TestAssignableNode.java
jiajunwang: The capacity is applied to "max capacity", that is also true when the default capacity is used.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
jiajunwang: Discussed with Hunter offline. I got the point now. Basically, we can potentially save the temp variable instanceCapaity with the suggested code. This is a valid point. But for this refresh logic, I prefer not to set and read the private field in the same function. Just a way to avoid confusing and reduce potential bug.

jiajunwang:
> Before checking it in, do you think we should leave a comment on what would happen if neither instance capacity nor default value is set?

Sure. Let me find a good place to comment on.

jiajunwang:
Commented on the instance capacity.

Approved by @narendly. I will merge the PR to the branch.

PR #417, opened 2019-08-15
Move partition health check method into dataAccessor layer

### LINKED ISSUE #416: `InstanceServiceImpl`'s getPartitionsHealth is too long and difficult to test
The idea is to wrap all the heavy lifting logics (especially those for getting cluster partition health zNode) in a separate class.

i3wangyi:
### Issues

fixes #416 

### Description

Move the method `generateClusterPartitionHealth` method into HelixDataAcessorWrapper, which is an ideal place to hide the complexity of ZK related operations.

### Tests

Moved the test method in TestInstanceService to a new test class called `TestHelixDataAccessorWrapper`

Start test :testInvalidHelixRestServerInitialization
10045 [main] ERROR org.apache.helix.rest.server.HelixRestServer  - Failed to initialize helix rest server. Tearing down.
10047 [main] ERROR org.apache.helix.rest.server.HelixRestServer  - Failed to initialize helix rest server. Tearing down.
End test :testInvalidHelixRestServerInitialization
Shut down zookeeper at port 2123 in thread main
Shut down zookeeper at port 2124 in thread main
[INFO] Tests run: 84, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 33.183 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 84, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  40.195 s
[INFO] Finished at: 2019-08-20T11:56:55-07:00
[INFO] --------------------------------------------------------------------

### Commits

### Documentation

### Code Quality


i3wangyi:
The refactor is not finalized. Because partitions health from ZK is not good enough, it will need to combine the logics from rest client. However, merely passing rest client won't easily solve the problem, we will end up having a lot of params.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
```diff
@@ -234,8 +235,7 @@ private StoppableCheck performCustomInstanceCheck(String clusterId, String insta
 
   private Map<String, StoppableCheck> performPartitionsCheck(List<String> instances,
       RESTConfig restConfig, Map<String, String> customPayLoads) {
-    //TODO: move the heavy partition health preparation in separate class
-    PartitionHealth clusterPartitionsHealth = generatePartitionHealthMapFromZK();
+    PartitionHealth clusterPartitionsHealth = _dataAccessor.getPartitionHealthFromZK();
     // update the health status for those expired partitions on instances
     Map<String, List<String>> expiredPartitionsByInstance =
```
dasahcc: I think we need to move this part as well. This is also part of read operation. We prefer to encapsulate the read logic (ZK read + direct API call) to one place.

i3wangyi: yeah, I thought about it as well. Ideally yes. The problem is I need to pass restconfig/baseurl/instances, a lot of params to the helix data accessor wrapper

dasahcc: I dont think we need many. The baseurl comes from RESTConfig. So the only needed params are RESTConfig and list of instances.

Also you dont have to pass it in when construct the data accessor wrapper. You can make it as function inpit in getPartitionHealth(...).

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
```diff
@@ -1,24 +1,158 @@
...
+    _restClient = CustomRestClientFactory.get();
+  }
+
+  HelixDataAccessorWrapper(ZKHelixDataAccessor dataAccessor, CustomRestClient client) {
+    super(dataAccessor);
+    _restClient = client;
+  }
+
+  public List<ExternalView> getExternalViews() {
+    return getChildNames(keyBuilder().externalViews()).stream()
```
dasahcc: You dont have to read it in that way. This causes more read request to ZK. It should be getChildValues(keyBuikder().externalViews()). If you would like to memorize it memorize it in a different way.

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
```diff
@@ -1,24 +1,158 @@
...
+    Map<String, Future<Map<String, Boolean>>> parallelTasks = new HashMap<>();
+    for (int i = 0; i < liveInstances.size(); i++) {
+      String liveInstance = liveInstances.get(i);
+      Optional<ZNRecord> maybeHealthRecord =
+          Optional.ofNullable(zkHealthReports.get(i)).map(HelixProperty::getRecord);
+      parallelTasks.put(liveInstance, POOL.submit(() -> {
+        if (maybeHealthRecord.isPresent()) {
+          return getPartitionsHealth(liveInstance, maybeHealthRecord.get(), restConfig,
+              customPayLoads);
+        } else {
```
dasahcc: else clause is not needed.

i3wangyi: It's different than the previous implementation. The private method `getPartitionsHealth` only submits queries when it's expired. The else case deals with when we don't have any record for the instance, it needs to submit a query of empty partitions list.

dasahcc: I am not saying totally remove the logic but just coding style. Since you change it in java 8, I will review it again.

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
```diff
@@ -1,24 +1,144 @@
...
+    // Only checks the instances are online with valid reports
+    List<String> liveInstances = getChildNames(keyBuilder().liveInstances());
+    // Make a parallel batch call for getting all healthreports from ZK.
+    List<HelixProperty> zkHealthReports = getProperty(liveInstances.stream()
+        .map(instance -> keyBuilder().healthReport(instance, PARTITION_HEALTH_KEY))
+        .collect(Collectors.toList()), false);
+    Map<String, Future<Map<String, Boolean>>> parallelTasks = new HashMap<>();
+    for (int i = 0; i < liveInstances.size(); i++) {
+      String liveInstance = liveInstances.get(i);
+      Optional<ZNRecord> maybeHealthRecord = zkHealthReports != null && zkHealthReports.size() > i
```
dasahcc: How does zkHealthReports.size() > i work? I am confused here. Was that the trick you did as .collect(Collectors.toList()), false)

If that's the case, can you add some comments here? It is hard to understand the code in this way.

i3wangyi: The **getProperty** method alone is deprecated, instead I used `getProperty(xxx, boolean throwException)`.
`List<HelixProperty> zkHealthReports` doesn't guarantee or doesn't explicitly say that it will always be of equal size of input parameters. E.g, I'd like to get 10 instances' healthReports, but the method only returns 5 objects. That's why I checked `zkHealthReports.size() > i` to avoid array index of bounds issue.
To be extra safe, 
It's possible the `zkHealthReports' is null, 'zkHealthReports' has less items than live instances parameter and each object in `zkHealthReports' is also possible null. All of these are covered in the ternary operator statement.

dasahcc: If Helix does not return you complete list, this is not an efficient way. You dont know which one is missed from the array list. Then you totally mess up the result. For example, you have 6 instances and you only get 3 ZNRecords for last 3 ones. With your logic, you will fill the 3 ZNRecords as first threes health report and querying next three ones. 

Please refer the Java doc of get in BaseDataAccessor interface. getProperty uses the get() to return list of HelixProperty: https://github.com/apache/helix/blob/master/helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java#L150

i3wangyi: 'getProperty' method was not a very good design but it's already there. I may have done redundant checks to be extra safe, it won't break the functionality. If I'm asking 6 instances' records, it returns {null, record1, null, null, record, null}, it will also work.

i3wangyi: I've changed the line to 
```
Optional<ZNRecord> maybeHealthRecord =
          Optional.ofNullable(zkHealthReports.get(i)).map(HelixProperty::getRecord);
```
If we agree the list will always return the same number of items as request parameters

i3wangyi:
The PR is ready to merge into master, approved by @dasahcc

Issue #421, opened 2019-08-19
Create definitions for soft/hard constraints and implementations


i3wangyi:
For Waged Rebalancer

PR #422, opened 2019-08-19
Redefine the hard/soft constraints

i3wangyi:
### Issues

- [ ] My PR addresses the following Helix issues and references them in the PR title:

(#421)

### Description

- Re-define the interfaces of hard/soft constraints
- A central place to keep relative weights of soft constraints

### Tests

### Commits

- [ ] My commits all reference appropriate Apache Helix GitHub issues in their subject lines, and I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":

### Documentation

### Code Quality

- [ ] My diff has been formatted using helix-style.xml


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraintWeightModel.java
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * Given the calculated scores map by soft constraints, get the sum of scores
+   * @param originScoresMap The origin scores of soft constraints
+   * @return The sum of double type
+   */
+  public double sumOfScores(Map<SoftConstraint, Float> originScoresMap) {
+    return originScoresMap.entrySet().stream().map(constraintToOriginScore -> {
+      float score = MIN_MAX_SCALER.scale(constraintToOriginScore.getValue());
+      float weight = MODEL.getOrDefault(constraintToOriginScore.getKey().getType(), 0f);
```
jiajunwang: How to ensure the input map includes all the constraint evaluation result? Shall we throw an exception if not?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -27,23 +27,31 @@
...
-   * The scoring function returns a score between MINIMAL_SCORE and MAXIMUM_SCORE, which is then weighted by the
-   * individual normalized constraint weights.
-   * Each individual constraint will define the meaning of MINIMAL_SCORE to MAXIMUM_SCORE differently.
-   */
-  float assignmentScore(AssignableNode node, AssignableReplica rep, ClusterContext clusterContext);
+  public SoftConstraint(Type type) {
+    _type = type;
+  }
+
+  enum Type {
```
jiajunwang: If we have 2 constraints with the same type defined, how to separately weight them?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -27,23 +27,31 @@
...
+public abstract class SoftConstraint {
+  private Type _type;
 
-  /**
-   * The scoring function returns a score between MINIMAL_SCORE and MAXIMUM_SCORE, which is then weighted by the
-   * individual normalized constraint weights.
-   * Each individual constraint will define the meaning of MINIMAL_SCORE to MAXIMUM_SCORE differently.
-   */
-  float assignmentScore(AssignableNode node, AssignableReplica rep, ClusterContext clusterContext);
+  public SoftConstraint(Type type) {
```
jiajunwang: Usually, we don't define a constructor for an abstract class. At least should make it protected.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -27,23 +27,31 @@
...
+    LEAST_USED_NODE,
+    LEAST_PARTITION_COUNT,
+    LEAST_MOVEMENTS
+  }
 
   /**
-   * Set the importance factor of the soft constraint.
-   * The more important it is, the more contribution it will make to the final evaluation.
-   * @param importance
+   * The scoring function returns a score between MINIMAL_SCORE and MAXIMUM_SCORE, which is then
```
jiajunwang: Where is the definition of MINIMAL_SCORE and MAXIMUM_SCORE? What will be the exception if this method returns any value out of the scope?
Please note in the comment.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -27,23 +27,31 @@
...
+   * Each individual constraint will define the meaning of MINIMAL_SCORE to MAXIMUM_SCORE
+   * differently.
+   * @return float value representing the score
    */
-  void setConstraintImportance(float importance);
+  public abstract float getAssignmentScore(AssignableNode node, AssignableReplica rep,
+      ClusterContext clusterContext);
 
-  float getConstraintImportance();
+  public Type getType() {
```
jiajunwang: protected shall be enough?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/HardConstraint.java
```diff
@@ -37,14 +39,23 @@
...
    */
-  boolean isAssignmentValid(AssignableNode node, AssignableReplica rep,
+  public abstract boolean isAssignmentValid(AssignableNode node, AssignableReplica replica,
       ClusterContext clusterContext);
 
-  /**
-   * @return Detail of the reason that the proposed assignment was rejected.
-   */
-  FailureReason getFailureReason();
+  public Type getType() {
```
jiajunwang: protected shall be enough?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/HardConstraint.java
```diff
@@ -37,14 +39,23 @@
...
+    return _type.toString();
+  }
+
   /**
-   * @return True if the proposed assignment is valid.
+   * Check if the replica could be assigned to the node
+   * @return True if the proposed assignment is valid; False otherwise
    */
-  boolean isAssignmentValid(AssignableNode node, AssignableReplica rep,
+  public abstract boolean isAssignmentValid(AssignableNode node, AssignableReplica replica,
```
jiajunwang: Why we need it to be public here? Can we just put the algorithm implementation class in the same package?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/HardConstraint.java
```diff
@@ -27,8 +27,10 @@
  * Evaluate a partition allocation proposal and return YES or NO based on the cluster context.
  * Any proposal fails one or more hard constraints will be rejected.
  */
-public interface HardConstraint {
-  enum FailureReason {
+public abstract class HardConstraint {
+  private Type _type;
```
jiajunwang: final

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/HardConstraint.java
```diff
@@ -37,14 +39,23 @@
     NODES_CONTAIN_SAME_PARTITION,
   }
 
+  public HardConstraint(Type failureReason) {
+    _type = failureReason;
+  }
+
+  public String getFailureReason() {
+    // by default just return the type, extend the method if more details are needed
+    return _type.toString();
```
jiajunwang: It might depends on the implementation logic.
We can have a default method with the basic message like this. So let's comment that a child class shall override this method if necessary.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/HardConstraint.java
```diff
@@ -37,14 +39,23 @@
     NODES_CONTAIN_SAME_PARTITION,
   }
 
+  public HardConstraint(Type failureReason) {
```
jiajunwang: protected?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/HardConstraint.java
```diff
@@ -27,8 +27,10 @@
  * Evaluate a partition allocation proposal and return YES or NO based on the cluster context.
  * Any proposal fails one or more hard constraints will be rejected.
  */
-public interface HardConstraint {
-  enum FailureReason {
+public abstract class HardConstraint {
+  private Type _type;
+
+  public enum Type {
```
jiajunwang: Can we also make it protected? Not sure, just feel it would be cleaner.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -27,23 +27,31 @@
  * Evaluate a partition allocation proposal and return a score within the normalized range.
  * A higher score means the proposal is more preferred.
  */
-public interface SoftConstraint {
-  float MIN_SCORE = -1000.0f;
-  float MAX_SCORE = 1000.0f;
+public abstract class SoftConstraint {
+  private Type _type;
```
jiajunwang: final

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraintWeightModel.java
i3wangyi: The MODEL is defined statically and not set dynamically on the fly. I think we should assume all types have a valid value or we're doing a bad job when assigning the weight.  Throw exception will fail all the rest operations I don't think it's good idea

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
i3wangyi: Type and constraint need to be 1 to 1 mapping. How can we have one type mapping to multiple constraints?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
i3wangyi: A default constructor in abstract class is a restriction that all the subclasses have to extend and set the type field. This way subclass won't be too casual without type.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/HardConstraint.java
i3wangyi: package-private makes more sense

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
i3wangyi: Just a comment wouldn't stop us from implementing any constraint at their will. A stronger constraint would be to allow us to implement any ranged soft constraints and the scaler scales the value to a reasonable range plus the weight. 

The MIN MAX scores are just copied from previous soft constraints' interfaces. How about just use Float.MIN_VALUE and Float.MAX_VALUE?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraintWeightModel.java
```diff
@@ -0,0 +1,47 @@
...
+
+  private ScoreScaler MIN_MAX_SCALER =
+      originScore -> (originScore - MIN_SCORE) / (MAX_SCORE - MIN_SCORE);
+
+  /**
+   * Given the calculated scores map by soft constraints, get the sum of scores
+   * @param originScoresMap The origin scores of soft constraints
+   * @return The sum of double type
+   */
+  double getSumOfScores(Map<SoftConstraint, Float> originScoresMap) {
```
dasahcc: Why we have inconsistent data type? If we think float is more efficient, let's use float as return type. It is mixing up double/float in the code.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraintWeightModel.java
```diff
@@ -0,0 +1,47 @@
+package org.apache.helix.controller.rebalancer.waged.constraints;
```
dasahcc: Have you set up the apache license header in the intellij? Once you set it up, it should be automatically pumped up.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
jiajunwang: I agree with 1:1 mapping. But this implementation is not protecting this. If it has to be one one mapping, shall we just take the Constraint class as the type?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
jiajunwang: That will make the scale function difficult to implement. We need a clear definition.
For now, comments or static vars should be enough. No one can really extend the constraint and deploy except for ourselves.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraintWeightModel.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Method to scale the origin score to a normalized range
+     * @param originScore The origin score of a range
+     * @return The normalized value between 0 - 1
+     */
+    float scale(float originScore);
+  }
+
+  private ScoreScaler MIN_MAX_SCALER =
+      originScore -> (originScore - MIN_SCORE) / (MAX_SCORE - MIN_SCORE);
```
jiajunwang: I think Float.MAX_VALUE - Float.MIN_VALUE will be overflow... At least it won't fit in a float var.
Please keep the original scale unless you have a good reason to change it.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraintWeightModel.java
jiajunwang: +1

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraintWeightModel.java
```diff
@@ -0,0 +1,50 @@
...
+   * Get the sum of normalized scores, given calculated scores map of soft constraints
+   * @param originScoresMap The origin scores of soft constraints
+   * @return The sum of double type
+   */
+  double getSumOfScores(Map<SoftConstraint, Float> originScoresMap) {
+    float sum = 0;
+    for (Map.Entry<SoftConstraint, Float> softConstraintScoreEntry : originScoresMap.entrySet()) {
+      float score = MIN_MAX_SCALER.scale(softConstraintScoreEntry.getValue());
+      float weight = MODEL.get(softConstraintScoreEntry.getKey().getType());
+      sum += score * weight;
```
jiajunwang: If you set the scale using MAX_VALUE or MIN_VALUE, you will need to be careful about overflow everywhere.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
i3wangyi: How about adding one more abstract method in SoftConstraint class. Child class either creates their own or inherit the base scale function from the abstract base class.
`ScaleFunction getScaleFunction();` 
Then the `WeightModel` class will basically
```
foreach softconstraint s:
     score = s.getScaleFunction().apply(s.getscore());
```
It's cleaner IMO and allow each child class to customize own its own.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
i3wangyi: If we use the class object then I guess the enum type definition is meaningless.

jiajunwang: Yeah, that's my question here. Do we really need the enum here, or not?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
jiajunwang: It is clean, but not addressing all problems.

If 2 constraints return with completely different scale scores (0-1 and 0-10000 for example), we expect the scale function does the magic and makes these scores comparable. But it might be difficult. So MAYBE, an easier way is just to comment and let the developers of the constraints to do the right thing.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
i3wangyi: We can go that route (no enum) but the enum was created in the first place, I do think it's cleaner. Back to the original simple interface design, we could also get away with enums as well since it doesn't also enforce 1:1 mapping

Then the weight map would consist of 
```
Map<SoftConstraint, Float> weights = 
 ImmutableMap.of(LeastUsedNode.class, 1.0f);
```
or 
```
Map<SoftConstraint, Float> weights = 
 ImmutableMap.of(instance of LeastUsedNode.class, 1.0f);
```

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
i3wangyi: Let the developers aware of the scaler function at the interface methods is usually better enforcement than asking for people to reference the comments.

jiajunwang: Agreed. In that case, can we enforce the input in the scale function? So we have some control on the evaluated score.
Anyway, please try to implement your proposal, we can discuss with the code.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/LeastPartitionCountConstraint.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.helix.controller.rebalancer.waged.model.AssignableNode;
+import org.apache.helix.controller.rebalancer.waged.model.AssignableReplica;
+import org.apache.helix.controller.rebalancer.waged.model.ClusterContext;
+
+/**
+ * Evaluate the proposed assignment according to the instance's partition count.
+ */
+class LeastPartitionCountConstraint extends SoftConstraint {
+  static LeastPartitionCountConstraint INSTANCE  = new LeastPartitionCountConstraint();
```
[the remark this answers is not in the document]
i3wangyi: Singleton instance has a better performance of GC. The downside is we lost the ability to customize the constraint of different score ranges, but I don't think we would need the requirement. I don't think there's any concurrent issue related to the class if it's singleton, can you specify?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/LeastPartitionCountConstraint.java
```diff
@@ -0,0 +1,55 @@
...
+
+  private LeastPartitionCountConstraint() {
+  }
+
+  /**
+   * Returns a score depending on the number of assignments on this node. The score is scaled evenly
+   * between the
+   * MIN_SCORE and the MAX_SCORE.
```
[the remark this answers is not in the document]
i3wangyi: sure, will do

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/LeastPartitionCountConstraint.java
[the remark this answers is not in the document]
jiajunwang: By concurrent issue, I mean if you need to keep some internal variable as a field in the constraint object, then we will need to do concurrent control. Because we might have multiple callers. I'm not saying it would be the case, just a thought.

Actually, the score range is a good point. We should have a consistent range. Given that saying, singleton could help.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraintWeightModel.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * The class retrieves the offline model that defines the relative importance of soft constraints.
+ */
+class SoftConstraintWeightModel {
+  private static Map<? extends SoftConstraint, Float> MODEL;
+
+  static {
+    // TODO either define the weights in property files or zookeeper node or static human input
```
jiajunwang: Just a reminder, we will need to be able to update the MODEL at runtime. For example, change the importance based on historical data without restarting the application.

I assume the rebalancer will pass some parameter to the algorithm instance. Or the instance will be able to load the update before calculation.

Anyway, I know this part is not done yet, but we need to implement it during the integration.

PR #426, opened 2019-08-21
Add BucketDataAccessor for large writes

### LINKED ISSUE #419: Add BucketDataAccessor to persist assignment metadata
For the new rebalancer (WAGED), we need to have a way to persist potentially large amounts of data (assignment metadata for the entire cluster) in ZooKeeper. This issue tracks the work of implementing a data accessor. See https://github.com/apache/helix/wiki/Storing-Assignment-Metadata-in-ZooKeeper

narendly:
### Issues

Resolves #419 

### Description

For the new WAGED rebalancer, it's necessary to have a data accessor that will allow writes of data exceeding 1MB to write assignment metadata. ZooKeeper's ZNode size is capped at 1MB, so BucketDataAccessor interface and ZkBucketDataAccessor help us achieve this.
Changelist:
1. Add BucketDataAccessor and ZkBucketDataAccessor
2. Add necessary serializers
3. Add an integration test against ZK

### Tests

**TestZkBucketDataAccessor**

[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.049 s - in org.apache.helix.manager.zk.TestZkBucketDataAccessor
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  9.498 s
[INFO] Finished at: 2019-09-06T16:14:41-07:00

### Commits

### Documentation

https://github.com/apache/helix/wiki/Storing-Assignment-Metadata-in-ZooKeeper

### Code Quality


helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.IOException;
+
+
+public interface BucketDataAccessor {
+
+  <T extends HelixProperty> boolean bucketWrite(String path, T value);
+
+  HelixProperty bucketRead(String path);
+
+  <T extends HelixProperty> boolean compressedBucketWrite(String path, T value) throws IOException;
```
jiajunwang: Question, why will a caller don't want the compress? With the buckets, the data won't be easy to read by a human anyway, right?

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
```diff
@@ -0,0 +1,35 @@
...
+
+  HelixProperty compressedBucketRead(String path);
+
+  void setBucketSize(int size);
```
jiajunwang: Shall we just final this var in the constructor?
Allow updating might increase the complicity.

helix-core/src/main/java/org/apache/helix/manager/zk/DummySerializer.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * An implementation of ZKSerializer that does nothing but pass the argument through. To be used
+ * with {@link ZkBucketDataAccessor}.
+ */
+public class DummySerializer implements ZkSerializer {
+  @Override
+  public byte[] serialize(Object data) throws ZkMarshallingError {
+    if (data instanceof byte[]) {
+      return (byte[]) data;
```
jiajunwang: Usually, we don't serialize an object like this. Better to implement the serializable interface.
The reason is that if you update the class in a new version, the serialized data you persisted in the older versions will be unreadable.
However, with the serializable interface methods, you have a way to address different versions gracefully.

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * ZNRecordMsgPackSerializer serializes ZNRecord objects into a byte array using MessagePack's
+ * serializer. Note that this serializer doesn't check for the size of the resulting binary.
+ */
+public class ZNRecordJacksonSerializer implements ZkSerializer {
+  private static Logger LOG = LoggerFactory.getLogger(ZNRecordJacksonSerializer.class);
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  @Override
+  public byte[] serialize(Object record) throws ZkMarshallingError {
```
jiajunwang: I see you already convert all exceptions to HelixException, do we still need to claim ZkMarshallingError here?

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
```diff
@@ -0,0 +1,71 @@
...
+    if (!(record instanceof ZNRecord)) {
```
jiajunwang: Check for null?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,178 @@
...
+    throw new UnsupportedOperationException("bucketWrite not supported!");
+  }
+
+  @Override
+  public HelixProperty bucketRead(String path) {
+    throw new UnsupportedOperationException("bucketRead not supported!");
+  }
+
+  @Override
+  public <T extends HelixProperty> boolean compressedBucketWrite(String path, T value)
```
jiajunwang: Both read and write involve multiple nodes, how we ensure the consistency?
Especially we might have a read happens in the middle of a write.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,178 @@
...
+      }
+    }
+    return true;
+  }
+
+  @Override
+  public HelixProperty compressedBucketRead(String path) {
+    // Retrieve the metadata
+    Map metadataMap;
+    byte[] metadata = (byte[]) _zkBaseDataAccessor.get(path, null, AccessOption.PERSISTENT);
```
jiajunwang: Will it be easier if the metadata is a ZNode? So we will have the accessor to read directly. No need to "OBJECT_MAPPER.readValue()".

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
i3wangyi: If it's an abstract class, the field can be of a field. Not a good idea for external to update at will

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
```diff
@@ -0,0 +1,35 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.io.IOException;
+
+
+public interface BucketDataAccessor {
```
i3wangyi: Can you add more detailed comments/javadocs in the interface class/methods?

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
```diff
@@ -0,0 +1,35 @@
...
+
+  <T extends HelixProperty> boolean bucketWrite(String path, T value);
```
i3wangyi: The write operation will also throw IOException as well, right?

One more, why not combine the 2 write operations together.
`<T extends HelixProperty> boolean bucketWrite(String path, T value, boolean isCompressEnabled) throws IOException;`

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
```diff
@@ -0,0 +1,35 @@
...
+
+  HelixProperty bucketRead(String path);
```
i3wangyi: `<T extends HelixProperty> T bucketRead(String path, Class<T> detailedPropertyType)` this way help you cast the result to the subclass without IDE warning.

helix-core/src/main/java/org/apache/helix/manager/zk/DummySerializer.java
```diff
@@ -0,0 +1,44 @@
...
+import org.I0Itec.zkclient.exception.ZkMarshallingError;
+import org.I0Itec.zkclient.serialize.ZkSerializer;
+import org.apache.helix.HelixException;
+
+
+/**
+ * An implementation of ZKSerializer that does nothing but pass the argument through. To be used
+ * with {@link ZkBucketDataAccessor}.
+ */
+public class DummySerializer implements ZkSerializer {
```
i3wangyi: What's the plan of the dummy implementation, do you plan to check it in as well?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,178 @@
...
+import org.apache.helix.ZNRecord;
+import org.apache.helix.manager.zk.client.DedicatedZkClientFactory;
+import org.apache.helix.manager.zk.client.HelixZkClient;
+import org.apache.helix.util.GZipCompressionUtil;
+import org.codehaus.jackson.map.ObjectMapper;
+
+public class ZkBucketDataAccessor implements BucketDataAccessor {
+
+  private final static ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static final String BUCKET_SIZE_KEY = "BUCKET_SIZE";
```
i3wangyi: nit: consistent with private static final ordering.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,178 @@
...
+  private static final String DATA_SIZE_KEY = "DATA_SIZE";
+
+  // 100 KB for default bucket size
+  private static final int DEFAULT_BUCKET_SIZE = 100 * 1024;
+  private int _bucketSize = DEFAULT_BUCKET_SIZE;
+  private ZkSerializer _zkSerializer;
+  private BaseDataAccessor _zkBaseDataAccessor;
+
+  public ZkBucketDataAccessor(String zkAddr) {
```
i3wangyi: I don't like the idea of doing more than necessary initialization works at the constructor level. It should be whoever uses the class' concerns. 

A simple setter like
```
public ZkBucketDataAccessor(ZkSerializer zkSerializer, BaseDataAccessor baseDataAccessor) {
   _zkSerializer = zkSerializer;
   _zkBaseDataAccessor = zkBaseDataAccessor;
}
```

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,178 @@
...
+
+public class ZkBucketDataAccessor implements BucketDataAccessor {
+
+  private final static ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static final String BUCKET_SIZE_KEY = "BUCKET_SIZE";
+  private static final String DATA_SIZE_KEY = "DATA_SIZE";
+
+  // 100 KB for default bucket size
+  private static final int DEFAULT_BUCKET_SIZE = 100 * 1024;
+  private int _bucketSize = DEFAULT_BUCKET_SIZE;
```
i3wangyi: Why a member variable _bucketSize when you already defined the static variable?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,178 @@
...
+      } else {
+        buckets.add(Arrays.copyOfRange(compressedRecord, ptr, ptr + _bucketSize));
+      }
+      ptr += _bucketSize;
+      counter++;
+    }
+
+    // Do an async set to ZK
+    boolean[] success;
+    success = _zkBaseDataAccessor.setChildren(paths, buckets, AccessOption.PERSISTENT);
```
i3wangyi: A very general practice people used when bulk writing is "retry", define some thresholds of MAX_RETRY_TIMES, it will make the bulk write more robust( when one write fails, try again)

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
```diff
@@ -0,0 +1,71 @@
...
+import org.I0Itec.zkclient.exception.ZkMarshallingError;
+import org.I0Itec.zkclient.serialize.ZkSerializer;
+import org.apache.helix.HelixException;
+import org.apache.helix.ZNRecord;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * ZNRecordMsgPackSerializer serializes ZNRecord objects into a byte array using MessagePack's
```
i3wangyi: The comment is different from the class name

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
```diff
@@ -0,0 +1,71 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import org.I0Itec.zkclient.exception.ZkMarshallingError;
+import org.I0Itec.zkclient.serialize.ZkSerializer;
+import org.apache.helix.HelixException;
+import org.apache.helix.ZNRecord;
+import org.codehaus.jackson.map.ObjectMapper;
```
i3wangyi: How performance is the objectmapper, and AFAIK `codehaus` is a deprecated package

narendly:
I will create a new PR since we aren't able to rebase.

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
narendly: I removed unsupported interface methods.

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
narendly: I agree. To make this simple, I will let the user define it in the constructor.

helix-core/src/main/java/org/apache/helix/manager/zk/DummySerializer.java
narendly: If I'm not misunderstanding your comment, it is meant to do nothing (as described in JavaDoc). 

This is because HelixZkClient requires some custom serializer (that implements ZkSerializer), but in this case we do not want any serialization because we serialize it in BucketDataAccessor before passing it into HelixZkClient.

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
narendly: ZkSerializer mandates it.

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
narendly: Please read: https://stackoverflow.com/questions/2950319/is-null-check-needed-before-calling-instanceof

Also I literally say in the comment below instanceof that null check is not needed...

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: I will add a field in the metadata. Will use that field to simulate some sort of reader-writer lock. Stay tuned.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: I agree that serializing the metadata might be an overkill. Will fix.

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
narendly: Added.

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
narendly: Removed bucketWrite instead.

helix-core/src/main/java/org/apache/helix/manager/zk/DummySerializer.java
narendly: See the comment above - there was a similar question.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: Fixed. Good catch

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
narendly: There aren't any IDE warnings... but for the sake of consistency (to make it consistent with write), I could also make the read method generic.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: I had something different in mind. My idea was make this ZkBucketDataAccessor's implementation details (such as how we serialize or what data accessor we use) hidden from the caller. This way we simplify a lot of code on the caller side and ZkBucketDataAccessor is free to use whichever method it wants to persist the data in ZK.

To put it another way, the rebalancer shouldn't have to worry about how the data is serialized and how we are accessing ZK.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: That was meant to be the default value, and there was a setter. But to keep thing simple, we will only allow the user to set the bucket size in the constructor.

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
narendly: Good catch!

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
narendly: Not great. I initially used the better-maintained half (fasterxml jackson) with MessagePack serialization, but hit a few issues in terms of backward compatibility and problems around deserializing. 

So I decided to just make it work first and optimize serialization later.

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
i3wangyi: Sorry I need to clarify more in more detail. What I've noticed in Helix is that the `String path` could point to a specific HelixProperty(IdealState, e.g). Then think about the behavior of the caller, it needs to 
```
IdealState idealState = (IdealState) accessor.bucketRead(IdealStatePath);
```
It'll of course trigger IDE warning (not safe). Casting at the method level will help you remove the warning.

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
i3wangyi: Cool, thanks for the explanation.

jiajunwang:
@narendly I had some discussion with Lei. We think it would be better if we can organize the bucket according to the resource information. In detail, the bucket may contain multiple resource assignments, but a single bucket can always be decompressed separately. This is different from the current implementation I believe. For now, we have to read all files to extract. So debugging would be difficult.

It won't change the interface. Just enhancing the znode structure and hopefully making SRE/Dev support easier. 

Note that in this case, we can put more information into the metadata file. For example, which bucket contains which resource's information.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: A retry like that is actually in the async sync logic (you could take a look at your leisure). I agree that the semantics could be improved here (I've left it as future work to polish it up with a TODO), but the bulk of what you're describing is already there.

narendly:
@jiajunwang 
I have some concerns about that approach as well. We already discussed that alternative during the design stage. To reiterate the points that were raised then:

1. It is more difficult to tackle the issue of extreme cases like a resource having too many partitions and too many resources having few partitions. This makes it harder to have a consistent bucketing policy.
2. You cited the benefit of aligning the bucketing policy and Helix's resources (resource/partition) as the ease of debugging. But we could always provide a tool for viewing. So that's not a blocker with the current approach.

Let's continue the discussion offline.

jiajunwang:
> @jiajunwang
> I have some concerns about that approach as well. We already discussed that alternative during the design stage. To reiterate the points that were raised then:
> 
> 1. It is more difficult to tackle the issue of extreme cases like a resource having too many partitions and too many resources having few partitions. This makes it harder to have a consistent bucketing policy.
> 2. You cited the benefit of aligning the bucketing policy and Helix's resources (resource/partition) as the ease of debugging. But we could always provide a tool for viewing. So that's not a blocker with the current approach.
> 
> Let's continue the discussion offline.

I personally have no strong preference for this. But I think there is a difference between requiring a different tool to debug and using the system provided unzip command to debug directly. On my understanding, this is more about convenience.

Shall we have a short meeting to discuss it? These two designs are not compatible. So let's confirm before proceeding. Please include Lei.

helix-core/src/main/java/org/apache/helix/manager/zk/DummySerializer.java
i3wangyi: If you have to instantiate a dummy serializer that does nothing in ZkBucketDataAccessor, I would suggest using an anonymous class .

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * Constructor that allows a custom bucket size.
+   * @param zkAddr
+   * @param bucketSize
+   */
+  public ZkBucketDataAccessor(String zkAddr, int bucketSize) {
+    HelixZkClient zkClient = DedicatedZkClientFactory.getInstance()
+        .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddr));
+    zkClient.setZkSerializer(new DummySerializer());
```
i3wangyi: Use anonymous class please if you have to initialize, 
```
zkClient.setZkSerialize(new ZKSerializer() {
   @override
  ....
});
```
A dummy public class means every developer can use it, and they will question why is it like this.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,183 @@
...
+  private ZkSerializer _zkSerializer;
+  private BaseDataAccessor _zkBaseDataAccessor;
+
+  /**
+   * Constructor that allows a custom bucket size.
+   * @param zkAddr
+   * @param bucketSize
+   */
+  public ZkBucketDataAccessor(String zkAddr, int bucketSize) {
+    HelixZkClient zkClient = DedicatedZkClientFactory.getInstance()
```
i3wangyi: I think the caller at least needs to pass the ZKClient instance. Then here you don't need to create one from the beginning.

I've seen you're relying on an integration test that directly writes/reads from real ZK data store, but I'd also suggest another constructor `new ZkBucketDataAccessor(mockjSerializer, mockBaseDataAccessor)` that sets the zkSerializer and baseDataAccesor directly, where you can mock any behaviors of the dependencies (e.g `when(mockBaseDataAccessor.setChildren()).thenThrowException`)

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,183 @@
...
+      } else {
+        buckets.add(Arrays.copyOfRange(compressedRecord, ptr, ptr + _bucketSize));
+      }
+      ptr += _bucketSize;
+      counter++;
+    }
+
+    // Do an async set to ZK
+    boolean[] success;
+    success = _zkBaseDataAccessor.setChildren(paths, buckets, AccessOption.PERSISTENT);
```
i3wangyi: why not 
```
boolean[] success = xxx
```
in one line?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,183 @@
...
+
+    // Return false if any of the writes failed
+    // TODO: Improve the failure handling
+    for (boolean s : success) {
+      if (!s) {
```
i3wangyi: Do you want to log which path fails to write? Also, what's the behavior if partial writes go through but some failed (impact on ZK datastore)?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,183 @@
...
+
+    byte[] compressedRecord = new byte[dataSize];
+
+    List<String> paths = new ArrayList<>();
+    for (int i = 0; i < numBuckets; i++) {
+      paths.add(path + "/" + i);
+    }
+
+    // Async get
+    List buckets = _zkBaseDataAccessor.get(paths, null, AccessOption.PERSISTENT, true);
```
i3wangyi: Type of list, do you get IDE warning?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBucketDataAccessor.java
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.helix.TestHelper;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.common.ZkTestBase;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+public class TestZkBucketDataAccessor extends ZkTestBase {
+
+  private String PATH = "/" + TestHelper.getTestClassName();
```
i3wangyi: nit:  use all upper case only for `private static final` fields.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBucketDataAccessor.java
```diff
@@ -0,0 +1,126 @@
...
+  private Map<String, String> MAP_FIELD;
+
+  private BucketDataAccessor _bucketDataAccessor;
+
+  @BeforeClass
+  public void beforeClass() {
+    _bucketDataAccessor = new ZkBucketDataAccessor(ZK_ADDR);
+
+    // Populate LIST_FIELD for R/W comparison
+    LIST_FIELD = new ArrayList<>();
```
i3wangyi: Can't you get rid of the beforeClass method and just set it inline?

For example,

```
private static final List<String> LIST_FIELDS = ImmutableList.of("1", "2");
private static final Map<String, String> MAP_FIELD = ImmutableMap.of("1", "2");
```

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -87,19 +90,30 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
    */
   public ZkBucketDataAccessor(String zkAddr) {
-    this(zkAddr, DEFAULT_BUCKET_SIZE);
+    this(zkAddr, DEFAULT_BUCKET_SIZE, DEFAULT_NUM_VERSIONS);
   }
 
   @Override
   public <T extends HelixProperty> boolean compressedBucketWrite(String path, T value)
       throws IOException {
+    lock(path);
```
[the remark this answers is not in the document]
narendly: @i3wangyi Can you remove this duplicate comment?
It is true that you don't need to establish multiple ZkClient connections for the use case you're referring to. However, this is a data accessor for large writes for more general use. That is why we consider the possibility of multiple writes and reads. This was the original design; the data accessor is not singly tied to the WAGED rebalancer. 

For our WAGED rebalancer use case, there will be a component called AssignmentMetadataStore that maintains only one instance of BucketDataAccessor that will be used across pipeline runs.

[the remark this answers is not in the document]
narendly: @i3wangyi https://github.com/apache/helix/issues/444
Here is the issue for the aforementioned component. This will create and maintain one instance of `BucketDataAccessor`. Once we have `BucketDataAccessor` (which is more of a generic component that could be used in other places), I will implement `AssignmentMetadataStore`.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,277 @@
...
+
+    // Do an async set to ZK
+    boolean[] success = _zkBaseDataAccessor.setChildren(paths, buckets, AccessOption.PERSISTENT);
+
+    // Return false if any of the writes failed
+    // TODO: Improve the failure handling
+    for (boolean s : success) {
+      if (!s) {
+        // release the lock
+        _zkBaseDataAccessor.remove(path + "/" + WRITE_LOCK_KEY, AccessOption.EPHEMERAL);
```
jiajunwang: Why not call unlock() here?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,277 @@
...
+  /**
+   * Constructor that uses a default bucket size.
+   * @param zkAddr
+   */
+  public ZkBucketDataAccessor(String zkAddr) {
+    this(zkAddr, DEFAULT_BUCKET_SIZE, DEFAULT_NUM_VERSIONS);
+  }
+
+  @Override
+  public <T extends HelixProperty> boolean compressedBucketWrite(String path, T value)
```
jiajunwang: There is actually only one case you will return false. And in that case, the caller will need to retry. So how about just though Exception on any failure?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,277 @@
...
+import org.apache.helix.BaseDataAccessor;
+import org.apache.helix.BucketDataAccessor;
+import org.apache.helix.HelixException;
+import org.apache.helix.HelixProperty;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.manager.zk.client.DedicatedZkClientFactory;
+import org.apache.helix.manager.zk.client.HelixZkClient;
+import org.apache.helix.util.GZipCompressionUtil;
+
+public class ZkBucketDataAccessor implements BucketDataAccessor {
```
jiajunwang: Based on the code, it seems we don't have a "GC" logic for the buckets. That means, if one certain property becomes smaller (hence fewer buckets in the newer version), the extra buckets in a previous property version won't be removed.

I do have some concerns about this point.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,277 @@
...
+        return data;
+      }
+    });
+    _zkBaseDataAccessor = new ZkBaseDataAccessor(zkClient);
+
+    // TODO: Optimize serialization with Jackson
+    // TODO: Or use a better binary serialization protocol
+    // TODO: Consider making this also binary
+    // TODO: Consider an async write for the metadata as well
+    HelixZkClient znRecordClient = DedicatedZkClientFactory.getInstance()
```
jiajunwang: Shall we use SharedZkClient? Note that each shared client can have separate serializer even they share the same connection. Which fits the requirement here very well.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,277 @@
...
+        }
+        throw new HelixException("ZkBucketDataAccesor only supports a byte array as an argument!");
+      }
+
+      @Override
+      public Object deserialize(byte[] data) throws ZkMarshallingError {
+        return data;
+      }
+    });
+    _zkBaseDataAccessor = new ZkBaseDataAccessor(zkClient);
```
jiajunwang: This accessor will be used to create the lock node. Please see my comments below. There could be problem if the unlock call fails. I think we shall re-connect (or recreate) the client if the unlock fails. That will ensure the ephemeral node gone.

In the extreme case, if the re-connect fails, this will prevent the accessor to be used anymore. Which might be a good thing because the caller will escalate the issue much sooner.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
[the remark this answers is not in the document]
jiajunwang: I was thinking about this point, for the genericity of the accessor, I think a ZK backed lock is necessary. Eventually, we will need to evaluate the extra cost, of course.

One minor thing may help with the performance a little bit, how about applying the lock after bucket calculation?

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
```diff
@@ -0,0 +1,45 @@
...
+   * @param value HelixProperty to write
+   * @param <T>
+   * @return
+   * @throws IOException
+   */
+  <T extends HelixProperty> boolean compressedBucketWrite(String path, T value) throws IOException;
+
+  /**
+   * Read a (large) HelixProperty that was written in buckets, compressed. If the write lock has
+   * been taken, it waits a default period of 60 seconds until the lock has been released.
```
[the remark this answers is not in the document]
narendly: This comment is no longer valid. The way it's coded up now is it just fails if it's locked. I originally thought it would be beneficial to have some sort of timeout. Let's hold an offline discussion.

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public class ZNRecordJacksonSerializer implements ZkSerializer {
+  private static Logger LOG = LoggerFactory.getLogger(ZNRecordJacksonSerializer.class);
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  @Override
+  public byte[] serialize(Object record) throws ZkMarshallingError {
+    if (!(record instanceof ZNRecord)) {
+      // null is NOT an instance of any class
+      LOG.error(
```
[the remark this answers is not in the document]
narendly: I agree that there'd be duplicate notification. Will remove the error log.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: We could clean up prior to writing. Will add this. I'll add it and also create a test case.

helix-core/src/main/java/org/apache/helix/task/WorkflowDispatcher.java
```diff
@@ -116,6 +116,9 @@ public void updateWorkflowStatus(String workflow, WorkflowConfig workflowCfg,
       if (isWorkflowStopped(workflowCtx, workflowCfg)) {
         workflowCtx.setWorkflowState(TaskState.STOPPED);
         _clusterDataCache.updateWorkflowContext(workflow, workflowCtx);
+      } else {
```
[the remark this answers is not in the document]
narendly: Will remove :)

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
```diff
@@ -0,0 +1,45 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.io.IOException;
+
+public interface BucketDataAccessor {
+
+  /**
+   * Write a (large) HelixProperty in buckets, compressed.
```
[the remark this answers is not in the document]
narendly: I removed "(large)". That's correct. In theory, this doesn't care how big the data is.

helix-core/src/main/java/org/apache/helix/BucketDataAccessor.java
```diff
@@ -0,0 +1,45 @@
...
+   * @param path path to which the metadata will be written to
+   * @param value HelixProperty to write
+   * @param <T>
+   * @return
```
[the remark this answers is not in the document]
narendly: Removed

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: Added the cleanup logic and added a test case to verify that cleanup is happening.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: Changed to SharedZkClientFactory.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  @Override
+  public <T extends HelixProperty> HelixProperty compressedBucketRead(String path,
+      Class<T> helixPropertySubType) {
+    return helixPropertySubType.cast(compressedBucketRead(path));
+  }
+
+  private HelixProperty compressedBucketRead(String path) {
+    tryLock(path);
```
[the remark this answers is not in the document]
narendly: No longer valid. No waiting at the accessor level.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,277 @@
...
+   * Returns the last string element in a split String array by /.
+   * @param path
+   * @return
+   */
+  private String extractIdFromPath(String path) {
+    String[] splitPath = path.split("/");
+    return splitPath[splitPath.length - 1];
+  }
+
+  private void tryLock(String path) {
```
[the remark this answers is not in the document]
narendly: Sure. I could make it return a boolean.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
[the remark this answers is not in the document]
i3wangyi: @jiajunwang @narendly 
It starts to make sense when the implementation of lock will be used in a distributed environment where instances other than helix controller itself will coordinate and write in order. Think about this use case then becomes actually a distributed lock, is using distributed lock really needed for the use case? Will we ever expose the endpoint to other parties?

My in-memory lock works in a single controller environment where the critical path write operations are only possible for one instance. Then singleton pattern + key based lock will work just fine, it'll be more performant since this implementation relies on ZK, which at the end of the day, is an I/O operation/network request, etc.

Does this understanding make sense? Have we cleared about our use cases?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,277 @@
...
+      metadataRecord = new ZNRecord(extractIdFromPath(path));
+    }
+    int lastSuccessIndex = (metadataRecord.getIntField(LAST_SUCCESS_KEY, -1) + 1) % _numVersions;
+    String dataPath = path + "/" + lastSuccessIndex;
+
+    // Take the ZNrecord and serialize it (get byte[])
+    byte[] serializedRecord = _zkSerializer.serialize(value.getRecord());
+
+    // Compress the byte[]
+    byte[] compressedRecord = GZipCompressionUtil.compress(serializedRecord);
```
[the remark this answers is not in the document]
narendly: While I agree that the method is getting long, it doesn't seem to cleanly separate. Let's keep it this way for now and let us clean up later.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: Let's use the other comment.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -0,0 +1,277 @@
...
+    // Enter the critical section
+    if (!_zkBaseDataAccessor.set(path + "/" + WRITE_LOCK_KEY, new byte[0],
+        AccessOption.EPHEMERAL)) {
+      throw new HelixException(String.format("Could not set an ephemeral node for path: %s", path));
+    }
+  }
+
+  private void unlock(String path) {
+    // Write succeeded, so release the lock
+    if (!_zkBaseDataAccessor.remove(path + "/" + WRITE_LOCK_KEY, AccessOption.EPHEMERAL)) {
```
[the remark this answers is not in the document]
narendly: Is there a way to disconnect and reconnect the data accessor?

Alternatively, we could have something that keeps trying up to N times with backoff if you feel that's a better option.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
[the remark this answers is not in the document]
i3wangyi: We've synced up offline. We do have such a use case where a custom tool or other components will read the same path and introduce a race-condition situation. In other words, the use case is in a distributed environment. Given the lack of a re-usable distributed lock module, the current implementation is a quick and light-weight solution to get things working.

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
```diff
@@ -0,0 +1,70 @@
...
+  public byte[] serialize(Object record) throws ZkMarshallingError {
+    if (!(record instanceof ZNRecord)) {
+      // null is NOT an instance of any class
+      throw new HelixException("Input object is not of type ZNRecord (was " + record + ")");
+    }
+    ZNRecord znRecord = (ZNRecord) record;
+
+    try {
+      return OBJECT_MAPPER.writeValueAsBytes(znRecord);
+    } catch (Exception e) {
```
i3wangyi: Could you catch the exact exception, it's too general

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
```diff
@@ -0,0 +1,70 @@
...
+  public Object deserialize(byte[] bytes) throws ZkMarshallingError {
+    if (bytes == null || bytes.length == 0) {
+      // reading a parent/null node
+      return null;
+    }
+
+    ZNRecord record;
+    try {
+      record = OBJECT_MAPPER.readValue(bytes, ZNRecord.class);
+    } catch (Exception e) {
```
i3wangyi: same comment here

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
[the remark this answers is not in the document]
jiajunwang: For this, you need to track the zkclient which is used when we create the accessor. That might be heavy, but I think we'd better ensure safety first.

As we discussed offline, let's add TODO for the solution for now.

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
narendly: Changed to IOException

helix-core/src/main/java/org/apache/helix/manager/zk/ZNRecordJacksonSerializer.java
narendly: Done.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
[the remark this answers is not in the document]
narendly: Will put a todo. Two options:
1. Use ephemeral owner ID
2. Implement something that's has a concept of a "lease" with a timeout.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBucketDataAccessor.java
```diff
@@ -112,6 +118,7 @@ public void testManyWritesWithVersionCounts() throws IOException {
 
     // Clean up
     bucketDataAccessor.compressedBucketDelete(path);
+    bucketDataAccessor.disconnect();
```
jiajunwang: np, add a check for verifying ZkClient closed after disconnect?

narendly: zkclient.close is a synchronized call. Also we don't expose the zk clients at the accessor level so that is difficult to do and probably not necessary.

narendly:
This PR is ready to be merged, approved by @i3wangyi @jiajunwang .

Add BucketDataAccessor for large writes
    
    For the new WAGED rebalancer, it's necessary to have a data accessor that will allow writes of data exceeding 1MB. ZooKeeper's ZNode size is capped at 1MB, so BucketDataAccessor interface and ZkBucketDataAccessor help us achieve this.
    Changelist:
    1. Add BucketDataAccessor and ZkBucketDataAccessor
    2. Add necessary serializers
    3. Add an integration test against ZK

Issue #402, opened 2019-08-13
Implement the WAGED rebalancer


jiajunwang:
Based on the existing components (or the interfaces), we can now implement the initial version of the WAGED rebalancer.

AC:
- Implement the rebalancer with unit tests.
- Roughly integrate with the rebalance pipeline. The final integration will be done when all the components implementation is ready.

PR #440, opened 2019-08-29
Record the replica objects in the AssignableNode in addition to the partition name

jiajunwang:
### Issues

#402 

### Description

The replica instances are required while the rebalance algorithm generating ResourceAssignment based on the AssignableNode instances.
Refine the methods of the AssignableNode for better code style and readability.
Also modify the related test cases to verify state information and new methods.

### Tests

Modified and run the Cluster Model tests.

Not necessary before fully integrated with the pipeline

### Commits

### Documentation

NA

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
@@ -33,6 +33,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static java.lang.Math.max;
 
@@ -51,16 +52,19 @@ public class AssignableNode {
   private Map<String, Integer> _maxCapacity;
   private int _maxPartition; // maximum number of the partitions that can be assigned to the node.
 
-  // proposed assignment tracking
-  // <resource name, partition name set>
-  private Map<String, Set<String>> _currentAssignments;
-  // <resource name, top state partition name>
-  private Map<String, Set<String>> _currentTopStateAssignments;
-  // <capacity key, capacity value>
-  private Map<String, Integer> _currentCapacity;
+  // A map of <resource name, <partition name, replica>> that tracks the replicas assigned to the node.
+  private Map<String, Map<String, AssignableReplica>> _currentAssignedReplicaMap;
+  // A map of <capacity key, capacity value> that tracks the current available node capacity
+  private Map<String, Integer> _currentCapacityMap;
   // The maximum capacity utilization (0.0 - 1.0) across all the capacity categories.
   private float _highestCapacityUtilization;
 
+  /**
+   * @param clusterConfig
+   * @param instanceConfig
+   * @param instanceName
+   * @param existingAssignment A collection of replicas that have been pre-allocated to the node.
+   */
   AssignableNode(ClusterConfig clusterConfig, InstanceConfig instanceConfig, String instanceName,
       Collection<AssignableReplica> existingAssignment) {
     _instanceName = instanceName;
@@ -68,9 +72,8 @@ public class AssignableNode {
   }
 
   private void reset() {
-    _currentAssignments = new HashMap<>();
-    _currentTopStateAssignments = new HashMap<>();
-    _currentCapacity = new HashMap<>();
+    _currentAssignedReplicaMap = new HashMap<>();
+    _currentCapacityMap = new HashMap<>();
     _highestCapacityUtilization = 0;
   }
 
@@ -80,8 +83,8 @@ public class AssignableNode {
    * refreshed. This is under the assumption that the capacity mappings of InstanceConfig and ResourceConfig could
    * subject to change. If the assumption is no longer true, this function should become private.
    *
-   * @param clusterConfig  - the Cluster Config of the cluster where the node is located
-   * @param instanceConfig - the Instance Config of the node
+   * @param clusterConfig      - the Cluster Config of the cluster where the node is located
+   * @param instanceConfig     - the Instance Config of the node
    * @param existingAssignment - all the existing replicas that are current assigned to the node
    */
   private void refresh(ClusterConfig clusterConfig, InstanceConfig instanceConfig,
@@ -92,7 +95,7 @@ public class AssignableNode {
     if (instanceCapacity.isEmpty()) {
       instanceCapacity = clusterConfig.getDefaultInstanceCapacityMap();
     }
-    _currentCapacity.putAll(instanceCapacity);
+    _currentCapacityMap.putAll(instanceCapacity);
     _faultZone = computeFaultZone(clusterConfig, instanceConfig);
     _instanceTags = new HashSet<>(instanceConfig.getTags());
     _disabledPartitionsMap = instanceConfig.getDisabledPartitionsMap();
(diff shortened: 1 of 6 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -51,10 +51,9 @@
   private Map<String, Integer> _maxCapacity;
   private int _maxPartition; // maximum number of the partitions that can be assigned to the node.
 
-  // proposed assignment tracking
-  // <resource name, partition name set>
-  private Map<String, Set<String>> _currentAssignments;
-  // <resource name, top state partition name>
+  // A map of <resource name, <partition name, state>> to track all the proposed partition assignment with state.
+  private Map<String, Map<String, String>> _currentAssignments;
+  // An index map of <resource name, top state partition name> for tracking the top state partitions.
```
narendly: 1. What do you mean by an "index map"?
2. When you say <resource name, top state partition name>, what is the top state partition name? Each resource maps to a Set of Strings, and these strings are names of nodes that have top state partitions? Could you clarify?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -51,10 +51,9 @@
   private Map<String, Integer> _maxCapacity;
   private int _maxPartition; // maximum number of the partitions that can be assigned to the node.
 
-  // proposed assignment tracking
-  // <resource name, partition name set>
-  private Map<String, Set<String>> _currentAssignments;
-  // <resource name, top state partition name>
+  // A map of <resource name, <partition name, state>> to track all the proposed partition assignment with state.
```
narendly: partition name? Do you mean the instance's name?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -287,10 +281,21 @@ private void assignNewBatch(Collection<AssignableReplica> replicas) {
...
-  private boolean addToAssignmentRecord(AssignableReplica replica,
-      Map<String, Set<String>> currentAssignments) {
-    return currentAssignments.computeIfAbsent(replica.getResourceName(), k -> new HashSet<>())
-        .add(replica.getPartitionName());
+  /**
+   * @return True if the replica has been added to record successfully. False if the replica already
+   * exists on this node.
+   */
+  private boolean addToAssignmentRecord(AssignableReplica replica) {
+    if (null == _currentAssignments.computeIfAbsent(replica.getResourceName(), k -> new HashMap<>())
```
narendly: Add some comment here - say that you're doing a null comparison and relying on the behavior of putIfAbsent(). 
I personally do not like this style and think it's awkward. It's not necessarily more readable or concise than doing a if (!contains()) check. I'd personally just go with a contains check. Or at least add more comments so it reads better.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -108,17 +107,14 @@ private void refresh(ClusterConfig clusterConfig, InstanceConfig instanceConfig,
    * @param assignableReplica - the replica to be assigned
    */
   void assign(AssignableReplica assignableReplica) {
-    if (!addToAssignmentRecord(assignableReplica, _currentAssignments)) {
+    if (addToAssignmentRecord(assignableReplica)) {
+      assignableReplica.getCapacity().entrySet().stream().forEach(
+          capacity -> updateCapacityAndUtilization(capacity.getKey(), capacity.getValue()));
+    } else {
```
narendly: Why don't we just throw this exception directly inside `addToAssignmentRecord()`? That way code is more concise. No need for this if statement. Also, we avoid having to return a boolean in `addToAssignmentRecord()`, which, to me at least, is quite unreadable...

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -138,15 +134,16 @@ void release(AssignableReplica assignableReplica) throws IllegalArgumentExceptio
           getInstanceName());
       return;
     }
-    Set<String> partitions = _currentAssignments.get(resourceName);
-    if (!partitions.contains(partitionName)) {
+    Map<String, String> partitionStateMap = _currentAssignments.get(resourceName);
```
narendly: Probably a good idea to add a comment here what is mapping to what.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -138,15 +134,16 @@ void release(AssignableReplica assignableReplica) throws IllegalArgumentExceptio
...
+    if (!partitionStateMap.containsKey(partitionName) || !partitionStateMap.get(partitionName)
+        .equals(assignableReplica.getReplicaState())) {
```
narendly: What if it's the second condition? If it's the second condition that's triggering this, then the log would be misleading.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -156,12 +153,12 @@ void release(AssignableReplica assignableReplica) throws IllegalArgumentExceptio
         .forEach(entry -> updateCapacityAndUtilization(entry.getKey(), -1 * entry.getValue()));
   }
 
-  public Map<String, Set<String>> getCurrentAssignmentsMap() {
+  public Map<String, Map<String, String>> getCurrentAssignmentsMap() {
```
narendly: Could we please annotate the public methods, and especially makes sure the return values are marked with what the strings in the map stand for?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -156,12 +153,12 @@ void release(AssignableReplica assignableReplica) throws IllegalArgumentExceptio
...
     return _currentAssignments;
   }
 
-  public Set<String> getCurrentAssignmentsByResource(String resource) {
-    return _currentAssignments.getOrDefault(resource, Collections.emptySet());
+  public Map<String, String> getCurrentAssignmentsByResource(String resource) {
```
narendly: Could we please annotate the public methods, and especially makes sure the return values are marked with what the strings in the map stand for? It's not so easy for someone like me to decode this :)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
jiajunwang: What you see is correct. We are already in a node class. No need to put instance name.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
jiajunwang: Maybe just change to "index" instead of "index map"?

Some background here, this is mainly for search (that's why I take it as an index). Note we can always derive this map from the _currentAssignments. But repeatedly calculating it is not efficient.

In addition, not sure how you are confused with the node name in this context. But this set is for "the partition names of replicas that are in the top state".  Maybe I shall just put all this in the comment.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
jiajunwang: Good idea. Will change.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
jiajunwang: partitionStateMap => <partition, state> map?
I think it is clear enough.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
jiajunwang: Will add "with state xxx" to the error message.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
jiajunwang: I agree more comments would be helpful here. About annotation, could you please specify? Example would be helpful. Not sure if Helix has such a use case that I am not aware of?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
jiajunwang: Sure, I agree. I think I went too far, haha. Will change it.

jiajunwang:
On my second thought, directly recording the replica instances would be easier. I have refactored the change, and tried to address the comments.
For the pending ones, please reply so I can ensure that my understanding is correct.

jiajunwang:
@i3wangyi Please take a look at this change, since you are working on the algorithm.

BTW, it seems that I can only request @narendly for reviewing but not @i3wangyi. Probably because @i3wangyi is not a committer yet.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
```diff
@@ -107,6 +107,18 @@ public int compareTo(AssignableReplica replica) {
     return 0;
   }
 
+  @Override
+  public boolean equals(Object obj) {
+    if (obj == null) {
```
narendly: if (obj == null || !(obj instanceof AssignableReplica)) {
  return false;
}
return compareTo((AssignableReplica) obj) == 0;

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -108,78 +111,106 @@ private void refresh(ClusterConfig clusterConfig, InstanceConfig instanceConfig,
...
+  /**
+   * @return The current available capacity.
+   */
   public Map<String, Integer> getCurrentCapacity() {
-    return _currentCapacity;
+    return _currentCapacityMap;
   }
 
+  /**
+   * @return The highest utilization number of the node among all the capacity category.
```
narendly: Nit: So this description is not super clear... could you explain a little more with an example and why we need this?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
jiajunwang: This is mainly for cleaner code. And better to ensure the same code style as the other equals() in the Helix project.

jiajunwang:

PR #450, opened 2019-09-05
Implement one of the soft constraints

i3wangyi:
### Issues

(#421)

### Description

**Least Used Instance Soft Constraint** 
The constraint measures the current usages of partitions on the instance versus the estimated max allowed partitions. 
Intuitively, we need to give a low score (discourage) if the # of partitions gets closer to upper bound; Give a high score (encourage) if # of partitions is way below the upper bound;

  

### Tests

unit test added for the class

### Commits

### Documentation

### Code Quality


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.helix.controller.rebalancer.waged.model.AssignableNode;
+import org.apache.helix.controller.rebalancer.waged.model.AssignableReplica;
+import org.apache.helix.controller.rebalancer.waged.model.ClusterContext;
+
+
+/**
+ * Evaluate by instance's current partition count and estimated max partition count
+ */
+class InstancePartitionsCountConstraint extends SoftConstraint {
```
jiajunwang: In my opinion, the original name is clearer because we know how the constraint evaluates briefly by looking at the name. This new name is OK, but in this case, please add more comments to say how it evaluates. Which case will have the highest score, which case will have the lowest score.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
```diff
@@ -0,0 +1,55 @@
...
+    if (currentPartitionCount == 0) {
+      return getMaxScore();
+    }
+    //When the node usage reaches the estimated max partition, return minimal score
+    if (currentPartitionCount >= estimatedMaxPartitionCount) {
+      return getMinScore();
+    }
+
+    // Return the usage percentage by estimation
+    return (estimatedMaxPartitionCount - currentPartitionCount) / (float) estimatedMaxPartitionCount;
```
jiajunwang: Sorry, I am a little bit confused here. I know we are going to normalize this score later, right? Will the final result be (0.0-1.0) or (minScore, maxScore) after the normalizing? My understanding is the eventual normalized result should be (0.0 - 1.0). Please let me know if I am wrong.

Based on this understanding, it is strange that you ignore the maxScore and minScore here.
I assume something like this:
  (estimatedMaxPartitionCount - currentPartitionCount) / (float) estimatedMaxPartitionCount * (getMaxScore() - getMinScore()) + getMinScore()

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
```diff
@@ -0,0 +1,55 @@
...
+  @Override
+  float evaluateAssignment(AssignableNode node, AssignableReplica replica, ClusterContext clusterContext) {
+    int estimatedMaxPartitionCount = clusterContext.getEstimatedMaxPartitionCount();
+    int currentPartitionCount = node.getAssignedReplicaCount();
+    // When the node is idle, return with the maxScore.
+    if (currentPartitionCount == 0) {
+      return getMaxScore();
+    }
+    //When the node usage reaches the estimated max partition, return minimal score
+    if (currentPartitionCount >= estimatedMaxPartitionCount) {
```
jiajunwang: One thing to clear is that estimatedMaxPartitionCount is not a hard limitation. It is evaluated as the expected partition count.

Based on this fact, the current implementation means all the nodes with more partition than the estimate partition count will be evaluated the same.
However, in reality, a node with one more partition would be a better placement than a node with 100 more partitions.
In my previous POC, I tried to make the scale wider. Basically, when the assigned partition count reaches the estimated count, the evaluated score would be (max - min) / 2 + min. This will make the algorithm return a more accurate result IMO.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -57,15 +58,19 @@
...
-   * weighted by the
-   * individual normalized constraint weights.
-   * Each individual constraint will define the meaning of MINIMAL_SCORE to MAXIMUM_SCORE
-   * differently.
-   * @return float value representing the score
+   * Evaluate and give a score for an potential assignment partition -> instance
+   * @return The score of the assignment in float value
    */
-  abstract float getAssignmentOriginScore(AssignableNode node, AssignableReplica replica,
+  abstract float evaluateAssignment(AssignableNode node, AssignableReplica replica,
```
jiajunwang: We are still using the normalize method you proposed earlier, right? I think we need a method name to differentiate this score from the normalized score. "evaluateAssignment" is not clear enough for me.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
i3wangyi: No, the min-max scaler function will guarantee you a final score within (0, 1). That's the value of a scaler function that leaves the implementation easier AS LONG AS the original score doesn't exceed the min/max and we don't have to worry about scaling and calculation ourselves.

i3wangyi:
> Another question, by looking at this soft constraint impl, I am wondering if we just enforce all the soft constraint implementation to return an evaluation between (0.0 - 1.0), will that be easier? At least we don't need "getScalerFunction", right?

How do you want to enforce the constraint in code? Using comment? Comments will not always guarantee the correct behavior, only self explainable code does. I think scaler function is necessary and a reminder to have developers aware of the range without looking into comments

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
i3wangyi: I leave the detailed comments in the method implementation. For people who would like to understand details, they could check the details while the java doc briefly tells which criteria is used in this constraint (Partitions count on Instance).

It makes more sense when we'd like to probably change the implementation details because we know this class will use partitions count as the evaluation criteria. How it is used can be changed.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
i3wangyi: (max - min) / 2 + min, what's the mathematical meaning behind it? 
Don't we basically want to penalize such an assignment by giving it a very low score? One single constraint shouldn't know about the global assignment status but only takes care THE particular assignment if it's good or not, isn't it?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
jiajunwang: @narendly LeastPartitionCountConstraint

jiajunwang: @i3wangyi You should at least mention that the constraint will evaluate the full node with a lower score. And we will never change this point. Otherwise, we will have new constraint instead.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.helix.controller.rebalancer.waged.model.AssignableReplica;
+import org.apache.helix.controller.rebalancer.waged.model.ClusterContext;
+
+
+/**
+ * Evaluate by instance's current partition count versus estimated max partition count
+ */
+class InstancePartitionsCountConstraint extends SoftConstraint {
+
+  InstancePartitionsCountConstraint() {
```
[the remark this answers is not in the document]
i3wangyi: The default constructor will be needed when algorithm factory class wants to create the instance. Right now, it's not used

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
jiajunwang: Have you read my comments in detail? I don't think so based on what you replyed.

estimatedMaxPartitionCount = 100.
Node A has current assignment 101.
Node B has current assignment 300.

Do you want to score these 2 nodes the same with this constraint or not?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
jiajunwang: >  the min-max scaler function will guarantee you a final score within (0, 1)
Don't forget the scale function also needs to ensure a uniform distribution of the values.

Let's go with an example.

Max = 1000, Min = -1000
And estimatedMaxPartitionCount = 100, currentPartitionCount = 1. Then, the score returned is 0.99.
And after scale, it would be (0.99 - -1000) / 2000 = 0.500495.
And the scaled score would always be between 0.5 ~ 0.5005 (except for the 2 corner cases), right? Do you think this is the right math we want to use?

i3wangyi: OK. If you want uniform distribution, you can create the soft constraint with minscore = 0, maxscore = 1, these can be overridden. Correct?

jiajunwang:
> > Another question, by looking at this soft constraint impl, I am wondering if we just enforce all the soft constraint implementation to return an evaluation between (0.0 - 1.0), will that be easier? At least we don't need "getScalerFunction", right?
> 
> How do you want to enforce the constraint in code? Using comment? Comments will not always guarantee the correct behavior, only self explainable code does. I think scaler function is necessary and a reminder to have developers aware of the range without looking into comments

For now, if I am the developer, and I return a value out of the scope of (minScore, maxScore), what will happen?
I think the max and min thresholds are also kind of comments, right?

I guess we will need to add a check to validate. However, in that case, we can just enforce the minScore = 0.0, maxScore = 1.0. And my proposal will be equal to this solution, right?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
jiajunwang: I think that is necessary. Or the result won't be comparable. And we should enforce it without providing such an option.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
i3wangyi: Based on previous implementation `(max - min) / 2 + min` and the one here `min()`, your examples would return an identical and fixed constant. It won't respect how much it's exceeding the estimated max partitions. Isn't the behavior what we want?

i3wangyi:
> > > Another question, by looking at this soft constraint impl, I am wondering if we just enforce all the soft constraint implementation to return an evaluation between (0.0 - 1.0), will that be easier? At least we don't need "getScalerFunction", right?
> > 
> > 
> > How do you want to enforce the constraint in code? Using comment? Comments will not always guarantee the correct behavior, only self explainable code does. I think scaler function is necessary and a reminder to have developers aware of the range without looking into comments
> 
> For now, if I am the developer, and I return a value out of the scope of (minScore, maxScore), what will happen?
> I think the max and min thresholds are also kind of comments, right?
> 
> I guess we will need to add a check to validate. However, in that case, we can just enforce the minScore = 0.0, maxScore = 1.0. And my proposal will be equal to this solution, right?

Whoever develops the specific constraint needs to respect the (min, max) range, it's not a question. If you want to enforce such hardcoded min/max in constraint definition,  I think only exposing the constructor below will be sufficient and clean
```
public LeastUsedPartitionsCount() {
   super(0, 1);
}
```

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
jiajunwang: That was a typo. should be (current - min) / 2 + min. But what I meant to say is more than this. Note the current count is hard limited by the hard constraint. Which could be much larger than estimatedMaxPartitionCount. It is up to you whether we want to score the larger count differently or not. Although I strongly suggest doing so.

Otherwise, at some point, if all nodes have partition count more than the estimatedMaxPartitionCount, this soft constraint will lose the capability to evaluate reasonablly.

i3wangyi: If we'd like to achieve different severeness of exceeding the estimated average,  in math, 
we can simply return `(estimatedMaxPartitionCount - currentPartitionCount) / (float) estimatedMaxPartitionCount;`.

The negative score is also acceptable.
The range becomes
(-FLOAT.MIN, 1)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
narendly: @jiajunwang @i3wangyi I think with the added explanation it's a lot easier understand. In my opinion it might be hard to capture everything in a name, so I'm fine with either. Thanks for adding the description.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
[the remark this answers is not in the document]
narendly: Okay.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -24,8 +24,9 @@
 import org.apache.helix.controller.rebalancer.waged.model.ClusterContext;
 
 /**
- * Evaluate a partition allocation proposal and return a score within the normalized range.
- * A higher score means the proposal is more preferred.
+ * The "soft" constraint evaluates the potential assignment by giving it score
```
narendly: the potential assignment -> the optimality of an assignment on a scale of 0 to maxScore.
giving it "a" score

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -57,17 +58,32 @@
...
+   * Child class only needs to care about how the score is implemented
+   * @return The score of the assignment in float value
    */
-  abstract float getAssignmentOriginScore(AssignableNode node, AssignableReplica replica,
+  protected abstract float getAssignmentScore(AssignableNode node, AssignableReplica replica,
       ClusterContext clusterContext);
 
+  /**
+   * Evaluate and give a score for an potential assignment partition -> instance
+   * It's the only exposed method to the external
```
narendly: "the external" -> "the caller"?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
```diff
@@ -0,0 +1,62 @@
...
+    if (currentPartitionCount == 0) {
+      return getMaxScore();
+    }
+
+    return estimatedMaxPartitionCount / (float) currentPartitionCount;
+  }
+
+  @Override
+  ScalerFunction getScalerFunction() {
+    return score -> score / getMaxScore();
```
jiajunwang: I still have the concern that this is not a uniform distribution.
If total partition count = 1000, and there are 10 nodes, the possible return score will be distributed in [0, 0.1], [1]. A score between (0.1, 1) will never be returned.

This probably makes the tuning work harder because the behavior of this constrain is harder to reasoning.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
```diff
@@ -0,0 +1,52 @@
...
+      ClusterContext clusterContext) {
+    float doubleEstimatedMaxPartitionCount = 2 * clusterContext.getEstimatedMaxPartitionCount();
+    float currentPartitionCount = node.getAssignedReplicaCount();
+    return Math.max((doubleEstimatedMaxPartitionCount - currentPartitionCount)
+        / doubleEstimatedMaxPartitionCount, 0);
+  }
+
+  @Override
+  ScalerFunction getNormalizeFunction() {
+    return score -> score * (getMaxScore() - getMinScore()) + getMinScore();
```
jiajunwang: Why override the getNormalizeFunction, it is equal to the abstract class soft constraint method, right?

i3wangyi: You're right, it's the same and verified by unit test

i3wangyi:
The PR is ready to merge into wagedRebalancer, approved by @jiajunwang @narendly

Issue #441, opened 2019-08-29
Enforce capacity key checks in the WAGED rebalancer.


jiajunwang:
The WAGED rebalancer replies on multiple layers of config (cluster, instance, and resource) to define the capacity and quota usages. The config logic and APIs have been implemented, but we don't have the logic to check yet.

Goals:
- The rebalancer shall fail the rebalance if the configuration does not fit the expectation or contains any conflict.
- On an empty capacity/weight configuration, the rebalancer shall still be able to work.

i3wangyi:
I suggest let's create a capacity class that takes care of everything (validation/initialization/update/), what do you think?

PR #451, opened 2019-09-05
Validate the instance capacity/partition weight configuration while constructing the assignable instances

jiajunwang:
### Issues

#441

### Description

Compare the configure items with the required capacity keys that are defined in the cluster config when build the assignable instances.
- According to the design, all the required capacity keys must appear in the instance capacity config.
- As for the partition weights, the corresponding weight item will be filled with value 0 if the required capacity key is not specified in the resource config.

### Tests

Modified the test cases in TestAssignableNode and TestAssignableReplica for the new validation.

Run all the WAGED rebalancer related tests. The code has not been connected with the other Helix component yet. So no need to run all mvn tests.

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
@@ -91,21 +91,7 @@ public class AssignableNode {
       Collection<AssignableReplica> existingAssignment) {
     reset();
 
-    List<String> requiredCapacityKeys = clusterConfig.getInstanceCapacityKeys();
-
-    Map<String, Integer> instanceCapacity = instanceConfig.getInstanceCapacityMap();
-    if (instanceCapacity.isEmpty()) {
-      instanceCapacity = clusterConfig.getDefaultInstanceCapacityMap();
-    }
-    // Remove all the non-required capacity items from the map.
-    instanceCapacity.keySet().retainAll(requiredCapacityKeys);
-    // All the required keys must exist in the instance config.
-    if (!instanceCapacity.keySet().containsAll(requiredCapacityKeys)) {
-      throw new HelixException(String.format(
-          "The required capacity keys %s is not fully configured in the instance %s capacity map %s.",
-          requiredCapacityKeys.toString(), _instanceName, instanceCapacity.toString()));
-    }
-
+    Map<String, Integer> instanceCapacity = fetchInstanceCapacity(clusterConfig, instanceConfig);
     _currentCapacityMap.putAll(instanceCapacity);
     _faultZone = computeFaultZone(clusterConfig, instanceConfig);
     _instanceTags = new HashSet<>(instanceConfig.getTags());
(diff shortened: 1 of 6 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -91,10 +91,21 @@ private void refresh(ClusterConfig clusterConfig, InstanceConfig instanceConfig,
       Collection<AssignableReplica> existingAssignment) {
     reset();
 
+    List<String> requiredCapacityKeys = clusterConfig.getInstanceCapacityKeys();
+
     Map<String, Integer> instanceCapacity = instanceConfig.getInstanceCapacityMap();
     if (instanceCapacity.isEmpty()) {
       instanceCapacity = clusterConfig.getDefaultInstanceCapacityMap();
     }
+    // Remove all the non-required capacity items from the map.
```
narendly: Nit: Could we move this check to a private method? I think this line of logic separates nicely.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
```diff
@@ -125,11 +125,9 @@ public static ClusterModel generateClusterModel(ResourceControllerDataProvider d
         // 2. if the resource does appear in the best possible assignment, need to reassign.
         if (clusterChanges
             .getOrDefault(HelixConstants.ChangeType.RESOURCE_CONFIG, Collections.emptySet())
-            .contains(resourceName)
-            || clusterChanges
+            .contains(resourceName) || clusterChanges
```
narendly: Why is this changing? Was this not formatted before?

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/model/TestAssignableReplica.java
```diff
@@ -96,4 +101,29 @@ public void testConstructRepliaWithResourceConfig() throws IOException {
     Assert.assertEquals(replica.getResourceInstanceGroupTag(), group);
     Assert.assertEquals(replica.getResourceMaxPartitionsPerInstance(), maxPartition);
   }
+
+  @Test
+  public void testIncompletePartitionWeightConfig() throws IOException {
```
narendly: Question:

If we don't have the key, are we treating it as 0 or erroring out?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
i3wangyi: I suggest creating a centralized capacity class that takes care of everything, (validation/init/updates), e.g, https://github.com/apache/helix/pull/452/files#diff-f45e90891858518de7280701a94da157

jiajunwang: @i3wangyi That is a good direction. But I believe we need some time to refine it. Let's use this implementation for now. In the near future, we are not planning to make capacity more complicated.
@narendly It is clear, but too many private methods are also confusing. I prefer to keep it in this way.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
jiajunwang: I think I manually formatted because of some comments. Anyway, the current version is automatically formatted.

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/model/TestAssignableReplica.java
jiajunwang: For replica, it would be 0.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
i3wangyi: It's already quite complicated to review. Make it a centralized IMO would only make things easier and more scalable to add/remove capacity keys.

jiajunwang: Sure, I don't have a strong preference here.

jiajunwang:

PR #453, opened 2019-09-08
Implement AssignmentMetadataStore

### LINKED ISSUE #444: Add Assignment Metadata Store for persisting assignment metadata
We want a component called `AssignmentMetadataStore` that maintains an instance of `BucketDataAccessor` (relevant issue here: https://github.com/apache/helix/issues/419 #419 ). This component will convert an internal representation of assignment metadata for multiple resources into one ZNRecord that will be the payload going into the write() of BucketDataAccessor.

narendly:
### Issues

Fixes #444 

### Description

AssignmentMetadataStore is a component for the new WAGED Rebalaner. It provides APIs that allows the rebalancer to read and write the baseline and best possible assignments using BucketDataAccessor.

Changelist:
1. Add AssignmentMetadataStore
2. Add an integration test: TestAssignmentMetadataStore

### Tests

TestAssignmentMetadataStore

$ mvn test -Dtest=AssignmentMetadataStore

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.666 s - in org.apache.helix.controller.rebalancer.waged.TestAssignmentMetadataStore
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  10.434 s
[INFO] Finished at: 2019-09-07T18:33:53-07:00

### Commits

### Documentation

https://github.com/apache/helix/wiki/Storing-Assignment-Metadata-in-ZooKeeper

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -80,9 +80,6 @@ public class AssignmentMetadataStore {
   }
 
   public void persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
-    // Update the in-memory reference
-    _globalBaseline = globalBaseline;
-
     // TODO: Make the write async?
     // Persist to ZK
     HelixProperty combinedAssignments = combineAssignments(BASELINE_KEY, globalBaseline);
@@ -92,13 +89,13 @@ public class AssignmentMetadataStore {
       // TODO: Improve failure handling
       throw new HelixException("Failed to persist baseline!", e);
     }
+
+    // Update the in-memory reference
+    _globalBaseline = globalBaseline;
   }
 
   public void persistBestPossibleAssignment(
       Map<String, ResourceAssignment> bestPossibleAssignment) {
-    // Update the in-memory reference
-    _bestPossibleAssignment = bestPossibleAssignment;
-
     // TODO: Make the write async?
     // Persist to ZK asynchronously
     HelixProperty combinedAssignments =
@@ -109,6 +106,9 @@ public class AssignmentMetadataStore {
       // TODO: Improve failure handling
       throw new HelixException("Failed to persist baseline!", e);
     }
+
+    // Update the in-memory reference
+    _bestPossibleAssignment = bestPossibleAssignment;
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -19,6 +19,16 @@ package org.apache.helix.controller.rebalancer.waged;
  * under the License.
  */
 
+import java.io.IOException;
+import java.util.Arrays;
+import org.I0Itec.zkclient.serialize.ZkSerializer;
+import org.apache.helix.BucketDataAccessor;
+import org.apache.helix.HelixException;
+import org.apache.helix.HelixManager;
+import org.apache.helix.HelixProperty;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.manager.zk.ZNRecordJacksonSerializer;
+import org.apache.helix.manager.zk.ZkBucketDataAccessor;
 import org.apache.helix.model.ResourceAssignment;
 
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -28,24 +38,114 @@
...
-  private Map<String, ResourceAssignment> _persistBestPossibleAssignment = new HashMap<>();
+  private static final String ASSIGNMENT_METADATA_KEY = "ASSIGNMENT_METADATA";
+  private static final String BASELINE_KEY = "BASELINE";
+  private static final String BEST_POSSIBLE_KEY = "BEST_POSSIBLE";
+  private static final ZkSerializer SERIALIZER = new ZNRecordJacksonSerializer();
+
+  private BucketDataAccessor _dataAccessor;
+  private String _baselinePath;
+  private String _bestPossiblePath;
+  private Map<String, ResourceAssignment> _globalBaseline;
```
i3wangyi: Actually I don't think these 2 private fields are necessary unless you want to have "lazy cache" functionality that "get" from DB once and persist.  But be aware of the approach that the client probably won't always be able to get the latest data

narendly: Could you re-word your comment? I don't think I'm understanding this clearly.

The reason I am keeping these two private fields is if there are no connection losses or controller switches, there wouldn't need to read everything from ZK since we'd already have it in-memory.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -28,24 +38,114 @@
...
+  private Map<String, ResourceAssignment> _bestPossibleAssignment;
+
+  public AssignmentMetadataStore(HelixManager helixManager) {
+    _dataAccessor = new ZkBucketDataAccessor(helixManager.getMetadataStoreConnectionString());
+    _baselinePath =
```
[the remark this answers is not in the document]
narendly: Thanks. Changed to 

`  private static final String BASELINE_TEMPLATE = "/%s/%s/BASELINE";
  private static final String BEST_POSSIBLE_TEMPLATE = "/%s/%s/BEST_POSSIBLE";`

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
i3wangyi: "Already have it in-memory" that's the idea of lazy cache: read from DB and keeps in memory. However, it's possible the data in memory is out of date, and you will need to consider when the client needs to refresh its "cached" data. 
To keep things simpler, if performance is not the bottleneck, always read from ZK will guarantee the correct up-to-date data. In this case, these 2 private fields are not necessary.
Also, curious that in the case of the controller switch, the memory is literally gone, isn't it?

jiajunwang: 1. Performance is critical here.
2. For AssignmentMetadataStore, there should be only one writer. The controller. So if we verify if the current manager is the leader, we shall be good.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -28,24 +38,116 @@
...
+        String.format(BASELINE_TEMPLATE, helixManager.getClusterName(), ASSIGNMENT_METADATA_KEY);
```
jiajunwang: Will it be cleaner if we make 3 methods instead? And then just call these methods whenever required.
1. getAssignmentMetadatPath(clusterName)
2. getBaselinePath(metadataPath)
3. getBestPossiblePath(metadataPath)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
narendly: @i3wangyi How could the data in memory be out-of-date? I think there's a gap in understanding somewhere...

i3wangyi: For example, there's a custom tool that updates the baseline and best possible mapping after this metadata class reads & caches the result from ZK. 
Also, take your exposed 2 methods (`getbasline() and persistBaseline()`). I understand the `persistBaseline` will overwrite the local private `_baseline` in the method body. In the concurrency environment, it will cause the race condition and it creates a read after write consistency issue (the in-memory data will not get updated).

narendly: Ah I see where the disconnect is: 
`AssignmentMetadataStore` is for WAGED rebalancer use only. That means that no other tool should update the assignment metadata. There would be only one instance of WAGED Rebalancer and this `AssignmentMetadataStore` will be a static object or a singleton, so we do not have to guard against concurrent writes.

i3wangyi: Got it. But there's no such constraint based on the class design. First, it's a public class; Second, it has a public constructor, meaning everyone could create an instance not subjective to waged rebalancer; I suggest enforcing it by singleton pattern (a factory constructor and private constructors).

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -28,24 +38,116 @@
...
-    return _persistGlobalBaseline;
+    // Return the in-memory baseline. If null, read from ZK. This is to minimize reads from ZK
+    if (_globalBaseline == null) {
+      try {
+        HelixProperty baseline =
+            _dataAccessor.compressedBucketRead(_baselinePath, HelixProperty.class);
+        _globalBaseline = splitAssignments(baseline);
+      } catch (HelixException e) {
+        // Metadata does not exist, so return an empty map
+        _globalBaseline = new HashMap<>();
```
jiajunwang: I would prefer throwing an exception. The rebalancer can decide what's the next step.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -28,24 +38,116 @@
...
+  }
+
+  public Map<String, ResourceAssignment> getBestPossibleAssignment() {
+    // Return the in-memory baseline. If null, read from ZK. This is to minimize reads from ZK
+    if (_bestPossibleAssignment == null) {
+      try {
+        HelixProperty baseline =
+            _dataAccessor.compressedBucketRead(_bestPossiblePath, HelixProperty.class);
+        _bestPossibleAssignment = splitAssignments(baseline);
+      } catch (HelixException e) {
```
jiajunwang: Same here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -28,24 +38,116 @@
...
+    _globalBaseline = globalBaseline;
+
+    // TODO: Make the write async?
+    // Persist to ZK
+    HelixProperty combinedAssignments = combineAssignments(BASELINE_KEY, globalBaseline);
+    try {
+      _dataAccessor.compressedBucketWrite(_baselinePath, combinedAssignments);
+    } catch (IOException e) {
+      // TODO: Improve failure handling
+      throw new HelixException("Failed to persist baseline!", e);
```
jiajunwang: Please either cleanup the new values fro the memory, or set the in-memory part after this write is done successfully.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -28,24 +38,116 @@
...
+    }
   }
 
-  public Map<String, ResourceAssignment> getBestPossibleAssignment() {
-    return _persistBestPossibleAssignment;
+  public void persistBestPossibleAssignment(
+      Map<String, ResourceAssignment> bestPossibleAssignment) {
+    // Update the in-memory reference
+    _bestPossibleAssignment.putAll(bestPossibleAssignment);
```
jiajunwang: _bestPossibleAssignment = bestPossibleAssignment is safer, right?
Or,
_bestPossibleAssignment.clear()
_bestPossibleAssignment.putAll(bestPossibleAssignment);

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
narendly: I think using the factory constructor is a bit of an overkill, but I think your point is valid for making it package private so that only waged rebalancer could use it :)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
narendly: Okay. I will throw an exception instead of returning an empty map by removing the try-catch block.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
narendly: Removed the try-catch.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
narendly: What do you mean? We update the reference in line 94. @jiajunwang

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
narendly: This was a mistake. This should look like persistBaseline. Fixed.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
narendly: @jiajunwang I want to make the paths static for now instead of generating it every time. It's not too clear whether we'll ever have growing number paths we write to but for now there are only two. Hard-coding gets the job done without incurring any more work.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
narendly: Clarified offline. Update the in-memory reference after the write via data accessor.

narendly:


Implement AssignmentMetadataStore

------

AssignmentMetadataStore is a component for the new WAGED Rebalaner. It provides APIs that allows the rebalancer to read and write the baseline and best possible assignments using BucketDataAccessor.

Changelist:

Add AssignmentMetadataStore
Add an integration test: TestAssignmentMetadataStore

PR #456, opened 2019-09-10
Add delayed rebalance and user-defined preference list features to the WAGED rebalancer.

jiajunwang:
### Issues

#402

### Description

Add delayed rebalance and user-defined preference list features to the WAGED rebalancer.
Add the delayed rebalance scheduling logic.
Add the necessary tests.

### Tests

TestDelayedWagedRebalance
TestDelayedWagedRebalanceWithDisabledInstance
TestDelayedWagedRebalanceWithRackaware
TestMixedModeWagedRebalance
TestZeroReplicaAvoidance (new test case added)

[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 986, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE

Re-run the failed tests:
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
@@ -20,12 +20,12 @@ package org.apache.helix.controller.rebalancer.util;
  */
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
 import org.apache.helix.HelixManager;
 import org.apache.helix.model.ClusterConfig;
 import org.apache.helix.model.IdealState;
@@ -33,8 +33,9 @@ import org.apache.helix.model.InstanceConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
 /**
- * The util to support delayed rebalance logic.
+ * The util for supporting delayed rebalance logic.
  */
 public class DelayedRebalanceUtil {
   private static final Logger LOG = LoggerFactory.getLogger(DelayedRebalanceUtil.class);
@@ -42,36 +43,29 @@ public class DelayedRebalanceUtil {
   private static RebalanceScheduler _rebalanceScheduler = new RebalanceScheduler();
 
   /**
-   * @return the rebalance delay based on the ClusterConfig configurations.
-   * Resource level configuration will be ignored.
-   */
-  public static long getRebalanceDelay(ClusterConfig clusterConfig) {
-    return getRebalanceDelay(null, clusterConfig);
-  }
-
-  /**
    * @return true if delay rebalance is configured and enabled in the ClusterConfig configurations.
-   * Resource level configuration will be ignored.
    */
   public static boolean isDelayRebalanceEnabled(ClusterConfig clusterConfig) {
-    return isDelayRebalanceEnabled(null, clusterConfig);
+    long delay = clusterConfig.getRebalanceDelayTime();
+    return (delay > 0 && clusterConfig.isDelayRebalaceEnabled());
   }
 
   /**
-   * @return all active instances (live instances plus offline-yet-active instances) while considering delay rebalance configurations.
+   * @return true if delay rebalance is configured and enabled in Resource IdealState and the
+   * ClusterConfig configurations.
    */
-  public static Set<String> getActiveInstances(Set<String> allNodes, Set<String> liveEnabledNodes,
-      Map<String, Long> instanceOfflineTimeMap, Set<String> liveNodes,
-      Map<String, InstanceConfig> instanceConfigMap, ClusterConfig clusterConfig) {
-    return getActiveInstances(allNodes, null, liveEnabledNodes, instanceOfflineTimeMap, liveNodes,
-        instanceConfigMap, clusterConfig.getRebalanceDelayTime(), clusterConfig);
+  public static boolean isDelayRebalanceEnabled(IdealState idealState,
+      ClusterConfig clusterConfig) {
+    long delay = getRebalanceDelay(idealState, clusterConfig);
+    return (delay > 0 && idealState.isDelayRebalanceEnabled() && clusterConfig
+        .isDelayRebalaceEnabled());
   }
 
   /**
    * @return the rebalance delay based on Resource IdealState and the ClusterConfig configurations.
    */
   public static long getRebalanceDelay(IdealState idealState, ClusterConfig clusterConfig) {
-    long delayTime = idealState == null ? -1 : idealState.getRebalanceDelay();
+    long delayTime = idealState.getRebalanceDelay();
     if (delayTime < 0) {
       delayTime = clusterConfig.getRebalanceDelayTime();
     }
@@ -79,30 +73,39 @@ public class DelayedRebalanceUtil {
   }
 
   /**
-   * @return true if delay rebalance is configured and enabled in Resource IdealState or the ClusterConfig configurations.
+   * @return all active instances (live instances plus offline-yet-active instances) while
+   * considering cluster delay rebalance configurations.
    */
-  public static boolean isDelayRebalanceEnabled(IdealState idealState,
-      ClusterConfig clusterConfig) {
-    long delay = getRebalanceDelay(idealState, clusterConfig);
-    return (delay > 0 && (idealState == null || idealState.isDelayRebalanceEnabled())
-        && clusterConfig.isDelayRebalaceEnabled());
+  public static Set<String> getActiveInstances(Set<String> allNodes, Set<String> liveEnabledNodes,
+      Map<String, Long> instanceOfflineTimeMap, Set<String> liveNodes,
+      Map<String, InstanceConfig> instanceConfigMap, ClusterConfig clusterConfig) {
+    if (!isDelayRebalanceEnabled(clusterConfig)) {
+      return Collections.emptySet();
+    }
+    return getActiveInstances(allNodes, liveEnabledNodes, instanceOfflineTimeMap, liveNodes,
+        instanceConfigMap, clusterConfig.getRebalanceDelayTime(), clusterConfig);
   }
 
   /**
-   * @return all active instances (live instances plus offline-yet-active instances) while considering delay rebalance configurations.
+   * @return all active instances (live instances plus offline-yet-active instances) while
+   * considering cluster and the resource delay rebalance configurations.
    */
   public static Set<String> getActiveInstances(Set<String> allNodes, IdealState idealState,
       Set<String> liveEnabledNodes, Map<String, Long> instanceOfflineTimeMap, Set<String> liveNodes,
       Map<String, InstanceConfig> instanceConfigMap, long delay, ClusterConfig clusterConfig) {
-    Set<String> activeInstances = new HashSet<>(liveEnabledNodes);
-
     if (!isDelayRebalanceEnabled(idealState, clusterConfig)) {
-      return activeInstances;
+      return Collections.emptySet();
     }
+    return getActiveInstances(allNodes, liveEnabledNodes, instanceOfflineTimeMap, liveNodes,
+        instanceConfigMap, delay, clusterConfig);
+  }
 
+  private static Set<String> getActiveInstances(Set<String> allNodes, Set<String> liveEnabledNodes,
+      Map<String, Long> instanceOfflineTimeMap, Set<String> liveNodes,
+      Map<String, InstanceConfig> instanceConfigMap, long delay, ClusterConfig clusterConfig) {
+    Set<String> activeInstances = new HashSet<>(liveEnabledNodes);
     Set<String> offlineOrDisabledInstances = new HashSet<>(allNodes);
     offlineOrDisabledInstances.removeAll(liveEnabledNodes);
-
     long currentTime = System.currentTimeMillis();
     for (String ins : offlineOrDisabledInstances) {
       long inactiveTime = getInactiveTime(ins, liveNodes, instanceOfflineTimeMap.get(ins), delay,
@@ -113,7 +116,6 @@ public class DelayedRebalanceUtil {
         activeInstances.add(ins);
       }
     }
-
     return activeInstances;
   }
 
@@ -121,7 +123,7 @@ public class DelayedRebalanceUtil {
    * @return The time when an offline or disabled instance should be treated as inactive.
    * Return -1 if it is inactive now.
    */
-  public static long getInactiveTime(String instance, Set<String> liveInstances, Long offlineTime,
+  private static long getInactiveTime(String instance, Set<String> liveInstances, Long offlineTime,
       long delay, InstanceConfig instanceConfig, ClusterConfig clusterConfig) {
     long inactiveTime = Long.MAX_VALUE;
 
@@ -157,12 +159,15 @@ public class DelayedRebalanceUtil {
   }
 
   /**
-   * Merge the new ideal preference list with the "active" mapping that is calculated based on the
-   * delayed rebalance configuration. The method will prioritize the active preference list so as to
-   * avoid unnecessary intermediate change.
+   * Merge the new ideal preference list with the delayed mapping that is calculated based on the
+   * delayed rebalance configurations.
+   * The method will prioritize the "active" preference list so as to avoid unnecessary transient
+   * state transitions.
    *
-   * @param newIdealPreferenceList  the ideal mapping that was calculated based on the current instance status
-   * @param newActivePreferenceList the mapping that was calculated based on the delayed instance status
+   * @param newIdealPreferenceList  the ideal mapping that was calculated based on the current
+   *                                instance status
+   * @param newDelayedPreferenceList the delayed mapping that was calculated based on the delayed
+   *                                 instance status
    * @param liveEnabledInstances    list of all the nodes that are both alive and enabled.
    * @param minActiveReplica        the minimum replica count to ensure a valid mapping.
    *                                If the active list does not have enough replica assignment,
@@ -172,25 +177,25 @@ public class DelayedRebalanceUtil {
    */
   public static Map<String, List<String>> getFinalDelayedMapping(
       Map<String, List<String>> newIdealPreferenceList,
-      Map<String, List<String>> newActivePreferenceList, Set<String> liveEnabledInstances,
+      Map<String, List<String>> newDelayedPreferenceList, Set<String> liveEnabledInstances,
       int minActiveReplica) {
     Map<String, List<String>> finalPreferenceList = new HashMap<>();
     for (String partition : newIdealPreferenceList.keySet()) {
       List<String> idealList = newIdealPreferenceList.get(partition);
-      List<String> activeList = newActivePreferenceList.get(partition);
+      List<String> delayedIdealList = newDelayedPreferenceList.get(partition);
 
       List<String> liveList = new ArrayList<>();
-      for (String ins : activeList) {
+      for (String ins : delayedIdealList) {
         if (liveEnabledInstances.contains(ins)) {
           liveList.add(ins);
         }
       }
 
       if (liveList.size() >= minActiveReplica) {
-        finalPreferenceList.put(partition, activeList);
+        finalPreferenceList.put(partition, delayedIdealList);
       } else {
         List<String> candidates = new ArrayList<>(idealList);
-        candidates.removeAll(activeList);
+        candidates.removeAll(delayedIdealList);
         for (String liveIns : candidates) {
           liveList.add(liveIns);
           if (liveList.size() >= minActiveReplica) {
@@ -210,7 +215,7 @@ public class DelayedRebalanceUtil {
    * @param replicaCount the expected active replica count.
    * @return the expected minimum active replica count that is required
    */
-  public static  int getMinActiveReplica(IdealState idealState, int replicaCount) {
+  public static int getMinActiveReplica(IdealState idealState, int replicaCount) {
     int minActiveReplicas = idealState.getMinActiveReplicas();
     if (minActiveReplicas < 0) {
       minActiveReplicas = replicaCount;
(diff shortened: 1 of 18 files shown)


jiajunwang:
Test to be added.
@dasahcc , Please take a look at this draft to verify if I'm doing the delayed rebalance correctly.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,215 @@
...
+
+/**
+ * The util to support delayed rebalance logic.
+ */
+public class DelayedRebalanceUtil {
+  /**
+   * @return the rebalance delay based on the ClusterConfig configurations.
+   * Resource level configuration will be ignored.
+   */
+  public static long getRebalanceDelay(ClusterConfig clusterConfig) {
```
narendly: Nit: can you put this getRebalanceDelay and the other getRebalanceDelay together so it's easier to see?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,215 @@
...
+    return getRebalanceDelay(null, clusterConfig);
+  }
+
+  /**
+   * @return true if delay rebalance is configured and enabled in the ClusterConfig configurations.
+   * Resource level configuration will be ignored.
+   */
+  public static boolean isDelayRebalanceEnabled(ClusterConfig clusterConfig) {
```
narendly: Put this right next to the other isDelayRebalanceEnabled?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,215 @@
...
+    return isDelayRebalanceEnabled(null, clusterConfig);
+  }
+
+  /**
+   * @return all active instances (live instances plus offline-yet-active instances) while considering delay rebalance configurations.
```
narendly: format this

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,215 @@
...
+   */
+  public static Set<String> getActiveInstances(Set<String> allNodes, Set<String> liveEnabledNodes,
```
narendly: Put this next to the other getActiveInstances

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * The util to support delayed rebalance logic.
+ */
+public class DelayedRebalanceUtil {
```
narendly: General comment on this util class:

Tip: Instead of passing in null as a parameter, just create another method (overloaded) without that parameter. That way you get to avoid passing nulls and the caller doesn't have to pass nulls around.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  /**
+   * Get the minimum active replica count threshold that allows delayed rebalance.
+   *
+   * @param idealState      the resource Ideal State
+   * @param replicaCount the expected active replica count.
+   * @return the expected minimum active replica count that is required
+   */
+  public static  int getMinActiveReplica(IdealState idealState, int replicaCount) {
```
narendly: This method signature has two spaces between static and int.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -82,7 +85,7 @@
   public WagedRebalancer(HelixManager helixManager) {
     this(
         // TODO init the metadata store according to their requirement when integrate,
-        // or change to final static method if possible.
+        //  or change to final static method if possible.
```
narendly: Nit: why a space?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -139,6 +141,11 @@ protected WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
 
     // Calculate the target assignment based on the current cluster status.
     Map<String, IdealState> newIdealStates = computeBestPossibleStates(clusterData, resourceMap);
+    // Apply user-defined preference list after the calculation.
+    // Note the user-defined list will be applied as an overwritten to the final result, instead of
+    // part of the rebalance algorithm.
+    newIdealStates.entrySet().stream().forEach(idealStateEntry -> applyUserDefinePreferenceList(
```
narendly: "applyUserDefinedPreferenceList"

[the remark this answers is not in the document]
narendly: +1 curious as well

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -349,4 +368,63 @@ private IdealState generateIdealStateWithAssignment(String resourceName,
...
+                  newActiveIdealState.getPreferenceLists(), activeInstances, DelayedRebalanceUtil
+                      .getMinActiveReplica(currentIdealState,
+                          currentIdealState.getReplicaCount(delayedActiveNodes.size())));
+          is.setPreferenceLists(finalPreferenceList);
+        }
+      }
+    }
+  }
+
+  private void applyUserDefinePreferenceList(ResourceConfig resourceConfig, IdealState idealState) {
```
narendly: "applyUserDefinePreferenceList"

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -349,4 +368,63 @@ private IdealState generateIdealStateWithAssignment(String resourceName,
...
+    if (resourceConfig != null) {
+      Map<String, List<String>> userDefinedPreferenceList = resourceConfig.getPreferenceLists();
+      if (!userDefinedPreferenceList.isEmpty()) {
+        LOG.info("Using user defined preference list for partitions: " + userDefinedPreferenceList
```
narendly: Wouldn't this be too verbose?

Wouldn't "Using user defined preference list for partitions" be enough for our purposes? Because we would have the preference list in ResourceConfig.

narendly:
Reminder to add tests!

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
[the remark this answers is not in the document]
jiajunwang: In the calculation, the assignment will be persisted into the AssignmentMetadata store. This will impact the calculation. If we want to treat the user-defined list as a temporary config, we shall not persist it.
It is debatable. So let me what's your thought here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
jiajunwang: This is mainly for avoiding duplicate code. I thought about this, I think we can do it reverse way. The method takes both instance config and cluster config shall call the method takes cluster config only. Then do an additional check based on that. I will change it.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: Not format the code yet.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: Makes sense. Will change.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
narendly: Either way there's some duplication. If you absolutely have to have duplication, we'd like to make it as clean as possible :)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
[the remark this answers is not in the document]
jiajunwang: Talked with Junkai offline, we agreed to keep the current design.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,269 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * The util for supporting delayed rebalance logic.
+ */
+public class DelayedRebalanceUtil {
+  private static final Logger LOG = LoggerFactory.getLogger(DelayedRebalanceUtil.class);
+
+  private static RebalanceScheduler _rebalanceScheduler = new RebalanceScheduler();
```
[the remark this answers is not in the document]
i3wangyi: Also, by using the class variable, you essentially only have one instance of the RebalanceScheduler across the entire application. Doing operations like scheduling or removing are in fact write/read operations, have you considered concurrency issues? If so, please specify it if you really want just one singleton instance.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -216,9 +231,26 @@ public void close() {
...
+          resourceStatePriorityMap,
+          getBaselineAssignment(_assignmentMetadataStore, currentStateOutput,
+              resourceMap.keySet()));
     }
-    return finalIdealState;
+    // Replace the assignment if user-defined preference list is configured.
+    // Note the user-defined list is intentionally applied to the final mapping after calculation.
+    // This is to avoid persisting it into the assignment store, which impacts the long term
+    // assignment evenness and partition movements.
+    finalIdealStateMap.entrySet().stream().forEach(
```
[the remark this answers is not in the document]
jiajunwang: I tried your suggestion. In short, I partially agree with you, but I think the current implementation is safer.

Major concern is that the change will make the idea state calculation logic separated. More specifically, the current code ensures the preference lists in the IdealState is calculated by the computeBestPossibleStates() only. And the later operation won't change the preference lists anymore. If we break this assumption, although there won't be any immediate issue, the future devs may be confused where they can get the finalized preference lists.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -196,17 +193,35 @@ public void close() {
       clusterChanges.putIfAbsent(HelixConstants.ChangeType.CLUSTER_CONFIG, Collections.emptySet());
     }
 
+    Set<String> activeNodes = new HashSet<>(clusterData.getEnabledLiveInstances());
```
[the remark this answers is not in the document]
jiajunwang: activeNodes could be overwritten to the delayed active nodes. So the name enabledLiveInstances here is not correct.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -196,17 +193,35 @@ public void close() {
...
+      delayedRebalanceSchedule(clusterData, delayedActiveNodes);
+      activeNodes = delayedActiveNodes;
+    }
+
     Map<String, ResourceAssignment> newAssignment =
-        partialRebalance(clusterData, clusterChanges, resourceMap, currentStateOutput);
+        partialRebalance(clusterData, clusterChanges, resourceMap, activeNodes,
+            currentStateOutput);
 
+    Map<String, Map<String, Integer>> resourceStatePriorityMap = new HashMap<>();
```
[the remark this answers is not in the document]
jiajunwang: Sure

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -196,17 +193,35 @@ public void close() {
...
+    if (DelayedRebalanceUtil.isDelayRebalanceEnabled(clusterConfig)) {
+      // If delayed rebalance is enabled, rebalance with the delayed active instance list.
+      Set<String> delayedActiveNodes = DelayedRebalanceUtil
+          .getActiveInstances(clusterData.getAllInstances(), activeNodes,
+              clusterData.getInstanceOfflineTimeMap(), clusterData.getLiveInstances().keySet(),
+              clusterData.getInstanceConfigMap(), clusterConfig);
+      // Schedule delayed rebalance in case no later cluster event happens when the delay time
+      // window passes.
+      delayedRebalanceSchedule(clusterData, delayedActiveNodes);
+      activeNodes = delayedActiveNodes;
```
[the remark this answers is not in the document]
jiajunwang: It will be simplified.
Actually, found one potential issue while I look into the code : )

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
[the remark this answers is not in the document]
jiajunwang: Essentially, I want to keep the same behavior as the current rebalancer. This logic is not nearly introduced.
I share the same concern, let's fix them in a separate PR.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,269 @@
...
+      delayTime = clusterConfig.getRebalanceDelayTime();
+    }
+    return delayTime;
+  }
+
+  /**
+   * @return all active instances (live instances plus offline-yet-active instances) while
+   * considering cluster delay rebalance configurations.
+   */
+  public static Set<String> getActiveInstances(Set<String> allNodes, Set<String> liveEnabledNodes,
```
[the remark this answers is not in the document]
jiajunwang: Sure. I will try to change all that I've touched. But I guess there could be more.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,269 @@
...
+  public static boolean isDelayRebalanceEnabled(ClusterConfig clusterConfig) {
+    long delay = clusterConfig.getRebalanceDelayTime();
+    return (delay > 0 && clusterConfig.isDelayRebalaceEnabled());
+  }
+
+  /**
+   * @return true if delay rebalance is configured and enabled in Resource IdealState and the
+   * ClusterConfig configurations.
+   */
+  public static boolean isDelayRebalanceEnabled(IdealState idealState,
```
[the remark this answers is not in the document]
jiajunwang: Actually, all methods in the util class have a similar issue. This is mainly because we do have 2 sets of logic due to the existing rebalancers. I think the comments of the method is clear enough for who read them.
Regarding which one to use, it depends on the usage. There is no absolute correctness.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,269 @@
...
+      ClusterConfig clusterConfig) {
+    long delay = getRebalanceDelay(idealState, clusterConfig);
+    return (delay > 0 && idealState.isDelayRebalanceEnabled() && clusterConfig
+        .isDelayRebalaceEnabled());
+  }
+
+  /**
+   * @return the rebalance delay based on Resource IdealState and the ClusterConfig configurations.
+   */
+  public static long getRebalanceDelay(IdealState idealState, ClusterConfig clusterConfig) {
```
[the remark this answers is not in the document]
jiajunwang: According to google:

de·lay
noun
a period of time by which something is late or postponed.

And about the 2nd problem, I used to have that one. But the content will be as simple as:
clusterConfig.getRebalanceDelayTime();
There is no need to have a util function for this.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+    return getActiveInstances(allNodes, liveEnabledNodes, instanceOfflineTimeMap, liveNodes,
+        instanceConfigMap, clusterConfig.getRebalanceDelayTime(), clusterConfig);
+  }
+
+  /**
+   * @return all active instances (live instances plus offline-yet-active instances) while
+   * considering cluster and the resource delay rebalance configurations.
+   */
+  public static Set<String> getActiveInstances(Set<String> allNodes, IdealState idealState,
```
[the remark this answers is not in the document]
jiajunwang: Not true, there is a branch inside. The logic is different. So one way is if-else in the code, which will require additional boolean to differentiate the conditions. And this boolean is not closely related to the method logic. So I don't like that style.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/DelayedRebalanceUtil.java
```diff
@@ -0,0 +1,269 @@
...
+import org.apache.helix.model.IdealState;
+import org.apache.helix.model.InstanceConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * The util for supporting delayed rebalance logic.
+ */
+public class DelayedRebalanceUtil {
```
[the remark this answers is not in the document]
jiajunwang: This util is not newly added. 80% of the logics have been covered by existing tests. The newly added 20% are coved by the newly added tests:

TestDelayedWagedRebalance
TestDelayedWagedRebalanceWithDisabledInstance
TestDelayedWagedRebalanceWithRackaware

Adding unit tests is a good idea. But not technically necessary because our integration test s are very complete. And to be clear, I'm not against the idea of adding the unit test. Just don't think it is necessary for the PR. All code has been covered in the new test cases.

jiajunwang:
I merged the code and re-run the tests.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -216,9 +226,26 @@ public void close() {
             "Fail to calculate the new IdealState for resource: " + resourceName,
             HelixRebalanceException.Type.INVALID_CLUSTER_STATUS, ex);
       }
-      finalIdealState.put(resourceName, newIdeaState);
+      finalIdealStateMap.put(resourceName, newIdeaState);
```
narendly: "newIdealState"

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/TestWagedRebalancer.java
```diff
@@ -112,7 +112,8 @@ protected ResourceControllerDataProvider setupClusterDataCache() throws IOExcept
   @Test
   public void testRebalance() throws IOException, HelixRebalanceException {
     _metadataStore.clearMetadataStore();
-    WagedRebalancer rebalancer = new WagedRebalancer(_metadataStore, _algorithm);
+    WagedRebalancer rebalancer =
+        new WagedRebalancer(_metadataStore, _algorithm, new DelayedAutoRebalancer(), null);
```
narendly: Create a constructor without the last field?

helix-core/src/test/java/org/apache/helix/tools/TestClusterVerifier.java
```diff
@@ -125,19 +125,42 @@ public void testDisablePartitionAndStopInstance() throws InterruptedException {
...
     Assert.assertTrue(strictMatchVerifier.verify(10000));
 
     // Disable partition for 1 instance, then Full-Auto ExternalView should not match IdealState.
-    _admin.enablePartition(false, _clusterName, _participants[0].getInstanceName(), FULL_AUTO_RESOURCES[0],
-        Lists.newArrayList(FULL_AUTO_RESOURCES[0] + "_0"));
+    _admin.enablePartition(false, _clusterName, _participants[0].getInstanceName(),
+        FULL_AUTO_RESOURCES[0], Lists.newArrayList(FULL_AUTO_RESOURCES[0] + "_0"));
     Thread.sleep(1000);
-    Assert.assertFalse(strictMatchVerifier.verify(3000));
+    Assert.assertTrue(strictMatchVerifier.verify(3000));
```
narendly: Um? Was there a change in the test logic?

jiajunwang: Yes, the previous verifier logic was not correct. It was comparing the list fields instead of mapping fields. We should depend on the mapping fields only.

jiajunwang:
Thanks for the reviewing, @narendly 
In this PR, I intentionally keep the delay logic untouched. Just moving to the util for sharing logic. The delay rebalance logic does not require any update, the only thing we need to modify is how the WAGED rebalancer uses the delay rebalance function.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: Good catch!

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/TestWagedRebalancer.java
jiajunwang: Sure.

jiajunwang:

PR #476, opened 2019-09-14
Adjust the expected replica count according to fault zone count.

jiajunwang:
### Issues

#402

### Description

The rebalancer should determine the expected replica count according to the fault zone instead of the node count only.

### Tests

Modified the existing cluster model tests.

NA, will run mvn test if check-in after the integration PR submitted.

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
@@ -62,10 +62,15 @@ public class ClusterModelProvider {
       Map<HelixConstants.ChangeType, Set<String>> clusterChanges,
       Map<String, ResourceAssignment> baselineAssignment,
       Map<String, ResourceAssignment> bestPossibleAssignment) {
+    // Construct all the assignable nodes and initialize with the allocated replicas.
+    Set<AssignableNode> assignableNodes =
+        parseAllNodes(dataProvider.getClusterConfig(), dataProvider.getInstanceConfigMap(),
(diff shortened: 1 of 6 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
```diff
@@ -74,10 +79,9 @@ public static ClusterModel generateClusterModel(ResourceControllerDataProvider d
         findToBeAssignedReplicas(replicaMap, clusterChanges, activeInstances,
             bestPossibleAssignment, allocatedReplicas);
 
-    // Construct all the assignable nodes and initialize with the allocated replicas.
-    Set<AssignableNode> assignableNodes =
-        parseAllNodes(dataProvider.getClusterConfig(), dataProvider.getInstanceConfigMap(),
-            activeInstances, allocatedReplicas);
+    // Update the allocated replicas to the assignable nodes.
+    assignableNodes.stream().forEach(node -> node.assignNewBatch(
+        allocatedReplicas.getOrDefault(node.getInstanceName(), Collections.emptySet())));
```
i3wangyi: nit: can we simplify just using get method, it's because all information is in the same method and you should know if  exists or not.

jiajunwang: Not necessary. If the node is newly added, there won't an existing assignment. And allocatedReplicas.get(node.getInstanceName()) will be null.

jiajunwang:

Issue #482, opened 2019-09-18
Improve the WAGED rebalancer soft constraint design


jiajunwang:
Our current implementation is based on POC work, which returns the evaluated score based on the ratio between current usage versus the estimated usage.
However, the estimated usage could be difficult to derive. For example, tags could impact expected usage.
In order to address these corner cases, we need to revisit all the soft constraint definition.
The requirement is basically:
1. Continous distribution.
2. Normalized to a [0.0 ~1.0] number.
3. Any two estimated scores should be comparable and different if the two potential assignment will cause different usages.

i3wangyi:
$$f(x) = 1/x$$ is likely to be the candidate function
![2560px-Hyperbola_one_over_x svg](https://user-images.githubusercontent.com/2618776/65187929-f7f03000-da21-11e9-9d7c-28e34fcbf8ac.png)

i3wangyi:
The simple inverse function has an issue where max value is not constrained.
I found the logistic regression more aligned with our requirement:
https://en.wikipedia.org/wiki/Logistic_function

![640px-Logistic-curve svg](https://user-images.githubusercontent.com/2618776/65189425-c1b4af80-da25-11e9-830e-9686ee5f81fc.png)

i3wangyi:
A better alternative function is $$f(x) = tanh(1/x)$$

The function plot looks like 
![save](https://user-images.githubusercontent.com/2618776/65278274-a064c980-dae0-11e9-8e57-8c55bf09679a.png)

It satisfies 
- continues distribution between (0, 1]
- The value decreases when x >= 0
- When x == 0, the value is infinitely close to 1 

The X in soft constraint mostly defines the usage of some aspect, and the lower usage is generally encouraged and higher usage is discouraged.

jiajunwang:
Which calculation is faster? And I would prefer sigmoid. I think it differentiates better when assigned partition count is a small number. And that is our major concerns.
We won't handle very full clusters usually.

i3wangyi:
<img src="https://latex.codecogs.com/svg.latex?\Large&space;tanh(x) = \frac{e^{2x} - 1}{e^{2x} + 1}}" title="tanh(x) = \frac{e^{2x} - 1}{e^{2x} + 1}" />

i3wangyi:
$$f(x) = tanh( 1 / 10x)$$
![save (1)](https://user-images.githubusercontent.com/2618776/65280418-2256f180-dae5-11e9-8799-d588167e6478.png)
[...]

PR #488, opened 2019-09-23
Implement soft constraint based on f(x) = tanh(1/x)

i3wangyi:
### Issues

(#482)
### Description

Pros:
The function squashes any x > 0 to a normalized range between (0, 1]
When x limits to 0, the f(x) = 1
No need to give any estimations about max capacity, max allowed partitions count, etc

Cons:
- It adds one additional parameter \alpha f(x) = tanh(1/(\alpha * x))
that changes the slope of plot function; Needs to tune the parameter as well as the weight importances between soft constraints

### Tests

[ERROR] testTagSetIncorrect(org.apache.helix.integration.TestAlertingRebalancerFailure)  Time elapsed: 180.143 s  <<< FAILURE!
        at org.apache.helix.integration.TestAlertingRebalancerFailure.checkResourceBestPossibleCalFailureState(TestAlertingRebalancerFailure.java:310)
        at org.apache.helix.integration.TestAlertingRebalancerFailure.testTagSetIncorrect(TestAlertingRebalancerFailure.java:174)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestWagedRebalance.testDisableInstance:225->validate:427->validateIsolation:439 expected:<3> but was:<4>
[INFO] 
[ERROR] Tests run: 995, Failures: 2, Errors: 0, Skipped: 3
[INFO] 
[INFO] Reactor Summary for Apache Helix 0.9.2-SNAPSHOT:
[INFO] 
[INFO] Apache Helix ....................................... SUCCESS [  0.480 s]
[INFO] Apache Helix :: Core ............................... FAILURE [  01:12 h]
[INFO] Apache Helix :: Admin Webapp ....................... SKIPPED
[INFO] Apache Helix :: Restful Interface .................. SKIPPED
[INFO] Apache Helix :: HelixAgent ......................... SKIPPED
[INFO] Apache Helix :: Recipes ............................ SKIPPED
[INFO] Apache Helix :: Recipes :: Rabbitmq Consumer Group . SKIPPED
[INFO] Apache Helix :: Recipes :: Rsync Replicated File Store SKIPPED
[INFO] Apache Helix :: Recipes :: distributed lock manager  SKIPPED
[INFO] Apache Helix :: Recipes :: distributed task execution SKIPPED
[INFO] Apache Helix :: Recipes :: service discovery ....... SKIPPED
[INFO] BUILD FAILURE
[INFO] Total time:  01:12 h

Passed the failing test sometimes when it's run standalone. The `TestWagedRebalancer` class is a flaky test.

### Commits

### Documentation

### Code Quality


i3wangyi:
@jiajunwang @narendly, this is the PR that optimizes the soft constraint function as we've reached an agreement with the meeting today.

The first several commits, let's focus on reviewing and refining the big picture of how the scoring method should work: interface, detailed tuning parameter setup, etc.

Once the big picture is agreed, I can focus on the detail implementations: unit test refining, etc.

The PR won't address the dynamical configurations of the soft constraints' weights as it's a different topic.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -25,46 +25,29 @@
...
- * of [minScore, maxScore]
+ * of (0, 1]
  * The higher the score, the better the assignment; Intuitively, the assignment is encouraged.
  * The lower score the score, the worse the assignment; Intuitively, the assignment is penalized.
  */
 abstract class SoftConstraint {
-  private float _maxScore = 1000f;
-  private float _minScore = -1000f;
+  // The hyper parameter in the normalize function; Child class can override and customize the value
+  private float alpha = 1.0f;
```
jiajunwang: As we discussed, the alpha might be determined by the cluster statue, right? For example, depends on the estimated partition count. Or you suggest we don't estimate but always use the fixed alpha?
Please comment on your plan.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -86,11 +69,11 @@ float getAssignmentNormalizedScore(AssignableNode node, AssignableReplica replic
...
   /**
-   * The default scaler function that squashes any score within (min_score, max_score) to (0, 1);
+   * The default normalize function f(x) = tanh(1/(\alpha * x)) that squashes any score >= 0 to (0, 1]
+   * The hyper parameter determines the converging rate of scores to 0 when X gets larger
    * Child class could override the method and customize the method on its own
-   * @return The MinMaxScaler instance by default
    */
   NormalizeFunction getNormalizeFunction() {
-    return (score) -> (score - getMinScore()) / (getMaxScore() - getMinScore());
+    return score -> (score == 0f? 1f: (float) Math.tanh(1 / (score * this.alpha)));
```
jiajunwang: So we assume all the soft constraints will use this formula? I don't think it will fit the partition movement constraint.
My suggestion is that, keep the scale (max and min), only change the constraints individually which we are calculating the percentage. Those are the applicable ones.

i3wangyi: The method is in parent class and child class like partition movement can override and have its own implementation without problems. It's reflected in my latest commit.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
i3wangyi: We can use fixed alpha =  1.0f at beginning, making it a private field gives us the ability to set the value from constructor if in the future we have the idea of how to tune it.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
```diff
@@ -38,59 +37,55 @@
...
-  private static final float MAX_SCORE = 1f;
-  private static final float MIN_SCORE = 0f;
-  //TODO: these factors will be tuned based on user's preference
+  // TODO: these factors will be tuned based on user's preference
   // This factor indicates the default score that is evaluated if only partition allocation matches
   // (states are different).
-  private static final float ALLOCATION_MATCH_FACTOR = 0.5f;
   // This factor indicates the contribution of the Baseline assignment matching to the final score.
-  private static final float BASELINE_MATCH_FACTOR = 0.25f;
+  private float _baseLineMatchFactor = 0.25f;
```
narendly: Could you please add a little more explanation about this? What does it mean for baselinematchfactor to be 0? What does it mean for it to be 1.0f?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ResourceTopStateAntiAffinityConstraint.java
```diff
@@ -29,25 +29,11 @@
...
-      return (getMaxScore() + getMinScore()) / 2.0f;
-    }
-
-    int curTopPartitionCountForResource = node.getAssignedTopStatePartitionsCount();
-    int doubleMaxTopStateCount = 2 * clusterContext.getEstimatedMaxTopStateCount();
-
-    return Math.max(
-        ((float) doubleMaxTopStateCount - curTopPartitionCountForResource) / doubleMaxTopStateCount,
-        0);
+    // if replica is top state, there're 1 more top state partition hosted on the instance if the assignment happens
```
narendly: there is?

Also why is it necessarily 1 or 0? I get that if it's not a top state replica, it should be 0. But if it a top state replica, anything above 0 would work? I'm just wondering why it's hard-coded at 1.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
narendly: 0.0f / 1.0f?

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/constraints/TestInstancePartitionsCountConstraint.java
```diff
@@ -46,18 +46,16 @@ public void testWhenInstanceIsIdle() {
   @Test
   public void testWhenInstanceIsFull() {
     when(_testNode.getAssignedReplicaCount()).thenReturn(10);
-    when(_clusterContext.getEstimatedMaxPartitionCount()).thenReturn(10);
     float score =
         _constraint.getAssignmentNormalizedScore(_testNode, _testReplica, _clusterContext);
-    Assert.assertEquals(score, 0.5f);
+    Assert.assertEquals(score, 0.099667996f);
```
narendly: Is this the calculation result of f(X) = tanh(1/x)? could we just make a note here so that we have the context later on when we need to fix the test?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ResourceTopStateAntiAffinityConstraint.java
i3wangyi: It's because if the potential assignment pair being evaluated (replica, node) is chosen, it adds "one" more top state partitions + existing top state partitions on the instance.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
i3wangyi: Where is the 0.0f in the numerator coming from?

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/constraints/TestInstancePartitionsCountConstraint.java
i3wangyi: Yeah, the value looks weird when it appears in the unit test class, thanks I will mark it in the java doc

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
i3wangyi: The factor, in general, defines the goodness of the assignment if it matches with baseline assignment; Actually, it shouldn't be part of the PR (it's more about refining the existing partition movement) so I reverted this part of changes. If in the next PR of refining the implementation, I will address it and add more comments.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
jiajunwang: One question, if the score could be any number, how it becomes comparable after Math.tanh(1 / (score * this.alpha))?
I know this formula ensures the return number is between 0, 1. But for example, if ScoreA is between 0 - 100, and ScoreB is between 1000 - 500000, even after the tanh(), how you justify their value is comparable?

This is a real concern, according to your previous code if one cluster has 10K partitions, and later more resource added and the count is 100K, part of your evaluation scores might be completely different regarding the actual scale.

What I'm thinking is that try to ensure the sable scale of all soft constraints. So the alpha needs to be configured according to each constraint separately. For instance, partition count constraint, the alpha should be calculated by a function which takes the estimate partition or total partition number as an input.

This algorithm is very sensitive, so we need to clearly justify all the detail designs. Moreover, any logic other than the configuration part should be treated as the final version at this stage.

i3wangyi: I agree. But the purpose of this PR is not to define the right alpha for each soft constraint but the scores in the same range. In my understanding, the tuning part will resolve the correct coefficients and alphas if necessary. We don't necessarily need to have comparable scores between soft constraints and it'll always be a problem to argue if A constraint is comparable to B constraint because they're implemented differently and independently. The final tuned "coefficients" or "weights" will take the absolute values into considerations and evens out the uncomparable part. The end result is not to compare "absolute" weight between each soft constraint but to generate the "trustworthy" metrics about the learned coefficients like "R2 scores", "MSE", etc. That tells us e.g 80% chance with this set of weights configured, we could achieve an evenness degree of 0.5 something like that. We could discuss offline if you still think the absolute comparable score numbers matter.

i3wangyi: Synced with @jiajunwang offline, we agreed to use the "usage percentage" instead of the absolute value of the usage as the input to the normalization method.

For each soft constraint, the "usage percentage" is roughly calculated by 
`usage percent = absolute usage value / estimate max usage`

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -86,11 +60,11 @@ float getAssignmentNormalizedScore(AssignableNode node, AssignableReplica replic
...
   /**
-   * The default scaler function that squashes any score within (min_score, max_score) to (0, 1);
+   * The default normalize function f(x) = tanh(1/(\alpha * x)) that squashes any score >= 0 to (0, 1]
+   * The hyper parameter determines the converging rate of scores to 0 when X gets larger
    * Child class could override the method and customize the method on its own
-   * @return The MinMaxScaler instance by default
    */
   NormalizeFunction getNormalizeFunction() {
-    return (score) -> (score - getMinScore()) / (getMaxScore() - getMinScore());
+    return score -> (score == 0f ? 1f : (float) Math.tanh(1 / score));
```
jiajunwang: I tried this formula a little bit. So the range is (0, 1], as expected. One concern is that while I try Math.tanh((float)20), the value returned by Math lib is 1. That means any utilization between 0% and 5% would be evaluated the same by this implementation, right?
Can we make it more accurate, at least to 1% or even 0.1%. I think we can setup the alpha for this, right?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/InstancePartitionsCountConstraint.java
```diff
@@ -27,22 +27,14 @@
...
 
   @Override
   protected float getAssignmentScore(AssignableNode node, AssignableReplica replica,
       ClusterContext clusterContext) {
     float doubleEstimatedMaxPartitionCount = 2 * clusterContext.getEstimatedMaxPartitionCount();
-    float currentPartitionCount = node.getAssignedReplicaCount();
-    return Math.max((doubleEstimatedMaxPartitionCount - currentPartitionCount)
-        / doubleEstimatedMaxPartitionCount, 0);
+    // use the percentage as the input for normalization method
+    return node.getAssignedReplicaCount() / doubleEstimatedMaxPartitionCount;
```
jiajunwang: We won't need double estimate with tanh().

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
i3wangyi: I see. You'd like to have a more distinguished value distribution from 1% - 100%. I could leave the alpha and let each soft constraint that's focused on **usage percentage** to have a default alpha of something.

jiajunwang: Yeah, my feeling is that the constraint should be more sensitive internally so even one partition assignment would cause a difference. Actually, let's ensure the score would be different when there is more than 0.1% usage. And please comment on this in the code so if we want to adjust alpha later, we will have a clue.
Thanks!

i3wangyi: A simple alpha = 1000 is able to have a sensitivity of 0.1%. In practice, do we have need 0.1% sensitivity or 1% is fine?

jiajunwang: Let do 0.1%. I personally feel safer with that as a start point : )

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/constraints/TestInstancePartitionsCountConstraint.java
```diff
@@ -28,36 +28,69 @@
...
+  @Test
+  public void testWhenExceedUsage() {
+    when(_testNode.getAssignedReplicaCount()).thenReturn(20);
+    when(_clusterContext.getEstimatedMaxPartitionCount()).thenReturn(10);
+    float score = _constraint.getAssignmentScore(_testNode, _testReplica, _clusterContext);
+    float normalizedScore =
         _constraint.getAssignmentNormalizedScore(_testNode, _testReplica, _clusterContext);
-    Assert.assertEquals(score, 1.0f);
+    Assert.assertEquals(score, 2f);
+    Assert.assertEquals(normalizedScore, 4.9999997E-4f);
```
narendly: Probably a good idea to add some description in front of the method. Also if you could explain what these values all mean so that the reader could understand. For example, a 3rd person reviewing the code won't have a clue what "4.9999997E-4f" means here and why it has to be that value :)

i3wangyi: The value is just a record of the normalized score, we don't need to understand the absolute value it's meaningless. The unit test serves two purposes in the PR:
1. A record of pre-computed value given the input. So that if anyone touches the scoring method and gets things changed, he would be aware of it cause the output value is different.
2. The test method tests score when 0% usage, 1% usage, 50% usage, 100%, and 200% usage of some "aspect". From the tests, we can see the final normalized scores are decreasing and in range (0, 1] that's good enough.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ResourceTopStateAntiAffinityConstraint.java
```diff
@@ -29,25 +29,14 @@
...
-  private static final float MIN_SCORE = 0f;
-
-  ResourceTopStateAntiAffinityConstraint() {
-    super(MAX_SCORE, MIN_SCORE);
-  }
 
   @Override
   protected float getAssignmentScore(AssignableNode node, AssignableReplica replica,
       ClusterContext clusterContext) {
-    if (!replica.isReplicaTopState()) {
```
jiajunwang: This means if the replica is not with the top state, this constraint should not score differently. Or the node with more master will tend to have less replicas (any state) because of this constraint.

Please keep the main logic unchanged here. Just update how we calculate the percentage.

i3wangyi: Does it mean if a replica is not in top state, as long as the constraint returns the same value (the value could just be anything, e.g, 0), it will not affect the algorithm result. Only evaluate when the replica is top state, since it makes the “top state” situations worse for a node.

jiajunwang: The point is the constraint should return the same value across all the nodes if the input replica is not the top state. Meaning all 0, or some random numbers. However, your change will return the value depends on getAssignedTopStatePartitionsCount(), this value is different from node to node.

i3wangyi: Return 0 as the default score to make it easier and added some comments on top of it. Merged with the latest changes.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/SoftConstraint.java
```diff
@@ -86,11 +70,12 @@ float getAssignmentNormalizedScore(AssignableNode node, AssignableReplica replic
...
-   * The default scaler function that squashes any score within (min_score, max_score) to (0, 1);
+   * The default normalize function f(x) = tanh(1/(\alpha * x)) that squashes any score >= 0 to (0,
+   * 1]
+   * The hyper parameter determines the converging rate of scores to 0 when X gets larger
    * Child class could override the method and customize the method on its own
-   * @return The MinMaxScaler instance by default
    */
   NormalizeFunction getNormalizeFunction() {
-    return (score) -> (score - getMinScore()) / (getMaxScore() - getMinScore());
+    return score -> (score == 0f ? 1f : (float) Math.tanh(1 / (this.alpha * score)));
```
jiajunwang: nit: this.alpha -> alpha ?

i3wangyi:
The PR is ready to merge into WagedRebalancer branch, approved by @jiajunwang

PR #490, opened 2019-09-25
Add latency metric components for WAGED rebalancer

### LINKED ISSUE #495: Add metrics for WAGED rebalancer
We need to add metrics for WAGED rebalancer. Since this rebalancer is stateful, we need to closely monitor latency-related metrics.

narendly:
### Issues

Fixes #495 

### Description

This is a metric framework to be used for analyzing the performance of WAGED rebalancer. It also adds a series of latency-related metrics. These metrics will be crucial in identifying performance bottlenecks in different stages of the assignment computation.

### Tests

**TestWagedRebalancerMetrics**

[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestEnableCompression.testEnableCompressionResource:109 expected:<true> but was:<false>
[ERROR]   TestControllerLeadershipChange.testMissingTopStateDurationMonitoring:119 expected:<true> but was:<false>
[ERROR] org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceTopoplogyAwareDisabled.test(org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceTopoplogyAwareDisabled)
[INFO]   Run 1: PASS
[ERROR]   Run 2: TestCrushAutoRebalanceTopoplogyAwareDisabled.test:65->TestCrushAutoRebalanceNonRack.test:159->TestCrushAutoRebalanceNonRack.validateIsolation:300 expected:<3> but was:<4>
[INFO] 
[ERROR] org.apache.helix.integration.rebalancer.PartitionMigration.TestWagedRebalancerMigration.testMigrateToWagedRebalancerWhileExpandCluster(org.apache.helix.integration.rebalancer.PartitionMigration.TestWagedRebalancerMigration)
[INFO]   Run 1: PASS
[INFO]   Run 2: PASS
[INFO]   Run 3: PASS
[ERROR]   Run 4: TestWagedRebalancerMigration.testMigrateToWagedRebalancerWhileExpandCluster » ThreadTimeout
[INFO]   Run 5: PASS
[INFO]   Run 6: PASS
[INFO] 
[ERROR]   TestWagedRebalance.testDisableInstance:225->validate:427->validateIsolation:439 expected:<3> but was:<4>
[INFO] 
[ERROR] Tests run: 989, Failures: 5, Errors: 0, Skipped: 8
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:12 h
[INFO] Finished at: 2019-10-03T13:30:41-07:00

--------------------------------
Individual runs: all tests pass when run individually.

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
@@ -19,7 +19,6 @@ package org.apache.helix.controller.rebalancer.waged;
  * under the License.
  */
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -106,9 +105,29 @@ public class WagedRebalancer {
         helixManager, metricCollector);
   }
 
+  /**
+   * This constructor will use null for HelixManager and MetricCollector. With null HelixManager,
+   * the rebalancer will rebalance solely based on CurrentStates. With null MetricCollector, the
+   * rebalancer will not emit JMX metrics.
+   * @param assignmentMetadataStore
+   * @param algorithm
+   * @param mappingCalculator
+   */
   protected WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
       RebalanceAlgorithm algorithm, MappingCalculator mappingCalculator) {
     this(assignmentMetadataStore, algorithm, mappingCalculator, null, null);
+  }
+
+  /**
+   * This constructor will use null for HelixManager and MetricCollector. With null HelixManager,
+   * the rebalancer will rebalance solely based on CurrentStates.
+   * @param assignmentMetadataStore
+   * @param algorithm
+   * @param metricCollector
+   */
+  protected WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
+      RebalanceAlgorithm algorithm, MetricCollector metricCollector) {
+    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer(), null, metricCollector);
   }
 
   private WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
@@ -126,12 +145,6 @@ public class WagedRebalancer {
     // allow rebalancer to proceed
     _metricCollector =
         metricCollector == null ? new WagedRebalancerMetricCollector() : metricCollector;
-  }
-
-  @VisibleForTesting
-  protected WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
-      RebalanceAlgorithm algorithm, MetricCollector metricCollector) {
-    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer(), null, metricCollector);
   }
 
   // Release all the resources.
@@ -499,7 +512,12 @@ public class WagedRebalancer {
     Map<String, ResourceAssignment> currentBestAssignment = Collections.emptyMap();
     if (assignmentMetadataStore != null) {
       try {
+        LatencyMetric stateReadLatency = _metricCollector.getMetric(
+            WagedRebalancerMetricCollector.WagedRebalancerMetricNames.StateReadLatencyGauge.name(),
+            LatencyMetric.class);
+        stateReadLatency.startMeasuringLatency();
         currentBestAssignment = assignmentMetadataStore.getBestPossibleAssignment();
+        stateReadLatency.endMeasuringLatency();
       } catch (HelixException ex) {
         // Report error. and use empty mapping instead.
         LOG.error("Failed to get the current best possible assignment.", ex);
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
@@ -19,6 +19,7 @@ package org.apache.helix.controller.rebalancer.waged;
  * under the License.
  */
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -48,6 +49,9 @@ import org.apache.helix.model.Partition;
 import org.apache.helix.model.Resource;
 import org.apache.helix.model.ResourceAssignment;
 import org.apache.helix.model.ResourceConfig;
+import org.apache.helix.monitoring.metrics.MetricCollector;
+import org.apache.helix.monitoring.metrics.WagedRebalancerMetricCollector;
+import org.apache.helix.monitoring.metrics.model.LatencyMetric;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -64,10 +68,8 @@ public class WagedRebalancer {
   // When any of the following change happens, the rebalancer needs to do a global rebalance which
   // contains 1. baseline recalculate, 2. partial rebalance that is based on the new baseline.
   private static final Set<HelixConstants.ChangeType> GLOBAL_REBALANCE_REQUIRED_CHANGE_TYPES =
-      ImmutableSet.of(
-          HelixConstants.ChangeType.RESOURCE_CONFIG,
-          HelixConstants.ChangeType.CLUSTER_CONFIG,
-          HelixConstants.ChangeType.INSTANCE_CONFIG);
+      ImmutableSet.of(HelixConstants.ChangeType.RESOURCE_CONFIG,
+          HelixConstants.ChangeType.CLUSTER_CONFIG, HelixConstants.ChangeType.INSTANCE_CONFIG);
   // The cluster change detector is a stateful object.
   // Make it static to avoid unnecessary reinitialization.
   private static final ThreadLocal<ResourceChangeDetector> CHANGE_DETECTOR_THREAD_LOCAL =
@@ -76,6 +78,7 @@ public class WagedRebalancer {
   private final MappingCalculator<ResourceControllerDataProvider> _mappingCalculator;
   private final AssignmentMetadataStore _assignmentMetadataStore;
   private final RebalanceAlgorithm _rebalanceAlgorithm;
+  private MetricCollector _metricCollector;
 
   private static AssignmentMetadataStore constructAssignmentStore(HelixManager helixManager) {
     AssignmentMetadataStore assignmentMetadataStore = null;
@@ -90,7 +93,8 @@ public class WagedRebalancer {
   }
 
   public WagedRebalancer(HelixManager helixManager,
-      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preferences) {
+      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preferences,
+      MetricCollector metricCollector) {
     this(constructAssignmentStore(helixManager),
         ConstraintBasedAlgorithmFactory.getInstance(preferences),
         // Use DelayedAutoRebalancer as the mapping calculator for the final assignment output.
@@ -99,16 +103,17 @@ public class WagedRebalancer {
         // TODO abstract and separate the main mapping calculator logic from DelayedAutoRebalancer
         new DelayedAutoRebalancer(),
         // Helix Manager is required for the rebalancer scheduler
-        helixManager);
+        helixManager, metricCollector);
   }
 
   protected WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
       RebalanceAlgorithm algorithm, MappingCalculator mappingCalculator) {
-    this(assignmentMetadataStore, algorithm, mappingCalculator, null);
+    this(assignmentMetadataStore, algorithm, mappingCalculator, null, null);
   }
 
   private WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
-      RebalanceAlgorithm algorithm, MappingCalculator mappingCalculator, HelixManager manager) {
+      RebalanceAlgorithm algorithm, MappingCalculator mappingCalculator, HelixManager manager,
+      MetricCollector metricCollector) {
     if (assignmentMetadataStore == null) {
       LOG.warn("Assignment Metadata Store is not configured properly."
           + " The rebalancer will not access the assignment store during the rebalance.");
@@ -117,6 +122,16 @@ public class WagedRebalancer {
     _rebalanceAlgorithm = algorithm;
     _mappingCalculator = mappingCalculator;
     _manager = manager;
+    // If metricCollector is null, instantiate a version that does not register metrics in order to
+    // allow rebalancer to proceed
+    _metricCollector =
+        metricCollector == null ? new WagedRebalancerMetricCollector() : metricCollector;
+  }
+
+  @VisibleForTesting
+  protected WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
+      RebalanceAlgorithm algorithm, MetricCollector metricCollector) {
+    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer(), null, metricCollector);
   }
 
   // Release all the resources.
@@ -138,8 +153,7 @@ public class WagedRebalancer {
       Map<String, Resource> resourceMap, final CurrentStateOutput currentStateOutput)
       throws HelixRebalanceException {
     if (resourceMap.isEmpty()) {
-      LOG.warn("There is no resource to be rebalanced by {}",
-          this.getClass().getSimpleName());
+      LOG.warn("There is no resource to be rebalanced by {}", this.getClass().getSimpleName());
       return Collections.emptyMap();
     }
 
@@ -169,7 +183,6 @@ public class WagedRebalancer {
             newStateMap == null ? Collections.emptyMap() : newStateMap);
       }
     }
-
     LOG.info("Finish computing new ideal states for resources: {}",
         resourceMap.keySet().toString());
     return newIdealStates;
@@ -191,24 +204,25 @@ public class WagedRebalancer {
               return itemKeys;
             }));
 
+    // Perform Global Baseline Calculation
     if (clusterChanges.keySet().stream()
-        .anyMatch(changeType -> GLOBAL_REBALANCE_REQUIRED_CHANGE_TYPES.contains(changeType))) {
+        .anyMatch(GLOBAL_REBALANCE_REQUIRED_CHANGE_TYPES::contains)) {
       refreshBaseline(clusterData, clusterChanges, resourceMap, currentStateOutput);
       // Inject a cluster config change for large scale partial rebalance once the baseline changed.
       clusterChanges.putIfAbsent(HelixConstants.ChangeType.CLUSTER_CONFIG, Collections.emptySet());
     }
 
-    Set<String> activeNodes = DelayedRebalanceUtil
-        .getActiveNodes(clusterData.getAllInstances(), clusterData.getEnabledLiveInstances(),
-            clusterData.getInstanceOfflineTimeMap(), clusterData.getLiveInstances().keySet(),
-            clusterData.getInstanceConfigMap(), clusterData.getClusterConfig());
+    Set<String> activeNodes = DelayedRebalanceUtil.getActiveNodes(clusterData.getAllInstances(),
+        clusterData.getEnabledLiveInstances(), clusterData.getInstanceOfflineTimeMap(),
+        clusterData.getLiveInstances().keySet(), clusterData.getInstanceConfigMap(),
+        clusterData.getClusterConfig());
 
     // Schedule (or unschedule) delayed rebalance according to the delayed rebalance config.
     delayedRebalanceSchedule(clusterData, activeNodes, resourceMap.keySet());
 
+    // Perform partial rebalance
     Map<String, ResourceAssignment> newAssignment =
-        partialRebalance(clusterData, clusterChanges, resourceMap, activeNodes,
-            currentStateOutput);
+        partialRebalance(clusterData, clusterChanges, resourceMap, activeNodes, currentStateOutput);
 
     // <ResourceName, <State, Priority>>
     Map<String, Map<String, Integer>> resourceStatePriorityMap = new HashMap<>();
@@ -238,16 +252,15 @@ public class WagedRebalancer {
     // some delayed rebalanced assignments.
     if (!activeNodes.equals(clusterData.getEnabledLiveInstances())) {
       applyRebalanceOverwrite(finalIdealStateMap, clusterData, resourceMap, clusterChanges,
-          resourceStatePriorityMap,
-          getBaselineAssignment(_assignmentMetadataStore, currentStateOutput,
-              resourceMap.keySet()));
+          resourceStatePriorityMap, getBaselineAssignment(_assignmentMetadataStore,
+              currentStateOutput, resourceMap.keySet()));
     }
     // Replace the assignment if user-defined preference list is configured.
     // Note the user-defined list is intentionally applied to the final mapping after calculation.
     // This is to avoid persisting it into the assignment store, which impacts the long term
     // assignment evenness and partition movements.
-    finalIdealStateMap.entrySet().stream().forEach(
-        idealStateEntry -> applyUserDefinedPreferenceList(
+    finalIdealStateMap.entrySet().stream()
+        .forEach(idealStateEntry -> applyUserDefinedPreferenceList(
             clusterData.getResourceConfig(idealStateEntry.getKey()), idealStateEntry.getValue()));
 
     return finalIdealStateMap;
@@ -258,19 +271,31 @@ public class WagedRebalancer {
       Map<HelixConstants.ChangeType, Set<String>> clusterChanges, Map<String, Resource> resourceMap,
       final CurrentStateOutput currentStateOutput) throws HelixRebalanceException {
     LOG.info("Start calculating the new baseline.");
+    LatencyMetric globalBaselineCalcLatency = _metricCollector.getMetric(
+        WagedRebalancerMetricCollector.WagedRebalancerMetricNames.GlobalBaselineCalcLatencyGauge
+            .name(),
+        LatencyMetric.class);
+    globalBaselineCalcLatency.startMeasuringLatency();
+    // Read the baseline from metadata store
     Map<String, ResourceAssignment> currentBaseline =
         getBaselineAssignment(_assignmentMetadataStore, currentStateOutput, resourceMap.keySet());
+
     // For baseline calculation
     // 1. Ignore node status (disable/offline).
     // 2. Use the baseline as the previous best possible assignment since there is no "baseline" for
     // the baseline.
-    Map<String, ResourceAssignment> newBaseline =
-        calculateAssignment(clusterData, clusterChanges, resourceMap, clusterData.getAllInstances(),
-            Collections.emptyMap(), currentBaseline);
+    Map<String, ResourceAssignment> newBaseline = calculateAssignment(clusterData, clusterChanges,
+        resourceMap, clusterData.getAllInstances(), Collections.emptyMap(), currentBaseline);
 
+    // Write the new baseline to metadata store
     if (_assignmentMetadataStore != null) {
       try {
+        LatencyMetric writeLatency = _metricCollector.getMetric(
+            WagedRebalancerMetricCollector.WagedRebalancerMetricNames.StateWriteLatencyGauge.name(),
+            LatencyMetric.class);
+        writeLatency.startMeasuringLatency();
         _assignmentMetadataStore.persistBaseline(newBaseline);
+        writeLatency.endMeasuringLatency();
       } catch (Exception ex) {
         throw new HelixRebalanceException("Failed to persist the new baseline assignment.",
             HelixRebalanceException.Type.INVALID_REBALANCER_STATUS, ex);
@@ -278,7 +303,7 @@ public class WagedRebalancer {
     } else {
       LOG.debug("Assignment Metadata Store is empty. Skip persist the baseline assignment.");
     }
-
+    globalBaselineCalcLatency.endMeasuringLatency();
     LOG.info("Finish calculating the new baseline.");
   }
 
@@ -288,20 +313,34 @@ public class WagedRebalancer {
       Set<String> activeNodes, final CurrentStateOutput currentStateOutput)
       throws HelixRebalanceException {
     LOG.info("Start calculating the new best possible assignment.");
+    LatencyMetric partialRebalanceLatency = _metricCollector.getMetric(
+        WagedRebalancerMetricCollector.WagedRebalancerMetricNames.PartialRebalanceLatencyGauge
+            .name(),
+        LatencyMetric.class);
+    partialRebalanceLatency.startMeasuringLatency();
+    // TODO: Consider combining the metrics for both baseline/best possible?
+    // Read the baseline from metadata store
     Map<String, ResourceAssignment> currentBaseline =
         getBaselineAssignment(_assignmentMetadataStore, currentStateOutput, resourceMap.keySet());
-    Map<String, ResourceAssignment> currentBestPossibleAssignment =
-        getBestPossibleAssignment(_assignmentMetadataStore, currentStateOutput,
-            resourceMap.keySet());
-    Map<String, ResourceAssignment> newAssignment =
-        calculateAssignment(clusterData, clusterChanges, resourceMap, activeNodes, currentBaseline,
-            currentBestPossibleAssignment);
+
+    // Read the best possible assignment from metadata store
+    Map<String, ResourceAssignment> currentBestPossibleAssignment = getBestPossibleAssignment(
+        _assignmentMetadataStore, currentStateOutput, resourceMap.keySet());
+
+    // Compute the new assignment
+    Map<String, ResourceAssignment> newAssignment = calculateAssignment(clusterData, clusterChanges,
+        resourceMap, activeNodes, currentBaseline, currentBestPossibleAssignment);
 
     if (_assignmentMetadataStore != null) {
       try {
+        LatencyMetric writeLatency = _metricCollector.getMetric(
+            WagedRebalancerMetricCollector.WagedRebalancerMetricNames.StateWriteLatencyGauge.name(),
+            LatencyMetric.class);
+        writeLatency.startMeasuringLatency();
         // TODO Test to confirm if persisting the final assignment (with final partition states)
         // would be a better option.
         _assignmentMetadataStore.persistBestPossibleAssignment(newAssignment);
+        writeLatency.endMeasuringLatency();
       } catch (Exception ex) {
         throw new HelixRebalanceException("Failed to persist the new best possible assignment.",
             HelixRebalanceException.Type.INVALID_REBALANCER_STATUS, ex);
@@ -309,20 +348,20 @@ public class WagedRebalancer {
     } else {
       LOG.debug("Assignment Metadata Store is empty. Skip persist the baseline assignment.");
     }
-
+    partialRebalanceLatency.endMeasuringLatency();
     LOG.info("Finish calculating the new best possible assignment.");
     return newAssignment;
   }
 
   /**
    * Generate the cluster model based on the input and calculate the optimal assignment.
-   * @param clusterData                the cluster data cache.
-   * @param clusterChanges             the detected cluster changes.
-   * @param resourceMap                the rebalancing resources.
-   * @param activeNodes                the alive and enabled nodes.
-   * @param baseline                   the baseline assignment for the algorithm as a reference.
+   * @param clusterData the cluster data cache.
+   * @param clusterChanges the detected cluster changes.
+   * @param resourceMap the rebalancing resources.
+   * @param activeNodes the alive and enabled nodes.
+   * @param baseline the baseline assignment for the algorithm as a reference.
    * @param prevBestPossibleAssignment the previous best possible assignment for the algorithm as a
-   *                                   reference.
+   *          reference.
    * @return the new optimal assignment for the resources.
    */
   private Map<String, ResourceAssignment> calculateAssignment(
@@ -415,7 +454,7 @@ public class WagedRebalancer {
    * @param currentStateOutput
    * @param resources
    * @return The current baseline assignment. If record does not exist in the
-   * assignmentMetadataStore, return the current state assignment.
+   *         assignmentMetadataStore, return the current state assignment.
    * @throws HelixRebalanceException
    */
   private Map<String, ResourceAssignment> getBaselineAssignment(
@@ -424,7 +463,12 @@ public class WagedRebalancer {
     Map<String, ResourceAssignment> currentBaseline = Collections.emptyMap();
     if (assignmentMetadataStore != null) {
       try {
+        LatencyMetric stateReadLatency = _metricCollector.getMetric(
+            WagedRebalancerMetricCollector.WagedRebalancerMetricNames.StateReadLatencyGauge.name(),
+            LatencyMetric.class);
+        stateReadLatency.startMeasuringLatency();
         currentBaseline = assignmentMetadataStore.getBaseline();
+        stateReadLatency.endMeasuringLatency();
       } catch (HelixException ex) {
         // Report error. and use empty mapping instead.
         LOG.error("Failed to get the current baseline assignment.", ex);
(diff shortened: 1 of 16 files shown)


jiajunwang:
Why additional framework is necessary? Can we use the DynamicMBeanProvider or simply the regular MBean provider?

narendly:
@jiajunwang We are using the same framework. We extend those classes so that it's easier to reason about in Waged Rebalancer.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -97,24 +100,25 @@ public WagedRebalancer(HelixManager helixManager,
         // Mapping calculator will translate the best possible assignment into the applicable state
         // mapping based on the current states.
         // TODO abstract and separate the main mapping calculator logic from DelayedAutoRebalancer
-        new DelayedAutoRebalancer());
+        new DelayedAutoRebalancer(), helixManager.getClusterName());
   }
 
   private WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
-      RebalanceAlgorithm algorithm, MappingCalculator mappingCalculator) {
+      RebalanceAlgorithm algorithm, MappingCalculator mappingCalculator, String clusterName) {
```
jiajunwang: The private constructor can accept the MetricCollector directly. So you don't need to bother how the MetricCollector is created in the rebalancer class.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -97,24 +100,25 @@ public WagedRebalancer(HelixManager helixManager,
...
     _rebalanceAlgorithm = algorithm;
     _mappingCalculator = mappingCalculator;
+    _metricCollector = new MetricCollector(clusterName, "WagedRebalancer");
   }
 
   @VisibleForTesting
   protected WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
       RebalanceAlgorithm algorithm) {
-    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer());
+    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer(), "");
```
jiajunwang: If the parameter is MetricCollector, you can just put a null here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -170,6 +174,8 @@ public void close() {
 
     LOG.info("Finish computing new ideal states for resources: {}",
         resourceMap.keySet().toString());
+    // Register all metrics
+    _metricCollector.register();
```
jiajunwang: Considering future improvement, I mean constructing the WAGED rebalance only once, please register in the constructor. Or even earlier. You can pass the registered _metricCollector to the WAGED rebalancer so it does not need to register again and again.

Another question is that I assume metric data has been updated here, why we register after the data has been collected?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/CountMetric.java
```diff
@@ -0,0 +1,41 @@
...
+ * under the License.
+ */
+
+import org.apache.helix.monitoring.mbeans.dynamicMBeans.SimpleDynamicMetric;
+
+/**
+ * Represents a count metric and defines methods to help with calculation. A count metric gives a
+ * gauge value of a certain property.
+ */
+public abstract class CountMetric<V> extends SimpleDynamicMetric<V> implements Metric {
```
jiajunwang: This looks more like a gauge instead of the counter.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -255,11 +261,14 @@ private void refreshBaseline(ResourceControllerDataProvider clusterData,
       Map<HelixConstants.ChangeType, Set<String>> clusterChanges, Map<String, Resource> resourceMap,
       final CurrentStateOutput currentStateOutput) throws HelixRebalanceException {
     LOG.info("Start calculating the new best possible assignment.");
+    LatencyMetric partialRebalanceMetric = new PartialRebalanceMetric("PartialRebalanceLatency");
+    _metricCollector.addMetric(partialRebalanceMetric);
```
jiajunwang: I would prefer adding metrics in the constructor and use getMetric() here. The code would be easier to understand.
Or if you prefer, we can just merge get and add methods.

The problem of the implementation is that, if the rebalancer calls partialRebalance more than once, the update metric object will not be recorded in the Metric Collector.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -255,11 +261,14 @@ private void refreshBaseline(ResourceControllerDataProvider clusterData,
...
+    partialRebalanceMetric.start();
```
jiajunwang: I like this design. Shall we call it startTiming or another more specific method name?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/LatencyMetric.java
```diff
@@ -0,0 +1,52 @@
...
+   * @param metricObject the metric object
+   */
+  public LatencyMetric(String metricName, Histogram metricObject) {
+    super(metricName, metricObject);
+  }
+
+  /**
+   * Starts measuring the latency.
+   */
+  public abstract void start();
```
jiajunwang: What will happen if a user calls start() twice consecutively?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/Metric.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Prints the metric along with its name.
+   */
+  String print();
+
+  /**
+   * Returns the underlying DynamicMetric.
+   */
+  DynamicMetric getDynamicMetric();
```
jiajunwang: Why we need this in the interface?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/PartialRebalanceMetric.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.TimeUnit;
+import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMetric;
+import org.apache.helix.monitoring.mbeans.dynamicMBeans.HistogramDynamicMetric;
+
+public class PartialRebalanceMetric extends LatencyMetric {
+  private static final long DEFAULT_RESET_INTERVAL_MS = 60 * 60 * 1000; // 1 hour
+
+  public PartialRebalanceMetric(String metricName) {
+    super(metricName, new Histogram(
+        new SlidingTimeWindowArrayReservoir(DEFAULT_RESET_INTERVAL_MS, TimeUnit.MILLISECONDS)));
```
jiajunwang: Could you please read the newly added interval configuration item here as well?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/MonitorDomainNames.java
```diff
@@ -28,5 +28,6 @@
   HelixThreadPoolExecutor,
   HelixCallback,
   RoutingTableProvider,
-  CLMParticipantReport
+  CLMParticipantReport,
+  WagedRebalancer
```
jiajunwang: Let's just call it Rebalancer? Maybe we want to fill other metrics regarding the legacy rebalancers as well?
I'm not sure about this, more like a question.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/PartialRebalanceMetric.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @Override
+  public void start() {
+    super._startTime = System.currentTimeMillis();
+  }
+
+  @Override
+  public void end() {
+    super._endTime = System.currentTimeMillis();
```
jiajunwang: Shouldn't you update the Histogram object here? Let me know if I missed something.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -255,11 +261,14 @@ private void refreshBaseline(ResourceControllerDataProvider clusterData,
       Map<HelixConstants.ChangeType, Set<String>> clusterChanges, Map<String, Resource> resourceMap,
       final CurrentStateOutput currentStateOutput) throws HelixRebalanceException {
     LOG.info("Start calculating the new best possible assignment.");
+    LatencyMetric partialRebalanceMetric = new PartialRebalanceMetric("PartialRebalanceLatency");
```
i3wangyi: Regardless of the code style and details, my concerns are:
Say you have implemented every single metric very well and they all can be visualized and analyzed, but there's no way to link separate metrics together from this design. In more detail, suppose our goal is to understand out of the entire pipeline, what's the latency breakdown for each major process?
Without the linkability, it's incomplete and biased since you cannot just add average latency metric up and get the total latency. 
If your goal is to understand the average latency of each step, regardless of how the total latency is, this design is fine. My concern is we cannot get much out of it,

For example, the waged rebalance pipeline consists of 3 steps, (step0, step1, step2). In one of the running process, the total latency is 100s (it breaks down into 5s, 80s, and 15s each); In the other running process, the total latency is 1000s (breaks down into 900s, 50s, 50s). In such an extreme case, how could we use the step0's standlone latency for the explanations?

narendly: @i3wangyi 
I'm not sure if I understand your concern entirely.

It actually is meant to be like that - this is just one metric we are adding. Here I've only added one metric just as an example, but every "step" of the pipeline like the way you referred to, will have an implementation. 

So you could say that we will at least have a Metric object for each step. We will also have a metric for the entire step. We'll emit appropriate gauges/counters/Histogram for more detailed analysis.

In short, we want to do exactly what you're describing. But at the code level, we won't have any dependency between metrics or hierarchy since JMX will treat every metric equally anyways (whether it represents step 0 or the entire duration of the pipeline, they are just another metric to JMX). Hope that clears things up.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/MonitorDomainNames.java
narendly: let's start with this and change it later on :)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/PartialRebalanceMetric.java
narendly: that's correct. i haven't implemented that logic yet since this was for just for demonstration.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/Metric.java
narendly: For helper methods for logging.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/PartialRebalanceMetric.java
narendly: Could you clarify? Not sure which one you're referring to. @jiajunwang

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/CountMetric.java
narendly: This will change in the later iteration. I will be changing the name for this.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/model/Metric.java
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Prints the metric along with its name.
+   */
+  String print();
+
+  /**
+   * Returns an appropriate value for the metric at the time of the call.
+   * @return metric value
+   */
+  long getMetricValue();
```
jiajunwang: It seems that this method is only for testing, right? I have some concern, the definition is not crystal clear. What is "appropriate"? What's the expected return for the histogram, or a counter? They are all different. For the user of the interface, it would hard to determine how to use it.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/model/LatencyMetric.java
```diff
@@ -0,0 +1,52 @@
...
+  protected long _startTime;
+  protected long _endTime;
+  protected String _metricName;
+
+  /**
+   * Instantiates a new Histogram dynamic metric.
+   * @param metricName the metric name
+   * @param metricObject the metric object
+   */
+  public LatencyMetric(String metricName, Histogram metricObject) {
```
jiajunwang: Can we just construct a new Histogram inside the constructor? I think there was some restriction preventing us from doing so. But I don't see any problem if we do it here.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -249,8 +253,15 @@ private boolean validateOfflineInstancesLimit(final ResourceControllerDataProvid
     // Init rebalancer with the rebalance preferences.
     Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preferences = cache.getClusterConfig()
         .getGlobalRebalancePreference();
+
+    // TODO: Add an appropriate entityName
+    if (METRIC_COLLECTOR_THREAD_LOCAL.get() == null) {
+      METRIC_COLLECTOR_THREAD_LOCAL.set(new MetricCollector(helixManager.getClusterName(), ""));
```
jiajunwang: I guess the lazy init is not required here.
Let's just put it to the constructor of the stage object.

It's always better to have the metrics code "disappear", you know.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/MonitorDomainNames.java
jiajunwang: No, the problem is that, once you have the name, changing it requires us to consider backward compatibility. Better to do it right on the first shot.

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/TestWagedRebalancer.java
```diff
@@ -110,9 +111,9 @@ protected ResourceControllerDataProvider setupClusterDataCache() throws IOExcept
   }
 
   @Test
-  public void testRebalance() throws IOException, HelixRebalanceException {
+  public void testRebalance() throws IOException, HelixRebalanceException, JMException {
     _metadataStore.clearMetadataStore();
-    WagedRebalancer rebalancer = new WagedRebalancer(_metadataStore, _algorithm);
+    WagedRebalancer rebalancer = new WagedRebalancer(_metadataStore, _algorithm, null);
```
jiajunwang: Not sure about all the usage, but if all test code is passing null, we can just remove this parameter from this constructor.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/model/Metric.java
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Resets the internal state of this metric.
+   */
+  void reset();
+
+  /**
+   * Prints the metric along with its name.
+   */
+  String print();
```
jiajunwang: toString() ?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/MetricCollector.java
```diff
@@ -0,0 +1,97 @@
...
+        "MetricCollector", _entityName);
+    doRegister(dynamicMetrics, new ObjectName(
+        String.format("%s:%s", MonitorDomainNames.WagedRebalancer.name(), mbeanName)));
+    _registered = true;
+    return this;
+  }
+
+  @Override
+  public String getSensorName() {
+    return String.format("%s.%s.%s", "MetricCollector", _clusterName, _entityName);
```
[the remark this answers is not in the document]
narendly: Added monitor domain name, cluster name, and entity name.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -90,31 +91,41 @@ private static AssignmentMetadataStore constructAssignmentStore(HelixManager hel
...
+      RebalanceAlgorithm algorithm) throws JMException {
+    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer(),
+        new WagedRebalancerMetricCollector());
   }
 
   @VisibleForTesting
-  protected WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
-      RebalanceAlgorithm algorithm) {
-    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer());
+  public WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
```
jiajunwang: Can we move the tests to the same package so this method can be protected?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -134,10 +145,9 @@ public void close() {
    */
   public Map<String, IdealState> computeNewIdealStates(ResourceControllerDataProvider clusterData,
       Map<String, Resource> resourceMap, final CurrentStateOutput currentStateOutput)
-      throws HelixRebalanceException {
+      throws HelixRebalanceException, JMException {
```
jiajunwang: Don't throw the monitor exception. Just log the error and ignore it. We don't want the monitor logic or bug to block business logic.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -90,31 +92,39 @@ private static AssignmentMetadataStore constructAssignmentStore(HelixManager hel
...
+      RebalanceAlgorithm algorithm, MappingCalculator mappingCalculator,
+      MetricCollector metricCollector) throws JMException {
     if (assignmentMetadataStore == null) {
       LOG.warn("Assignment Metadata Store is not configured properly."
           + " The rebalancer will not access the assignment store during the rebalance.");
     }
     _assignmentMetadataStore = assignmentMetadataStore;
     _rebalanceAlgorithm = algorithm;
     _mappingCalculator = mappingCalculator;
+    _metricCollector = metricCollector;
```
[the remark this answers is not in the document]
jiajunwang: As I checked the code about how _metricCollector is leveraged, there is no null check after this point.
Since we still have a public constructor that accepts the instance, we need to check for null to prevent NPE here. Or check for null everywhere it is used. Either way is fine to me.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -391,7 +429,13 @@ private void validateInput(ResourceControllerDataProvider clusterData,
     Map<String, ResourceAssignment> currentBaseline = Collections.emptyMap();
     if (assignmentMetadataStore != null) {
       try {
+        _metricCollector.getMetric(
+            WagedRebalancerMetricCollector.WagedRebalancerMetricNames.StateReadLatencyGauge.name(),
+            LatencyMetric.class).startMeasuringLatency();
         currentBaseline = assignmentMetadataStore.getBaseline();
+        _metricCollector.getMetric(
+            WagedRebalancerMetricCollector.WagedRebalancerMetricNames.StateReadLatencyGauge.name(),
+            LatencyMetric.class).endMeasuringLatency();
```
jiajunwang: Can we have temp var for the metric object?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/MetricCollector.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.helix.HelixException;
+import org.apache.helix.controller.rebalancer.waged.metrics.model.Metric;
+import org.apache.helix.monitoring.mbeans.MonitorDomainNames;
+import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMBeanProvider;
+import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMetric;
+
+/**
+ * Collects and manages all metrics that implement the {@link Metric} interface.
+ */
+public class MetricCollector extends DynamicMBeanProvider {
```
jiajunwang: It might be safer to mark this class as abstract.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/WagedRebalancerMetricCollector.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  public WagedRebalancerMetricCollector(String clusterName) throws JMException {
+    super(clusterName, WAGED_REBALANCER_ENTITY_NAME);
+    createMetrics();
+    register();
+  }
+
+  @VisibleForTesting
+  public WagedRebalancerMetricCollector() throws JMException {
```
jiajunwang: Comment the metrics won't be registered while using this constructor.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/WagedRebalancerMetricCollector.java
```diff
@@ -0,0 +1,76 @@
...
+    LatencyMetric globalBaselineCalcLatencyGauge = new RebalanceLatencyGauge(
+        WagedRebalancerMetricNames.GlobalBaselineCalcLatencyGauge.name(), getResetIntervalInMs());
+    LatencyMetric partialRebalanceLatencyGauge = new RebalanceLatencyGauge(
+        WagedRebalancerMetricNames.PartialRebalanceLatencyGauge.name(), getResetIntervalInMs());
+    LatencyMetric stateReadLatencyGauge = new RebalanceLatencyGauge(
+        WagedRebalancerMetricNames.StateReadLatencyGauge.name(), getResetIntervalInMs());
+    LatencyMetric stateWriteLatencyGauge = new RebalanceLatencyGauge(
+        WagedRebalancerMetricNames.StateWriteLatencyGauge.name(), getResetIntervalInMs());
+
+    // Add to MetricCollector
```
jiajunwang: nit: Add metrics to the WagedRebalancerMetricCollector.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/metrics/implementation/RebalanceLatencyGauge.java
```diff
@@ -0,0 +1,94 @@
...
+    _endTime = 0L;
+  }
+
+  @Override
+  public String toString() {
+    return String.format("Metric %s's latency is %d", _metricName, getLastEmittedMetricValue());
+  }
+
+  /**
+   * Returns an appropriate metric value at the time of the call.
```
jiajunwang: nit: update the comments, please.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -104,7 +109,7 @@ public Object call() {
   }
 
   private BestPossibleStateOutput compute(ClusterEvent event, Map<String, Resource> resourceMap,
-      CurrentStateOutput currentStateOutput) {
+      CurrentStateOutput currentStateOutput) throws JMException {
```
jiajunwang: Same here, we don't want any monitor logic to block business. No matter how broken it could be.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -249,8 +254,16 @@ private boolean validateOfflineInstancesLimit(final ResourceControllerDataProvid
     // Init rebalancer with the rebalance preferences.
     Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preferences = cache.getClusterConfig()
         .getGlobalRebalancePreference();
+
+    // TODO: Add an appropriate entityName
```
jiajunwang: I think this has been done.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -117,6 +122,16 @@ private WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
     _rebalanceAlgorithm = algorithm;
     _mappingCalculator = mappingCalculator;
     _manager = manager;
+    // If metricCollector is null, instantiate a version that does not register metrics in order to
+    // allow rebalancer to proceed
+    _metricCollector =
+        metricCollector == null ? new WagedRebalancerMetricCollector() : metricCollector;
+  }
+
+  @VisibleForTesting
```
jiajunwang: As you suggested in the other PR, it is not preferred to use "@VisibleForTesting" method, right? Shall we remove this annotation and move it closer to the other protected method?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -226,16 +247,25 @@ private void refreshBaseline(ResourceControllerDataProvider clusterData,
...
-            Collections.emptyMap(), currentBaseline);
+    Map<String, ResourceAssignment> newBaseline = calculateAssignment(clusterData, clusterChanges,
+        resourceMap, clusterData.getAllInstances(), Collections.emptyMap(), currentBaseline);
 
+    // Write the new baseline to metadata store
+    _metricCollector.getMetric(WagedRebalancerMetricNames.GlobalStateWriteLatencyGauge.name(),
+        LatencyMetric.class).startMeasuringLatency();
     if (_assignmentMetadataStore != null) {
       try {
         _assignmentMetadataStore.persistBaseline(newBaseline);
```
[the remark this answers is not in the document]
jiajunwang: The BucketAccessor will be a generic component, right? What's the plan to track it's latency if it is used somewhere else outside the rebalancer?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -447,7 +490,7 @@ private void validateInput(ResourceControllerDataProvider clusterData,
    * @param currentStateOutput
    * @param resources
    * @return The current best possible assignment. If record does not exist in the
-   * assignmentMetadataStore, return the current state assignment.
+   *         assignmentMetadataStore, return the current state assignment.
    * @throws HelixRebalanceException
    */
   private Map<String, ResourceAssignment> getBestPossibleAssignment(
```
jiajunwang: No metric for the this read latency?

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/MockAssignmentMetadataStore.java
```diff
@@ -33,7 +33,7 @@
   private Map<String, ResourceAssignment> _persistGlobalBaseline = new HashMap<>();
   private Map<String, ResourceAssignment> _persistBestPossibleAssignment = new HashMap<>();
 
-  MockAssignmentMetadataStore() {
+  public MockAssignmentMetadataStore() {
```
jiajunwang: Isn't the test class that uses it in the same package?

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/TestWagedRebalancer.java
```diff
@@ -47,9 +47,8 @@
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.anyString;
-import static org.mockito.Mockito.when;
+import static org.mockito.Matchers.*;
```
jiajunwang: I think wildcard is not preferred.

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/model/AbstractTestClusterModel.java
```diff
@@ -116,11 +115,11 @@ protected ResourceControllerDataProvider setupClusterDataCache() throws IOExcept
     // 4. Mock two resources, each with 2 partitions on the default instance.
     // The instance will have the following partitions assigned
     // Resource 1:
-    //          partition 1 - MASTER
-    //          partition 2 - SLAVE
+    // partition 1 - MASTER
```
jiajunwang: Please try not to change it.
Or fill the line with "--------------"
Otherwise it becomes less readable.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/MetricCollector.java
```diff
@@ -0,0 +1,92 @@
...
+  public DynamicMBeanProvider register() throws JMException {
+    // First cast all Metric objects to DynamicMetrics
+    Collection<DynamicMetric<?, ?>> dynamicMetrics = new HashSet<>();
+    _metricMap.values().forEach(metric -> dynamicMetrics.add(metric.getDynamicMetric()));
+
+    // TODO: Create helper methods for mbeanName and ObjectName
+    String mbeanName =
+        String.format("%s=%s, %s=%s", CLUSTER_NAME_KEY, _clusterName, ENTITY_NAME_KEY, _entityName);
+    doRegister(dynamicMetrics,
+        new ObjectName(String.format("%s:%s", MonitorDomainNames.Rebalancer.name(), mbeanName)));
```
jiajunwang: Question, is MetricCollector designed for Rebalancer metrics only?
If not, here it is hardcoded.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
```diff
@@ -0,0 +1,94 @@
...
+  @Override
+  public void startMeasuringLatency() {
+    reset();
+    _startTime = System.currentTimeMillis();
+  }
+
+  @Override
+  public void endMeasuringLatency() {
+    if (_endTime != 0L) {
+      throw new HelixException(
```
jiajunwang: This may fail the business logic. Please don't throw Exceptions.
If _endTime != 0, we can log the error and ignore this recording.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
```diff
@@ -0,0 +1,94 @@
...
+    _metricName = metricName;
+  }
+
+  /**
+   * Calling this method multiple times would simply overwrite the previous state. This is because
+   * the rebalancer could fail at any point, and we want it to recover gracefully by resetting the
+   * internal state of this metric.
+   */
+  @Override
+  public void startMeasuringLatency() {
```
jiajunwang: I think you will need to add sync keyword for startMeasuringLatency and endMeasuringLatency.
There is no guarantee this metric object is used in only one thread.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
```diff
@@ -0,0 +1,94 @@
...
+    updateValue(_lastEmittedMetricValue);
+  }
+
+  @Override
+  public String getMetricName() {
+    return _metricName;
+  }
+
+  @Override
+  public void reset() {
```
jiajunwang: Why this is public?
Consider the following call sequence:
1. startMeasuringLatency()
2. reset()
3. endMeasuringLatency(), then you record a huge number as the latency.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -818,6 +818,7 @@ private void dropRebalancedRunningTasks(Map<String, SortedSet<Integer>> newAssig
             paMap.put(pId, new PartitionAssignment(instance, TaskPartitionState.DROPPED.name()));
             jobContext.setPartitionState(pId, TaskPartitionState.DROPPED);
             // Now it will be dropped and be rescheduled
+            // drop it from new assignemnt
```
jiajunwang: Unrelated change?

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/model/AbstractTestClusterModel.java
```diff
@@ -38,7 +37,7 @@
 import org.mockito.Mockito;
 import org.testng.annotations.BeforeClass;
 
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
jiajunwang: Still wildcard. I think @pkuwm has already added it to style file. Please try importing the latest file.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
```diff
@@ -41,7 +41,7 @@
   // enlarge the overall weight of the evenness constraints compared with the movement constraint.
   // TODO: Tune or make the following factor configurable.
   private static final int EVENNESS_PREFERENCE_NORMALIZE_FACTOR = 50;
-  private static final Map<String, Float> MODEL = new HashMap<>() {
+  private static final Map<String, Float> MODEL = new HashMap<String, Float>() {
```
i3wangyi: Thanks for the fix

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
```diff
@@ -223,4 +225,13 @@ protected void incrementSimpleDynamicMetric(SimpleDynamicMetric<Long> metric) {
   protected void incrementSimpleDynamicMetric(SimpleDynamicMetric<Long> metric, long value) {
     metric.updateValue(metric.getValue() + value);
   }
+
+  /**
+   * Return the interval length for the underlying reservoir used by the MBean metric configured
+   * in the system env variables. If not found, use default value.
+   */
+  protected Long getResetIntervalInMs() {
```
i3wangyi: use primitive type long

helix-core/src/main/java/org/apache/helix/monitoring/metrics/MetricCollector.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMBeanProvider;
+import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMetric;
+
+/**
+ * Collects and manages all metrics that implement the {@link Metric} interface.
+ */
+public abstract class MetricCollector extends DynamicMBeanProvider {
+  private static final String CLUSTER_NAME_KEY = "ClusterName";
+  private static final String ENTITY_NAME_KEY = "EntityName";
+  private String _clusterName;
```
i3wangyi: private final of the fields?

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
narendly: Would this work?

```
  @Override
  public void endMeasuringLatency() {
    if (_startTime == 0L || _endTime != 0L) {
      LOG.error(
          "Needs to call startMeasuringLatency first! Ignoring and resetting the metric. Metric name: {}",
          _metricName);
      reset();
    }
    _endTime = System.currentTimeMillis();
    _lastEmittedMetricValue = _endTime - _startTime;
    updateValue(_lastEmittedMetricValue);
  }
```

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
narendly: Good catch.

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/model/AbstractTestClusterModel.java
narendly: Fixed: `import static org.mockito.Mockito.when;`

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
jiajunwang: Agreed.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
jiajunwang: No, in this case, instead of resetting and continue the logic, I would prefer reset and skip the updateValue().

How about:
1. reset set both value to be -1.
2. if _startTime is -1, ignore the endMeasuringLatency calls.
3. After update value, force reset.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -253,20 +258,41 @@ private boolean validateOfflineInstancesLimit(final ResourceControllerDataProvid
...
+        LogUtil.logWarn(logger, _eventId, String.format(
+            "MetricCollector instantiation failed! WagedRebalancer will not emit metrics due to JMException %s",
+            e));
+      }
+    }
+
     // TODO avoid creating the rebalancer on every rebalance call for performance enhancement
-    WagedRebalancer wagedRebalancer = new WagedRebalancer(helixManager, preferences);
+    WagedRebalancer wagedRebalancer =
+        new WagedRebalancer(helixManager, preferences, METRIC_COLLECTOR_THREAD_LOCAL.get());
```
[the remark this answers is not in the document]
narendly: null.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
narendly: This is from another PR. We should fix it after this gets merged with master. Otherwise, there will be a merge conflict.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/MetricCollector.java
narendly: Fixed.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
narendly: Also need to check _endTime != VALUE_NOT_SET

narendly:

    Add WAGED rebalancer metric framework and latency metric implementation
    
    Changelist:
    1. Add WAGED rebalancer metric interface
    2. Implement latency-related metrics
    3. Integrate latency metrics into WAGED rebalancer
    4. Add tests

PR #506, opened 2019-10-10
Increase parallelism for ZkBucketDataAccessor

### LINKED ISSUE #509: Allow multiple readers for BucketDataAccessor
Low Parallelism
Currently, ZkBasedBucketDataAccessor uses a locking mechanism implemented using ZooKeeper's ephemeral nodes. This is similar to a mutex, and allows for one entity to either read/write. Since there could only be one ongoing operation, the users of this data accessor such as WAGED rebalancer suffers from low parallelism. More specifically, it is not possible to have verifiers read from the metadata written to ZK while some RW is happening for testing/verification purposes.

Latency
Overall latency of the rebalance pipeline (BestPossibleCalcStage) could be reduced if we could parallelize read and write operations. The current implementation, due to locking, has to linearize all read and write operations.

Added Complexity from Locking
Inevitably, there will be unexpected failures. When we use locks in a distributed manner, we need to account for all failure scenarios related to connection issues. This makes coding a little more difficult because of many try-catch statements.

narendly:
### Issues

Resolves #509 

### Description

This diff improves parallelism and throughput for ZkBucketDataAccessor. It implements the following ideas:
1. Optimistic Concurrency Control
2. Monotonically Increasing Version Number
3. Garbage Collection of Stale Metadata
4. TTL for stale metadata

### Tests

**TestZkBucketDataAccessor**

[ERROR] Tests run: 1036, Failures: 2, Errors: 0, Skipped: 3, Time elapsed: 4,911.756 s <<< FAILURE! - in TestSuite
[ERROR] testEnableCompressionResource(org.apache.helix.integration.TestEnableCompression)  Time elapsed: 300.37 s  <<< FAILURE!
org.testng.internal.thread.ThreadTimeoutException: Method org.testng.internal.TestNGMethod.testEnableCompressionResource() didn't finish within the time-out 300000
	at org.apache.helix.integration.TestEnableCompression.testEnableCompressionResource(TestEnableCompression.java:107)

[ERROR] testTaskPerformanceMetrics(org.apache.helix.monitoring.mbeans.TestTaskPerformanceMetrics)  Time elapsed: 5.274 s  <<< FAILURE!
	at org.apache.helix.monitoring.mbeans.TestTaskPerformanceMetrics.testTaskPerformanceMetrics(TestTaskPerformanceMetrics.java:119)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestEnableCompression.testEnableCompressionResource:107 » ThreadTimeout Method...
[ERROR]   TestTaskPerformanceMetrics.testTaskPerformanceMetrics:119 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1036, Failures: 2, Errors: 0, Skipped: 3
[INFO] 
[INFO] BUILD FAILURE

------------------------

START testEnableCompressionResource at Thu Oct 10 16:36:01 PDT 2019
START TestEnableCompression_testEnableCompressionResource at Thu Oct 10 16:36:01 PDT 2019
true: ClusterStateVerifier$BestPossAndExtViewZkVerifier(TestEnableCompression_testEnableCompressionResource@localhost:2183): wait 16988ms to verify
compressed paths:[/TestEnableCompression_testEnableCompressionResource/EXTERNALVIEW/TestResource, /TestEnableCompression_testEnableCompressionResource/IDEALSTATES/TestResource]
END TestEnableCompression_testEnableCompressionResource at Thu Oct 10 16:36:20 PDT 2019
END testEnableCompressionResource at Thu Oct 10 16:36:20 PDT 2019, took: 18859ms.
START testTaskPerformanceMetrics at Thu Oct 10 16:36:25 PDT 2019
END testTaskPerformanceMetrics at Thu Oct 10 16:36:35 PDT 2019, took: 10288ms.
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 38.943 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS

### Commits

### Documentation

https://github.com/apache/helix/wiki/Concurrency-and-Parallelism-for-BucketDataAccessor

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
@@ -26,8 +26,12 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ExecutorService;
+import java.util.TimerTask;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import org.I0Itec.zkclient.DataUpdater;
 import org.I0Itec.zkclient.exception.ZkMarshallingError;
 import org.I0Itec.zkclient.exception.ZkNoNodeException;
@@ -41,13 +45,9 @@ import org.apache.helix.manager.zk.client.DedicatedZkClientFactory;
 import org.apache.helix.manager.zk.client.HelixZkClient;
 import org.apache.helix.util.GZipCompressionUtil;
 import org.codehaus.jackson.map.ObjectMapper;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
-  private static final Logger LOG = LoggerFactory.getLogger(ZkBucketDataAccessor.class);
-
-  private static final long DEFAULT_VERSION_TTL = 60000L; // 1 min
+  private static final long DEFAULT_VERSION_TTL = TimeUnit.MINUTES.toMillis(1L); // 1 min
   private static final String BUCKET_SIZE_KEY = "BUCKET_SIZE";
   private static final String DATA_SIZE_KEY = "DATA_SIZE";
   private static final String METADATA_KEY = "METADATA";
@@ -55,10 +55,11 @@ public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
   private static final String LAST_WRITE_KEY = "LAST_WRITE";
   private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
   // Thread pool for deleting stale versions
-  private static final ExecutorService GC_THREAD_POOL = Executors.newFixedThreadPool(1);
+  private static final ScheduledExecutorService GC_THREAD = Executors.newScheduledThreadPool(1);
+  // Concurrent map to keep track of <Root path, ScheduledFuture (gc task)> pairs
+  private static final Map<String, ScheduledFuture> GC_TASK_MAP = new ConcurrentHashMap<>();
 
-  // 100 KB for default bucket size
-  private static final int DEFAULT_BUCKET_SIZE = 50 * 1024;
+  private static final int DEFAULT_BUCKET_SIZE = 50 * 1024; // 50KB
   private final int _bucketSize;
   private final long _versionTTL;
   private ZkSerializer _zkSerializer;
@@ -69,6 +70,7 @@ public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
    * Constructor that allows a custom bucket size.
    * @param zkAddr
    * @param bucketSize
+   * @param versionTTL in ms
    */
   public ZkBucketDataAccessor(String zkAddr, int bucketSize, long versionTTL) {
     _zkClient = DedicatedZkClientFactory.getInstance()
@@ -102,37 +104,36 @@ public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
   }
 
   @Override
-  public <T extends HelixProperty> boolean compressedBucketWrite(String path, T value)
+  public <T extends HelixProperty> boolean compressedBucketWrite(String rootPath, T value)
       throws IOException {
-    final long timestamp = System.currentTimeMillis();
-
     DataUpdater<byte[]> lastWriteVersionUpdater = dataInZk -> {
       if (dataInZk == null || dataInZk.length == 0) {
         // No last write version exists, so start with 0
-        return ("0_" + timestamp).getBytes();
+        return "0".getBytes();
       }
-      // Last write exists, so increment and write it back with a timestamp
+      // Last write exists, so increment and write it back
+      // **String conversion is necessary to make it display in ZK (zooinspector)**
       String lastWriteVersionStr = new String(dataInZk);
-      long lastWriteVersion = extractVersion(lastWriteVersionStr);
+      long lastWriteVersion = Long.parseLong(lastWriteVersionStr);
       lastWriteVersion++;
-      return (lastWriteVersion + "_" + timestamp).getBytes();
+      return String.valueOf(lastWriteVersion).getBytes();
     };
 
     // 1. Increment lastWriteVersion using DataUpdater
-    ZkBaseDataAccessor.AccessResult result = _zkBaseDataAccessor
-        .doUpdate(path + "/" + LAST_WRITE_KEY, lastWriteVersionUpdater, AccessOption.PERSISTENT);
+    ZkBaseDataAccessor.AccessResult result = _zkBaseDataAccessor.doUpdate(
+        rootPath + "/" + LAST_WRITE_KEY, lastWriteVersionUpdater, AccessOption.PERSISTENT);
     if (result._retCode != ZkBaseDataAccessor.RetCode.OK) {
       throw new HelixException(
-          String.format("Failed to write the write version at path: %s!", path));
+          String.format("Failed to write the write version at path: %s!", rootPath));
     }
 
     // Successfully reserved a version number
     byte[] binaryVersion = (byte[]) result._updatedValue;
-    final String timestampedVersion = new String(binaryVersion);
-    final long version = extractVersion(timestampedVersion);
+    String versionStr = new String(binaryVersion);
+    final long version = Long.parseLong(versionStr);
 
-    // 2. Write to the incremented last write version with timestamp for TTL
-    String versionedDataPath = path + "/" + timestampedVersion;
+    // 2. Write to the incremented last write version
+    String versionedDataPath = rootPath + "/" + versionStr;
 
     // Take the ZNRecord and serialize it (get byte[])
     byte[] serializedRecord = _zkSerializer.serialize(value.getRecord());
@@ -172,7 +173,7 @@ public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
     for (boolean s : success) {
       if (!s) {
         throw new HelixException(
-            String.format("Failed to write the data buckets for path: %s", path));
+            String.format("Failed to write the data buckets for path: %s", rootPath));
       }
     }
 
@@ -180,27 +181,27 @@ public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
     DataUpdater<byte[]> lastSuccessfulWriteVersionUpdater = dataInZk -> {
       if (dataInZk == null || dataInZk.length == 0) {
         // No last write version exists, so write version from this write
-        return timestampedVersion.getBytes();
+        return versionStr.getBytes();
       }
       // Last successful write exists so check if it's smaller than my number
-      String timestampedLastWriteVersion = new String(dataInZk);
-      long lastWriteVersion = extractVersion(timestampedLastWriteVersion);
+      String lastWriteVersionStr = new String(dataInZk);
+      long lastWriteVersion = Long.parseLong(lastWriteVersionStr);
       if (lastWriteVersion < version) {
         // Smaller, so I can overwrite
-        return timestampedVersion.getBytes();
+        return versionStr.getBytes();
       } else {
         // Greater, I have lagged behind. Return null and do not write
         return null;
       }
     };
-    if (!_zkBaseDataAccessor.update(path + "/" + LAST_SUCCESSFUL_WRITE_KEY,
+    if (!_zkBaseDataAccessor.update(rootPath + "/" + LAST_SUCCESSFUL_WRITE_KEY,
         lastSuccessfulWriteVersionUpdater, AccessOption.PERSISTENT)) {
-      throw new HelixException(
-          String.format("Failed to write the last successful write metadata at path: %s!", path));
+      throw new HelixException(String
+          .format("Failed to write the last successful write metadata at path: %s!", rootPath));
     }
 
-    // 5. Garbage collect stale versions
-    GC_THREAD_POOL.submit(() -> deleteStaleVersions(path, version));
+    // 5. Update the timer for GC
+    updateGCTimer(rootPath);
     return true;
   }
 
@@ -306,20 +307,40 @@ public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
     disconnect();
   }
 
+  private void updateGCTimer(String rootPath) {
+    // Reset the timer
+    ScheduledFuture future = GC_TASK_MAP.get(rootPath);
+    if (future != null) {
+      future.cancel(false);
+    }
+
+    TimerTask gcTask = new TimerTask() {
+      @Override
+      public void run() {
+        deleteStaleVersions(rootPath);
+        // GC complete so remove myself from task map
+        GC_TASK_MAP.remove(rootPath);
+      }
+    };
+
+    // Schedule the gc task with TTL
+    ScheduledFuture gcFuture = GC_THREAD.schedule(gcTask, _versionTTL, TimeUnit.MILLISECONDS);
+    GC_TASK_MAP.put(rootPath, gcFuture);
+  }
+
   /**
-   * Deletes all versions that are older than the currentVersion given.
-   * @param path
-   * @param currentVersion
+   * Deletes all stale versions.
+   * @param rootPath
    */
-  private void deleteStaleVersions(String path, long currentVersion) {
+  private void deleteStaleVersions(String rootPath) {
     // Get all children names under path
-    List<String> children = _zkBaseDataAccessor.getChildNames(path, AccessOption.PERSISTENT);
+    List<String> children = _zkBaseDataAccessor.getChildNames(rootPath, AccessOption.PERSISTENT);
     if (children == null || children.isEmpty()) {
       // The whole path has been deleted so return immediately
       return;
     }
-    filterChildrenNames(children, currentVersion);
-    List<String> pathsToDelete = getPathsToDelete(path, children);
+    filterChildrenNames(rootPath, children);
+    List<String> pathsToDelete = getPathsToDelete(rootPath, children);
     for (String pathToDelete : pathsToDelete) {
       // TODO: Should be batch delete but it doesn't work. It's okay since this runs async
       _zkBaseDataAccessor.remove(pathToDelete, AccessOption.PERSISTENT);
@@ -327,72 +348,23 @@ public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
   }
 
   /**
-   * Filter out currentVersion and non-version children names from the children list given.
+   * Filter out non-version children names and last successful write version from the children list
+   * given.
+   * @param rootPath
    * @param children
-   * @param currentVersion
    */
-  private void filterChildrenNames(List<String> children, long currentVersion) {
+  private void filterChildrenNames(String rootPath, List<String> children) {
     // Leave out metadata
     children.remove(LAST_SUCCESSFUL_WRITE_KEY);
     children.remove(LAST_WRITE_KEY);
 
-    // Keep the children that are not expired and exclude currentVersion from deletion
-    long currTime = System.currentTimeMillis();
-    children.removeIf(childName -> currTime < extractTimestamp(childName) + _versionTTL
-        || extractVersion(childName) >= currentVersion);
-  }
-
-  /**
-   * Extracts timestamp from the znode name.
-   * @param name
-   * @return
-   */
-  private long extractTimestamp(String name) {
-    // Path structure: ~~/ver_timestamp
-    String[] parts = name.split("_");
-    if (parts.length == 2) {
-      String timestampStr = parts[1];
-      long timestamp;
-      try {
-        timestamp = Long.parseLong(timestampStr);
-        return timestamp;
-      } catch (Exception e) {
-        LOG.warn(
-            "Failed to parse timestamp for ZNode name: {}. Returning Long.MAX_VALUE for timestamp!",
-            name);
-        return Long.MAX_VALUE;
-      }
+    // Get the last successful write version and exclude from removal
+    byte[] binaryLastSuccessfulWrite = _zkBaseDataAccessor
+        .get(rootPath + "/" + LAST_SUCCESSFUL_WRITE_KEY, null, AccessOption.PERSISTENT);
+    if (binaryLastSuccessfulWrite != null) {
+      String lastSuccessfulWriteVersion = new String(binaryLastSuccessfulWrite);
+      children.remove(lastSuccessfulWriteVersion);
     }
-    LOG.warn(
-        "ZNode with an invalid name detected while extracting timestamp. ZNode name: {}. Returning Long.MAX_VALUE for timestamp!",
-        name);
-    return Long.MAX_VALUE;
-  }
-
-  /**
-   * Extracts the version from the znode name.
-   * @param name
-   * @return
-   */
-  private long extractVersion(String name) {
-    String[] parts = name.split("_");
-    if (parts.length == 2) {
-      String versionStr = parts[0];
-      long version;
-      try {
-        version = Long.parseLong(versionStr);
-        return version;
-      } catch (Exception e) {
-        LOG.warn(
-            "Failed to parse version for ZNode name: {}. Returning Long.MAX_VALUE for version!",
-            name);
-        return Long.MAX_VALUE;
-      }
-    }
-    LOG.warn(
-        "ZNode with an invalid name detected while extracting version. ZNode name: {}. Returning Long.MAX_VALUE for version!",
-        name);
-    return Long.MAX_VALUE;
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
@@ -20,14 +20,14 @@ package org.apache.helix.manager.zk;
  */
 
 import com.google.common.collect.ImmutableMap;
-import com.google.common.primitives.Ints;
+import com.google.common.primitives.Longs;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import org.I0Itec.zkclient.DataUpdater;
 import org.I0Itec.zkclient.exception.ZkMarshallingError;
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -202,126 +214,130 @@ public void disconnect() {
...
-   * Returns the last string element in a split String array by /.
+   * Deletes all versions that are older than the currentVersion given.
    * @param path
-   * @return
+   * @param currentVersion
    */
-  private String extractIdFromPath(String path) {
-    String[] splitPath = path.split("/");
-    return splitPath[splitPath.length - 1];
+  private void deleteStaleVersions(String path, String currentVersion) {
```
jiajunwang: I think we are not going to apply the read retry if the in-progress read fails because of deleting, right?
In this case, we should still have some protection mechanism to prevent read failure. One way is to delay the delete for 30s, for example. Meaning we always delete the older version if it has been rolled out for more than 30s. In this case, 99.99% readers that reads the old version would be finished for sure.

This could be tricky. We can have a meeting with the team to brainstorm.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +83,117 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
+    // 4. Update lastSuccessfulWriteVersion using Updater
+    DataUpdater<byte[]> lastSuccessfulWriteVersionUpdater = dataInZk -> {
+      if (dataInZk == null || dataInZk.length == 0) {
+        // No last write version exists, so write version from this write
+        return Longs.toByteArray(version);
       }
-      return true;
+      // Last successful write exists so check if it's smaller than my number
+      long lastWriteVersion = Longs.fromByteArray(dataInZk);
+      if (lastWriteVersion < version) {
```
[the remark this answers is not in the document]
narendly: It's a long. It will never overflow. Assuming we do a write every nanosecond, we have 292 years.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +83,117 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
+        // Smaller, so I can overwrite
+        return Longs.toByteArray(version);
+      } else {
+        // Greater, I have lagged behind. Return the existing data
+        return dataInZk;
```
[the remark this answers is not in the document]
narendly: Good suggestion.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: Retries should be taken care of by the caller. And adding the delay may not be the cleanest solution. At any rate, I feel that that is out of scope of this PR.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +83,117 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
   public ZkBucketDataAccessor(String zkAddr) {
-    this(zkAddr, DEFAULT_BUCKET_SIZE, DEFAULT_NUM_VERSIONS);
+    this(zkAddr, DEFAULT_BUCKET_SIZE);
   }
 
   @Override
   public <T extends HelixProperty> boolean compressedBucketWrite(String path, T value)
       throws IOException {
+    AtomicLong versionRef = new AtomicLong();
+    DataUpdater<byte[]> lastWriteVersionUpdater = dataInZk -> {
```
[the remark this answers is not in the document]
narendly: _baseDataAccessor only takes a byte array. I think that having 1 data accessor is more beneficial.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +83,117 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
+      // Last write exists, so increment and write it back
+      long lastWriteVersion = Longs.fromByteArray(dataInZk);
+      lastWriteVersion++;
+      // Set the AtomicReference
+      versionRef.set(lastWriteVersion);
+      return Longs.toByteArray(lastWriteVersion);
+    };
+
+    // 1. Increment lastWriteVersion using DataUpdater
+    if (!_zkBaseDataAccessor.update(path + "/" + LAST_WRITE_KEY, lastWriteVersionUpdater,
```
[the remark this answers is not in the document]
narendly: Sure. That also works.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +83,117 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
-        boolean[] success =
-            _zkBaseDataAccessor.setChildren(paths, buckets, AccessOption.PERSISTENT);
-        // Exception and fail the write if any failed
-        for (boolean s : success) {
-          if (!s) {
-            throw new HelixException(
-                String.format("Failed to write the data buckets for path: %s", path));
-          }
-        }
+    // 3. Data write succeeded, so write the metadata by first serializing to byte array
```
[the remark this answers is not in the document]
narendly: Done.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +83,117 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
+    };
+    if (!_zkBaseDataAccessor.update(path + "/" + LAST_SUCCESSFUL_WRITE_KEY,
+        lastSuccessfulWriteVersionUpdater, AccessOption.PERSISTENT)) {
+      throw new HelixException(
+          String.format("Failed to write the last successful write metadata at path: %s!", path));
     }
-    throw new HelixException(String.format("Could not acquire lock for write. Path: %s", path));
+
+    // 5. Garbage collect stale versions
+    new Thread(() -> deleteStaleVersions(path, Long.toString(version))).start();
```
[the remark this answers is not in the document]
narendly: Added a thread pool of size 1. Might have to bound the size of the queue but I'm not too worried about that for now.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -202,126 +218,156 @@ public void disconnect() {
...
+   */
+  private void filterChildrenNames(List<String> children, long currentVersion) {
+    // Leave out metadata
+    children.remove(LAST_SUCCESSFUL_WRITE_KEY);
+    children.remove(LAST_WRITE_KEY);
+
+    // Keep the children that are not expired and exclude currentVersion from deletion
+    long currTime = System.currentTimeMillis();
+    children.removeIf(childName -> currTime < extractTimestamp(childName) + _versionTTL
+        || extractVersion(childName) == currentVersion);
```
jiajunwang: extractVersion(childName) >= currentVersion

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -202,126 +218,156 @@ public void disconnect() {
...
+   * @param currentVersion
+   */
+  private void filterChildrenNames(List<String> children, long currentVersion) {
+    // Leave out metadata
+    children.remove(LAST_SUCCESSFUL_WRITE_KEY);
+    children.remove(LAST_WRITE_KEY);
+
+    // Keep the children that are not expired and exclude currentVersion from deletion
+    long currTime = System.currentTimeMillis();
+    children.removeIf(childName -> currTime < extractTimestamp(childName) + _versionTTL
```
jiajunwang: In case some other random nodes are added to the folder, will extractTimestamp(childName) and extractVersion(childName) fail with exception? I think we should remove those random nodes as well as a clean up. But that also means we need to validate if their names are valid or not before the filtering.

jiajunwang: On more comments, this usage of versionTTL may leave some garbage uncollected if no new clean up is triggered. Why not just sleep in the thread for versionTTL long and then clean up everything before the current version?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -202,126 +218,156 @@ public void disconnect() {
...
-   * Returns the last string element in a split String array by /.
+   * Deletes all versions that are older than the currentVersion given.
    * @param path
-   * @return
+   * @param currentVersion
    */
-  private String extractIdFromPath(String path) {
-    String[] splitPath = path.split("/");
-    return splitPath[splitPath.length - 1];
+  private void deleteStaleVersions(String path, long currentVersion) {
```
jiajunwang: I'm fine with the unbounded queue of the threadpool. But we don't really need this currentVersion, do we?

We can always read the most recent LAST_SUCCESSFUL_WRITE version before processing.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: Added validation. If invalid, it would just return Long.MAX_VALUE and add a warn log. Alternatively, I could have made it so that we return -1L so they will get cleaned up. Either way as long as things don't break.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBucketDataAccessor.java
```diff
@@ -69,17 +98,43 @@ public void afterClass() {
...
-    record.setListField(NAME_KEY, LIST_FIELD);
-    record.setMapField(NAME_KEY, MAP_FIELD);
     Assert.assertTrue(_bucketDataAccessor.compressedBucketWrite(PATH, new HelixProperty(record)));
   }
 
+  @Test(dependsOnMethods = "testCompressedBucketWrite")
+  public void testMultipleWrites() throws Exception {
+    int count = 50;
+
+    // Write 10 times
```
narendly: TODO: Fix comment

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
i3wangyi: I think the GC thread submitted immediately after the last successful update will likely introduce a *non-repeatable read* or *read skew* problem. 
Think about another transaction T2 firstly reads the data of last successful write version = 0 concurrently, T1 finishes the last step, try to update the last successful version = 1 and cleaned up the older versions (including 0),m committed. T2 wouldn't be able to read the version = 0 repeatedly after T1 commits.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +87,121 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
   @Override
   public <T extends HelixProperty> boolean compressedBucketWrite(String path, T value)
       throws IOException {
-    // Take the ZNrecord and serialize it (get byte[])
+    final long timestamp = System.currentTimeMillis();
+
+    DataUpdater<byte[]> lastWriteVersionUpdater = dataInZk -> {
+      if (dataInZk == null || dataInZk.length == 0) {
+        // No last write version exists, so start with 0
+        return ("0_" + timestamp).getBytes();
```
i3wangyi: What's your idea behind using an "integer + timestamp" combination pattern? Since timestamp itself is monotonically Increasing, and it's already used as Txid or version id. Can we get rid of the combination pattern, to just make things simpler?

i3wangyi: Also, I'm interesting to know why two different paths are needed: last write key + last successful write key. Can we simply just keep the "last successful key"? For example, remove the 1st step of asking for a version, directly use the timestamp as the version id. Once all steps succeed, overwrite the last successful key with the timestamp

narendly: Please see https://github.com/apache/helix/wiki/Concurrency-and-Parallelism-for-BucketDataAccessor for details.
Timestamp cannot be used for monotonically-increasing sequences because of clock drift.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +87,121 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
+      String lastWriteVersionStr = new String(dataInZk);
+      long lastWriteVersion = extractVersion(lastWriteVersionStr);
+      lastWriteVersion++;
+      return (lastWriteVersion + "_" + timestamp).getBytes();
+    };
+
+    // 1. Increment lastWriteVersion using DataUpdater
+    ZkBaseDataAccessor.AccessResult result = _zkBaseDataAccessor
+        .doUpdate(path + "/" + LAST_WRITE_KEY, lastWriteVersionUpdater, AccessOption.PERSISTENT);
+    if (result._retCode != ZkBaseDataAccessor.RetCode.OK) {
```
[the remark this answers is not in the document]
narendly: _retCode is a public field.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +87,121 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
+      throw new HelixException(
+          String.format("Failed to write the write version at path: %s!", path));
+    }
+
+    // Successfully reserved a version number
+    byte[] binaryVersion = (byte[]) result._updatedValue;
```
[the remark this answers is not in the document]
narendly: It's a public field.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: @i3wangyi That's why we have TTL for versions.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
i3wangyi: Can you explain how the clock drift could happen in the helix controller, in the same data-center? We're not writing across data-centers, aren't we? e.g UTC time.

narendly: It's possible if we consider cases where there could be multiple writers. Of course, in the case where there's one writer, using timestamp would work and would simplify things. But the purpose of this change is to increase parallelism, and that means we want to allow multiple writers (whether they are in the same data centers or not) as well.

Even if we had multiple writers in the same data center, NTP usually doesn't guarantee strict ordering of timestamps either.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
jiajunwang: @narendly For the 1) point, we can use batch read to minimize the extra latency. 2) point, I think it would preferred if the thread can delete a more recent version, as long as it is not the current one. There won't be a clean way to ensure all the old versions are cleaned up. There will be a race-condition. And the propagation latency of ZK might make it worse. So my point is to decouple the cleanup logic from the business as much as possible. So no matter when we want to trigger a cleanup, we can just call the method without worrying about side effects.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -202,126 +222,188 @@ public void disconnect() {
...
+    // Path structure: ~~/ver_timestamp
+    String[] parts = name.split("_");
+    if (parts.length == 2) {
+      String timestampStr = parts[1];
+      long timestamp;
+      try {
+        timestamp = Long.parseLong(timestampStr);
+        return timestamp;
+      } catch (Exception e) {
+        LOG.warn(
```
jiajunwang: If one random node exists, I guess this warning will flood our log file. Shall we do a debug log here?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -202,126 +222,188 @@ public void disconnect() {
...
+   * @param name
    * @return
    */
-  private boolean tryLock(String path) {
-    // Check if another write is taking place and if not, create an ephemeral node to simulate
-    // acquiring of a lock
-    return !_zkBaseDataAccessor.exists(path + "/" + WRITE_LOCK_KEY, AccessOption.EPHEMERAL)
-        && _zkBaseDataAccessor.set(path + "/" + WRITE_LOCK_KEY, new byte[0],
-            AccessOption.EPHEMERAL);
+  private long extractTimestamp(String name) {
```
jiajunwang: For performance, can we return long[2] so that the name is not split twice?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
jiajunwang: > On more comments, this usage of versionTTL may leave some garbage uncollected if no new clean up is triggered. Why not just sleep in the thread for versionTTL long and then clean up everything before the current version?

Can you reply to this one? I think scheduling the cleanup task for the future (after TTL) would be a cleaner way. It reduces the possibility of garbage leftovers. Moreover, you won't need to check (or maybe recording) the Timestamp in this way.

narendly: @jiajunwang Sure. We could have a "delete scheduler" for cleanup.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +87,121 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
-          LOG.warn("Failed to clean up previous bucketed data in data path: {}", dataPath);
-        }
+    List<String> paths = new ArrayList<>();
+    List<byte[]> buckets = new ArrayList<>();
+
+    int ptr = 0;
+    int counter = 0;
+    while (counter < numBuckets) {
+      paths.add(versionedDataPath + "/" + counter);
+      if (counter == numBuckets - 1) {
```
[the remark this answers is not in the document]
narendly: Good point. Will use persistent_sequential.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +87,121 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
+        DATA_SIZE_KEY, Integer.toString(compressedRecord.length));
+    byte[] binaryMetadata = OBJECT_MAPPER.writeValueAsBytes(metadata);
+    paths.add(versionedDataPath + "/" + METADATA_KEY);
+    buckets.add(binaryMetadata);
+
+    // Do an async set to ZK
+    boolean[] success = _zkBaseDataAccessor.setChildren(paths, buckets, AccessOption.PERSISTENT);
+    // Exception and fail the write if any failed
+    for (boolean s : success) {
+      if (!s) {
```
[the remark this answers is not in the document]
narendly: This is okay - it will be cleaned up by GC thread.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -202,126 +222,188 @@ public void disconnect() {
...
+    List<String> children = _zkBaseDataAccessor.getChildNames(path, AccessOption.PERSISTENT);
+    if (children == null || children.isEmpty()) {
+      // The whole path has been deleted so return immediately
+      return;
+    }
+    filterChildrenNames(children, currentVersion);
+    List<String> pathsToDelete = getPathsToDelete(path, children);
+    for (String pathToDelete : pathsToDelete) {
+      // TODO: Should be batch delete but it doesn't work. It's okay since this runs async
+      _zkBaseDataAccessor.remove(pathToDelete, AccessOption.PERSISTENT);
```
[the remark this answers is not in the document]
narendly: Yes

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
i3wangyi: I'm concerned about the introduced first round trip of request/response, think about the latency added with a trade-off with the minor cases of timestamp not strictly increasing (I believe it happens rarely and many docs claim the usage of a simple timestamp in MVCC). It won't reduce the ability of multi-writers but will make the design simpler and performance better, I don't think the point is discussed well given our use cases.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
i3wangyi: Talked with @narendly offline, the idea of zookeeper TTL node interests me. I found starting from 3.5.3 zookeeper does support TTL nodes https://zookeeper.apache.org/doc/r3.5.4-beta/zookeeperProgrammers.html#TTL+Nodes. 

If using this feature, we can get rid of the custom GC thread at all. Here are some naive steps:
1. When a writer creates a node, it needs to be PERSISTENT
2. However, by checking the states of the nodes older than the newly created nodes. Update these nodes with TTL.

It sounds good in theory (the controller code looks clean, no GC thread), the concern is in reality, we don't even have that api supported. The code changes are even more. Just FYI.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -19,56 +19,60 @@
...
-  private BaseDataAccessor _zkBaseDataAccessor;
-  private BaseDataAccessor<ZNRecord> _znRecordBaseDataAccessor;
+  private ZkBaseDataAccessor<byte[]> _zkBaseDataAccessor;
 
   /**
    * Constructor that allows a custom bucket size.
    * @param zkAddr
    * @param bucketSize
-   * @param numVersions number of versions to store in ZK
+   * @param versionTTL in ms
```
jiajunwang: versionTTLms? Or some name includes the "ms" part.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -85,103 +89,120 @@ public Object deserialize(byte[] data) throws ZkMarshallingError {
...
+    List<String> paths = new ArrayList<>();
+    List<byte[]> buckets = new ArrayList<>();
+
+    int ptr = 0;
+    int counter = 0;
+    while (counter < numBuckets) {
+      paths.add(versionedDataPath + "/" + counter);
+      if (counter == numBuckets - 1) {
+        // Special treatment for the last bucket
+        buckets.add(
```
jiajunwang: nit:

int offset = ptr + (counter == numBuckets - 1)? (compressedRecord.length % _bucketSize) : _bucketSize
Arrays.copyOfRange(compressedRecord, ptr, offset);

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -202,126 +223,158 @@ public void disconnect() {
...
-        }
+    // Async get
+    List<byte[]> buckets = _zkBaseDataAccessor.get(paths, null, AccessOption.PERSISTENT, true);
+
+    // Combine buckets into one byte array
+    int copyPtr = 0;
+    for (int i = 0; i < numBuckets; i++) {
+      if (i == numBuckets - 1) {
+        // Special treatment for the last bucket
+        System.arraycopy(buckets.get(i), 0, compressedRecord, copyPtr, dataSize % bucketSize);
```
jiajunwang: Same here, you can simplify this block.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -19,56 +19,60 @@
...
-  private static final int DEFAULT_NUM_VERSIONS = 2;
+  private static final int DEFAULT_BUCKET_SIZE = 50 * 1024; // 50KB
+  private static final long DEFAULT_VERSION_TTL = TimeUnit.MINUTES.toMillis(1L); // 1 min
   private static final String BUCKET_SIZE_KEY = "BUCKET_SIZE";
   private static final String DATA_SIZE_KEY = "DATA_SIZE";
-  private static final String WRITE_LOCK_KEY = "WRITE_LOCK";
-  private static final String LAST_SUCCESS_KEY = "LAST_SUCCESS";
+  private static final String METADATA_KEY = "METADATA";
+  private static final String LAST_SUCCESSFUL_WRITE_KEY = "LAST_SUCCESSFUL_WRITE";
+  private static final String LAST_WRITE_KEY = "LAST_WRITE";
```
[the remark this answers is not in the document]
narendly: Added!

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: Changed

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: I prefer writing it out for readability.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
narendly: I like writing it out for readability..

narendly:

narendly:
Using the following commit message.

This diff improves parallelism and throughput for ZkBucketDataAccessor. It implements the following ideas:

Optimistic Concurrency Control
Monotonically Increasing Version Number
Garbage Collection of Stale Metadata
TTL for stale metadata

Issue #508, opened 2019-10-10
The WAGED rebalancer should fail the cluster model creating if resource usage config is invalid


jiajunwang:
The current assumption is that on invalid resource usage config, the rebalancer assumes zero usage for the resource partitions.
After discussing with our potential customers, this could be dangerous. We need to apply a more strict check for resource usage, in addition to the instance capacity.

PR #511, opened 2019-10-14
More strict partition weight validation while creating the cluster model.

jiajunwang:
### Issues

#508

### Description

More strict partition weight validation while creating the cluster model.
1. If any capacity key is not configured in the Resource Config (or default weight) as the partition weight, the config is invalid.
2. If any partition weight is configured with a negative number, the config is invalid.
Note that the rebalancer will not compute a new assignment if any capacity/weight config is invalid.

This change is based on our discussion with potential customers. The expectation is to try our best to avoid confusing rebalance behavior.

### Tests

TestAssignableReplica
TestClusterConfig

[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 1044, Failures: 3, Errors: 0, Skipped: 4
[INFO] 
[INFO] BUILD FAILURE

Re-run

[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 31.022 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS

### Commits

### Documentation

NA

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
@@ -160,8 +160,6 @@ public class AssignableReplica implements Comparable<AssignableReplica> {
         capacityMap.getOrDefault(ResourceConfig.DEFAULT_PARTITION_KEY, new HashMap<>())));
 
     List<String> requiredCapacityKeys = clusterConfig.getInstanceCapacityKeys();
-    // Remove the non-required capacity items.
-    partitionCapacity.keySet().retainAll(requiredCapacityKeys);
     // If any required capacity key is not configured in the resource config, fail the model creating.
     if (!partitionCapacity.keySet().containsAll(requiredCapacityKeys)) {
       throw new HelixException(String.format(
@@ -169,6 +167,8 @@ public class AssignableReplica implements Comparable<AssignableReplica> {
           requiredCapacityKeys.toString(), resourceConfig.getResourceName(), partitionName,
           partitionCapacity.toString()));
     }
+    // Remove the non-required capacity items.
+    partitionCapacity.keySet().retainAll(requiredCapacityKeys);
 
     return partitionCapacity;
   }
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.helix.HelixException;
 import org.apache.helix.model.ClusterConfig;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
@@ -151,22 +152,24 @@ public class AssignableReplica implements Comparable<AssignableReplica> {
 
     Map<String, Integer> partitionCapacity = capacityMap.get(partitionName);
     if (partitionCapacity == null) {
-      partitionCapacity = capacityMap.get(ResourceConfig.DEFAULT_PARTITION_KEY);
+      partitionCapacity =
+          capacityMap.getOrDefault(ResourceConfig.DEFAULT_PARTITION_KEY, new HashMap<>());
     }
-    if (partitionCapacity == null) {
-      LOG.warn("The capacity usage of the specified partition {} is not configured in the Resource"
-          + " Config {}. No default partition capacity is configured either. Will proceed with"
-          + " empty capacity configuration.", partitionName, resourceConfig.getResourceName());
-      partitionCapacity = new HashMap<>();
+
+    for (Map.Entry<String, Integer> capacityEntry : clusterConfig.getDefaultPartitionWeightMap()
+        .entrySet()) {
+      partitionCapacity.putIfAbsent(capacityEntry.getKey(), capacityEntry.getValue());
     }
 
     List<String> requiredCapacityKeys = clusterConfig.getInstanceCapacityKeys();
     // Remove the non-required capacity items.
     partitionCapacity.keySet().retainAll(requiredCapacityKeys);
-    // If any required capacity key is not configured in the resource config, fill the partition
-    // capacity map with 0 usage.
-    for (String capacityKey : requiredCapacityKeys) {
-      partitionCapacity.putIfAbsent(capacityKey, 0);
+    // If any required capacity key is not configured in the resource config, fail the model creating.
+    if (!partitionCapacity.keySet().containsAll(requiredCapacityKeys)) {
+      throw new HelixException(String.format(
+          "The required capacity keys %s are not fully configured int the resource %s partition %s weight map %s.",
+          requiredCapacityKeys.toString(), resourceConfig.getResourceName(), partitionName,
+          partitionCapacity.toString()));
     }
 
     return partitionCapacity;
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
@@ -150,16 +150,14 @@ public class AssignableReplica implements Comparable<AssignableReplica> {
               .getResourceName(), ex);
     }
 
-    Map<String, Integer> partitionCapacity = capacityMap.get(partitionName);
-    if (partitionCapacity == null) {
-      partitionCapacity =
-          capacityMap.getOrDefault(ResourceConfig.DEFAULT_PARTITION_KEY, new HashMap<>());
-    }
-
-    for (Map.Entry<String, Integer> capacityEntry : clusterConfig.getDefaultPartitionWeightMap()
-        .entrySet()) {
-      partitionCapacity.putIfAbsent(capacityEntry.getKey(), capacityEntry.getValue());
-    }
+    // Fetch the capacity of partition from 3 possible sources according to the following priority.
+    // 1. The partition capacity that is explicitly configured in the resource config.
+    // 2. Or, the default partition capacity that is configured under partition name DEFAULT_PARTITION_KEY in the resource config.
+    // 3. If the default partition capacity that is configured in the cluster config contains more capacity keys, fill the capacity map with those additional values.
+    Map<String, Integer> partitionCapacity =
+        new HashMap<>(clusterConfig.getDefaultPartitionWeightMap());
+    partitionCapacity.putAll(capacityMap.getOrDefault(partitionName,
+        capacityMap.getOrDefault(ResourceConfig.DEFAULT_PARTITION_KEY, new HashMap<>())));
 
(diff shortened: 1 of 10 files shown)


helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -705,49 +706,80 @@ public void setInstanceCapacityKeys(List<String> capacityKeys) {
...
    * @throws IllegalArgumentException - when any of the data value is a negative number or when the map is empty
-   *
-   * This information is required by the global rebalancer.
+   *                                  This information is required by the global rebalancer.
    * @see <a href="Rebalance Algorithm">
-   *   https://github.com/apache/helix/wiki/Design-Proposal---Weight-Aware-Globally-Even-Distribute-Rebalancer#rebalance-algorithm-adapter
-   *   </a>
+   * https://github.com/apache/helix/wiki/Design-Proposal---Weight-Aware-Globally-Even-Distribute-Rebalancer#rebalance-algorithm-adapter
+   * </a>
    * If the instance capacity is not configured in neither Instance Config nor Cluster Config, the
```
narendly: neither -> either

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -705,49 +706,80 @@ public void setInstanceCapacityKeys(List<String> capacityKeys) {
...
    * cluster topology is considered invalid. So the rebalancer may stop working.
```
narendly: Is this description valid anymore? We aren't throwing an exception anymore.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: changed

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: We throw an exception while the set map parameters are invalid.
As for the rebalancer, we are not throwing an exception. But the new assignment won't be calculated (rebalance stopped) on invalid cluster status. So the description is still valid.

jiajunwang:
> Suggestion on the description of the PR. You say "a more strict check", but you don't actually describe it. So the reviewer has no clue what the "more strict check" should be. I think if you could actually describe it, something like, "previously, if a resource's weight is not set, we consider it 0. But we will change that behavior so that if a resource's weight is not set, we will throw an exception and halt the rebalancing."
> Then the reviewer knows what to watch out for and whether the code change actually reflects that change.

Sure. Updated the description.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
```diff
@@ -151,22 +152,24 @@ public static String generateReplicaKey(String resourceName, String partitionNam
 
     Map<String, Integer> partitionCapacity = capacityMap.get(partitionName);
     if (partitionCapacity == null) {
-      partitionCapacity = capacityMap.get(ResourceConfig.DEFAULT_PARTITION_KEY);
+      partitionCapacity =
```
narendly: Let's clearly document the behavior here as a comment - I think what we used to have as a warn log could be changed to an inline comment here. Also, let's say that if partition capacity usage is not found in ResourceConfig, we will try to get it from ClusterConfig.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
```diff
@@ -151,22 +152,24 @@ public static String generateReplicaKey(String resourceName, String partitionNam
...
-          + " Config {}. No default partition capacity is configured either. Will proceed with"
-          + " empty capacity configuration.", partitionName, resourceConfig.getResourceName());
-      partitionCapacity = new HashMap<>();
+
+    for (Map.Entry<String, Integer> capacityEntry : clusterConfig.getDefaultPartitionWeightMap()
+        .entrySet()) {
+      partitionCapacity.putIfAbsent(capacityEntry.getKey(), capacityEntry.getValue());
     }
 
     List<String> requiredCapacityKeys = clusterConfig.getInstanceCapacityKeys();
```
narendly: Would it be better to just iterate over `requiredCapacityKeys` in the first place starting in line 159?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
```diff
@@ -151,22 +152,24 @@ public static String generateReplicaKey(String resourceName, String partitionNam
...
     // Remove the non-required capacity items.
     partitionCapacity.keySet().retainAll(requiredCapacityKeys);
-    // If any required capacity key is not configured in the resource config, fill the partition
-    // capacity map with 0 usage.
-    for (String capacityKey : requiredCapacityKeys) {
-      partitionCapacity.putIfAbsent(capacityKey, 0);
+    // If any required capacity key is not configured in the resource config, fail the model creating.
+    if (!partitionCapacity.keySet().containsAll(requiredCapacityKeys)) {
```
narendly: Could we do this check before `partitionCapacity.keySet().retainAll(requiredCapacityKeys);`?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
```diff
@@ -151,22 +152,24 @@ public static String generateReplicaKey(String resourceName, String partitionNam
...
+      throw new HelixException(String.format(
+          "The required capacity keys %s are not fully configured int the resource %s partition %s weight map %s.",
```
narendly: int the resource -> typo

Also, I think it would be good if you could put ":" or something so we could easily tell which one's a variable name. (like resource: %s, partition: %s, etc..)

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -705,49 +706,78 @@ public void setInstanceCapacityKeys(List<String> capacityKeys) {
...
+      return capacityData.entrySet().stream().collect(
+          Collectors.toMap(entry -> entry.getKey(), entry -> Integer.parseInt(entry.getValue())));
+    }
+    return Collections.emptyMap();
+  }
 
+  public void setDefaultCapacityMap(ClusterConfigProperty capacityPropertyType,
+      Map<String, Integer> capacityDataMap) throws IllegalArgumentException {
+    if (capacityDataMap == null || capacityDataMap.size() == 0) {
+      throw new IllegalArgumentException("Default capacity data is empty");
```
narendly: `capacityDataMap.size() == 0`

I think this might actually be a valid thing for users to do, so we shouldn't throw an exception? Are they not allowed to empty out the default capacity map? They should be allowed to for stricter capacity planning, correct?

Or we provide a separate method for `reset/removeDefaultCapacityMap`?

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/model/TestAssignableReplica.java
```diff
@@ -102,6 +103,24 @@ public void testConstructReplicaWithResourceConfig() throws IOException {
     Assert.assertEquals(replica.getResourceMaxPartitionsPerInstance(), maxPartition);
   }
 
+  @Test
```
narendly: Nit: I think adding some JavaDoc here would be nice: 

E.g.) Tests that if default partition weight map is configured in ClusterConfig and NOT in ResourceConfig, AssignableReplica actually will get the default weight from ClusterConfig even though it's not set in ResourceConfig.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
```diff
@@ -166,19 +166,71 @@ public void testSetInstanceCapacityMap() {
...
+        DEFAULT_PARTITION_WEIGHT_MAP.name()), weightDataMapString);
+  }
+
+  @Test(expectedExceptions = IllegalArgumentException.class, expectedExceptionsMessageRegExp = "Default capacity data is empty")
+  public void testSetPartitionWeightMapEmpty() {
+    Map<String, Integer> weightDataMap = new HashMap<>();
+
+    ClusterConfig testConfig = new ClusterConfig("testConfig");
+    testConfig.setDefaultPartitionWeightMap(weightDataMap);
+  }
```
narendly: Do we allow users to reset or remove? Or would they have to explicitly set everything to 0 (which could be quite tedious..)

jiajunwang: This is equal to:

Map<String, Integer> weightDataMap = clusterConfig.getDefaultPartitionWeightMap();
weightDataMap.remove(...) or weightDataMap.put(...)
testConfig.setDefaultPartitionWeightMap(weightDataMap);

As for the initial value, they need to be explicitly set to 0. This is to ensure the rebalancer understands user's requirement correctly.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: Good point. I will change it. I think I will need to do the same for participant default capacity.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
jiajunwang: Sure I can. But I want to understand the reason of your suggestion?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
jiajunwang: No. In that case, I need to check if the key does not exist on the map. And even that's the case, I will still need to keep the later check to validate. So it will just increase the complicity.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
jiajunwang: Sure. Sounds good.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
jiajunwang: Will change it.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
jiajunwang: If what you meant is how the users are supposed to remove the default config, they can do so by putting an empty map to the config.

narendly: If you allow an empty map, then I think it would work. I see the change above. Thanks for making that change to allow an empty map to go through in the set function.

jiajunwang:

Issue #515, opened 2019-10-16
PropertyStore accessor that supports read/write content from/to ZK /PROPERTYSTORE


i3wangyi:
- Implement PropertyStore read API
- Implement PropertyStore write API

PR #516, opened 2019-10-16
Implement the propertyStore read endpoint

i3wangyi:
### Issues

(#515)
### Description

Implement the propertyStore read endpoint 
Example: http://xxxx/clusters/TestCluster/propertyStore/*<PATH>*
- The read method accepts a simple path parameter called "path" of the format "abc/abc/abc"
- The path is validated using regex
- The read operation parses the byte array content from a static bytearray zk base data accessor, if it's ZNRecord format, return ZnRecord; Otherwise, return {"content": <payload>}

Sample curl tests:
```
curl "http://localhost:8100/admin/v2/clusters/TestCluster/propertyStore/dummy" -X GET
{
  "/TestCluster/PROPERTYSTORE/dummy" : "dummyContent"
}

curl "http://localhost:8100/admin/v2/clusters/TestCluster/propertyStore/empty" -X GET   
{
  "/TestCluster/PROPERTYSTORE/empty" : ""
}

curl "http://localhost:8100/admin/v2/clusters/TestCluster/propertyStore/znrecord" -X GET
{
  "id" : "TestCluster",
  "simpleFields" : {
  },
  "mapFields" : {
  },
  "listFields" : {
  }
}

curl "http://localhost:8100/admin/v2/clusters/TestCluster/propertyStore/nonexist" -X GET
The property store path /TestCluster/PROPERTYSTORE/nonexist doesn't exist
```

### Tests

TestPropertyStoreAccessor

[INFO] Tests run: 90, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 118.383 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 90, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  02:07 min
[INFO] Finished at: 2019-11-27T09:56:06-08:00

### Commits

### Documentation

### Code Quality


helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,68 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Path("/clusters/{clusterId}/propertyStore")
+public class PropertyStoreAccessor extends AbstractHelixResource {
+  private static Logger LOG = LoggerFactory.getLogger(PropertyStoreAccessor.class);
+
+  @GET
+  public Response getPropertyByPath(@PathParam("clusterId") String clusterId,
+      @QueryParam("path") String urlEncodedPath) {
```
narendly: I'm thinking that for the ease of use from the user's perspective, something like this might be cleaner:

Making the path a PathParam instead of a QueryParam (since we are really querying for an entity, path param makes sense):


 ```
  @GET
  @PathParam("{path : .+}") // escapes slashes
  public Response getPropertyByPath(@PathParam("clusterId") String clusterId,
      @PathParam("path") String znodePath)
```

What do you think? In this case, the user can directly pass in the path as in

`http://endpoint.linkedin.biz/clusters/CLUSTER/propertyStore/PATH
`
But we'd need to review what we define PATH should be. Should it be the complete path? or the path under /PROPERTYSTORE/? That would be a good thing to discuss.

Comment (pkuwm) [inline helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java]: @i3wangyi I was working on this a few weeks back. My idea is to use regex to parse the propertyStore path using `{path: .+}` so `/propertyStore/aa/bb/cc` would be parsed to get path = "/aa/bb/cc". Path doesn't have to be encoded or decoded.

File: helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+    String propertyStoreRootPath = PropertyPathBuilder.propertyStore(clusterId);
+    HelixPropertyStore<ZNRecord> propertyStore = new ZkHelixPropertyStore<>(
+        (ZkBaseDataAccessor<ZNRecord>) getDataAccssor(clusterId).getBaseDataAccessor(),
+        propertyStoreRootPath, Collections.emptyList());
+    try {
+      urlEncodedPath = decode(urlEncodedPath);
+    } catch (UnsupportedEncodingException e) {
+      LOG.error("The path {} is mis-inputted for cluster {}", urlEncodedPath, clusterId);
+      throw new HelixException("The path needs to be url encoded from of '/%/%/.../'");
```
pkuwm: Can we return a readable response to client, including failed status code, error message, etc? Like:
{
  "status": "failed",
  "message": "Path is invalid."
}

narendly: I would also consider whether we should return a valid error response instead of an exception here. Let's discuss this further to establish Helix API Service REST convention to figure out what the best thing to do should be here :)

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: It's the first idea for me as well. However, realizing the PATH parameter can be something nasty like "a/b/c/d", the url becomes unnatural `"http://endpoint.linkedin.biz/clusters/CLUSTER/propertyStore/a/b/c/d"`. Something like this will just confuse people in practice. The URL encode is natural to HTTP request to escape the '/'

i3wangyi: The problem is transmitting the "/aa/bb/cc" in the Http request URL. Something like this `"http://endpoint.linkedin.biz/clusters/CLUSTER/propertyStore/a/b/c/d"` doesn't look a legit case to me. Do you think the regex can help stop the directory like "/"?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: This type of exception falls into the category of "wrong client input parameter" and there's nothing wrong internally in the server. It makes sense to throw/return exception in this case. However, I may choose to use some standard http exception with proper http status code.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
pkuwm: I think it looks fine. For example, in Hadoop WebHDFS API, the path slash is also presented in the api url. 
```
curl -i -b cookie.jar "http://host:port/webhdfs/v1/tmp/file.txt?op=GET_BLOCK_LOCATIONS"
```
About encoding the path, if only the web ui uses the api, it is ok, as the front end could encode the path. But I don't think it is convenient for users to use the api manually, if users still have to run a script like python to encode the path. I always believe we need to consider user friendly.

i3wangyi: Thanks for the example of Hadoop API. I will make the change see if the PathParam is a doable approach (it should be since others already implement it)

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,76 @@
...
+public class PropertyStoreAccessor extends AbstractHelixResource {
+  private static Logger LOG = LoggerFactory.getLogger(PropertyStoreAccessor.class);
+
+  @GET
+  @Path("{path: .+}")
+  public Response getPropertyByPath(@PathParam("clusterId") String clusterId,
+      @PathParam("path") String path) {
+    path = "/" + path;
+    if (!isPathValid(path)) {
+      LOG.error("The path {} is mis-inputted for cluster {}", path, clusterId);
```
narendly: I think this sounds better:
"This propertyStore path {} is invalid for cluster {}"...

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,76 @@
...
+import org.slf4j.LoggerFactory;
+
+
+@Path("/clusters/{clusterId}/propertyStore")
+public class PropertyStoreAccessor extends AbstractHelixResource {
+  private static Logger LOG = LoggerFactory.getLogger(PropertyStoreAccessor.class);
+
+  @GET
+  @Path("{path: .+}")
+  public Response getPropertyByPath(@PathParam("clusterId") String clusterId,
```
narendly: Could we add a block comment for the function clearly defining what the path should be?

For example, if the ZK path is as follows:

/PROPERTYSTORE/abc/abc/abc

The path must be:

"/abc/abc/abc"

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,76 @@
...
+    if (!isPathValid(path)) {
+      LOG.error("The path {} is mis-inputted for cluster {}", path, clusterId);
+      throw new WebApplicationException(Response.status(Response.Status.BAD_REQUEST).entity(
+          "Invalid path string. Valid path strings use slash as the directory separator and names the location of ZNode")
+          .build());
+    }
+    String propertyStoreRootPath = PropertyPathBuilder.propertyStore(clusterId);
+    HelixPropertyStore<ZNRecord> propertyStore = new ZkHelixPropertyStore<>(
+        (ZkBaseDataAccessor<ZNRecord>) getDataAccssor(clusterId).getBaseDataAccessor(),
+        propertyStoreRootPath, Collections.emptyList());
```
narendly: Should we define a `getPropertyStore` function just like how we did for `getDataAccessor` so that we won't have to create one every time?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,76 @@
...
+    ZNRecord record = propertyStore.get(path, null, AccessOption.PERSISTENT);
```
narendly: Here's a concern:

What if the ZNode under a PropertyStore is not a valid ZNRecord ZNode? In that case, this will throw an exception.

I think the idea that we discussed was to either:
1) return an appropriate error response saying the data is invalid. I am leaning towards a 500.
2) Or return the ZNode binary (byte[]) in a ZNRecord format. The JSON would look like this:

{
  simpleField:
   "ZNODE_NAME" : "xxxxxx..."
}

What do you think? 

@dasahcc @jiajunwang @lei-xia Thoughts?

helix-rest/src/test/java/org/apache/helix/rest/server/TestPropertyStoreAccessor.java
```diff
@@ -0,0 +1,51 @@
...
+  @Test
+  public void testGetPropertyStore() {
+    String path = "/TaskRebalancer/Workflow_0/Context";
+    String result =
+        new JerseyUriRequestBuilder("clusters/{}/propertyStore" + path).format(TEST_CLUSTER)
+            .isBodyReturnExpected(true).get(this);
+    Assert.assertFalse(result.isEmpty());
+  }
+
+  //TODO: implement the method
```
narendly: What is this TODO for?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: How about returning the raw data (`byte[]`) and let the clients parse the content. Helix-rest only supports raw read/write APIs and don't make any assumption about the underlying data structure.

narendly: @i3wangyi  That might be problematic. Many users of Helix REST already assume the JSON format as the format of the return value. And as a matter of fact, that is the established protocol for Helix REST endpoints. The format I suggested above

```
{
simpleField:
"ZNODE_NAME" : "xxxxxx..."
}
```
is a flexible format that meets all requirements.

i3wangyi: It's fair. The problem with our current accessors is they all assume ZNRecord property as the returned value. I've updated the PR using a pure ZKClient to retrieve propertyStore data. 
Please take a quick look and see if it's acceptable behavior. In the meantime, I'm figuring out how to add proper integration test to validate non-znRecord data. @narendly

helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java
```diff
@@ -0,0 +1,50 @@
...
+ * under the License.
+ */
+
+import org.apache.helix.ZNRecord;
+import org.apache.helix.manager.zk.client.HelixZkClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class PropertyStoreService {
```
narendly: I see what you're doing, but I don't think a whole class is necessary for this purpose. See my comment above.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,85 @@
...
+      @PathParam("path") String path) {
+    path = "/" + path;
+    if (!isPathValid(path)) {
+      LOG.error("The propertyStore path {} is invalid for cluster {}", path, clusterId);
+      throw new WebApplicationException(Response.status(Response.Status.BAD_REQUEST).entity(
+          "Invalid path string. Valid path strings use slash as the directory separator and names the location of ZNode")
+          .build());
+    }
+    String propertyStoreRootPath = PropertyPathBuilder.propertyStore(clusterId);
+    PropertyStoreService propertyStoreService = new PropertyStoreService(getHelixZkClient());
```
narendly: Here's an idea instead of creating PropertyStoreService:

1. Add `getPropertyStore()` method in `AbstractHelixResource` so that we don't create one every time this endpoint gets called.
2. Try to read the given path using `propertyStore` using a try-catch
3. Upon failure, try to use a `ZkClient` (use `getHelixZkClient()`) to construct a ZNRecord and return it using

```
    ZNRecord znRecord = new ZNRecord(path);
    znRecord.setSimpleField(path, content.toString());
    return znRecord;
```.
This would be sufficient. I don't think creating another class/abstraction is necessary since the logic is simple and self-contained.

Comment (i3wangyi) [reply helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java]: I've read through the impl of ZKHelixPropertyStore and it's more complicated than the rest requirement (read and write only) and only serves for ZNRecord. Thus I don't think we'll need to have a method `getPropertyStore()` of getting the property store in rest. 

I could implement one without the "service" and only use the raw zKClient, which I think it's less readable and testable.

File: helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java (hunk 8 above)
Comment (i3wangyi) [reply helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java]: Please see changes at the latest commit, if you feel without the class the code is cleaner we can go with that route

Comment (narendly) [reply helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java]: We should try to avoid using ZkClient directly to read and write. I did see your comment. I think it's possible to do this without this new class and using getPropertyStore().

- You could create a private object using new ZkHelixPropertyStore<ZNRecord>(). If you think this is hard, you could use ZkBaseDataAccessor<ZNRecord> as the last resort.
- If ZNode we are reading is not in the ZNRecord format, we could catch that exception and do appropriate processing (say, include it in the mapField of Json).

The second step actually would need some work. You could define a dummy serializer that simply reads the byte array and converts it to a String value in the Json. 

Does this make sense? Let me know if you have any questions.

File: helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java (hunk 9 above)
Comment (narendly) [reply helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java]: - I still don't think "Service" is necessary.
- We should avoid using the raw ZkClient.
See my comment above.

File: helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java (hunk 8 above)
Comment (i3wangyi) [reply helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java]: Let's use ZKClient directly at this PR, it's just simple. My latest commit the Service class is no longer used at all. I've taken a look at the ZkHelixPropertyStore class which provides a lot of overheads that are not needed at all in Helix API service. 

Ideally, I will prefer to use ZkBaseDataAccessor with the Non-ZnRecord custom serializer support(add a getBaseDataAccesor method in abstract class)  (will be my next PR).

Comment (narendly) [reply helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java]: @i3wangyi The reason to avoid using ZkClient directly is because we need to allow for the possibility of redirection at the base data accessor level. 
Consider a case where Helix uses multiple metadata stores in the future?

Comment (i3wangyi) [reply helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java]: I couldn't agree with you on the long long future plan, without detailed migration plans it's hard to convince ppl to give up the usage of a public class (zkClient) plus the usage is only with Helix Rest. The introduced overhead from ZKPropertyStore class where it registers cache/ cache update thread, ..., will make it more complicated to debug than the simple ZKClient method. I'm afraid the usage becomes a magic blackbox when the return value is NULL you don't know it means the value is empty or not a znRecord record.
I will prefer adding a todo here once we have the ZkBaseDataAccessor with the Non-ZnRecord custom serializer support, I could use the baseDataAccessror than the ZkClient.

Comment (narendly) [reply helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java]: Let us discuss using ZkBaseDataAccessor so that we could achieve this without using ZkClient directly here. I think it's possible. Let us sync in person.

Comment (i3wangyi) [reply helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java]: Does it mean the PR will need to wait on the baseDataAccesor's non-znRecord methods ready? It's likely to be so. Any ideas to expedite and run things in parallel?

Comment (narendly) [reply helix-rest/src/main/java/org/apache/helix/rest/server/service/PropertyStoreService.java]: We could have the first PR to be the addition of the "custom" ZkClient that just reads byte[]. This will be the building block for all of the other PRs. 

That shouldn't be too complicated, and upon completion of that, you could rebase other PRs to be able to leverage that. Does that sound good to you?

File: helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java [hunk 10]
```diff
@@ -0,0 +1,102 @@
...
+   * @param path path parameter is like "abc/abc/abc" in the URL
+   * @return JSON object as the response
+   */
+  @GET
+  @Path("{path: .+}")
+  public Response getPropertyByPath(@PathParam("clusterId") String clusterId,
+      @PathParam("path") String path) {
+    path = "/" + path;
+    if (!isPathValid(path )) {
+      LOG.error("The propertyStore path {} is invalid for cluster {}", path, clusterId);
```
lei-xia: For use input error, do we need to log error?  This is not really  any  error in our system, right?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,102 @@
...
+        } catch (ClassCastException e) {
+          LOG.warn("The content of node at path {} is not ZNRecord format", recordPath);
+        }
+        // fallback to a default and simple ZNRecord
+        ZNRecord znRecord = new ZNRecord(recordPath);
+        znRecord.setSimpleField(recordPath, new String(bytes));
+        return znRecord;
+      }
+    };
+    ZkBaseDataAccessor<ZNRecord> propertyStoreDataAccessor = getZkBaseDataAccessor(propertyStoreDataSerializer);
```
lei-xia: getZkBaseDataAccessor() won't work as expected here, right? You create a  new ZkSerializer here every time, and inside getZkBaseDataAccessor()  the map just use the reference to the serializer as the key, so it will always create a new BaseDataAccessor and add into the  map, which will cause ZkClient leak.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,102 @@
...
+          "Invalid path string. Valid path strings use slash as the directory separator and names the location of ZNode")
+          .build());
+    }
+    final String recordPath = PropertyPathBuilder.propertyStore(clusterId) + path;
+    ZkSerializer propertyStoreDataSerializer = new ZkSerializer() {
+      @Override
+      public byte[] serialize(Object content)
+          throws ZkMarshallingError {
+        // No need, it's used only for write-api
+        return new byte[0];
```
lei-xia: This is confusion, if the serialize is not used, let us throw an exception  (such as UnsupportedMethodException?)

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -117,9 +121,20 @@ public HelixDataAccessor getDataAccssor(String clusterName) {
     }
   }
 
+  public ZkBaseDataAccessor<ZNRecord> getZkBaseDataAccessor(ZkSerializer serializer) {
+    synchronized (_zkBaseDataAccessorBySerializer) {
+      if (!_zkBaseDataAccessorBySerializer.containsKey(serializer)) {
```
lei-xia: The key of this map is the reference to the ZkSerializer, your later code where this method is called always  creates a new ZkSerializer and passes to this  method, that  will always  create  a new  BaseDataAccessor, and  since the basedataAccessor created here never get closed, it will cause  ZkClient leak.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -117,9 +121,20 @@ public HelixDataAccessor getDataAccssor(String clusterName) {
...
+        ZkBaseDataAccessor<ZNRecord> baseDataAccessor = new ZkBaseDataAccessor<>(_zkAddr, serializer);
+        _zkBaseDataAccessorBySerializer.put(serializer, baseDataAccessor);
+      }
+    }
+    return _zkBaseDataAccessorBySerializer.get(serializer);
```
lei-xia: Should we close the BaseDataAccessor somewhere (instead of keeping it  forever until the server being shutdown)?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
narendly: +1

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
narendly: Did you mean to do double-lock checking here? Like I pointed out in a different PR, we need to follow the format given here:

https://stackoverflow.com/questions/3578604/how-to-solve-the-double-checked-locking-is-broken-declaration-in-java

If you synchronize at the beginning, it will not allow parallel access to the map, which affect parallelism and therefore performance.

narendly: +1

I think what @lei-xia meant was that since you create a new ZkSerializer (propertyStoreDataSerializer) every time the endpoint is called, `_zkBaseDataAccessorBySerializer.containsKey(serializer)` will always return false and you'll end up creating a new ZkBaseDataAccessor.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -117,9 +121,20 @@ public HelixDataAccessor getDataAccssor(String clusterName) {
     }
   }
 
+  public ZkBaseDataAccessor<ZNRecord> getZkBaseDataAccessor(ZkSerializer serializer) {
```
narendly: Here's an alternative design: What do you think about creating a base data accesor on-demand every time this endpoint gets called? The REST server should be a stateless API server in general anyways, so I don't see anything vastly wrong with that approach. 

And the way we create ZkBaseDataAccessor, we use a SharedZkClient, so we know we aren't creating a ZkConnection every time (we would be re-using the ZkConnection)?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,102 @@
...
+      @PathParam("path") String path) {
+    path = "/" + path;
+    if (!isPathValid(path )) {
+      LOG.error("The propertyStore path {} is invalid for cluster {}", path, clusterId);
+      throw new WebApplicationException(Response.status(Response.Status.BAD_REQUEST).entity(
+          "Invalid path string. Valid path strings use slash as the directory separator and names the location of ZNode")
+          .build());
+    }
+    final String recordPath = PropertyPathBuilder.propertyStore(clusterId) + path;
+    ZkSerializer propertyStoreDataSerializer = new ZkSerializer() {
```
narendly: We are creating a new object here?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
i3wangyi: @lei-xia @narendly Thanks for catching it! I was thinking that a new baseDataAccessor per each zkSerializer (based on the content of serializer itself). My latest commit addresses the loophole of multiple zkConnections where the propertyStore serializer's hashcode is identical. To answer the question "shall we close the Zkconnection", I don't think the cost is much given the fact current HelixZkClient inside Helix-Rest server is never closed. As long as the helix-rest service is up and running, it keeps a constant connection to the zk cluster.

i3wangyi: Fixed. It's merely the existing implementations I copied from other methods in the same class.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: A new object IMO is still needed because each request's property store path is different.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: Updated the commit, actually, it will be used if we're going to add the propertyStore write API

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
i3wangyi: It's a workable solution. I took a look at the code, the difficulty is getting `String zkAddress` in accessor class. The current pool also creates only one zkConnection as well, isn't it?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
i3wangyi: *revise* my comment elsewhere, the `close` method in the class will close the connection when server is shut down. I don't see the right time in the rest server when we should close the connection: if closing it after request finished, the next time, it needs to establish a new connection as well.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: answered in the comment. The duplicated connections can be saved if the hashcode of ZkSerializers is the same.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: IMO it's a minor. Anyway, it gives us the ability to debug when things go wrong, right? Plus it's a rest server, we don't have the same concern of polluting the logs in an event-driven helix-controller world.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,117 @@
...
+      // fallback to a default and simple ZNRecord
+      ZNRecord znRecord = new ZNRecord(_path);
+      znRecord.setSimpleField(_path, new String(bytes));
+      return znRecord;
+    }
+
+    // The hash code is designed to be same for all PropertyStoreSerializer instance
+    @Override
+    public int hashCode() {
+      return "PropertyStoreSerializer".hashCode();
```
lei-xia: this.getClass().getName().hashCode()?

i3wangyi: done.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
lei-xia: I suggest to change it to INFO. ERROR in log means something is wrong, but this is expected, nothing actual is wrong.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
narendly: Sorry I don't understand your comment. Could you please clarify?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/AbstractResource.java
```diff
@@ -130,6 +130,10 @@ private String errorMsgToJson(String error) {
   }
 
   protected Response JSONRepresentation(Object entity) {
+    if (entity == null) {
+      // return empty content when entity is null
+      return OK("");
+    }
```
narendly: 1) Why is this necessary all of a sudden?
2) Shouldn't we do checks like this in toJson function?

i3wangyi: Because I noticed getting the data from a non-existed path gives me the string result 
"null". It should not be the expected

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
i3wangyi: For your idea, the propertyStoreAccessor needs to `new ZkBaseDataAccessor(_zkAddress)`
every time, the `_zkAddress` is a private field. We need to understand deeper that the ZkBaseDataAccessor constructor uses the same ZkConnection to realize that we're not creating multiple ZkConnections. We'd better decouple the logic and this kind of pre-knowledge. 

My implementation of the proposed "zkBaseDataAccessor" pool also creates only one instance of zkConnection for propertyStoreAccessor as well, isn't it?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/AbstractResource.java
narendly: I don't understand your comment. What's wrong with returning null when the ZNode doesn't exist?

If you return return OK(""), then doesn't that mean that there exists a ZNode with empty content?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
narendly: Ping?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
narendly: > We need to understand deeper that the ZkBaseDataAccessor constructor uses the same ZkConnection to realize that we're not creating multiple ZkConnections. We'd better decouple the logic and this kind of pre-knowledge.

You could figure this out by reading how SharedZkClient is implemented. It shares one ZkConnection.

> My implementation of the proposed "zkBaseDataAccessor" pool also creates only one instance of zkConnection for propertyStoreAccessor as well, isn't it?

I believe your current implementation creates a ZkBaseDataAccessor instance per serializer. In a pathological case, I call your read method with 20K serializers, then you would have 20K different ZkBaseDataAccessors?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
narendly: Ping - see my comment about 20K different serializers?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
i3wangyi: My current implementation only creates one instance for all operations related to propertyStore, it's verified in the unit test.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/AbstractResource.java
i3wangyi: Currently, we're returning string "null" if the data path doesn't exist. One can interpret as 1. the content in the data path is "null" 2. The data path doesn't exist. Isn't it confusing?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: The input is wrong thus log level is ERROR. If clients want to know the error, we could grep from the server logs. 
It's actually very common usage, I don't think it's a serious issue at all

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,129 @@
...
+      return (byte[]) o;
+    }
+
+    // used for reading the raw content of property store path to ZnRecord format
+    @Override
+    public Object deserialize(byte[] bytes)
+        throws ZkMarshallingError {
+      if (bytes == null || bytes.length == 0) {
+        throw new ZkMarshallingError("Data unavailable from path: " + _path);
+      }
```
narendly: Do we throw a ZkMarshallingError if the ZNode content is empty?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,129 @@
...
+      // first, try to deserialize the bytearray into ZnRecord using {@link ZNRecordSerializer}
+      ZNRecord content = (ZNRecord) ZN_RECORD_SERIALIZER.deserialize(bytes);
+      // if first trial fails, fallback to return a simple/default znRecord
```
narendly: "fails"? shouldn't this throw an exception? and we should catch it?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,129 @@
...
+      if (content == null) {
```
narendly: Does null mean that there is data and ZNRecordSerializer wasn't able to deserialize it?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,129 @@
...
+        ZNRecord znRecord = new ZNRecord(_path);
+        znRecord.setSimpleField(_path, new String(bytes));
```
narendly: Not sure if _path is the right key here. Could we use a different key to make it easier for users to use?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,129 @@
...
+        return znRecord;
+      }
+      return content;
+    }
+
+    // The hashCode and equals methods' implementations ensure all instances are the same
+    @Override
```
narendly: If you're going to keep one object, I don't understand why you chose to have a map for it. Wouldn't AtomicReference or just a plain private variable be sufficient?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/AbstractResource.java
dasahcc: This should not be the right behavior. We need to return 400 for path not found instead of 200 for OK.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi:
```
if (bytes == null || bytes.length == 0) {
      // reading a parent/null node
      return null;
    }

    ByteArrayInputStream bais = new ByteArrayInputStream(bytes);

    ObjectMapper mapper = new ObjectMapper();
    DeserializationConfig deserializationConfig = mapper.getDeserializationConfig();
    deserializationConfig.set(DeserializationConfig.Feature.AUTO_DETECT_FIELDS, true);
    deserializationConfig.set(DeserializationConfig.Feature.AUTO_DETECT_SETTERS, true);
    deserializationConfig.set(DeserializationConfig.Feature.FAIL_ON_UNKNOWN_PROPERTIES, true);
    try {
      //decompress the data if its already compressed
      if (GZipCompressionUtil.isCompressed(bytes)) {
        byte[] uncompressedBytes = GZipCompressionUtil.uncompress(bais);
        bais = new ByteArrayInputStream(uncompressedBytes);
      }
      ZNRecord zn = mapper.readValue(bais, ZNRecord.class);

      return zn;
    } catch (Exception e) {
      logger.error("Exception during deserialization of bytes: " + new String(bytes), e);
      return null;
    }
```
It's restricted by current implementation of ZNRecordSerializer, you can see 
1. No exception will be thrown
2. Only null will tell us it failed to deserialize.
I'm not very comfortable to change the implementation in ZNRecordSerializer class

i3wangyi:
> Why this PR contains customized serializer change? I would suggest to separate the PR. 1 for customized serializer change and another for propertyStore REST.

The property store accessor just re-uses the existing features by the custom serializer and I think it's needed for parsing the content under PropertyStore. The PR is all about property store read api only.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: Say if you'd like to provide a custom serializer for reading the content in the MetaDataStore, the map is the place for keeping accessor instances. But I agree, we need to be extra careful not causing resource leakages.

dasahcc:
> > Why this PR contains customized serializer change? I would suggest to separate the PR. 1 for customized serializer change and another for propertyStore REST.
> 
> The property store accessor just re-uses the existing features by the custom serializer and I think it's needed for parsing the content under PropertyStore. The PR is all about property store read api only.

I am confused about that. For propertyStore, it uses ZNRecord. Default serializer should be fine. Why it depends on the customized serializer.

Even we need that, we should separate them out. Have a separate customized serializer first. Let's review that. After that, let's talk about the propertyStore support. Dont merge them together.

i3wangyi:
> > > Why this PR contains customized serializer change? I would suggest to separate the PR. 1 for customized serializer change and another for propertyStore REST.
> > 
> > 
> > The property store accessor just re-uses the existing features by the custom serializer and I think it's needed for parsing the content under PropertyStore. The PR is all about property store read api only.
> 
> I am confused about that. For propertyStore, it uses ZNRecord. Default serializer should be fine. Why it depends on the customized serializer.
> 
> Even we need that, we should separate them out. Have a separate customized serializer first. Let's review that. After that, let's talk about the propertyStore support. Dont merge them together.

Does the property store always use ZNrecord? My understanding of the scope is to provide generic rest endpoints for users to read/write property store content no matter what format it is.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/AbstractResource.java
i3wangyi: Fixed. Returning 404 when the path doesn't exist

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: replied in other thread. It's limited by the current ZNRecordSerializer implementation

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: Updated, use a "default" as the key

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: It's no longer a concern given we'll verify path existence first

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -46,9 +48,9 @@
 
   // 1 Cluster name will correspond to 1 helix data accessor
   private final Map<String, HelixDataAccessor> _helixDataAccessorPool;
-
   // 1 Cluster name will correspond to 1 task driver
   private final Map<String, TaskDriver> _taskDriverPool;
+  final Map<ZkSerializer, ZkBaseDataAccessor<ZNRecord>> _zkBaseDataAccessorBySerializer;
```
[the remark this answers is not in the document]
i3wangyi: resolved. Remove the map

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: updated.

dasahcc:
> > > > Why this PR contains customized serializer change? I would suggest to separate the PR. 1 for customized serializer change and another for propertyStore REST.
> > > 
> > > 
> > > The property store accessor just re-uses the existing features by the custom serializer and I think it's needed for parsing the content under PropertyStore. The PR is all about property store read api only.
> > 
> > 
> > I am confused about that. For propertyStore, it uses ZNRecord. Default serializer should be fine. Why it depends on the customized serializer.
> > Even we need that, we should separate them out. Have a separate customized serializer first. Let's review that. After that, let's talk about the propertyStore support. Dont merge them together.
> 
> Does the property store always use ZNrecord? My understanding of the scope is to provide generic rest endpoints for users to read/write property store content no matter what format it is.


As I said, even if we need the part of customized serializer logic, let's have one PR for customized serializer. After it has been checkedin, let's have the implementation of PropertyStore.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,112 @@
...
+
+    // used for reading the raw content of property store path to ZnRecord format
+    @Override
+    public Object deserialize(byte[] bytes)
+        throws ZkMarshallingError {
+      // firstly, try to deserialize the bytearray into ZnRecord using {@link ZNRecordSerializer}
+      ZNRecord content = (ZNRecord) ZN_RECORD_SERIALIZER.deserialize(bytes);
+      // if first trial fails, fallback to return a simple/default znRecord
+      if (content == null) {
+        ZNRecord znRecord = new ZNRecord(DEFAULT_KEY);
```
dasahcc: Why we need create another ZNRecord? If it is failed to deserialize by ZNRecord, we can just simply return a json response instead of a complete ZNRecord with empty list fields and map fields.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,112 @@
...
+   * @return JSON object as the response
+   */
+  @GET
+  @Path("{path: .+}")
+  public Response getPropertyByPath(@PathParam("clusterId") String clusterId,
+      @PathParam("path") String path) {
+    path = "/" + path;
+    if (!isPathValid(path)) {
+      LOG.info("The propertyStore path {} is invalid for cluster {}", path, clusterId);
+      throw new WebApplicationException(Response.status(Response.Status.BAD_REQUEST).entity(
```
dasahcc: We do have badRequest() response function.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,112 @@
...
+   * Sample HTTP URLs:
+   *  http://<HOST>/clusters/{clusterId}/propertyStore/<PATH>
+   * It refers to the /PROPERTYSTORE/<PATH> in Helix metadata store
+   * @param clusterId The cluster Id
+   * @param path path parameter is like "abc/abc/abc" in the URL
+   * @return JSON object as the response
+   */
+  @GET
+  @Path("{path: .+}")
+  public Response getPropertyByPath(@PathParam("clusterId") String clusterId,
```
dasahcc: I think we should limit the path to under PROPERTYSTORE folder. @narendly This is what we agreed with new API change, right?

dasahcc:
> > > > > Why this PR contains customized serializer change? I would suggest to separate the PR. 1 for customized serializer change and another for propertyStore REST.
> > > > 
> > > > 
> > > > The property store accessor just re-uses the existing features by the custom serializer and I think it's needed for parsing the content under PropertyStore. The PR is all about property store read api only.
> > > 
> > > 
> > > I am confused about that. For propertyStore, it uses ZNRecord. Default serializer should be fine. Why it depends on the customized serializer.
> > > Even we need that, we should separate them out. Have a separate customized serializer first. Let's review that. After that, let's talk about the propertyStore support. Dont merge them together.
> > 
> > 
> > Does the property store always use ZNrecord? My understanding of the scope is to provide generic rest endpoints for users to read/write property store content no matter what format it is.
> 
> As I said, even if we need the part of customized serializer logic, let's have one PR for customized serializer. After it has been checkedin, let's have the implementation of PropertyStore.

Was confused with another PR for customized serialize. Synced up offline. This change should be self-contained.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: thanks! updated, the other NOT_FOUND response should be a valid case (badRequest() returns http code 400)

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: It's @narendly's idea, many clients already assume the content to be ZnRecord in their client-side library; I'm fine either one

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: Yes, client only needs to pass the subpath after `PROPERTYSTORE` and you can check the line `final String recordPath = PropertyPathBuilder.propertyStore(clusterId) + path;`.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: no longer valid.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
i3wangyi: no longer valid

i3wangyi:
latest mvn test under helix-rest folder, 

[INFO] Tests run: 90, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 31.416 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 90, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  38.012 s
[INFO] Finished at: 2019-11-25T17:55:22-08:00
****

Will rebase from master and resolve conflicts later

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
narendly: I am fine with either. Let's just return a JSON. That way, we could clearly differentiate whether it was a raw binary stream read or a ZNRecord read.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,112 @@
...
+
+    // used for reading the raw content of property store path to ZnRecord format
+    @Override
+    public Object deserialize(byte[] bytes)
+        throws ZkMarshallingError {
+      // firstly, try to deserialize the bytearray into ZnRecord using {@link ZNRecordSerializer}
+      ZNRecord content = (ZNRecord) ZN_RECORD_SERIALIZER.deserialize(bytes);
+      // if first trial fails, fallback to return a simple/default znRecord
+      if (content == null) {
+        ZNRecord znRecord = new ZNRecord(DEFAULT_KEY);
```
narendly: Instead of using "default", could we use the name of the ZNode instead as the key and the binary as the value of the JSON returned?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
narendly: Correct.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,112 @@
...
+          "Invalid path string. Valid path strings use slash as the directory separator and names the location of ZNode");
+    }
+    final String recordPath = PropertyPathBuilder.propertyStore(clusterId) + path;
+    ZkBaseDataAccessor<ZNRecord> propertyStoreDataAccessor = getPropertyStoreAccessor();
+    if (propertyStoreDataAccessor.exists(recordPath, AccessOption.PERSISTENT)) {
+      ZNRecord record = propertyStoreDataAccessor.get(recordPath, null, AccessOption.PERSISTENT);
+      return JSONRepresentation(record);
+    } else {
+      throw new WebApplicationException(Response.status(Response.Status.NOT_FOUND)
+          .entity("The property store path " + recordPath + " doesn't exist").build());
```
narendly: Nit: String.format?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -35,10 +36,12 @@
 import org.apache.helix.manager.zk.ZkClient;
 import org.apache.helix.manager.zk.client.HelixZkClient;
 import org.apache.helix.manager.zk.client.SharedZkClientFactory;
+import org.apache.helix.rest.server.resources.helix.PropertyStoreAccessor;
 import org.apache.helix.task.TaskDriver;
 import org.apache.helix.tools.ClusterSetup;
 
 public class ServerContext {
+  private static ZkBaseDataAccessor<ZNRecord> PROPERTY_STORE_ACCESSOR = null;
```
narendly: Isn't ServerContext itself kind of a static concept in Helix REST? Any reason you're making this a static var?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
narendly: I see. It's returning null on Exception. Could you document this as an inline comment - that ZNRecordSerializer's behavior is that it will return null on decompress exception?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
i3wangyi: No specific reason, it also works with simple `private ZkBaseDataAccessor<ZNRecord> propertyStoreAccessor`

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: done.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: Synced-up offline. It seems a bytearray based zkBaseDataAccessor suits the use case better.  I've updated the PR and tested the results.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: Updated in the comment

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,93 @@
...
+      byte[] bytes = propertyStoreDataAccessor.get(recordPath, null, AccessOption.PERSISTENT);
+      ZNRecord znRecord = (ZNRecord) ZN_RECORD_SERIALIZER.deserialize(bytes);
+      // The ZNRecordSerializer returns null when exception occurs in deserialization method
+      if (znRecord == null) {
+        ObjectNode jsonNode = OBJECT_MAPPER.createObjectNode();
+        jsonNode.put(recordPath, new String(bytes));
+        return JSONRepresentation(jsonNode);
+      } else {
+        return JSONRepresentation(znRecord);
+      }
```
narendly: Is this else clause necessary?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+    final String recordPath = PropertyPathBuilder.propertyStore(clusterId) + path;
+    ZkBaseDataAccessor<byte[]> propertyStoreDataAccessor = getByteArrayDataAccessor();
+    if (propertyStoreDataAccessor.exists(recordPath, AccessOption.PERSISTENT)) {
+      byte[] bytes = propertyStoreDataAccessor.get(recordPath, null, AccessOption.PERSISTENT);
+      ZNRecord znRecord = (ZNRecord) ZN_RECORD_SERIALIZER.deserialize(bytes);
+      // The ZNRecordSerializer returns null when exception occurs in deserialization method
+      if (znRecord == null) {
+        ObjectNode jsonNode = OBJECT_MAPPER.createObjectNode();
+        jsonNode.put(recordPath, new String(bytes));
```
narendly: Do you think it would be easier to use "content" or some fixed constant string for the key? Or do you think it'd be easier if we put the path here?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -118,6 +120,29 @@ public HelixDataAccessor getDataAccssor(String clusterName) {
     }
   }
 
+  public ZkBaseDataAccessor<byte[]> getByteArrayBaseStoreAccessor() {
```
narendly: Nit: fix the method name?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: good catch. fixed.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
i3wangyi: updated using "content" as the key

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
i3wangyi: done.

i3wangyi:
latest mvn test under helix-rest
```
9827 [pool-61-thread-4] ERROR org.apache.helix.rest.server.service.InstanceServiceImpl  - Custom client side instance level health check for TestCluster/instance0.linkedin.com_1235 failed.
java.io.IOException
Shut down zookeeper at port 2123 in thread main
Shut down zookeeper at port 2124 in thread main
[INFO] Tests run: 90, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 31.323 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 90, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  37.076 s
[INFO] Finished at: 2019-11-26T18:42:44-08:00

```

helix-rest/src/test/java/org/apache/helix/rest/server/TestPropertyStoreAccessor.java
```diff
@@ -0,0 +1,116 @@
...
+    Assert.assertEquals(record.getId(), TEST_ZNRECORD.getId());
+  }
+
+  @Test
+  public void testGetPropertyStoreWithTestStringData() throws IOException {
+    String actual =
+        new JerseyUriRequestBuilder("clusters/{}/propertyStore/NonZnRecord").format(TEST_CLUSTER)
+            .isBodyReturnExpected(true).get(this);
+    JsonNode jsonNode = OBJECT_MAPPER.readTree(actual);
+    String payLoad = jsonNode.get("content").getValueAsText();
```
narendly: Nit: Make "content" a constant?

i3wangyi: done

i3wangyi:
The PR is ready to be merged into master, approved by @narendly. Thanks!

Commit message:
```
Implement the propertyStore read endpoint 
Example: http://xxxx/clusters/TestCluster/propertyStore/*<PATH>*
- The read method accepts a simple path parameter called "path" of the format "abc/abc/abc"
- The path is validated using regex
- The read operation parses the byte array content from a static bytearray zk base data accessor, if it's ZNRecord format, return ZnRecord; Otherwise, return {"content": <payload>}
```

The latest mvn test in helix-rest
```
[INFO] Tests run: 90, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 118.383 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 90, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  02:07 min
[INFO] Finished at: 2019-11-27T09:56:06-08:00
```

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -118,6 +120,29 @@ public HelixDataAccessor getDataAccssor(String clusterName) {
...
 
+  public ZkBaseDataAccessor<byte[]> getByteArrayBaseDataAccessor() {
+    if (_byteArrayBaseDataAccessor == null) {
+      synchronized (this) {
+        if (_byteArrayBaseDataAccessor == null) {
+          _byteArrayBaseDataAccessor = new ZkBaseDataAccessor<>(_zkAddr, new ZkSerializer() {
+            @Override
+            public byte[] serialize(Object o)
+                throws ZkMarshallingError {
+              return (byte[]) o;
```
narendly: Since we don't support this yet, let's throw an unsupported exception?

i3wangyi: done.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -110,14 +113,40 @@ public ConfigAccessor getConfigAccessor() {
...
+    if (_byteArrayBaseDataAccessor == null) {
+      synchronized (this) {
+        if (_byteArrayBaseDataAccessor == null) {
+          _byteArrayBaseDataAccessor = new ZkBaseDataAccessor<>(_zkAddr, new ZkSerializer() {
+            @Override
+            public byte[] serialize(Object o)
+                throws ZkMarshallingError {
+              // TODO: will implement the write operation when it's needed
+              throw new UnsupportedOperationException(
+                  "The write operation hasn't been supported yet");
```
narendly: "Serialize is not supported yet!"

PR #528, opened 2019-10-23
Suggested fixes by iCR, OpenRefactory, Inc.

openrefactory:
OpenRefactory introduces fixers, that automatically fix safety, security, reliability and compliance issues. We ran OpenRefactory's Intelligent Code Repair (iCR) analysis engine on your code. We are reporting a few sample fixes that were generated.
The goal of the this fix is to resolve **Unchecked Method Return Value**, **Bad Flow in Switch** and **Weak Library Usage** problem. 

### Unchecked Method Return Value fixes:
In file: `TestFullAutoNodeTagging.java`, method:` withinAverage ` returns one or more error return value(s) that have not been checked properly. iCR suggested that the checks are done properly.

### Bad Flow in Switch fixes:
In file: `AbstractTaskDispatcher.java`, class: `AbstractTaskDispatcher`, method: `updatePreviousAssignedTasksStatus`  has a case statement that is not terminated by a break . iCR suggested changes in 1 files to resolve the problem.

In file: `TestExecutor.java`, class:` TestExecutor`, method:` getValueType`  has a case statement that is not terminated by a break . iCR suggested changes in 1 files to resolve the problem.

### Weak Library Usage fixes:
In file: `TaskExecutionDemo.java`, class: `TaskExecutionDemo,` method: `populateDummyData`  uses Random object but it may not be truly random. It should be replaced with SecureRandom. . iCR suggested changes in 1 files to resolve the problem.

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
+++ b/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
@@ -303,12 +303,14 @@ public abstract class AbstractTaskDispatcher {
 
             // Also release resources for these tasks
             assignableInstanceManager.release(instance, taskConfig, quotaType);
+            break;
           } else if (jobState == TaskState.IN_PROGRESS
               && (jobTgtState != TargetState.STOP && jobTgtState != TargetState.DELETE)) {
             // Job is in progress, implying that tasks are being re-tried, so set it to RUNNING
             paMap.put(pId,
                 new JobRebalancer.PartitionAssignment(instance, TaskPartitionState.RUNNING.name()));
             assignedPartitions.get(instance).add(pId);
+            break;
           }
         }
 
--- a/helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestFullAutoNodeTagging.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestFullAutoNodeTagging.java
@@ -459,6 +459,9 @@ public class TestFullAutoNodeTagging extends ZkUnitTestBase {
         upperBound = 2;
       }
       int average = computeAverage(countMap);
+      // OR Warning: Handle Error Code
+      if (average == -1) {
+      }
       for (String participantName : countMap.keySet()) {
         int count = countMap.get(participantName);
         if (count < average - 1 || count > average + upperBound) {
--- a/helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestFullAutoNodeTagging.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestFullAutoNodeTagging.java
@@ -459,8 +459,8 @@ public class TestFullAutoNodeTagging extends ZkUnitTestBase {
         upperBound = 2;
       }
       int average = computeAverage(countMap);
-      // OR Warning: Handle Error Code
       if (average == -1) {
+        return false;
       }
       for (String participantName : countMap.keySet()) {
         int count = countMap.get(participantName);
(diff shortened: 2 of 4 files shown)


helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestFullAutoNodeTagging.java
```diff
@@ -459,6 +459,9 @@ private boolean withinAverage(Map<String, Integer> countMap, boolean isEmptyAllo
         upperBound = 2;
       }
       int average = computeAverage(countMap);
+      // OR Warning: Handle Error Code
```
jiajunwang: I don't think the average would be -1 here. Since countMap size has been checked above.
Anyway, in this case, we should return false. Could you please modify it to:
if (average == -1) {
  return false;
}

openrefactory: Modification is done.

narendly:
Also please be advised that you might want to use a production-grade linux environment to run the mvn test suite.

recipes/task-execution/src/main/java/org/apache/helix/taskexecution/TaskExecutionDemo.java
```diff
@@ -121,7 +121,7 @@ private static void populateDummyData(TaskResultStore taskResultStore) throws Ex
     float fraudProbability = 0.01f;
     float clickProbability = 0.01f;
     int numImps = NUM_IMP_EVENTS;
-    Random rand = new Random();
+    SecureRandom rand = new SecureRandom();
```
pkuwm: I wonder if SecureRandom is a must here. Are the random numbers here for security or cryptographic purpose?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -303,12 +303,14 @@ public void updatePreviousAssignedTasksStatus(
...
             // Also release resources for these tasks
             assignableInstanceManager.release(instance, taskConfig, quotaType);
+            break;
           } else if (jobState == TaskState.IN_PROGRESS
               && (jobTgtState != TargetState.STOP && jobTgtState != TargetState.DELETE)) {
             // Job is in progress, implying that tasks are being re-tried, so set it to RUNNING
             paMap.put(pId,
                 new JobRebalancer.PartitionAssignment(instance, TaskPartitionState.RUNNING.name()));
             assignedPartitions.get(instance).add(pId);
+            break;
```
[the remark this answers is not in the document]
openrefactory: Thanks for your feedback. Our software checks every possible path within a switch case. If inside any "if", "else if" or "else" the break is not introduced mistakenly, it suggests fix there. This fix suggestions are auto generated by our software.

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestFullAutoNodeTagging.java
```diff
@@ -459,6 +459,9 @@ private boolean withinAverage(Map<String, Integer> countMap, boolean isEmptyAllo
         upperBound = 2;
       }
       int average = computeAverage(countMap);
+      if (average == -1) {
+        return false;
+      }
```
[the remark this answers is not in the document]
openrefactory: Our software checks what are the return values of the method: `countMap`. If there is any error check value returned from the method and that value is not handled, our software suggests a fix.

openrefactory:
@jiajunwang we are getting this failure while running the command "mvn test" on helix's master branch. Would you mind suggesting something to solve this issue? 

An image has been attached below.

mvn-build

narendly:
@openrefactory That is a known unstable test. If that is the only failure, this PR should be ready to be merged

Issue #533, opened 2019-10-24
Add custom serialize/deserialize support for non-ZnRecord data


i3wangyi:
1. Propose and fix the design of the new method signatures
2. Add the implementations and tests

PR #534, opened 2019-10-24
Update zkBaseDataAccessor with custom serializer support

i3wangyi:
### Issues

#533

### Description

Update zkBaseDataAccessor with custom serializer support
- new constructor takes in the custom serializer as the parameter

### Tests

testSyncCreateWithCustomSerializer
testCustomAccessorCreateZnRecord
testDefaultAccessorCreateCustomData

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestBucketizedResource.testBounceDisableAndDrop:189 expected:<false> but was:<true>
[ERROR]   TestDrop.testBasic:113->assertEmptyCSandEV:62 expected:<true> but was:<false>
[ERROR]   TestZkConnectionLost.testLostZkConnection:140 » Helix Workflow "testLostZkConn...
[ERROR]   TestWtCacheAsyncOpMultiThread.testHappyPathZkCacheBaseDataAccessor:202 » IllegalState
[ERROR]   TestZkCacheSyncOpSingleThread.testZkCacheCallbackExternalOpNoChroot:96 » IllegalState
[INFO] 
[ERROR] Tests run: 885, Failures: 6, Errors: 0, Skipped: 0
[INFO] 

The single failed test passed when running in IDE

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
@@ -257,4 +257,9 @@ public interface BaseDataAccessor<T> {
    * reset the cache if any, when session expiry happens
    */
   void reset();
+
+  /**
+   * Close the opened zkConnection
+   */
+  void close();
 }
--- a/helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
@@ -259,7 +259,7 @@ public interface BaseDataAccessor<T> {
   void reset();
 
   /**
-   * Close the opened zkConnection
+   * Close the connection to the metadata store
    */
   void close();
 }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -136,14 +136,14 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    */
   @Override
   public boolean create(String path, T record, int options) {
-    AccessResult result = doCreate(path, record, options, true);
+    AccessResult result = doCreate(_zkClient, path, record, options);
     return result._retCode == RetCode.OK;
   }
 
   /**
    * sync create. In order to support non-ZnRecord creation, the instance of {@link HelixZkClient} needs to be a parameter
    */
-  public AccessResult doCreate(String path, Object record, int options, boolean isZnRecord) {
+  public AccessResult doCreate(HelixZkClient zkClient, String path, Object record, int options) {
     AccessResult result = new AccessResult();
     CreateMode mode = AccessOption.getMode(options);
     if (mode == null) {
@@ -152,7 +152,6 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
       return result;
     }
 
-    HelixZkClient zkClient = isZnRecord ? _zkClient : getNonZNRecordZkClient();
     boolean retry;
     do {
       retry = false;
@@ -165,7 +164,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
         // this will happen if parent node does not exist
         String parentPath = HelixUtil.getZkParentPath(path);
         try {
-          AccessResult res = doCreate(parentPath, null, AccessOption.PERSISTENT, isZnRecord);
+          AccessResult res = doCreate(zkClient, parentPath, null, AccessOption.PERSISTENT);
           result._pathCreated.addAll(res._pathCreated);
           RetCode rc = res._retCode;
           if (rc == RetCode.OK || rc == RetCode.NODE_EXISTS) {
@@ -205,7 +204,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    */
   @Override
   public boolean set(String path, T record, int expectVersion, int options) {
-    AccessResult result = doSet(path, record, expectVersion, options, true);
+    AccessResult result = doSet(_zkClient, path, record, expectVersion, options);
     return result._retCode == RetCode.OK;
   }
 
@@ -220,14 +219,14 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    * Sync set operation with custom serializer support
    */
   public boolean set(String path, Object record, int options, int expectVersion, ZkSerializer zkSerializer) {
-    AccessResult result = doSet(path, zkSerializer.serialize(record), expectVersion, options, false);
+    AccessResult result = doSet(getNonZNRecordZkClient(), path, zkSerializer.serialize(record), expectVersion, options);
     return result._retCode == RetCode.OK;
   }
 
   /**
    * sync set
    */
-  public AccessResult doSet(String path, Object record, int expectVersion, int options, boolean isZnRecord) {
+  public AccessResult doSet(HelixZkClient zkClient, String path, Object record, int expectVersion, int options) {
     AccessResult result = new AccessResult();
 
     CreateMode mode = AccessOption.getMode(options);
@@ -237,12 +236,11 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
       return result;
     }
 
-    HelixZkClient helixZkClient = isZnRecord ? _zkClient : getNonZNRecordZkClient();
     boolean retry;
     do {
       retry = false;
       try {
-        Stat stat = helixZkClient.writeDataGetStat(path, record, expectVersion);
+        Stat stat = zkClient.writeDataGetStat(path, record, expectVersion);
         DataTree.copyStat(stat, result._stat);
       } catch (ZkNoNodeException e) {
         // node not exists, try create if expectedVersion == -1; in this case, stat will not be set
@@ -253,7 +251,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
         }
         try {
           // may create recursively
-          AccessResult res = doCreate(path, record, options, true);
+          AccessResult res = doCreate(zkClient, path, record, options);
           result._pathCreated.addAll(res._pathCreated);
           RetCode rc = res._retCode;
           switch (rc) {
@@ -329,7 +327,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
           T newData = updater.update(null);
           RetCode rc;
           if (newData != null) {
-            AccessResult res = doCreate(path, newData, options, true);
+            AccessResult res = doCreate(_zkClient, path, newData, options);
             result._pathCreated.addAll(res._pathCreated);
             rc = res._retCode;
           } else {
@@ -402,7 +400,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    */
   public boolean create(String path, Object data, int options, ZkSerializer serializer) {
     byte[] bytes = serializer.serialize(data);
-    AccessResult result = doCreate(path, bytes, options, false);
+    AccessResult result = doCreate(getNonZNRecordZkClient(), path, bytes, options);
     return result._retCode == RetCode.OK;
   }
 
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -362,6 +362,9 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
       if (AccessOption.isThrowExceptionIfNotExist(options)) {
         throw ex;
       }
+    } catch (ZkMarshallingError ex) {
+      LOG.error("Failed to deserialize the byte array", ex);
+      throw ex;
     }
     return content;
   }
@@ -370,8 +373,14 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    * Sync create method with custom serializer support
    */
   public boolean create(String path, Object data, int options, ZkSerializer serializer) {
-    AccessResult result = doCreate(_nonZNRecordClient, path, serializer.serialize(data), options);
-    return result._retCode == RetCode.OK;
+    try {
+      byte[] bytes = serializer.serialize(data);
+      AccessResult result = doCreate(_nonZNRecordClient, path, bytes, options);
+      return result._retCode == RetCode.OK;
+    } catch (ZkMarshallingError ex) {
+      LOG.error("Failed to serialize the data object", ex);
+      throw ex;
+    }
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -88,8 +88,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
   private static Logger LOG = LoggerFactory.getLogger(ZkBaseDataAccessor.class);
 
   private final HelixZkClient _zkClient;
-  // onDemand zkClient for read/write non-znRecord data path;
-  // minimal change to support custom serializer
+  //ZkClient that will be lazily instantiated if there are operations for non-ZNRecord data
   private HelixZkClient _nonZNRecordZkClient = null;
 
   public ZkBaseDataAccessor(HelixZkClient zkClient) {
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -26,13 +26,16 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+
 import org.I0Itec.zkclient.DataUpdater;
 import org.I0Itec.zkclient.IZkChildListener;
 import org.I0Itec.zkclient.IZkDataListener;
 import org.I0Itec.zkclient.exception.ZkBadVersionException;
 import org.I0Itec.zkclient.exception.ZkException;
+import org.I0Itec.zkclient.exception.ZkMarshallingError;
 import org.I0Itec.zkclient.exception.ZkNoNodeException;
 import org.I0Itec.zkclient.exception.ZkNodeExistsException;
+import org.I0Itec.zkclient.serialize.ZkSerializer;
 import org.apache.helix.AccessOption;
 import org.apache.helix.BaseDataAccessor;
 import org.apache.helix.HelixException;
@@ -43,6 +46,7 @@ import org.apache.helix.manager.zk.ZkAsyncCallbacks.ExistsCallbackHandler;
 import org.apache.helix.manager.zk.ZkAsyncCallbacks.GetDataCallbackHandler;
 import org.apache.helix.manager.zk.ZkAsyncCallbacks.SetDataCallbackHandler;
 import org.apache.helix.manager.zk.client.HelixZkClient;
+import org.apache.helix.manager.zk.zookeeper.ZkConnection;
 import org.apache.helix.store.zk.ZNode;
 import org.apache.helix.util.HelixUtil;
 import org.apache.zookeeper.CreateMode;
@@ -84,12 +88,33 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
   private static Logger LOG = LoggerFactory.getLogger(ZkBaseDataAccessor.class);
 
   private final HelixZkClient _zkClient;
+  private final HelixZkClient _nonZNRecordClient;
 
   public ZkBaseDataAccessor(HelixZkClient zkClient) {
     if (zkClient == null) {
       throw new NullPointerException("zkclient is null");
     }
     _zkClient = zkClient;
+    _nonZNRecordClient = new ZkClient.Builder().setConnection(
+        new ZkConnection(zkClient.getServers(), ZkClient.DEFAULT_SESSION_TIMEOUT))
+        .setZkSerializer(new ZkSerializer() {
+          @Override
+          public byte[] serialize(Object data) throws ZkMarshallingError {
+            if (data instanceof byte[]) {
+              return (byte[]) data;
+            }
+            throw new HelixException("Only support a byte array as an argument!");
+          }
+
+          @Override
+          public Object deserialize(byte[] data) throws ZkMarshallingError {
+            return data;
+          }
+        }).build();
+  }
+
+  public HelixZkClient getZkClient() {
+    return _zkClient;
   }
 
   /**
@@ -97,14 +122,14 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    */
   @Override
   public boolean create(String path, T record, int options) {
-    AccessResult result = doCreate(path, record, options);
+    AccessResult result = doCreate(_zkClient, path, record, options);
     return result._retCode == RetCode.OK;
   }
 
   /**
-   * sync create
+   * sync create. In order to support non-ZnRecord creation, the instance of {@link HelixZkClient} needs to be a parameter
    */
-  public AccessResult doCreate(String path, T record, int options) {
+  public AccessResult doCreate(HelixZkClient zkClient, String path, Object record, int options) {
     AccessResult result = new AccessResult();
     CreateMode mode = AccessOption.getMode(options);
     if (mode == null) {
@@ -117,16 +142,15 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
     do {
       retry = false;
       try {
-        _zkClient.create(path, record, mode);
+        zkClient.create(path, record, mode);
         result._pathCreated.add(path);
-
         result._retCode = RetCode.OK;
         return result;
       } catch (ZkNoNodeException e) {
         // this will happen if parent node does not exist
         String parentPath = HelixUtil.getZkParentPath(path);
         try {
-          AccessResult res = doCreate(parentPath, null, AccessOption.PERSISTENT);
+          AccessResult res = doCreate(zkClient, parentPath, record, AccessOption.PERSISTENT);
           result._pathCreated.addAll(res._pathCreated);
           RetCode rc = res._retCode;
           if (rc == RetCode.OK || rc == RetCode.NODE_EXISTS) {
@@ -198,7 +222,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
         }
         try {
           // may create recursively
-          AccessResult res = doCreate(path, record, options);
+          AccessResult res = doCreate(_zkClient, path, record, options);
           result._pathCreated.addAll(res._pathCreated);
           RetCode rc = res._retCode;
           switch (rc) {
@@ -274,7 +298,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
           T newData = updater.update(null);
           RetCode rc;
           if (newData != null) {
-            AccessResult res = doCreate(path, newData, options);
+            AccessResult res = doCreate(_zkClient, path, newData, options);
             result._pathCreated.addAll(res._pathCreated);
             rc = res._retCode;
           } else {
@@ -324,6 +348,30 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
       }
     }
     return data;
+  }
+
+  /**
+   * Sync get method with custom serializer support
+   */
+  public Object get(String path, Stat stat, int options, ZkSerializer serializer) {
+    byte[] content = null;
+    try {
+      content = _nonZNRecordClient.readData(path, stat);
+      return serializer.deserialize(content);
+    } catch (ZkNoNodeException ex) {
+      if (AccessOption.isThrowExceptionIfNotExist(options)) {
+        throw ex;
+      }
+    }
+    return content;
+  }
+
+  /**
+   * Sync create method with custom serializer support
+   */
+  public boolean create(String path, Object data, int options, ZkSerializer serializer) {
+    AccessResult result = doCreate(_nonZNRecordClient, path, serializer.serialize(data), options);
+    return result._retCode == RetCode.OK;
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -111,10 +111,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    * @param zkAddress The zookeeper address
    */
   public ZkBaseDataAccessor(String zkAddress) {
-    _zkClient = SharedZkClientFactory.getInstance()
-        .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
-            new HelixZkClient.ZkClientConfig()
-                .setZkSerializer(new ZNRecordSerializer()));
+    this(zkAddress, new ZNRecordSerializer());
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -32,7 +32,6 @@ import org.I0Itec.zkclient.IZkChildListener;
 import org.I0Itec.zkclient.IZkDataListener;
 import org.I0Itec.zkclient.exception.ZkBadVersionException;
 import org.I0Itec.zkclient.exception.ZkException;
-import org.I0Itec.zkclient.exception.ZkMarshallingError;
 import org.I0Itec.zkclient.exception.ZkNoNodeException;
 import org.I0Itec.zkclient.exception.ZkNodeExistsException;
 import org.I0Itec.zkclient.serialize.ZkSerializer;
@@ -88,9 +87,8 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
   private static Logger LOG = LoggerFactory.getLogger(ZkBaseDataAccessor.class);
 
   private final HelixZkClient _zkClient;
-  //ZkClient that will be lazily instantiated if there are operations for non-ZNRecord data
-  private HelixZkClient _nonZNRecordZkClient = null;
 
+  //TODO: to be deprecated for client usage
   public ZkBaseDataAccessor(HelixZkClient zkClient) {
     if (zkClient == null) {
       throw new NullPointerException("zkclient is null");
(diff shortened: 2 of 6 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/GenericSerializer.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+/**
+ * The generic interface serialize/deserialize custom data type of T to/from byte[]
+ * @param <T> The custom data type
+ */
+public interface GenericSerializer<T> {
+  byte[] serialize(T data);
+  T deserilize(byte[] content);
+}
```
narendly: This serializer interface is not needed. Let us remove it.

It will have to be either PathBasedZkSerializer (Helix library) or ZkSerializer (ZK library). Let me confirm and update here in this PR.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -326,6 +327,21 @@ public T get(String path, Stat stat, int options) {
...
+
+  public <K> boolean create(String path, Stat stat, int options, GenericSerializer<K> serializer) {
+    //TODO
+    return false;
+  }
+
+  public <K> boolean set(String path, Stat stat, int options, GenericSerializer<K> serializer) {
+    //TODO
+    return false;
+  }
```
narendly: Since both PathBasedZkSerializer and ZkSerializer just use Java's generic Object and byte array, we do not need to introduce another generic type K. So it should look like the following:

`public Object get(String path, Stat stat, int options, ZkSerializer (or PathBasedZkSerializer) customSerializer);`

Also, let us split the PR so that we have separate PRs for each public method if you think that's easier?

narendly:
@i3wangyi Let us use ZkSerializer for this.

helix-core/src/main/java/org/apache/helix/manager/zk/GenericSerializer.java
i3wangyi: Sure, thanks for the confirmation!

narendly:
@i3wangyi Also, let's create a branch off the tip of master and work against that branch since this PR itself is not complete :)

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -101,6 +104,54 @@ public boolean create(String path, T record, int options) {
...
+      LOG.error("Invalid create mode. options: " + options);
+      result._retCode = RetCode.ERROR;
+      return result;
+    }
+
+    boolean retry;
+    do {
+      retry = false;
+      try {
+        _zkClient.create(path, record, ZooDefs.Ids.OPEN_ACL_UNSAFE, mode, zkSerializer);
```
narendly: This is different from what we discussed.

What we agreed on was to have another HelixZkClient that uses a "dummy" serializer, correct? And we add the serialization logic in `create(String path, T record, int options, ZkSerializer zkSerializer)`?

i3wangyi: The latest change has covered the usage of a dummy serializer zkClient, please help take a look and there're details I believe needs to be adjusted.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -117,16 +142,15 @@ public AccessResult doCreate(String path, T record, int options) {
...
         result._pathCreated.add(path);
-
         result._retCode = RetCode.OK;
         return result;
       } catch (ZkNoNodeException e) {
         // this will happen if parent node does not exist
         String parentPath = HelixUtil.getZkParentPath(path);
         try {
-          AccessResult res = doCreate(parentPath, null, AccessOption.PERSISTENT);
+          AccessResult res = doCreate(zkClient, parentPath, record, AccessOption.PERSISTENT);
```
narendly: Why was null changed to record?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -84,27 +88,48 @@ public AccessResult() {
...
   @Override
   public boolean create(String path, T record, int options) {
-    AccessResult result = doCreate(path, record, options);
+    AccessResult result = doCreate(_zkClient, path, record, options);
     return result._retCode == RetCode.OK;
   }
 
   /**
-   * sync create
+   * sync create. In order to support non-ZnRecord creation, the instance of {@link HelixZkClient} needs to be a parameter
```
narendly: Format?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -326,6 +350,30 @@ public T get(String path, Stat stat, int options) {
...
 
+  /**
+   * Sync get method with custom serializer support
+   */
+  public Object get(String path, Stat stat, int options, ZkSerializer serializer) {
+    byte[] content = null;
+    try {
+      content = _nonZNRecordClient.readData(path, stat);
+      return serializer.deserialize(content);
+    } catch (ZkNoNodeException ex) {
```
narendly: Will ZkMarshallingError (failed to deserialize) be caught in the users of get()?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -188,6 +191,40 @@ public void testSyncCreate() {
...
+        return bytes;
+      }
+    };
+    boolean success = accessor.create(path, recordData, AccessOption.PERSISTENT, zkSerializer);
+    Assert.assertTrue(success);
+    Object data = accessor.get(path, null, AccessOption.PERSISTENT, zkSerializer);
+    Assert.assertEquals(new String((byte[]) data), "DummyContent");
+    System.out.println("END " + testName + " at " + new Date(System.currentTimeMillis()));
+  }
+
```
narendly: Could we also add a case for get()?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: Fixed. Probably a mistake

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: It is the "formatted" comment.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
i3wangyi: The simple get test is already verified after creation, do you think is it still needed for adding one specific case for get?

narendly: In that case, I'm fine with not adding a test case for get.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
narendly: Reminder to remove catch clause for ZkMarshallingError.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -326,6 +350,39 @@ public T get(String path, Stat stat, int options) {
...
+   * Sync create method with custom serializer support
+   */
+  public boolean create(String path, Object data, int options, ZkSerializer serializer) {
+    try {
+      byte[] bytes = serializer.serialize(data);
+      AccessResult result = doCreate(_nonZNRecordClient, path, bytes, options);
+      return result._retCode == RetCode.OK;
+    } catch (ZkMarshallingError ex) {
+      LOG.error("Failed to serialize the data object", ex);
+      throw ex;
```
narendly: Reminder: remove catch clause ZkMarshallingError.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -84,27 +88,48 @@ public AccessResult() {
...
+          @Override
+          public Object deserialize(byte[] data) throws ZkMarshallingError {
+            return data;
+          }
+        }).build();
+  }
+
+  public HelixZkClient getZkClient() {
+    return _zkClient;
   }
```
narendly: Should we just make _zkClient a protected field instead of creating a public getter?

i3wangyi: ZkCacheBaseDataAccessor is not a child class of ZkBaseDataAccessor, so making it protected won't help ZkCacheBaseDataAccessor get access to the _zkClient

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: done

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: I've used a simple boolean parameter to avoid using the public get method in the latest version.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -92,19 +98,42 @@ public ZkBaseDataAccessor(HelixZkClient zkClient) {
     _zkClient = zkClient;
   }
 
+  private synchronized HelixZkClient getNonZNRecordClient() {
```
narendly: Are we doing a lazy initialization? If so, please document this clearly in the code why lazy initialization is required.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -92,19 +98,42 @@ public ZkBaseDataAccessor(HelixZkClient zkClient) {
...
+    if (_nonZNRecordClient == null) {
```
narendly: I think the use of synchronization primitives could be changed to allow for more parallelism?

Instead of synchronizing on the whole method, could we do:

```
private HelixZkClient getNonZNRecordZkClient() {
  if (_nonZnRecordClient == null) {
    synchronized (`_nonZnRecordZkClient`) {
      // create...
  else {
    return `_nonZnRecordZkClient`;
```

Synchronizing the whole method might significantly reduce parallelism. We only need to synchronize when we need to modify the underlying object, which is `_nonZnRecordZkClient`;

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -92,19 +98,42 @@ public ZkBaseDataAccessor(HelixZkClient zkClient) {
...
+      _nonZNRecordClient = new ZkClient.Builder().setConnection(
```
narendly: Let's use make it clear in the variable name that this is a ZK client.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -92,19 +98,42 @@ public ZkBaseDataAccessor(HelixZkClient zkClient) {
...
+    AccessResult result = doCreate(path, record, options, true);
     return result._retCode == RetCode.OK;
   }
 
   /**
-   * sync create
+   * sync create. In order to support non-ZnRecord creation, the instance of {@link HelixZkClient} needs to be a parameter
    */
-  public AccessResult doCreate(String path, T record, int options) {
+  public AccessResult doCreate(String path, Object record, int options, boolean isZnRecord) {
```
narendly: Could we possibly tweak the method signature like it was before instead of a boolean flag? I think

`doCreate(HelixZkClient zkClient, Object record, int options);` was a better structure.

And getNonZnRecordZkClient() would be called in your get(), set(), and create().

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -166,14 +195,30 @@ public boolean set(String path, T record, int options) {
...
+  public boolean set(String path, Object record, int options, int expectVersion, ZkSerializer zkSerializer) {
+    AccessResult result = doSet(path, zkSerializer.serialize(record), expectVersion, options, false);
     return result._retCode == RetCode.OK;
   }
 
   /**
    * sync set
    */
-  public AccessResult doSet(String path, T record, int expectVersion, int options) {
+  public AccessResult doSet(String path, Object record, int expectVersion, int options, boolean isZnRecord) {
```
narendly: public AccessResult doSet(HelixZkClient zkClient, String path, Object record, int expectVersion, int options) {

would be a better structure because it allows for more flexibility (vs 2 options when a boolean is used).

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: However, it will go back to the previous problem when ZKCachedDataAccesor wants to use `doCreate` method directly and it needs to get the instance of ZKClient.

narendly: CachedBaseDataAccessor could just pass in `_zkclient`, correct?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: done

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: done

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: done.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: ZkCachedBaseDataAccessor doesn't have the `_zkClient` instance, ZkCachedBaseDataAccessor is not a child class of zkBaseDataAccessor somehow. Going this route will create the public method of getting zkClient.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -92,19 +99,51 @@ public ZkBaseDataAccessor(HelixZkClient zkClient) {
     _zkClient = zkClient;
   }
 
+  /**
+   * Lazy initialization is used to reduce the cost of unnecessary opened zkConnections
+   * @return singleton instance of non-ZnRecord ZkClient
+   */
+  private HelixZkClient getNonZNRecordZkClient() {
+    if (_nonZNRecordZkClient == null) {
+      synchronized (ZkBaseDataAccessor.class) {
```
narendly: Is there a reason we're synchronizing on this? I think when we synchronize on the class, we usually do `synchronized (this)`?

My previous comment is suggesting that we synchronize on the nonZNRecordZkClient object.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: I take the comment back, sorry I just noticed the cacheAccessor also has its own instance of zkClient.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -92,19 +99,51 @@ public ZkBaseDataAccessor(HelixZkClient zkClient) {
...
+        _nonZNRecordZkClient = new ZkClient.Builder().setConnection(
```
narendly: We may need to do a null check again here to avoid multiple runs of the creation logic by other threads that were blocking.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: synchronized on the nullable object could lead to NPE, I like the synchronized(this) idea. Sounds good!

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: Good catch!

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: The method signatures get fixed in the latest commit.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
narendly: Great. You're correct - let's use synchronized(this) since the ZkClient could be null in this case.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -84,6 +88,9 @@ public AccessResult() {
   private static Logger LOG = LoggerFactory.getLogger(ZkBaseDataAccessor.class);
 
   private final HelixZkClient _zkClient;
+  // onDemand zkClient for read/write non-znRecord data path;
```
narendly: read/write "to" non-ZnRecord data

-> remove "path" because there's no such thing as ZnRecord path or non-ZnRecord path.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -84,6 +88,9 @@ public AccessResult() {
...
+  // minimal change to support custom serializer
```
narendly: This comment is confusing and not necessary. Let's remove this. Or just say

"ZkClient that will be lazily instantiated if there are operations for non-ZNRecord data"

i3wangyi: updated.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: fixed

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: After running the mvn test again, I found another issue with this approach: 
The zkCacheBaseDataAccessor's zkClient == null in some cases (depending on which constructor it is using). So some tests start to fail because of NPE, I feel directly using `_zkclient` is risky in zkCacheBaseDataAccessor. A public get method is safer in that case.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -92,19 +98,52 @@ public ZkBaseDataAccessor(HelixZkClient zkClient) {
     _zkClient = zkClient;
   }
 
+  /**
+   * Lazy initialization is used to reduce the cost of unnecessary opened zkConnections
+   * @return singleton instance of non-ZnRecord ZkClient
```
narendly: Here's another idea. If keeping many active ZkConnections is expensive (and how? memory intensive?), could we close the connection after each operation is done, and connect the ZkClient right before the start of an operation?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -50,11 +66,6 @@ public void afterMethod() {
     }
   }
 
-  @AfterClass
-  public void after() {
-    int a =1;
-  }
```
narendly: Could we make sure we have an AfterClass where we close all outstanding/existing ZkClient connections? This is very important for the stability of our tests.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -34,11 +37,24 @@
 import org.apache.helix.manager.zk.ZkBaseDataAccessor.RetCode;
 import org.apache.zookeeper.data.Stat;
 import org.testng.Assert;
-import org.testng.annotations.AfterClass;
 import org.testng.annotations.AfterMethod;
 import org.testng.annotations.Test;
 
 public class TestZkBaseDataAccessor extends ZkUnitTestBase {
+  private static final ZkSerializer CUSTOM_ZK_SERIALIZER = new ZkSerializer() {
```
narendly: Add a quick comment saying what this custom serializer does?

"A serializer that takes a String and converts it to a byte array."

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -92,19 +98,52 @@ public ZkBaseDataAccessor(HelixZkClient zkClient) {
...
-    AccessResult result = doCreate(path, record, options);
+    AccessResult result = doCreate(_zkClient, path, record, options);
     return result._retCode == RetCode.OK;
   }
 
   /**
-   * sync create
+   * sync create. In order to support non-ZnRecord creation, the instance of {@link HelixZkClient} needs to be a parameter
    */
-  public AccessResult doCreate(String path, T record, int options) {
```
narendly: We need to keep the old public API method signature. Otherwise, this diff will not be backward-compatible. Could you refactor so that we still have all the old public API methods that has the same behavior?

narendly: For example, here you could do:

```
public AccessResult doCreate(String path, T record, int options) {
    doCreate(_zkClient, path, record, options);
}
```

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -166,14 +204,29 @@ public boolean set(String path, T record, int options) {
...
+   */
+  public boolean set(String path, Object record, int options, int expectVersion, ZkSerializer zkSerializer) {
+    AccessResult result = doSet(getNonZNRecordZkClient(), path, zkSerializer.serialize(record), expectVersion, options);
     return result._retCode == RetCode.OK;
   }
 
   /**
    * sync set
    */
-  public AccessResult doSet(String path, T record, int expectVersion, int options) {
```
narendly: We need to keep the old public API method signature. Otherwise, this diff will not be backward-compatible. Could you refactor so that we still have all the old public API methods that has the same behavior?

See my comment on doCreate. If we do this, we won't have to change the code in other places like ZkHelixCacheDataAccessor.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java
```diff
@@ -232,7 +232,7 @@ public boolean create(String path, T data, int options) {
       try {
         cache.lockWrite();
         ZkBaseDataAccessor<T>.AccessResult result =
-            _baseAccessor.doCreate(serverPath, data, options);
+            _baseAccessor.doCreate(_zkclient, serverPath, data, options);
```
narendly: If we keep the old public API instead of removing, we shouldn't need to change this at all. In other words, we shouldn't see any diffs in CacheBaseDataAccessor.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java
```diff
@@ -263,7 +263,7 @@ public boolean set(String path, T data, int expectVersion, int options) {
       if (cache != null) {
         cache.lockWrite();
         ZkBaseDataAccessor<T>.AccessResult result =
-            _baseAccessor.doSet(serverPath, data, expectVersion, options);
+            _baseAccessor.doSet(_zkclient, serverPath, data, expectVersion, options);
```
narendly: If we keep the old public API instead of removing, we shouldn't need to change this at all. In other words, we shouldn't see any diffs in CacheBaseDataAccessor.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -146,7 +157,7 @@ public void testSyncDoSet() {
     ZNRecord record = new ZNRecord("submsg_0");
     ZkBaseDataAccessor<ZNRecord> accessor = new ZkBaseDataAccessor<ZNRecord>(_gZkClient);
 
-    AccessResult result = accessor.doSet(path, record, -1, AccessOption.PERSISTENT);
+    AccessResult result = accessor.doSet(_gZkClient, path, record, -1, AccessOption.PERSISTENT);
```
narendly: If we keep the old public API instead of removing, we shouldn't need to change this at all. In other words, we shouldn't see any diffs here.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
narendly: Thanks for catching it. In that case, I think using a getter is okay.

narendly:
Also @dasahcc had an idea of pushing this change to the ZkClient level. Let's discuss again offline.

i3wangyi:
> Also @dasahcc had an idea of pushing this change to the ZkClient level. Let's discuss again offline.

Don't come back saying pushing the change into the lower layer in ZKClient because it was one of my previous implementations last week (while you said it's not a design we agreed on). For the sake of minimal changes inside ZkClient we have this 2nd non-znRecord zkClient idea

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: I think the option has been discussed last week and I have reasons against it: worse case it will open/close many zk connections and each round of operation is I/O expensive. Performance is a problem and the cost of open/close zk connections. Compared with that, the one single additional zkConnection by lazy initialization at least only add one additional zkConnection and the performance is better.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: This is a really good idea, thanks!

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
i3wangyi: done.

narendly:
> > Also @dasahcc had an idea of pushing this change to the ZkClient level. Let's discuss again offline.
> 
> Don't come back saying pushing the change into the lower layer in ZKClient because it was one of my previous implementations last week (while you said it's not a design we agreed on). For the sake of minimal changes inside ZkClient we have this 2nd non-znRecord zkClient idea

I will explain. @dasahcc 's idea is a little different from what you had here before, which was to have a ZkClient take in a serializer, and do a setSerializer() for each operation with a custom serializer. His idea involves creating an overloaded serialize() and deserialize() method in the ZkClient so you don't have to change the serializer every time.

Either way works, but we could discuss pros and cons of each method.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -85,13 +88,20 @@ public AccessResult() {
...
 
+  //TODO: to be deprecated for client usage
   public ZkBaseDataAccessor(HelixZkClient zkClient) {
     if (zkClient == null) {
       throw new NullPointerException("zkclient is null");
     }
     _zkClient = zkClient;
   }
 
+  public ZkBaseDataAccessor(String zkAddress, ZkSerializer zkSerializer) {
```
narendly: Should we provide` public ZkBaseDataAccessor(String zkAddress)` as well? This would be the default constructor using ZnRecordSerializer when a custom serializer is not needed.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -85,13 +88,20 @@ public AccessResult() {
...
+    _zkClient = new ZkClient.Builder()
+        .setConnection(new ZkConnection(zkAddress, ZkClient.DEFAULT_SESSION_TIMEOUT))
+        .setZkSerializer(zkSerializer).build();
```
narendly: We talked about using SharedZkClientFactory here, right?

i3wangyi: Sure, I will check out the usage. It's one of the missing details with the current commit, I quickly draft it to simply show-case the new idea to see how the overall structure looks!

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -188,6 +199,33 @@ public void testSyncCreate() {
     System.out.println("END " + testName + " at " + new Date(System.currentTimeMillis()));
   }
 
+  @Test
+  public void testSyncCreateWithCustomSerializer() {
```
lei-xia: How did you verify the dataAccessor actually uses the serializer given by the caller, not the default one?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: done.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
i3wangyi: The updated test takes a list serializer/deserilaizer as the input parameter, it's different from default one. Do you think we need to test the constructor if it really overrides the zkClient with custom serializer?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -188,6 +194,51 @@ public void testSyncCreate() {
...
```
narendly: We need to test the regular constructor without a custom serializer.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
narendly: Working with @i3wangyi to add more test cases to test this.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -1114,4 +1136,12 @@ public void unsubscribeChildChanges(String path, IZkChildListener childListener)
   public void reset() {
     // Nothing to do
   }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void close() {
+    _zkClient.close();
```
narendly: Could you please add a null check and isClosed() check (if available)?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -188,6 +212,72 @@ public void testSyncCreate() {
...
+  @Test
+  public void testDefaultAccessorCreateCustomData() {
+    String className = TestHelper.getTestClassName();
+    String methodName = TestHelper.getTestMethodName();
+    String testName = className + "_" + methodName;
+    System.out.println("START " + testName + " at " + new Date(System.currentTimeMillis()));
+
+    String path = String.format("/%s/%s", _rootPath, "msg_0");
+
+    ZkBaseDataAccessor defaultAccessor = new ZkBaseDataAccessor(ZK_ADDR);
```
narendly: Please close this defaultAccessor at the end of the test to prevent a ZkClient leak.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -188,6 +212,72 @@ public void testSyncCreate() {
...
+  @Test
+  public void testCustomAccessorCreateZnRecord() {
+    String className = TestHelper.getTestClassName();
+    String methodName = TestHelper.getTestMethodName();
+    String testName = className + "_" + methodName;
+    System.out.println("START " + testName + " at " + new Date(System.currentTimeMillis()));
+
+    String path = String.format("/%s/%s", _rootPath, "msg_0");
+
+    ZkBaseDataAccessor customDataAccessor = new ZkBaseDataAccessor(ZK_ADDR, LIST_SERIALIZER);
```
narendly: Please close this accessor at the end of the test to prevent a ZkClient leak.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -188,6 +212,72 @@ public void testSyncCreate() {
...
+  @Test
+  public void testSyncCreateWithCustomSerializer() {
+    String className = TestHelper.getTestClassName();
+    String methodName = TestHelper.getTestMethodName();
+    String testName = className + "_" + methodName;
+    System.out.println("START " + testName + " at " + new Date(System.currentTimeMillis()));
+
+    String path = String.format("/%s/%s", _rootPath, "msg_0");
+
+    ZkBaseDataAccessor<List<Integer>> accessor = new ZkBaseDataAccessor<>(ZK_ADDR, LIST_SERIALIZER);
```
narendly: Please close this accessor at the end of the test to prevent a ZkClient leak.

helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
```diff
@@ -257,4 +257,9 @@
    * reset the cache if any, when session expiry happens
    */
   void reset();
+
+  /**
+   * Close the opened zkConnection
```
narendly: Please remove the mention of ZK-related things in this interface. BaseDataAccessor is a generic interface - it shouldn't assume the use of ZK.

Something like "Close the connection to the metadata store" would be more appropriate.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: done.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
i3wangyi: done.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
i3wangyi: done.

helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
i3wangyi: done.

i3wangyi:
@narendly mvn test founds out some issues (after adding the close method), I've addressed them (very minor). Please take a look. 

The PR is ready to be merged into Master, approved by @narendly 

commit message:
Update zkBaseDataAccessor with custom serializer support

Issue #518, opened 2019-10-22
Stabilize the WAGED rebalancer rebalance result


jiajunwang:
It is noticed during the unit test that the WAGED rebalancer triggers movements when there is an only minor change in the cluster. For example, node down or resource ideal state change (the mapping field change which should be ignored).

These oversensitive rebalance make the assignment unstable. We need to fix the rebalancer so it only computes a new assignment when it is really necessary.

PR #535, opened 2019-10-25
Adjust the replica rebalance calculating ordering method to avoid a static order.

jiajunwang:
### Issues

#518 

### Description

Adjust the replica rebalance calculating ordering to avoid static order.

The problem of a static order is that the same set of replicas will always be the ones that are moved or state transited during the rebalance.
This randomize won't change the algorithm's performance. But it will help the Helix to eliminate very unstable partitions.

### Tests

NA

[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 1045, Failures: 2, Errors: 0, Skipped: 3
[INFO] 
[INFO] BUILD FAILURE

Rerun
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 24.373 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS

### Commits

### Documentation

NA

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
@@ -24,6 +24,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
@@ -134,7 +135,6 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
         .collect(Collectors.toList());
   }
 
-  // TODO investigate better ways to sort replicas. One option is sorting based on the creation time.
   private List<AssignableReplica> getOrderedAssignableReplica(ClusterModel clusterModel) {
     Map<String, Set<AssignableReplica>> replicasByResource = clusterModel.getAssignableReplicaMap();
     List<AssignableReplica> orderedAssignableReplicas =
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -159,12 +159,19 @@ private float getAssignmentNormalizedScore(AssignableNode node, AssignableReplic
...
-            }
+            // If state prioritizes are the same, try to randomize the replicas order. Otherwise,
+            // the same replicas might always be moved in the rebalance. This is because their
+            // rebalance calculating will always happen at the critical moment while the cluster is
+            // almost full.
+            //
+            // Note that to ensure the algorithm is deterministic with the same inputs, do not use
+            // Random functions here. Use hashcode based on the cluster topology information to get
+            // a controlled randomized order is good enough.
+            long replicaHash1 = Objects
```
dasahcc: Is this hashing algorithm deterministic?

jiajunwang: It is java hashing based. Just a wrapper.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -162,12 +162,19 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
           int statePriority1 = replica1.getStatePriority();
           int statePriority2 = replica2.getStatePriority();
           if (statePriority1 == statePriority2) {
-            // If state prioritizes are the same, compare the names.
-            if (resourceName1.equals(resourceName2)) {
-              return replica1.getPartitionName().compareTo(replica2.getPartitionName());
-            } else {
-              return resourceName1.compareTo(resourceName2);
-            }
+            // If state prioritizes are the same, try to randomize the replicas order. Otherwise,
```
narendly: prioritizes -> priorities

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -162,12 +162,19 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+            // the same replicas might always be moved in the rebalance. This is because their
+            // rebalance calculating will always happen at the critical moment while the cluster is
+            // almost full.
```
narendly: Why is the same order a problem? Do we not want it to be deterministic?

jiajunwang: The problem of fixed order is "the same replicas might always be moved in the rebalance", as mentioned in the comments.

If you read the following sentence in the comments, the order is still deterministic given the input is the same. There is no randomness.

narendly: What's confusing me is that usually we want a static order in rebalance so we know what to expect, and that way it's easier for us to debug, track, and test. That has been the case with CrushED rebalancer where we sort all resources and partitions by lexical order.

So the fact that you're avoiding the static order in which replicas are given to the rebalancer is counter-intuitive. So I'd like to understand why it's different in the WAGED rebalancer setting. Also, the description claims that the static ordering makes the rebalance unstable. How so?

jiajunwang: I see.
The thing is that we still need a stable order given the TOPOLOGY is the same. But, we don't want a fixed order for the WAGED rebalancer.
Because of the greedy algorithm, there always be an unstable window while we try to place the replica one by one. For example, if you have 10 replicas to be assigned, usually, the rebalancer always tries to move replica 9,10. If we have a fixed order, the same replica will be moved again and again. And due to the new placement tend to be unstable, this double the possibility that the replicas being moved. So as a result, these replicas will almost be certain to be moved again in the next round of rebalance. That, as I noted in the comments, makes the assignment unstable.
The details are hard to note down in the comments. I plan to explain later in some blogs in detail.

jiajunwang:

Issue #539, opened 2019-10-25
Add instance utilization metrics


pkuwm:
We need to monitor instance utilization in purpose of understanding why rebalance is triggered and usage of the instance's resource. 

CapacityGauge  | Gauge of the capacity for each CAPACITY key. 
UsageGauge  | Gauge of the real time partition weight reservation for each CAPACITY key on the participant.
WeightGauge | Gauge of the average partition weight for each CAPACITY keys.Note that per-partition weight metrics would be more accurate. But the number of metrics would be too large to support.
MaxCapacityUsageGauge

PR #548, opened 2019-10-29
Add max capacity usage metric for instance monitor.

pkuwm:
### Issues

- [x] My PR addresses the following Helix issues and references them in the PR description:

Implements one part of #539 : MaxCapacityUsageGauge

### Description

- [x] Here are some details about my PR, including screenshots of any UI changes:
We need to monitor instance utilization in purpose of understanding what the max instance capacity usage is and knowing the status of the instance.

Changes:
1. Change instance monitor to extend dynamic metric.
2. Add APIs for get/update MaxCapacityUsage
3. Add an API in cluster status monitor to update max capacity usage.
4. Add unit tests.

### Tests

- [x] The following tests are written for this issue:
TestIntanceMonitor
TestClusterStatusMonitor.testUpdateMaxCapacityUsage

- [x] The following is the result of the "mvn test" command on the appropriate module:
mvn test in helix-core

[ERROR] Failures:
[ERROR]   TestDisable.testDisableNodeCustomIS:97->Object.hashCode:-2 » ThreadTimeout Met...
[ERROR]   TestJobQueueCleanUp.testJobQueueAutoCleanUp » ThreadTimeout Method org.testng....
[ERROR]   TestTaskPerformanceMetrics.testTaskPerformanceMetrics:118 expected:<true> but was:<false>
[ERROR]   TestGetLastScheduledTaskExecInfo.testGetLastScheduledTaskExecInfo:73 expected:<COMPLETED> but was:<RUNNING>
[ERROR]   TestClusterVerifier.testResourceSubset:211 » ThreadTimeout Method org.testng.i...
[INFO]
[ERROR] Tests run: 1047, Failures: 5, Errors: 0, Skipped: 4
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:19 h
[INFO] Finished at: 2019-10-28T21:52:22-07:00
[INFO] ------------------------------------------------------------------------

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 8.297 s - in org.apache.helix.task.TestGetLastScheduledTaskExecInfo
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0

[...]

PR #557, opened 2019-10-30
Add instance capacity gauge

pkuwm:
### Issues

Implements one part of #539 : InstanceCapacityGauge

### Description

We need to monitor instance utilization in purpose of understanding what the instance capacity is.

Changes:
1. Change instance monitor to update capacity.
2. Change typo of API updateAttributeInfo in dynamicMetric
3. Combine max usage and instance capacity update into one api in cluster status monitor.
4. Add unit tests.

### Tests

TestClusterStatusMonitor.testUpdateInstanceCapacity

[INFO] Results:
[INFO]
[ERROR] Failures:
[INFO]
[ERROR] Tests run: 1061, Failures: 1, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:11 h
[INFO] Finished at: 2020-01-14T18:25:06-08:00

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 10.123 s - in org.apache.helix.integration.task.TestJobQueueCleanUp
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  14.906 s
[INFO] Finished at: 2020-01-14T19:08:42-08:00

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
@@ -21,7 +21,6 @@ package org.apache.helix.controller.stages;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -249,7 +248,7 @@ public class CurrentStateComputationStage extends AbstractBaseStage {
         // Only use the resources in ideal states to parse all replicas.
         Map<String, IdealState> idealStateMap = dataProvider.getIdealStates();
         Map<String, Resource> resourceToMonitorMap = resourceMap.entrySet().stream()
-            .filter(resourceName -> idealStateMap.containsKey(resourceName))
+            .filter(idealStateMap::containsKey)
             .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
 
         Map<String, ResourceAssignment> currentStateAssignment =
@@ -257,16 +256,16 @@ public class CurrentStateComputationStage extends AbstractBaseStage {
         ClusterModel clusterModel = ClusterModelProvider.generateClusterModelFromExistingAssignment(
             dataProvider, resourceToMonitorMap, currentStateAssignment);
 
-        Map<String, Double> maxUsageMap = new HashMap<>();
         for (AssignableNode node : clusterModel.getAssignableNodes().values()) {
           String instanceName = node.getInstanceName();
+          // There is no new usage adding to this node, so an empty map is passed in.
           double usage = node.getProjectedHighestUtilization(Collections.emptyMap());
-          maxUsageMap.put(instanceName, usage);
+          clusterStatusMonitor
+              .updateInstanceCapacityStatus(instanceName, usage, node.getMaxCapacity());
         }
-
-        clusterStatusMonitor.updateInstanceMaxUsage(maxUsageMap);
       } catch (Exception ex) {
-        LOG.error("Failed to report instance capacity metrics.", ex);
+        LOG.error("Failed to report instance capacity metrics. Exception message: {}",
+            ex.getMessage());
       }
 
       return null;
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
@@ -367,24 +367,25 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
   }
 
   /**
-   * Update max capacity usage for per instance. Before calling this API, we assume the instance
-   * monitors are already registered in ReadClusterDataStage. If the monitor is not registered, this
-   * max usage update will fail.
+   * Updates instance capacity status for per instance, including max usage and capacity of each
+   * capacity key. Before calling this API, we assume the instance monitors are already registered
+   * in ReadClusterDataStage. If the monitor is not registered, this instance capacity status update
+   * will fail.
    *
-   * @param maxUsageMap a map of max capacity usage, {instance: maxCapacityUsage}
+   * @param instanceName This instance name
+   * @param maxUsage Max capacity usage of this instance
+   * @param capacityMap A map of this instance capacity, {capacity key: capacity value}
    */
-  public void updateInstanceMaxUsage(Map<String, Double> maxUsageMap) {
-    synchronized (_instanceMonitorMap) {
-      for (Map.Entry<String, Double> entry : maxUsageMap.entrySet()) {
-        InstanceMonitor monitor = _instanceMonitorMap.get(entry.getKey());
-        if (monitor == null) {
-          LOG.warn("Failed to update max usage because instance monitor is not found, instance: {}.",
-              entry.getKey());
-          continue;
-        }
-        monitor.updateMaxCapacityUsage(entry.getValue());
-      }
+  public void updateInstanceCapacityStatus(String instanceName, double maxUsage,
+      Map<String, Integer> capacityMap) {
+    InstanceMonitor monitor = _instanceMonitorMap.get(instanceName);
+    if (monitor == null) {
+      LOG.warn("Failed to update instance capacity status because instance monitor is not found, "
+          + "instance: {}.", instanceName);
+      return;
     }
+    monitor.updateMaxCapacityUsage(maxUsage);
+    monitor.updateCapacity(capacityMap);
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/InstanceMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/InstanceMonitor.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import javax.management.JMException;
 import javax.management.ObjectName;
 
@@ -41,7 +42,7 @@ public class InstanceMonitor extends DynamicMBeanProvider {
   /**
    * Metric names for instance capacity.
    */
-  public enum InstanceMonitorMetrics {
+  public enum InstanceMonitorMetric {
     // TODO: change the metric names with Counter and Gauge suffix and deprecate old names.
     TOTAL_MESSAGE_RECEIVED_COUNTER("TotalMessageReceived"),
     ENABLED_STATUS_GAUGE("Enabled"),
@@ -49,9 +50,9 @@ public class InstanceMonitor extends DynamicMBeanProvider {
     DISABLED_PARTITIONS_GAUGE("DisabledPartitions"),
     MAX_CAPACITY_USAGE_GAUGE("MaxCapacityUsageGauge");
 
-    private String metricName;
+    private final String metricName;
 
-    InstanceMonitorMetrics(String name) {
+    InstanceMonitorMetric(String name) {
       metricName = name;
     }
 
@@ -75,6 +76,9 @@ public class InstanceMonitor extends DynamicMBeanProvider {
   private SimpleDynamicMetric<Long> _onlineStatusGauge;
   private SimpleDynamicMetric<Double> _maxCapacityUsageGauge;
 
+  // A map of dynamic capacity Gauges. The map's keys could change.
+  private final Map<String, SimpleDynamicMetric<Long>> _dynamicCapacityMetricsMap;
+
   /**
    * Initialize the bean
    * @param clusterName the cluster to monitor
@@ -85,24 +89,39 @@ public class InstanceMonitor extends DynamicMBeanProvider {
     _participantName = participantName;
     _tags = ImmutableList.of(ClusterStatusMonitor.DEFAULT_TAG);
     _initObjectName = objectName;
+    _dynamicCapacityMetricsMap = new ConcurrentHashMap<>();
 
     createMetrics();
   }
 
   private void createMetrics() {
     _totalMessagedReceivedCounter = new SimpleDynamicMetric<>(
-        InstanceMonitorMetrics.TOTAL_MESSAGE_RECEIVED_COUNTER.metricName(), 0L);
+        InstanceMonitorMetric.TOTAL_MESSAGE_RECEIVED_COUNTER.metricName(), 0L);
 
     _disabledPartitionsGauge =
-        new SimpleDynamicMetric<>(InstanceMonitorMetrics.DISABLED_PARTITIONS_GAUGE.metricName(),
+        new SimpleDynamicMetric<>(InstanceMonitorMetric.DISABLED_PARTITIONS_GAUGE.metricName(),
             0L);
     _enabledStatusGauge =
-        new SimpleDynamicMetric<>(InstanceMonitorMetrics.ENABLED_STATUS_GAUGE.metricName(), 0L);
+        new SimpleDynamicMetric<>(InstanceMonitorMetric.ENABLED_STATUS_GAUGE.metricName(), 0L);
     _onlineStatusGauge =
-        new SimpleDynamicMetric<>(InstanceMonitorMetrics.ONLINE_STATUS_GAUGE.metricName(), 0L);
+        new SimpleDynamicMetric<>(InstanceMonitorMetric.ONLINE_STATUS_GAUGE.metricName(), 0L);
     _maxCapacityUsageGauge =
-        new SimpleDynamicMetric<>(InstanceMonitorMetrics.MAX_CAPACITY_USAGE_GAUGE.metricName(),
+        new SimpleDynamicMetric<>(InstanceMonitorMetric.MAX_CAPACITY_USAGE_GAUGE.metricName(),
             0.0d);
+  }
+
+  private List<DynamicMetric<?, ?>> buildAttributeList() {
+    List<DynamicMetric<?, ?>> attributeList = Lists.newArrayList(
+        _totalMessagedReceivedCounter,
+        _disabledPartitionsGauge,
+        _enabledStatusGauge,
+        _onlineStatusGauge,
+        _maxCapacityUsageGauge
+    );
+
+    attributeList.addAll(_dynamicCapacityMetricsMap.values());
+
+    return attributeList;
   }
 
   @Override
@@ -183,7 +202,7 @@ public class InstanceMonitor extends DynamicMBeanProvider {
   }
 
   /**
-   * Update max capacity usage for this instance.
+   * Updates max capacity usage for this instance.
    * @param maxUsage max capacity usage of this instance
    */
   public synchronized void updateMaxCapacityUsage(double maxUsage) {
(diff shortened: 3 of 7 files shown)


pkuwm:
Will rebase the PR once #548 is merged.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -367,22 +367,28 @@ public void increaseMessageReceived(List<Message> messages) {
...
+   * capacity key. Before calling this API, we assume the instance monitors are already registered
+   * in ReadClusterDataStage. If the monitor is not registered, this instance capacity status update
+   * will fail. Keys for maxUsageMap and capacityMap should be the same.
    *
-   * @param maxUsageMap a map of max capacity usage, {instance: maxCapacityUsage}
+   * @param maxUsageMap A map of max capacity usage, {instance: maxCapacityUsage}
+   * @param capacityMap A map of instance capacity, {instance: {capacity key: capacity value}}
    */
-  public void updateInstanceMaxUsage(Map<String, Double> maxUsageMap) {
+  public void updateInstanceCapacityStatus(Map<String, Double> maxUsageMap,
```
pkuwm: I combine max usage and instance capacity update into one api, in the purpose of to avoid one more for loop and synchronized lock for instance capacity update. The downside is, we assume keys (instance names) are the same. If you think we should separate the apis, please comment.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
```diff
@@ -103,7 +103,7 @@ protected synchronized boolean doRegister(Collection<DynamicMetric<?, ?>> dynami
    * @param description description of the MBean
    * @param dynamicMetrics the DynamicMetrics
    */
-  private void updateAttributtInfos(Collection<DynamicMetric<?, ?>> dynamicMetrics,
+  public void updateAttributeInfo(Collection<DynamicMetric<?, ?>> dynamicMetrics,
```
pkuwm: This might need to be a separate PR.. Let me know if it is neccessary.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/InstanceMonitor.java
```diff
@@ -199,6 +203,39 @@ protected synchronized double getMaxCapacityUsageGauge() {
...
+
+      // Update MBean's all attributes.
+      List<DynamicMetric<?, ?>> attributeList = Lists.newArrayList(
+          _totalMessagedReceivedCounter,
+          _disabledPartitionsGauge,
+          _enabledStatusGauge,
+          _onlineStatusGauge,
+          _maxCapacityUsageGauge
+      );
+      attributeList.addAll(_capacityMetricsMap.values());
```
jiajunwang: Recreating the whole list is expensive, can we have a map to track all metrics? A map would be easier for you to maintain.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
jiajunwang: 1. protected?
2. Thanks for fixing the typo, let use "updateAttributtesInfo"?
3. You will need to add synchronized if it is called somewhere else outside this class. Actually we need to add this to all the methods that are reading or updating _attributeMap and _mBeanInfo.
And please don't lock globally, a separate lock for _attributeMap and _mBeanInfo would be good enough.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -367,22 +367,28 @@ public void increaseMessageReceived(List<Message> messages) {
...
+      Map<String, Map<String, Integer>> capacityMap) {
     synchronized (_instanceMonitorMap) {
       for (Map.Entry<String, Double> entry : maxUsageMap.entrySet()) {
```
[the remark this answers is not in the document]
pkuwm: Right. I've put comments for this. 2 for loops increase a bit more time. Will change it.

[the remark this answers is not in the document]
pkuwm: If we only see where the api is called here, the two maps' keys are the same. But we can't guarantee future calls would have the same keys. Why I did this is to try to combine 2 updates into one api and reduce one for loop. But if we need 2 for loops, I would like 2 different APIs separately.

pkuwm:
@jiajunwang PR updated. Thanks for reviewing!

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
```diff
@@ -186,7 +186,7 @@ public AttributeList getAttributes(String[] attributes) {
   }
 
   @Override
-  public MBeanInfo getMBeanInfo() {
+  public synchronized MBeanInfo getMBeanInfo() {
```
jiajunwang: As we discussed, this sync won't work. Let's just lock the attributes list in the direct access method.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -242,16 +248,18 @@ private void reportInstanceCapacityMetrics(ClusterStatusMonitor clusterStatusMon
...
         for (AssignableNode node : clusterModel.getAssignableNodes().values()) {
           String instanceName = node.getInstanceName();
           double usage = node.getProjectedHighestUtilization(Collections.emptyMap());
-          maxUsageMap.put(instanceName, usage);
+          clusterStatusMonitor
+              .updateInstanceCapacityStatus(instanceName, usage, node.getMaxCapacity());
         }
-
-        clusterStatusMonitor.updateInstanceMaxUsage(maxUsageMap);
+      } catch (HelixException ex) {
```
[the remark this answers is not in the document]
pkuwm: I actually wanted to make it more specific. Exception to me, is too generic, which I try to avoid.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -242,16 +248,18 @@ private void reportInstanceCapacityMetrics(ClusterStatusMonitor clusterStatusMon
         ClusterModel clusterModel = ClusterModelProvider.generateClusterModelFromCurrentState(
             dataProvider, resourceMap, currentStateAssignment);
 
-        Map<String, Double> maxUsageMap = new HashMap<>();
         for (AssignableNode node : clusterModel.getAssignableNodes().values()) {
           String instanceName = node.getInstanceName();
           double usage = node.getProjectedHighestUtilization(Collections.emptyMap());
```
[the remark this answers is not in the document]
pkuwm: Sure.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
pkuwm: How about locking a separate object for both _MBeanInfo and _attributeMap? Once one of them is accessed, we lock both. In this case, we could make sure MBeanInfo and attributes are bound together. 
```
private final Object lockObject = new Object();

synchronized (lockObject) {
}

File: helix-core/src/main/java/org/apache/helix/monitoring/mbeans/InstanceMonitor.java [hunk 8]
```diff
@@ -67,14 +74,12 @@ public String metricName() {
...
-  // Counters
-  private SimpleDynamicMetric<Long> _totalMessagedReceivedCounter;
-
-  // Gauges
-  private SimpleDynamicMetric<Long> _enabledStatusGauge;
-  private SimpleDynamicMetric<Long> _disabledPartitionsGauge;
-  private SimpleDynamicMetric<Long> _onlineStatusGauge;
-  private SimpleDynamicMetric<Double> _maxCapacityUsageGauge;
+  // Capacity Gauges
+  private final Map<String, SimpleDynamicMetric<Long>> _capacityMetricsMap;
```
jiajunwang: I like the design. But it could be confusing why _capacityMetricsMap is not part of the _longMetricsMap.
Can we make it more differentiate?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
```diff
@@ -158,41 +155,39 @@ public synchronized void unregister() {
   }
 
   @Override
-  public Object getAttribute(String attribute)
-      throws AttributeNotFoundException, MBeanException, ReflectionException {
```
jiajunwang: Please keep these exceptions in the signature. We might want to complete the logic here by throwing the right exceptions moving forward.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
```diff
@@ -158,41 +155,39 @@ public synchronized void unregister() {
...
+      Object value = getAttribute(attributeName);
+      attributeList.add(new Attribute(attributeName, value));
     }
     return attributeList;
   }
 
   @Override
   public MBeanInfo getMBeanInfo() {
-    return _mBeanInfo;
+    synchronized (_lock) {
```
jiajunwang: As we discussed, this sync seems not working. Just remove it?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
```diff
@@ -158,41 +155,39 @@ public synchronized void unregister() {
...
+      return _mBeanInfo;
+    }
   }
 
   @Override
-  public void setAttribute(Attribute attribute) throws AttributeNotFoundException,
-      InvalidAttributeValueException, MBeanException, ReflectionException {
```
jiajunwang: I know you want to keep the code clean, but this list will remind us what to do if we implement this method in the future. So I would prefer to keep them.
No strong preference. You might want more reviewers to check it : )

helix-core/src/main/java/org/apache/helix/monitoring/metrics/MetricValueType.java
```diff
@@ -0,0 +1,29 @@
...
+ * under the License.
+ */
+
+/**
+ * Represents type of mbean metric's value.
+ * DOUBLE: metric value type is double.
+ * LONG: metric value type is long.
+ */
+public enum MetricValueType {
+  DOUBLE, LONG
```
jiajunwang: What's the difference between this type andy the original Double and Long?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
```diff
@@ -59,6 +55,7 @@
 
   // Attribute name to the DynamicMetric object mapping
   private final Map<String, DynamicMetric> _attributeMap = new HashMap<>();
+  private final Object _lock = new Object();
```
jiajunwang: Have you tried what we discussed that we don't lock at all? Since the strong consistency is not possible anyway.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/MetricValueType.java
pkuwm: Why I want to have this enum type is, I want to avoid using `InstanceOf` or `typeOf` to check Double or Long types. This enum makes it easier to check Double or Long without impacting the performance like InstanceOf.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
pkuwm: In my opinion, an implemented method could throw less exceptions, but not more exceptions than the interface. If the implementation doesn't throw any exception, we don't have to put them with the implementation.
If we want to implement this method in the future, we could definitely refer to the interface.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/InstanceMonitor.java
pkuwm: Currently I don't have a better name for this. Maybe we can add clearer comments for it?

jiajunwang: Sure, it should work.
Or call _capacityMetricsMap -> _dynamicCapacityMetricsMap to emphasis that it is changeable?

helix-core/src/main/java/org/apache/helix/monitoring/metrics/MetricValueType.java
jiajunwang: How about:
ENABLED_STATUS_GAUGE("Enabled", Long.class)
And you don't even need that switch-case block, instead just call the default constructor of the class to create the initial value.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -367,24 +367,25 @@ public void increaseMessageReceived(List<Message> messages) {
...
-      for (Map.Entry<String, Double> entry : maxUsageMap.entrySet()) {
-        InstanceMonitor monitor = _instanceMonitorMap.get(entry.getKey());
-        if (monitor == null) {
-          LOG.warn("Failed to update max usage because instance monitor is not found, instance: {}.",
-              entry.getKey());
-          continue;
-        }
-        monitor.updateMaxCapacityUsage(entry.getValue());
-      }
+  public void updateInstanceCapacityStatus(String instanceName, double maxUsage,
```
narendly: Will changing the name of a public method break backward-compatibility?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/InstanceMonitor.java
```diff
@@ -42,23 +43,29 @@
   /**
    * Metric names for instance capacity.
    */
-  public enum InstanceMonitorMetrics {
+  public enum InstanceMonitorMetric {
```
narendly: Will changing the name of a public enum break backward-compatibility?

pkuwm: No. This class was added in a previous commit for adding the MaxCapacityUsageGauge. It is not yet deployed into production.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
pkuwm: No. It is not yet deployed. This api was added one week ago.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/InstanceMonitor.java
pkuwm: I also considered _dynamicCapacityMetricsMap / _mutableCapacityMetricsMap. I will change the name and also add more comments about the mutable map.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
pkuwm: Discussed offline. Will consider making it simpler and removing the lock.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
```diff
@@ -159,28 +163,30 @@ public synchronized void unregister() {
...
+      throws AttributeNotFoundException {
     if (SENSOR_NAME_TAG.equals(attribute)) {
       return getSensorName();
     }
 
-    if (!_attributeMap.containsKey(attribute)) {
-      return null;
+    DynamicMetric metric = _attributeMap.get(attribute);
+    if (metric == null) {
+      throw new AttributeNotFoundException("Attribute[" + attribute + "] is not found.");
```
jiajunwang: Have you confirmed what will be the output when the get method returns this Exception?
Here's what I want to test:
1. Make the MBeanInfo async with the list. See what is the value and if the fetching value from the MBean server works or not.
2. Fix the async part, then try to fetch the value again. See if the reporting logic still works fine and if we can fetch the right data again.
I just want to be safe.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
pkuwm: Right, already added back and implemented.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
```diff
@@ -103,25 +100,30 @@ protected synchronized boolean doRegister(Collection<DynamicMetric<?, ?>> dynami
    * @param description description of the MBean
    * @param dynamicMetrics the DynamicMetrics
    */
-  private void updateAttributtInfos(Collection<DynamicMetric<?, ?>> dynamicMetrics,
+  protected void updateAttributesInfo(Collection<DynamicMetric<?, ?>> dynamicMetrics,
       String description) {
-    _attributeMap.clear();
+    if (dynamicMetrics == null || dynamicMetrics.isEmpty()) {
```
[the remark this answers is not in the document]
pkuwm: I think so, as well. The original code checks null, so I do it to avoid changing the logic. If you also agree on not checking null, I can remove the check.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
pkuwm: Let me do test it.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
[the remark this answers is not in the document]
jiajunwang: Check for null looks good to me. I mean "dynamicMetrics.isEmpty()" is not necessary.
This is a way to remove the metrics in the list.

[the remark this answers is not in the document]
pkuwm: Changed.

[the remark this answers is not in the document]
pkuwm: Using an empty collection to remove the metrics is kind of hacky way to me. Let's confirm:
1. Do we allow null or empty to delete metrics? The original code allows this.
2. If we don't allow null to delete metrics, we will do nothing if it is null. We allow empty collection to delete metrics.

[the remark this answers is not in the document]
jiajunwang: Or you can disallow it and add another method to clear the list. Both work for me : )

jiajunwang:
@pkuwm So this PR has been pending for a while. When you have time, could you please rebase and notify us if it is ready again for the reviewers.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/dynamicMBeans/DynamicMBeanProvider.java
pkuwm: Test results:
1. MBeanInfo async with the list.
Running 2 threads: thread 1 awaits while it is updating attributes "A" and "B" (not completed); meanwhile, thread 2 updates attributes "C", "D" and completes. Finally, "C", "D" could be fetched correctly, while getting "A" and "B" throws AttributeNotFoundException. thread 1 completes updating. Fetching "A", "B" correctly, and throwing AttributeNotFoundException for fetching "C", "D"
2. Fix async.
Has run a few tests. It works fine and confirms to fetch the data correctly.

pkuwm:

```
We need to monitor instance utilization in purpose of understanding what the instance capacity is.

Change list:
- Change instance monitor to update capacity
- Change getAttribute to throw AttributeNotFoundException in DynamicMBeanProvider
- Combine max usage and instance capacity update into one method in cluster status monitor
- Add unit test
```

PR #564, opened 2019-11-02
Avoid redundant writes in AssignmentMetadataStore

### LINKED ISSUE #562: AssignmentMetadataStore should skip updating if the new version is identical to the old one
Currently, our AssignmentMetadataStore always update to ZK on any new version of assignment data. This is not necessary and increases ZK traffic.
We should compare the new version with the in-memory cache and avoid write if updating is not necessary.

narendly:
### Issues

Fixes #562 

### Description

For the WAGED rebalancer, we persist the cluster's mapping via AssignmentMetadataStore every pipeline. However, if there are no changes made to the new assignment from the old assignment, this write is not necessary. This diff checks whether they are equal and skips the write if old and new assignments are the same.

### Tests

testAvoidingRedundantWrite

[ERROR] Tests run: 1046, Failures: 2, Errors: 0, Skipped: 3, Time elapsed: 3,848.329 s <<< FAILURE! - in TestSuite
[ERROR] testForceDeleteJobFromJobQueue(org.apache.helix.integration.task.TestDeleteJobFromJobQueue)  Time elapsed: 0.24 s  <<< FAILURE!

[ERROR] testEnableCompressionResource(org.apache.helix.integration.TestEnableCompression)  Time elapsed: 131.218 s  <<< FAILURE!
	at org.apache.helix.integration.TestEnableCompression.testEnableCompressionResource(TestEnableCompression.java:108)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 1046, Failures: 2, Errors: 0, Skipped: 3
[INFO] 
[INFO] BUILD FAILURE

--------------------------------------------
Individual runs passed.

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -42,15 +42,15 @@ public class AssignmentMetadataStore {
   private static final String ASSIGNMENT_METADATA_KEY = "ASSIGNMENT_METADATA";
   private static final String BASELINE_TEMPLATE = "/%s/%s/BASELINE";
   private static final String BEST_POSSIBLE_TEMPLATE = "/%s/%s/BEST_POSSIBLE";
-  private static final String BASELINE_KEY = "BASELINE";
-  private static final String BEST_POSSIBLE_KEY = "BEST_POSSIBLE";
+  static final String BASELINE_KEY = "BASELINE";
+  static final String BEST_POSSIBLE_KEY = "BEST_POSSIBLE";
   private static final ZkSerializer SERIALIZER = new ZNRecordJacksonSerializer();
 
-  private BucketDataAccessor _dataAccessor;
-  private String _baselinePath;
-  private String _bestPossiblePath;
-  private Map<String, ResourceAssignment> _globalBaseline;
-  private Map<String, ResourceAssignment> _bestPossibleAssignment;
+  BucketDataAccessor _dataAccessor;
+  String _baselinePath;
+  String _bestPossiblePath;
+  Map<String, ResourceAssignment> _globalBaseline;
+  Map<String, ResourceAssignment> _bestPossibleAssignment;
 
   AssignmentMetadataStore(String metadataStoreAddrs, String clusterName) {
     this(new ZkBucketDataAccessor(metadataStoreAddrs), clusterName);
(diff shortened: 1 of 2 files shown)


helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/TestAssignmentMetadataStore.java
```diff
@@ -81,7 +90,7 @@ public void beforeClass() throws Exception {
     _manager.connect();
 
     // create AssignmentMetadataStore
-    _store = new AssignmentMetadataStore(_manager.getMetadataStoreConnectionString(),
+    _store = new MockAssignmentMetadataStore(_manager.getMetadataStoreConnectionString(),
```
jiajunwang: Why we need the mock store here? If it is not testing against the original class, this test case won't be valid, right? Imaging someone modifies the real store class without touching this mock, the test case will still pass.

I guess you create this mock class for 2 reasons:
1. To expose the internal field. For this, we can create some protected methods that can be used to access these private fields.
2. To add the counters. For this, we might have to introduce a child class, but instead of re-write the whole method logic, we should at least call super.persistBaseline(), for example, to ensure the origin class'es method logic is tested here.

jiajunwang: Moreover, I think even we need this mock, it should be used for testAvoidingRedundantWrite() only. As for the other test method, they should be still using the original assignment store class.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -42,15 +42,15 @@
   private static final String ASSIGNMENT_METADATA_KEY = "ASSIGNMENT_METADATA";
   private static final String BASELINE_TEMPLATE = "/%s/%s/BASELINE";
   private static final String BEST_POSSIBLE_TEMPLATE = "/%s/%s/BEST_POSSIBLE";
-  private static final String BASELINE_KEY = "BASELINE";
-  private static final String BEST_POSSIBLE_KEY = "BEST_POSSIBLE";
+  static final String BASELINE_KEY = "BASELINE";
+  static final String BEST_POSSIBLE_KEY = "BEST_POSSIBLE";
```
jiajunwang: Please refer to my previous comments, I don't think you need to remove private if the mock class is extending the methods by calling super.method() first.

narendly:

PR #569, opened 2019-11-05
Fix partitions double charged for pending message

### LINKED ISSUE #568: Double charging for existing pending message
In the IntermediateCalStage, Helix does a double charge for existing pending message. For example, when you have a message pending for the partition, it will be charged for pending ST first. Then when we do charge for recovery/load balance partitions, they will be charged again.

dasahcc:
### Issues

Fixes #568 

### Description

Current Helix will charge the partitions with pending message twice due to not remove the partition from partitions need recovery/load balance set when they are charged for pending message.

This fix will fix the problem for recovery/load/ANY rebalance type charged for pending messages.

### Tests

Adding pending message piece for a partition.

[ERROR] Failures: 
[ERROR]   TestPartitionMovementThrottle.testResourceThrottle:162->getMaxParallelTransitionCount:376 » ConcurrentModification
[INFO] 

[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 77.865 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
@@ -363,7 +363,8 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
     }
 
     chargePendingTransition(resource, currentStateOutput, throttleController,
-        partitionsNeedRecovery, partitionsNeedLoadBalance, cache);
+        partitionsNeedRecovery, partitionsNeedLoadBalance, cache,
+        bestPossiblePartitionStateMap, intermediatePartitionStateMap);
 
     // Perform recovery balance
     Set<Partition> recoveryThrottledPartitions =
@@ -461,7 +462,9 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
    */
   private void chargePendingTransition(Resource resource, CurrentStateOutput currentStateOutput,
       StateTransitionThrottleController throttleController, Set<Partition> partitionsNeedRecovery,
-      Set<Partition> partitionsNeedLoadbalance, ResourceControllerDataProvider cache) {
+      Set<Partition> partitionsNeedLoadbalance, ResourceControllerDataProvider cache,
+      PartitionStateMap bestPossiblePartitionStateMap,
+      PartitionStateMap intermediatePartitionStateMap) {
     String resourceName = resource.getResourceName();
 
     // check and charge pending transitions
@@ -487,10 +490,20 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
           String pendingState = pendingMap.get(instance);
           if (pendingState != null && !pendingState.equals(currentState)
               && !cache.getDisabledInstancesForPartition(resourceName, partition.getPartitionName())
-                  .contains(instance)) {
+              .contains(instance)) {
             // Only charge this instance if the partition is not disabled
             throttleController.chargeInstance(rebalanceType, instance);
             shouldChargePartition = true;
+            // If there is a pending state transition for the partition, that means that an assignment
+            // has already been made and the state transition message has already been sent out for the partition
+            // in a previous pipeline run. We must honor this and reflect it by charging for the pending state transition message.
+
+            // Since the assignment has already been made for the pending message, we do a special treatment
+            // for it by setting the best possible state directly in intermediatePartitionStateMap so that the pending
+            // message won't be double-assigned or double-charged in recovery or load balance.
+            handlePendingStateTransitionsForThrottling(partition, partitionsNeedRecovery,
+                partitionsNeedLoadbalance, rebalanceType, bestPossiblePartitionStateMap,
+                intermediatePartitionStateMap);
           }
         }
         if (shouldChargePartition) {
@@ -943,6 +956,32 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
         }
       }
       return matchedState;
+    }
+  }
+
+  /**
+   * Handle a partition with a pending message so that the partition will not be double-charged or double-assigned during recovery and load balance.
+   * @param partition
+   * @param partitionsNeedRecovery
+   * @param partitionsNeedLoadbalance
+   * @param rebalanceType
+   */
+  private void handlePendingStateTransitionsForThrottling(Partition partition,
+      Set<Partition> partitionsNeedRecovery, Set<Partition> partitionsNeedLoadbalance,
+      RebalanceType rebalanceType, PartitionStateMap bestPossiblePartitionStateMap,
+      PartitionStateMap intermediatePartitionStateMap) {
+    // Pass the best possible state directly into intermediatePartitionStateMap
+    // This is safe to do so because we already have a pending transition for this partition, implying that the assignment has been made in previous pipeline
+    intermediatePartitionStateMap
+        .setState(partition, bestPossiblePartitionStateMap.getPartitionMap(partition));
+    // Remove the partition's name from the set of partition (names) that need to be charged and assigned to prevent double-processing
+    switch (rebalanceType) {
+    case RECOVERY_BALANCE:
+      partitionsNeedRecovery.remove(partition);
+      break;
+    case LOAD_BALANCE:
+      partitionsNeedLoadbalance.remove(partition);
+      break;
     }
   }
 }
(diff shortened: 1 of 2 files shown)


helix-core/src/test/java/org/apache/helix/controller/stages/TestIntermediateStateCalcStage.java
```diff
@@ -248,26 +255,28 @@ public void testWithClusterConfigChange() {
 
     IntermediateStateOutput output = event.getAttribute(AttributeName.INTERMEDIATE_STATE.name());
 
-    for (String resource : resourceSet) {
+    for (String resource : resources) {
       // Note Assert.assertEquals won't work. If "actual" is an empty map, it won't compare
       // anything.
-      Assert.assertTrue(output.getPartitionStateMap(resource).getStateMap()
-          .equals(expectedResult.getPartitionStateMap(resource).getStateMap()));
+      Assert.assertEquals(output.getPartitionStateMap(resource).getStateMap(), expectedResult.getPartitionStateMap(resource).getStateMap());
```
i3wangyi: This is my test script to verify that some partitions do get double-charged, but it seems after the fix, there's no special assertions to claim that these partitions are not in the output?

helix-core/src/test/java/org/apache/helix/controller/stages/TestIntermediateStateCalcStage.java
```diff
@@ -77,6 +80,11 @@ public void testNoStateMissing() {
           if (resource.endsWith("0")) {
             // Regular recovery balance
             currentStateOutput.setCurrentState(resource, partition, instanceName, "OFFLINE");
+            // add blocked state transition messages
```
i3wangyi: Could you add some comments about the expected results after adding these pending ST messages? Like partition 'p0' shouldn't be in the output

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -487,10 +490,18 @@ private void chargePendingTransition(Resource resource, CurrentStateOutput curre
...
             // Only charge this instance if the partition is not disabled
             throttleController.chargeInstance(rebalanceType, instance);
             shouldChargePartition = true;
+            // Remove the partition marked as partition need recovery/load balance. Otherwise, they
+            // will be double charged when they already charged for pending messages.
+            // At the same time, we need to update the intermediate map since it was supposed to
+            // be updated when the recovery rebalance throttling piece.
+            cleanPartitionsAreCharged(partition, partitionsNeedRecovery, partitionsNeedLoadbalance,
+                rebalanceType);
+            intermediatePartitionStateMap
```
i3wangyi: bestPossiblePartitionStateMap and intermediatePartitionStateMap updates are addition to the bug fix, right? Why is it needed?

helix-core/src/test/java/org/apache/helix/controller/stages/TestIntermediateStateCalcStage.java
dasahcc: I am not sure whether you understand the scenario. Once the pending messages added to resource0 and the throttling set to 3. p0 and p1 will be charged for pending and only 1 quota leftover. Then for recovery throttle, p0 will take the quota and p1 will be throttled. So it is unexpected result.

With the fix, the expected result will match the computed result.

helix-core/src/test/java/org/apache/helix/controller/stages/TestIntermediateStateCalcStage.java
dasahcc: I think you do not understand the expect result. If they are throttled, their final mapping should be their current states.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
dasahcc: This is used for assignment. Intermediate assign should be null at first beginning. It will be filled with either BestPossible or CurrentState when it has been throttled or not.

helix-core/src/test/java/org/apache/helix/controller/stages/TestIntermediateStateCalcStage.java
i3wangyi: My observations of the test code are that in this final Assertion stage, other than the format change, the content doesn't change at all. It shouldn't be the case because you modified the behavior of adding the pending ST messages, right? Something or some comments need to be here. Otherwise, without adding or adding the ST messages, the test script stays same?
If I missed something, please let me know.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
i3wangyi: Or maybe put it in another question: what if we don't set the intermediatePartitionStateMap after the fix, what's wrong with it? Or shall we add the update logic if we don't have the fix?

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -945,4 +956,27 @@ private int getIdealStateMatched(Partition partition) {
...
+  private void cleanPartitionsAreCharged(Partition partition, Set<Partition> partitionsNeedRecovery,
+      Set<Partition> partitionsNeedLoadbalance, RebalanceType rebalanceType) {
+    switch (rebalanceType) {
+    case RECOVERY_BALANCE:
+      partitionsNeedRecovery.remove(partition);
+      break;
+    case LOAD_BALANCE:
+      partitionsNeedLoadbalance.remove(partition);
+      break;
+    case ANY:
```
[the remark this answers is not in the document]
dasahcc: Good catch!

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
narendly: I think we could tweak the structure to minimize confusion in the future. Here's a suggestion:

- First, please use this for the inline comment:

// If there is a pending state transition for the partition, that means that an assignment has already been made and the state transition message has already been sent out for the partition in a previous pipeline run. We must honor this and reflect it by charging for the pending state transition message.
// Since the assignment has already been made for the pending message, we do a special treatment for it by setting the best possible state directly in intermediatePartitionStateMap so that the pending message won't be double-assigned or double-charged in recovery or load balance.

- And please merge the two lines of logic by creating a method (instead of cleanPartitionsAreCharged):

```
  /**
   * Pretreat a partition with a pending message so that the partition will not be double-charged or double-assigned during recovery and load balance.
   * @param partition
   * @param partitionsNeedRecovery
   * @param partitionsNeedLoadbalance
   * @param rebalanceType
   */
  private void preAssignPendingTransitions(Partition partition, Set<Partition> partitionsNeedRecovery,
      Set<Partition> partitionsNeedLoadbalance, RebalanceType rebalanceType, PartitionStateMap bestPossiblePartitionStateMap,
      PartitionStateMap intermediatePartitionStateMap) {
    // Pass the best possible state directly into intermediatePartitionStateMap
    // This is safe to do so because we already have a pending transition for this partition, implying that the assignment has been made in previous pipeline
    intermediatePartitionStateMap.setState(partition, bestPossiblePartitionStateMap.getPartitionMap(partition));
    // Remove the partition's name from the set of partition (names) that need to be charged and assigned to prevent double-processing
    switch (rebalanceType) {
    case RECOVERY_BALANCE:
      partitionsNeedRecovery.remove(partition);
      break;
    case LOAD_BALANCE:
      partitionsNeedLoadbalance.remove(partition);
      break;
    }
  }
```

I think this way we make it 100% clear what our intentions are and why we're making this change.

dasahcc: Good comments and craftmanship, Hunter!

dasahcc: Just make the name handlePendingStateTransitionsForThrottling to make it clear. Pretreat sounds more like a hacky way.

narendly: @dasahcc Sure. I think that name makes sense.

dasahcc:
This PR is ready to be merged, approved by @i3wangyi @narendly @jiajunwang

PR #570, opened 2019-11-05
Add Java API for adding and validating resources for WAGED rebalancer

### LINKED ISSUE #553: Need Java API to create resource with weight configured and rebalance the resource.
With the new WAGED rebalancer, users are required to provide with more information such as partition weight when they create resources. It would be helpful if we have an improved API in Helix admin to support this operation.

narendly:
### Issues

Fixes #553 

### Description

Add Java API for adding and validating resources for WAGED rebalancer. This is a set of convenience APIs provided through HelixAdmin the user could use to more easily add resources and validate them.

### Tests

testAddResourceWithWeightAndValidation
testEnableWagedRebalance

[ERROR] Tests run: 1051, Failures: 3, Errors: 0, Skipped: 3, Time elapsed: 3,899.986 s <<< FAILURE! - in TestSuite
[ERROR] testEnableCompressionResource(org.apache.helix.integration.TestEnableCompression)  Time elapsed: 150.607 s  <<< FAILURE!
	at org.apache.helix.integration.TestEnableCompression.testEnableCompressionResource(TestEnableCompression.java:108)

[ERROR] testMsgTriggeredRebalance(org.apache.helix.controller.stages.TestRebalancePipeline)  Time elapsed: 11.216 s  <<< FAILURE!
	at org.apache.helix.controller.stages.TestRebalancePipeline.testMsgTriggeredRebalance(TestRebalancePipeline.java:221)

[ERROR] testIncompleteInstanceCapacity(org.apache.helix.controller.rebalancer.waged.model.TestAssignableNode)  Time elapsed: 0.016 s  <<< FAILURE!
org.testng.TestException: 

The exception was thrown with the wrong message: expected "The required capacity keys: \[item2, item1, item3, AdditionalCapacityKey\] are not fully configured in the instance: testInstanceId, capacity map: \{item2=40, item1=20, item3=30\}." but got "The required capacity keys: [item2, item1, item3, AdditionalCapacityKey] are not fully configured in the instance: testInstanceConfigId, capacity map: {item2=40, item1=20, item3=30}."
	at org.apache.helix.controller.rebalancer.waged.model.TestAssignableNode.testIncompleteInstanceCapacity(TestAssignableNode.java:278)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestAssignableNode.testIncompleteInstanceCapacity » Test 
The exception was th...
[ERROR]   TestRebalancePipeline.testMsgTriggeredRebalance:221 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1051, Failures: 3, Errors: 0, Skipped: 3

--------

Fixed TestAssignableNode in this PR - PASS
TestRebalancePipeline - PASS
TestEnableCompression - PASS

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
@@ -30,6 +30,7 @@ import org.apache.helix.model.HelixConfigScope;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.MaintenanceSignal;
+import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
 
 /*
@@ -576,4 +577,28 @@ public interface HelixAdmin {
    * Release resources
    */
   void close();
+
+  /**
+   * Adds a resource with IdealState and ResourceConfig to be rebalanced by WAGED rebalancer with validation.
+   * @param clusterName
+   * @param idealState
+   * @param resourceConfig
+   * @return true if the resource has been added successfully. False otherwise
+   */
+  boolean addResourceWithWeight(String clusterName, IdealState idealState, ResourceConfig resourceConfig);
+
+  /**
+   * Batch-enables Waged rebalance for the names of resources given.
+   * @param clusterName
+   * @param resourceNames
+   * @return
+   */
+  boolean enableWagedRebalance(String clusterName, List<String> resourceNames);
+
+  /**
+   * Validates the resources to see if their weight configs have been set properly.
+   * @param resourceNames
+   * @return for each resource, true if the weight configs have been set properly, false otherwise
+   */
+  Map<String, Boolean> validateForWagedRebalance(String clusterName, List<String> resourceNames);
 }
--- a/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
@@ -22,6 +22,7 @@ package org.apache.helix;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+
 import org.apache.helix.model.ClusterConstraints;
 import org.apache.helix.model.ClusterConstraints.ConstraintType;
 import org.apache.helix.model.ConstraintItem;
@@ -32,6 +33,7 @@ import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.MaintenanceSignal;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
+
 
 /*
  * Helix cluster management
@@ -580,12 +582,17 @@ public interface HelixAdmin {
 
   /**
    * Adds a resource with IdealState and ResourceConfig to be rebalanced by WAGED rebalancer with validation.
+   * Validation includes the following:
+   * 1. Check ResourceConfig has the WEIGHT field
+   * 2. Check that all capacity keys from ClusterConfig are set up in the WEIGHT field
+   * 3. Check that all ResourceConfig's weightMap fields have all of the capacity keys
    * @param clusterName
    * @param idealState
    * @param resourceConfig
    * @return true if the resource has been added successfully. False otherwise
    */
-  boolean addResourceWithWeight(String clusterName, IdealState idealState, ResourceConfig resourceConfig);
+  boolean addResourceWithWeight(String clusterName, IdealState idealState,
+      ResourceConfig resourceConfig);
 
   /**
    * Batch-enables Waged rebalance for the names of resources given.
@@ -597,8 +604,25 @@ public interface HelixAdmin {
 
   /**
    * Validates the resources to see if their weight configs have been set properly.
+   * Validation includes the following:
+   * 1. Check ResourceConfig has the WEIGHT field
+   * 2. Check that all capacity keys from ClusterConfig are set up in the WEIGHT field
+   * 3. Check that all ResourceConfig's weightMap fields have all of the capacity keys
    * @param resourceNames
    * @return for each resource, true if the weight configs have been set properly, false otherwise
    */
-  Map<String, Boolean> validateForWagedRebalance(String clusterName, List<String> resourceNames);
+  Map<String, Boolean> validateResourcesForWagedRebalance(String clusterName,
+      List<String> resourceNames);
+
+  /**
+   * Validates the instances to ensure their weights in InstanceConfigs have been set up properly.
+   * Validation includes the following:
+   * 1. If default instance capacity is not set, check that the InstanceConfigs have the CAPACITY field
+   * 2. Check that all capacity keys defined in ClusterConfig are present in the CAPACITY field
+   * @param clusterName
+   * @param instancesNames
+   * @return
+   */
+  Map<String, Boolean> validateInstancesForWagedRebalance(String clusterName,
+      List<String> instancesNames);
 }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -57,6 +57,7 @@ import org.apache.helix.ZNRecord;
 import org.apache.helix.controller.rebalancer.DelayedAutoRebalancer;
 import org.apache.helix.controller.rebalancer.strategy.CrushEdRebalanceStrategy;
 import org.apache.helix.controller.rebalancer.strategy.RebalanceStrategy;
+import org.apache.helix.controller.rebalancer.util.WagedValidationUtil;
 import org.apache.helix.controller.rebalancer.waged.WagedRebalancer;
 import org.apache.helix.controller.rebalancer.waged.model.AssignableNode;
 import org.apache.helix.controller.rebalancer.waged.model.AssignableReplica;
@@ -1700,10 +1701,6 @@ public class ZKHelixAdmin implements HelixAdmin {
         continue;
       }
       ResourceConfig resourceConfig = _configAccessor.getResourceConfig(clusterName, resourceName);
-      if (resourceConfig == null || !resourceConfig.isValid()) {
-        result.put(resourceName, false);
-        continue;
-      }
       result.put(resourceName,
           validateWeightForResourceConfig(clusterConfig, resourceConfig, idealState));
     }
@@ -1729,7 +1726,7 @@ public class ZKHelixAdmin implements HelixAdmin {
         result.put(instanceName, false);
         continue;
       }
-      AssignableNode.validateAndGetInstanceCapacity(clusterConfig, instanceConfig);
+      WagedValidationUtil.validateAndGetInstanceCapacity(clusterConfig, instanceConfig);
       result.put(instanceName, true);
     }
 
@@ -1745,32 +1742,39 @@ public class ZKHelixAdmin implements HelixAdmin {
    */
   private boolean validateWeightForResourceConfig(ClusterConfig clusterConfig,
       ResourceConfig resourceConfig, IdealState idealState) {
+    // If ResourceConfig is null AND the default partition weight map is defined, we consider this resource valid
+    if (resourceConfig == null) {
+      return !clusterConfig.getDefaultPartitionWeightMap().isEmpty();
+      // ResourceConfig is null and no default partition weight map has been defined. Not valid.
+    }
+
     // Parse the entire capacityMap from ResourceConfig
     Map<String, Map<String, Integer>> capacityMap;
     try {
       capacityMap = resourceConfig.getPartitionCapacityMap();
     } catch (IOException ex) {
-      throw new IllegalArgumentException(
-          "Invalid partition capacity configuration of resource: " + resourceConfig
-              .getResourceName(), ex);
+      logger.error("Invalid partition capacity configuration of resource: {}", idealState.getResourceName(), ex);
+      return false;
     }
 
-    // Remove DEFAULT key
     Set<String> capacityMapSet = new HashSet<>(capacityMap.keySet());
     boolean hasDefaultCapacity = capacityMapSet.contains(ResourceConfig.DEFAULT_PARTITION_KEY);
+    // Remove DEFAULT key
     capacityMapSet.remove(ResourceConfig.DEFAULT_PARTITION_KEY);
 
     // Make sure capacityMap contains all partitions defined in IdealState
     // Here, IdealState has not been rebalanced, so listFields might be null, in which case, we would get an emptyList from getPartitionSet()
     // So check using numPartitions instead
+    // This check allows us to fail early on instead of having to loop through all partitions
     if (capacityMapSet.size() != idealState.getNumPartitions() && !hasDefaultCapacity) {
-      throw new IllegalArgumentException(String.format(
-          "ResourceConfig for %s does not have all partitions defined in PartitionCapacityMap!",
-          idealState.getResourceName()));
+      logger.error(
+          "ResourceConfig for {} does not have all partitions defined in PartitionCapacityMap!",
+          idealState.getResourceName());
+      return false;
     }
 
     // Loop through all partitions and validate
-    capacityMap.keySet().forEach(partitionName -> AssignableReplica
+    capacityMap.keySet().forEach(partitionName -> WagedValidationUtil
         .validateAndGetPartitionCapacity(partitionName, resourceConfig, capacityMap,
             clusterConfig));
     return true;
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -1680,6 +1680,17 @@ public class ZKHelixAdmin implements HelixAdmin {
       throw new HelixException("Resource name list is invalid!");
     }
 
+    // First, ensure that all instances are valid
+    HelixDataAccessor accessor =
+        new ZKHelixDataAccessor(clusterName, new ZkBaseDataAccessor<>(_zkClient));
+    Builder keyBuilder = accessor.keyBuilder();
+    List<String> liveInstances = accessor.getChildNames(keyBuilder.liveInstances());
+    if (validateInstancesForWagedRebalance(clusterName, liveInstances).containsValue(false)) {
+      throw new HelixException(String
+          .format("Instance capacities haven't been configured properly for cluster %s",
+              clusterName));
+    }
+
     Map<String, Boolean> result = new HashMap<>();
     ClusterConfig clusterConfig = _configAccessor.getClusterConfig(clusterName);
     for (String resourceName : resourceNames) {
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -75,6 +75,7 @@ import org.apache.helix.model.Message;
 import org.apache.helix.model.Message.MessageState;
 import org.apache.helix.model.Message.MessageType;
 import org.apache.helix.model.PauseSignal;
+import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
 import org.apache.helix.tools.DefaultIdealStateCalculator;
 import org.apache.helix.util.HelixUtil;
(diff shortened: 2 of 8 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1579,137 @@ public void close() {
...
+      // TODO This might be okay because of default weight?
+      throw new HelixException("ResourceConfig is null!");
+    }
+
+    // Order in which a resource should be added:
+    // 1. Validate the weights in ResourceConfig against ClusterConfig
+    // Check that all capacity keys in ClusterConfig are set up in every partition in ResourceConfig's WEIGHT field
+    if (!validateWeightForResourceConfig(_configAccessor.getClusterConfig(clusterName),
+        resourceConfig)) {
+      return false;
```
jiajunwang: Shall we throw Exceptions with the error information here?

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -576,4 +577,28 @@ void rebalance(String clusterName, String resourceName, int replica, String keyP
    * Release resources
    */
   void close();
+
+  /**
+   * Adds a resource with IdealState and ResourceConfig to be rebalanced by WAGED rebalancer with validation.
```
jiajunwang: Let's add the validation logic briefly for these new methods here.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1579,137 @@ public void close() {
...
+      ResourceConfig resourceConfig) {
+    // Null checks
+    if (idealState == null) {
+      throw new HelixException("IdealState is null!");
+    }
+    if (resourceConfig == null) {
+      // TODO This might be okay because of default weight?
+      throw new HelixException("ResourceConfig is null!");
+    }
+
```
jiajunwang: One more check to confirm the input IS and Config have the same resource name.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1579,137 @@ public void close() {
...
+    rebalance(clusterName, idealState.getResourceName(), Integer.parseInt(idealState.getReplicas()),
+        idealState.getResourceName(), idealState.getInstanceGroupTag());
+
+    return true;
+  }
+
+  @Override
+  public boolean enableWagedRebalance(String clusterName, List<String> resourceNames) {
+    // Null checks
+    if (clusterName == null || clusterName.isEmpty()) {
```
jiajunwang: nit, do the same for the addResourceWithWeight method?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1579,137 @@ public void close() {
...
+        new ZKHelixDataAccessor(clusterName, new ZkBaseDataAccessor<>(_zkClient));
+    Builder keyBuilder = accessor.keyBuilder();
+    List<IdealState> idealStates = accessor.getChildValues(keyBuilder.idealStates());
+    List<String> nullIdealStates = new ArrayList<>();
+    for (int i = 0; i < idealStates.size(); i++) {
+      if (idealStates.get(i) == null) {
+        nullIdealStates.add(resourceNames.get(i));
+      } else {
+        idealStates.get(i)
+            .setRebalancerClassName("org.apache.helix.controller.rebalancer.WagedRebalancer");
```
jiajunwang: Don't hardcode, WagedRebalancer.class.getName()

jiajunwang: In this case, you will need to change the rebalance mode to FULL-AUTO.
Or we just filter the non-FA resources out. What do you think?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1579,137 @@ public void close() {
...
+    if (clusterName == null || clusterName.isEmpty()) {
+      throw new HelixException("Cluster name is invalid!");
+    }
+    if (resourceNames == null || resourceNames.isEmpty()) {
+      throw new HelixException("Resource name list is invalid!");
+    }
+
+    Map<String, Boolean> result = new HashMap<>();
+    ClusterConfig clusterConfig = _configAccessor.getClusterConfig(clusterName);
+    for (String resourceName : resourceNames) {
```
jiajunwang: We will need to validate the instances as well.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1579,137 @@ public void close() {
...
+    return result;
+  }
+
+  /**
+   * Validates ResourceConfig's weight field against the given ClusterConfig.
+   * @param clusterConfig
+   * @param resourceConfig
+   * @return true if ResourceConfig has all the required fields. False otherwise.
+   */
+  private boolean validateWeightForResourceConfig(ClusterConfig clusterConfig,
```
jiajunwang: For this validation, please try to reuse the following code by some refactoring. Or we will have duplicate code to maintain.
- AssignableReplica.fetchCapacityUsage
- AssignableNode.fetchInstanceCapacity

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
jiajunwang: Can we have the same method checking for both? When enabling the WAGED rebalancer, we need to ensure everything is good. There is no case that we need to check the resources only.

As for the instance list, I don't think we need the user to input. Because the controller will rebalance based on the full instance list anyway.

narendly: So make sure that all instances have been validated?

narendly: Check added.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
narendly: Refactored into public static methods.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -576,4 +579,50 @@ void rebalance(String clusterName, String resourceName, int replica, String keyP
...
+  /**
+   * Validates the instances to ensure their weights in InstanceConfigs have been set up properly.
+   * Validation includes the following:
+   * 1. If default instance capacity is not set, check that the InstanceConfigs have the CAPACITY field
+   * 2. Check that all capacity keys defined in ClusterConfig are present in the CAPACITY field
+   * @param clusterName
+   * @param instancesNames
+   * @return
+   */
+  Map<String, Boolean> validateInstancesForWagedRebalance(String clusterName,
```
jiajunwang: This is different from what we agreed with the potential customer. But I can see why you made this choice. I don't have a strong preference here. I will forward this PR to them so they can take a look.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -505,4 +516,107 @@ public void testLegacyEnableDisablePartition() {
...
+    // validate
+    validationResult = admin.validateResourcesForWagedRebalance(clusterName,
+        Collections.singletonList(testResourcePrefix));
+    Assert.assertEquals(validationResult.size(), 1);
+    Assert.assertFalse(validationResult.get(testResourcePrefix));
+    try {
+      admin.addResourceWithWeight(clusterName, idealState, resourceConfig);
+    } catch (Exception e) {
+      // OK since ClusterConfig is empty
+    }
```
pkuwm: `Assert.fail()` is missing.
This is one of the cons using the traditional try...catch... to verify an expected exception. I am working on upgrading testng to support `assertThrows`, which looks much better than try..catch..

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -505,4 +516,107 @@ public void testLegacyEnableDisablePartition() {
...
+    ResourceConfig resourceConfig = new ResourceConfig(testResourcePrefix);
+    // validate
+    Map<String, Boolean> validationResult = admin.validateResourcesForWagedRebalance(clusterName,
+        Collections.singletonList(testResourcePrefix));
+    Assert.assertEquals(validationResult.size(), 1);
+    Assert.assertFalse(validationResult.get(testResourcePrefix));
+    try {
+      admin.addResourceWithWeight(clusterName, idealState, resourceConfig);
+      Assert.fail();
+    } catch (Exception e) {
```
pkuwm: If this is to verify resourceConfig is empty, are we expecting a HelixException, instead of a general Exception? HelixException may look more accurate in this case.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1586,193 @@ public void close() {
...
+    Map<String, Boolean> result = new HashMap<>();
+    ClusterConfig clusterConfig = _configAccessor.getClusterConfig(clusterName);
+    for (String resourceName : resourceNames) {
+      IdealState idealState = getResourceIdealState(clusterName, resourceName);
+      if (idealState == null || !idealState.isValid()) {
+        result.put(resourceName, false);
+        continue;
+      }
+      ResourceConfig resourceConfig = _configAccessor.getResourceConfig(clusterName, resourceName);
+      if (resourceConfig == null || !resourceConfig.isValid()) {
```
jiajunwang: It's a little bit tricky here. Even the resourceConfig is null, as long as the default weight is configured in the cluster config, we can still rebalance.
If there is no resource confic, shall we just pass null to validateWeightForResourceConfig and let it decide if it is overall valid configuration?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1586,193 @@ public void close() {
...
+  /**
+   * Validates ResourceConfig's weight field against the given ClusterConfig.
+   * @param clusterConfig
+   * @param resourceConfig
+   * @param idealState
+   * @return true if ResourceConfig has all the required fields. False otherwise.
+   */
+  private boolean validateWeightForResourceConfig(ClusterConfig clusterConfig,
+      ResourceConfig resourceConfig, IdealState idealState) {
+    // Parse the entire capacityMap from ResourceConfig
```
jiajunwang: As I comment above, I would prefer if this logic is also inside AssignableReplica .validateAndGetPartitionCapacity so we don't need this duplicate code here.
For this, we might need to add a null check in the validateAndGetPartitionCapacity. But I think it worths it.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
narendly: @jiajunwang Can you clearly point out how it's different? What was the original requirement and how this implementation is different?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
narendly: No. It's expecting an IllegalArgumentException

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
narendly: Good catch. This must slipped through the cracks.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1586,193 @@ public void close() {
...
+    Map<String, Map<String, Integer>> capacityMap;
+    try {
+      capacityMap = resourceConfig.getPartitionCapacityMap();
+    } catch (IOException ex) {
+      throw new IllegalArgumentException(
+          "Invalid partition capacity configuration of resource: " + resourceConfig
+              .getResourceName(), ex);
+    }
+
+    // Remove DEFAULT key
```
jiajunwang: I think validateAndGetPartitionCapacity already handles this default keys, and in addition, cluster config default weight config. Can we just rely on that logic?

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
jiajunwang: To be clear, we agreed to have a single API for validating everything. I see you have done this in the validateResourcesForWagedRebalance method. So not sure if we still need this one in addition.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1586,193 @@ public void close() {
...
+  public boolean addResourceWithWeight(String clusterName, IdealState idealState,
+      ResourceConfig resourceConfig) {
+    // Null checks
+    if (clusterName == null || clusterName.isEmpty()) {
+      throw new HelixException("Cluster name is null or empty!");
+    }
+    if (idealState == null || !idealState.isValid()) {
+      throw new HelixException("IdealState is null or invalid!");
+    }
+    if (resourceConfig == null || !resourceConfig.isValid()) {
```
pkuwm: `ResourceConfig.isValid()` always returns true. Is this expected or a bug? How to verify resourceConfig is empty?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
pkuwm: But from the code here: https://github.com/apache/helix/blob/890ec65f2f04d77cd05020c17415701bb62fc8ef/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java#L1601

A HelixException is thrown. Did I miss any place where an IllegalArgumentException is thrown?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
narendly: This is okay. I think if we need to add more logic here, we can in the future. However, that work isn't strictly in the scope of this diff.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -173,7 +179,7 @@ public void dropInstance(String clusterName, InstanceConfig instanceConfig) {
           // does not repeatedly write instance history)
           logger.warn("Retrying dropping instance {} with exception {}",
               instanceConfig.getInstanceName(), e.getCause().getMessage());
-          retryCnt ++;
+          retryCnt++;
```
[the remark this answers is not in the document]
narendly: I apologize. I think the IDE decided to format the entire file. I will be careful next time.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1586,193 @@ public void close() {
...
+    }
+
+    HelixDataAccessor accessor =
+        new ZKHelixDataAccessor(clusterName, new ZkBaseDataAccessor<>(_zkClient));
+    Builder keyBuilder = accessor.keyBuilder();
+    List<IdealState> idealStates = accessor.getChildValues(keyBuilder.idealStates());
+    List<String> nullIdealStates = new ArrayList<>();
+    for (int i = 0; i < idealStates.size(); i++) {
+      if (idealStates.get(i) == null) {
+        nullIdealStates.add(resourceNames.get(i));
```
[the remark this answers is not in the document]
narendly: This was done on purpose and a design choice I made. I think this API is only useful if we actually let the user know which ones are not configured correctly. I print this in the exception message (see line 1658).

Otherwise, the user has to potentially go through hundreds of resources to be able to get them ready for WAGED rebalance.

What do you think?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1586,193 @@ public void close() {
...
+    }
+    if (resourceNames == null || resourceNames.isEmpty()) {
+      throw new HelixException("Resource name list is invalid!");
+    }
+
+    // First, ensure that all instances are valid
+    HelixDataAccessor accessor =
+        new ZKHelixDataAccessor(clusterName, new ZkBaseDataAccessor<>(_zkClient));
+    Builder keyBuilder = accessor.keyBuilder();
+    List<String> liveInstances = accessor.getChildNames(keyBuilder.liveInstances());
```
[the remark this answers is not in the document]
narendly: Thanks for the catch. Will fix.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1586,193 @@ public void close() {
...
+        result.put(resourceName, false);
+        continue;
+      }
+      ResourceConfig resourceConfig = _configAccessor.getResourceConfig(clusterName, resourceName);
+      if (resourceConfig == null || !resourceConfig.isValid()) {
+        result.put(resourceName, false);
+        continue;
+      }
+      result.put(resourceName,
+          validateWeightForResourceConfig(clusterConfig, resourceConfig, idealState));
```
narendly: TODO: use a try-catch to get the boolean result for each resource because validateWeightForResourceConfig() will throw an exception and never return false.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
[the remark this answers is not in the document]
jiajunwang: Makes sense.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
jiajunwang: I think we made it clear in the other thread. I think this API design is good.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
jiajunwang: That makes sense. I should improve this in my PR actually. This should speed it up little bit. Thanks for pointing this out.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1578,4 +1587,196 @@ public void close() {
...
+   * @param clusterConfig
+   * @param resourceConfig
+   * @param idealState
+   * @return true if ResourceConfig has all the required fields. False otherwise.
+   */
+  private boolean validateWeightForResourceConfig(ClusterConfig clusterConfig,
+      ResourceConfig resourceConfig, IdealState idealState) {
+    // If ResourceConfig is null AND the default partition weight map is defined, we consider this resource valid
+    if (resourceConfig == null) {
+      return !clusterConfig.getDefaultPartitionWeightMap().isEmpty();
```
jiajunwang: Add one more check to verify if the default weight map contains all the required keys.

narendly: I enhanced the check. Could you take a look and confirm that the logic is sound? @jiajunwang

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1742,18 +1742,32 @@ public boolean enableWagedRebalance(String clusterName, List<String> resourceNam
...
+      }
+      // If ResourceConfig is null AND the default partition weight map is defined, and the map has all the required keys, we consider this valid since the default weights will be used
+      // Need to check the map contains all the required keys
+      if (clusterConfig.getDefaultPartitionWeightMap().keySet()
+          .containsAll(clusterConfig.getInstanceCapacityKeys())) {
+        // Contains all the required keys, so consider it valid since it will use the default weights
+        return true;
+      }
+      logger.error(
+          "ClusterConfig's default partition weight map doesn't have all the required keys!");
```
jiajunwang: nit, "ResourceConfig for {} is null, and ClusterConfig's default partition weight map doesn't have all the required keys!"

narendly: Updated.

narendly:

Add Java API for adding and validating resources for WAGED rebalancer

Add Java API methods for adding and validating resources for WAGED rebalancer. This is a set of convenience APIs provided through HelixAdmin the user could use to more easily add resources and validate them for WAGED rebalance usage.

Issue #563, opened 2019-11-01
Improve the WAGED rebalancer calculating speed.


jiajunwang:
During our recent tests, we notice the new WAGED rebalancer calculating is not fast enough. This issue is for tracking all the performance tests and improvements.

jiajunwang:
1. Potential algorithm inefficiency because of the max utilization calculating.
2. The rebalancer should skip all calculations while nothing has been changed.

PR #586, opened 2019-11-12
Improve the WAGED rebalancer performance.

jiajunwang:
### Issues

#563 

### Description

This change improves the rebalance's speed by 2x to 5x depends on the host capacity.
1. Parallelism the loop processing whenever possible and do not change the logic.
2. Avoid some duplicate logic in the loop. Extract the calculation outside the loop.
The logic of rebalancer is not changed in this change.

### Tests

No business logic change, so test with the existing test cases.

[ERROR] Failures: 
[INFO]   Run 4: PASS
[INFO]   Run 5: PASS
[INFO] 
[ERROR]   TestZKUtil.testChildrenOperations:78 expected:<2> but was:<3>
[ERROR]   TestGetLastScheduledTaskExecInfo.testGetLastScheduledTaskExecInfo:70 expected:<1573527192711> but was:<1573527191553>
[INFO] 
[ERROR] Tests run: 1045, Failures: 5, Errors: 0, Skipped: 3
[INFO] 

Run-run

[INFO] Tests run: 37, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 26.058 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 37, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS

### Commits

### Documentation

NA

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
@@ -84,7 +84,17 @@ public class WagedRebalancer {
   private final HelixManager _manager;
   private final MappingCalculator<ResourceControllerDataProvider> _mappingCalculator;
   private final AssignmentMetadataStore _assignmentMetadataStore;
+
   private final MetricCollector _metricCollector;
+  private final CountMetric _rebalanceFailureCount;
+  private final CountMetric _globalBaselineCalcCounter;
+  private final LatencyMetric _globalBaselineCalcLatency;
+  private final LatencyMetric _writeLatency;
+  private final CountMetric _partialRebalanceCounter;
+  private final LatencyMetric _partialRebalanceLatency;
+  private final LatencyMetric _stateReadLatency;
+  private final BaselineDivergenceGauge _baselineDivergenceGauge;
+
   private RebalanceAlgorithm _rebalanceAlgorithm;
   private Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> _preference =
       NOT_CONFIGURED_PREFERENCE;
@@ -155,10 +165,38 @@ public class WagedRebalancer {
     _rebalanceAlgorithm = algorithm;
     _mappingCalculator = mappingCalculator;
     _manager = manager;
+
     // If metricCollector is null, instantiate a version that does not register metrics in order to
     // allow rebalancer to proceed
     _metricCollector =
         metricCollector == null ? new WagedRebalancerMetricCollector() : metricCollector;
+    _rebalanceFailureCount = _metricCollector.getMetric(
+        WagedRebalancerMetricCollector.WagedRebalancerMetricNames.RebalanceFailureCounter.name(),
+        CountMetric.class);
+    _globalBaselineCalcCounter = _metricCollector.getMetric(
+        WagedRebalancerMetricCollector.WagedRebalancerMetricNames.GlobalBaselineCalcCounter.name(),
+        CountMetric.class);
+    _globalBaselineCalcLatency = _metricCollector.getMetric(
+        WagedRebalancerMetricCollector.WagedRebalancerMetricNames.GlobalBaselineCalcLatencyGauge
+            .name(),
+        LatencyMetric.class);
+    _partialRebalanceCounter = _metricCollector.getMetric(
+        WagedRebalancerMetricCollector.WagedRebalancerMetricNames.PartialRebalanceCounter.name(),
+        CountMetric.class);
+    _partialRebalanceLatency = _metricCollector.getMetric(
+        WagedRebalancerMetricCollector.WagedRebalancerMetricNames.PartialRebalanceLatencyGauge
+            .name(),
+        LatencyMetric.class);
+    _writeLatency = _metricCollector.getMetric(
+        WagedRebalancerMetricCollector.WagedRebalancerMetricNames.StateWriteLatencyGauge.name(),
+        LatencyMetric.class);
+    _stateReadLatency = _metricCollector.getMetric(
+        WagedRebalancerMetricCollector.WagedRebalancerMetricNames.StateReadLatencyGauge.name(),
+        LatencyMetric.class);
+    _baselineDivergenceGauge = _metricCollector.getMetric(
+        WagedRebalancerMetricCollector.WagedRebalancerMetricNames.BaselineDivergenceGauge.name(),
+        BaselineDivergenceGauge.class);
+
     _changeDetector = new ResourceChangeDetector(true);
   }
 
@@ -209,10 +247,7 @@ public class WagedRebalancer {
     } catch (HelixRebalanceException ex) {
       LOG.error("Failed to calculate the new assignments.", ex);
       // Record the failure in metrics.
-      CountMetric rebalanceFailureCount = _metricCollector.getMetric(
-          WagedRebalancerMetricCollector.WagedRebalancerMetricNames.RebalanceFailureCounter.name(),
-          CountMetric.class);
-      rebalanceFailureCount.increment(1L);
+      _rebalanceFailureCount.increment(1L);
 
       HelixRebalanceException.Type failureType = ex.getFailureType();
       if (failureType.equals(HelixRebalanceException.Type.INVALID_REBALANCER_STATUS) || failureType
@@ -236,22 +271,23 @@ public class WagedRebalancer {
     // Construct the new best possible states according to the current state and target assignment.
     // Note that the new ideal state might be an intermediate state between the current state and
     // the target assignment.
-    for (IdealState is : newIdealStates.values()) {
-      String resourceName = is.getResourceName();
+    newIdealStates.values().parallelStream().forEach(idealState -> {
+      String resourceName = idealState.getResourceName();
       // Adjust the states according to the current state.
-      ResourceAssignment finalAssignment = _mappingCalculator.computeBestPossiblePartitionState(
-          clusterData, is, resourceMap.get(resourceName), currentStateOutput);
+      ResourceAssignment finalAssignment = _mappingCalculator
+          .computeBestPossiblePartitionState(clusterData, idealState, resourceMap.get(resourceName),
+              currentStateOutput);
 
       // Clean up the state mapping fields. Use the final assignment that is calculated by the
       // mapping calculator to replace them.
-      is.getRecord().getMapFields().clear();
+      idealState.getRecord().getMapFields().clear();
       for (Partition partition : finalAssignment.getMappedPartitions()) {
         Map<String, String> newStateMap = finalAssignment.getReplicaMap(partition);
         // if the final states cannot be generated, override the best possible state with empty map.
-        is.setInstanceStateMap(partition.getPartitionName(),
+        idealState.setInstanceStateMap(partition.getPartitionName(),
             newStateMap == null ? Collections.emptyMap() : newStateMap);
       }
-    }
+    });
     LOG.info("Finish computing new ideal states for resources: {}",
         resourceMap.keySet().toString());
     return newIdealStates;
@@ -296,8 +332,8 @@ public class WagedRebalancer {
       Set<String> activeNodes, final CurrentStateOutput currentStateOutput)
       throws HelixRebalanceException {
     getChangeDetector().updateSnapshots(clusterData);
-    // Get all the changed items' information
-    Map<HelixConstants.ChangeType, Set<String>> clusterChanges =
+    // Get all the changed items' information. Filter for the items that have content changed.
+    final Map<HelixConstants.ChangeType, Set<String>> clusterChanges =
         getChangeDetector().getChangeTypes().stream()
             .collect(Collectors.toMap(changeType -> changeType, changeType -> {
               Set<String> itemKeys = new HashSet<>();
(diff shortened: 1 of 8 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -296,27 +332,21 @@ public void close() {
       Set<String> activeNodes, final CurrentStateOutput currentStateOutput)
       throws HelixRebalanceException {
     getChangeDetector().updateSnapshots(clusterData);
-    // Get all the changed items' information
-    Map<HelixConstants.ChangeType, Set<String>> clusterChanges =
+    // Get all the changed items' information. Filter for the items that have content changed.
+    final Map<HelixConstants.ChangeType, Set<String>> clusterChanges =
```
narendly: can you parallel stream this?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -296,27 +332,21 @@ public void close() {
...
               Set<String> itemKeys = new HashSet<>();
               itemKeys.addAll(getChangeDetector().getAdditionsByType(changeType));
               itemKeys.addAll(getChangeDetector().getChangesByType(changeType));
               itemKeys.addAll(getChangeDetector().getRemovalsByType(changeType));
               return itemKeys;
-            }));
-    // Filter for the items that have content changed.
-    clusterChanges =
-        clusterChanges.entrySet().stream().filter(changeEntry -> !changeEntry.getValue().isEmpty())
+            })).entrySet().stream().filter(changeEntry -> !changeEntry.getValue().isEmpty())
```
narendly: can you parallel stream this?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -362,47 +392,41 @@ public void close() {
...
-            LatencyMetric.class);
-        writeLatency.startMeasuringLatency();
-        _assignmentMetadataStore.persistBaseline(newBaseline);
-        writeLatency.endMeasuringLatency();
-      } catch (Exception ex) {
-        throw new HelixRebalanceException("Failed to persist the new baseline assignment.",
-            HelixRebalanceException.Type.INVALID_REBALANCER_STATUS, ex);
+      final CurrentStateOutput currentStateOutput)
+      throws HelixRebalanceException {
+    if (clusterChanges.keySet().stream()
```
narendly: can you parallel stream this?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
```diff
@@ -142,21 +142,21 @@ void release(AssignableReplica replica) throws IllegalArgumentException {
...
     removedReplica.getCapacity().entrySet().stream()
-        .forEach(entry -> updateCapacityAndUtilization(entry.getKey(), -1 * entry.getValue()));
+        .forEach(entry -> updateRemainingCapacity(entry.getKey(), -1 * entry.getValue()));
   }
 
   /**
    * @return A set of all assigned replicas on the node.
    */
-  public Set<AssignableReplica> getAssignedReplicas() {
+  Set<AssignableReplica> getAssignedReplicas() {
```
narendly: can you parallel stream this?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -104,16 +104,17 @@ public OptimalAssignment calculate(ClusterModel clusterModel) throws HelixRebala
...
     if (candidateNodes.isEmpty()) {
       optimalAssignment.recordAssignmentFailure(replica,
           Maps.transformValues(hardConstraintFailures, this::convertFailureReasons));
       return Optional.empty();
     }
 
-    Function<AssignableNode, Double> calculatePoints =
-        (candidateNode) -> getAssignmentNormalizedScore(candidateNode, replica, clusterContext);
-
-    return candidateNodes.stream().max(Comparator.comparing(calculatePoints));
```
i3wangyi: can you try `candidateNodes.parallelStream().max(Comparator.comparing(calculatePoints));`?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -170,10 +176,8 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
             // Note that to ensure the algorithm is deterministic with the same inputs, do not use
             // Random functions here. Use hashcode based on the cluster topology information to get
             // a controlled randomized order is good enough.
-            Long replicaHash1 = (long) Objects
-                .hash(replica1.toString(), clusterModel.getAssignableNodes().keySet());
-            Long replicaHash2 = (long) Objects
-                .hash(replica2.toString(), clusterModel.getAssignableNodes().keySet());
+            Integer replicaHash1 = replicaHashCodeMap.get(replica1.toString());
```
i3wangyi: nit: primitive int

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -146,6 +147,11 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
     Map<String, ResourceAssignment> baselineAssignment =
         clusterModel.getContext().getBaselineAssignment();
 
+    Map<String, Integer> replicaHashCodeMap = orderedAssignableReplicas.parallelStream().collect(
+        Collectors.toMap(AssignableReplica::toString,
+            replica -> Objects.hash(replica.toString(), clusterModel.getAssignableNodes().keySet()),
```
i3wangyi: It would be simpler if you add a `hashcode()` method inside the AssignableReplica class.
It becomes 
```
Map<String, Integer> replicaHashCodeMap = orderedAssignableReplicas.parallelStream().collect(
        Collectors.toMap(AssignableReplica::toString, AssignableReplica::hashCode);
```

i3wangyi: And do you really need this level of parallism? If so, you might want to update stream() in line 142 as well

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
i3wangyi: Let's be aware of using parallelStream in non-expensive operations because parallelStream itself is more expensive than simple stream

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
```diff
@@ -37,6 +37,7 @@
 public class AssignableReplica implements Comparable<AssignableReplica> {
   private static final Logger LOG = LoggerFactory.getLogger(AssignableReplica.class);
 
+  private final String _replicaKey;
```
i3wangyi: _replicaId is more appropriate

i3wangyi: or simpler `String _id`

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: As Yi pointed out, parallelStream is costly. So I tested multiple places. The conclusion is that only high-level parallelism is worthwhile. Doing it too much slow things down.

For this one, the change types are not a long list. So the gain won't beat the thread managing overhead.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: same here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: Still the same

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableNode.java
jiajunwang: Since it is already inside a parallelStream in the caller's logic, do parallel here will slow it down dramatically.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/AssignableReplica.java
jiajunwang: Not really, the key is not unique, so it is not really an Id. I don't want to imply that this is a unique Id that one can use as a hash key.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
jiajunwang: I tried, the current one is the fastest speed that I can get.
I think if you compare multiple times, the comparator will do calculating again and again. So pre-calculate the result is much faster somehow.

FYI, with your change:
generateClusterModel Took: 302
rebalanceAlgorithm.calculate took 9399
Finish calculating an assignment. Took: 9701
generateClusterModel Took: 122
rebalanceAlgorithm.calculate took 8546
Finish calculating an assignment. Took: 8668

With my change:
generateClusterModel Took: 295
rebalanceAlgorithm.calculate took 6520
Finish calculating an assignment. Took: 6815
generateClusterModel Took: 115
rebalanceAlgorithm.calculate took 6123
Finish calculating an assignment. Took: 6238

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
jiajunwang: That will break some logic, there is no real hashcode logic for replica. Here the code just serves for ordering. If I add it to the hashcode() it will impact the maps that use AssignableRepilca as the key.

Regarding 142, since it is not computing-intensive, it seems not improved a lot when I tried. So I just don't do it.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
jiajunwang: I like the compareTo method with Integer.

i3wangyi: Integer class will possibly throw NPE and that's generally not a good idea using it .

jiajunwang: The fact is that I need to use the map for avoiding duplicate calculating. If I use map, I have to put Integer inside. If that is null, we will be seeing NPE no matter we are using Integer or int in this logic.
This is a private method, and inside this class, we know the value won't be null.

i3wangyi:
The fact of time complexity doesn't change: O(# of replicas * # nodes) and we're assigning each replica one by one. However, it may not be true (2 replicas assigned to 2 different nodes don't conflict with each other and change the cluster context). The time complexity can be improved to 
O( # of replicas / avg batch assign size  * # nodes)

Another idea to boost the performance:
```
while unassignedReplicas > 0:
    # parallel
    list of [node, replica] proposed assginments =  parallelComputeSoftConstraints(nodes, unassignedReplicas) 
   list of [node, replica] unconfliced assignments = filter(proposed assignments)
   applu(unconfliced assignments)
```

jiajunwang:
> The fact of time complexity doesn't change: O(# of replicas * # nodes) and we're assigning each replica one by one. However, it may not be true (2 replicas assigned to 2 different nodes don't conflict with each other and change the cluster context). The time complexity can be improved to
> O( # of replicas / avg batch assign size * # nodes)
> 
> Another idea to boost the performance:
> 
> ```
> while unassignedReplicas > 0:
>     # parallel
>     list of [node, replica] proposed assginments =  parallelComputeSoftConstraints(nodes, unassignedReplicas) 
>    list of [node, replica] unconfliced assignments = filter(proposed assignments)
>    applu(unconfliced assignments)
> ```

I tied. But to my surprise, the time spent was not reduced obviously. It would be a little bit faster, but the logic would be much more complicated. So I think better to postpone this change since the other plan will have a larger return.
For this PR, I don't want to change the logic.
I will have another PR for the improvement that does change the logic.

i3wangyi:
> > The fact of time complexity doesn't change: O(# of replicas * # nodes) and we're assigning each replica one by one. However, it may not be true (2 replicas assigned to 2 different nodes don't conflict with each other and change the cluster context). The time complexity can be improved to
> > O( # of replicas / avg batch assign size * # nodes)
> > Another idea to boost the performance:
> > ```
> > while unassignedReplicas > 0:
> >     # parallel
> >     list of [node, replica] proposed assginments =  parallelComputeSoftConstraints(nodes, unassignedReplicas) 
> >    list of [node, replica] unconfliced assignments = filter(proposed assignments)
> >    applu(unconfliced assignments)
> > ```
> 
> I tied. But to my surprise, the time spent was not reduced obviously. It would be a little bit faster, but the logic would be much more complicated. So I think better to postpone this change since the other plan will have a larger return.
> For this PR, I don't want to change the logic.
> I will have another PR for the improvement that does change the logic.

Sounds good. I agree the code change of the plan will be a lot given the current structure. I guess the reason it doesn't improve much is that many assignments are overlapped (many replicas want to assign to the same node based on the greedy properties).

jiajunwang:
This PR is ready to be merged, approved by @i3wangyi and @narendly 

Description:

This change improves the rebalance's speed by 2x to 5x depends on the host capacity.
1. Parallelism the loop processing whenever possible and help to improve the performance. This does not change the logic.
2. Avoid some duplicate logic in the loop. Put the calculation outside the loop and only do it once..

PR #601, opened 2019-11-13
Add new constructors and close() to Helix API

### LINKED ISSUE #593: Update close() for ZkHelixAdmin
close() needs to be updated so it doesn't close external ZkClient

### LINKED ISSUE #594: Update ZkCacheBaseDataAccessor's stop()
stop() needs to be updated to close baseDataAccessor as well when it's been initialized with a baseDataAccessor

### LINKED ISSUE #595: Add close() to ZkHelixClusterVerifier
Need to add close() in order to shut down the internal ZKClient to prevent leaks.

### LINKED ISSUE #596: Update ZkBaseDataAccessor's close()
Update close() so that the logic distinguishes an internal ZkClient and an external ZkClient. We shouldn't close an external ZkClient.

narendly:
### Issues

Resolves #593 
Resolves #594 
Resolves #595 
Resolves #596 

### Description

In order to allow users to avoid having to create, manage, and close a HelixZkClient instance when using Helix data access APIs, we add new constructors and close() methods. The new constructors will allow for instantiation purely based on a ZkAddress, and internally, either Shared or Dedicated HelixZkClient will be used.

Also, the close() method ensures that we only close if we generated a ZkClient internally. External ZkClients, we can't close arbitrarily because they could be used in other active Helix APIs.

### Tests

No additional tests required - no new logic added.

[ERROR] Failures:
[INFO]   Run 4: PASS
[INFO]   Run 5: PASS
[INFO]
[INFO]
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  57:28 min
[INFO] Finished at: 2019-11-14T19:44:05-08:00

------
Two tests pass when run individually.

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -90,12 +90,16 @@ public class ZKHelixAdmin implements HelixAdmin {
 
   private final HelixZkClient _zkClient;
   private final ConfigAccessor _configAccessor;
+  // true if ZKHelixAdmin was instantiated with a HelixZkClient, false otherwise
+  // This is used for close() to determine how ZKHelixAdmin should close the underlying ZkClient
+  private boolean _usesExternalZkClient;
 
   private static Logger logger = LoggerFactory.getLogger(ZKHelixAdmin.class);
 
   public ZKHelixAdmin(HelixZkClient zkClient) {
     _zkClient = zkClient;
     _configAccessor = new ConfigAccessor(zkClient);
+    _usesExternalZkClient = true;
   }
 
   public ZKHelixAdmin(String zkAddress) {
@@ -107,6 +111,7 @@ public class ZKHelixAdmin implements HelixAdmin {
         .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress), clientConfig);
     _zkClient.waitUntilConnected(timeOutInSec, TimeUnit.SECONDS);
     _configAccessor = new ConfigAccessor(_zkClient);
+    _usesExternalZkClient = false;
   }
 
   @Override
@@ -1571,11 +1576,13 @@ public class ZKHelixAdmin implements HelixAdmin {
     return instances;
   }
 
+  /**
+   * Closes the ZkClient only if it was generated internally.
+   */
   @Override
   public void close() {
-    if (_zkClient != null) {
+    if (_zkClient != null && !_usesExternalZkClient) {
       _zkClient.close();
     }
   }
-
 }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -92,10 +92,11 @@ public class ZKHelixAdmin implements HelixAdmin {
   private final ConfigAccessor _configAccessor;
   // true if ZKHelixAdmin was instantiated with a HelixZkClient, false otherwise
   // This is used for close() to determine how ZKHelixAdmin should close the underlying ZkClient
-  private boolean _usesExternalZkClient;
+  private final boolean _usesExternalZkClient;
 
   private static Logger logger = LoggerFactory.getLogger(ZKHelixAdmin.class);
 
+  @Deprecated
   public ZKHelixAdmin(HelixZkClient zkClient) {
     _zkClient = zkClient;
     _configAccessor = new ConfigAccessor(zkClient);
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -105,9 +105,10 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    * @param zkAddress The zookeeper address
    */
   public ZkBaseDataAccessor(String zkAddress, ZkSerializer zkSerializer) {
-    _zkClient = SharedZkClientFactory.getInstance()
-        .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
-            new HelixZkClient.ZkClientConfig().setZkSerializer(zkSerializer));
+    _zkClient = SharedZkClientFactory.getInstance().buildZkClient(
+        new HelixZkClient.ZkConnectionConfig(zkAddress),
+        new HelixZkClient.ZkClientConfig().setZkSerializer(zkSerializer));
+    _usesExternalZkClient = false;
   }
 
   /**
@@ -1147,7 +1148,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    */
   @Override
   public void close() {
-    if (_zkClient != null) {
+    if (_zkClient != null && !_usesExternalZkClient) {
       _zkClient.close();
     }
   }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -87,12 +87,17 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
   private static Logger LOG = LoggerFactory.getLogger(ZkBaseDataAccessor.class);
 
   private final HelixZkClient _zkClient;
+  // true if ZkBaseDataAccessor was instantiated with a HelixZkClient, false otherwise
+  // This is used for close() to determine how ZkBaseDataAccessor should close the underlying
+  // ZkClient
+  private boolean _usesExternalZkClient;
 
   public ZkBaseDataAccessor(HelixZkClient zkClient) {
     if (zkClient == null) {
       throw new NullPointerException("zkclient is null");
     }
     _zkClient = zkClient;
+    _usesExternalZkClient = true;
   }
 
   /**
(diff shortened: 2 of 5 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -90,12 +90,16 @@
 
   private final HelixZkClient _zkClient;
   private final ConfigAccessor _configAccessor;
+  // true if ZKHelixAdmin was instantiated with a HelixZkClient, false otherwise
+  // This is used for close() to determine how ZKHelixAdmin should close the underlying ZkClient
+  private boolean _usesExternalZkClient;
```
jiajunwang: final

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -87,22 +87,28 @@ public AccessResult() {
   private static Logger LOG = LoggerFactory.getLogger(ZkBaseDataAccessor.class);
 
   private final HelixZkClient _zkClient;
+  // true if ZkBaseDataAccessor was instantiated with a HelixZkClient, false otherwise
+  // This is used for close() to determine how ZkBaseDataAccessor should close the underlying
+  // ZkClient
+  private boolean _usesExternalZkClient;
```
jiajunwang: same, final

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
```diff
@@ -48,6 +48,10 @@
   protected static int DEFAULT_PERIOD = 500;
 
   protected final HelixZkClient _zkClient;
+  // true if ZkHelixClusterVerifier was instantiated with a HelixZkClient, false otherwise
+  // This is used for close() to determine how ZkHelixClusterVerifier should close the underlying
+  // ZkClient
+  private boolean _usesExternalZkClient;
```
jiajunwang: final

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
narendly: Added

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
narendly: Added

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
narendly: Added

i3wangyi:
The strange flag `usedExternalZkClient` will need to be added to each and every accessor that wraps the ZkClient by your proposal, right? 
My idea is to leave a clean implementation for all the `close` method in those accessors, create a boolean field inside ZkClient `isShared` something. As long as it's created by SharedZkClientFactory it will be true, and the close method will respect the boolean to perform the actual close action.

narendly:
> The strange flag `usedExternalZkClient` will need to be added to each and every accessor that wraps the ZkClient by your proposal, right?
> My idea is to leave a clean implementation for all the `close` method in those accessors, create a boolean field inside ZkClient `isShared` something. As long as it's created by SharedZkClientFactory it will be true, and the close method will respect the boolean to perform the actual close action.

It's more about whether ZkClient we're using is external or not. Whether it's a SharedZkClient or not has little to do with this.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -90,12 +90,16 @@
...
 
   private static Logger logger = LoggerFactory.getLogger(ZKHelixAdmin.class);
 
   public ZKHelixAdmin(HelixZkClient zkClient) {
```
lei-xia: Please mark this constructor as @Deprecated.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -87,22 +87,28 @@ public AccessResult() {
...
 
   public ZkBaseDataAccessor(HelixZkClient zkClient) {
```
lei-xia: @Deprecated

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
lei-xia: this can be  final too.

i3wangyi:
> > The strange flag `usedExternalZkClient` will need to be added to each and every accessor that wraps the ZkClient by your proposal, right?
> > My idea is to leave a clean implementation for all the `close` method in those accessors, create a boolean field inside ZkClient `isShared` something. As long as it's created by SharedZkClientFactory it will be true, and the close method will respect the boolean to perform the actual close action.
> 
> It's more about whether ZkClient we're using is external or not. Whether it's a SharedZkClient or not has little to do with this.

Then the scope of PR is to remove zkConnection leak by avoiding clients maintain their own external zkClients cause we force them to use the raw ZkAddress constructor.
Could you aggregate all the possible "leakable" constructors and mark them as @Deprecated at once and add the boolean flag for all of them?

narendly:
> > > The strange flag `usedExternalZkClient` will need to be added to each and every accessor that wraps the ZkClient by your proposal, right?
> > > My idea is to leave a clean implementation for all the `close` method in those accessors, create a boolean field inside ZkClient `isShared` something. As long as it's created by SharedZkClientFactory it will be true, and the close method will respect the boolean to perform the actual close action.
> > 
> > 
> > It's more about whether ZkClient we're using is external or not. Whether it's a SharedZkClient or not has little to do with this.
> 
> Then the scope of PR is to remove zkConnection leak by avoiding clients maintain their own external zkClients cause we force them to use the raw ZkAddress constructor.
> Could you aggregate all the possible "leakable" constructors and mark them as @deprecated at once and add the boolean flag for all of them?

OK.

i3wangyi:
> > > > The strange flag `usedExternalZkClient` will need to be added to each and every accessor that wraps the ZkClient by your proposal, right?
> > > > My idea is to leave a clean implementation for all the `close` method in those accessors, create a boolean field inside ZkClient `isShared` something. As long as it's created by SharedZkClientFactory it will be true, and the close method will respect the boolean to perform the actual close action.
> > > 
> > > 
> > > It's more about whether ZkClient we're using is external or not. Whether it's a SharedZkClient or not has little to do with this.
> > 
> > 
> > Then the scope of PR is to remove zkConnection leak by avoiding clients maintain their own external zkClients cause we force them to use the raw ZkAddress constructor.
> > Could you aggregate all the possible "leakable" constructors and mark them as @deprecated at once and add the boolean flag for all of them?
> 
> OK.

Notice a lot of codes are duplicated, what do you think of adding an `AbstractAccessor` or `CloseableAccessor` class for DRY? I feel even it's safe to do even with the current implementation.

narendly:
> > > > > The strange flag `usedExternalZkClient` will need to be added to each and every accessor that wraps the ZkClient by your proposal, right?
> > > > > My idea is to leave a clean implementation for all the `close` method in those accessors, create a boolean field inside ZkClient `isShared` something. As long as it's created by SharedZkClientFactory it will be true, and the close method will respect the boolean to perform the actual close action.
> > > > 
> > > > 
> > > > It's more about whether ZkClient we're using is external or not. Whether it's a SharedZkClient or not has little to do with this.
> > > 
> > > 
> > > Then the scope of PR is to remove zkConnection leak by avoiding clients maintain their own external zkClients cause we force them to use the raw ZkAddress constructor.
> > > Could you aggregate all the possible "leakable" constructors and mark them as @deprecated at once and add the boolean flag for all of them?
> > 
> > 
> > OK.
> 
> Notice a lot of codes are duplicated, what do you think of adding an `AbstractAccessor` or `CloseableAccessor` class for DRY? I feel even it's safe to do even with the current implementation.

I appreciate the idea, but each accessor class has a different implementation/logic for initialization and close() method, so if we do add something like that, it would be for the sake of adding a superclass, not to reduce the amount of duplicate code. If we want to do a larger-scale redesign, then we need to think about this more carefully because these are public APIs that are already being used in production. So let's hold that thought since it's not a must-have.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
narendly: Done

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
narendly: Done

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
```diff
@@ -188,6 +194,12 @@ public boolean verifyByPolling() {
     return verifyByPolling(DEFAULT_TIMEOUT, DEFAULT_PERIOD);
   }
 
+  public void close() {
```
jiajunwang: nit, do we need to change any test cases to call close? Or they are all using the constructor with a HelixZkClient input?

narendly: I am running mvn test to be sure :)
But this is a new method, so it shouldn't affect existing tests.

narendly: mvn test passed.

PR #606, opened 2019-11-15
Modify ZkUtil methods so that they accept ZkAddress as parameter

### LINKED ISSUE #605: Replace methods in ZkUtil so that it does not require HelixZkClient
Replace methods in ZkUtil so that it does not require HelixZkClient

This is so that the users of ZkUtil do not have to create HelixZkClient to use the utility methods.

narendly:
### Issues

Resolves #605 

### Description

Replace methods in ZkUtil so that it does not require HelixZkClient

This is so that the users of ZkUtil do not have to create HelixZkClient to use the utility methods.

### Tests

Covered by TestZKUtil.java

[INFO] Results:
[INFO] 
[ERROR] Failures: 
[INFO]   Run 4: PASS
[INFO]   Run 5: PASS
[INFO] 
[INFO] 
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  56:42 min
[INFO] Finished at: 2019-11-15T00:20:58-08:00

-----------------
Both pass when run individually.

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
@@ -36,6 +36,10 @@ import org.slf4j.LoggerFactory;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.data.Stat;
 
+/**
+ * Using this ZKUtil class for production purposes is NOT recommended since a lot of the static
+ * methods require a ZkClient instance to be passed in.
+ */
 public final class ZKUtil {
   private static Logger logger = LoggerFactory.getLogger(ZKUtil.class);
   private static int RETRYLIMIT = 3;
@@ -43,6 +47,13 @@ public final class ZKUtil {
   private ZKUtil() {
   }
 
+  /**
+   * Note: this method will create a dedicated ZkClient on the fly. Creating and closing a
+   * ZkConnection is a costly operation - use it at your own risk!
+   * @param clusterName
+   * @param zkAddress
+   * @return
+   */
   public static boolean isClusterSetup(String clusterName, String zkAddress) {
     HelixZkClient zkClient = getHelixZkClient(zkAddress);
     boolean result = isClusterSetup(clusterName, zkClient);
@@ -50,7 +61,6 @@ public final class ZKUtil {
     return result;
   }
 
-  @Deprecated
   public static boolean isClusterSetup(String clusterName, HelixZkClient zkClient) {
     if (clusterName == null) {
       logger.info("Fail to check cluster setup : cluster name is null!");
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
```diff
@@ -42,6 +43,14 @@
   private ZKUtil() {
   }
 
+  public static boolean isClusterSetup(String clusterName, String zkAddress) {
+    HelixZkClient zkClient = getHelixZkClient(zkAddress);
+    boolean result = isClusterSetup(clusterName, zkClient);
+    zkClient.close();
```
dasahcc: Just a throught, shall we have a static ZKClient map instead of creating/closing for each request? We can close all of them with a close() function? Or setup a timeout for idle usage?

narendly: Another approach is to have a static dedicated ZK client. But I made the design choice to create a ZK client and closing at the end of the method for the following reasons:
1. ZKUtil's methods are static and therefore should be stateless.
2. The usages for ZKUtil should generally be a one-time thing.
3. It is not desirable to have a static connection running at all times; you'd never know when the connection should be closed.

This is a tradeoff I made considering the points above. Of course, the drawback of this approach is that creating a dedicated connection for each static method call is not a lightweight operation.

What do people think? @lei-xia @jiajunwang ?

dasahcc: I buy the idea of one time access of different ZKs. But I think we should do following things:
1. Make clear Java doc to let user understand it could cause more resource usage (ZKClient open/close).
2. We should not deprecate the old API as long as that can provide user an option to now open/close ZKClient frequently while they would like to use the similar functionalities.

Also we need to clearly define the role of this Util and follow the rule for future development.

narendly: @dasahcc 

Those are reasonable points. What I'll do is that 
1. I will remove @deprecated annotation from existing methods.
2. I will add in the Javadoc for newly added methods saying that it will create a ZK connection every time. 
3. In general, ZKUtil should be used for testing and Helix internal uses only. I will clarify this in the Javadoc for the class.

kaisun2000: > I buy the idea of one time access of different ZKs. But I think we should do following things:
> 
> 1. Make clear Java doc to let user understand it could cause more resource usage (ZKClient open/close).
> 2. We should not deprecate the old API as long as that can provide user an option to now open/close ZKClient frequently while they would like to use the similar functionalities.
> 
> Also we need to clearly define the role of this Util and follow the rule for future development.

@dasahcc, actually we are targeting to hide ZkClient from Helix powered applications. So that later they don't see them anymore. Instead, they see HelixManager, HelixAdmin, ConfigAccessor, HelixDataAccessor, HelixPropertyStore, HelixVerifiyer and the last one is ZkUtil.

narendly: @dasahcc Your points have been addressed.

narendly:


```
Replace methods in ZkUtil so that it does not require HelixZkClient

This is so that the users of ZkUtil do not have to create HelixZkClient to use the utility methods.
```

PR #630, opened 2019-11-22
New API to create an ephemeral node that is session aware

pkuwm:
* Creates an ephemeral node that is session aware.
* If the ephemeral node is created successfully, returns true.
* If session is expired, returns false.
* If connection is timed out or interrupted, exception is thrown.

I will complete the descriptions once the code itself is agreed. We will focus on the code logic in this draft.

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -569,6 +569,71 @@ public class ZkClient implements Watcher {
   }
 
   /**
+   * Creates a session aware ephemeral node.
+   * If the ephemeral node is created successfully, returns true.
+   * If session is expired, returns false.
+   * If connection is timed out or interrupted, exception is thrown.
+   *
+   * @param path path of the ephemeral node being created
+   * @param data data of the ephemeral node being created
+   * @param sessionId session id of the zookeeper client
+   * @return <code>true</code> if the input session is not expired and the ephemeral node is created
+   *         successfully; <code>false</code> if the input session is expired.
+   * @throws ZkInterruptedException
+   *           if operation is interrupted, or a required reconnection gets interrupted
+   * @throws IllegalArgumentException
+   *           if called from anything except the ZooKeeper event thread
+   * @throws ZkException
+   *           if any ZooKeeper exception occurs
+   * @throws RuntimeException
+   *           if any other exception occurs
+   */
+  public boolean createEphemeral(final String path, final Object data, final long sessionId)
+      throws ZkInterruptedException, IllegalArgumentException, ZkException, RuntimeException {
+    if (path == null) {
+      throw new NullPointerException("Path must not be null.");
+    }
+
+    long startTime = System.currentTimeMillis();
+    try {
+      final byte[] serializedData = (data == null ? null : serialize(data, path));
+      checkDataSizeLimit(serializedData);
+
+      boolean successful = retryUntilConnected(() -> {
+        // Acquire zk event lock to make sure no new session is established,
+        // while the session is being checked and the ephemeral node is being created.
+        acquireEventLock();
+        try {
+          // Check if current session is still the same with the input session.
+          // If not, it means the input session is expired, we will not create an ephemeral node,
+          // and return false.
+          if (getSessionId() != sessionId) {
+            return false;
+          }
+          // Call zookeeper to create an ephemeral node.
+          getConnection()
+              .create(path, serializedData, ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL);
+          return true;
+        } finally {
+          getEventLock().unlock();
+        }
+      });
+
+      if (!successful) {
+        recordFailure(path, ZkClientMonitor.AccessType.WRITE);
+        return false;
+      }
+      record(path, serializedData, startTime, ZkClientMonitor.AccessType.WRITE);
+      return true;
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -568,6 +568,71 @@ public void createEphemeral(final String path, final Object data)
     create(path, data, CreateMode.EPHEMERAL);
   }
 
+  /**
+   * Creates a session aware ephemeral node.
```
narendly: Please explain what "session aware" means here as well?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -568,6 +568,71 @@ public void createEphemeral(final String path, final Object data)
...
+   * @throws ZkInterruptedException
+   *           if operation is interrupted, or a required reconnection gets interrupted
+   * @throws IllegalArgumentException
+   *           if called from anything except the ZooKeeper event thread
+   * @throws ZkException
+   *           if any ZooKeeper exception occurs
+   * @throws RuntimeException
+   *           if any other exception occurs
+   */
+  public boolean createEphemeral(final String path, final Object data, final long sessionId)
```
pkuwm: Lei's comment: 
```
lei-xia 2 minutes ago
Many of the code in this method is duplicated with the one in create() (line 522). Another option is to extend that method to take a sessionID (also wrap the content of that method as a new private method to avoid expose it to be public API)?
```

My reply: I tried to reduce the duplicate code. There are 2 things to consider:
1. Because the input data is an object, we need to serialize it. If so, we need to change the api parameters like:
create(final String path, `byte[] data`, final List<ACL> acl, final CreateMode mode, Callable<String> callable);
And we will need to change create() implementation, whether or not we want to do?
2. Object data needs to be serialized and passed to the final create().
I will show you the change for the above idea and see if we could accept it.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
pkuwm: Yes, will do.

kaisun2000: "Creates an ephemeral node that is session aware.
If the ephemeral node is created successfully, returns true.
If session is expired, returns false.
If connection is timed out or interrupted, exception is thrown."

The description is not very accurate. The gist is that given a session id, we need to make sure the created ephemeral node must be in this session. Otherwise, the creation should fail. 
ZkClient has a auto-reconnect feature. This feature makes this create ephemeral node in this session operation not that easy. 
We need handle session expired, disconnected state change. In additional close(). 

So all in all the following semantic should be true. 

"
given a session ID, the api should return true it can create an ephemeral node in this session.
If met irrecoverable condition, the API should return false.
If met recoverable  condition, the API should retry and create the ephemeral node.
"

Here, irrecoverable condition can be: session expired, new session established, zkclient closed, interrrupted, keeper exceptions such as invalid parameters etc.

Recoverable condition can be: session disconnect. (Note, session can still be reconnected with the same session id potentially later.)

pkuwm:
Lei's comments: https://github.com/pkuwm/helix/commit/f6feb711c36c88994dc2abc6f1c021de59ff2ed3

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
pkuwm: Right, thanks for the meaningful description. This is what I am planning to complete. Once the code itself is agreed by us, I will complete the description.

lei-xia:
Can we have only 1 PR and keep all comments in one place.  It is  a bit confusing here.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
pkuwm: @lei-xia 's comment:
`
Give a more clear description what does this extra sessionId in parameters mean, such as, this method allows you to specify an expected sessionId, if the current sessionId of underlying ZK connection does not match the given sessionId, the creation will fail ....
`

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -568,6 +568,71 @@ public void createEphemeral(final String path, final Object data)
...
 
+  /**
+   * Creates a session aware ephemeral node.
+   * If the ephemeral node is created successfully, returns true.
+   * If session is expired, returns false.
+   * If connection is timed out or interrupted, exception is thrown.
+   *
+   * @param path path of the ephemeral node being created
+   * @param data data of the ephemeral node being created
+   * @param sessionId session id of the zookeeper client
```
pkuwm: Lei @lei-xia 
`
This is should be the expected session Id of the ZK connection, not necessarily the sessionID of current ZK Connection, rephrase it to make it less confusing. And also make it clear that if the sessionID of current ZK connection does not match the given (expected) sessionId, the operation will fail.
`

pkuwm:
@dasahcc In this PR, I don’t move the code retryUntilConnect to
createEmphemeNode, instead I reuse retryUntilConnect to minimize the
changes.

PR #632, opened 2019-11-26
Asynchronously calculating the Baseline

jiajunwang:
### Issues

#563 

### Description

Enable the Baseline calculation to be asynchronously done.
This will greatly fasten the rebalance speed. Basically, the WAGED rebalancer will firstly partial rebalance to recover the invalid replica allocations (for example, the ones that are on a disabled instance). Then it calculates the new baseline by global rebalancing.

### Tests

Most test cases are still valid to cover this workflow modification. Modify the following two since the assumption is different due to synchronously calculation.
TestMixedModeAutoRebalance
TestMixedModeAutoWagedRebalance

[WARNING] Tests run: 1058, Failures: 0, Errors: 0, Skipped: 3, Time elapsed: 4,126.389 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[WARNING] Tests run: 1058, Failures: 0, Errors: 0, Skipped: 3
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 01:08 h
[INFO] Finished at: 2019-12-19T14:41:07-08:00

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -92,11 +92,14 @@ public class AssignmentMetadataStore {
     return _bestPossibleAssignment;
   }
 
-  public void persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
+  /**
+   * @return true if a new baseline was persisted.
+   */
+  public boolean persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
     // TODO: Make the write async?
     // If baseline hasn't changed, skip writing to metadata store
     if (compareAssignments(_globalBaseline, globalBaseline)) {
-      return;
+      return false;
     }
     // Persist to ZK
     HelixProperty combinedAssignments = combineAssignments(BASELINE_KEY, globalBaseline);
@@ -109,14 +112,18 @@ public class AssignmentMetadataStore {
 
     // Update the in-memory reference
     _globalBaseline = globalBaseline;
+    return true;
   }
 
-  public void persistBestPossibleAssignment(
+  /**
+   * @return true if a new best possible assignment was persisted.
+   */
+  public boolean persistBestPossibleAssignment(
       Map<String, ResourceAssignment> bestPossibleAssignment) {
     // TODO: Make the write async?
     // If bestPossibleAssignment hasn't changed, skip writing to metadata store
     if (compareAssignments(_bestPossibleAssignment, bestPossibleAssignment)) {
-      return;
+      return false;
     }
     // Persist to ZK
     HelixProperty combinedAssignments =
@@ -130,6 +137,7 @@ public class AssignmentMetadataStore {
 
     // Update the in-memory reference
     _bestPossibleAssignment = bestPossibleAssignment;
+    return true;
   }
 
   protected void finalize() {
@@ -179,7 +187,7 @@ public class AssignmentMetadataStore {
    * @param newAssignment
    * @return true if they are the same. False otherwise or oldAssignment is null
    */
-  private boolean compareAssignments(Map<String, ResourceAssignment> oldAssignment,
+  protected boolean compareAssignments(Map<String, ResourceAssignment> oldAssignment,
       Map<String, ResourceAssignment> newAssignment) {
     // If oldAssignment is null, that means that we haven't read from/written to
     // the metadata store yet. In that case, we return false so that we write to metadata store.
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -94,7 +94,10 @@ public class AssignmentMetadataStore {
 
   /**
    * @return true if a new baseline was persisted.
+   * @throws HelixException if the method failed to persist the baseline.
    */
+  // TODO: Enhance the return value so it is more intuitive to understand when the persist fails and
+  // TODO: when it is skipped.
   public boolean persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
     // TODO: Make the write async?
     // If baseline hasn't changed, skip writing to metadata store
@@ -117,7 +120,10 @@ public class AssignmentMetadataStore {
 
   /**
    * @return true if a new best possible assignment was persisted.
+   * @throws HelixException if the method failed to persist the baseline.
    */
+  // TODO: Enhance the return value so it is more intuitive to understand when the persist fails and
+  // TODO: when it is skipped.
   public boolean persistBestPossibleAssignment(
       Map<String, ResourceAssignment> bestPossibleAssignment) {
     // TODO: Make the write async?
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
@@ -235,7 +235,7 @@ public class WagedRebalancer {
   // Release all the resources.
   public void close() {
     if (_baselineCalculateExecutor != null) {
-      _baselineCalculateExecutor.shutdown();
+      _baselineCalculateExecutor.shutdownNow();
     }
     if (_assignmentMetadataStore != null) {
       _assignmentMetadataStore.close();
@@ -498,7 +498,7 @@ public class WagedRebalancer {
     LOG.info("Global baseline calculation completed and has been persisted into metadata store.");
 
     if (isBaselineChanged && shouldSchedulePartialRebalance) {
-      LOG.info("Schedule a new rebalance after the new baseline calculated.");
+      LOG.info("Schedule a new rebalance after the new baseline calculation has finished.");
       RebalanceUtil.scheduleOnDemandPipeline(clusterName, 0L, false);
     }
   }
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
@@ -26,6 +26,10 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 
 import com.google.common.collect.ImmutableMap;
@@ -54,6 +58,7 @@ import org.apache.helix.monitoring.metrics.WagedRebalancerMetricCollector;
 import org.apache.helix.monitoring.metrics.implementation.BaselineDivergenceGauge;
 import org.apache.helix.monitoring.metrics.model.CountMetric;
 import org.apache.helix.monitoring.metrics.model.LatencyMetric;
+import org.apache.helix.util.RebalanceUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -79,6 +84,8 @@ public class WagedRebalancer {
       .of(ClusterConfig.GlobalRebalancePreferenceKey.EVENNESS, -1,
           ClusterConfig.GlobalRebalancePreferenceKey.LESS_MOVEMENT, -1);
 
+  // To calculate the baseline asynchronously
+  private final ExecutorService _baselineCalculateExecutor;
   private final ResourceChangeDetector _changeDetector;
   private final HelixManager _manager;
   private final MappingCalculator<ResourceControllerDataProvider> _mappingCalculator;
@@ -93,6 +100,8 @@ public class WagedRebalancer {
   private final LatencyMetric _partialRebalanceLatency;
   private final LatencyMetric _stateReadLatency;
   private final BaselineDivergenceGauge _baselineDivergenceGauge;
+
+  private boolean _asyncBaselineCalculation;
 
   // Note, the rebalance algorithm field is mutable so it should not be directly referred except for
   // the public method computeNewIdealStates.
@@ -109,7 +118,8 @@ public class WagedRebalancer {
   }
 
   public WagedRebalancer(HelixManager helixManager,
-      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preference) {
+      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preference,
+      boolean isAsyncGlobalRebalance) {
     this(helixManager == null ? null
             : constructAssignmentStore(helixManager.getMetadataStoreConnectionString(),
                 helixManager.getClusterName()), ConstraintBasedAlgorithmFactory.getInstance(preference),
@@ -127,7 +137,8 @@ public class WagedRebalancer {
         // CurrentState-based rebalancing. 2. Tests that require instrumenting the rebalancer for
         // verifying whether the cluster has converged.
         helixManager == null ? null
-            : new WagedRebalancerMetricCollector(helixManager.getClusterName()));
+            : new WagedRebalancerMetricCollector(helixManager.getClusterName()),
+        isAsyncGlobalRebalance);
     _preference = ImmutableMap.copyOf(preference);
   }
 
@@ -140,7 +151,7 @@ public class WagedRebalancer {
    */
   protected WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
       RebalanceAlgorithm algorithm) {
-    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer(), null, null);
+    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer(), null, null, false);
   }
 
   /**
@@ -152,12 +163,13 @@ public class WagedRebalancer {
    */
   protected WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
       RebalanceAlgorithm algorithm, MetricCollector metricCollector) {
-    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer(), null, metricCollector);
+    this(assignmentMetadataStore, algorithm, new DelayedAutoRebalancer(), null, metricCollector,
+        false);
   }
 
   private WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
       RebalanceAlgorithm algorithm, MappingCalculator mappingCalculator, HelixManager manager,
-      MetricCollector metricCollector) {
+      MetricCollector metricCollector, boolean asyncBaselineCalculation) {
     if (assignmentMetadataStore == null) {
       LOG.warn("Assignment Metadata Store is not configured properly."
           + " The rebalancer will not access the assignment store during the rebalance.");
@@ -199,23 +211,30 @@ public class WagedRebalancer {
         BaselineDivergenceGauge.class);
 
     _changeDetector = new ResourceChangeDetector(true);
+
+    _baselineCalculateExecutor = Executors.newSingleThreadExecutor();
+    _asyncBaselineCalculation = asyncBaselineCalculation;
   }
 
-  // Update the rebalancer preference configuration if the new preference is different from the
-  // current preference configuration.
-  public void updatePreference(
-      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> newPreference) {
-    if (_preference.equals(NOT_CONFIGURED_PREFERENCE) || _preference.equals(newPreference)) {
-      // 1. if the preference was not configured during constructing, no need to update.
-      // 2. if the preference equals to the new preference, no need to update.
-      return;
+  // Update the rebalancer configuration if the new options are different from the current
+  // configuration.
+  public synchronized void updateRebalanceOptions(
+      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> newPreference,
+      boolean asyncBaselineCalculation) {
+    _asyncBaselineCalculation = asyncBaselineCalculation;
+    // 1. if the preference was not configured during constructing, no need to update.
+    // 2. if the preference equals to the new preference, no need to update.
+    if (!_preference.equals(NOT_CONFIGURED_PREFERENCE) && !_preference.equals(newPreference)) {
+      _rebalanceAlgorithm = ConstraintBasedAlgorithmFactory.getInstance(newPreference);
+      _preference = ImmutableMap.copyOf(newPreference);
     }
-    _rebalanceAlgorithm = ConstraintBasedAlgorithmFactory.getInstance(newPreference);
-    _preference = ImmutableMap.copyOf(newPreference);
   }
 
   // Release all the resources.
   public void close() {
+    if (_baselineCalculateExecutor != null) {
+      _baselineCalculateExecutor.shutdown();
+    }
     if (_assignmentMetadataStore != null) {
       _assignmentMetadataStore.close();
     }
@@ -413,28 +432,57 @@ public class WagedRebalancer {
             HelixRebalanceException.Type.INVALID_CLUSTER_STATUS, ex);
       }
 
-      calculateAndUpdateBaseline(clusterModel, algorithm);
+      final boolean waitForGlobalRebalance = !_asyncBaselineCalculation;
+      final String clusterName = clusterData.getClusterName();
+      // Calculate the Baseline assignment for global rebalance.
+      Future<Boolean> result = _baselineCalculateExecutor.submit(() -> {
+        try {
+          // Note that we should schedule a new partial rebalance if the following calculation does
+          // not wait until the new baseline is calculated.
+          // So set doSchedulePartialRebalance to be !waitForGlobalRebalance
+          calculateAndUpdateBaseline(clusterModel, algorithm, !waitForGlobalRebalance, clusterName);
+        } catch (HelixRebalanceException e) {
+          LOG.error("Failed to calculate baseline assignment!", e);
+          return false;
+        }
+        return true;
+      });
+      if (waitForGlobalRebalance) {
+        try {
+          if (!result.get()) {
+            throw new HelixRebalanceException("Failed to calculate for the new Baseline.",
+                HelixRebalanceException.Type.FAILED_TO_CALCULATE);
+          }
+        } catch (InterruptedException | ExecutionException e) {
+          throw new HelixRebalanceException("Failed to execute new Baseline calculation.",
+              HelixRebalanceException.Type.FAILED_TO_CALCULATE, e);
+        }
+      }
     }
   }
 
   /**
    * Calculate and update the Baseline assignment
    * @param clusterModel
-   * @param algorithm
+   * @param doSchedulePartialRebalance True if the call should trigger a following partial rebalance
+   *                                   so the new Baseline could be applied to cluster.
+   * @param clusterName
    * @throws HelixRebalanceException
    */
-  private void calculateAndUpdateBaseline(ClusterModel clusterModel, RebalanceAlgorithm algorithm)
+  private void calculateAndUpdateBaseline(ClusterModel clusterModel, RebalanceAlgorithm algorithm,
+      boolean doSchedulePartialRebalance, String clusterName)
       throws HelixRebalanceException {
     LOG.info("Start calculating the new baseline.");
     _globalBaselineCalcCounter.increment(1L);
     _globalBaselineCalcLatency.startMeasuringLatency();
 
+    boolean isbaselineUpdated = false;
     Map<String, ResourceAssignment> newBaseline = calculateAssignment(clusterModel, algorithm);
     // Write the new baseline to metadata store
     if (_assignmentMetadataStore != null) {
       try {
         _writeLatency.startMeasuringLatency();
-        _assignmentMetadataStore.persistBaseline(newBaseline);
+        isbaselineUpdated = _assignmentMetadataStore.persistBaseline(newBaseline);
         _writeLatency.endMeasuringLatency();
       } catch (Exception ex) {
         throw new HelixRebalanceException("Failed to persist the new baseline assignment.",
@@ -445,6 +493,11 @@ public class WagedRebalancer {
     }
     _globalBaselineCalcLatency.endMeasuringLatency();
     LOG.info("Finish calculating the new baseline.");
+
+    if (isbaselineUpdated && doSchedulePartialRebalance) {
+      LOG.info("Schedule a new rebalance after the new baseline calculated.");
+      RebalanceUtil.scheduleOnDemandPipeline(clusterName, 0l, false);
+    }
   }
 
   private Map<String, ResourceAssignment> partialRebalance(
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
@@ -101,7 +101,7 @@ public class WagedRebalancer {
   private final LatencyMetric _stateReadLatency;
   private final BaselineDivergenceGauge _baselineDivergenceGauge;
 
-  private boolean _asyncGlobalRebalance;
+  private boolean _asyncGlobalRebalanceEnabled;
 
   // Note, the rebalance algorithm field is mutable so it should not be directly referred except for
   // the public method computeNewIdealStates.
@@ -119,7 +119,7 @@ public class WagedRebalancer {
 
   public WagedRebalancer(HelixManager helixManager,
       Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preference,
-      boolean isAsyncGlobalRebalance) {
+      boolean isAsyncGlobalRebalanceEnabled) {
     this(helixManager == null ? null
             : constructAssignmentStore(helixManager.getMetadataStoreConnectionString(),
                 helixManager.getClusterName()), ConstraintBasedAlgorithmFactory.getInstance(preference),
@@ -138,7 +138,7 @@ public class WagedRebalancer {
         // verifying whether the cluster has converged.
         helixManager == null ? null
             : new WagedRebalancerMetricCollector(helixManager.getClusterName()),
-        isAsyncGlobalRebalance);
+        isAsyncGlobalRebalanceEnabled);
     _preference = ImmutableMap.copyOf(preference);
   }
 
@@ -169,7 +169,7 @@ public class WagedRebalancer {
 
   private WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
       RebalanceAlgorithm algorithm, MappingCalculator mappingCalculator, HelixManager manager,
-      MetricCollector metricCollector, boolean asyncGlobalRebalance) {
+      MetricCollector metricCollector, boolean isAsyncGlobalRebalanceEnabled) {
     if (assignmentMetadataStore == null) {
       LOG.warn("Assignment Metadata Store is not configured properly."
           + " The rebalancer will not access the assignment store during the rebalance.");
@@ -213,17 +213,16 @@ public class WagedRebalancer {
     _changeDetector = new ResourceChangeDetector(true);
 
     _baselineCalculateExecutor = Executors.newSingleThreadExecutor();
-    _asyncGlobalRebalance = asyncGlobalRebalance;
+    _asyncGlobalRebalanceEnabled = isAsyncGlobalRebalanceEnabled;
   }
 
   // Update the global rebalance mode to be asynchronous or synchronous
-  public void setGlobalRebalanceAsyncMode(boolean asyncGlobalRebalance) {
-    _asyncGlobalRebalance = asyncGlobalRebalance;
+  public void setGlobalRebalanceAsyncMode(boolean isAsyncGlobalRebalanceEnabled) {
+    _asyncGlobalRebalanceEnabled = isAsyncGlobalRebalanceEnabled;
   }
 
-  // Update the rebalancer pereference if the new options are different from the current
-  // pereference.
-  public synchronized void updateRebalancePereference(
+  // Update the rebalancer preference if the new options are different from the current preference.
+  public synchronized void updateRebalancePreference(
       Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> newPreference) {
     // 1. if the preference was not configured during constructing, no need to update.
     // 2. if the preference equals to the new preference, no need to update.
@@ -435,14 +434,15 @@ public class WagedRebalancer {
             HelixRebalanceException.Type.INVALID_CLUSTER_STATUS, ex);
       }
 
-      final boolean waitForGlobalRebalance = !_asyncGlobalRebalance;
+      final boolean waitForGlobalRebalance = !_asyncGlobalRebalanceEnabled;
       final String clusterName = clusterData.getClusterName();
       // Calculate the Baseline assignment for global rebalance.
       Future<Boolean> result = _baselineCalculateExecutor.submit(() -> {
         try {
-          // Note that we should schedule a new partial rebalance if the following calculation does
-          // not wait until the new baseline is calculated.
-          // So set doSchedulePartialRebalance to be !waitForGlobalRebalance
+          // Note that we should schedule a new partial rebalance for a future rebalance pipeline if
+          // the planned partial rebalance in the current rebalance pipeline won't wait for the new
+          // baseline being calculated.
+          // So set shouldSchedulePartialRebalance to be !waitForGlobalRebalance
           calculateAndUpdateBaseline(clusterModel, algorithm, !waitForGlobalRebalance, clusterName);
         } catch (HelixRebalanceException e) {
           LOG.error("Failed to calculate baseline assignment!", e);
@@ -467,25 +467,25 @@ public class WagedRebalancer {
   /**
    * Calculate and update the Baseline assignment
    * @param clusterModel
-   * @param doSchedulePartialRebalance True if the call should trigger a following partial rebalance
+   * @param shouldSchedulePartialRebalance True if the call should trigger a following partial rebalance
    *                                   so the new Baseline could be applied to cluster.
    * @param clusterName
    * @throws HelixRebalanceException
    */
   private void calculateAndUpdateBaseline(ClusterModel clusterModel, RebalanceAlgorithm algorithm,
-      boolean doSchedulePartialRebalance, String clusterName)
+      boolean shouldSchedulePartialRebalance, String clusterName)
       throws HelixRebalanceException {
     LOG.info("Start calculating the new baseline.");
     _baselineCalcCounter.increment(1L);
     _baselineCalcLatency.startMeasuringLatency();
 
-    boolean isbaselineUpdated = false;
+    boolean isBaselineChanged = false;
     Map<String, ResourceAssignment> newBaseline = calculateAssignment(clusterModel, algorithm);
     // Write the new baseline to metadata store
     if (_assignmentMetadataStore != null) {
       try {
         _writeLatency.startMeasuringLatency();
-        isbaselineUpdated = _assignmentMetadataStore.persistBaseline(newBaseline);
+        isBaselineChanged = _assignmentMetadataStore.persistBaseline(newBaseline);
         _writeLatency.endMeasuringLatency();
       } catch (Exception ex) {
         throw new HelixRebalanceException("Failed to persist the new baseline assignment.",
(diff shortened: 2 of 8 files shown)


helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -121,6 +126,7 @@
           .put(GlobalRebalancePreferenceKey.LESS_MOVEMENT, 1).build();
   private final static int MAX_REBALANCE_PREFERENCE = 10;
   private final static int MIN_REBALANCE_PREFERENCE = 0;
+  private final static boolean DEFAULT_ASYNC_GLOBAL_REBALANCE = true;
```
narendly: DEFAULT_GLOBAL_REBALANCE_ASYNC_(MODE)_ENABLED? It reads better when the value is a boolean.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -822,6 +828,19 @@ public void setGlobalRebalancePreference(Map<GlobalRebalancePreferenceKey, Integ
     return DEFAULT_GLOBAL_REBALANCE_PREFERENCE;
   }
 
+  /**
+   * Set the asynchronous global rebalance option.
+   * @param isAsync true if the global rebalance should be calculated asynchronously
+   */
+  public void setAsyncGlobalRebalance(boolean isAsync) {
```
narendly: Rename to `enableAsyncGlobalRebalance`?

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -822,6 +828,19 @@ public void setGlobalRebalancePreference(Map<GlobalRebalancePreferenceKey, Integ
...
+    _record.setBooleanField(ClusterConfigProperty.ASYNC_GLOBAL_REBALANCE.name(), isAsync);
+  }
+
+  public boolean getAsyncGlobalRebalance() {
```
narendly: Rename to `isAsyncGlobalRebalanceEnabled`?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -220,17 +221,24 @@ public GenericHelixController(String clusterName, Set<Pipeline.Type> enabledPipe
   class RebalanceTask extends TimerTask {
     final HelixManager _manager;
     final ClusterEventType _clusterEventType;
+    private final boolean _requestCacheRefresh;
```
narendly: Please follow a proper boolean naming convention. In this case, `shouldRefreshCache` or something like this would read much better.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -220,17 +221,24 @@ public GenericHelixController(String clusterName, Set<Pipeline.Type> enabledPipe
...
     private long _nextRebalanceTime;
 
     public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType) {
       this(manager, clusterEventType, -1);
+    }
 
+    public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType,
```
narendly: I am going to make an argument that the change around RebalanceTask/GenericHelixController could be a separate PR. Could you please consider making it a separate PR to make the development/review go smoother and more modular?

helix-core/src/main/java/org/apache/helix/controller/changedetector/ResourceChangeSnapshot.java
```diff
@@ -130,18 +132,22 @@ private IdealState trimIdealState(IdealState originalIdealState) {
...
       case FULL_AUTO:
         // For FULL_AUTO resources, both map fields and list fields are not considered as data input
         // for the controller. The controller will write to these two types of fields for persisting
         // the assignment mapping.
-        trimmedIdealState.getRecord().setListFields(Collections.emptyMap());
-        trimmedIdealState.getRecord().setMapFields(Collections.emptyMap());
-        break;
+        idealStateRecord.setListFields(idealStateRecord.getListFields().keySet().stream().collect(
+            Collectors.toMap(partition -> partition, partition -> Collections.emptyList())));
+        // Continue to clean up map fields.
```
narendly: Rather, say mapFields will be cleaned up in case SEMI_AUTO so the reader knows where to look.

helix-core/src/main/java/org/apache/helix/controller/changedetector/ResourceChangeSnapshot.java
```diff
@@ -130,18 +132,22 @@ private IdealState trimIdealState(IdealState originalIdealState) {
     // Clone the IdealState to avoid modifying the objects in the Cluster Data Cache, which might
     // be used by the other stages in the pipeline.
     IdealState trimmedIdealState = new IdealState(originalIdealState.getRecord());
+    ZNRecord idealStateRecord = trimmedIdealState.getRecord();
```
narendly: Nit: `trimmedIdealStateRecord`

helix-core/src/main/java/org/apache/helix/controller/changedetector/ResourceChangeSnapshot.java
```diff
@@ -130,18 +132,22 @@ private IdealState trimIdealState(IdealState originalIdealState) {
...
+      // WARNING: the IdealState copy constructor is not really deep copy. So we should not modify
+      // the values directly or the cached values will be changed.
       case FULL_AUTO:
         // For FULL_AUTO resources, both map fields and list fields are not considered as data input
         // for the controller. The controller will write to these two types of fields for persisting
         // the assignment mapping.
-        trimmedIdealState.getRecord().setListFields(Collections.emptyMap());
-        trimmedIdealState.getRecord().setMapFields(Collections.emptyMap());
-        break;
+        idealStateRecord.setListFields(idealStateRecord.getListFields().keySet().stream().collect(
```
narendly: This whole change could also be a self-contained PR where you actually explain why the change is needed. 

For example, in the PR description for this change, you could explain why Collections.emptyMap() doesn't work and you have to explicitly set an emptyList for each partition key. Wouldn't you think that would make things a lot easier for the reviewer, if you were the one reviewing the code?

helix-core/src/main/java/org/apache/helix/monitoring/metrics/model/LatencyMetric.java
```diff
@@ -28,8 +28,6 @@
  * how long a particular stage in the logic took in milliseconds.
  */
 public abstract class LatencyMetric extends HistogramDynamicMetric implements Metric<Long> {
-  protected long _startTime;
-  protected long _endTime;
```
narendly: This should be in a separate PR with changes involving the other Metric classes.

jiajunwang:
> This is a preliminary review more on style and structuring the PR. I'd love to get to the async logic, but there seem to be a few things that should be sorted out before we jump deep into that..
> 
> Overall, I'd like to reiterate the point that we need to keep the PR size small and modular. I have demonstrated what could be separated into self-contained PRs. You will see the review go a lot faster and you will get better reviews.
> 
> This way, we all win because reviewers will be happy to review the code for quality, and as a PR submitter, you'll see the review go a lot faster, and we all get a better quality algorithm/product.

First of all, thanks for the review. I agree with you.
This PR was not intended to be so large. But I find myself changing more and more while tests keep failing. Overall, the changes are all related. In other word, if we are not going to do async baseline calculation, all the additional change won't be necessary.

To reduce overhead, I plan to re-organize the commits in this PR first. And then, if applicable, I will create additional PRs for the preparation commits. But those smaller changes won't have enough motivation if we review them separately. So I plan to just link those PRs to this one for the reviewers to understand the background.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
```diff
@@ -30,6 +31,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(RebalanceLatencyGauge.class);
   private static final long VALUE_NOT_SET = -1;
   private long _lastEmittedMetricValue = VALUE_NOT_SET;
+  private final ThreadLocal<Long> _startTime;
```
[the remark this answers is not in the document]
jiajunwang: Note the 5th point in the description:
"Enable the latency metric to report correct numbers even used in multi-threads."

If you don't understand, we can talk. I will split this part into a separate PR later. Maybe you will understand by then. And I totally agree with you that we should minimize the change. That is what I'm doing.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -94,7 +94,12 @@
     // The preference of the rebalance result.
     // EVENNESS - Evenness of the resource utilization, partition, and top state distribution.
     // LESS_MOVEMENT - the tendency of keeping the current assignment instead of moving the partition for optimal assignment.
-    REBALANCE_PREFERENCE
+    REBALANCE_PREFERENCE,
+    // Specify if the WAGED rebalancer should asynchronously calculating the global rebalance result.
```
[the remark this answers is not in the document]
jiajunwang: It is not really rebalancing. It is calculating the result and persists it. No replica is moved.
I still prefer the current version.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -94,7 +94,12 @@
...
+    // Note that asynchronously calculating will reduce the rebalance delay but may cause more
```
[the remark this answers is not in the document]
jiajunwang: I thought it is clear enough. But if you say so, let me add one line.

"Asynchronous baseline calculation means the partial rebalance is done based on an older baseline. So the result would be intermediate and could be changed again later."

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -94,7 +94,12 @@
...
+    // partition movements/
+    // Default to be true.
+    ASYNC_GLOBAL_REBALANCE
```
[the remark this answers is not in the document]
jiajunwang: You mean have a different enum for WAGED rebalancer options? I think it's a good idea.
One thing we might need to discuss is whether we need and how to differentiate this enum in the ZNode content. Adding prefix seems to be too much overhead.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: Sounds good. BTW, could you please also paste this convention so I can take a look?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: Agreed, that's on my list.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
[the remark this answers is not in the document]
jiajunwang: On my second thought, I don't think this config item should be bound with the internal logic too much. So let's just assume it is the "global rebalance" for simplicity. And I will make the change.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: I had a discussion with Junkai before, "enableXXXXXX" in our code is a bad naming. It is very confusing if you are disabling the option but you are still calling enableXXXX() method.
I would change it to be setAsyncGlobalRebalanceMode.

jiajunwang:
I tried to split the potential PRs into different commits. I will send them separately later. Please review those incoming PRs first. Leave this one to the end : )

jiajunwang:
Please review the following PRs first:
https://github.com/apache/helix/pull/639

narendly:
> I tried to split the potential PRs into different commits. I will send them separately later. Please review those incoming PRs first. Leave this one to the end : )

Very helpful! I think splitting PRs for review purposes is a great strategy - makes it a lot easier to give and receive reviews. I really appreciate it :)

Please make sure all these PRs are in sync though - we don't want to miss any changes for small PRs in the master PR.

jiajunwang:
So after all the prerequisite PRs have been checked in, we can finally review this final one to make baseline calculation async.
I have updated the PR. A new mvn test is in progress. I will update with the result later.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -92,11 +92,14 @@ protected AssignmentMetadataStore(BucketDataAccessor bucketDataAccessor, String
     return _bestPossibleAssignment;
   }
 
-  public void persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
+  /**
+   * @return true if a new baseline was persisted.
```
i3wangyi: The java doc is incomplete. 
1. No mention of the input parameter
2. parameter name is inconsistent with method name
3. I feel a more generic `persistAssignment(Map<String, ResourceAssignment> assignments)` is sufficient.  Keep the difference implemented elsewhere

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -109,7 +118,8 @@ private static AssignmentMetadataStore constructAssignmentStore(String metadataS
   }
 
   public WagedRebalancer(HelixManager helixManager,
-      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preference) {
+      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preference,
+      boolean isAsyncGlobalRebalance) {
```
i3wangyi: At least we need to keep the name consistent, sometimes it's global, sometimes it's baseline

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -199,23 +211,30 @@ private WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
...
-  // current preference configuration.
-  public void updatePreference(
-      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> newPreference) {
-    if (_preference.equals(NOT_CONFIGURED_PREFERENCE) || _preference.equals(newPreference)) {
-      // 1. if the preference was not configured during constructing, no need to update.
-      // 2. if the preference equals to the new preference, no need to update.
-      return;
+  // Update the rebalancer configuration if the new options are different from the current
+  // configuration.
+  public synchronized void updateRebalanceOptions(
```
i3wangyi: Make two methods? If client only wants to update the boolean but not update the preference, do you need the client to memorize the previous preference?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -413,28 +432,57 @@ private void globalRebalance(ResourceControllerDataProvider clusterData,
             HelixRebalanceException.Type.INVALID_CLUSTER_STATUS, ex);
       }
 
-      calculateAndUpdateBaseline(clusterModel, algorithm);
+      final boolean waitForGlobalRebalance = !_asyncBaselineCalculation;
```
i3wangyi: Prefer to use a consistent boolean variable, _waitForGlobalRebalance is sufficient without `_asyncBaselineCalculation`

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -413,28 +432,57 @@ private void globalRebalance(ResourceControllerDataProvider clusterData,
...
+      // Calculate the Baseline assignment for global rebalance.
+      Future<Boolean> result = _baselineCalculateExecutor.submit(() -> {
+        try {
+          // Note that we should schedule a new partial rebalance if the following calculation does
+          // not wait until the new baseline is calculated.
+          // So set doSchedulePartialRebalance to be !waitForGlobalRebalance
+          calculateAndUpdateBaseline(clusterModel, algorithm, !waitForGlobalRebalance, clusterName);
+        } catch (HelixRebalanceException e) {
+          LOG.error("Failed to calculate baseline assignment!", e);
+          return false;
```
i3wangyi: What's the value of returning a boolean without using it, how about `Future<Void>`

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: We have spent lots of effort on sync the names. Note that the process is called global rebalance. And the result is called baseline assignment. Here, the name is accurate.
If you feel the name does not match what I noted here, please comment. And I will fix it.

jiajunwang: On that saying, let me change the field name to be asyncGlobalRebalance. To avoid saying the same thing in two ways.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: Sounds good.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: _waitForGlobalRebalance at the rebalancer level does not make much sense.
And we need to pass it into 2 methods with different values. To ensure these 2 methods get the same boolean value, we need to assign it to some final local variable.

The alternative way is localAsyncBoolean = _asyncBoolean;
But I feel changing the name a little bit would help itself explain.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: The result is used to determine which error message to output.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
jiajunwang: The parameter is very obvious, I think saying that I'm passing a global baseline assignment by a parameter named globalBaseline is too verbose.
It is debatable if we should name it baseline or global baseline. I'm fine with both. But it is not the key part of this PR, so I don't want to extend my change's scope. Please feel free to change it.

One generic method is not enough. Unless we add an enum to point which type of assignment it is. I'd prefer to keep this unchanged.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -443,8 +494,13 @@ private void calculateAndUpdateBaseline(ClusterModel clusterModel, RebalanceAlgo
     } else {
       LOG.debug("Assignment Metadata Store is null. Skip persisting the baseline assignment.");
     }
-    _globalBaselineCalcLatency.endMeasuringLatency();
+    _baselineCalcLatency.endMeasuringLatency();
     LOG.info("Finish calculating the new baseline.");
+
+    if (isbaselineUpdated && doSchedulePartialRebalance) {
+      LOG.info("Schedule a new rebalance after the new baseline calculated.");
+      RebalanceUtil.scheduleOnDemandPipeline(clusterName, 0l, false);
```
narendly: Question about this: If we schedule an OnDemandPipeline, wouldn't this also trigger another global baseline calculation as well as a partial rebalance?

I am worried that this might actually become an infinite loop that triggers global baseline calculation non-stop... Or is that false flag preventing that?

Also, nit: `0l -> 0L`?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -413,28 +435,57 @@ private void globalRebalance(ResourceControllerDataProvider clusterData,
             HelixRebalanceException.Type.INVALID_CLUSTER_STATUS, ex);
       }
 
-      calculateAndUpdateBaseline(clusterModel, algorithm);
+      final boolean waitForGlobalRebalance = !_asyncGlobalRebalance;
+      final String clusterName = clusterData.getClusterName();
+      // Calculate the Baseline assignment for global rebalance.
+      Future<Boolean> result = _baselineCalculateExecutor.submit(() -> {
+        try {
+          // Note that we should schedule a new partial rebalance if the following calculation does
```
narendly: What are you exactly referring to by the "following calculation"? I'm assuming the async global baseline calculation ?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -413,28 +435,57 @@ private void globalRebalance(ResourceControllerDataProvider clusterData,
...
+          // not wait until the new baseline is calculated.
+          // So set doSchedulePartialRebalance to be !waitForGlobalRebalance
+          calculateAndUpdateBaseline(clusterModel, algorithm, !waitForGlobalRebalance, clusterName);
+        } catch (HelixRebalanceException e) {
+          LOG.error("Failed to calculate baseline assignment!", e);
+          return false;
+        }
+        return true;
```
narendly: Should we also catch a general Exception so that the thread doesn't die just in case here?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -413,28 +435,57 @@ private void globalRebalance(ResourceControllerDataProvider clusterData,
...
+      });
+      if (waitForGlobalRebalance) {
+        try {
+          if (!result.get()) {
+            throw new HelixRebalanceException("Failed to calculate for the new Baseline.",
```
narendly: Probably good to indicate here that the "async" calculation failed?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -413,28 +435,57 @@ private void globalRebalance(ResourceControllerDataProvider clusterData,
...
+                HelixRebalanceException.Type.FAILED_TO_CALCULATE);
+          }
+        } catch (InterruptedException | ExecutionException e) {
```
narendly: Should we catch a general Exception here also so that the rebalancer doesn't die (or we might already be catching outside this class already)?

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -121,14 +121,16 @@ public BestPossibleStateCalcStage(WagedRebalancer wagedRebalancer) {
   }
 
   private WagedRebalancer getWagedRebalancer(HelixManager helixManager,
-      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preferences) {
+      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preferences,
+      boolean isAsyncGlobalRebalance) {
```
narendly: isAsyncGlobalRebalance -> asyncGlobalRebalanceEnabled?

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -121,14 +121,16 @@ public BestPossibleStateCalcStage(WagedRebalancer wagedRebalancer) {
...
     // Create WagedRebalancer instance if it hasn't been already initialized
     if (_wagedRebalancer == null) {
-      _wagedRebalancer = new WagedRebalancer(helixManager, preferences);
+      _wagedRebalancer = new WagedRebalancer(helixManager, preferences, isAsyncGlobalRebalance);
     } else {
-      // Since the preference can be updated at runtime, try to update the algorithm preference
-      // before returning the rebalancer.
-      _wagedRebalancer.updatePreference(preferences);
+      // Since the rebalance configure can be updated at runtime, try to update the rebalancer
```
narendly: configure -> configuration

narendly: Question: Do we explicitly provide an API to update the rebalancer configuration? Or do users just update the ClusterConfig? I am hoping that this would make it into the user manual...

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -281,8 +283,10 @@ private boolean validateOfflineInstancesLimit(final ResourceControllerDataProvid
 
     Map<String, IdealState> newIdealStates = new HashMap<>();
 
+    ClusterConfig clusterConfig = cache.getClusterConfig();
     WagedRebalancer wagedRebalancer =
-        getWagedRebalancer(helixManager, cache.getClusterConfig().getGlobalRebalancePreference());
+        getWagedRebalancer(helixManager, clusterConfig.getGlobalRebalancePreference(),
+            clusterConfig.isGlobalRebalanceAsyncModeEnabled());
```
narendly: I like the name for this method, `isGlobalRebalanceAsyncModeEnabled` :)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -92,11 +92,14 @@ protected AssignmentMetadataStore(BucketDataAccessor bucketDataAccessor, String
...
+   */
+  public boolean persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
     // TODO: Make the write async?
     // If baseline hasn't changed, skip writing to metadata store
     if (compareAssignments(_globalBaseline, globalBaseline)) {
-      return;
+      return false;
```
narendly: This behavior might be confusing. This is not ideal... usually people would think if I get a false, that means that a write to metadata store failed (for example, ZK write failure). But here, that's not necessarily the case. In this case, we decide not to write it on purpose because there's no diff.

How should we differentiate this behavior clearly? No diff != Write failure and should make each case be expressed distinctly?

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/MockAssignmentMetadataStore.java
```diff
@@ -41,17 +41,19 @@
     return _persistGlobalBaseline;
   }
 
-  public void persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
+  public boolean persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
     _persistGlobalBaseline = globalBaseline;
+    return true;
```
narendly: This behavior might be confusing. This is not ideal... usually people would think if I get a false, that means that a write to metadata store failed (for example, ZK write failure). But here, that's not necessarily the case. In this case, we decide not to write it on purpose because there's no diff.

How should we differentiate this behavior clearly? No diff != Write failure and should make each case be expressed distinctly?

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestMixedModeAutoRebalance.java
```diff
@@ -111,112 +109,125 @@ public void beforeClass() throws Exception {
...
+      }, 2000);
+
+      ExternalView ev =
+          _gSetupTool.getClusterManagementTool().getResourceExternalView(CLUSTER_NAME, dbName);
+      IdealState is =
+          _gSetupTool.getClusterManagementTool().getResourceIdealState(CLUSTER_NAME, dbName);
+      validateMinActiveAndTopStateReplica(is, ev, _replica, NUM_NODE);
+    } finally {
+      _gSetupTool.getClusterManagementTool().dropResource(CLUSTER_NAME, dbName);
+      _clusterVerifier.verify(5000);
```
narendly: 1. Should we make 5000 a constant?
2. Does verify() really check that the cluster is gone? Shouldn't we explicitly check for the ZNodes and see if they return null when read or false on an exist() call?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -92,11 +92,14 @@ protected AssignmentMetadataStore(BucketDataAccessor bucketDataAccessor, String
...
   }
 
-  public void persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
+  /**
+   * @return true if a new baseline was persisted.
+   */
+  public boolean persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
     // TODO: Make the write async?
     // If baseline hasn't changed, skip writing to metadata store
     if (compareAssignments(_globalBaseline, globalBaseline)) {
```
pkuwm: Can we move this check outside of this method? My 2 cents, this persisting method only does persisting. True if successful, false if fails. If baseline is not changed, it should be decided outside of this method. If you call this method to persist, I will Do my job to persist it no matter it is not changed or not.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -109,14 +112,18 @@ public void persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
...
 
-  public void persistBestPossibleAssignment(
+  /**
+   * @return true if a new best possible assignment was persisted.
+   */
+  public boolean persistBestPossibleAssignment(
       Map<String, ResourceAssignment> bestPossibleAssignment) {
     // TODO: Make the write async?
     // If bestPossibleAssignment hasn't changed, skip writing to metadata store
     if (compareAssignments(_bestPossibleAssignment, bestPossibleAssignment)) {
```
pkuwm: Same as above in baseline persisting

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -79,21 +84,25 @@
...
   private final HelixManager _manager;
   private final MappingCalculator<ResourceControllerDataProvider> _mappingCalculator;
   private final AssignmentMetadataStore _assignmentMetadataStore;
 
   private final MetricCollector _metricCollector;
   private final CountMetric _rebalanceFailureCount;
-  private final CountMetric _globalBaselineCalcCounter;
-  private final LatencyMetric _globalBaselineCalcLatency;
+  private final CountMetric _baselineCalcCounter;
+  private final LatencyMetric _baselineCalcLatency;
```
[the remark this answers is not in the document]
jiajunwang: Please refer to Yi's comment, "globalRebalance" or "baselineCalc", we should pick only one. I personally can go with either one. But "globalBaseline" is mixed up and confusing.
Let me know which one do you prefer?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -79,21 +84,25 @@
...
   private final LatencyMetric _writeLatency;
   private final CountMetric _partialRebalanceCounter;
   private final LatencyMetric _partialRebalanceLatency;
   private final LatencyMetric _stateReadLatency;
   private final BaselineDivergenceGauge _baselineDivergenceGauge;
 
+  private boolean _asyncGlobalRebalance;
```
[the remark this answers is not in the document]
jiajunwang: Sure

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -199,23 +211,33 @@ private WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
...
+  public void setGlobalRebalanceAsyncMode(boolean asyncGlobalRebalance) {
+    _asyncGlobalRebalance = asyncGlobalRebalance;
   }
 
-  // Update the rebalancer preference configuration if the new preference is different from the
-  // current preference configuration.
-  public void updatePreference(
+  // Update the rebalancer pereference if the new options are different from the current
+  // pereference.
+  public synchronized void updateRebalancePereference(
```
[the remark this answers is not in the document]
jiajunwang: yeah, a typo here. Will fix.
Let's just keep it preference. We can potentially merge the options as I did originally. But for now, I don't have the strong motivation of doing that.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -199,23 +211,33 @@ private WagedRebalancer(AssignmentMetadataStore assignmentMetadataStore,
...
     }
-    _rebalanceAlgorithm = ConstraintBasedAlgorithmFactory.getInstance(newPreference);
-    _preference = ImmutableMap.copyOf(newPreference);
   }
 
   // Release all the resources.
   public void close() {
+    if (_baselineCalculateExecutor != null) {
+      _baselineCalculateExecutor.shutdown();
+    }
```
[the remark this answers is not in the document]
jiajunwang: That is to ensure the shutdown also cancel the ongoing runnable. My original thought is that we don't cancel the ongoing calculation. Since there are at most one in running state. So it should not take a long time. In addition, since the change detector already finds something so we started calculating a new baseline, it might be worthy to finish the process instead of just ignore those changes.
What do you think?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -413,28 +435,57 @@ private void globalRebalance(ResourceControllerDataProvider clusterData,
...
+  private void calculateAndUpdateBaseline(ClusterModel clusterModel, RebalanceAlgorithm algorithm,
+      boolean doSchedulePartialRebalance, String clusterName)
       throws HelixRebalanceException {
     LOG.info("Start calculating the new baseline.");
-    _globalBaselineCalcCounter.increment(1L);
-    _globalBaselineCalcLatency.startMeasuringLatency();
+    _baselineCalcCounter.increment(1L);
+    _baselineCalcLatency.startMeasuringLatency();
 
+    boolean isbaselineUpdated = false;
```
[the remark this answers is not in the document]
jiajunwang: Naming is hard. So the meaning of this boolean is if a new baseline (different from the older one) has been persisted. It is not about successful or not.
didNewBselineUpdated?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: No, unless there are other cluster changes happened which triggers a baseline calculation anyway.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: Yeah, will change it

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: If this thread fails because of other exception, we will handle in the future.get(). We are fine as long as the threadpool is still alive.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: if waitForGlobalRebalance == true, it is not asynchronously done.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: It should die on an unknown exception. And the pipeline will handle as if we are using another rebalancer.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
jiajunwang: The user just updates in the cluster config. Please consider this method as internal usage only.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
jiajunwang: That's what you suggested, I also think it is a clean one : )

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
jiajunwang: Write failure -> exception
Write skipped -> return false
How about this?

jiajunwang: We have such logic in ZkClient.updater()

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/MockAssignmentMetadataStore.java
jiajunwang: Let's keep the same discussion in the same comment.

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestMixedModeAutoRebalance.java
jiajunwang: 1. I agree, but I didn't find a good place to put it. If in this file, we only have 2 references. If move it somewhere else, then different test case might have a different requirement. So I'd prefer to keep the test case easier for now.
2. It just needs to ensure the resource is gone.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
jiajunwang: Please refer to Hunter's comments. Let's discuss the same problem in the same thread.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
jiajunwang: Please refer to Hunter's comments. Let's discuss the same problem in the same thread.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -413,28 +435,57 @@ private void globalRebalance(ResourceControllerDataProvider clusterData,
             HelixRebalanceException.Type.INVALID_CLUSTER_STATUS, ex);
       }
 
-      calculateAndUpdateBaseline(clusterModel, algorithm);
+      final boolean waitForGlobalRebalance = !_asyncGlobalRebalance;
+      final String clusterName = clusterData.getClusterName();
+      // Calculate the Baseline assignment for global rebalance.
```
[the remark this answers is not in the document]
jiajunwang: It is not necessarily true. If waitForGlobalRebalance is true, it is not done asynchronously.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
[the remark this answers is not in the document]
jiajunwang: How about isBaselineChanged?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -443,8 +494,13 @@ private void calculateAndUpdateBaseline(ClusterModel clusterModel, RebalanceAlgo
     } else {
       LOG.debug("Assignment Metadata Store is null. Skip persisting the baseline assignment.");
     }
-    _globalBaselineCalcLatency.endMeasuringLatency();
+    _baselineCalcLatency.endMeasuringLatency();
     LOG.info("Finish calculating the new baseline.");
+
+    if (isbaselineUpdated && doSchedulePartialRebalance) {
+      LOG.info("Schedule a new rebalance after the new baseline calculated.");
```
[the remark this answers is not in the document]
jiajunwang: We have the 2 logs here. I agree to make the first one "Global baseline calculation completed and has been persisted into metadata store." So we have the complete message when combined.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: 0L it is now : )

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
jiajunwang: I just removed my previous comment about ZK updater. That was not right.
What @pkuwm mentioned in the other comment can help with it. But moving this check outside is too verbose and complicated the store class design.
Let me know if you think the following behavior is acceptable or not:
Write failure -> throws exception
Write skipped -> return false

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -413,28 +434,58 @@ private void globalRebalance(ResourceControllerDataProvider clusterData,
             HelixRebalanceException.Type.INVALID_CLUSTER_STATUS, ex);
       }
 
-      calculateAndUpdateBaseline(clusterModel, algorithm);
+      final boolean waitForGlobalRebalance = !_asyncGlobalRebalanceEnabled;
```
narendly: Nit: we flip the boolean a lot here - do you think that's necessary? Can we just use one boolean?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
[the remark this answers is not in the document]
narendly: Actually, I think what I meant here is "in a different thread". But it's kind of self-explanatory, so please disregard this comment.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -443,8 +494,13 @@ private void calculateAndUpdateBaseline(ClusterModel clusterModel, RebalanceAlgo
     } else {
       LOG.debug("Assignment Metadata Store is null. Skip persisting the baseline assignment.");
     }
-    _globalBaselineCalcLatency.endMeasuringLatency();
-    LOG.info("Finish calculating the new baseline.");
+    _baselineCalcLatency.endMeasuringLatency();
+    LOG.info("Global baseline calculation completed and has been persisted into metadata store.");
+
+    if (isBaselineChanged && shouldSchedulePartialRebalance) {
+      LOG.info("Schedule a new rebalance after the new baseline calculated.");
```
narendly: Nit: "after the new baseline calculated." -> "after the new baseline calculation has finished"

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestMixedModeAutoRebalance.java
narendly: Okay. As for 2, I actually misread and thought you were doing a dropCluster(). verify makes sense in this case. Since this is a test, I'm fine with keeping 5000 for now, but it would be better to create a timeout constant either inside the verifier or in this class since I do see a lot of magic numbers here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
jiajunwang: Added a TODO here. Also modified comments.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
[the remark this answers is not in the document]
jiajunwang: It might be for someone. It brings some value to use the same name everywhere.
Anyway, if you find any globalBaseline in the code that I leftover, feel free to comment or just change it.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: It has to be done once, since the inputs for 2 methods are different.
After comparing multiple possible combinations, I feel the current way requires fewer comments.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
jiajunwang: Changed

jiajunwang:

PR #635, opened 2019-11-29
Fix a potential issue in the ResourceChangeSnapshot.

jiajunwang:
### Issues

- [x] My PR addresses the following Helix issues and references them in the PR description:

#563 

### Description

- [x] Here are some details about my PR, including screenshots of any UI changes:

Prerequisite of #632 

The trim logic in the ResourceChangeSnapshot for cleaning up the IdealState should not clear the whole map. This will cause the WAGED rebalancer to ignore changes such as new partitions into the partition list.
Modify the test case accordingly.

### Tests

- [x] The following tests are written for this issue:

TestResourceChangeDetector

- [ ] The following is the result of the "mvn test" command on the appropriate module:

Please refer to #632 's test result.

### Commits

- [x] My commits all reference appropriate Apache Helix GitHub issues in their subject lines, and I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Documentation

- [x] In case of new functionality, my PR adds documentation in the following wiki page:

NA

### Code Quality

- [x] My diff has been formatted using helix-style.xml

PR #636, opened 2019-11-29
Simply and enhance the RebalanceLatencyGauge so it can be used in multi-threads.

jiajunwang:
### Issues

#563 

### Description

Prerequisite of #632 

The previous design of RebalanceLatencyGauge won't support asynchronous metric data emitting. This PR adds support by using a ThreadLocal object.
The metric logic is not changed.

### Tests

NA

### Commits

### Documentation

NA

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
@@ -19,9 +19,10 @@ package org.apache.helix.monitoring.metrics.implementation;
  * under the License.
  */
 
+import java.util.concurrent.TimeUnit;
+
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.SlidingTimeWindowArrayReservoir;
-import java.util.concurrent.TimeUnit;
 import org.apache.helix.monitoring.metrics.model.LatencyMetric;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -30,6 +31,7 @@ public class RebalanceLatencyGauge extends LatencyMetric {
   private static final Logger LOG = LoggerFactory.getLogger(RebalanceLatencyGauge.class);
   private static final long VALUE_NOT_SET = -1;
   private long _lastEmittedMetricValue = VALUE_NOT_SET;
+  private final ThreadLocal<Long> _startTime;
 
   /**
    * Instantiates a new Histogram dynamic metric.
@@ -39,6 +41,7 @@ public class RebalanceLatencyGauge extends LatencyMetric {
     super(metricName, new Histogram(
         new SlidingTimeWindowArrayReservoir(slidingTimeWindow, TimeUnit.MILLISECONDS)));
     _metricName = metricName;
+    _startTime = ThreadLocal.withInitial(() -> VALUE_NOT_SET);
   }
 
   /**
@@ -50,7 +53,7 @@ public class RebalanceLatencyGauge extends LatencyMetric {
   @Override
   public void startMeasuringLatency() {
     reset();
-    _startTime = System.currentTimeMillis();
+    _startTime.set(System.currentTimeMillis());
   }
 
   /**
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
```diff
@@ -50,24 +53,24 @@ public RebalanceLatencyGauge(String metricName, long slidingTimeWindow) {
...
       LOG.error(
           "Needs to call startMeasuringLatency first! Ignoring and resetting the metric. Metric name: {}",
           _metricName);
-      reset();
       return;
     }
-    _endTime = System.currentTimeMillis();
-    _lastEmittedMetricValue = _endTime - _startTime;
-    updateValue(_lastEmittedMetricValue);
+    synchronized (this) {
```
narendly: Question: could we possibly make `_lastEmittedMetricValue `a `ThreadLocal `as well and remove this `synchronized `primitive?

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
```diff
@@ -30,6 +31,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(RebalanceLatencyGauge.class);
   private static final long VALUE_NOT_SET = -1;
   private long _lastEmittedMetricValue = VALUE_NOT_SET;
+  private final ThreadLocal<Long> _startTime;
```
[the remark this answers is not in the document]
jiajunwang: Sorry I didn't see those comments. But it's a good suggestion. Let me add it.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
jiajunwang: No, the caller is usually reading the data from a different thread. If you make this output value thread local, you will always get the init value (-1). So it has to be a normal value.
The same to the metric object too.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
```diff
@@ -30,6 +31,8 @@
   private static final Logger LOG = LoggerFactory.getLogger(RebalanceLatencyGauge.class);
   private static final long VALUE_NOT_SET = -1;
   private long _lastEmittedMetricValue = VALUE_NOT_SET;
+  // Use threadlocal here so the start time can be updated and recorded in multi-threads.
+  private final ThreadLocal<Long> _startTime;
```
i3wangyi: Wouldn't it be easier for each thread creating a new instance of RebalanceLatencyGauge, that you don't need to handle the concurrency complexities? It's more error-prone as we all know 
But I do see the usage pattern is the one single instance RebalanceLatencyGauge will be long-lived in the memory - if it's desired, can we make sure the class usage always follows the singleton pattern?

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
```diff
@@ -50,24 +54,24 @@ public RebalanceLatencyGauge(String metricName, long slidingTimeWindow) {
   @Override
   public void startMeasuringLatency() {
     reset();
-    _startTime = System.currentTimeMillis();
+    _startTime.set(System.currentTimeMillis());
   }
 
   /**
    * WARNING: this method is not thread-safe.
```
i3wangyi: nit: update the comment? After your change, the class should be thread-safe implementation, right?

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
jiajunwang: It is necessary because we don't want multiple gauge objects. That will require us to assign different sensor names to those metrics. That would be very hard to manage.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
jiajunwang: Good catch. Will fix.

helix-core/src/main/java/org/apache/helix/monitoring/metrics/implementation/RebalanceLatencyGauge.java
i3wangyi: Thanks for the clarification. Is it possible to make the class a singleton pattern and limit the pubic constructor usages? @jiajunwang

jiajunwang:

jiajunwang:

PR #637, opened 2019-11-29
Add new WAGED rebalancer config item "GLOBAL_REBALANCE_ASYNC_MODE".

jiajunwang:
### Issues

#563 

### Description

Prerequisite of #632 

This option will be used by the WAGED rebalancer to determine if the global rebalance should be performed asynchronously.
Also, refine the other WAGED rebalancer configuration items so they are separately categorized.

### Tests

TestClusterConfig

### Commits

### Documentation

NA

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
+++ b/helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
@@ -83,13 +83,12 @@ public class ClusterConfig extends HelixProperty {
     DISABLED_INSTANCES,
 
     // Specifies job types and used for quota allocation
-    QUOTA_TYPES
-  }
+    QUOTA_TYPES,
 
-  /**
-   * Configurable characteristics of the WAGED rebalancer.
-   */
-  public enum WagedRebalancerConfigProperty {
+    /**
+     * Configurable characteristics of the WAGED rebalancer.
+     * TODO: Split the WAGED rebalancer configuration items to the other config file.
+     */
     // The required instance capacity keys for resource partition assignment calculation.
     INSTANCE_CAPACITY_KEYS,
     // The default instance capacity if no capacity is configured in the Instance Config node.
@@ -100,11 +99,14 @@ public class ClusterConfig extends HelixProperty {
     // EVENNESS - Evenness of the resource utilization, partition, and top state distribution.
     // LESS_MOVEMENT - the tendency of keeping the current assignment instead of moving the partition for optimal assignment.
     REBALANCE_PREFERENCE,
-    // Specify if the WAGED rebalancer should asynchronously perform global rebalance.
-    // Note that asynchronous calculation will reduce the rebalance delay but may cause more
-    // partition movements. This is because the partial rebalance will be preformed with an stale
-    // baseline. The rebalance result would be an intermediate one and could be changed again when
-    // a new baseline is calculated.
+    // Specify if the WAGED rebalancer should asynchronously perform the global rebalance, which is
+    // in general slower than the partial rebalance.
+    // Note that asynchronous global rebalance calculation will reduce the controller rebalance
+    // delay. But it may cause more partition movements. This is because the partial rebalance will
+    // be performed with a stale baseline. The rebalance result would be an intermediate one and
+    // could be changed again when a new baseline is calculated.
+    // For more details, please refer to
+    // https://github.com/apache/helix/wiki/Weight-aware-Globally-Evenly-distributed-Rebalancer#rebalance-coordinator
     //
     // Default to be true.
     GLOBAL_REBALANCE_ASYNC_MODE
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -94,7 +99,15 @@
     // The preference of the rebalance result.
     // EVENNESS - Evenness of the resource utilization, partition, and top state distribution.
     // LESS_MOVEMENT - the tendency of keeping the current assignment instead of moving the partition for optimal assignment.
-    REBALANCE_PREFERENCE
+    REBALANCE_PREFERENCE,
+    // Specify if the WAGED rebalancer should asynchronously perform global rebalance.
+    // Note that asynchronous calculation will reduce the rebalance delay but may cause more
+    // partition movements. This is because the partial rebalance will be preformed with an stale
```
narendly: preformed -> performed?
an stale -> a stale

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -94,7 +99,15 @@
...
+    // baseline. The rebalance result would be an intermediate one and could be changed again when
+    // a new baseline is calculated.
+    //
```
narendly: I think it would be a good idea to add:

1. why synchronous baseline calculation would necessarily be slower (seems like you could make an argument that it's actually not slow because it will just compute based on the latest (non-stale) baseline). In an async case, since you calculate based on a stale baseline, your results may not be up-to-date, so you have to move more partitions, so overall are you losing out on speed?

2. The actual order of calculation. Something like:
- synchronous: 1) partial rebalance -> 2) global baseline calculation based on result of 1 -> 3) best possible calculation (all in a linear fashion)
- asynchronous: 1) (a) partial rebalance + (b) async global baseline calculation based on previous best possible before partial rebalance -> 2) best possible calculation based on result of (a) and (b).

Please let me know if this understanding is accurate. The reason I'm being very verbose and detailed is because I want to make sure this doesn't become a "mystery" piece of logic.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -822,6 +840,19 @@ public void setGlobalRebalancePreference(Map<GlobalRebalancePreferenceKey, Integ
     return DEFAULT_GLOBAL_REBALANCE_PREFERENCE;
   }
 
+  /**
+   * Set the asynchronous global rebalance mode.
+   * @param isAsync true if the global rebalance should be performed asynchronously
+   */
+  public void setGlobalRebalanceAsyncMode(boolean isAsync) {
```
narendly: Nit: `enableGlobalRebalanceAsyncMode`?

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: Good catch.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: It is not easy to make it clear here. So I think we just add a link to the github wiki page.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: Please see my reply in #632. enableXXXX is not a preferred name.
Considering enableXXXX(false), it is very easy to get lost if you are enabling it or disabling it.

setXXXX is more straight forward.

jiajunwang:
I rollback part of the change. The additional enum in the cluster config is dangerous. When you only have one enum in the config, there is no way a dev can create duplicate config item key. But when you have more than one enums, conflicting keys become possible. We'd better not do it now.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -822,6 +838,19 @@ public void setGlobalRebalancePreference(Map<GlobalRebalancePreferenceKey, Integ
     return DEFAULT_GLOBAL_REBALANCE_PREFERENCE;
   }
 
+  /**
+   * Set the asynchronous global rebalance mode.
```
i3wangyi: nit: the method will be used mostly by client library, could you make the java doc more concrete? How often the async rebalance will be triggered, estimated impact, default value and so on?

jiajunwang: This part has been documented in the GLOBAL_REBALANCE_ASYNC_MODE enum items. I think that would be a better place for detail.

jiajunwang:

jiajunwang:

PR #638, opened 2019-11-29
Decouple the event type and the scheduled rebalance cache refresh option.

jiajunwang:
### Issues

#563 

### Description

Prerequisite of #632 

The previous design is that both on-demand and periodic rebalance scheduling task will request for a cache refresh. This won't be always true moving forward.
For example, the WAGED rebalancer async baseline calculating requests for a scheduled rebalance. But cache refresh won't be necessary.
This PR does not change any business logic. It prepares for future feature change.

### Tests

NA

### Commits

### Documentation

NA

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
@@ -221,7 +222,7 @@ public class GenericHelixController implements IdealStateChangeListener,
   class RebalanceTask extends TimerTask {
     final HelixManager _manager;
     final ClusterEventType _clusterEventType;
-    private final boolean _shouldRefreshCache;
+    private final Optional<Boolean> _shouldRefreshCache;
     private long _nextRebalanceTime;
 
     public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType) {
@@ -230,11 +231,11 @@ public class GenericHelixController implements IdealStateChangeListener,
 
     public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType,
         long nextRebalanceTime) {
-      this(manager, clusterEventType, nextRebalanceTime, true);
+      this(manager, clusterEventType, nextRebalanceTime, Optional.empty());
     }
 
     public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType,
-        long nextRebalanceTime, boolean shouldRefreshCache) {
+        long nextRebalanceTime, Optional<Boolean> shouldRefreshCache) {
       _manager = manager;
       _clusterEventType = clusterEventType;
       _nextRebalanceTime = nextRebalanceTime;
@@ -248,7 +249,11 @@ public class GenericHelixController implements IdealStateChangeListener,
     @Override
     public void run() {
       try {
-        if (_shouldRefreshCache) {
+        // If refresh option is specified, there is no need to check the event type.
+        if ((!_shouldRefreshCache.isPresent() && (
+            _clusterEventType.equals(ClusterEventType.PeriodicalRebalance) || _clusterEventType
+                .equals(ClusterEventType.OnDemandRebalance))) ||
+            (_shouldRefreshCache.isPresent() && _shouldRefreshCache.get())) {
           requestDataProvidersFullRefresh();
 
           HelixDataAccessor accessor = _manager.getHelixDataAccessor();
@@ -395,7 +400,7 @@ public class GenericHelixController implements IdealStateChangeListener,
 
     RebalanceTask newTask =
         new RebalanceTask(_helixManager, ClusterEventType.OnDemandRebalance, rebalanceTime,
-            shouldRefreshCache);
+            Optional.of(shouldRefreshCache));
 
     _onDemandRebalanceTimer.schedule(newTask, delay);
     logger.info("Scheduled instant pipeline run for cluster {}." , _helixManager.getClusterName());
@@ -710,8 +715,8 @@ public class GenericHelixController implements IdealStateChangeListener,
               forceRebalance(manager, ClusterEventType.RetryRebalance);
             } else {
               _asyncTasksThreadPool
-                  .schedule(new RebalanceTask(manager, ClusterEventType.RetryRebalance, -1, false),
-                      delay, TimeUnit.MILLISECONDS);
+                  .schedule(new RebalanceTask(manager, ClusterEventType.RetryRebalance), delay,
+                      TimeUnit.MILLISECONDS);
             }
             logger.info("Retry rebalance pipeline with delay " + delay + "ms for cluster: " + _clusterName);
           }
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -365,6 +365,15 @@ public class GenericHelixController implements IdealStateChangeListener,
   /**
    * Schedule an on demand rebalance pipeline.
    * @param delay
+   */
+  @Deprecated
+  public void scheduleOnDemandRebalance(long delay) {
+    scheduleOnDemandRebalance(delay, true);
+  }
+
+  /**
+   * Schedule an on demand rebalance pipeline.
+   * @param delay
    * @param shouldRefreshCache true if refresh the cache before scheduling a rebalance.
    */
   public void scheduleOnDemandRebalance(long delay, boolean shouldRefreshCache) {
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -222,7 +222,7 @@ public class GenericHelixController implements IdealStateChangeListener,
   class RebalanceTask extends TimerTask {
     final HelixManager _manager;
     final ClusterEventType _clusterEventType;
-    private final Optional<Boolean> _shouldRefreshCache;
+    private final Optional<Boolean> _shouldRefreshCacheOption;
     private long _nextRebalanceTime;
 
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -220,17 +221,24 @@ public GenericHelixController(String clusterName, Set<Pipeline.Type> enabledPipe
   class RebalanceTask extends TimerTask {
     final HelixManager _manager;
     final ClusterEventType _clusterEventType;
+    private final boolean _shouldRefreshCache;
     private long _nextRebalanceTime;
 
     public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType) {
       this(manager, clusterEventType, -1);
```
narendly: Nit: could we create a constant for `REBALANCE_IMMEDIATELY` or `IMMEDIATE_REBALANCE` with value -1?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -358,8 +365,9 @@ public void scheduleRebalance(long rebalanceTime) {
   /**
    * Schedule an on demand rebalance pipeline.
    * @param delay
+   * @param shouldRefreshCache true if refresh the cache before scheduling a rebalance.
    */
-  public void scheduleOnDemandRebalance(long delay) {
+  public void scheduleOnDemandRebalance(long delay, boolean shouldRefreshCache) {
```
narendly: Constructor change in a public method may cause backward-compatibility issues. Could we keep the original public method and add this overloaded one?

If so, what should the default behavior be for `public void scheduleOnDemandRebalance(long delay)`?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -240,8 +248,7 @@ public long getNextRebalanceTime() {
     @Override
     public void run() {
       try {
-        if (_clusterEventType.equals(ClusterEventType.PeriodicalRebalance) || _clusterEventType
-            .equals(ClusterEventType.OnDemandRebalance)) {
+        if (_shouldRefreshCache) {
```
narendly: Is this a behavior change (since it could be risky in different parts of the controller pipeline - as in it might refresh cache when it doesn't need to and increase latency)?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -692,8 +701,8 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
               forceRebalance(manager, ClusterEventType.RetryRebalance);
             } else {
               _asyncTasksThreadPool
-                  .schedule(new RebalanceTask(manager, ClusterEventType.RetryRebalance), delay,
-                      TimeUnit.MILLISECONDS);
+                  .schedule(new RebalanceTask(manager, ClusterEventType.RetryRebalance, -1, false),
```
narendly: Nit: -1 -> constant (as mentioned above) for readability?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: Good idea. But should not be a change in this PR. Let me create an issue for that.

jiajunwang: https://github.com/apache/helix/issues/640

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: Good catch. The old behavior is shouldRefreshCache = true.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: I tried to keep everything work as the old day. Logic only is changed if the code is using the new methods.
Please let me know if you find any outliers.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: Let me do this after the merge, tracking in issue #640

jiajunwang:
> I have a feeling that this should actually require more changes in other parts of the code since there are many places that use scheduledRebalance functionality. Is that true? If so, should we identify and review each use case?

You are right, if anyone relies on, for example ClusterEventType.RetryRebalance, to trigger a rebalance without a cache refresh, the old application code will trigger unnecessary refresh with the new lib.
An alternative way is that we add a optional config to the Rebalance Task. So that it takes effect when it's specified, or the logic will be kept the same. But that will make the code too complicated. I would prefer not doing that since guessing the refresh requirement based on the event type is not quite decent logic anyway. But I'm open to change.
@lei-xia  Could you please also comment here? Thanks.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -220,17 +221,24 @@ public GenericHelixController(String clusterName, Set<Pipeline.Type> enabledPipe
   class RebalanceTask extends TimerTask {
     final HelixManager _manager;
     final ClusterEventType _clusterEventType;
+    private final boolean _shouldRefreshCache;
```
i3wangyi: nit: set default value, if not, Java assumes it's false by default

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -220,17 +221,24 @@ public GenericHelixController(String clusterName, Set<Pipeline.Type> enabledPipe
...
     private long _nextRebalanceTime;
 
     public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType) {
       this(manager, clusterEventType, -1);
+    }
 
+    public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType,
```
i3wangyi: can we use a simple Builder pattern; It's package-private class so I assume the constructor will be called in very limited places.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -359,7 +366,17 @@ public void scheduleRebalance(long rebalanceTime) {
...
   public void scheduleOnDemandRebalance(long delay) {
+    scheduleOnDemandRebalance(delay, true);
+  }
+
+  /**
+   * Schedule an on demand rebalance pipeline.
+   * @param delay
+   * @param shouldRefreshCache true if refresh the cache before scheduling a rebalance.
+   */
+  public void scheduleOnDemandRebalance(long delay, boolean shouldRefreshCache) {
```
i3wangyi: Please correct me if I misunderstand the boolean value's purpose, I understand it's useful if wagedRebalancer would like to calculate a global rebalance on a "snapshot", "old" version of the cache object. Is it harder to maintain and debug the cache object if we sometimes have an update-to-date cache and sometimes an old version cache? 
Is it possible to take the snapshot of cache object at some time and give it to waged rebalancer for the global calculation?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: I tried to change it to optional object. So we have strict backward compatibility.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: A Builder pattern would be overdesign here since this class is private.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: It is in our roadmap. But take a snapshot of the cache is a large project itself. For now, we don't have that feature yet.

jiajunwang:
I tried to make the new parameter optional, and now the logic should be strict backward compatible .

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -220,17 +222,24 @@ public GenericHelixController(String clusterName, Set<Pipeline.Type> enabledPipe
...
+    }
 
+    public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType,
+        long nextRebalanceTime) {
+      this(manager, clusterEventType, nextRebalanceTime, Optional.empty());
     }
 
-    public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType, long nextRebalanceTime) {
+    public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType,
+        long nextRebalanceTime, Optional<Boolean> shouldRefreshCache) {
```
i3wangyi: I will keep the constructor parameter `boolean shouldRefreshCache` and `_shouldRefreshCache = Optional.of(shouldRefreshCache)`. It makes it easier to use whoever creates the Task object

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -220,17 +222,29 @@ public GenericHelixController(String clusterName, Set<Pipeline.Type> enabledPipe
...
+      this(manager, clusterEventType, nextRebalanceTime, Optional.empty());
     }
 
-    public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType, long nextRebalanceTime) {
+    public RebalanceTask(HelixManager manager, ClusterEventType clusterEventType,
+        long nextRebalanceTime, boolean shouldRefreshCache) {
+      this(manager, clusterEventType, nextRebalanceTime, Optional.of(shouldRefreshCache));
+    }
+
+    private RebalanceTask(HelixManager manager, ClusterEventType clusterEventType,
```
i3wangyi: The private constructor is not needed if you simply init the value in-line ` private final Optional<Boolean> _shouldRefreshCacheOption = Optional.empty();`, right?

jiajunwang:

jiajunwang:

PR #642, opened 2019-12-03
Fix handleNewSession creating ephemeral node with expired session

### LINKED ISSUE #641: Zk session race condition when creating a live instance
When a storage node's network adapter has issues, network connection is lost, which causes around 5-10 Zookeeper sessions to become expired. Reconnect events are created after the expiration. Eventually this node has 40 minutes busy resetting the storage node's StateModel while helix controller regards this node as online, so helix does not move partitions mastership to other storage node. This caused 40 minutes down time for users of these partitions.

Root cause is zk session race condition:
Zk session may become expired and change before creating a live instance. So when a live instance(ephemeral node) is being created, if the expected session is expired, we should NOT create the ephemeral node.

pkuwm:
### Issues

Fixes #641 

When a storage node's network adapter has issues, network connection is lost, which causes around 5-10 Zookeeper sessions to become expired. Reconnect events are created after the expiration. Eventually this node has 40 minutes busy resetting the storage node's StateModel while helix controller regards this node as online, so helix does not move partitions mastership to other storage node. This caused 40 minutes down time for users of these partitions.

Root cause is zk session race condition:
Zk session may become expired and change before creating a live instance. So when a live instance(ephemeral node) is being created, if the expected session is expired, we should NOT create the ephemeral node.

### Description

Change list:
1. Change API `create()` in ZkClient to accept zk sessionId as a parameter. Add a new public API createEphemeral(final String path, final Object data, final String sessionId) to create a session aware ephemeral node.
2. Fire a new session event only right after the first SyncConnected is received.
3. Change code logic to handle new session operations that need to be completed in the expected session.

Some other comments from @lei-xia : https://github.com/pkuwm/helix/pull/1

### Tests

TestHandleNewSession.testDiscardExpiredSessions
TestHandleNewSession.testSessionExpiredWhenResetHandlers
TestRawZkClient.testConnectionLossWhileCreateEphemeral
TestRawZkClient.testCreateEphemeralWithMismatchedSession
TestRawZkClient.testCreateEphemeralWithValidSession
TestRawZkClient.testRetryUntilConnectedAfterConnectionLoss

mvn test at helix-core. 

[INFO] Tests run: 896, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3,339.05 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 896, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  55:43 min
[INFO] Finished at: 2020-01-05T14:19:20-08:00

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -112,7 +112,7 @@ public class ParticipantManager {
   public void handleNewSession() throws Exception {
     // Check participant's session is still valid.
     // If not, skip handling new session for this participant.
-    final String zkClientHexSession = _zkclient.getHexSessionId();
+    final String zkClientHexSession = ZKUtil.toHexSessionId(_zkclient.getSessionId());
     if (!zkClientHexSession.equals(_sessionId)) {
       LOG.warn("Skip handling new session for participant. There is a session mismatch: "
               + "participant manager session = {}, zk client session = {}", _sessionId,
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -66,7 +66,6 @@ public class ParticipantManager {
   final PropertyKey.Builder _keyBuilder;
   final String _clusterName;
   final String _instanceName;
-  final String _sessionId;
   final int _sessionTimeout;
   final ConfigAccessor _configAccessor;
   final InstanceType _instanceType;
@@ -77,14 +76,17 @@ public class ParticipantManager {
   final LiveInstanceInfoProvider _liveInstanceInfoProvider;
   final List<PreConnectCallback> _preConnectCallbacks;
 
+  private final String _sessionId;
+
   public ParticipantManager(HelixManager manager, HelixZkClient zkclient, int sessionTimeout,
-      LiveInstanceInfoProvider liveInstanceInfoProvider, List<PreConnectCallback> preConnectCallbacks) {
+      LiveInstanceInfoProvider liveInstanceInfoProvider, List<PreConnectCallback> preConnectCallbacks,
+      final String sessionId) {
     _zkclient = zkclient;
     _manager = manager;
     _clusterName = manager.getClusterName();
     _instanceName = manager.getInstanceName();
     _keyBuilder = new PropertyKey.Builder(_clusterName);
-    _sessionId = manager.getSessionId();
+    _sessionId = sessionId;
     _sessionTimeout = sessionTimeout;
     _configAccessor = manager.getConfigAccessor();
     _instanceType = manager.getInstanceType();
@@ -97,28 +99,15 @@ public class ParticipantManager {
   }
 
   /**
-   * Handles a new session for a participant.
+   * Handles a new session for a participant. The new session's id is passed in when participant
+   * manager is created, as it is required in
+   * {@link org.apache.helix.manager.zk.zookeeper.ZkClient#createEphemeral(String, Object, String)}
+   * to prevent ephemeral node creation from session race condition: ephemeral node is created by an
+   * expired or unexpected session.
    *
-   * @deprecated
-   * This is deprecated. Instead, please use {@link #handleNewSession(String)} which takes care of
-   * potential session race condition issue.
-   *
-   * @throws Exception
-   */
-  @Deprecated
-  public void handleNewSession() throws Exception {
-    handleNewSession(null);
-  }
-
-  /**
-   * Handles a new session for a participant. The new session's id has to be passed in, as it is
-   * required in {@link #createLiveInstance(String)} to prevent ephemeral node creation from session
-   * race condition: ephemeral node is created by an expired or unexpected session.
-   *
-   * @param sessionId the new session's ID
    * @throws Exception if any exception occurs
    */
-  public void handleNewSession(final String sessionId) throws Exception {
+  public void handleNewSession() throws Exception {
     joinCluster();
 
     /**
@@ -130,7 +119,7 @@ public class ParticipantManager {
 
     // TODO create live instance node after all the init works done --JJ
     // This will help to prevent controller from sending any message prematurely.
-    createLiveInstance(sessionId);
+    createLiveInstance();
     carryOverPreviousCurrentState();
 
     /**
@@ -176,7 +165,7 @@ public class ParticipantManager {
     }
   }
 
-  private void createLiveInstance(final String sessionId) {
+  private void createLiveInstance() {
     String liveInstancePath = _keyBuilder.liveInstance(_instanceName).getPath();
     LiveInstance liveInstance = new LiveInstance(_instanceName);
     liveInstance.setSessionId(_sessionId);
@@ -200,8 +189,12 @@ public class ParticipantManager {
     do {
       retry = false;
       try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), sessionId);
+        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), _sessionId);
         LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance.getEphemeralOwner());
+      } catch (ZkSessionMismatchedException e) {
+        throw new HelixException(
+            "Failed to create live instance, path: " + liveInstancePath + ", session: "
+                + _sessionId, e);
       } catch (ZkNodeExistsException e) {
         LOG.warn("found another instance with same instanceName: " + _instanceName + " in cluster "
             + _clusterName);
@@ -214,7 +207,7 @@ public class ParticipantManager {
            */
           retry = true;
         } else {
-          String ephemeralOwner = Long.toHexString(stat.getEphemeralOwner());
+          String ephemeralOwner = ZKUtil.toHexSessionId(stat.getEphemeralOwner());
           if (ephemeralOwner.equals(_sessionId)) {
             /**
              * update sessionId field in live-instance if necessary
@@ -258,15 +251,19 @@ public class ParticipantManager {
      */
     if (retry) {
       try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), sessionId);
+        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), _sessionId);
         LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance
             .getEphemeralOwner());
+      } catch (ZkSessionMismatchedException e) {
+        throw new HelixException(
+            "Failed to create live instance, path: " + liveInstancePath + ", session: "
+                + _sessionId, e);
+      } catch (ZkNodeExistsException e) {
+        throw new HelixException(
+            "Instance name: " + _instanceName + " already has a live-instance in cluster: "
+                + _clusterName);
       } catch (Exception e) {
-        String errorMessage =
-            "instance: " + _instanceName + " already has a live-instance in cluster "
-                + _clusterName;
-        LOG.error(errorMessage);
-        throw new HelixException(errorMessage);
+        throw new HelixException(e);
       }
     }
 
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -108,6 +108,16 @@ public class ParticipantManager {
    * @throws Exception if any exception occurs
    */
   public void handleNewSession() throws Exception {
+    // Check participant's session is still valid.
+    // If not, stop to handle new session for this participant.
+    final String zkClientHexSession = _zkclient.getHexSessionId();
+    if (!zkClientHexSession.equals(_sessionId)) {
+      throw new HelixException(
+          "Failed to handle new session for participant. There is a session mismatch: "
+              + "participant manager session = " + _sessionId + ", zk client session = "
+              + zkClientHexSession);
+    }
+
     joinCluster();
 
     /**
@@ -193,7 +203,7 @@ public class ParticipantManager {
         LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance.getEphemeralOwner());
       } catch (ZkSessionMismatchedException e) {
         throw new HelixException(
-            "Failed to create live instance, path: " + liveInstancePath + ", session: "
+            "Failed to create live instance, path: " + liveInstancePath + ", expected session: "
                 + _sessionId, e);
       } catch (ZkNodeExistsException e) {
         LOG.warn("found another instance with same instanceName: " + _instanceName + " in cluster "
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -97,10 +97,28 @@ public class ParticipantManager {
   }
 
   /**
-   * Handle new session for a participang.
+   * Handles a new session for a participant.
+   *
+   * @deprecated
+   * This is deprecated. Instead, please use {@link #handleNewSession(String)} which takes care of
+   * potential session race condition issue.
+   *
    * @throws Exception
    */
+  @Deprecated
   public void handleNewSession() throws Exception {
+    handleNewSession(null);
+  }
+
+  /**
+   * Handles a new session for a participant. The new session's id has to be passed in, as it is
+   * required in {@link #createLiveInstance(String)} to prevent ephemeral node creation from session
+   * race condition: ephemeral node is created by an expired or unexpected session.
+   *
+   * @param sessionId the new session's ID
+   * @throws Exception if any exception occurs
+   */
+  public void handleNewSession(final String sessionId) throws Exception {
     joinCluster();
 
     /**
@@ -112,7 +130,7 @@ public class ParticipantManager {
 
     // TODO create live instance node after all the init works done --JJ
     // This will help to prevent controller from sending any message prematurely.
-    createLiveInstance();
+    createLiveInstance(sessionId);
     carryOverPreviousCurrentState();
 
     /**
@@ -158,7 +176,7 @@ public class ParticipantManager {
     }
   }
 
-  private void createLiveInstance() {
+  private void createLiveInstance(final String sessionId) {
     String liveInstancePath = _keyBuilder.liveInstance(_instanceName).getPath();
     LiveInstance liveInstance = new LiveInstance(_instanceName);
     liveInstance.setSessionId(_sessionId);
@@ -182,7 +200,7 @@ public class ParticipantManager {
     do {
       retry = false;
       try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord());
+        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), sessionId);
         LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance.getEphemeralOwner());
       } catch (ZkNodeExistsException e) {
         LOG.warn("found another instance with same instanceName: " + _instanceName + " in cluster "
@@ -240,7 +258,7 @@ public class ParticipantManager {
      */
     if (retry) {
       try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord());
+        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), sessionId);
         LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance
             .getEphemeralOwner());
       } catch (Exception e) {
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -23,10 +23,8 @@ import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.TimeUnit;
 
 import org.I0Itec.zkclient.DataUpdater;
-import org.I0Itec.zkclient.exception.ZkNodeExistsException;
 import org.apache.helix.AccessOption;
 import org.apache.helix.BaseDataAccessor;
 import org.apache.helix.ConfigAccessor;
@@ -51,7 +49,6 @@ import org.apache.helix.model.StateModelDefinition;
 import org.apache.helix.model.builder.HelixConfigScopeBuilder;
 import org.apache.helix.participant.StateMachineEngine;
 import org.apache.helix.participant.statemachine.ScheduledTaskStateModelFactory;
-import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -110,14 +107,14 @@ public class ParticipantManager {
    * @throws Exception if any exception occurs
    */
   public void handleNewSession() throws Exception {
-    // Check participant's session is still valid.
+    // Check zk session of this participant is still valid.
     // If not, skip handling new session for this participant.
     final String zkClientHexSession = ZKUtil.toHexSessionId(_zkclient.getSessionId());
     if (!zkClientHexSession.equals(_sessionId)) {
-      LOG.warn("Skip handling new session for participant. There is a session mismatch: "
-              + "participant manager session = {}, zk client session = {}", _sessionId,
-          zkClientHexSession);
-      return;
+      throw new HelixException(
+          "Failed to handle new session for participant. There is a session mismatch: "
+              + "participant manager session = " + _sessionId + ", zk client session = "
+              + zkClientHexSession);
     }
 
     joinCluster();
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -76,6 +76,8 @@ public class ParticipantManager {
   final LiveInstanceInfoProvider _liveInstanceInfoProvider;
   final List<PreConnectCallback> _preConnectCallbacks;
 
+  // zk session id should be immutable after participant manager is created. This is to avoid
+  // session race condition when handling new session for the participant.
   private final String _sessionId;
 
   public ParticipantManager(HelixManager manager, HelixZkClient zkclient, int sessionTimeout,
@@ -109,13 +111,13 @@ public class ParticipantManager {
    */
   public void handleNewSession() throws Exception {
     // Check participant's session is still valid.
-    // If not, stop to handle new session for this participant.
+    // If not, skip handling new session for this participant.
     final String zkClientHexSession = _zkclient.getHexSessionId();
     if (!zkClientHexSession.equals(_sessionId)) {
-      throw new HelixException(
-          "Failed to handle new session for participant. There is a session mismatch: "
-              + "participant manager session = " + _sessionId + ", zk client session = "
-              + zkClientHexSession);
+      LOG.warn("Skip handling new session for participant. There is a session mismatch: "
+              + "participant manager session = {}, zk client session = {}", _sessionId,
+          zkClientHexSession);
+      return;
     }
 
     joinCluster();
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -197,97 +197,18 @@ public class ParticipantManager {
       }
     }
 
-    boolean retry;
-    do {
-      retry = false;
-      try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), _sessionId);
-        LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance.getEphemeralOwner());
-      } catch (ZkSessionMismatchedException e) {
-        throw new HelixException(
-            "Failed to create live instance, path: " + liveInstancePath + ", expected session: "
-                + _sessionId, e);
-      } catch (ZkNodeExistsException e) {
-        LOG.warn("found another instance with same instanceName: " + _instanceName + " in cluster "
-            + _clusterName);
-
-        Stat stat = new Stat();
-        ZNRecord record = _zkclient.readData(liveInstancePath, stat, true);
-        if (record == null) {
-          /**
-           * live-instance is gone as we check it, retry create live-instance
-           */
-          retry = true;
-        } else {
-          String ephemeralOwner = ZKUtil.toHexSessionId(stat.getEphemeralOwner());
-          if (ephemeralOwner.equals(_sessionId)) {
-            /**
-             * update sessionId field in live-instance if necessary
-             */
-            LiveInstance curLiveInstance = new LiveInstance(record);
-            if (!curLiveInstance.getEphemeralOwner().equals(_sessionId)) {
-              /**
-               * in last handle-new-session,
-               * live-instance is created by new zkconnection with stale session-id inside
-               * just update session-id field
-               */
-              LOG.info("overwriting session-id by ephemeralOwner: " + ephemeralOwner
-                  + ", old-sessionId: " + curLiveInstance.getEphemeralOwner() + ", new-sessionId: "
-                  + _sessionId);
-
-              curLiveInstance.setSessionId(_sessionId);
-              _zkclient.writeData(liveInstancePath, curLiveInstance.getRecord());
-            }
-          } else {
-            /**
-             * wait for a while, in case previous helix-participant exits unexpectedly
-             * and its live-instance still hangs around until session timeout
-             */
-            try {
-              TimeUnit.MILLISECONDS.sleep(_sessionTimeout + 5000);
-            } catch (InterruptedException ex) {
-              LOG.warn("Sleep interrupted while waiting for previous live-instance to go away.", ex);
-            }
-            /**
-             * give a last try after exit while loop
-             */
-            retry = true;
-            break;
-          }
-        }
-      }
-    } while (retry);
-
-    /**
-     * give a last shot
-     */
-    if (retry) {
-      try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), _sessionId);
-        LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance
-            .getEphemeralOwner());
-      } catch (ZkSessionMismatchedException e) {
-        throw new HelixException(
-            "Failed to create live instance, path: " + liveInstancePath + ", session: "
-                + _sessionId, e);
-      } catch (ZkNodeExistsException e) {
-        throw new HelixException(
-            "Instance name: " + _instanceName + " already has a live-instance in cluster: "
-                + _clusterName);
-      } catch (Exception e) {
-        throw new HelixException(e);
-      }
+    try {
+      _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), _sessionId);
+      LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance.getEphemeralOwner());
+    } catch (ZkSessionMismatchedException e) {
+      throw new HelixException(
+          "Failed to create live instance, path: " + liveInstancePath + ", expected session: "
+              + _sessionId, e);
     }
 
     ParticipantHistory history = getHistory();
     history.reportOnline(_sessionId, _manager.getVersion());
     persistHistory(history);
-
-    if (!liveInstance.getEphemeralOwner().equals(liveInstance.getSessionId())) {
-      LOG.warn(
-          "Session ID {} (Deprecated) in the znode does not match the Ephemeral Owner session ID {}. Will use the Ephemeral Owner session ID.",
-          liveInstance.getSessionId(), liveInstance.getEphemeralOwner());
-    }
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -23,8 +23,10 @@ import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 import org.I0Itec.zkclient.DataUpdater;
+import org.I0Itec.zkclient.exception.ZkNodeExistsException;
 import org.apache.helix.AccessOption;
 import org.apache.helix.BaseDataAccessor;
 import org.apache.helix.ConfigAccessor;
@@ -49,6 +51,7 @@ import org.apache.helix.model.StateModelDefinition;
 import org.apache.helix.model.builder.HelixConfigScopeBuilder;
 import org.apache.helix.participant.StateMachineEngine;
 import org.apache.helix.participant.statemachine.ScheduledTaskStateModelFactory;
+import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -99,12 +102,10 @@ public class ParticipantManager {
 
   /**
    * Handles a new session for a participant. The new session's id is passed in when participant
-   * manager is created, as it is required in
-   * {@link org.apache.helix.manager.zk.zookeeper.ZkClient#createEphemeral(String, Object, String)}
-   * to prevent ephemeral node creation from session race condition: ephemeral node is created by an
-   * expired or unexpected session.
+   * manager is created, as it is required to prevent ephemeral node creation from session race
+   * condition: ephemeral node is created by an expired or unexpected session.
    *
-   * @throws Exception if any exception occurs
+   * @throws Exception
    */
   public void handleNewSession() throws Exception {
     // Check zk session of this participant is still valid.
@@ -128,6 +129,8 @@ public class ParticipantManager {
 
     // TODO create live instance node after all the init works done --JJ
     // This will help to prevent controller from sending any message prematurely.
+    // Live instance creation also checks if the expected session is valid or not. Live instance
+    // should not be created by an expired zk session.
     createLiveInstance();
     carryOverPreviousCurrentState();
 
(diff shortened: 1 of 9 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1099,8 +1098,14 @@ public void handleStateChanged(KeeperState state) {
 
   @Override
   public void handleNewSession() throws Exception {
```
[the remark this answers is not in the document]
pkuwm: I was hesitated to do so as there are other places using the api like MockParticipantManager and ZkTestHelper. And also it is a public API. I thought we could deprecate it after we change the usage in MockParticipantManager and ZkTestHelper. That may bring in more changes in tests. Not sure whether or not we want to do it in another PR. We can discuss.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1129,13 +1134,13 @@ public void handleNewSession() throws Exception {
 
     switch (_instanceType) {
     case PARTICIPANT:
-      handleNewSessionAsParticipant();
+      handleNewSessionAsParticipant(sessionId);
       break;
     case CONTROLLER:
       handleNewSessionAsController();
```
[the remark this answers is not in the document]
pkuwm: Sure will add TODO.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1162,14 +1167,15 @@ public void handleNewSession() throws Exception {
...
+  void handleNewSessionAsParticipant(final String sessionId) throws Exception {
     if (_participantManager != null) {
       _participantManager.reset();
     }
     _participantManager =
         new ParticipantManager(this, _zkclient, _sessionTimeout, _liveInstanceInfoProvider,
             _preConnectCallbacks);
-    _participantManager.handleNewSession();
+
+    _participantManager.handleNewSession(sessionId);
```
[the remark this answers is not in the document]
pkuwm: The cached _sessionId may change and may be different with the passed in sessionId.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -165,6 +166,27 @@ public void recordPathStat(Stat stat, OptionalLong notificationTime) {
     }
   }
 
+  private abstract class SessionAwareCallable<T> implements Callable<T> {
```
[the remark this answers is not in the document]
pkuwm: Agreed.. A lot of comments to explain it.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -803,12 +872,16 @@ private Stat getStat(final String path, final boolean watch) {
   }
 
   protected void processStateChanged(WatchedEvent event) {
```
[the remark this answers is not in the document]
pkuwm: I think it can be deprecated as it is protected and nowhere else is using it.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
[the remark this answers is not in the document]
pkuwm: Yes, it is doable. I don't have a preference, but it seems a bit better to make it final and initialize it in the constructor.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -531,12 +558,29 @@ public String create(final String path, Object datat, final List<ACL> acl, final
...
+
+      /*
+       * We pass different implementation of callable.
+       * 1. If the session id is null, it means the operation is NOT session aware.
+       * In this case, we will create a regular callable to create an ephemeral node.
+       * 2. Otherwise, the operation is session aware. We will use a SessionAwareCallable to
+       * create an ephemeral node, which is guaranteed to be created in the expected session.
+       */
+      String actualPath;
+      if (sessionId == null) {
```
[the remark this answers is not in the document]
pkuwm: I understand your concern. My initial idea was to keep minimal changes. There are other createEphemeral() APIs that don't pass a sessionId. My thought was to keep the those APIs for now and keep the logic for them.

If we need to consider session aware for all createEphemeral() APIs, we have to add new APIs with sessionId createEphemeral(path, acl, sessionId) and change the code logic, which would make this PR bigger.

To make this logic more accurate, maybe we can do:
```
if (sessionId != null && mode == CreateMode.EPHEMERAL) {
   SessionAware;
} esle {
   NonSessionAware;
}
```

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -521,6 +543,11 @@ public String create(final String path, Object data, final CreateMode mode)
    */
   public String create(final String path, Object datat, final List<ACL> acl, final CreateMode mode)
       throws IllegalArgumentException, ZkException {
+    return create(path, datat, acl, mode, null);
+  }
+
+  private String create(final String path, Object datat, final List<ACL> acl, final CreateMode mode,
+      final String sessionId) throws IllegalArgumentException, ZkException {
```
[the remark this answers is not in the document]
pkuwm: Right, I had the same idea. On the other side, I realized that we had metrics recording later in create() and retryUntilConnected(). If we check session id earlier, that means createEphemeral fails and we also need to record metrics. Thus adds more code to this early checking block. At the same time, I didn't think checking in retryUntilConnected() takes significant overhead (it will create a callable object, of course). So I decided not to do early check.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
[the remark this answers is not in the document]
jiajunwang: Checking both conditions looks like a better option for me.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -158,7 +162,7 @@ private void joinCluster() {
     }
   }
 
-  private void createLiveInstance() {
+  private void createLiveInstance(final String sessionId) {
```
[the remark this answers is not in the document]
jiajunwang: I don't think so. If you pass the _sessionId as a parameter, even it is changed later at random time by the manager logic, the parameter won't be changed. Note that you are passing the String.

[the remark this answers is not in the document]
jiajunwang: Discussed offline, it seems my purpose will leave a gap which might cause a race condition. It seems that we have to add this new interface.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkSessionAwareStateListener.java
```diff
@@ -0,0 +1,27 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import org.I0Itec.zkclient.IZkStateListener;
+
+
+public interface ZkSessionAwareStateListener extends IZkStateListener {
```
jiajunwang: As discussed offline, we don't need this new interface. Instead, let's create a private listener class with the same logic to simplify the code.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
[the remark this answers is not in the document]
narendly: My 2 cents from our discussion offline.

Since we've had some issues with keeping the sessionId as an internal state (like _sessionId as a private variable), I prefer to take it in as the method parameter to do away with any state. This way, we don't treat sessionId as an internal state and its lifecycle will coincide with that of the method.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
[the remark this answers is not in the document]
narendly: Should we require ZooKeeper in the constructor as well?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -165,6 +166,27 @@ public void recordPathStat(Stat stat, OptionalLong notificationTime) {
     }
   }
 
+  private abstract class SessionAwareCallable<T> implements Callable<T> {
+    protected volatile String sessionId;
+    protected volatile ZooKeeper zookeeper;
```
narendly: When this Callable gets instantiated with a constructor that takes in both sessionId and zookeeper, please make sure you validate them as well. We could even do the check for whether sessionId given and ZooKeeper's sessionId match in the constructor and fail early?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -165,6 +166,27 @@ public void recordPathStat(Stat stat, OptionalLong notificationTime) {
...
+
+    public String getSessionId() {
+      return sessionId;
+    }
+
+    public void setZookeeper(ZooKeeper zk) {
+      zookeeper = zk;
+    }
+
+    public ZooKeeper getZookeeper() {
```
narendly: Are getters really necessary?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
narendly: Actually, let's be careful here. We should only fail if mode is ephemeral.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -568,6 +612,31 @@ public void createEphemeral(final String path, final Object data)
     create(path, data, CreateMode.EPHEMERAL);
   }
 
+  /**
+   * Create a session aware ephemeral node. Given an expected non-zero session id, if the ephemeral
```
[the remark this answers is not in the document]
narendly: Nit: Probably should check for an empty string as well.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -858,21 +931,30 @@ private void reconnect() {
     }
   }
 
-  private void fireNewSessionEvents() {
+  private void fireNewSessionEvents(final String sessionId) {
```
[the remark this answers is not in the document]
narendly: +1. Or put a clear TODO and create a corresponding issue and link it here?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -858,21 +931,30 @@ private void reconnect() {
...
     for (final IZkStateListener stateListener : _stateListener) {
-      _eventThread.send(new ZkEvent("New session event sent to " + stateListener) {
+      _eventThread.send(new ZkEvent("New session event sent to " + stateListener, sessionId) {
```
narendly: Nit: String.format?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -858,21 +931,30 @@ private void reconnect() {
...
 
   protected void fireStateChangedEvent(final KeeperState state) {
+    fireStateChangedEvent(state, null);
+  }
+
+  private void fireStateChangedEvent(final KeeperState state, final String sessionId) {
     for (final IZkStateListener stateListener : _stateListener) {
-      _eventThread.send(new ZkEvent("State changed to " + state + " sent to " + stateListener) {
+      String description = "State changed to " + state + " sent to " + stateListener;
+      _eventThread.send(new ZkEvent(description, sessionId) {
```
narendly: Nit: String.format?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -97,10 +97,28 @@ public ParticipantManager(HelixManager manager, HelixZkClient zkclient, int sess
...
+  /**
+   * Handles a new session for a participant. The new session's id has to be passed in, as it is
+   * required in {@link #createLiveInstance(String)} to prevent ephemeral node creation from session
+   * race condition: ephemeral node is created by an expired or unexpected session.
+   *
+   * @param sessionId the new session's ID
+   * @throws Exception if any exception occurs
+   */
+  public void handleNewSession(final String sessionId) throws Exception {
     joinCluster();
```
kaisun2000: Before joinCluster, we can check if the sessionId is the same as current ZkClient sessionId. If it is not, no need to do anything except printing a log. 
This is an outdated one by the time of execution.

helix-core/src/main/java/org/apache/helix/manager/zk/client/HelixZkClient.java
```diff
@@ -97,20 +97,34 @@ default void unsubscribeStateChanges(IZkStateListener listener) {
...
 
+  void createEphemeral(final String path, final String sessionId);
+
   void createEphemeral(final String path, final List<ACL> acl);
 
+  void createEphemeral(final String path, final List<ACL> acl, final String sessionId);
+
   String create(final String path, Object data, final CreateMode mode);
 
-  String create(final String path, Object datat, final List<ACL> acl, final CreateMode mode);
```
kaisun2000: nit, not need to make this correction. This make confuse people. Either way, up to you.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -101,6 +101,10 @@ public ParticipantManager(HelixManager manager, HelixZkClient zkclient, int sess
    * @throws Exception
    */
   public void handleNewSession() throws Exception {
```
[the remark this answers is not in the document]
lei-xia: Anyone external customer uses this class? This class is supposed to be internal class, can you confirm it? We can directly remove the method if no external customer refers this class.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -182,7 +200,7 @@ private void createLiveInstance() {
     do {
       retry = false;
       try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord());
+        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), sessionId);
         LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance.getEphemeralOwner());
       } catch (ZkNodeExistsException e) {
```
lei-xia: Why do we catch other exception and throw out HelixException (like what we do in line 264-270?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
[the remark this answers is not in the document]
pkuwm: @lei-xia I did a quick check code search on LinkedIn's internal codebase. It doesn't seem to be used within LinkedIn. But not sure about external users outside linkedin. I am hesitated to just remove it since it is public. That is why we spent so much effort on backward compatibility. If we all agree on removing this deprecated method, I will remove it. @jiajunwang @kaisun2000 @narendly

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -240,7 +258,7 @@ private void createLiveInstance() {
      */
     if (retry) {
       try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord());
+        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), sessionId);
         LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance
             .getEphemeralOwner());
       } catch (Exception e) {
```
lei-xia: The error message is not quite accurate here, it may be caused by  other failures, let us catch NodeExistException separately. For other Exception, print out the exact cause from the exception.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -697,10 +706,16 @@ void createClient() throws Exception {
     int retryCount = 0;
     while (retryCount < 3) {
       try {
-        // TODO: synchronize this block and wait for the new non-zero session ID updated.
-        _zkclient.waitUntilConnected(_connectionInitTimeout, TimeUnit.MILLISECONDS);
-        handleStateChanged(KeeperState.SyncConnected);
-        handleNewSession();
+        // TODO: might be better to create a new method to return session id in waitUntilConnected.
+        synchronized (this) {
+          if (!_zkclient.waitUntilConnected(_connectionInitTimeout, TimeUnit.MILLISECONDS)) {
```
lei-xia: When waitUntilConnected return false, does it always mean it is timeout-ed (not other failures)?  Will this method throw out any other exceptions?  In general, we need to either catch all exceptions, or skip them to the upper-level calls, handle partial set of failure cases could be confusing.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -697,10 +706,16 @@ void createClient() throws Exception {
     int retryCount = 0;
     while (retryCount < 3) {
       try {
-        // TODO: synchronize this block and wait for the new non-zero session ID updated.
-        _zkclient.waitUntilConnected(_connectionInitTimeout, TimeUnit.MILLISECONDS);
-        handleStateChanged(KeeperState.SyncConnected);
-        handleNewSession();
+        // TODO: might be better to create a new method to return session id in waitUntilConnected.
+        synchronized (this) {
```
lei-xia: Also be caution waitUntilConnected() is a blocking call, make sure the lock is released within that method if it goes to sleep.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -980,13 +997,13 @@ void waitUntilConnected() {
         continue;
       }
 
-      _sessionId = Long.toHexString(_zkclient.getSessionId());
+      _sessionId = _zkclient.getHexSessionId(_zkclient.getSessionId());
```
lei-xia: Why do we need this?   if we really need a different method for Hex Id, then why not _zkClient.getHexSessionId directly return the Hex(_sessionId)? This embeded calls looks a bit weird

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1134,7 +1163,17 @@ public void handleNewSession(final String sessionId) throws Exception {
     // Wait until we get a non-zero session id. Otherwise, getSessionId() might be null.
     waitUntilConnected();
 
-    // TODO: filter out stale sessions here.
+    /*
+     * Filter out stale sessions. If a session id is not null and not the same as current session
+     * id, this session is expired. With this filtering, expired sessions are NOT handled,
+     * so performance is expected to improve.
+     */
+    if (sessionId != null && !getSessionId().equals(sessionId)) {
```
lei-xia: what happens if sessionId is null, will all following calls  function correctly if it is null? say can createLiveInstance() handle it properly if sessionId passed to it is null?

helix-core/src/main/java/org/apache/helix/manager/zk/client/HelixZkClient.java
```diff
@@ -172,6 +186,15 @@ void asyncSetData(final String path, Object datat, final int version,
...
   long getSessionId();
 
+  /**
+   * Gets a session id in hexadecimal notation from a session id in long type.
+   * Ex. 0x1000a5ceb930004 is returned.
+   *
+   * @param sessionId session id in long type
+   * @return String representation of session id in hexadecimal notation.
+   */
+  String getHexSessionId(long sessionId);
```
lei-xia: Either make it return directly, i.e, getHexSessionId(), otherwise, this is more like a util function (just doing a conversion, it may not  be  in ZkClient. or at least, if you think it should be util function somehow, then it should be defined something like:  static String toHexSessionId(long sessionId);

helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
```diff
@@ -202,7 +202,7 @@ public void process(WatchedEvent event) {
 
     String newSessionId = Long.toHexString(curZookeeper.getSessionId());
     LOG.info("After session expiry. sessionId: " + newSessionId + ", zk: " + curZookeeper);
-    Assert.assertNotSame(newSessionId, oldSessionId, "Fail to expire current session, zk: "
+    Assert.assertFalse(newSessionId.equals(oldSessionId), "Fail to expire current session, zk: "
```
lei-xia: What is difference here?

pkuwm: `assertNotSame` checks two objects don't refer to the same object. But here we actually would like to check if these two sessions are equaled or not. If we use `assertNotSame`, this would always pass because these two sessionId strings are different objects.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: Like what I mention in the comment `TODO: after removing I0ItecIZkStateListenerHelixImpl, null session should be checked and discarded.`, null is acceptable for backward compatibility. If null is passed in, we will keep the behavior as before, which may cause race condition.

kaisun2000:
One general comments:
Currently, the code for create session aware ephemeral node is looking good. 
However, we can't simply reply on curren ZKHellixManager code. 
The reason is that the handleNewSession delivery time is different. For non-shared client, we actually  deliver them only when synconnected state change is notified. (This is to get the session_id for ZkClient event queue). 

To validate ZKHelixManager working, we should add two different test cases, one for participant role, and the other for controller role. At session expiration and disconnect time, the handleNewSession would be delivered correctly. No missing, no duplication and thread-safe. 

Let us add these tests.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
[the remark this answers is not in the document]
pkuwm: I re-think about this. I don't think we need to change this method. I don't want to create a new method handleNewSession(sessionId). But would pass sessionId to the participant manager when creating the participant manager. Because each time ZkHelixManager.handleNewSessionAsParticipant() would create a new participant manager. I think it is safe to just pass sessionId to participant manager. So I could keep minimal changes, given the correctness.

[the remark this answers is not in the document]
lei-xia: we can not make sure we won't change methods from any public classes, that will mean we won't be able to change anything. We should treat this class as not a API, so it should be fine to change the method.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -77,14 +76,17 @@
   final LiveInstanceInfoProvider _liveInstanceInfoProvider;
   final List<PreConnectCallback> _preConnectCallbacks;
 
+  private final String _sessionId;
```
jiajunwang: This is probably OK, but one minor concern is none backward compatible that you change this field to private.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -97,10 +99,25 @@ public ParticipantManager(HelixManager manager, HelixZkClient zkclient, int sess
...
-   * @throws Exception
+   * Handles a new session for a participant. The new session's id is passed in when participant
+   * manager is created, as it is required in
+   * {@link org.apache.helix.manager.zk.zookeeper.ZkClient#createEphemeral(String, Object, String)}
+   * to prevent ephemeral node creation from session race condition: ephemeral node is created by an
+   * expired or unexpected session.
+   *
+   * @throws Exception if any exception occurs
    */
   public void handleNewSession() throws Exception {
```
jiajunwang: I might have been confused, but shouldn't we have a new handleNewSession method that takes a session Id as parameter?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
jiajunwang: Agreed, rollback is safer.
But consider the possible rollback fails condition, shall we try to create the new live instance node first? (If this is not the current behavior)

helix-core/src/main/java/org/apache/helix/manager/zk/ZkSessionMismatchedException.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public class ZkSessionMismatchedException extends ZkException {
+
+    private static final long serialVersionUID = 1L;
+
+    public ZkSessionMismatchedException() {
+        super();
+    }
+
+    public ZkSessionMismatchedException(KeeperException cause) {
```
jiajunwang: I don't think we will have a KeeperException to cause ZkSessionMismatchedException. Since ZkSessionMismatchedException is basically invented by us in this PR. So this construction might be unnecessary.

helix-core/src/main/java/org/apache/helix/manager/zk/client/HelixZkClient.java
```diff
@@ -172,6 +186,15 @@ void asyncSetData(final String path, Object datat, final int version,
...
   long getSessionId();
 
+  /**
+   * Gets the zookeeper's session id of this zk client in hexadecimal notation.
+   * This session could be valid, expired, or "0"(initial value before zk session is established).
+   * Ex. 1000a5ceb930004 is returned.
+   *
+   * @return String representation of session id in hexadecimal notation.
+   */
+  String getHexSessionId();
```
jiajunwang: FYI, I think we want to expose as little information as possible from this interface. We leave getSessionId mainly for backward compatibility. If there is no critical usage, please do not expose more internal session ID (which is changeable) to the ZkClient user.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -556,15 +644,55 @@ public String create(final String path, Object datat, final List<ACL> acl, final
...
-        public String call() throws Exception {
-          return getConnection().create(path, data, acl, mode);
+      final byte[] dataBytes = dataObject == null ? null : serialize(dataObject, path);
+      checkDataSizeLimit(dataBytes);
+
+      /*
+       * We pass different implementation of callable.
+       * 1. If the session id is null, it means the operation is NOT session aware.
+       * In this case, we will implement a regular callable to create a node.
+       * 2. Otherwise, the operation is session aware. We will use a  to
```
jiajunwang: I cannot read it. "We will use a WHAT to create...."

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -556,15 +644,55 @@ public String create(final String path, Object datat, final List<ACL> acl, final
...
+       * create a node which is guaranteed to be created in the expected session.
+       */
+      final String actualPath = retryUntilConnected(() -> {
+        ZooKeeper zooKeeper = ((ZkConnection) getConnection()).getZookeeper();
+
+        /*
+         * 1. If the session id is NOT null and create mode is EPHEMERAL or EPHEMERAL_SEQUENTIAL,
```
jiajunwang: Please do meaningful comments instead of reading the code for me. The conditions are very clear below.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -556,15 +644,55 @@ public String create(final String path, Object datat, final List<ACL> acl, final
...
-      final byte[] data = datat == null ? null : serialize(datat, path);
-      checkDataSizeLimit(data);
-      String actualPath = retryUntilConnected(new Callable<String>() {
-        @Override
-        public String call() throws Exception {
-          return getConnection().create(path, data, acl, mode);
+      final byte[] dataBytes = dataObject == null ? null : serialize(dataObject, path);
+      checkDataSizeLimit(dataBytes);
+
+      /*
```
jiajunwang: I feel it would be much cleaner if you split the session check and mode check related logic to a separate method. All the comments could be much simpler as well.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -835,8 +1080,28 @@ protected void processStateChanged(WatchedEvent event) {
...
+    if (event.getState() == KeeperState.SyncConnected) {
+      if (!_isNewSessionEventFired) {
+        /*
+         * Once the new zookeeper session receives the first SyncConnected state,
+         * the zookeeper session is established and connected to zookeeper service.
+         * So the session id is available and non-zero. Now we can fire new session events.
+         */
+        fireNewSessionEvents();
+        /*
+         * Set it true to avoid firing events again for the same session next time
```
jiajunwang: How to ensure this flag is for the "same session" only? I didn't see you check if it is the "same session", or reset the flag.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1166,6 +1432,12 @@ private void acquireEventLock() {
     }
     try {
       while (true) {
+        // Because ConnectionLossException and SessionExpiredException are caught but not thrown,
```
jiajunwang: nit, just my personal opinion, this is too verbose. This var is very clear to be recording and logging what exception we see during the operation.

pkuwm: We don't expose the exception in the message. Since we just need the code in the message, I will change it to a code type so we don't record the exception itself.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
pkuwm: Resetting the flag is after reconnect:
https://github.com/apache/helix/blob/b485d1f708e3308fa46a0ce88dda36607d93e186/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java#L1143-L1145

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -886,19 +1141,20 @@ private void reconnect() {
   }
 
   private void fireNewSessionEvents() {
+    final String sessionId = getHexSessionId(getSessionId());
```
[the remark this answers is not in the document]
pkuwm: Added a double check before firing new session events.
```
    if (event.getState() == KeeperState.SyncConnected) {
      if (!_isNewSessionEventFired && !"0".equals(getHexSessionId())) {
        /*
         * Before the new zookeeper instance is connected to the zookeeper service and its session
         * is established, its session id is 0.
         * New session event is not fired until the new zookeeper session receives the first
         * SyncConnected state(the zookeeper session is established).
         * Now the session id is available and non-zero, and we can fire new session events.
         */
        fireNewSessionEvents();
```

helix-core/src/test/java/org/apache/helix/manager/zk/TestRawZkClient.java
```diff
@@ -461,4 +467,247 @@ void testPendingRequestGauge() throws Exception {
...
+    try {
+      final CountDownLatch countDownLatch = new CountDownLatch(1);
+      final AtomicBoolean running = new AtomicBoolean(true);
+
+      final Thread creationThread = new Thread(() -> {
+        while (running.get()) {
+          try {
+            // Create ephemeral node in the expected session id.
+            zkClient.createEphemeral(path, data, expectedSessionId);
+          } catch (ZkTimeoutException e) {
```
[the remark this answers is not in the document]
pkuwm: the connectionloss keeper exception is not concerted to ZKTimeoutException. But each time before retrying, the operation time is checked. If timed out, ZKTimeoutException is thrown. And this is why I add the retry cause message to the exception: to help this test, as well.
```
        // before attempting a retry, check whether retry timeout has elapsed
        if (System.currentTimeMillis() - operationStartTime > _operationRetryTimeoutInMillis) {
          throw new ZkTimeoutException("Operation cannot be retried because of retry timeout ("
              + _operationRetryTimeoutInMillis + " milli seconds). Retry was caused by "
              + retryCauseCode);
        }
```

helix-core/src/test/java/org/apache/helix/manager/zk/TestRawZkClient.java
```diff
@@ -461,4 +467,247 @@ void testPendingRequestGauge() throws Exception {
...
+    }, 1000L));
+
+    try {
+      // Try to create ephemeral node with the original session.
+      // This creation should NOT be successful because the original session is already expired.
+      _zkClient.createEphemeral(path, "Hello Helix", originalHexSessionId);
+      Assert.fail("Ephemeral node should not be created by the expired session.");
+    } catch (ZkSessionMismatchedException expected) {
+      // Expected because there is a session mismatch.
+    }
```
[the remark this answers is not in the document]
pkuwm: The purpose of this test is to test a mismatched session won't create an ephemeral node. With the check logic, I think ZkSessionMismatchedException is expected. So I don't think we need to fail this test if any other exception is thrown.

helix-core/src/test/java/org/apache/helix/manager/zk/TestRawZkClient.java
```diff
@@ -461,4 +467,247 @@ void testPendingRequestGauge() throws Exception {
...
+    Assert.assertFalse(_zkClient.exists(path));
+
+    // Expire the original session.
+    ZkTestHelper.expireSession(_zkClient);
+
+    // Wait until the ZkClient has got a new session.
+    Assert.assertTrue(TestHelper.verify(() -> {
+      try {
+        // New session id should be greater than expired session id.
+        return _zkClient.getSessionId() > originalSessionId;
```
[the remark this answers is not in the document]
pkuwm: Yes. I reused the old code. Corrected it.

helix-core/src/test/java/org/apache/helix/manager/zk/TestRawZkClient.java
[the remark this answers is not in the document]
kaisun2000: ` try { }
catch (ZkSessionMismatchedException expected) {
} 
catch (Exception eNotExpected) {
Assert.fail("unexpected exception" + e);
}`
If other exception happens, there is something wrong. You want to catch it. Otherwise, it is deemed as right, which is not the case.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
kaisun2000: @jiajunwang, let us document this invariant so that people who later change the code knows what is the assumption here.

A lot of times later people who wrote the code would forget about the invariant they assumed when they wrote the code.

helix-core/src/test/java/org/apache/helix/manager/zk/TestRawZkClient.java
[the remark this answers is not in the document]
pkuwm: It is fine. Added. Thanks.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -697,10 +706,16 @@ void createClient() throws Exception {
...
+        // TODO: might be better to create a new method to return session id in waitUntilConnected.
+        synchronized (this) {
+          if (!_zkclient.waitUntilConnected(_connectionInitTimeout, TimeUnit.MILLISECONDS)) {
+            throw new ZkTimeoutException(
+                "Unable to connect to zookeeper server within timeout: " + _connectionInitTimeout
+                    + " ms.");
+          }
+          handleStateChanged(KeeperState.SyncConnected);
+          handleNewSession(_zkclient.getHexSessionId(_zkclient.getSessionId()));
+        }
```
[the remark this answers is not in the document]
kaisun2000: Walked through the code offline together. Yes, pkuwm is right. 

For dedicated ZKClient, the first synconnected state change would be ignored.

This is the current behavior. We may revisit this assertion in Horizontal Scaling effort. Let us document this one in ZkClient constructor managingZkConnection part.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
kaisun2000: can _zkClient.getSessionId() == 0? 

I think they can. We should do nothing here then.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
kaisun2000: For ZkSesssionMismatch Exception, we should not throw, we should just cleanup (if there is anything) and then return.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -182,7 +200,7 @@ private void createLiveInstance() {
     do {
       retry = false;
       try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord());
+        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), sessionId);
         LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance.getEphemeralOwner());
       } catch (ZkNodeExistsException e) {
         LOG.warn("found another instance with same instanceName: " + _instanceName + " in cluster "
```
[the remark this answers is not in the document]
pkuwm: Right. Completed in a later commit.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
pkuwm: If we don't throw exception and just return, createLiveInstance may be treated as success and so carryOverPreviousCurrentState and setupMsgHandler will run. So I would like to follow the original logic to throw an exception if the ephemeral creation fails. The helix exception would be handled by ZkEventThread's run method: log the event's running failure.

pkuwm:
> One general comments:
> Currently, the code for create session aware ephemeral node is looking good.
> However, we can't simply reply on curren ZKHellixManager code.
> The reason is that the handleNewSession delivery time is different. For non-shared client, we actually deliver them only when synconnected state change is notified. (This is to get the session_id for ZkClient event queue).
> 
> To validate ZKHelixManager working, we should add two different test cases, one for participant role, and the other for controller role. At session expiration and disconnect time, the handleNewSession would be delivered correctly. No missing, no duplication and thread-safe.
> 
> Let us add these tests.

Synced up with @kaisun2000.

- For ADMINISTRATOR type using shared zkclient, handleNewSessionAsParticipant or handleNewSessionAsController is not run in handle new session. There is not meaningful things to do in handle new session. So it is fine for shared zkclient.
- For the other types using dedicated zkclient, existing and newly added tests already cover the cases.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -97,10 +98,25 @@ public ParticipantManager(HelixManager manager, HelixZkClient zkclient, int sess
   }
 
   /**
-   * Handle new session for a participang.
-   * @throws Exception
+   * Handles a new session for a participant. The new session's id is passed in when participant
+   * manager is created, as it is required in
+   * {@link org.apache.helix.manager.zk.zookeeper.ZkClient#createEphemeral(String, Object, String)}
```
jiajunwang: The explanation is good here, but this link to the code is not necessary. And maybe hard for us to maintain.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -178,89 +194,18 @@ private void createLiveInstance() {
       }
     }
 
-    boolean retry;
-    do {
-      retry = false;
-      try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord());
-        LOG.info("LiveInstance created, path: " + liveInstancePath + ", sessionId: " + liveInstance.getEphemeralOwner());
-      } catch (ZkNodeExistsException e) {
```
jiajunwang: Although it should be impossible that we see this exception anymore, shall we still catch it and throw HelixException?

jiajunwang: Or we can keep the wait and retry logic.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -697,10 +699,19 @@ void createClient() throws Exception {
     int retryCount = 0;
     while (retryCount < 3) {
       try {
-        // TODO: synchronize this block and wait for the new non-zero session ID updated.
-        _zkclient.waitUntilConnected(_connectionInitTimeout, TimeUnit.MILLISECONDS);
+        // TODO: get session id from waitUntilConnected to avoid race condition
```
jiajunwang: This is OK. Just a reminder that we need to finish this TODO until the bug is finished.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1134,9 +1159,31 @@ public void handleNewSession(final String sessionId) throws Exception {
...
     waitUntilConnected();
 
-    // TODO: filter out stale sessions here.
+    /*
+     * Filter out stale sessions. If a session id is not null and not the same as current session
+     * id, this session is expired. With this filtering, expired sessions are NOT handled,
+     * so performance is expected to improve.
+     */
+    if (sessionId != null && !getSessionId().equals(sessionId)) {
+      LOG.info("Session is expired and not handled. Expected: {}. Actual: {}.", sessionId,
```
jiajunwang: nit, LOG.warn

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1134,9 +1159,31 @@ public void handleNewSession(final String sessionId) throws Exception {
...
+    /*
+     * When a null session id is passed in, we will take current session's id for following
+     * operations. Please note that current session might not be the one we expect to handle,
+     * because the one we expect might be already expired when the zk event is waiting in the
+     * event queue. Why we use current session here is for backward compatibility with the old
+     * method handleNewSession().
+     */
+    if (sessionId == null) {
+      sessionId = getSessionId();
+      LOG.warn("Session id: <null> is passed in. Current session id: {} will be used.", sessionId);
```
jiajunwang: nit, LOG.debug?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -703,8 +939,6 @@ public void process(WatchedEvent event) {
         if (event.getState() == KeeperState.Expired) {
           getEventLock().getZNodeEventCondition().signalAll();
           getEventLock().getDataChangedCondition().signalAll();
-          // We also have to notify all listeners that something might have changed
-          fireAllEvents();
```
jiajunwang: It is in the finally block. With this change, when getShutdownTrigger() == true, some notifications might be missed.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
pkuwm: Sure. Will address this link.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
pkuwm: The wait and retry logic is kept.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: Will submit a new PR for this.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: Resolved.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: Resolved

pkuwm:
Test results:

[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestAddNodeAfterControllerStart.testDistributed:165 » ThreadTimeout Method org...
[INFO]
[ERROR] Tests run: 896, Failures: 2, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:05 h
[INFO] Finished at: 2020-01-07T18:48:16-08:00


[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 7.891 s - in org.apache.helix.integration.TestAddNodeAfterControllerStart
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  12.397 s
[INFO] Finished at: 2020-01-07T21:02:37-08:00


[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 10.03 s - in org.apache.helix.integration.task.TestJobQueueCleanUp
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  14.887 s
[INFO] Finished at: 2020-01-07T21:13:52-08:00

pkuwm:
This PR is ready to be merged, approved by @kaisun2000 @jiajunwang 

Final commit message:
```
There is zk session race condition in zk helix manager. Zk session may expire and change during handling new session. If the expected session is expired when handling new session, the ephemeral node should NOT be created.

Change list:
- Change API create() in ZkClient to accept zk sessionId as a parameter. Add a new public API createEphemeral(final String path, final Object data, final String sessionId) to create an ephemeral node by the expected zk session.
- Fire new session events and all events(child/data change events) to right after the first SyncConnected state is received in zk client.
- Filter out stale sessions in handling new session.
- Add multiple unit tests to cover the cases.
```

jiajunwang:
TestAddNodeAfterControllerStart is rarely seen in the failure test list. Could you please run this test multiple times (use the scripts/runSingleTest.sh), or the whole mvn tests again if possible to ensure we are not introducing any problem with the change?

pkuwm:
I did run this test 10 times which all passed when I was surprised to see
this failure first time. Let me check the result of my recent run tonight.

PR #647, opened 2019-12-09
Add java API to create cluster with CloudConfig

### LINKED ISSUE #646: TODO: Add relevant java Level APIs for cloudConfig creation
The java related APIs for Cluster Creation with CloudConfig need to be added.

alirezazamani:
### Issues
Fixes #646.

### Description
In this commit the below APIs have been added.
1- AddCluster with CloudCOnfig API.
2- addCloudConfig to existing cluster.
3- Update CloudConfig to update existing Cloud Config

Several tests have been added to validate these APIs.

### Tests

New tests have been added.
TestConfigAccessor.testUpdateCloudConfig
TestZkHelixAdmin.testAddCloudConfig
TestClusterSetup.testAddClusterWithInvalidCloudConfig
TestClusterSetup.testAddClusterWithValidCloudConfig
TestClusterSetup.testAddClusterAzureProvider

Test Result 1: mvn test
[INFO] Tests run: 900, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3,228.171 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 900, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  53:52 min
[INFO] Finished at: 2019-12-09T13:52:08-08:00

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
@@ -591,7 +591,7 @@ public class ConfigAccessor {
    * @return The instance of {@link CloudConfig}
    */
   public CloudConfig getCloudConfig(String clusterName) {
-    if (!ZKUtil.isClusterSetup(clusterName, zkClient)) {
+    if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
       throw new HelixException(
           String.format("Failed to get config. cluster: %s is not setup.", clusterName));
     }
@@ -604,7 +604,7 @@ public class ConfigAccessor {
       return null;
     }
 
-    return new CloudConfig(record);
+    return new CloudConfig.Builder(record).build();
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
@@ -22,7 +22,7 @@ package org.apache.helix;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
-
+import org.apache.helix.model.CloudConfig;
 import org.apache.helix.model.ClusterConstraints;
 import org.apache.helix.model.ClusterConstraints.ConstraintType;
 import org.apache.helix.model.ConstraintItem;
@@ -379,6 +379,19 @@ public interface HelixAdmin {
    * @param resourceName
    */
   void dropResource(String clusterName, String resourceName);
+
+  /**
+   * Add cloud config to the cluster.
+   * @param clusterName
+   * @param cloudConfig
+   */
+  void addCloudConfig(String clusterName, CloudConfig cloudConfig);
+
+  /**
+   * Remove the Cloud Config for specific cluster
+   * @param clusterName
+   */
+  void removeCloudConfig(String clusterName);
 
   /**
    * Get a list of state model definitions in a cluster
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -59,6 +59,7 @@ import org.apache.helix.controller.rebalancer.strategy.CrushEdRebalanceStrategy;
 import org.apache.helix.controller.rebalancer.strategy.RebalanceStrategy;
 import org.apache.helix.manager.zk.client.HelixZkClient;
 import org.apache.helix.manager.zk.client.SharedZkClientFactory;
+import org.apache.helix.model.CloudConfig;
 import org.apache.helix.model.ClusterConfig;
 import org.apache.helix.model.ClusterConstraints;
 import org.apache.helix.model.ClusterConstraints.ConstraintType;
@@ -1023,6 +1024,33 @@ public class ZKHelixAdmin implements HelixAdmin {
 
     accessor.removeProperty(keyBuilder.idealStates(resourceName));
     accessor.removeProperty(keyBuilder.resourceConfig(resourceName));
+  }
+
+  @Override
+  public void addCloudConfig(String clusterName, CloudConfig cloudConfig) {
+    logger.info("Add CloudConfig to cluster {}, CloudConfig is {}.", clusterName,
+        cloudConfig.toString());
+
+    if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
+      throw new HelixException("cluster " + clusterName + " is not setup yet");
+    }
+
+    CloudConfig.Builder builder = new CloudConfig.Builder(cloudConfig);
+    CloudConfig cloudConfigBuilder = builder.build();
+
+    ZKHelixDataAccessor accessor =
+        new ZKHelixDataAccessor(clusterName, new ZkBaseDataAccessor<ZNRecord>(_zkClient));
+    Builder keyBuilder = accessor.keyBuilder();
+    accessor.setProperty(keyBuilder.cloudConfig(), cloudConfigBuilder);
+  }
+
+  @Override
+  public void removeCloudConfig(String clusterName) {
+    logger.info("Remove Cloud Config for cluster {}.", clusterName);
+    HelixDataAccessor accessor =
+        new ZKHelixDataAccessor(clusterName, new ZkBaseDataAccessor<ZNRecord>(_zkClient));
+    Builder keyBuilder = accessor.keyBuilder();
+    accessor.removeProperty(keyBuilder.cloudConfig());
   }
 
   @Override
--- a/helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
+++ b/helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
@@ -30,6 +30,9 @@ import org.apache.helix.cloud.constants.CloudProvider;
  * Cloud configurations
  */
 public class CloudConfig extends HelixProperty {
+
+  public static final String CLOUD_CONFIG_KW = "CloudConfig";
+
   /**
    * Configurable characteristics of a cloud.
    * NOTE: Do NOT use this field name directly, use its corresponding getter/setter in the
@@ -52,39 +55,22 @@ public class CloudConfig extends HelixProperty {
 
   /**
    * Instantiate the CloudConfig for the cloud
-   * @param cluster
    */
-  public CloudConfig(String cluster) {
-    super(cluster);
+  private CloudConfig() {
+    super(CLOUD_CONFIG_KW);
   }
 
   /**
-   * Instantiate with a pre-populated record
-   * @param record a ZNRecord corresponding to a cloud configuration
+   * The constructor from the ZNRecord.
+   * @param record
    */
-  public CloudConfig(ZNRecord record) {
-    super(record);
+  private CloudConfig(ZNRecord record) {
+    super(CLOUD_CONFIG_KW);
+    _record.setSimpleFields(record.getSimpleFields());
+    _record.setListFields(record.getListFields());
+    _record.setMapFields(record.getMapFields());
   }
 
-  /**
-   * Instantiate the config using each field individually.
-   * Users should use CloudConfig.Builder to create CloudConfig.
-   * @param cluster
-   * @param enabled
-   * @param cloudID
-   */
-  public CloudConfig(String cluster, boolean enabled, CloudProvider cloudProvider, String cloudID,
-      List<String> cloudInfoSource, String cloudProcessorName) {
-    super(cluster);
-    _record.setBooleanField(CloudConfigProperty.CLOUD_ENABLED.name(), enabled);
-    _record.setSimpleField(CloudConfigProperty.CLOUD_PROVIDER.name(), cloudProvider.name());
-    _record.setSimpleField(CloudConfigProperty.CLOUD_ID.name(), cloudID);
-    if (cloudProvider.equals(CloudProvider.CUSTOMIZED)) {
-      _record
-          .setSimpleField(CloudConfigProperty.CLOUD_INFO_PROCESSOR_NAME.name(), cloudProcessorName);
-      _record.setListField(CloudConfigProperty.CLOUD_INFO_SOURCE.name(), cloudInfoSource);
-    }
-  }
 
   /**
    * Enable/Disable the CLOUD_ENABLED field.
@@ -135,28 +121,11 @@ public class CloudConfig extends HelixProperty {
   }
 
   /**
-   * Set the CLOUD_INFO_PROCESSOR_NAME field.
-   * @param cloudInfoProcessorName
-   */
-  public void setCloudInfoFProcessorName(String cloudInfoProcessorName) {
-    _record.setSimpleField(CloudConfigProperty.CLOUD_INFO_PROCESSOR_NAME.name(),
-        cloudInfoProcessorName);
-  }
-
-  /**
    * Get the CLOUD_INFO_PROCESSOR_NAME field.
    * @return CLOUD_INFO_PROCESSOR_NAME field.
    */
   public String getCloudInfoProcessorName() {
     return _record.getSimpleField(CloudConfigProperty.CLOUD_INFO_PROCESSOR_NAME.name());
-  }
-
-  /**
-   * Set the CLOUD_PROVIDER field.
-   * @param cloudProvider
-   */
-  public void setCloudProvider(CloudProvider cloudProvider) {
-    _record.setSimpleField(CloudConfigProperty.CLOUD_PROVIDER.name(), cloudProvider.name());
   }
 
   /**
@@ -167,32 +136,28 @@ public class CloudConfig extends HelixProperty {
     return _record.getSimpleField(CloudConfigProperty.CLOUD_PROVIDER.name());
   }
 
+
   public static class Builder {
-    private String _clusterName = null;
-    private CloudProvider _cloudProvider;
-    private boolean _cloudEnabled = DEFAULT_CLOUD_ENABLED;
-    private String _cloudID;
-    private List<String> _cloudInfoSources;
-    private String _cloudInfoProcessorName;
+    private ZNRecord _record;
 
     public CloudConfig build() {
       validate();
-      return new CloudConfig(_clusterName, _cloudEnabled, _cloudProvider, _cloudID,
-          _cloudInfoSources, _cloudInfoProcessorName);
+      return new CloudConfig(_record);
     }
 
     /**
      * Default constructor
      */
     public Builder() {
+      _record = new ZNRecord(CLOUD_CONFIG_KW);
     }
 
     /**
-     * Constructor with Cluster Name as input
-     * @param clusterName
+     * Instantiate with a pre-populated record
+     * @param record a ZNRecord corresponding to a cloud configuration
      */
-    public Builder(String clusterName) {
-      _clusterName = clusterName;
+    public Builder(ZNRecord record) {
+      _record = record;
     }
 
     /**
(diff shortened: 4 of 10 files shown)


helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -176,6 +179,30 @@ public void close() {
     }
   }
 
+  public void addCluster(String clusterName, boolean overwritePrevious, boolean isCloudEnabled)
```
zhangmeng916: Thinking this thoroughly, I believe we need more parameters than "isCloudEnabled". cloudID is dynamically generated every time, and we do not have default value for it. With that said, it makes more sense for users to be able to input "cloudProvider" and "cloudId" when the cloud is "enabled". We should assume that we always have "Azure" as default.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -176,6 +179,30 @@ public void close() {
...
+      throws HelixException {
+    if (isCloudEnabled) {
+      String cloudID = System.getProperty(SystemPropertyKeys.CLOUD_ID);
```
zhangmeng916: Let's discuss where the system property should be used. If you call the property key here, it means that the default value for Azure is put in the cfg2 of Helix-rest, which is not quite proper. These cloud features are not part of Helix Rest configuration. Ideally it should be in participant side, and put in our client's code cfg2 (e.g. Espresso storage node). 
In cluster creation, we just populate whatever we have into Zookeeper. The remaining is participant's work.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
pkuwm: Just my 2 cents: to me, passing a boolean param to a method to determine the behavior is kind of code smell.
See https://martinfowler.com/bliki/FlagArgument.html
Can we have a better design for this?

alirezazamani:
This PR is pending for further discussion...

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1025,6 +1026,23 @@ public void dropResource(String clusterName, String resourceName) {
     accessor.removeProperty(keyBuilder.resourceConfig(resourceName));
   }
 
+  @Override
+  public void addCloudConfig(String clusterName, CloudConfig cloudConfig) {
+    logger.info("Add CloudConfig to cluster {}.", clusterName);
```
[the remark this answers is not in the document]
alirezazamani: Done. Some of the fields ave been logged. Other fields such as ProcessorName and Sources have not been added since they might be null and logging them can cause an exception.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
alirezazamani: Fixed. Also tight now we have more input parameter.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
alirezazamani: Fixed based on our discussion.

helix-core/src/test/java/org/apache/helix/tools/TestClusterSetup.java
```diff
@@ -436,4 +441,82 @@ public void testDisableResource() throws Exception {
...
+
+
+  @Test(dependsOnMethods = "testAddClusterWithValidCloudConfig")
+  public void testAddClusterAzureProvider() throws Exception {
+    String className = TestHelper.getTestClassName();
+    String methodName = TestHelper.getTestMethodName();
+    String clusterName = className + "_" + methodName;
+
+    System.setProperty(SystemPropertyKeys.CLOUD_ID, "TestID");
+    System.setProperty(SystemPropertyKeys.CLOUD_INFO_SOURCE, "TestURL");
```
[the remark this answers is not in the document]
alirezazamani: Fixed.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -176,6 +179,37 @@ public void close() {
...
+        CloudConfig.Builder builder = new CloudConfig.Builder(clusterName);
+        builder.setCloudEnabled(true);
+        builder.setCloudID(cloudID);
+        builder.setCloudProvider(CloudProvider.valueOf(cloudProvider));
+        builder.setCloudInfoSources(cloudInfoSource);
+        builder.setCloudInfoProcessorName(cloudProcessorName);
+
+        // The builder checks the fields and generate exception with proper message.
+        CloudConfig validatedCloudConfig = builder.build();
+        addCluster(clusterName, overwritePrevious);
```
zhangmeng916: You can combine this "addCluster" with the following one.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -176,6 +179,37 @@ public void close() {
...
+        _admin.addCloudConfig(clusterName, validatedCloudConfig);
+      } else {
+        throw new HelixException(
+            "Cannot add the cluster, cloud isn enabled and cloudConfig is null!");
```
zhangmeng916: typo here

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1025,6 +1026,25 @@ public void dropResource(String clusterName, String resourceName) {
     accessor.removeProperty(keyBuilder.resourceConfig(resourceName));
   }
 
+  @Override
+  public void addCloudConfig(String clusterName, CloudConfig cloudConfig) {
+    logger.info("Add CloudConfig to cluster {}, CloudEnabled {}, CloudID {}, CloudProvider{}",
```
zhangmeng916: We should log the whole config instead of only 3 fields.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -176,6 +179,37 @@ public void close() {
...
+        // Validate the cloudConfig in the builder
+        // If it is invalid, do not even create a cluster
+        CloudConfig.Builder builder = new CloudConfig.Builder(clusterName);
+        builder.setCloudEnabled(true);
+        builder.setCloudID(cloudID);
+        builder.setCloudProvider(CloudProvider.valueOf(cloudProvider));
+        builder.setCloudInfoSources(cloudInfoSource);
+        builder.setCloudInfoProcessorName(cloudProcessorName);
+
+        // The builder checks the fields and generate exception with proper message.
```
zhangmeng916: You don't need this comment here, as the "validation" is already mentioned in the previous comment.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -176,6 +179,37 @@ public void close() {
...
+      if (cloudConfig != null) {
+        String cloudID = cloudConfig.getCloudID();
+        String cloudProvider = cloudConfig.getCloudProvider();
+        List<String> cloudInfoSource = cloudConfig.getCloudInfoSources();
+        String cloudProcessorName = cloudConfig.getCloudInfoProcessorName();
+
+        // Validate the cloudConfig in the builder
+        // If it is invalid, do not even create a cluster
+        CloudConfig.Builder builder = new CloudConfig.Builder(clusterName);
+        builder.setCloudEnabled(true);
```
zhangmeng916: It's better to use chained builder so that you only need to have one line of code. I noticed some existing ones don't use chain. But for concise code, let's use it.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
alirezazamani: fixed.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
alirezazamani: Fixed.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
alirezazamani: Synced offline. Fixed.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -176,6 +179,32 @@ public void close() {
     }
   }
 
+  public void addCluster(String clusterName, boolean overwritePrevious, boolean isCloudEnabled,
```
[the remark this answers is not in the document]
alirezazamani: Addressed.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -673,6 +673,41 @@ private void updateClusterConfig(String clusterName, ClusterConfig clusterConfig
...
+    }
+
+    HelixConfigScope scope =
+        new HelixConfigScopeBuilder(ConfigScopeProperty.CLOUD).forCluster(clusterName).build();
+    String zkPath = scope.getZkPath();
+
+    if (overwrite) {
+      ZKUtil.createOrReplace(_zkClient, zkPath, cloudConfig.getRecord(), true);
+    } else {
+      ZKUtil.createOrUpdate(_zkClient, zkPath, cloudConfig.getRecord(), true, true);
```
jiajunwang: Any case, now or in the future, we need to support updates? I don't think so. So shall we just simplify the code and get rid of the private method and the overwrite option?

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -380,6 +380,13 @@ void addStateModelDef(String clusterName, String stateModelDef, StateModelDefini
    */
   void dropResource(String clusterName, String resourceName);
 
+  /**
+   *
+   * @param clusterName
+   * @param cloudConfig
+   */
+  void addCloudConfig(String clusterName, CloudConfig cloudConfig);
```
jiajunwang: So we never remove the config?
For a complete operation set, I'd prefer we have a delete cluster config interface. Or call it setCloudConfig so it can be removed by passing null parameter.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -207,6 +207,16 @@ public Builder(CloudConfig cloudConfig) {
       _cloudInfoProcessorName = cloudConfig.getCloudInfoProcessorName();
     }
 
+    public Builder(String cluster, boolean enabled, CloudProvider cloudProvider, String cloudID,
+        List<String> cloudInfoSource, String cloudProcessorName) {
+      _clusterName = cluster;
```
jiajunwang: nit, let's try to keep the parameter names the same with the fields' names.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -207,6 +207,16 @@ public Builder(CloudConfig cloudConfig) {
       _cloudInfoProcessorName = cloudConfig.getCloudInfoProcessorName();
     }
 
+    public Builder(String cluster, boolean enabled, CloudProvider cloudProvider, String cloudID,
```
jiajunwang: Why we need this constructor while all the fields can be specified with the specific methods?

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -176,6 +179,31 @@ public void close() {
...
+        List<String> cloudInfoSources = cloudConfig.getCloudInfoSources();
+        String cloudProcessorName = cloudConfig.getCloudInfoProcessorName();
+
+        // Validate the cloudConfig in the builder
+        // If it is invalid, do not even create a cluster
+        CloudConfig.Builder builder = new CloudConfig.Builder(clusterName, true,
+            CloudProvider.valueOf(cloudProvider), cloudID, cloudInfoSources, cloudProcessorName);
+
+        CloudConfig validatedCloudConfig = builder.build();
+        addCluster(clusterName, overwritePrevious);
```
jiajunwang: So as I mentioned above, if we have the validation method, we don't need to have 2 lines of addCluster, right?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1025,6 +1026,25 @@ public void dropResource(String clusterName, String resourceName) {
...
+  @Override
+  public void addCloudConfig(String clusterName, CloudConfig cloudConfig) {
+    logger.info("Add CloudConfig to cluster {}, CloudConfig is {}.", clusterName,
+        cloudConfig.toString());
+
+    if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
+      throw new HelixException("cluster " + clusterName + " is not setup yet");
+    }
+
+    CloudConfig.Builder builder = new CloudConfig.Builder(cloudConfig);
```
jiajunwang: Why do we need to copy build a new object for this logic? If it is for validation, please make that method callable here. A builder should not be used for validation.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -176,6 +179,31 @@ public void close() {
...
+      } else {
+        addCluster(clusterName, overwritePrevious);
+      }
+    } else {
+      throw new HelixException("Cannot add the cluster, cloudConfig is null!");
+    }
+  }
+
   public void addCluster(String clusterName, boolean overwritePrevious) {
     _admin.addCluster(clusterName, overwritePrevious);
```
jiajunwang: In general, usually, we do this in a reverse way. The method with less parameters should call the method with more parameters. I think that is why you have the awkward condition check in the method above.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
alirezazamani: I think we need it. For example, when we keep the cluster and remove all the instances and we would like to have new instances to join the cluster with based on new configuration.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
alirezazamani: Ok removeCloudConfig is added.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
alirezazamani: Fixed.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
alirezazamani: I just added this constructor for simplicity. It is not needed but if we can avoid calling several set function and use this function instead, the code will be more readable. If you have strong preference about it, I can remove it though.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
alirezazamani: Addressed.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
alirezazamani: Good suggestion. Fixed.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
alirezazamani: Addressed.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
jiajunwang: This is not we discussed, right? We just need to call validate. And shall we make validate a static method for the CloudConfig?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1025,6 +1026,34 @@ public void dropResource(String clusterName, String resourceName) {
     accessor.removeProperty(keyBuilder.resourceConfig(resourceName));
   }
 
+  @Override
+  public void addCloudConfig(String clusterName, CloudConfig cloudConfig) {
```
jiajunwang: This interface design is a little bit awkward. If we already have a cluster name in the CloudConfig, why cannot we just rely on that one? Now we might have 2 different clusterNames, and the caller is required to read the document for understanding where to put the name...

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -207,6 +228,16 @@ public Builder(CloudConfig cloudConfig) {
       _cloudInfoProcessorName = cloudConfig.getCloudInfoProcessorName();
     }
 
+    public Builder(String clusterName, boolean enabled, CloudProvider cloudProvider, String cloudID,
+        List<String> cloudInfoSource, String cloudInfoProcessorName) {
+      _clusterName = clusterName;
```
jiajunwang: As we discussed offline, the clustername here is just used for the ZNRecord Id. This brings some concerns.
1. The admin API will have the cluster  name as an input, then another cluster name passed through the cloud config. This is duplicate and potentially conflict information.
2. The ZNRecord id itself will be the same as the cluster config node id. Which is not good.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -673,6 +673,41 @@ private void updateClusterConfig(String clusterName, ClusterConfig clusterConfig
...
+
+
+  /**
+   * Update CloudConfig of the given cluster.
+   * The value of field in current config will be replaced with the value of the same field in given config if it
+   * presents. If there is new field in given config but not in current config, the field will be added into
+   * the current cloud config.
+   * The list fields and map fields will be replaced as a single entry.
+   *
+   * The current Cloud config will be replaced with the given cloudConfig.
```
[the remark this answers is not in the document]
alirezazamani: Removed. Thanks.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -380,6 +380,19 @@ void addStateModelDef(String clusterName, String stateModelDef, StateModelDefini
    */
   void dropResource(String clusterName, String resourceName);
 
+  /**
+   *
```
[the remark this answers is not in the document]
alirezazamani: Added.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -176,13 +179,25 @@ public void close() {
     }
   }
 
-  public void addCluster(String clusterName, boolean overwritePrevious) {
-    _admin.addCluster(clusterName, overwritePrevious);
+  public void addCluster(String clusterName, boolean overwritePrevious, CloudConfig cloudConfig)
+      throws HelixException {
+    if (cloudConfig != null) {
+      cloudConfig.validate();
```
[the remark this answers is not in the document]
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
alirezazamani: Fixed. Cluster name has been removed.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
alirezazamani: Fixed. Cluster name has been removed.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -167,6 +168,26 @@ public String getCloudProvider() {
     return _record.getSimpleField(CloudConfigProperty.CLOUD_PROVIDER.name());
   }
 
+
+  public void validate() {
```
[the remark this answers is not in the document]
alirezazamani: As we discussed offline, we will have one validation method in builder and constructors for cloud config have been changed to private method. Thanks for the suggestion. I think the code makes more sense now.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -52,30 +55,23 @@
...
    * Users should use CloudConfig.Builder to create CloudConfig.
-   * @param cluster
    * @param enabled
+   * @param cloudProvider
    * @param cloudID
+   * @param cloudInfoSource
+   * @param cloudProcessorName
    */
-  public CloudConfig(String cluster, boolean enabled, CloudProvider cloudProvider, String cloudID,
+  private CloudConfig(boolean enabled, CloudProvider cloudProvider, String cloudID,
```
dasahcc: Let's make it more generic. In the future, if we have more entries, we may need change the constructor for adding new arguments.

My suggestion will be taking the ZnRecord as the input instead of having all the entries as arguments. Also that helps to reduce the place containing specific entries copying logic.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -90,7 +86,7 @@ public CloudConfig(String cluster, boolean enabled, CloudProvider cloudProvider,
    * Enable/Disable the CLOUD_ENABLED field.
    * @param enabled
    */
-  public void setCloudEnabled(boolean enabled) {
+  private void setCloudEnabled(boolean enabled) {
```
dasahcc: These methods are/can not used by any place. Instead of removing all, we can keep some of the them as public ones. We need to leave the room to let user change some of the fields, such as cloudenable, source fields.

For validation purpose, we can make all constructors private and creating through Builder.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -188,11 +184,17 @@ public Builder() {
...
-    public Builder(String clusterName) {
-      _clusterName = clusterName;
+    public Builder(ZNRecord record) {
+      _cloudEnabled = record.getBooleanField(CloudConfigProperty.CLOUD_ENABLED.name(), false);
+      _cloudProvider =
+          CloudProvider.valueOf(record.getSimpleField(CloudConfigProperty.CLOUD_PROVIDER.name()));
+      _cloudID = record.getSimpleField(CloudConfigProperty.CLOUD_ID.name());
+      _cloudInfoSources = record.getListField(CloudConfigProperty.CLOUD_INFO_SOURCE.name());
+      _cloudInfoProcessorName =
+          record.getSimpleField(CloudConfigProperty.CLOUD_INFO_PROCESSOR_NAME.name());
```
dasahcc: We can keep a record as internal member.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -207,9 +209,13 @@ public Builder(CloudConfig cloudConfig) {
       _cloudInfoProcessorName = cloudConfig.getCloudInfoProcessorName();
     }
 
-    public Builder setClusterName(String v) {
-      _clusterName = v;
-      return this;
+    public Builder(boolean enabled, CloudProvider cloudProvider, String cloudID,
```
dasahcc: This is not necessary. The functionality is duplicated with set methods.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
alirezazamani: Fixed. Thanks for the suggestion.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
alirezazamani: Fixed. I made 3 of them that users might use as public. I added a test for those methods as well.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
alirezazamani: Fixed. Thanks.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -167,122 +136,108 @@ public String getCloudProvider() {
...
     }
 
     private void validate() {
-      if (_cloudEnabled) {
-        if (_cloudID == null) {
-          throw new HelixException(
-              "This Cloud Configuration is Invalid. The CloudID is missing from the config.");
-        }
-        if (_cloudProvider == null) {
+      if (this.getCloudID() == null) {
```
dasahcc: Is "this" necessary?

alirezazamani: Fixed.

helix-core/src/main/java/org/apache/helix/SystemPropertyKeys.java
```diff
@@ -38,4 +38,5 @@
 
   // MBean monitor for helix.
   public static final String HELIX_MONITOR_TIME_WINDOW_LENGTH_MS = "helix.monitor.slidingTimeWindow.ms";
+
```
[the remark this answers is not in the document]
alirezazamani: Fixed. Thanks.

alirezazamani:
This PR is ready to be merged, approved by @zhangmeng916 and @dasahcc.

Final Commit message:
Title: 
Add java API to create cluster with CloudConfig

Body:
In this commit the below APIs have been added.
1- AddCluster with CloudCOnfig API.
2- addCloudConfig to existing cluster.
3- Update CloudConfig to update existing Cloud Config.
Several tests have been added to test these APIs.

Issue #650, opened 2019-12-11
WAGED rebalancer moves extra partitions when new node joins the cluster


jiajunwang:
Due to the greedy algorithm that is used in the WAGED rebalancer, it might assign some partitions quite randomly when the cluster is empty. This usually leads to extra (unnecessary) movements in the final result. As we tested, the most impacted scenario is new node joining.

PR #651, opened 2019-12-11
Improve the algorithm so it prioritizes the assignment to the idle nodes when the constraint evaluation results are the same

jiajunwang:
### Issues

#650 

### Description

This is to get rid of the randomness when the constraint-based evaluation result is a tie. Usually, when the algorithm picks up the nodes with the same score randomly, more partition movements will be triggered. A better strategy is filling the idle nodes (AKA the newly added nodes first). So those assignments won't cause a chain reaction that the other partition placements have to be altered.

### Tests

We are using the same correctness based tests to cover this logic change.
For the performance-related tests, we will do that in the long run test later. The report will be provided by then.

[ERROR] Failures: 
[INFO]   Run 4: PASS
[INFO]   Run 5: PASS
[INFO] 
[ERROR]   TestZKUtil.testChildrenOperations:78 expected:<2> but was:<3>
[ERROR]   TestGetLastScheduledTaskExecInfo.testGetLastScheduledTaskExecInfo:63->setupTasks:115 NullPointer
[INFO] 
[ERROR] Tests run: 1048, Failures: 6, Errors: 0, Skipped: 3

Re-run:
[INFO] Tests run: 40, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 37.635 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 40, Failures: 0, Errors: 0, Skipped: 0

### Commits

### Documentation

TBD, we will conclude the detail design soon after the algorithm has been tested and finalized.

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
@@ -122,13 +122,9 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
           if (scoreCompareResult == 0) {
             // If the evaluation scores of 2 nodes are the same, the algorithm assigns the replica
             // to the idle node first.
-            boolean isNodeOneIdle = !busyInstances.contains(nodeEntry1.getKey().getInstanceName());
-            boolean isNodeTwoIdle = !busyInstances.contains(nodeEntry2.getKey().getInstanceName());
-            if (isNodeOneIdle != isNodeTwoIdle) {
-              return isNodeOneIdle ? 1 : -1;
-            } else {
-              return 0;
-            }
+            int idleScore1 = busyInstances.contains(nodeEntry1.getKey().getInstanceName()) ? 0 : 1;
+            int idleScore2 = busyInstances.contains(nodeEntry2.getKey().getInstanceName()) ? 0 : 1;
+            return idleScore1 - idleScore2;
           } else {
             return scoreCompareResult;
           }
@@ -219,6 +215,10 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
     return orderedAssignableReplicas;
   }
 
+  /**
+   * @param assignments A collection of resource replicas assignment.
+   * @return A set of instance names that have at least one replica assigned in the input assignments.
+   */
   private Set<String> getBusyInstances(Collection<ResourceAssignment> assignments) {
     return assignments.stream().flatMap(
         resourceAssignment -> resourceAssignment.getRecord().getMapFields().values().stream()
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
@@ -20,6 +20,7 @@ package org.apache.helix.controller.rebalancer.waged.constraints;
  */
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterContext.java
```diff
@@ -104,6 +106,15 @@
     _estimatedMaxTopStateCount = estimateAvgReplicaCount(totalTopStateReplicas, instanceCount);
     _baselineAssignment = baselineAssignment;
     _bestPossibleAssignment = bestPossibleAssignment;
+
+    _busyInstances = _bestPossibleAssignment.values().stream().flatMap(
```
i3wangyi: Why use bestPossibleAssignment?  BestPossibleAssignment may not be the current cluster mapping. Shouldn't the `isBusy` boolean information determined during the initNewBatch in ClusterModelProvider?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -113,8 +112,24 @@ public OptimalAssignment calculate(ClusterModel clusterModel) throws HelixRebala
...
     return candidateNodes.parallelStream().map(node -> new HashMap.SimpleEntry<>(node,
         getAssignmentNormalizedScore(node, replica, clusterContext)))
-        .max(Comparator.comparingDouble((scoreEntry) -> scoreEntry.getValue()))
-        .map(Map.Entry::getKey);
+        .max((nodeEntry1, nodeEntry2) -> {
+          int scoreCompareResult =  nodeEntry1.getValue().compareTo(nodeEntry2.getValue());
+          if (scoreCompareResult == 0) {
+            // If the evaluation scores of 2 nodes are the same, the algorithm assigns the replica
+            // to the idle node first.
+            boolean isNodeOneIdle =
```
i3wangyi: It can be simplified. 
```
int score1 = node1.isBusy() ? 1 : 0 * score1;
int score2 = node2.isBusy() ?  1: 0 * score2;
return score1 - score2;
```

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterContext.java
jiajunwang: Current mapping is never directly used in the algorithm. To WAGED, the best possible assignment is treated as the current cluster mapping.

Given that saying, it cannot be determined by iniNewBatch. Since that input would already be processed by the rebalance scope calculating method. So it is not the one that we want.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
jiajunwang: That logic will be different, note that the score is still the most important thing. If I change it as you suggested, the busy node will always win regardless of the score. As I explained to you earlier, this should only change the behavior when the evaluation cannot tell the difference between 2 nodes.

i3wangyi: My suggestion of the multiply is after you check if the score equals.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterContext.java
i3wangyi: Is it because of yesterday's my question: if the cluster's present (current mapping) is a determined intermediate result from the past's (init mapping) calculation, the future of the cluster (final mapping) needs to be determined by the "past"'s calculation **best possible mapping** rather than current mapping? It sounds complex but I feel it's one of the trickiest things to understand and relate in code in the event-driven stateful system.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterContext.java
```diff
@@ -104,6 +106,15 @@
...
+        resourceAssignment -> resourceAssignment.getRecord().getMapFields().values().stream()
```
i3wangyi: This is a bit of general advice FYI. 
I feel concerned about manipulating a map of nested map data structure in multiple places, it's in general hard to read & follow. A centralized data model of bestPossibleAssignement, baselineAssignment, even current assignment would be ideal. Something like
```
public class ClusterAssignment {
   
   public xxx findDifference(ClusterAssignment other) {
   }
   public Set<String> getBusyInstances() {
    .... 
   }
}
ClusterAssignment bestPossible;
ClusterAssignment baseLine;
```

It would be easier to unit test and have as many as things centralized in one place

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterContext.java
jiajunwang: If I get your question correctly, you are right. To be simple, the current state (which is intermediate) is never read in the whole rebalance calculation. It is mainly compared and processed by the other rebalance pipeline stages.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
jiajunwang: I see. let me have a try.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterContext.java
jiajunwang: I agree. Just this seems to be a too large change for this PR. Let me try the best to make it clearer with less code change. I have moved the calculation to the algorithm implementation for now. Later, as you mentioned if we have more requirements, we can split it out to a class.

jiajunwang:

Issue #652, opened 2019-12-12
Add Helix Cloud Properties


zhangmeng916:
When the participant of Helix is running inside a cloud environment and would like to auto register to its cluster, there are a couple of properties needed, e.g. which function to use to fetch/parse the cloud instance information, what is the source for retrieving cloud instance information, as well as some user defined properties, e.g. http timeout, max retries, etc. We should have a globally available Helix Cloud properties defined so that different classes may use it.

PR #653, opened 2019-12-12
add Helix properties factory and class

zhangmeng916:
### Issues

(#[652](https://github.com/apache/helix/issues/652))
### Description

When the participant of Helix is running inside a cloud environment and would like to auto register to its cluster, there are a couple of properties needed, e.g. which function to use to fetch/parse the cloud instance information, what is the source for retrieving cloud instance information, as well as some user defined properties, e.g. http timeout, max retries, etc. We nee to have a globally available Helix Cloud properties defined so that different classes may use it. 
This PR creates a singleton Helix cloud properties factory, and a Helix cloud properties class. In Helix cloud properties, properties are read from different sources and combined to store in the memory. The local properties will be used later by participants and cloud instance information processor.

### Tests
[ERROR] Tests run: 895, Failures: 3, Errors: 0, Skipped: 0, Time elapsed: 3,773.486 s <<< FAILURE! - in TestSuite
[ERROR] testWorkflowFailureJobThreshold(org.apache.helix.integration.task.TestJobFailureDependence)  Time elapsed: 300.015 s  <<< FAILURE!
org.testng.internal.thread.ThreadTimeoutException: Method org.testng.internal.TestNGMethod.testWorkflowFailureJobThreshold() didn't finish within the time-out 300000

[ERROR] testWorkflowJobFail(org.apache.helix.integration.task.TestWorkflowTermination)  Time elapsed: 10.043 s  <<< FAILURE!
org.apache.helix.HelixException: Workflow "testWorkflowJobFail" context is empty or not in states: "[FAILED]", current state: "IN_PROGRESS"
	at org.apache.helix.integration.task.TestWorkflowTermination.testWorkflowJobFail(TestWorkflowTermination.java:223)

[ERROR] testClusterStatusMonitorLifecycle(org.apache.helix.monitoring.TestClusterStatusMonitorLifecycle)  Time elapsed: 6.413 s  <<< FAILURE!
	at org.apache.helix.monitoring.TestClusterStatusMonitorLifecycle.testClusterStatusMonitorLifecycle(TestClusterStatusMonitorLifecycle.java:290)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestJobFailureDependence.testWorkflowFailureJobThreshold » ThreadTimeout Metho...
[ERROR]   TestWorkflowTermination.testWorkflowJobFail:223 » Helix Workflow "testWorkflow...
[ERROR]   TestClusterStatusMonitorLifecycle.testClusterStatusMonitorLifecycle:290 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 895, Failures: 3, Errors: 0, Skipped: 0
[INFO] 
[INFO] Reactor Summary for Apache Helix 0.9.2-SNAPSHOT:
[INFO] 
[INFO] Apache Helix ....................................... SUCCESS [  0.295 s]
[INFO] Apache Helix :: Core ............................... FAILURE [  01:02 h]
[INFO] Apache Helix :: Admin Webapp ....................... SKIPPED
[INFO] Apache Helix :: Restful Interface .................. SKIPPED
[INFO] Apache Helix :: HelixAgent ......................... SKIPPED
[INFO] Apache Helix :: Recipes ............................ SKIPPED
[INFO] Apache Helix :: Recipes :: Rabbitmq Consumer Group . SKIPPED
[INFO] Apache Helix :: Recipes :: Rsync Replicated File Store SKIPPED
[INFO] Apache Helix :: Recipes :: distributed lock manager  SKIPPED
[INFO] Apache Helix :: Recipes :: distributed task execution SKIPPED
[INFO] Apache Helix :: Recipes :: service discovery ....... SKIPPED
[INFO] BUILD FAILURE
[INFO] Total time:  01:02 h
[INFO] Finished at: 2019-12-11T20:27:48-08:00
[ERROR] 
[ERROR] Please refer to /Users/mnzhang/helix/helix-core/target/surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] 
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :helix-core

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -581,13 +581,14 @@ public ClusterConfig getClusterConfig(String clusterName) {
     return new ClusterConfig(record);
   }
 
+
```
narendly: Nit: redundant line?

helix-core/src/main/java/org/apache/helix/HelixCloudProperties.java
```diff
@@ -0,0 +1,108 @@
...
+   * To combine cloud properties from different sources, firstly, the cloud config will be examined, if the provider has its
+   * properties defined in Helix, then the corresponding file under helix-core/src/main/resources/ will be read. Finally,
+   * user input properties will be merged. If there is any duplicate key existing, the user defined property will have
+   * higher priority and override other properties
+   * @param userProperties
+   * @param cloudConfig
+   */
+  public HelixCloudProperties(Properties userProperties, CloudConfig cloudConfig) {
+    switch (cloudConfig.getCloudProvider()) {
+      case "AZURE":
```
narendly: Question: Has this been formatted properly? I believe there shouldn't be an indentation between switch and case?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * hold helix cloud properties
+ */
+public class HelixCloudProperty {
+  private static final Logger logger = LoggerFactory.getLogger(HelixCloudProperty.class.getName());
+  private final String AZURE_CLOUD_PROPERTY_FILE = SystemPropertyKeys.AZURE_CLOUD_PROPERTIES;
```
narendly: private static final for constant declaration?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+
+  private Boolean _isCloudEnabled;
```
narendly: Nit: does this need to be a big Boolean? Can we use a primitive type?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Initialize Helix Cloud Property based on the provider
+   * @param
+   */
+  public HelixCloudProperty(CloudConfig cloudConfig) {
+    this.setCloudEndabled(cloudConfig.isCloudEnabled());
+    this.setCloudId(cloudConfig.getCloudID());
+    this.setCloudProvider(cloudConfig.getCloudProvider());
+    switch (cloudConfig.getCloudProvider()) {
```
pkuwm: Need a default case?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+
+  private Boolean _isCloudEnabled;
+  private String _cloudId;
+  private String _cloudProvider;
+  private List<String> _cloudInfoSources;
+  private String _cloudInfoProcessorName;
+  private String _maxRetry;
+  private String _cloudConnectionTimeout;
+  private String _cloudRequestTimeout;
+  private Properties _customizedProperties = new Properties();
```
narendly: Nit: Like we discussed previously, I think it would be great if you could have a block comment describing what these are and with an example.

For example,

/**
* _cloudProvider: name of the cloud provider. E.g.) "Azure", "GCP", etc.
*
*
...

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Initialize Helix Cloud Property based on the provider
+   * @param
+   */
+  public HelixCloudProperty(CloudConfig cloudConfig) {
+    this.setCloudEndabled(cloudConfig.isCloudEnabled());
+    this.setCloudId(cloudConfig.getCloudID());
+    this.setCloudProvider(cloudConfig.getCloudProvider());
```
narendly: Nit: are these `this`s necessary?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
pkuwm: Might be more clear to use a primitive boolean?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+    switch (cloudConfig.getCloudProvider()) {
+      case "AZURE":
+        Properties azureProperties = new Properties();
+        try {
+          InputStream stream =
+              Thread.currentThread().getContextClassLoader().getResourceAsStream(AZURE_CLOUD_PROPERTY_FILE);
+          azureProperties.load(stream);
+        } catch (Exception e) {
+          String errMsg = "fail to open properties file: " + AZURE_CLOUD_PROPERTY_FILE;
```
narendly: "failed"

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+import org.apache.helix.model.CloudConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * hold helix cloud properties
+ */
+public class HelixCloudProperty {
+  private static final Logger logger = LoggerFactory.getLogger(HelixCloudProperty.class.getName());
```
narendly: Nit: I think the convention we use is LOG for logger.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -245,6 +253,11 @@ public ZKHelixManager(String clusterName, String instanceName, InstanceType inst
 
     _stateListener = stateListener;
 
+    if (helixParticipantProperty != null) {
+      _helixParticipantProperty = helixParticipantProperty;
```
pkuwm: No need?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+        Properties azureProperties = new Properties();
+        try {
+          InputStream stream =
+              Thread.currentThread().getContextClassLoader().getResourceAsStream(AZURE_CLOUD_PROPERTY_FILE);
+          azureProperties.load(stream);
+        } catch (Exception e) {
+          String errMsg = "fail to open properties file: " + AZURE_CLOUD_PROPERTY_FILE;
+          throw new IllegalArgumentException(errMsg, e);
+        }
+        logger.info("load helix Azure cloud properties: " + azureProperties);
```
narendly: Could you consider revising this to "Successfully loaded Helix Azure cloud properties"?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
pkuwm: Static final for this constant?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+        this.setCloudInfoSources(Collections.singletonList(azureProperties.getProperty("CLOUD_INFO_SOURCE")));
+        this.setCloudInfoProcessorName(azureProperties.getProperty("CLOUD_INFO_PROCESSFOR_NAME"));
+        this.setCloudMaxRetry(azureProperties.getProperty("RETRY_MAX"));
+        this.setCloudConnectionTimeout(azureProperties.getProperty("CONNECTION_TIMEOUT_MS"));
+        this.setCloudRequestTimeout(azureProperties.getProperty("REQUEST_TIMEOUT_MS"));
```
narendly: Instead of hard-coding, could we either have a separate constant class or define them as constants in this class?

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -0,0 +1,59 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Singleton factory that builds different types of Helix property, e.g. participant property, controller property, etc.
+ */
+public final class HelixPropertyFactory {
+  private static final Logger logger = LoggerFactory.getLogger(HelixPropertyFactory.class);
+  private final String HELIX_PARTICIPANT_PROPERTY_FILE = SystemPropertyKeys.HELIX_PARTICIPANT_PROPERTIES;
```
pkuwm: If constant is expected, declare static final ?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Initialize Helix Cloud Property based on the provider
+   * @param
+   */
+  public HelixCloudProperty(CloudConfig cloudConfig) {
+    this.setCloudEndabled(cloudConfig.isCloudEnabled());
+    this.setCloudId(cloudConfig.getCloudID());
+    this.setCloudProvider(cloudConfig.getCloudProvider());
+    switch (cloudConfig.getCloudProvider()) {
+      case "AZURE":
```
narendly: Instead of a hard-coded string, could we have an enum or a constant for this?

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public HelixCloudProperty getHelixCloudProperty() {
+    return _helixCloudProperty;
+  }
+
+  public String getVersion() {
+    return _version;
+  }
+
+  public Integer getHealthReportLatency() {
```
pkuwm: Primitive value may be enough?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+    _cloudConnectionTimeout = cloudConnectionTimeout;
+  }
+
+  public void setCloudRequestTimeout(String cloudRequestTimeout) {
+    _cloudRequestTimeout = cloudRequestTimeout;
+  }
+
+  public void setCustomizedCloudProperties(Properties customizedCloudProperties) {
+    _customizedProperties.putAll(customizedCloudProperties);
+  }
```
narendly: Ideally every public method/API should have a pertinent JavaDoc to go with explaining what they do and what these strings look like (examples)

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Properties;
+import org.apache.helix.model.CloudConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * hold helix cloud properties
```
narendly: This JavaDoc doesn't tell me any more than the class name. Could we try to make this more complete?

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+import java.util.Properties;
+import org.apache.helix.model.CloudConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * hold participant properties
```
narendly: This JavaDoc doesn't tell me any more than the class name. Could we try to make this more complete?

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+public class HelixParticipantProperty {
+  private static final Logger logger = LoggerFactory.getLogger(HelixParticipantProperty.class.getName());
```
narendly: Consider using LOG?

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
```diff
@@ -0,0 +1,70 @@
...
+  private String _version;
+  private Integer _healthReportLatency;
+  private HelixCloudProperty _helixCloudProperty;
```
narendly: Should this be final?

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
```diff
@@ -0,0 +1,70 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * hold participant properties
+ */
+public class HelixParticipantProperty {
+  private static final Logger logger = LoggerFactory.getLogger(HelixParticipantProperty.class.getName());
+  private String _version;
+  private Integer _healthReportLatency;
```
narendly: I have no idea what this is for since there are no descriptions, but we usually use `long` for latency values?

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
```diff
@@ -0,0 +1,70 @@
...
+  private HelixCloudProperty _helixCloudProperty;
+
+  /**
+   * Initialize Helix Participant Property with default value
+   * @param
```
narendly: This JavaDoc is missing the param field.

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
```diff
@@ -0,0 +1,70 @@
...
+ * hold participant properties
+ */
+public class HelixParticipantProperty {
+  private static final Logger logger = LoggerFactory.getLogger(HelixParticipantProperty.class.getName());
+  private String _version;
+  private Integer _healthReportLatency;
+  private HelixCloudProperty _helixCloudProperty;
+
+  /**
+   * Initialize Helix Participant Property with default value
```
narendly: Where are we getting the default values from?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,150 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(HelixCloudProperty.class.getName());
+  private final String AZURE_CLOUD_PROPERTY_FILE = SystemPropertyKeys.AZURE_CLOUD_PROPERTIES;
+
+  private Boolean _isCloudEnabled;
+  private String _cloudId;
+  private String _cloudProvider;
+  private List<String> _cloudInfoSources;
+  private String _cloudInfoProcessorName;
+  private String _maxRetry;
+  private String _cloudConnectionTimeout;
```
pkuwm: Did you consider these timeout as long type?

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
```diff
@@ -0,0 +1,70 @@
...
+    _helixCloudProperty = helixCloudProperty;
+  }
+
+  public void setVersion(String version) {
+    _version = version;
+  }
+
+  public void setHealthReportLatency(String latency) {
+    _healthReportLatency = Integer.valueOf(latency);
+  }
```
narendly: Could we add a proper JavaDoc for public API methods?

```
/** 
 * Get the current value of the foo property.
 * The foo property controls the initial guess used by the bla algorithm in
 * {@link #bla}
 * @return The initial guess used by {@link #bla}
 */
int getFoo() {
  return foo;
}
```
From:
https://stackoverflow.com/questions/211041/do-you-use-javadoc-for-every-method-you-write

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.helix.model.CloudConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Singleton factory that builds different types of Helix property, e.g. participant property, controller property, etc.
+ */
+public final class HelixPropertyFactory {
+  private static final Logger logger = LoggerFactory.getLogger(HelixPropertyFactory.class);
```
narendly: LOG?

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
narendly: private static final?

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private static class SingletonHelper {
+    private static final HelixPropertyFactory INSTANCE = new HelixPropertyFactory();
+  }
+
+  public static HelixPropertyFactory getInstance() {
+    return SingletonHelper.INSTANCE;
+  }
+
+  public HelixParticipantProperty getHelixParticipantProperty(String zkAddress, String clusterName) {
```
narendly: JavaDoc?

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -0,0 +1,59 @@
...
+    ConfigAccessor configAccessor = new ConfigAccessor(zkAddress);
+    CloudConfig cloudConfig = configAccessor.getCloudConfig(clusterName);
+    Properties properties = new Properties();
+    try {
+      InputStream stream =
+          Thread.currentThread().getContextClassLoader().getResourceAsStream(HELIX_PARTICIPANT_PROPERTY_FILE);
+        properties.load(stream);
+    } catch (Exception e) {
+      String errMsg = "fail to open properties file: " + HELIX_PARTICIPANT_PROPERTY_FILE;
```
narendly: "HelixPropertyFactory failed to ..." to distinguish it from the other log.

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -0,0 +1,59 @@
...
+      throw new IllegalArgumentException(errMsg, e);
+    }
+    logger.info("load helix participant properties: " + properties);
```
narendly: HelixPropertyFactory successfully loaded Helix Participant properties...?

helix-core/src/main/java/org/apache/helix/SystemPropertyKeys.java
```diff
@@ -1,6 +1,10 @@
 package org.apache.helix;
 
 public class SystemPropertyKeys {
+  public static final String HELIX_PARTICIPANT_PROPERTIES = "helix-participant.properties";
+
+  public static final String AZURE_CLOUD_PROPERTIES = "azure-cloud.properties";
+
```
narendly: Let's add some documentation for what, how, when these are going to be used?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -76,9 +77,16 @@
   final StateMachineEngine _stateMachineEngine;
   final LiveInstanceInfoProvider _liveInstanceInfoProvider;
   final List<PreConnectCallback> _preConnectCallbacks;
+  final HelixParticipantProperty _helixParticipantProperty;
 
   public ParticipantManager(HelixManager manager, HelixZkClient zkclient, int sessionTimeout,
       LiveInstanceInfoProvider liveInstanceInfoProvider, List<PreConnectCallback> preConnectCallbacks) {
+    this(manager, zkclient, sessionTimeout, liveInstanceInfoProvider, preConnectCallbacks, null);
```
narendly: Is passing `null` a safe thing to do? Wouldn't it be better to pass in an empty instance of `HelixParticipantProperty` to avoid potential NPEs?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -202,6 +205,11 @@ public ZKHelixManager(String clusterName, String instanceName, InstanceType inst
 
   public ZKHelixManager(String clusterName, String instanceName, InstanceType instanceType,
       String zkAddress, HelixManagerStateListener stateListener) {
+    this(clusterName, instanceName, instanceType, zkAddress,stateListener, null);
```
narendly: Is passing null a safe thing to do?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -202,6 +205,11 @@ public ZKHelixManager(String clusterName, String instanceName, InstanceType inst
...
+  }
+
+  public ZKHelixManager(String clusterName, String instanceName, InstanceType instanceType,
+      String zkAddress, HelixManagerStateListener stateListener, HelixParticipantProperty helixParticipantProperty) {
```
narendly: ZKHelixManager is not only used by Participants. It could be created in many different modes such as Controller, Participant, Controller Participant, and Spectator. So I do not think adding HelixParticipantProperty in the constructor is a good idea because you're adding specificity to a general class. Using a more generic type would be better.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -245,6 +253,11 @@ public ZKHelixManager(String clusterName, String instanceName, InstanceType inst
 
     _stateListener = stateListener;
 
+    if (helixParticipantProperty != null) {
+      _helixParticipantProperty = helixParticipantProperty;
+    } else {
+      _helixParticipantProperty = HelixPropertyFactory.getInstance().getHelixParticipantProperty(zkAddress, clusterName);
+    }
```
narendly: This could be refactored once we get rid of null's.

helix-core/src/main/resources/helix-participant.properties
```diff
@@ -0,0 +1,25 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+clustermanager.version=${project.version}
```
narendly: Question: why do we need to have the version in the property? How will it be used?

helix-core/src/main/resources/helix-participant.properties
```diff
@@ -0,0 +1,25 @@
...
+
+minimum_supported_version.batch_message=0.6.1
+minimum_supported_version.participant=0.4
+
+helixmanager.participantHealthReport.reportLatency=60*1000
```
narendly: Is this going to be a string or a long?

helix-core/src/main/java/org/apache/helix/HelixCloudProperties.java
zhangmeng916: This file was removed.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
zhangmeng916: The logic here is that if user provides helixParticipantProperty, we will use the provided one. Otherwise, we create a HelixParticipantProperty with default value.

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
zhangmeng916: The default value means the value set for the fields if user does not override them. They come from cloud config and property file.

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,158 @@
...
+
+
+/**
+ * Hold helix cloud properties read from CloudConfig and user defined files. Clients may override the fields from their application.
+ */
+public class HelixCloudProperty {
+  private static final Logger LOG = LoggerFactory.getLogger(HelixCloudProperty.class.getName());
+  private static final String AZURE_CLOUD_PROPERTY_FILE = SystemPropertyKeys.AZURE_CLOUD_PROPERTIES;
+  private static final String CLOUD_INFO_SOURCE = "CLOUD_INFO_SOURCE";
+  private static final String CLOUD_INFO_PROCESSFOR_NAME = "cloud_info_processor_name";
```
jiajunwang: Should we use capitalized letters for consistency?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,158 @@
...
+        setCloudInfoSources(Collections.singletonList(azureProperties.getProperty(CLOUD_INFO_SOURCE)));
+        setCloudInfoProcessorName(azureProperties.getProperty(CLOUD_INFO_PROCESSFOR_NAME));
+        setCloudMaxRetry(azureProperties.getProperty(RETRY_MAX));
+        setCloudConnectionTimeout(azureProperties.getProperty(CONNECTION_TIMEOUT_MS));
+        setCloudRequestTimeout(azureProperties.getProperty(REQUEST_TIMEOUT_MS));
+      case CUSTOMIZED:
+        setCloudInfoSources(cloudConfig.getCloudInfoSources());
+        setCloudInfoProcessorName(cloudConfig.getCloudInfoProcessorName());
+      default:
+        LOG.info("unrecognized cloud provider: " + cloudConfig.getCloudProvider());
```
jiajunwang: Unsupported cloud provider ?

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,158 @@
...
+import org.apache.helix.cloud.constants.CloudProvider;
+import org.apache.helix.model.CloudConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Hold helix cloud properties read from CloudConfig and user defined files. Clients may override the fields from their application.
+ */
+public class HelixCloudProperty {
```
jiajunwang: Shall we make all the set methods to be private if they are not used anywhere else?

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.helix.model.CloudConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Hold Helix participant properties. The participant properties further hold Helix cloud properties and some other properties
+ * specific for the participant.
+ */
+public class HelixParticipantProperty {
```
jiajunwang: Same for the set methods, is private good enough?

helix-core/src/main/java/org/apache/helix/api/cloud/CloudInstanceInformation.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Generic interface for cloud instance information
+ */
+public interface CloudInstanceInformation {
+  /**
+   * Get the the value of a specific cloud instance field by name
+   * @return the value of the field
+   */
+  String get(String key);
```
jiajunwang: Shouldn't it be String get(CloudInstanceField key); ? Otherwise, what's the point of having the enum here?

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformation.java
```diff
@@ -0,0 +1,77 @@
...
+      return new AzureCloudInstanceInformation(_cloudInstanceInfoMap);
+    }
+
+    /**
+     * Default constructor
+     */
+    public Builder() {
+    }
+
+    public Builder setInstanceName(String v) {
```
jiajunwang: Please use a meaningful parameter name for the method.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformation.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  public static class Builder {
+    private Map<String, String> _cloudInstanceInfoMap = null;
+
+    public AzureCloudInstanceInformation build() {
+      return new AzureCloudInstanceInformation(_cloudInstanceInfoMap);
+    }
+
+    /**
```
jiajunwang: nit, Default constructor is optional, we can remove this.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
jiajunwang: HelixManagerProperty would be better.

jiajunwang: And we might want to deprecate the older constructor.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -0,0 +1,290 @@
...
+     */
+    public Builder(CloudConfig cloudConfig) {
+      _cloudEnabled = cloudConfig.isCloudEnabled();
+      _cloudProvider = CloudProvider.valueOf(cloudConfig.getCloudProvider());
+      _cloudID = cloudConfig.getCloudID();
+      _cloudInfoSources = cloudConfig.getCloudInfoSources();
+      _cloudInfoProcessorName = cloudConfig.getCloudInfoProcessorName();
+    }
+
+    public Builder setClusterName(String v) {
```
jiajunwang: meaningful param name, please

helix-core/src/main/resources/azure-cloud.properties
```diff
@@ -0,0 +1,25 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+cloud_info_resource = "http://169.254.169.254/metadata/instance?api-version=2019-06-04"
```
jiajunwang: Is this for testing? Please comment if this will be the default value for all of our customers.

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
zhangmeng916: I changed them to be all lower case so that the properties files can use lower case.

helix-core/src/main/java/org/apache/helix/api/cloud/CloudInstanceInformation.java
zhangmeng916: Let's discuss more about this part. The original purpose is that we allow customized implementation to have different fields other than the enum.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
zhangmeng916: It's in Ali's code. I've rebased, and it should be gone.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -0,0 +1,75 @@
...
+  private HelixCloudProperty _helixCloudProperty;
+
+  /**
+   * Initialize Helix manager property with default value
+   * @param properties helix manager related properties input as a map
+   * @param cloudConfig cloudConfig read from Zookeeper
+   */
+  public HelixManagerProperty(Properties properties, CloudConfig cloudConfig) {
+    if (cloudConfig != null) {
+      _helixCloudProperty = new HelixCloudProperty(cloudConfig);
```
jiajunwang: I think we discussed this once, if we have a boolean "isInCloud" (I think we had a better name, but I don't remember), then this nullable value can be handled more gracefully.
At least we can check if the "isInCloud" == true, null cloudconfig input should trigger an exception.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkClusterManager.java
```diff
@@ -63,18 +63,11 @@ public void testController() throws Exception {
...
     }
 
+    TestHelper.setupEmptyCluster(_gZkClient, clusterName);
+
     ZKHelixManager controller =
         new ZKHelixManager(clusterName, null, InstanceType.CONTROLLER, ZK_ADDR);
 
-    try {
-      controller.connect();
-      Assert.fail("Should throw HelixException if initial cluster structure is not setup");
```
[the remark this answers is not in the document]
zhangmeng916: The problem with the previous test is that the call to get default Helix manager property will fail because we have not set it up yet. The old ZKHelixManager constructor has no dependency on cluster set up, but the new one needs to get the cloud config from it.

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,175 @@
...
+
+  public long getCloudRequestTimeout() {
+    return _cloudRequestTimeout;
+  }
+
+  public Properties getCustomizedCloudProperties() {
+    return _customizedCloudProperties;
+  }
+
+  private void setCloudEndabled(boolean isCloudEnabled) {
```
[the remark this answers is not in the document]
zhangmeng916: These should be public, as clients may need to set them.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
zhangmeng916: The cloudEnabled flag is actually in cloud config itself. Remember that we talk about we do not want to touch cluster config, and we put all cloud related configs in a separate znode.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
zhangmeng916: I mark it as deprecated.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
jiajunwang: As we discussed, my major concern is getHelixCloudProperty() will return null. This makes the class harder to use.

helix-core/src/main/java/org/apache/helix/HelixParticipantProperty.java
zhangmeng916: The client will need to override them.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkClusterManager.java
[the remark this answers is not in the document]
jiajunwang: I understand, but in this case, please find an alternative way to test instead of removing it.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
zhangmeng916: Changed to have a default cloudconfig value.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
zhangmeng916: already modified.

helix-core/src/main/java/org/apache/helix/api/cloud/CloudInstanceInformation.java
zhangmeng916: Let's keep the flexibility for now. We can make it restricted later.

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,179 @@
...
+
+  public void setCloudMaxRetry(String cloudMaxRetry) {
+    _cloudMaxRetry = Integer.valueOf(cloudMaxRetry);
+  }
+
+  public void setCloudConnectionTimeout(String cloudConnectionTimeout) {
+    _cloudConnectionTimeout = Long.valueOf(cloudConnectionTimeout);
+  }
+
+  public void setCloudRequestTimeout(String cloudRequestTimeout) {
```
pkuwm: From the comment of the class: the fields could be changed by clients. So I think the parameter type could be `long` so it is consistent with the field type. And clients don't have to convert a long type timeout to a string.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Properties;
+import org.apache.helix.model.CloudConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Hold Helix participant properties. The participant properties further hold Helix cloud properties and some other properties
+ * specific for the participant.
+ */
+public class HelixManagerProperty {
```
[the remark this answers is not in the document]
zhangmeng916: Typo, changed to "manager". There is already HelixManangerProperties in the repo, so we keep singular for all property in our new design.

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,179 @@
...
+  private static final String REQUEST_TIMEOUT_MS = "request_timeout_ms";
+
+  // Denote whether the instance is considered as in a cloud environment.
+  private boolean _isCloudEnabled;
+
+  // Unique id of the cloud environment where the instance is in.
+  private String _cloudId;
+
+  // Cloud environment provider, e.g. Azure, AWS, GCP, etc.
+  private String _cloudProvider;
```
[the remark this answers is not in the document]
zhangmeng916: I tried to change it to CloudProvider. I agree that this is more restricted and clear in this file. However, in cloud config, the function getCloudProvider() returns a string. As it is already a string instead of Enum, it makes less value to have a Enum in this file as it needs to call that function anyway. Later maybe we can consider them together to see whether we would like to change.

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,184 @@
...
+  // Other customized properties that may be used.
+  private Properties _customizedCloudProperties = new Properties();
+
+  /**
+   * Initialize Helix Cloud Property based on the provider
+   * @param
+   */
+  public HelixCloudProperty(CloudConfig cloudConfig) {
+    setCloudEndabled(cloudConfig.isCloudEnabled());
+    if (cloudConfig.isCloudEnabled() == true) {
```
jiajunwang: " == true" is redundent code. Please remove.

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -0,0 +1,184 @@
...
+        setCloudMaxRetry(Integer.valueOf(azureProperties.getProperty(CLOUD_MAX_RETRY)));
+        setCloudConnectionTimeout(Long.valueOf(azureProperties.getProperty(CONNECTION_TIMEOUT_MS)));
+        setCloudRequestTimeout(Long.valueOf(azureProperties.getProperty(REQUEST_TIMEOUT_MS)));
+        break;
+      case CUSTOMIZED:
+        setCloudInfoSources(cloudConfig.getCloudInfoSources());
+        setCloudInfoProcessorName(cloudConfig.getCloudInfoProcessorName());
+        break;
+      default:
+        LOG.info("Unsupported cloud provider: {}", cloudConfig.getCloudProvider());
```
jiajunwang: nit, usually, if the error has been thrown to the caller, I will suggest don't log here.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -0,0 +1,74 @@
...
+  private String _version;
+  private long _healthReportLatency;
+  private HelixCloudProperty _helixCloudProperty;
+
+  /**
+   * Initialize Helix manager property with default value
+   * @param properties helix manager related properties input as a map
+   * @param cloudConfig cloudConfig read from Zookeeper
+   */
+  public HelixManagerProperty(Properties properties, CloudConfig cloudConfig) {
```
jiajunwang: nit, "properties" -> "helixManagerProperties" to be more specific?

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * Retrieve Helix manager property. It returns the property object with default values.
+   * Clients may override these values.
+   */
+  public HelixManagerProperty getHelixManagerProperty(String zkAddress, String clusterName) {
+    ConfigAccessor configAccessor = new ConfigAccessor(zkAddress);
+    CloudConfig cloudConfig = configAccessor.getCloudConfig(clusterName);
+    if (cloudConfig == null) {
+      CloudConfig.Builder builder = new CloudConfig.Builder(clusterName);
```
jiajunwang: nit, could be simpler by adding a buildEmptyCloudConfig method which returns a static cloudconfig instance which has been configured as cloudEnabled = false.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkClusterManager.java
```diff
@@ -54,27 +54,28 @@
...
       _gZkClient.deleteRecursively("/" + clusterName);
     }
 
-    ZKHelixManager controller =
-        new ZKHelixManager(clusterName, null, InstanceType.CONTROLLER, ZK_ADDR);
-
     try {
-      controller.connect();
-      Assert.fail("Should throw HelixException if initial cluster structure is not setup");
+      ZKHelixManager controller =
```
jiajunwang: This is concerning. Considering if our customer code try-catch in the same old way that we have, their try-catch logic is required to be updated as well. This is not backward compatible and may fail our customer's code. Let's talk and see if there is a way to lazily load the configs so the behavior could be kept the same.

zhangmeng916: I changed the logic so that backward compatibility can be maintained. Basically, current logic allow constructing a ZKHelixManager without setting up the cluster. So we have to allow a default cloud config setup before cluster is set up. I think it is ok as we just set enabled to be false.

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+    Properties properties = new Properties();
+    try {
+      InputStream stream = Thread.currentThread().getContextClassLoader()
+          .getResourceAsStream(HELIX_PARTICIPANT_PROPERTY_FILE);
+      properties.load(stream);
+    } catch (IOException e) {
+      String errMsg = String.format("failed to open Helix participant properties file: %s",
+          HELIX_PARTICIPANT_PROPERTY_FILE);
+      LOG.error(errMsg);
```
jiajunwang: nit, like I suggested above, we don't need this error to be logged since it will be thrown.

zhangmeng916:
Final commit message "Add Helix property factory and class"

zhangmeng916:
Final commit message 
Add Helix property factory and class.
- Add Helix property singleton factory. The factory returns Helix property with default values, and clients may override these values;
- Add Helix manager property. It further holds Helix cloud properties and some other properties specific to the manager, defined in helix-manager.properties file.
- Add Helix cloud property. It holds cloud related properties, which comes from cloud config and client specified file, e.g., for Azure, the file is azure-cloud.properties.

PR #655, opened 2019-12-12
Add an option for using dedicated ZkConnection in ZkBaseDataAccessor

### LINKED ISSUE #654: Add a flag for using dedicated ZkConnection in ZkBaseDataAccessor
There are usage patterns where a user needs to create ephemeral nodes using a ZkBaseDataAccessor. We need to support this use case since we want users to avoid using ZkClient directly, so we do it in ZkBaseDataAccessor, which is Helix's wrapper API around the raw ZkClient.

narendly:
### Issues

Resolves #654 

### Description

There are usage patterns where a user needs to create ephemeral nodes using a ZkBaseDataAccessor. We need to support this use case since we want users to avoid using ZkClient directly, so we do it in ZkBaseDataAccessor, which is Helix's wrapper API around the raw ZkClient.

The default behavior would be to use a shared ZK connection resource.

### Tests

Tests for Dedicated and Shared ZkClients already exist in the repository.

[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestCardDealingAdjustmentAlgorithmV2.testAlgorithmConstructor:127 expected:<9> but was:<6>
[ERROR]   Run 1: TestCardDealingAdjustmentAlgorithmV2.testComputeMappingForDifferentReplicas:273 Total movements: 4 != expected 8, replica: 1
[ERROR]   Run 2: TestCardDealingAdjustmentAlgorithmV2.testComputeMappingForDifferentReplicas:273 Total movements: 4 != expected 8, replica: 2
[ERROR]   Run 3: TestCardDealingAdjustmentAlgorithmV2.testComputeMappingForDifferentReplicas:273 Total movements: 14 != expected 21, replica: 3
[INFO]   Run 4: PASS
[INFO]   Run 5: PASS
[INFO] 
[ERROR]   TestRebalanceRunningTask.testFixedTargetTaskAndEnabledRebalanceAndNodeAdded:298 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 883, Failures: 3, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE

------------------------
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 19.178 s - in org.apache.helix.integration.task.TestRebalanceRunningTask
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  23.770 s
[INFO] Finished at: 2019-12-12T08:56:34-08:00

[INFO] Tests run: 27, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.469 s - in org.apache.helix.controller.strategy.crushMapping.TestCardDealingAdjustmentAlgorithmV2
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 27, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  5.009 s
[INFO] Finished at: 2019-12-12T08:58:18-08:00

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -57,6 +57,24 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
+
+  // Designates which mode ZkBaseDataAccessor should be created in. If not specified, it will be
+  // created on SHARED mode.
+  enum ZkClientType {
+    /*
+     * When ZkBaseDataAccessor is created with the DEDICATED type, it supports ephemeral node
+     * creation, callback functionality, and session management. But note that this is more
+     * resource-heavy since it creates a dedicated ZK connection so should be used sparingly only
+     * when the aforementioned features are needed.
+     */
+    DEDICATED,
+    /*
+     * When ZkBaseDataAccessor is created with the SHARED type, it only supports CRUD
+     * functionalities. This will be the default mode of creation.
+     */
+    SHARED
+  }
+
   enum RetCode {
     OK,
     NODE_EXISTS,
@@ -110,7 +128,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    * @param zkAddress The zookeeper address
    */
   public ZkBaseDataAccessor(String zkAddress, ZkSerializer zkSerializer) {
-    this(zkAddress, zkSerializer, false);
+    this(zkAddress, zkSerializer, ZkClientType.SHARED);
   }
 
   /**
@@ -121,7 +139,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    * @param zkAddress The zookeeper address
    */
   public ZkBaseDataAccessor(String zkAddress, PathBasedZkSerializer pathBasedZkSerializer) {
-    this(zkAddress, pathBasedZkSerializer, false);
+    this(zkAddress, pathBasedZkSerializer, ZkClientType.SHARED);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -141,6 +157,53 @@ public boolean create(String path, T record, int options) {
...
+   * Creates a ZkBaseDataAccessor with a custom implementation of ZkSerializer.
+   * If dedicated flag is set to true, it will use a dedicated ZkConnection, which allows ephemeral
+   * node creation, callbacks, and session management.
+   * If dedicated flag is set to false, it will use a shared ZkConnection, which only allows simple
+   * CRUD operations to ZooKeeper.
+   * @param zkAddress
+   * @param zkSerializer
+   * @param dedicated
+   */
+  public ZkBaseDataAccessor(String zkAddress, ZkSerializer zkSerializer, boolean dedicated) {
```
pkuwm: Just my 2 cents: to me, passing a boolean param to a method to determine the behavior is kind of code smell.
See
* https://softwareengineering.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior
* https://martinfowler.com/bliki/FlagArgument.html

Can we have a better design for this? In my opinion, at least, we can create an enum and pass it:
```
enum ZkClientType {
  DedicatedZkClient, SharedZkClient;
}

ZkBaseDataAccessor(String zkAddress, ZkSerializer zkSerializer, ZkClientType type);
```
I think this would look better than this ugly boolean flag..

narendly: Sure. I agree with this suggestion. Let me define an enum.

narendly: Enum added.

jiajunwang:
I revisited the design doc about our ongoing ZK scalability project, it seems that we will migrate this accessor layer to be "zk realm aware". In that case, the raw ZkAddress will mostly be upgraded to a realm address, right? By then, I'm not sure if we still need the type to differentiate or the realm information would be rich enough for us to determine?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -56,6 +57,24 @@
 import org.slf4j.LoggerFactory;
 
 public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
+
+  // Designates which mode ZkBaseDataAccessor should be created in. If not specified, it will be
+  // created on SHARED mode.
+  enum ZkClientType {
```
jiajunwang: I like this type. But I would prefer if we can put it in the HelixZkClientFactory layer. There are multiple options, either we add a method in a default factory implementation so it will generate both types of the ZkClient accordingly. Or we have util for that purpose to minimize the changes.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -103,35 +122,103 @@ public ZkBaseDataAccessor(HelixZkClient zkClient) {
...
+  /**
+   * Creates a ZkBaseDataAccessor with {@link org.apache.helix.ZNRecord} as the data model.
+   * If DEDICATED, it will use a dedicated ZkConnection, which allows ephemeral
+   * node creation, callbacks, and session management.
+   * If SHARED, it will use a shared ZkConnection, which only allows simple
+   * CRUD operations to ZooKeeper.
+   * @param zkAddress
+   * @param zkClientType
+   */
+  public ZkBaseDataAccessor(String zkAddress, ZkClientType zkClientType) {
```
jiajunwang: There could be a small problem here. I reviewed our user's code where this issue was originally discussed, they are using HelixPropertyStore, which implement BaseDataAccessor. So at least, we will need to change ZkHelixPropertyStore as well.
Moreover, shall we make this method to the BaseDataAccessor interface? I think we might need a more generic design. Since the interface itself has been polluted by the ZK related concept, it could be a good time for us to clean them.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
i3wangyi: public enum? Otherwise, how client get access to the enum in the constructor? 
I think for the ease of client, the builder pattern is more straightforward to read. Otherwise, one needs to read all the comments of enums and NOTE sessions of each constructor to figure out the right one.
`ZkBaseDataAccessor.Builder(zkAddress).setSerializer(XXX).buildDedicate()`
`ZkBaseDataAccessor.Builder(zkAddress).setSerializer(XXX).buildShared()`

narendly: The naming for this was deliberate: Moving forward, we won't necessarily be using HelixZkClientFactory. There will be other ZkClient APIs that are multiple ZK-aware that will serve as the ZkClient layer for this accessor. Does this make sense?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
narendly: I don't agree that we should include this in the interface. The BaseDataAccessor interface does *not* assume the use of ZooKeeper. The change here is specific to ZooKeeper.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -103,35 +122,103 @@ public ZkBaseDataAccessor(HelixZkClient zkClient) {
...
+  /**
+   * Creates a ZkBaseDataAccessor with a custom implementation of ZkSerializer.
+   * If DEDICATED, it will use a dedicated ZkConnection, which allows ephemeral
+   * node creation, callbacks, and session management.
+   * If SHARED, it will use a shared ZkConnection, which only allows simple
+   * CRUD operations to ZooKeeper.
+   * @param zkAddress
+   * @param zkSerializer
+   */
+  public ZkBaseDataAccessor(String zkAddress, ZkSerializer zkSerializer,
```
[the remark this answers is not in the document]
narendly: Yes. @i3wangyi also pointed this out. I see the value in that proposition, but since that would be considered refactoring, I'd like to get to it at a later time.

narendly:

narendly:
There are usage patterns where a user needs to create ephemeral nodes using a ZkBaseDataAccessor. We need to support this use case since we want users to avoid using ZkClient directly, so we do it in ZkBaseDataAccessor, which is Helix's wrapper API around the raw ZkClient.

The default behavior would be to use a shared ZK connection resource.

Issue #661, opened 2019-12-18
Remove all listeners when leader controller transits to standby


i3wangyi:
The experiment shows when a leader controller transits to standby mode, the majority performance bottleneck (90%+) spent on removing the existing listeners.
Further analysis shows the listeners' removal step spent most of the time on `getChildren()` request for some paths (CURRENTSTATES, e.g). 

However, the GET request is unnecessary in this use case. When the controller is the leader of a cluster, all the listeners persisted in memory have persisted the up-to-date information of all the zookeeper paths.

Thus, the issue tracks the progress of making use of existing "unsubscribeAll" method in zkClient ot gain the performance improvement

PR #662, opened 2019-12-18
Mitigate controller leadership switch latency - improve leader -> standby

i3wangyi:
### Issues

#661 

### Description

- The code change will skip the getChildren() request on listener's removal, and the zkClient's unsubscribeAll will be used instead of unsubscribing a single path;
- It also includes some log optimization works.

### Tests

`testControllerConnectThenSessionExpire` in `TestControllerLeaderhshipChange`

[ERROR] Tests run: 890, Failures: 3, Errors: 0, Skipped: 0, Time elapsed: 3,909.929 s <<< FAILURE! - in TestSuite
[ERROR] testDeleteStoppingStuckWorkflowForcefully(org.apache.helix.integration.task.TestForceDeleteWorkflow)  Time elapsed: 60.645 s  <<< FAILURE!
	at org.apache.helix.integration.task.TestForceDeleteWorkflow.testDeleteStoppingStuckWorkflowForcefully(TestForceDeleteWorkflow.java:279)

[ERROR] testForceDeleteJobFromJobQueue(org.apache.helix.integration.task.TestDeleteJobFromJobQueue)  Time elapsed: 0.498 s  <<< FAILURE!

[ERROR] testStateTransitionTimeoutByClusterLevel(org.apache.helix.integration.paticipant.TestStateTransitionTimeoutWithResource)  Time elapsed: 38.119 s  <<< FAILURE!
	at org.apache.helix.integration.paticipant.TestStateTransitionTimeoutWithResource.testStateTransitionTimeoutByClusterLevel(TestStateTransitionTimeoutWithResource.java:196)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestStateTransitionTimeoutWithResource.testStateTransitionTimeoutByClusterLevel:196 expected:<true> but was:<false>
[ERROR]   TestForceDeleteWorkflow.testDeleteStoppingStuckWorkflowForcefully:279 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 890, Failures: 3, Errors: 0, Skipped: 0

Failed tests get passed running individually in IDE

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -116,7 +116,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
   private HelixCallbackMonitor _monitor;
 
   // TODO: make this be per _manager or per _listener instaed of per callbackHandler -- Lei
-  private CallbackProcessor _batchCallbackProcessor;
+  public CallbackProcessor _batchCallbackProcessor;
   private boolean _watchChild = true; // Whether we should subscribe to the child znode's data
                                       // change.
 
@@ -165,7 +165,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
     }
   }
 
-  class CallbackProcessor
+  public class CallbackProcessor
       extends DedupEventProcessor<NotificationContext.Type, NotificationContext> {
     private CallbackHandler _handler;
 
@@ -501,23 +501,20 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
 
   private void subscribeForChanges(NotificationContext.Type callbackType, String path,
       boolean watchChild) {
-    logger.info("Subscribing changes listener to path: " + path + ", type: " + callbackType
-        + ", listener: " + _listener);
+    logger
+        .info("Subscribing changes listener on path: {}, callbackType: {}, listener: {}, isWatchChild: {}",
+            path, callbackType, _listener, watchChild);
 
     long start = System.currentTimeMillis();
     if (_eventTypes.contains(EventType.NodeDataChanged)
         || _eventTypes.contains(EventType.NodeCreated)
         || _eventTypes.contains(EventType.NodeDeleted)) {
-      logger.info("Subscribing data change listener to path: " + path);
       subscribeDataChange(path, callbackType);
     }
 
     if (_eventTypes.contains(EventType.NodeChildrenChanged)) {
-      logger.info("Subscribing child change listener to path:" + path);
       subscribeChildChange(path, callbackType);
       if (watchChild) {
-        logger.info("Subscribing data change listener to all children for path:" + path);
-
         try {
           switch (_changeType) {
           case CURRENT_STATE:
@@ -583,8 +580,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
    * exists
    */
   public void init() {
-    logger.info("initializing CallbackHandler: " + this.toString() + " content: " + getContent());
-
+    logger.info("Initializing CallbackHandler, content: {}", getContent());
     if (_batchModeEnabled) {
       synchronized (this) {
         if (_batchCallbackProcessor != null) {
@@ -697,8 +693,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
   }
 
   void reset(boolean isShutdown) {
-    logger.info("Resetting CallbackHandler: {}. Is resetting for shutdown: {}.", this.toString(),
-        isShutdown);
+    logger.info("Reset CallbackHandler. isShutdown: {}. Content: {}", isShutdown, getContent());
     try {
       _ready = false;
       synchronized (this) {
@@ -718,6 +713,21 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
     } catch (Exception e) {
       String msg = "Exception while resetting the listener:" + _listener;
       ZKExceptionHandler.getInstance().handle(msg, e);
+    }
+  }
+
+  /**
+   * The package private method only closes the batchCallbackProcessor thread
+   * *Caution*: currently it's only used during disconnecting from ZK and the
+   * listeners unsubscription will be taken care by zkClient directly
+   */
+  void closeCallbackProcessor() {
+    _ready = false;
+    synchronized (this) {
+      if (_batchCallbackProcessor != null) {
+        _batchCallbackProcessor.shutdown();
+        _batchCallbackProcessor = null;
+      }
     }
   }
 
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -352,10 +352,8 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
 
     // This allows the listener to work with one change at a time
     synchronized (_manager) {
-      if (logger.isInfoEnabled()) {
-        logger.info(Thread.currentThread().getId() + " START:INVOKE " + _path + " listener:"
-            + _listener + " type: " + type);
-      }
+      logger.info(Thread.currentThread().getId() + " START:INVOKE " + _path + " listener:"
+          + _listener + " type: " + type);
 
       if (!_expectTypes.contains(type)) {
         logger.warn("Callback handler received event in wrong order. Listener: " + _listener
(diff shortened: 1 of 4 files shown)


i3wangyi:
Current Leader -> Standby diagram:
Current existing implementatio

i3wangyi:
After applying the changes
After changes

dasahcc:
Logically, it sounds good to me. But I think this is not the majority contribution in LeaderSwitch when controller is trying acquiring leadership, right? 

How much it reduces the overhead?

i3wangyi:
> Logically, it sounds good to me. But I think this is not the majority contribution in LeaderSwitch when controller is trying acquiring leadership, right?
> 
> How much it reduces the overhead?

No, the optimization is for leader -> standby, one data point shows a 10X latency improvement (from 1700ms -> 170ms); For optimizing the other part: standby -> leader, there're different options, I will explore them later.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -721,6 +716,21 @@ void reset(boolean isShutdown) {
     }
   }
 
+  /**
+   * The package private method only closes the batchCallbackProcessor thread
+   * *Caution*: currently it's only used during disconnecting from ZK and the
+   * listeners unsubscription will be taken care by zkClient directly
+   */
+  void closeCallbackProcessor() {
```
dasahcc: More precisely, it only closes batchProcessor. Rename the method

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -780,13 +780,10 @@ public void disconnect() {
...
       // TODO reset user defined handlers only
       // TODO Fix the issue that when connection disconnected, reset handlers will be blocked. -- JJ
-      // This is because reset logic contains ZK operations.
-      resetHandlers(true);
-
-      if (_leaderElectionHandler != null) {
-        _leaderElectionHandler.reset(true);
+      // remove every subscribed listener
+      if (_zkclient != null) {
+        _zkclient.unsubscribeAll();
```
dasahcc: Have you make sure the listeners will be add back once handleNewSession?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -583,8 +580,7 @@ private void subscribeForChanges(NotificationContext.Type callbackType, String p
    * exists
    */
   public void init() {
-    logger.info("initializing CallbackHandler: " + this.toString() + " content: " + getContent());
-
+    logger.info("Initializing CallbackHandler, content: {}", getContent());
```
jiajunwang: Why reduce the information here? The logger is constructed with CallbackHandler class, it is not clear for us which child class it might be if you remove the this.toString().

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -697,8 +693,7 @@ public void reset() {
   }
 
   void reset(boolean isShutdown) {
-    logger.info("Resetting CallbackHandler: {}. Is resetting for shutdown: {}.", this.toString(),
-        isShutdown);
+    logger.info("Reset CallbackHandler. isShutdown: {}. Content: {}", isShutdown, getContent());
```
jiajunwang: Same here, callback type is necessary.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -501,23 +501,20 @@ private void subscribeForChangesAsyn(NotificationContext.Type callbackType, Stri
...
-        + ", listener: " + _listener);
+    logger
+        .info("Subscribing changes listener on path: {}, callbackType: {}, listener: {}, isWatchChild: {}",
+            path, callbackType, _listener, watchChild);
 
     long start = System.currentTimeMillis();
     if (_eventTypes.contains(EventType.NodeDataChanged)
         || _eventTypes.contains(EventType.NodeCreated)
         || _eventTypes.contains(EventType.NodeDeleted)) {
-      logger.info("Subscribing data change listener to path: " + path);
```
jiajunwang: Why removing these logs? While debugging, how do we know if this listener is configured with what kind of _eventTypes list? Since there are still some concerns or bug reports regarding callbacks, we'd better keep the detail logs.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -780,13 +780,10 @@ public void disconnect() {
 
       // TODO reset user defined handlers only
       // TODO Fix the issue that when connection disconnected, reset handlers will be blocked. -- JJ
-      // This is because reset logic contains ZK operations.
-      resetHandlers(true);
```
jiajunwang: So we are not sending FINALIZE events on the reset anymore? I think this is problematic.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
i3wangyi: These logs are actually redundant. The private subscribeForChanges() will eventually call other private methods 
```
  private void subscribeChildChange(String path, NotificationContext.Type callbackType) {
    if (callbackType == NotificationContext.Type.INIT
        || callbackType == NotificationContext.Type.CALLBACK) {
      if (logger.isDebugEnabled()) {
        logger.debug(_manager.getInstanceName() + " subscribes child-change. path: " + path
            + ", listener: " + _listener);
      }
      _zkClient.subscribeChildChanges(path, this);
    } else if (callbackType == NotificationContext.Type.FINALIZE) {
      logger.info(_manager.getInstanceName() + " unsubscribe child-change. path: " + path
          + ", listener: " + _listener);

      _zkClient.unsubscribeChildChanges(path, this);
    }
  }

  private void subscribeDataChange(String path, NotificationContext.Type callbackType) {
    if (callbackType == NotificationContext.Type.INIT
        || callbackType == NotificationContext.Type.CALLBACK) {
      if (logger.isDebugEnabled()) {
        logger.debug(_manager.getInstanceName() + " subscribe data-change. path: " + path
            + ", listener: " + _listener);
      }
      _zkClient.subscribeDataChanges(path, this);
    } else if (callbackType == NotificationContext.Type.FINALIZE) {
      logger.info(_manager.getInstanceName() + " unsubscribe data-change. path: " + path
          + ", listener: " + _listener);

      _zkClient.unsubscribeDataChanges(path, this);
    }
  }
```
To perform the actual listener subscription. I think keeping the verbose log in the last step is sufficient enough and you know helix.log is flooded with all listeners logs

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
i3wangyi: The `getContent()` method has the richest information to tell everything
```
public String getContent() {
    return "CallbackHandler{" + "_watchChild=" + _watchChild + ", _preFetchEnabled="
        + _preFetchEnabled + ", _batchModeEnabled=" + _batchModeEnabled + ", _path='" + _path + '\''
        + ", _listener=" + _listener + ", _changeType=" + _changeType + ", _manager=" + _manager
        + ", _zkClient=" + _zkClient + '}';
  }
```

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
i3wangyi: The callback type is not passed in from the caller in the beginning. Do you want to pass that information? And getContent gives you all the information

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
i3wangyi: done.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
i3wangyi: Yeah. That's the design of using `unsubscribeAll`. Because the current callback handler's FINALIZE method is pretty expensive (involving read all children node under currentstates then remove them one by one). 
I could perform tests that prove without the FINALIZE event, HelixManager could clean up all information when `disconnect()`. And the single callback handler could still use the FINALIZE event to un-register itself.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
i3wangyi: Good point! I will add the test (expire the session and compare the callhandlers before and after)

i3wangyi: #681 is checked and it will be used to test details of this change

i3wangyi: Added the test case of validating callback handlers before and after session expiry in `TestControllerLeaderhshipChange`

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang: getContent should be good enough. Good point.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
jiajunwang: The thing is that our customers also depend on the FINALIZE event to do the cleanup. I believe with this change disconnect will clean up all the listeners. But you break the agreement.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -501,23 +499,20 @@ private void subscribeForChangesAsyn(NotificationContext.Type callbackType, Stri
 
   private void subscribeForChanges(NotificationContext.Type callbackType, String path,
       boolean watchChild) {
-    logger.info("Subscribing changes listener to path: " + path + ", type: " + callbackType
-        + ", listener: " + _listener);
+    logger.info(
+        "START:INVOKE subscribing changes listener on path: {}, callbackType: {}, listener: {}, isWatchChild: {}",
```
jiajunwang: START:INVOKE/END:INVOKE are for the callbacks. Please don't use it here. It will be confusing to whoever debug later.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -721,6 +714,21 @@ void reset(boolean isShutdown) {
     }
   }
 
+  /**
+   * The package private method only closes the batchCallbackProcessor thread
+   * *Caution*: currently it's only used during disconnecting from ZK and the
+   * listeners unsubscription will be taken care by zkClient directly
+   */
+  void closeBatchCallbackProcessor() {
```
jiajunwang: Besides my concern of not sending FINALIZE, this method is a subset of void reset(boolean isShutdown); I think you can avoid duplicate code.

Of course, please address my concern about not having the FINALIZE first. Then we can revisit this one.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
i3wangyi: @jiajunwang I understand your concern. With the current helix structure & library dependency, many times there's really no easy way to satisfy everything while still keeping the code clean & short IMO, even we overall agree with the overall design here.  Let's take it offline.

I will try to resolve rest concerns as much as I could and propose a new draft code structure

jiajunwang: Sure, feel free to call for a short meeting with more people. My point is that we shall never trade-off correctness or functionality for performance.

PR #671, opened 2019-12-24
Add ZooKeeperAccessor to helix-rest

### LINKED ISSUE #667: Add a path-based get() to helix-rest
We want to enhance helix-rest's functionality and rebrand it so that it will become Helix API Service. This means that we will also have REST endpoints that provide a more general usage of ZooKeeper (CRUD with a RW path for example).

Deliverable:
1. Provide a REST endpoint that does a get() based on a ZK path

### LINKED ISSUE #668: Add a path-based exists() to helix-rest
We want to enhance helix-rest's functionality and rebrand it so that it will become Helix API Service. This means that we will also have REST endpoints that provide a more general usage of ZooKeeper (CRUD with a RW path for example).

Deliverable:

    Provide a REST endpoint that does a get() based on a ZK path (no watcher functionality needed)

### LINKED ISSUE #669: Add a path-based getChildren() to helix-rest
We want to enhance helix-rest's functionality and rebrand it so that it will become Helix API Service. This means that we will also have REST endpoints that provide a more general usage of ZooKeeper (CRUD with a RW path for example).

Deliverable:
1. Add the getChildren REST endpoint in helix-rest

narendly:
### Issues

Fixes #667 
Fixes #668 
Fixes #669 

### Description

In order to provide basic ZkClient functionalities (CRUD) in helix-rest, we want to add a ZooKeeperAccessor class that supports endpoints like get(), getChildren(), and exists().

Changelist:
1. Fix the double-checked locking idiom for byte array ZkBaseDataAccessor
2. Add ZooKeeperAccessor
3. Add tests

### Tests

TestZooKeeperAccessor.java

[INFO] Tests run: 93, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 21.96 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 93, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  27.300 s
[INFO] Finished at: 2020-01-13T10:02:47-08:00

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
@@ -47,8 +47,8 @@ public class ServerContext {
   private ZKHelixAdmin _zkHelixAdmin;
   private ClusterSetup _clusterSetup;
   private ConfigAccessor _configAccessor;
-  // The lazy initialized base data accessor that reads/writes byte array to ZK
-  private ZkBaseDataAccessor<byte[]> _byteArrayBaseDataAccessor;
+  // A lazily-initialized base data accessor that reads/writes byte array to ZK
+  private volatile ZkBaseDataAccessor<byte[]> _byteArrayZkBaseDataAccessor;
   // 1 Cluster name will correspond to 1 helix data accessor
   private final Map<String, HelixDataAccessor> _helixDataAccessorPool;
   // 1 Cluster name will correspond to 1 task driver
@@ -122,27 +122,34 @@ public class ServerContext {
     }
   }
 
-  public ZkBaseDataAccessor<byte[]> getByteArrayBaseDataAccessor() {
-    if (_byteArrayBaseDataAccessor == null) {
-      synchronized (this) {
-        if (_byteArrayBaseDataAccessor == null) {
-          _byteArrayBaseDataAccessor = new ZkBaseDataAccessor<>(_zkAddr, new ZkSerializer() {
-            @Override
-            public byte[] serialize(Object o)
-                throws ZkMarshallingError {
-              throw new UnsupportedOperationException("Serialize is not supported yet!");
-            }
-
-            @Override
-            public Object deserialize(byte[] bytes)
-                throws ZkMarshallingError {
-              return bytes;
-            }
-          });
-        }
-      }
+  /**
+   * Returns a lazily-instantiated ZkBaseDataAccessor for the byte array type.
+   * @return
+   */
+  public ZkBaseDataAccessor<byte[]> getByteArrayZkBaseDataAccessor() {
+    ZkBaseDataAccessor<byte[]> byteArrayZkBaseDataAccessor = _byteArrayZkBaseDataAccessor;
+    if (byteArrayZkBaseDataAccessor != null) { // First check (no locking)
+      return byteArrayZkBaseDataAccessor;
     }
-    return _byteArrayBaseDataAccessor;
+
+    synchronized (this) {
+      if (_byteArrayZkBaseDataAccessor == null) { // Second check (with locking)
+        _byteArrayZkBaseDataAccessor = new ZkBaseDataAccessor<>(_zkAddr, new ZkSerializer() {
+          @Override
+          public byte[] serialize(Object o)
+              throws ZkMarshallingError {
+            throw new UnsupportedOperationException("serialize() is not supported yet!");
+          }
+
+          @Override
+          public Object deserialize(byte[] bytes)
+              throws ZkMarshallingError {
+            return bytes;
+          }
+        });
+      }
+      return _byteArrayZkBaseDataAccessor;
+    }
   }
 
   public void close() {
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
@@ -138,7 +138,8 @@ public class ServerContext {
           @Override
           public byte[] serialize(Object o)
               throws ZkMarshallingError {
-            throw new UnsupportedOperationException("serialize() is not supported yet!");
+            // TODO: Support serialize for write methods if necessary
+            throw new UnsupportedOperationException("serialize() is not supported.");
           }
 
           @Override
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
@@ -48,6 +48,7 @@ public class ServerContext {
   private ClusterSetup _clusterSetup;
   private ConfigAccessor _configAccessor;
   // A lazily-initialized base data accessor that reads/writes byte array to ZK
+  // TODO: Only read (deserialize) is supported at this time. This baseDataAccessor should support write (serialize) as needs arise
   private volatile ZkBaseDataAccessor<byte[]> _byteArrayZkBaseDataAccessor;
   // 1 Cluster name will correspond to 1 helix data accessor
   private final Map<String, HelixDataAccessor> _helixDataAccessorPool;
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
@@ -30,6 +30,7 @@ import org.apache.helix.PropertyPathBuilder;
 import org.apache.helix.ZNRecord;
 import org.apache.helix.manager.zk.ZNRecordSerializer;
 import org.apache.helix.manager.zk.ZkBaseDataAccessor;
+import org.apache.helix.rest.server.resources.zookeeper.ZooKeeperAccessor;
 import org.codehaus.jackson.node.ObjectNode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -55,8 +56,8 @@ public class PropertyStoreAccessor extends AbstractHelixResource {
   public Response getPropertyByPath(@PathParam("clusterId") String clusterId,
       @PathParam("path") String path) {
     path = "/" + path;
-    if (!isPathValid(path)) {
-      LOG.info("The propertyStore path {} is invalid for cluster {}", path, clusterId);
+    if (!ZooKeeperAccessor.isPathValid(path)) {
+      LOG.error("The propertyStore path {} is invalid for cluster {}", path, clusterId);
       return badRequest(
           "Invalid path string. Valid path strings use slash as the directory separator and names the location of ZNode");
     }
@@ -76,19 +77,5 @@ public class PropertyStoreAccessor extends AbstractHelixResource {
       throw new WebApplicationException(Response.status(Response.Status.NOT_FOUND)
           .entity(String.format("The property store path %s doesn't exist", recordPath)).build());
     }
-  }
-
-  /**
-   * Valid matches:
-   * /
-   * /abc
-   * /abc/abc/abc/abc
-   * Invalid matches:
-   * null or empty string
-   * /abc/
-   * /abc/abc/abc/abc/
-   **/
-  private static boolean isPathValid(String path) {
-    return path.matches("^/|(/[\\w-]+)+$");
   }
 }
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
@@ -57,7 +57,7 @@ public class PropertyStoreAccessor extends AbstractHelixResource {
       @PathParam("path") String path) {
     path = "/" + path;
     if (!ZooKeeperAccessor.isPathValid(path)) {
-      LOG.error("The propertyStore path {} is invalid for cluster {}", path, clusterId);
+      LOG.info("The propertyStore path {} is invalid for cluster {}", path, clusterId);
       return badRequest(
           "Invalid path string. Valid path strings use slash as the directory separator and names the location of ZNode");
     }
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
@@ -19,24 +19,21 @@ package org.apache.helix.rest.server.resources.zookeeper;
  * under the License.
  */
 
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 
+import com.google.common.collect.ImmutableMap;
 import org.apache.helix.AccessOption;
-import org.apache.helix.HelixAdmin;
-import org.apache.helix.ZNRecord;
 import org.apache.helix.manager.zk.ZkBaseDataAccessor;
 import org.apache.helix.rest.common.ContextPropertyKeys;
 import org.apache.helix.rest.server.ServerContext;
 import org.apache.helix.rest.server.resources.AbstractResource;
-import org.apache.helix.rest.server.resources.helix.ClusterAccessor;
-import org.codehaus.jackson.node.ObjectNode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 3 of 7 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
```diff
@@ -55,8 +56,8 @@
   public Response getPropertyByPath(@PathParam("clusterId") String clusterId,
       @PathParam("path") String path) {
     path = "/" + path;
-    if (!isPathValid(path)) {
-      LOG.info("The propertyStore path {} is invalid for cluster {}", path, clusterId);
+    if (!ZooKeeperAccessor.isPathValid(path)) {
+      LOG.error("The propertyStore path {} is invalid for cluster {}", path, clusterId);
```
pkuwm: Nit, I don't think this error message is appropriate. I remembered @lei-xia reviewed this in the previous ticket, this path invalidation is not an error in the server. So keeping info (or warn) may be good enough.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -0,0 +1,181 @@
...
+   * Returns a list of children ZNode names given the path for the parent ZNode.
+   * @param path
+   * @return list of child ZNodes
+   */
+  private Response getChildren(String path) {
+    if (!isPathValid(path)) {
+      String errMsg = "getChildren(): The given path {} is not a valid ZooKeeper path!" + path;
+      LOG.error(errMsg);
+      return badRequest(errMsg);
+    }
```
pkuwm: 1. {} doesn't apply here.
2. This is a code block that could be reused in other methods. Can we make it a method to reduce duplicate code?
3. Shall we really need the log level at `error`? Maybe other reviewers could give opinions. @lei-xia @jiajunwang @i3wangyi

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -0,0 +1,181 @@
...
+          // a Base64 String! The REST client (user) must convert the resulting String back into
+          // a byte[] using Base64.
+          return JSONRepresentation(binaryResult);
+        case getStringData:
+          Map<String, String> stringResult =
+              ImmutableMap.of(ZooKeeperCommand.getStringData.name(), new String(bytes));
+          return JSONRepresentation(stringResult);
+        default:
+          LOG.error("Unsupported command :" + command);
+          return badRequest("Unsupported command :" + command);
```
pkuwm: Nit, space after the colon might be easier to read in the log, `command: abcd`?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -0,0 +1,181 @@
...
+
+    // Need to prepend a "/" since JAX-RS regex removes it
+    path = "/" + path;
+    switch (cmd) {
+      case exists:
+        return exists(path);
+      case getBinaryData:
+        return getData(path, cmd);
+      case getStringData:
+        return getData(path, cmd);
```
pkuwm: These 2 get data commands could be combined?
```
       case getBinaryData:
       case getStringData:
         return getData(path, cmd);
```
Or, can we split getData() to have 2 methods: getBinaryData() and getStringData(). I prefer this. Since we already check the command types here, it would be cleaner just to call the exact APIs, instead of checking the command in getData() again, which is passing the command enum and use an if-else to check.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PropertyStoreAccessor.java
narendly: That makes sense. This is not the error from the server's perspective. Will change it back to info.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
narendly: Will combine.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
narendly: Added

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
narendly: 1. Fixed. Good catch!
2. This code needs some customization to differentiate the log message. If you have a better idea on how to refactor, please give a suggestion.
3. This should be an info call as mentioned above since it doesn't indicate there was an error on the server logic.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  @GET
+  @Path("{path: .+}")
+  public Response get(@PathParam("path") String path, @QueryParam("command") String commandStr) {
+    ZooKeeperCommand cmd;
+    try {
+      cmd = ZooKeeperCommand.valueOf(commandStr);
+    } catch (Exception e) {
+      return badRequest("Invalid ZooKeeper command: " + commandStr);
```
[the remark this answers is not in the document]
narendly: I agree. Either option looks good.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -0,0 +1,181 @@
...
+ * ZooKeeperAccessor provides methods for accessing ZooKeeper resources (ZNodes).
+ * It provides basic ZooKeeper features supported by ZkClient.
+ */
+@Path("/zookeeper")
+public class ZooKeeperAccessor extends AbstractResource {
+  private static final Logger LOG = LoggerFactory.getLogger(ZooKeeperAccessor.class.getName());
+  private ZkBaseDataAccessor<byte[]> _zkBaseDataAccessor;
+
+  public enum ZooKeeperCommand {
+    exists, getBinaryData, getStringData, getChildren
```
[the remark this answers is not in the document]
narendly: Good point. But this is the convention for Helix rest. Also I think it looks slightly better in a REST call as in

?command=getChildren 
vs
?command=GET_CHILDREN

helix-rest/src/test/java/org/apache/helix/rest/server/TestZooKeeperAccessor.java
```diff
@@ -0,0 +1,160 @@
...
+  public void testExists()
+      throws IOException {
+    String path = "/path";
+    Assert.assertFalse(_testBaseDataAccessor.exists(path, AccessOption.PERSISTENT));
+    Map<String, Boolean> result;
+    String data = new JerseyUriRequestBuilder("zookeeper{}?command=exists").format(path)
+        .isBodyReturnExpected(true).get(this);
+    result = OBJECT_MAPPER.readValue(data, HashMap.class);
+    Assert.assertTrue(result.containsKey("exists"));
+    Assert.assertEquals(result.get("exists"), Boolean.FALSE);
```
NealSun96: Nit: could be assertFalse (and assertTrue later in this test)

narendly: Fixed.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -0,0 +1,183 @@
...
+      case exists:
+        return exists(path);
+      case getBinaryData:
+      case getStringData:
+        return getData(path, cmd);
+      case getChildren:
+        return getChildren(path);
+      default:
+        LOG.error("Unsupported command :" + commandStr);
+        return badRequest("Unsupported command :" + commandStr);
```
pkuwm: Nit, space after colon? Just for friendly user experience.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
pkuwm: 2. I think for the response, the method name is not necessary. I would not expect the java method name in a Rest API response because my command is already provided in the API. Instead, the message "The given path is not a valid ZooKeeper path: " is good enough. 
In the server log, I agree it is OK to add the info for clearly debugging. And maybe a more generic message "Getting children" is better than the method name itself?

4. One more thing, from the user perspective, I am not sure if we would like provide a specific message for the case with trailing slash: `/abc/`, which I think it might be easy to make a mistake by adding the trailing slash.
Shall we provide a message for in the bad request specifically to help users realize the error? If a user thinks a path with trailing slash is OK, it would be confusing and time costly for the user to troubleshoot.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -0,0 +1,183 @@
...
+  private Response getData(String path, ZooKeeperCommand command) {
+    if (!isPathValid(path)) {
+      String errMsg = "getData(): The given path is not a valid ZooKeeper path: " + path;
+      LOG.info(errMsg);
+      return badRequest(errMsg);
+    }
+
+    if (_zkBaseDataAccessor.exists(path, AccessOption.PERSISTENT)) {
+      byte[] bytes = _zkBaseDataAccessor.get(path, null, AccessOption.PERSISTENT);
+      String s = new String(bytes);
```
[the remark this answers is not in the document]
narendly: Good catch. Removed

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
narendly: Fixed

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
narendly: 2. I moved the common path-checking logic to the get() function.
3. This sounds like a nice feature, but we could add this in the future. The Javadoc contains this and the user manual will also contain this information, so I don't think it's completely necessary right now.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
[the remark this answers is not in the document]
narendly: I think it's the right direction, but I don't think it's completely necessary since what we have right now works and doesn't harm readability. Let's consider making this change when it becomes necessary and keep things simple?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -0,0 +1,174 @@
...
+      case exists:
+        return exists(path);
+      case getBinaryData:
+      case getStringData:
+        return getData(path, cmd);
+      case getChildren:
+        return getChildren(path);
+      default:
+        LOG.error("Unsupported command: " + commandStr);
+        return badRequest("Unsupported command: " + commandStr);
```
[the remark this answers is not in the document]
narendly: Done

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -0,0 +1,183 @@
...
+          // a Base64 String! The REST client (user) must convert the resulting String back into
+          // a byte[] using Base64.
+          return JSONRepresentation(binaryResult);
+        case getStringData:
+          Map<String, String> stringResult =
+              ImmutableMap.of(ZooKeeperCommand.getStringData.name(), new String(bytes));
+          return JSONRepresentation(stringResult);
+        default:
+          LOG.error("Unsupported command : " + command);
+          return badRequest("Unsupported command : " + command);
```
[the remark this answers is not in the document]
narendly: Done

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -47,8 +47,8 @@
   private ZKHelixAdmin _zkHelixAdmin;
   private ClusterSetup _clusterSetup;
   private ConfigAccessor _configAccessor;
-  // The lazy initialized base data accessor that reads/writes byte array to ZK
-  private ZkBaseDataAccessor<byte[]> _byteArrayBaseDataAccessor;
+  // A lazily-initialized base data accessor that reads/writes byte array to ZK
```
jiajunwang: Is it only supporting read at this version? If so, can we update the name accordingly? _byteArrayZkReadOnlyAccessor?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -122,27 +122,34 @@ public HelixDataAccessor getDataAccssor(String clusterName) {
...
+      return byteArrayZkBaseDataAccessor;
+    }
+
+    synchronized (this) {
+      if (_byteArrayZkBaseDataAccessor == null) { // Second check (with locking)
+        _byteArrayZkBaseDataAccessor = new ZkBaseDataAccessor<>(_zkAddr, new ZkSerializer() {
+          @Override
+          public byte[] serialize(Object o)
+              throws ZkMarshallingError {
+            throw new UnsupportedOperationException("serialize() is not supported yet!");
```
jiajunwang: nit, I think "yet" is not necessary. Just say "serialize() is not supported!" or "write is not supported!"?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -0,0 +1,176 @@
...
+
+
+/**
+ * ZooKeeperAccessor provides methods for accessing ZooKeeper resources (ZNodes).
+ * It provides basic ZooKeeper features supported by ZkClient.
+ */
+@Path("/zookeeper")
+public class ZooKeeperAccessor extends AbstractResource {
+  private static final Logger LOG = LoggerFactory.getLogger(ZooKeeperAccessor.class.getName());
+  private ZkBaseDataAccessor<byte[]> _zkBaseDataAccessor;
```
jiajunwang: The reason we have this field is for using it in the private methods, right?
Passing the accessor directly to those private methods would be simpler and help to decouple the logic. Can we do that?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
jiajunwang: Sorry that I haven't reviewed this part earlier. But I would actually prefer having 2 separate private methods.
1. In this way, the only method that knows cmd meaning is this one. The private method does not need to check cmd.
2. I can see the logic is almost the same. So we can actually have one private method that returns a byte array. Then wrap it with String constructor and then construct the response here.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
jiajunwang: Since the comment here does not match what the accessor can do, can we at least comment or add a TODO here?

narendly: Will add.

narendly:

narendly:
Final commit message:

Add ZooKeeperAccessor to helix-rest

In order to provide basic ZkClient functionalities (CRUD) in helix-rest, we want to add a ZooKeeperAccessor class that supports endpoints like get(), getChildren(), and exists().

Changelist:

Fix the double-checked locking idiom for byte array ZkBaseDataAccessor
Add ZooKeeperAccessor
Add tests

PR #677, opened 2020-01-09
Add method to wait and return established session's ID

### LINKED ISSUE #676: Add a method to return established session in zkclient
Current zkclient's waitUntilConnected does not return a zk session ID. zkClient's getSessionId() could bring in session race condition: session A is connected in waitUntilConnected, but when zkClient.getSession() is called in zkHelixManager, session A might be already expired and so zkClient.getSession() gets session B. This session ID is critical for the firs time handling new session after zkclient is created in zkHelixManager.

pkuwm:
### Issues

Fixes #676 

### Description

Current zkclient's waitUntilConnected does not return a zk session ID. zkClient's getSessionId() could bring in session race condition: session A is connected in waitUntilConnected, but when zkClient.getSession() is called in zkHelixManager, session A might be already expired and so zkClient.getSession() gets session B. This session ID is critical for the firs time handling new session after zkclient is created in zkHelixManager.

Solution: add a new method waitForEstablishedSession() to wait for SynCconnected state and return the session id before unlocking the eventLock.

### Tests

testWaitForEstablishedSession

Tests all pass.

[INFO] Tests run: 897, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3,384.61 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 897, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  56:24 min
[INFO] Finished at: 2020-01-08T22:38:20-08:00

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
@@ -699,19 +699,15 @@ public class ZKHelixManager implements HelixManager, IZkStateListener {
     int retryCount = 0;
     while (retryCount < 3) {
       try {
-        // TODO: get session id from waitUntilConnected to avoid race condition
-        if (!_zkclient.waitUntilConnected(_connectionInitTimeout, TimeUnit.MILLISECONDS)) {
-          throw new ZkTimeoutException(
-              "Unable to connect to zookeeper server within timeout: " + _connectionInitTimeout
-                  + " ms.");
-        }
+        final long sessionId =
+            _zkclient.waitForEstablishedSession(_connectionInitTimeout, TimeUnit.MILLISECONDS);
         handleStateChanged(KeeperState.SyncConnected);
         /*
          * This listener is subscribed after SyncConnected and firing new session events,
          * which means this listener has not yet handled new session, so we have to handle new
          * session here just for this listener.
          */
-        handleNewSession(ZKUtil.toHexSessionId(_zkclient.getSessionId()));
+        handleNewSession(ZKUtil.toHexSessionId(sessionId));
         break;
       } catch (HelixException e) {
         LOG.error("fail to createClient.", e);
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/client/HelixZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/client/HelixZkClient.java
@@ -6,6 +6,7 @@ import java.util.concurrent.TimeUnit;
 import org.I0Itec.zkclient.DataUpdater;
 import org.I0Itec.zkclient.IZkChildListener;
 import org.I0Itec.zkclient.IZkDataListener;
+import org.I0Itec.zkclient.exception.ZkTimeoutException;
 import org.I0Itec.zkclient.serialize.SerializableSerializer;
 import org.I0Itec.zkclient.serialize.ZkSerializer;
 import org.apache.helix.manager.zk.BasicZkSerializer;
@@ -181,6 +182,28 @@ public interface HelixZkClient {
 
   // ZK state control
   boolean waitUntilConnected(long time, TimeUnit timeUnit);
+
+  /**
+   * Waits for SyncConnected state and returns a valid session ID(non-zero). The implementation of
+   * this method should wait for SyncConnected state and ZK session to be established, and should
+   * guarantee the established session's ID is returned before keeper state changes.
+   *
+   * Please note: this default implementation may have race condition issue and return an unexpected
+   * session ID that is zero or another new session's ID. The default implementation is for backward
+   * compatibility purpose.
+   *
+   * @param timeout Max waiting time for connecting to ZK server.
+   * @param timeUnit Time unit for the timeout.
+   * @return A valid ZK session ID which is non-zero.
+   */
+  default long waitForEstablishedSession(long timeout, TimeUnit timeUnit) {
+    if (!waitUntilConnected(timeout, timeUnit)) {
+      throw new ZkTimeoutException(
+          "Failed to get established session because connecting to ZK server has timed out in "
+              + timeout + " " + timeUnit);
+    }
+    return getSessionId();
+  }
 
   String getServers();
 
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -1371,12 +1371,11 @@ public class ZkClient implements Watcher {
   public long waitForEstablishedSession(long timeout, TimeUnit timeUnit) {
     validateCurrentThread();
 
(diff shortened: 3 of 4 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -2136,4 +2146,24 @@ public int hashCode() {
...
     }
   }
+
+  private void validateCurrentThread() {
+    if (_zookeeperEventThread != null && Thread.currentThread() == _zookeeperEventThread) {
+      throw new IllegalArgumentException("Must not be done in the zookeeper event thread.");
+    }
+  }
+
+  private void waitForKeeperStateUntilDeadline(KeeperState keeperState, Date deadline) {
```
jiajunwang: Why not put the lock/check current thread logic inside this method as well?

pkuwm: Synced with @jiajunwang offline. Either works OK. We agree to add a locking layer for returning session, so there are less code changes and it is easier to maintain.

pkuwm:

Commit message:
```
zkClient's getSessionId() could bring in session race condition: session A is connected in waitUntilConnected, but when zkClient.getSession() is called in zkHelixManager, session A might be already expired and so zkClient.getSession() gets session B. This session ID is critical for the firs time handling new session after zkclient is created in zkHelixManager.

Solution: add a new method waitForEstablishedSession() to wait for SynCconnected state and return the session id before unlocking the eventLock.

Change list:
- Add a new method waitForEstablishedSession()
- Add a unit test to cover the new method.
```

helix-core/src/main/java/org/apache/helix/manager/zk/client/HelixZkClient.java
```diff
@@ -182,6 +183,28 @@ void asyncSetData(final String path, Object datat, final int version,
...
+   */
+  default long waitForEstablishedSession(long timeout, TimeUnit timeUnit) {
+    if (!waitUntilConnected(timeout, timeUnit)) {
+      throw new ZkTimeoutException(
+          "Failed to get established session because connecting to ZK server has timed out in "
+              + timeout + " " + timeUnit);
+    }
+    return getSessionId();
+  }
+
```
kaisun2000: Why we need this default implementation that still have race condition?  Why default can't be the correct one? Or the default implementation just throw exception telling people they should implement the correct version?

kaisun2000:
Overall, let me summarize the history, the invariant which should lead to a working solution and where can be further improved. 

The history was about auto-reconnect. Here, ZkClient contains a connection object and connection object wrap a native Zookeeper instance (client object representing a session). ZkClient also has its own eventthread (Note this is different from Zookeeper instance eventthread. ZkClient pass its process() to native Zookeeper and this process() would run on native Zookeeper event thread processing state change and data change notification, and put the event of state change or data change to ZkClient event thread for processing. ZkClient further provides auto-reconnect functionality as Zookeeper client can have disconnected state and reconnect later automatically. However, if Zookeeper client can't reconnect in time, it would be expired by the server. This is why ZkClient would do auto-reconnect. 

Here, the invariants are the following:
1/ Each native Zookeeper object conceptually denotes one session, would have only one session ID, it would not change over time.
2/ The Zookeeper object would only get its session ID the first time connected (SynConnected state change notification arrives) Otherwise, you would see the sessionID as 0.

This invariant would be important for us to reason the correctness of the code. 

So reconnect only happens when sessionExpired event notification received and a new Zookeeper object would be created.

Note before this fix, the bug lies in the fact that an event thrown by zookeeper native thread would be later processed by ZkClient event thread. By the time ZkClient event thread process the event, they can happen when new session (thus new Zookeeper object) is on. 

The general idea is to add the sessionID information with the event running on ZKClient event thread. Thus they can detect stale session and act accordingly. (Mainly for ephemeral node creation case)

Before this batch of change, the handleNewSession() callback from ZKHelixManager would be called when sessionExpired notification delivered to the ZKClient. The thinking is that because auto-connect would happen anyway. 
In this batch of change, we need to pass the sesssionID to handleNewSession(). Thus, we can not call handleNewSession() immediately for the reason that before the first synconnect notification to the new session arrives, there is no valid sessionID. 
Thus, we made the change to deliver handleNewSession() only when the synconnected is delivered for the new Zookeeper object when reconnect happens. 

All the above are done before this change. 

However, there is still an issue. The first time when zkClient is created in the ZKHelixManager, handleNewSession would only be subscribed a little bit later. There is a possibility that this handleNewSession would be later than the first SynConnect get delivered. That is why waitForEstablishedSession() is added in this diff. 

Actually, the alternative approach can be that creating ZkClient and add the subscriber (handleNewSession in this case) can be an atomic operation. That way, it would miss the first handleNewSession. This can be a potential better alternative.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -699,19 +699,15 @@ void createClient() throws Exception {
     int retryCount = 0;
     while (retryCount < 3) {
       try {
-        // TODO: get session id from waitUntilConnected to avoid race condition
-        if (!_zkclient.waitUntilConnected(_connectionInitTimeout, TimeUnit.MILLISECONDS)) {
-          throw new ZkTimeoutException(
-              "Unable to connect to zookeeper server within timeout: " + _connectionInitTimeout
-                  + " ms.");
-        }
+        final long sessionId =
```
kaisun2000: line 698 _zkclient.subscribeStateChanges(this); 
at this implementation should always happen later than first synconnected for the reason that ZkClient constructor would call waitTillConnected to guarantee the first synconnect is delivered before proceeding. 

Thus first handleNewSession is not delivered from first Synconnect.

By the time handlNewSession is called synchronously from here, it does not necessarily have a valid sessionID. Here, we use waitForEstablishedSession to get a valid sessionID at the time. Otherwise, you have chance to get sessionID 0.  

@jiajunwang, can you confirm if my reasoning is right here?

kaisun2000: The other issue identified with HuiZhi is that 698 _zkclient.subscribeStateChanges(this) is not safe for another reason. 

The event can be like this:
1/ Zookeeper (client) created, and syconnect comes and then the constructor of ZKClient finishes.
2/ line 698 _zkclient.subscribeStateChanges(this) executed.
3/ Zookeeper expired and reconnect happens in ZKClient, then synconnnected get delivered.
4/ HandleNewSesssion get delivered.
5/ _zkCllient.waitForEstablishedSession executed and handleNewSession get called the second time. 

So the gist is still we should provide a constructor of ZKCLient with the stateChanges callback register at one shot. @jiajunwang what is your take?

jiajunwang: As we discussed offline, the issue is not newly introduced in this PR. And it is not becoming worse. I agree that we need to fix it. But not in this PR.

PR #681, opened 2020-01-15
Integration test for controller connect and disconnect

i3wangyi:
### Issues

#661 

### Description

**what** 
Add the integration test case to simulate the controller node becomes on/off leader for the test cluster; Also add the case to simulate the leadership switch case.
**why**
- To unblock the progress on PR https://github.com/apache/helix/pull/662 and help us understand the risks  when the leadership switch happens

| Is controller the leader |  Important checkpoints |
| --- | --- |
|Yes | 1. CallBack Handlers are all active and correctly registered 2. The LEADER node on ZK reflects the right leadership ~~3. Controller timer tasks are all active~~ 4. ZK connection is open | 
|No  | 1. CallBack Handlers list is empty 2. The LEADER node on ZK is either empty or showing a different leader ~~3. Controller timer tasks are all disabled~~ 4. ZK connection can be open or closed |

- To help us verify some important properties when the controller is the leader and not the leader
- It also serves a quick easy way to figure out the latency on leader -> standby and standby -> leader
- TODO: There's no good way to check if controller timer tasks are all stopped without adding a public method only for test purpose
- The diff also contains certain code style changes 
### Tests

TestControllerLeadershipChange
 - testControllerConnectThenDisconnect
 - testWhenControllerAlreadyExists
 - testWhenLeadershipSwitch

[ERROR] Tests run: 899, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3,819.652 s <<< FAILURE! - in TestSuite
[ERROR] testForceDeleteJobFromJobQueue(org.apache.helix.integration.task.TestDeleteJobFromJobQueue)  Time elapsed: 0.458 s  <<< FAILURE!

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 899, Failures: 1, Errors: 0, Skipped: 0
[INFO] 

The failing test passed when running individually in IDE

### Commits

### Documentation

### Code Quality


helix-core/src/main/java/org/apache/helix/HelixTimerTask.java
```diff
@@ -32,4 +32,9 @@
    * Stop a timer task
    */
   public abstract void stop();
+
+  /**
+   * Validate if the timer task is stopped
+   */
+  public abstract boolean isStopped();
```
pkuwm: Would this break a class which extends HelixTimerTask in an external package?

jiajunwang: If this is for test only, please keep it in the implementation classes and keep it protected.

helix-core/src/test/java/org/apache/helix/integration/controller/TestControllerLeadershipChange.java
```diff
@@ -20,29 +20,136 @@
...
+    super.beforeClass();
+    _gSetupTool.addCluster(CLUSTER_NAME, true);
+    // add some instances
+    _gSetupTool.addInstanceToCluster(CLUSTER_NAME, "TestInstance");
+    _gSetupTool.addResourceToCluster(CLUSTER_NAME, "TestResource", 10, "MasterSlave");
+  }
 
   @Test
-  public void testMissingTopStateDurationMonitoring() throws Exception {
+  public void testOnSingleController() {
```
jiajunwang: I think we have the same test case, or a similar test case. For example, TestStandAloneCMMain.java, there is more related test cases. The existing tests are actually more complicated than this one.
I'm fine if you want to re-org the test so they are together. But can we remove the old duplicated tests and ensure we keep the most complicated cases in our test list?

i3wangyi: Thanks for mentioning it! I was unaware of the existing test case at all (it's not trivial to find out). Will merge the similarities part later

helix-core/src/main/java/org/apache/helix/HelixTimerTask.java
i3wangyi: The only exposed interface from ZkHelixManager is `List<HelixTimerTask> _controllerTimerTasks = xxx` and the actual implementation class is `static class StatusDumpTask` which is not exposed neither. How could I create a protected method and make it only visible for the test?

helix-core/src/test/java/org/apache/helix/integration/controller/TestControllerLeadershipChange.java
i3wangyi: Synced offline. We both agree that the existing test is outdated and already broken. The new test case will cover the one since it's deleted in the PR

helix-core/src/main/java/org/apache/helix/HelixTimerTask.java
i3wangyi: One of the options is to cast the HelixTimerTask to a wrapper subclass only visible in test folders. And it assumes the objects in the list will be of type `StatusDumpTask`

helix-core/src/test/java/org/apache/helix/integration/controller/TestControllerLeadershipChange.java
```diff
@@ -30,19 +31,157 @@
...
+    controller.syncStart();
+    verifyControllerIsLeader(controller);
+    System.out.println(System.currentTimeMillis() - start + "ms spent on becoming the leader");
+
+    start = System.currentTimeMillis();
+    controller.syncStop();
+    verifyControllerIsNotLeader(controller);
+    verifyZKDisconnected(controller);
+
+    System.out.println(
```
jiajunwang: Please keep the test cases as silent as possible. Only log or throw error asserts on failure case.

helix-core/src/test/java/org/apache/helix/integration/controller/TestControllerLeadershipChange.java
```diff
@@ -117,16 +258,17 @@ public void testMissingTopStateDurationMonitoring() throws Exception {
...
+        .getAttribute(resourceMBeanObjectName, "PartitionTopStateHandoffDurationGauge.Max") < 500);
 
     participant.syncStop();
     manager1.disconnect();
     manager2.disconnect();
     deleteCluster(clusterName);
   }
 
-  private void setLeader(HelixManager manager) throws Exception {
+  private void setLeader(HelixManager manager)
```
jiajunwang: nit, I believe the latest style file has changed exception throws to non-wrapping. Please update your local style configuration.

i3wangyi: good catch. updated

helix-core/src/test/java/org/apache/helix/integration/controller/TestControllerLeadershipChange.java
i3wangyi: I do need these output of recording the time duration spent on L -> S and S -> L (to prove the improvements). Modify the console print to LOG so it won't be printed out at least during `mvn test`

i3wangyi:
The PR is ready to be merged into Master, approved by @jiajunwang 

Commit:
Add integration tests for verifying controller's behavior on/off leadership of a cluster

PR #682, opened 2020-01-15
Implement sharding key trie node class

### LINKED ISSUE #701: Create TrieRoutingData - a trie structure representing Zk path sharding keys and associated realm addresses
In order to support an upcoming service that offers "Zk path sharding key" look-ups, we are creating a data structure that allows fast string search with the said sharding keys. Common operations we are trying to support include "given a Zk path, find all sharding keys that contain the path" and "given a Zk path, find the sharding key contained by the path". We are implementing a trie structure, because comparing to plain string comparisons, a trie structure allows faster execution of the operations similar to the example ones.

NealSun96:
### Issues

Fixes #701 

### Description

Created a class that represents the trie nodes used for Zk path sharding keys look-ups. The class itself is a clean representation of a generic trie node, and doesn't include any additional logic by design.  

### Tests

[INFO] Tests run: 93, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 19.549 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 93, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  26.465 s
[INFO] Finished at: 2020-01-24T14:02:10-08:00

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/ShardingKeyTreeNode.java
```diff
@@ -0,0 +1,74 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ShardingKeyTreeNode {
```
i3wangyi: TrieNode more exactly, isn't it?

jiajunwang:
Could we please have the issue, full description that includes the motivation of this PR? If it is for a new feature, some kind of design would be preferred.

helix-core/src/main/java/org/apache/helix/ShardingKeyTreeNode.java
NealSun96: Agreed. Renaming accordingly.

narendly:
> Could we please have the issue, full description that includes the motivation of this PR? If it is for a new feature, some kind of design would be preferred.
+1. Please review pull request guidelines on the sidebar.

narendly:
@nealsun-linkedin Please fix the title accordingly.

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,66 @@
...
+ * under the License.
+ */
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ShardingKeyTrieNode {
+  private Map<String, ShardingKeyTrieNode> children;
+  private Map<String, String> data;
+  private String curPath;
```
narendly: Is this field necessary?

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,66 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ShardingKeyTrieNode {
+  private Map<String, ShardingKeyTrieNode> children;
+  private Map<String, String> data;
```
narendly: Why is the data field a map vs just a plain string? Or rather, you could just have the `name` field.

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,66 @@
...
+  private String curPath;
+
+  public ShardingKeyTrieNode(String curPath) {
+    this.children = new HashMap<String, ShardingKeyTrieNode>();
+    this.data = new HashMap<String, String>();
+    this.curPath = curPath;
```
narendly: Nit: `this` is not necessary in this context. Let's remove all `this`'s

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,66 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ShardingKeyTrieNode {
```
narendly: In terms of defining APIs, please review the following pages:

https://docs.oracle.com/javase/7/docs/api/javax/swing/tree/TreeNode.html
https://fasterxml.github.io/jackson-core/javadoc/2.5/com/fasterxml/jackson/core/TreeNode.html

They list some commonly-used APIs for classes like this and you could also learn about what convention is used for naming the methods.

We are not going to need all the methods in these examples so let's start simple, but please try to follow the convention as that will help with reducing confusion.

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,66 @@
+package org.apache.helix;
```
narendly: Please create a separate package for this instead of putting it in `org.apache.helix`. Since this will be used by helix-rest, I am fine with creating a package called "metadata-store" or something like that in helix-rest for the time being.

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,66 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.HashMap;
+import java.util.Map;
+
```
narendly: Could we add unit tests for this class?

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
narendly: I think it would be better if this trie was immutable. Should we make these fields `final`?

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
narendly: Could we consider using the Builder pattern here? That might make it easier for users to construct it.

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
narendly: https://medium.com/@hudsonmendes/immutable-the-builder-pattern-vs-messy-pojos-or-pocos-e6b5be894a9d

NealSun96: In the original design, we wanted a leaf node to store a key-value mapping of data, which includes ZkRealm. I'm not sure what a `name` is in this context.

narendly: What is the original design? You could also publish your rough (or finalized) design in the GitHub wiki if you'd like to refer to it in your PRs.

You _could_ use a map to store data, but it would be good to list out some examples in the form of JavaDoc.

There are pros and cons. If you choose to use one map for all data (for example, name, ZK connect string, etc.), then it'd be hard to require some of these fields (and rightly so - would a TrieNode be valid without a name?). We could consider these items carefully and determine what the right way to represent the data is.

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
pkuwm: I agree to add common used APIs to this TrieNode class, like what Hunter already mentioned. The common used APIs will definitely help your development later.

helix-core/src/main/java/org/apache/helix/ShardingKeyTrieNode.java
NealSun96: Per offline discussion, we are adding two fields, namely `name` and `isTerminal`, and we are changing the `data` field to be a `zkRealm` string field. All fields are final.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,115 @@
...
+      validate();
+      return new ShardingKeyTrieNode(_children, _isLeaf, _name, _zkRealmAddress);
+    }
+
+    public Builder setChildren(final Map<String, ShardingKeyTrieNode> children) {
+      _children = children;
+      return this;
+    }
+
+    public Builder setisLeaf(final boolean isLeaf) {
```
narendly: Nit: setIsLeaf or setLeaf?

narendly: Also as a note: you don't need to make this isLeaf (parameter) final. But it's okay.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,115 @@
...
+      if (_name == null || _name.isEmpty()) {
+        throw new IllegalArgumentException("name cannot be null or empty");
+      }
+      if (_isLeaf && (_zkRealmAddress == null || _zkRealmAddress.isEmpty())) {
+        throw new IllegalArgumentException("zkRealmAddress cannot be null or empty when the node is a terminal node");
+      }
+    }
+  }
+
+
```
narendly: Nit: empty line?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,115 @@
...
+  public ShardingKeyTrieNode(Map<String, ShardingKeyTrieNode> children, boolean isLeaf, String name,
+      String zkRealmAddress) {
+    _children = children;
+    _isLeaf = isLeaf;
+    _name = name;
+    _zkRealmAddress = zkRealmAddress;
+  }
+
+  public static class Builder {
```
narendly: Nit: move this Builder class to the very end of this class.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,115 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.Map;
+
+/**
+ * A trie node that is used to represent Zk path sharding keys. Terminal nodes contain ZkRealm addresses.
+ */
+public class ShardingKeyTrieNode {
```
narendly: Looks good so far - could we add tests for this?

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/TestShardingKeyTrieNode.java
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Testing the proper creation case, building a sharding key trie like so:
+   *     a
+   *    / \
+   *   b  e
+   *  / \
+   * c  d
+   */
+  @Test
+  public void testCreation() {
```
narendly: Could we also add a "mock" write as if a parser was translating what's in ZK into a prefix tree? Some sort of tree traversal would be good to add here.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,117 @@
...
+  public ShardingKeyTrieNode(Map<String, ShardingKeyTrieNode> children, boolean isLeaf, String name,
+      String zkRealmAddress) {
+    _children = children;
+    _isLeaf = isLeaf;
+    _name = name;
+    _zkRealmAddress = zkRealmAddress;
+  }
+
+  public Map<String, ShardingKeyTrieNode> getChildren() {
+    return _children;
```
dasahcc: Let's wrap it as unmodifiable map. Otherwise, it will be changed.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,117 @@
...
+    private boolean _isLeaf;
+    private String _name;
+    private String _zkRealmAddress;
+
+    public ShardingKeyTrieNode build() {
+      validate();
+      return new ShardingKeyTrieNode(_children, _isLeaf, _name, _zkRealmAddress);
+    }
+
+    public Builder setChildren(final Map<String, ShardingKeyTrieNode> children) {
```
dasahcc: This is children map can be dynamically changed? For example, for some services which have large amount of zk paths, if there is a new path added, would you prefer to add a new path or rebuild everything in memory?

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/TestShardingKeyTrieNode.java
NealSun96: Per offline discussion, the translation/parsing test will be provided in the parser class.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
NealSun96: Good catch, thank you! Updated.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
NealSun96: As of the current design, if changes occur on the source of truth, the entire trie will be rebuilt based on the source of truth. The reasoning of this design is as follows: 
1. This is a clean way of updating the in-memory trie; no need to worry about syncing;
2. updates to the routing data (such as adding a new path) are very rare (so much so our current design of the cached routing data is a static mapping, instead of a dynamically updated cache);
3. the cost of building a trie is not very high - given that updating is rare, the cost to rebuild the trie is not a bottleneck to anything;
4. simplicity.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public boolean isLeaf() {
+    return _isLeaf;
+  }
+
+  public String getName() {
+    return _name;
+  }
+
+  public String getZkRealmAddress() {
```
jiajunwang: Is it the case that this address won't be valid or even initialized if this is not a leaf node? If so, can we create a leaf class which extends this one and make this method available only there?
If this is not the case, please ignore what I said.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,118 @@
...
+  public ShardingKeyTrieNode(Map<String, ShardingKeyTrieNode> children, boolean isLeaf, String name,
+      String zkRealmAddress) {
+    _children = children;
+    _isLeaf = isLeaf;
+    _name = name;
+    _zkRealmAddress = zkRealmAddress;
+  }
+
+  public Map<String, ShardingKeyTrieNode> getChildren() {
+    return Collections.unmodifiableMap(_children);
```
jiajunwang: 1. The map is final
2. There is no set method to update the map.
Why not directly put it in an unmodifiable map in the constructor? So we don't need to create a map on every get.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
NealSun96: My instinct is saying a leaf class extension may be an overkill. This is because we can make an argument that the `children` variable won't be valid or even initialized if it's a leaf node, then it seems that we should have two class: `leafNode` and `nonLeafNode`, to avoid the misuse of `children`. `LeafNode` has `zkRealmAddress`, and `nonLeafNode` has `children`; both of these classes inherit from a parent class.
If we look at that purposed parent class, its only variable is `name`, which seems to be too broad for a class to me.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
NealSun96: Makes sense. Let me make the update.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
```diff
@@ -0,0 +1,118 @@
...
+  public boolean isLeaf() {
+    return _isLeaf;
+  }
+
+  public String getName() {
+    return _name;
+  }
+
+  public String getZkRealmAddress() {
+    return _zkRealmAddress;
```
narendly: This class should check if self is leaf and throw an exception if it's not. Also make that behavior clear in the JavaDoc as well.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ShardingKeyTrieNode.java
jiajunwang: What I proposed are TreeNode and LeafNode. LeafNode is the child of the TreeNode. getZkRealmAddress() is the method of LeafNode, TreeNode only supports the tree-related operations. I think it is easier because the caller does not call isLeaf() to check then call getZkRealmAddress() to fetch the information.

For easier access, every TreeNode can have a method called getLeafNodes() which returns all the leaf nodes under its branch.

Then,
1. When you adding a new realm address, construct a LeafNode then adding to the right TreeNode.
2. When you search for a node, find the TreeNode first, then call getLeafNodes() which will return all the leaves.

jiajunwang: I think it will simplify the class.

NealSun96:
Replying to both comments from @jiajunwang and @narendly here, as both comments revolve around `getZkRealmAddress()`. 
I agree that `getZkRealmAddress()` should not be used for non-leaf nodes. As of right now, I have added the exception throwing behavior per @narendly 's suggestion, however, at this moment I do not agree with this approach. Throwing an exception implies that `getZkRealmAddress()` should not be called under certain conditions, and that `zkRealmAddress` should not exist under these conditions; it is a better practice to ensure that such a function is unavailable in the first place (under these conditions), and that the object is always in a valid state. 
I believe @jiajunwang 's proposed design is one of the only ways to achieve this, but it'll be a complete redo on the current design. Per @narendly 's suggestions, I purpose to move on with the current design for now, and revisit this class as more usage to this class is consolidated. With more usage, we will have a better context on how the class should behave, which can help making a decision. This class has been stalling the progress on other components for longer than planned, which we should avoid.

jiajunwang:
> Replying to both comments from @jiajunwang and @narendly here, as both comments revolve around `getZkRealmAddress()`.
> I agree that `getZkRealmAddress()` should not be used for non-leaf nodes. As of right now, I have added the exception throwing behavior per @narendly 's suggestion, however, at this moment I do not agree with this approach. Throwing an exception implies that `getZkRealmAddress()` should not be called under certain conditions, and that `zkRealmAddress` should not exist under these conditions; it is a better practice to ensure that such a function is unavailable in the first place (under these conditions), and that the object is always in a valid state.
> I believe @jiajunwang 's proposed design is one of the only ways to achieve this, but it'll be a complete redo on the current design. Per @narendly 's suggestions, I purpose to move on with the current design for now, and revisit this class as more usage to this class is consolidated. With more usage, we will have a better context on how the class should behave, which can help making a decision. This class has been stalling the progress on other components for longer than planned, which we should avoid.

@NealSun96 Let's do the right thing. Schedule is something we need to concern. But if you leave this kind of known potential issues at this early stage, it tends to cost you much more time later to fix later.

NealSun96:
This PR is replaced by #706 , closing.

PR #688, opened 2020-01-17
Fix the watcher leakage issue

### LINKED ISSUE #687: ZkClient re-installs the watcher for a non-existence path when the node is deleted
*Root cause* 

--
pathStatRecord.recordPathStat(getStat(path, true), notificationTime);
--

The getStat(path, true) method not only returns the Stat of the path but also adds the exists watcher (exists watcher can be added even the path is non-existed). The path parameter in the context of node deleted action is the path of deleted node, which means it adds the watcher to a non-existence zookeeper path.

i3wangyi:
### Issues

(FIX #687)

### Description

ZkClient will re-install the watcher even when the node path is deleted and it leaks to zookeeper server-side watcher leakages. 
This PR tries to address the leakage issue.

**Detailed Explanation**
* Current implementation re-installs the watcher explicitly when handling either child or data event
* The updates will re-install the watcher explicitly immediately after receiving the WatchedEvent, if it's NodeDeleted event, no watcher will be set; Then in the actual data handling part, no more explicit watcher will be set.

### Tests

TestZKWatch
- testSubscribeDataChange
- testSubscribeChildChange
- testSubscribeNonExistPath

[ERROR] Failures: 
[ERROR]   TestDelayedAutoRebalanceWithRackaware.testMinimalActiveReplicaMaintain:89->TestDelayedAutoRebalance.testMinimalActiveReplicaMaintain:131->ZkTestBase.validateMinActiveAndTopStateReplica:446 Test-DB-2_1 missing LEADER replica expected:<true> but was:<false>
[ERROR]   TestHelixZkClient.testSharingZkClient:168 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 901, Failures: 2, Errors: 0, Skipped: 4
 
`TestDelayedAutoRebalanceWithRackaware.testMinimalActiveReplicaMaintain` is a flaky test and failing.
`TestHelixZkClient.testSharingZkClient` fails because the previously installed watch is deleted and we need to re-install the watch on purpose to make the test happy

-- after the fix 
[ERROR] Tests run: 901, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 3,846.059 s <<< FAILURE! - in TestSuite
[ERROR] testSubscribeDataChange(org.apache.helix.manager.zk.TestZKWatch)  Time elapsed: 0.031 s  <<< FAILURE!
java.lang.AssertionError: expected:<0> but was:<1>
	at org.apache.helix.manager.zk.TestZKWatch.testSubscribeDataChange(TestZKWatch.java:81)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestZKWatch.testSubscribeDataChange:81 expected:<0> but was:<1>
[INFO] 
[ERROR] Tests run: 901, Failures: 1, Errors: 0, Skipped: 2

The single failed test passed running in IDE

-- After merge with latest master branch
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestNoThrottleDisabledPartitions.testNoThrottleOnDisabledPartition:293 expected:<true> but was:<false>
[ERROR]   TestZKWatch.testSubscribeDataChange:81 expected:<0> but was:<1>
[INFO]
[ERROR] Tests run: 1082, Failures: 3, Errors: 0, Skipped: 2

The failed tests passed when running individually

#### Final 
[ERROR] Tests run: 1083, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 4,588.244 s <<< FAILURE! - in TestSuite
[ERROR] testMissingTopStateDurationMonitoring(org.apache.helix.integration.controller.TestControllerLeadershipChange)  Time elapsed: 4.705 s  <<< FAILURE!
	at org.apache.helix.integration.controller.TestControllerLeadershipChange.testMissingTopStateDurationMonitoring(TestControllerLeadershipChange.java:262)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 1083, Failures: 1, Errors: 0, Skipped: 0

The single failed test passed in IDE

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -11,7 +11,6 @@
 package org.apache.helix.manager.zk.zookeeper;
 
 import java.lang.reflect.Method;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
@@ -905,19 +904,6 @@ public class ZkClient implements Watcher {
         || event.getType() == Event.EventType.NodeCreated
         || event.getType() == Event.EventType.NodeChildrenChanged;
 
-    if (!stateChanged) {
-      // Need to re-install the watchers since ZK watchers are one-time-triggers
-      // Need to re-install before handling the event to minimize impact on missing events
-      if (event.getType() == EventType.NodeChildrenChanged) {
-        watchForChilds(event.getPath());
-      } else if (event.getType() != EventType.NodeDeleted) {
-        watchForData(event.getPath());
-      } else {
-        LOG.debug("Path {} is deleted", event.getPath());
-      }
-    }
-
-
     getEventLock().lock();
     try {
       // We might have to install child change event listener if a new node was created
@@ -967,10 +953,10 @@ public class ZkClient implements Watcher {
 
   private void fireAllEvents() {
     for (Entry<String, Set<IZkChildListener>> entry : _childListener.entrySet()) {
-      fireChildChangedEvents(entry.getKey(), entry.getValue());
+      fireChildChangedEvents(entry.getKey(), entry.getValue(), true);
     }
     for (Entry<String, Set<IZkDataListenerEntry>> entry : _dataListener.entrySet()) {
-      fireDataChangedEvents(entry.getKey(), entry.getValue(), OptionalLong.empty());
+      fireDataChangedEvents(entry.getKey(), entry.getValue());
     }
   }
 
@@ -1255,6 +1241,7 @@ public class ZkClient implements Watcher {
 
   private void processDataOrChildChange(WatchedEvent event, long notificationTime) {
     final String path = event.getPath();
+    final boolean pathExists = event.getType() != EventType.NodeDeleted;
 
     if (event.getType() == EventType.NodeChildrenChanged || event.getType() == EventType.NodeCreated
         || event.getType() == EventType.NodeDeleted) {
@@ -1262,7 +1249,7 @@ public class ZkClient implements Watcher {
       if (childListeners != null && !childListeners.isEmpty()) {
         // TODO recording child changed event propagation latency as well. Note this change will
         // introduce additional ZK access.
-        fireChildChangedEvents(path, childListeners);
+        fireChildChangedEvents(path, childListeners, pathExists);
       }
     }
 
@@ -1270,13 +1257,17 @@ public class ZkClient implements Watcher {
         || event.getType() == EventType.NodeCreated) {
       Set<IZkDataListenerEntry> listeners = _dataListener.get(path);
       if (listeners != null && !listeners.isEmpty()) {
-        fireDataChangedEvents(event.getPath(), listeners, OptionalLong.of(notificationTime));
+        fireDataChangedEvents(event.getPath(), listeners, OptionalLong.of(notificationTime), pathExists);
       }
     }
   }
 
+  private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry> listeners) {
+    fireDataChangedEvents(path, listeners, OptionalLong.empty(), true);
+  }
+
   private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry> listeners,
-      final OptionalLong notificationTime) {
+      final OptionalLong notificationTime, boolean pathExists) {
     try {
       final ZkPathStatRecord pathStatRecord = new ZkPathStatRecord(path);
       // Trigger listener callbacks
@@ -1285,30 +1276,28 @@ public class ZkClient implements Watcher {
             + listener.getDataListener() + " prefetch data: " + listener.isPrefetchData()) {
           @Override
           public void run() throws Exception {
-            if (!pathStatRecord.pathChecked()) {
-              pathStatRecord.recordPathStat(getStat(path), notificationTime);
-            }
-            //TODO: Use event type to check if path exists and save network request
-            if (!pathStatRecord.pathExists()) {
-              // no znode found at the path, trigger data deleted handler.
+            if (!pathExists) {
               listener.getDataListener().handleDataDeleted(path);
-            } else {
-              Object data = null;
-              if (listener.isPrefetchData()) {
-                if (LOG.isDebugEnabled()) {
-                  LOG.debug("Prefetch data for path: {}", path);
-                }
-                try {
-                  // TODO: the data is redundantly read multiple times when multiple listeners exist
-                  data = readData(path, null, true);
-                } catch (ZkNoNodeException e) {
-                  LOG.warn("Prefetch data for path: {} failed.", path, e);
-                  listener.getDataListener().handleDataDeleted(path);
-                  return;
-                }
-              }
-              listener.getDataListener().handleDataChange(path, data);
+              return;
             }
+            if (!pathStatRecord.pathChecked()) {
+              pathStatRecord.recordPathStat(getStat(path, true), notificationTime);
+            }
+            Object data = null;
+            if (listener.isPrefetchData()) {
+              if (LOG.isDebugEnabled()) {
+                LOG.debug("Prefetch data for path: {}", path);
+              }
+              try {
+                // TODO: the data is redundantly read multiple times when multiple listeners exist
+                data = readData(path, null, true);
+              } catch (ZkNoNodeException e) {
+                LOG.warn("Prefetch data for path: {} failed.", path, e);
+                listener.getDataListener().handleDataDeleted(path);
+                return;
+              }
+            }
+            listener.getDataListener().handleDataChange(path, data);
           }
         });
       }
@@ -1317,9 +1306,7 @@ public class ZkClient implements Watcher {
     }
   }
 
-  private void fireChildChangedEvents(final String path, Set<IZkChildListener> childListeners) {
-    //TODO: Use event type to check if path exists and save network request
-    boolean pathExists = exists(path);
+  private void fireChildChangedEvents(final String path, Set<IZkChildListener> childListeners, boolean pathExists) {
     try {
       for (final IZkChildListener listener : childListeners) {
         _eventThread.send(new ZkEvent("Children of " + path + " changed sent to " + listener) {
@@ -1328,8 +1315,7 @@ public class ZkClient implements Watcher {
             List<String> children = null;
             if (pathExists) {
               try {
-                //TODO: duplicate read happen when multiple child listener exists
-                // if path exists, still necessary to catch the NoNodeException? (the exception occurs under race-condition)
+                //TODO: duplicate reads when multiple child listener exists
                 children = getChildren(path);
                 if (children != null) {
                   for (String child : children) {
@@ -1340,7 +1326,6 @@ public class ZkClient implements Watcher {
                 }
               } catch (ZkNoNodeException e) {
                 LOG.warn("Get children under path: {} failed.", path, e);
-                //TODO: still handle child change if getting children failed?
                 // Continue trigger the change handler
               }
             }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -11,6 +11,7 @@
 package org.apache.helix.manager.zk.zookeeper;
 
 import java.lang.reflect.Method;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
@@ -1287,6 +1288,7 @@ public class ZkClient implements Watcher {
             if (!pathStatRecord.pathChecked()) {
               pathStatRecord.recordPathStat(getStat(path), notificationTime);
             }
+            //TODO: Use event type to check if path exists and save network request
             if (!pathStatRecord.pathExists()) {
               // no znode found at the path, trigger data deleted handler.
               listener.getDataListener().handleDataDeleted(path);
@@ -1316,20 +1318,29 @@ public class ZkClient implements Watcher {
   }
 
   private void fireChildChangedEvents(final String path, Set<IZkChildListener> childListeners) {
+    //TODO: Use event type to check if path exists and save network request
     boolean pathExists = exists(path);
     try {
       for (final IZkChildListener listener : childListeners) {
         _eventThread.send(new ZkEvent("Children of " + path + " changed sent to " + listener) {
           @Override
           public void run() throws Exception {
-            List<String> children = null;
+            List<String> children = new ArrayList<>();
             if (pathExists) {
               try {
+                //TODO: duplicate read happen when multiple child listener exists
+                // if path exists, still necessary to catch the NoNodeException? (the exception occurs under race-condition)
                 children = getChildren(path);
               } catch (ZkNoNodeException e) {
                 LOG.warn("Get children under path: {} failed.", path, e);
+                //TODO: still handle child change if getting children failed?
                 // Continue trigger the change handler
               }
+            }
+            for (String child : children) {
+              // add the exists watcher for all child path, it's to prevent watcher missing
+              // in case of node recreation shortly after deletion
+              watchForData(path + "/" + child);
             }
             listener.handleChildChange(path, children);
           }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -1325,22 +1325,24 @@ public class ZkClient implements Watcher {
         _eventThread.send(new ZkEvent("Children of " + path + " changed sent to " + listener) {
           @Override
           public void run() throws Exception {
-            List<String> children = new ArrayList<>();
+            List<String> children = null;
             if (pathExists) {
               try {
                 //TODO: duplicate read happen when multiple child listener exists
                 // if path exists, still necessary to catch the NoNodeException? (the exception occurs under race-condition)
                 children = getChildren(path);
+                if (children != null) {
+                  for (String child : children) {
+                    // add the exists watcher for all child path, it's to prevent watcher missing
+                    // in case of node recreation shortly after deletion
+                    watchForData(path + "/" + child);
+                  }
+                }
               } catch (ZkNoNodeException e) {
                 LOG.warn("Get children under path: {} failed.", path, e);
                 //TODO: still handle child change if getting children failed?
                 // Continue trigger the change handler
               }
-            }
-            for (String child : children) {
-              // add the exists watcher for all child path, it's to prevent watcher missing
-              // in case of node recreation shortly after deletion
-              watchForData(path + "/" + child);
             }
             listener.handleChildChange(path, children);
           }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -1297,6 +1297,7 @@ public class ZkClient implements Watcher {
                   LOG.debug("Prefetch data for path: {}", path);
                 }
                 try {
+                  // TODO: the data is redundantly read multiple times when multiple listeners exist
                   data = readData(path, null, true);
                 } catch (ZkNoNodeException e) {
                   LOG.warn("Prefetch data for path: {} failed.", path, e);
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -904,12 +904,18 @@ public class ZkClient implements Watcher {
         || event.getType() == Event.EventType.NodeCreated
         || event.getType() == Event.EventType.NodeChildrenChanged;
 
-    if (event.getType() == Event.EventType.NodeDeleted) {
-      if (LOG.isDebugEnabled()) {
-        String path = event.getPath();
-        LOG.debug(path);
+    if (!stateChanged) {
+      // Need to re-install the watchers since ZK watchers are one-time-triggers
+      // Need to re-install before handling the event to minimize impact on missing events
+      if (event.getType() == EventType.NodeChildrenChanged) {
+        watchForChilds(event.getPath());
+      } else if (event.getType() != EventType.NodeDeleted) {
+        watchForData(event.getPath());
+      } else {
+        LOG.debug("Path {} is deleted", event.getPath());
       }
     }
+
 
     getEventLock().lock();
     try {
@@ -1278,9 +1284,8 @@ public class ZkClient implements Watcher {
             + listener.getDataListener() + " prefetch data: " + listener.isPrefetchData()) {
           @Override
           public void run() throws Exception {
-            // Reinstall watch before listener callbacks to check the znode status
             if (!pathStatRecord.pathChecked()) {
-              pathStatRecord.recordPathStat(getStat(path, true), notificationTime);
+              pathStatRecord.recordPathStat(getStat(path), notificationTime);
             }
             if (!pathStatRecord.pathExists()) {
               // no znode found at the path, trigger data deleted handler.
@@ -1310,19 +1315,14 @@ public class ZkClient implements Watcher {
   }
 
   private void fireChildChangedEvents(final String path, Set<IZkChildListener> childListeners) {
+    boolean pathExists = exists(path);
     try {
-      final ZkPathStatRecord pathStatRecord = new ZkPathStatRecord(path);
       for (final IZkChildListener listener : childListeners) {
         _eventThread.send(new ZkEvent("Children of " + path + " changed sent to " + listener) {
           @Override
           public void run() throws Exception {
-            // Reinstall watch before listener callbacks to check the znode status
-            if (!pathStatRecord.pathChecked()) {
-              pathStatRecord.recordPathStat(getStat(path, hasListeners(path)),
-                  OptionalLong.empty());
-            }
             List<String> children = null;
-            if (pathStatRecord.pathExists()) {
+            if (pathExists) {
               try {
                 children = getChildren(path);
               } catch (ZkNoNodeException e) {
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -1313,7 +1313,7 @@ public class ZkClient implements Watcher {
           @Override
           public void run() throws Exception {
             if (!pathStatRecord.pathChecked()) {
-              pathStatRecord.recordPathStat(getStat(path, hasListeners(path) && !pathExists),
+              pathStatRecord.recordPathStat(getStat(path, hasListeners(path) && pathExists),
                   OptionalLong.empty());
             }
             List<String> children = null;
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -1282,7 +1282,7 @@ public class ZkClient implements Watcher {
             if (!pathStatRecord.pathChecked()) {
               pathStatRecord.recordPathStat(getStat(path, pathExists), notificationTime);
             }
-            if (!pathExists) {
+            if (!pathStatRecord.pathExists()) {
               listener.getDataListener().handleDataDeleted(path);
             } else {
               Object data = null;
@@ -1321,7 +1321,7 @@ public class ZkClient implements Watcher {
                   OptionalLong.empty());
             }
             List<String> children = null;
-            if (pathExists) {
+            if (!pathStatRecord.pathExists()) {
               try {
                 //TODO: duplicate reads when multiple child listener exists
                 children = getChildren(path);
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -1320,13 +1320,6 @@ public class ZkClient implements Watcher {
               try {
                 //TODO: duplicate reads when multiple child listener exists
                 children = getChildren(path);
-                if (children != null) {
-                  for (String child : children) {
-                    // add the exists watcher for all child path, it's to prevent watcher missing
-                    // in case of node recreation shortly after deletion
-                    watchForData(path + "/" + child);
-                  }
-                }
               } catch (ZkNoNodeException e) {
                 LOG.warn("Get children under path: {} failed.", path, e);
                 // Continue trigger the change handler
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
@@ -959,7 +959,7 @@ public class ZkClient implements Watcher {
       fireChildChangedEvents(entry.getKey(), entry.getValue(), true);
     }
     for (Entry<String, Set<IZkDataListenerEntry>> entry : _dataListener.entrySet()) {
-      fireDataChangedEvents(entry.getKey(), entry.getValue());
+      fireDataChangedEvents(entry.getKey(), entry.getValue(), OptionalLong.empty(), true);
     }
   }
 
@@ -1265,10 +1265,6 @@ public class ZkClient implements Watcher {
     }
   }
 
-  private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry> listeners) {
-    fireDataChangedEvents(path, listeners, OptionalLong.empty(), true);
-  }
-
   private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry> listeners,
       final OptionalLong notificationTime, boolean pathExists) {
     try {
@@ -1321,7 +1317,7 @@ public class ZkClient implements Watcher {
                   OptionalLong.empty());
             }
             List<String> children = null;
-            if (!pathStatRecord.pathExists()) {
+            if (pathStatRecord.pathExists()) {
               try {
                 //TODO: duplicate reads when multiple child listener exists
                 children = getChildren(path);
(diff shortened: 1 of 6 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1261,31 +1258,62 @@ private void processDataOrChildChange(WatchedEvent event, long notificationTime)
 
     if (event.getType() == EventType.NodeDataChanged || event.getType() == EventType.NodeDeleted
         || event.getType() == EventType.NodeCreated) {
+      boolean isPathExist = event.getType() != EventType.NodeDeleted;
```
narendly: nit: doesPathExist or pathExists

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1261,31 +1258,62 @@ private void processDataOrChildChange(WatchedEvent event, long notificationTime)
...
       Set<IZkDataListenerEntry> listeners = _dataListener.get(path);
       if (listeners != null && !listeners.isEmpty()) {
-        fireDataChangedEvents(event.getPath(), listeners, OptionalLong.of(notificationTime));
+        fireDataChangedEvents(event.getPath(), listeners, notificationTime, isPathExist);
       }
     }
   }
 
+  private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry> listeners) {
```
narendly: Could you add some javadoc on how this is different from the other method?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1261,31 +1258,62 @@ private void processDataOrChildChange(WatchedEvent event, long notificationTime)
...
+      };
+
+      _eventThread.send(zkEvent);
+    }
+  }
+
   private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry> listeners,
-      final OptionalLong notificationTime) {
+      final long notificationTime, final boolean isPathExist) {
+    final ZkPathStatRecord pathStatRecord = new ZkPathStatRecord(path);
```
narendly: Question: main Helix paths (like /IDEALSTATES, /CONFIG, etc) do not get deleted, so I am wondering if this is really meaningful?

I think handling deleted/non-existent ZNodes might be more effective for children ZNodes? Like not the main IDEALSTATES path, but the IdealState ZNodes inside it.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: done

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: The truth is: currently both *exists* listener and *child* listener are listening on the path (e.g /IDEALSTATES/db32). When the db32 path is removed, both types of listeners will get triggered and they should be (cannot avoid it) unless doing additional checks on the path (check if it's Helix main path or not).

The child change handler needs to handle the child node change. Then a re-installation of a child handler needs to happen.

The data change handler also gets triggered to handle the data change. In the deletion case, because of the existing code, we'll add the unwanted watcher again onto the deleted path. With the fix, it will avoid adding the unwanted watcher and the logic applies to every path in Znode

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -145,32 +145,28 @@ public int hashCode() {
...
+      if (isPathChecked()) {
+        return;
+      }
+      _isChecked.set(true);
+      if (_monitor != null && stat != null) {
         long updateTime = Math.max(stat.getCtime(), stat.getMtime());
-        if (notificationTime.getAsLong() > updateTime) {
-          _monitor.recordDataPropagationLatency(_path, notificationTime.getAsLong() - updateTime);
+        if (notificationTime > updateTime) {
+          _monitor.recordDataPropagationLatency(_path, notificationTime - updateTime);
```
dasahcc: I did not see the value of changing this. If you feel this is necessary, please make it for separate PR as it does impact logic and purely code refactoring.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1261,31 +1257,75 @@ private void processDataOrChildChange(WatchedEvent event, long notificationTime)
...
-        fireDataChangedEvents(event.getPath(), listeners, OptionalLong.of(notificationTime));
+        fireDataChangedEvents(event.getPath(), listeners, notificationTime, pathExists);
+      }
+    }
+  }
+
+  /**
+   * The method is called when state change occurs and we need to handle existing path data handlers
+   */
+  private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry> listeners) {
```
dasahcc: I do see some duplicated code between these three methods. Can you combine the logic to optimize the code structure.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -145,32 +145,28 @@ public int hashCode() {
 
   private class ZkPathStatRecord {
     private final String _path;
-    private Stat _stat = null;
-    private boolean _checked = false;
+    private AtomicBoolean _isChecked = new AtomicBoolean(false);
```
jiajunwang: Can you please explain why we need to change this value to Atomic?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1261,31 +1257,75 @@ private void processDataOrChildChange(WatchedEvent event, long notificationTime)
 
     if (event.getType() == EventType.NodeDataChanged || event.getType() == EventType.NodeDeleted
         || event.getType() == EventType.NodeCreated) {
+      boolean pathExists = event.getType() != EventType.NodeDeleted;
```
jiajunwang: This might be a potential bug. For example,
T1: node deleted, and NodeDeleted is sent.
T2: node is re-created
T3: The NodeDeleted event is processed, pathExists is set to false. And it is **NOT** the truth!

This might be OK, since we will have another event coming later. But it may introduce a substantial change of the ZkClient behavior. So I would prefer if we don't do it. The minimum fix itself does not require this enhancement. Please correct me if I am wrong.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1261,31 +1257,75 @@ private void processDataOrChildChange(WatchedEvent event, long notificationTime)
...
+    try {
+      for (final IZkDataListenerEntry listener : listeners) {
+        ZkEvent zkEvent = new ZkEvent(
+            "Data of " + path + " changed sent to " + listener.getDataListener()
+                + " prefetch data: " + listener.isPrefetchData()) {
+          @Override
+          public void run()
+              throws Exception {
+            Object data;
+            // TODO: are we fetching the data multiple times?
```
jiajunwang: The answer is yes. There is a solution, can we do the same as the other methods so as to avoid multiple reads?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: It was my simple try to make it thread-safe, the atomic has the ability though; Do you have concrete suggestions on how to use it?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: It's an interesting case though I don't think it will introduce any new substantial change to the behavior. Given ZK's consistency, the node recreation event will guarantee to arrive later than the node deletion event. Let me explain fundamentally what happens on `handleNodeDeleted`, it unsubscribes the listener entry in zkClient (it can be confirmed that all operations are all in-memory operations without any ZK relations); 
When the T2 node re-creation event gets processed, it will re-subscribe the listener just like it has never seen it before.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: I will make it a formal TODO and let's first focus on the watcher leakage issue :)

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
pkuwm: But is it really necessary to use atomic? Is there a clue that there would be concurrency issue?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
pkuwm: I think you don't have to check pathExits here and pass it to fireDataChangedEvents(). Instead, you can check the condition within `fireDataChangedEvents`. The pro is the api can avoid the boolean param and 2 logic branches in `fireDataChangedEvents`. If it is a DELETED event, you will call `handleDataDeleted`.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: @pkuwm I agree it's not necessary (my understanding of this code currently involves no multi-threading; To make it thread-safe with a simple atomic boolean is merely a simple try. 
@jiajunwang @pkuwm The latest updates involve little changes on the existing code, it's no longer needed plus I don't really want to improve (refactor) the existing ones :)

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: No longer valid comment since the new change doesn't modify the existing codes that much

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: The latest changes will not need to touch the existing codes too much, so it's no longer valid. Could you take another look, thanks?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1310,26 +1318,30 @@ public void run() throws Exception {
...
           @Override
           public void run() throws Exception {
-            // Reinstall watch before listener callbacks to check the znode status
-            if (!pathStatRecord.pathChecked()) {
-              pathStatRecord.recordPathStat(getStat(path, hasListeners(path)),
-                  OptionalLong.empty());
-            }
-            List<String> children = null;
-            if (pathStatRecord.pathExists()) {
+            List<String> children = new ArrayList<>();
```
pkuwm: Do you change this list initialization from null to an instance for the later for loop? But do you expect the listener will be removed in any case in `handleChildChange`? From the code in `handleChildChange`, if `children` is not null, the listener will not be removed.

i3wangyi: Thanks for mentioning it! You're talking about CallbackHandler's implementation, right? I just checked you're right, that CallbackHandler does rely on the NULL to remove the listener.
To me, it's very odd cause it obviously violates the decoupling principle while 2 different classes are so tightly correlated, also it doesn't make sense to pass a NULL variable. 

I will put a TODO and leave the current code as it is. But still, need to add the child exists watcher part.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -904,13 +904,6 @@ public void process(WatchedEvent event) {
         || event.getType() == Event.EventType.NodeCreated
         || event.getType() == Event.EventType.NodeChildrenChanged;
 
-    if (event.getType() == Event.EventType.NodeDeleted) {
```
jiajunwang: Do we have a new place to debug log delete event? I didn't see it in this PR. Please confirm. If we don't have, let's keep it here.

jiajunwang: Feel free to enrich the log message though.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1309,22 +1306,24 @@ public void run() throws Exception {
...
-  private void fireChildChangedEvents(final String path, Set<IZkChildListener> childListeners) {
+  private void fireChildChangedEvents(final String path, Set<IZkChildListener> childListeners, boolean pathExists) {
     try {
-      final ZkPathStatRecord pathStatRecord = new ZkPathStatRecord(path);
       for (final IZkChildListener listener : childListeners) {
         _eventThread.send(new ZkEvent("Children of " + path + " changed sent to " + listener) {
           @Override
           public void run() throws Exception {
-            // Reinstall watch before listener callbacks to check the znode status
-            if (!pathStatRecord.pathChecked()) {
```
jiajunwang: This will remove our previous improvement for multiple listeners. Why we remove it?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1309,22 +1306,24 @@ public void run() throws Exception {
...
-            if (pathStatRecord.pathExists()) {
+            if (pathExists) {
               try {
+                //TODO: duplicate reads when multiple child listener exists
                 children = getChildren(path);
+                if (children != null) {
+                  for (String child : children) {
+                    // add the exists watcher for all child path, it's to prevent watcher missing
+                    // in case of node recreation shortly after deletion
+                    watchForData(path + "/" + child);
```
jiajunwang: This is not necessary, it has been done in the "listener.handleChildChange(path, children);" later.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1278,29 +1276,28 @@ private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry>
             + listener.getDataListener() + " prefetch data: " + listener.isPrefetchData()) {
           @Override
           public void run() throws Exception {
-            // Reinstall watch before listener callbacks to check the znode status
+            if (!pathExists) {
+              listener.getDataListener().handleDataDeleted(path);
+              return;
+            }
             if (!pathStatRecord.pathChecked()) {
               pathStatRecord.recordPathStat(getStat(path, true), notificationTime);
```
jiajunwang: You avoided the notificationTime report on the deleting events with this change. Please keep the original order.

jiajunwang: This will invalidate our propagation latency metric, BTW.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
jiajunwang: You can just change the condition from hasListeners(path) to pathExists && hasListeners(path), right?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: updated. The removal is not intended

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: I checked the code in `recordPathStat`, basically, the line `pathStatRecord.recordPathStat(getStat(path, hasListeners(path)), OptionalLong.empty());` does nothing! This is because notificationTime is optional and the propogation latency is never recorded whne child change events. 

I feel like leaving this piece of code here will just make ourselves confused. And if in the future, we do need to record propagation latency for child events. We could add it back. 

Second, `getStat(path, hasListeners(path))` will add a exists watcher for the parent path (child listener). Don't we want to add watcher type of `child`? The code here looks problematic for me, please clarify the original purpose if I misunderstand it. Thanks!

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: That's true... However, I do feel it's odd to connect the relationship between two separate classes as it violates the separation of concern. I could either delete the added part or just add the change but remove the code path in CallbackListener. Which one do you prefer? 
To make things easier, I will mark it as a TODO (IMO, Re-install watchers should be the concern of ZkClient(Watcher) class, not CallbackHandler).

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: Thanks for catching it, fixed it.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1278,29 +1279,28 @@ private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry>
...
           @Override
           public void run() throws Exception {
-            // Reinstall watch before listener callbacks to check the znode status
             if (!pathStatRecord.pathChecked()) {
-              pathStatRecord.recordPathStat(getStat(path, true), notificationTime);
+              pathStatRecord.recordPathStat(getStat(path, pathExists), notificationTime);
             }
-            if (!pathStatRecord.pathExists()) {
-              // no znode found at the path, trigger data deleted handler.
+            if (!pathExists) {
```
kaisun2000: Seems to me all the change from line 1285 is not necessary, right? We can still use the old code. 

The change may not be wrong, but seems to necessary. 
Or did I miss something here?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
kaisun2000: Seems to me that we can keep the same style as fireDataChangeEvent

`
if (!pathStatRecord.pathChecked()) {
   pathStatRecord.recordPathStat(getStat(path, pathExists), OptionalLong.empty());
}

// The rest leave it the same 
`

i3wangyi: @kaisun2000 Yes. However, as I replied earlier, this line `pathStatRecord.recordPathStat(getStat(path, pathExists), OptionalLong.empty());` does nothing (need to read the exact implemenation) when `OptionalLong.empty()` is passed in. We're not recording propogation latency for child listeners anyway. It only confuses us AND the `getStat()` method will possibly add another leaked watch

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: Yes, we can keep the old code. I just changed the 
`if { } else {}` to `if () return; xxx` to make the code less indent. No function change

kaisun2000: How about let us keep the old code. The thing is that later if people looking at the change, it would take some time to figure out there is not really logical change here. 

This is more of a beautification work. This may look better, but for fix, can be confusing.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
kaisun2000: I see. Is it because `if (_monitor != null && stat != null && **notificationTime.isPresent())**` in the pathStatRecord.recordPathState() implementation?

Then my question is that before this diff, line 1321 `pathStatRecord.recordPathStat(getStat(path, hasListeners(path)), OptionalLong.empty());` still does not really emit metrics either. So my proposal keeps the old behavior? Or did I miss something here?

i3wangyi: @kaisun2000 Yes, keep the same code is fine; Is it really necessary? Why not add it later once we do need to record propagation latency?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: Sure, reverted the style

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
kaisun2000: This is actually debatable. I am more into thinking that " the fix" just fix the problem, don't introduce non essential change. Or change the metric reporting behavior. One fix for one issue.

That said, I don't have strong opinion here. Check with JJ. I am fine either way.

i3wangyi: all right, I will revert this part of change

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -960,10 +956,10 @@ public void process(WatchedEvent event) {
 
   private void fireAllEvents() {
     for (Entry<String, Set<IZkChildListener>> entry : _childListener.entrySet()) {
-      fireChildChangedEvents(entry.getKey(), entry.getValue());
+      fireChildChangedEvents(entry.getKey(), entry.getValue(), true);
     }
     for (Entry<String, Set<IZkDataListenerEntry>> entry : _dataListener.entrySet()) {
-      fireDataChangedEvents(entry.getKey(), entry.getValue(), OptionalLong.empty());
+      fireDataChangedEvents(entry.getKey(), entry.getValue());
```
kaisun2000: Another question, this is something I am not sure. No true or false passed in here as 3rd parameter in line 962.  THe function signature has not default value. What value would be then for this invocation of fireDataChangeEvents?

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1309,21 +1309,21 @@ public void run() throws Exception {
...
           public void run() throws Exception {
-            // Reinstall watch before listener callbacks to check the znode status
             if (!pathStatRecord.pathChecked()) {
-              pathStatRecord.recordPathStat(getStat(path, hasListeners(path)),
+              pathStatRecord.recordPathStat(getStat(path, hasListeners(path) && !pathExists),
                   OptionalLong.empty());
             }
             List<String> children = null;
-            if (pathStatRecord.pathExists()) {
+            if (pathExists) {
```
kaisun2000: nit: if (pathStatRecord.pathExists()) would also work here.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1278,12 +1279,10 @@ private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry>
...
           @Override
           public void run() throws Exception {
-            // Reinstall watch before listener callbacks to check the znode status
             if (!pathStatRecord.pathChecked()) {
-              pathStatRecord.recordPathStat(getStat(path, true), notificationTime);
+              pathStatRecord.recordPathStat(getStat(path, pathExists), notificationTime);
             }
-            if (!pathStatRecord.pathExists()) {
-              // no znode found at the path, trigger data deleted handler.
+            if (!pathExists) {
```
kaisun2000: nit: if (!pathStatRecord.pathExists()) also works here.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: This function will be invoked when session expiry. I guess it will be used to initialize the states; Thus it makes sense to assume the path exists

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: updated

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: updated

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
kaisun2000: I see. By default it is true.

i3wangyi: Actually I could remove the overloading method by passing the default value here

kaisun2000:
Look good to me. Don't forget to run the unit test one more time.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -1278,12 +1275,10 @@ private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry>
             + listener.getDataListener() + " prefetch data: " + listener.isPrefetchData()) {
           @Override
           public void run() throws Exception {
-            // Reinstall watch before listener callbacks to check the znode status
```
dasahcc: Can you add a comment like this? We reinstall watch based on the node has been deleted or not.

i3wangyi: updated!

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
```diff
@@ -960,10 +956,10 @@ public void process(WatchedEvent event) {
 
   private void fireAllEvents() {
     for (Entry<String, Set<IZkChildListener>> entry : _childListener.entrySet()) {
-      fireChildChangedEvents(entry.getKey(), entry.getValue());
+      fireChildChangedEvents(entry.getKey(), entry.getValue(), true);
```
jiajunwang: So while the client is reconnected, it will just reinstall the watcher without checking, right? Will we still see watcher leakage?

helix-core/src/test/java/org/apache/helix/manager/zk/client/TestHelixZkClient.java
```diff
@@ -163,6 +163,8 @@ public void handleDataDeleted(String s) {
       // expected to be here.
     }
 
+    // client B needs to re-install the data watch
+    sharedZkClientB.watchForData(TEST_PATH);
```
jiajunwang: Usually, I'm very cautious about changing the existing test case. This indicates that we changed the behavior. So after this change, the user will need to resubscribe the watch once it is deleted, right? Regardless of the Helix reserved node that has children change listeners to cover, how about the other zk path that we allow the user to subscribe listener? The listeners will be idle after the node has been removed once, right? This breaks the feature IMO.

Overall, I think we will need to make it optional. 1. It is fine that we change the behavior for the reserved patch with additional children change listener on the parent node. But for the other random path from user's requests, we should keep the old behavior.

i3wangyi: Your concern is valid! To think about it, actually, it is the reason why native zookeeper has 2 sets of path related watcher: exists type and data type. To make a distinction between two, IMO, we either create a new `subscribeXXXDataChange` method or add a parameter like `subscribeDataChange(xx, xx, boolean listenOnNonExistPath = True/False`. 

What's your opinion? In fact, at the current stage,  do we have users listen on random paths? If not, I think we're creating problems for ourselves.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: Good question. I think with the default value, it will reinstall the watcher whatever left on the previous session; It's likely the node deletion happens after session expiry, in this case, it will leak; One mitigation approach is to perform EXISTS check when handling new session, but it's always possible a race-condition exists (that the node deletion happen after the exists check).  One good thing though comes with new session is that all the "leaked" watchers will be gone.

Do you think a set of exists check on each path from original session should be performed or it's OK to trade-off between performance and excessive watchers.

jiajunwang: 1. I still think the fix should be bumping up ZK version. If the concern is not super urgent, we should do that. Clean and no side effects.
2. Otherwise, given this issue, a background GC thread would be easier. What I'm thinking is that we have a running timeout for the existing check-in that thread for each watched path.

- If delete event, timeout set to zero and trigger GC immediately.
- Otherwise, the GC thread will validate the path after the timeout.

helix-core/src/test/java/org/apache/helix/manager/zk/client/TestHelixZkClient.java
jiajunwang: One way is keeping the existing subscription methods the same. But we created some private/protected methods for the HelixManager to use.
Alternatively, as you said, adding options to the method. It is not optimal but compared with adding more methods, I would prefer to keep it simple now. One improvement is that we extend the listeners so the one keeps existing after delete would be a different listener type internally. Then the callback knows what to do. It helps to keep the methods simple.

helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java
i3wangyi: 1. I don't think to bump up the ZK version and use the new API to delete ZK watcher explicitly would resolve the problem. 

Let me explain why: if the node is re-created after deleting the watcher request is submitted, without the parent path's child listener, there's no way the ZkClient would be able to receive new events from the path. It's the same dilemma this fix is currently facing.

2. A background GC thread is an over-kill solution. First It's difficult to determine the right TTL; Second, it comes with the additional thread cost.

My argument point is that it's not always valid to say we need to be 100% stick to the old behavior, especially when the old behavior is incorrect and leading to the excessive amount of leakage.
The easiest way is to release a newer Helix version and educate the client with the caveat **"Once a node is deleted, the listener on the path will be inactive even the node is re-created; The workaround is to have a child listener on the parent path or re-subscribe the path"**. The statement IMO makes sense to most clients. 

As to how many clients will be suffering from the behavior change, I don't have a good answer but the experience told me the number is minor if the client is using Helix as a cluster management tool instead of a generic ZKClient proxy. 

Do these all make sense? @kaisun2000 @dasahcc please feel free to leave any thoughts about this, cause it will change the design fundamentally and won't be able to help Venice reduce the watcher pains in any near future.

jiajunwang: The first point is for us to simplify the GC. I agree it won't help with the listener becoming inactive case.
And, about the second part, a behavior change is fine as long as it is reasonable. However, the main concern is that, even with the current design, leakage is still possible, right? Since the watcher removing depends on the notification, if no notification arrived, or the client missed, the watcher will be leaked. That's why I think this design is not good enough. Feel free to fix the current design if there is a way to avoid all leakages.

kaisun2000: If I understand correctly, @jiajunwang concern is that reconnect happen and fireAllEvent would be triggered and install the watcher for all the paths in subscriber list. The thing is that there can be case that in the disconnecting period, the zkpath can be deleted. This can still leak another path. This is very good point. 

The first propose fix by @jiajunwang is:
1/ use exists to re-subscribe in re-connecting phase anyway.
2/ if path is not there, delete the watcher for this path.
This require 3.5 zk feature though. Not feasible now.

For the GC approach, what I can see is that this may still be a problem. First you check if the subscribe path is there. If it is not there, you delete it. The problem is that these are two steps operation to ZK. The problem is between the two steps, the user may create the path and install watcher as one operation. Now you delete it. This is even worse than leaking. 

The second step of the first proposal has the same problem. 

Actually I thought about this before, if Zookeeper does not have atomic check if the path not there then delete watcher, it is hard for us to implement this correctly.
@jiajunwang, what is your take?

Maybe let us note this issue and have an enhancement task to track it?

i3wangyi: @jiajunwang I agree that during re-connecting, it's still possible to leak the watcher if the in-memory listeners contain already deleted paths; I will think about the case and see if it can be overcome. But we all know at least the case here is not the major leakage point.

In fact, the long comment is more about the other one you claimed about changing the existing behavior cause I have to modify the integration test case.

i3wangyi: Discussed offline and we reached the consensus that the leakage during session expiry is a potential issue and since the impact of it is minor, we'll create a ticket to track the fixing progress in the future

helix-core/src/test/java/org/apache/helix/manager/zk/client/TestHelixZkClient.java
i3wangyi: Discussed offline and we agreed that the behavior change is necessary and the client needs to be aware of the change.

i3wangyi:
The PR is ready to be merged into master, approved by @jiajunwang @kaisun2000 

Commit message:
```
 Fix the zk watcher leakage issue when znode is deleted
    - Leverage the EventType (NodeDeleted) to verify if the path is deleted
    - If it's deleted, no need to re-install the watcher
```

Issue #526, opened 2019-10-23
Need a better way to verify (wait) if the partition assignment converged or not.


jiajunwang:
Currently, we use the verifier to check if the partitions in a cluster have been allocated to the ideal placement. The verifier compares the current states vs. the IdealState mapping. This method itself is fine, but we don't have a good way to know when can the verifier starts to wait.
For example, if the cluster change just happens, and the verifier starts waiting before the rebalance calculation is done, the result will be returned prematurely.

So, a better way to tell when controller has finished calculation and then kick off the verifier's check logic is required. This will help us eliminate the hardcoded thread sleeps in the tests, and stabilize the tests.

narendly:
+1. Thanks for creating an issue for this.

pkuwm:
Note down an idea: verifier listens on a Znode, when controller completes calculation, it sends a message or updates the Znode so verifier knows controllers completes calculation.

@jiajunwang's concern: 
About why it's still not a complete idea, consider the following scenario:
event A processed in controller
test case trigger some changes
controller notify for event A processed
test case proceed prematurely (event B has not been processed yet)

Maybe we can create one Znode for each event, and verifier needs to check if the message/znode data is what it's looking for to proceed?

Issue #660, opened 2019-12-17
Controller leader switch may trigger some additional (unexpected) rebalance in the WAGED rebalancer


jiajunwang:
1. Since the WAGED rebalancer determines the rebalance scope by comparing in-memory cluster data cache, it is possible that a controller leader switch impacts this scope calculating. Then the scope could be enlarged or shrank.
2. In addition, if the leader switch happened while a rebalance pipeline is proceeding, we might see 2 versions output of the best possible state from the different controller (need to confirm if the pipeline would continue once the leadership switch happens. If that's the case, some partitions might be moved back and forth.

To solve this,
1. Considering reset the data change detector on new session handling.
2. We need to confirm that a rebalance pipeline will be stopped once the leadership of the controller has been switched.

jiajunwang:
As an initial investigation, this case will only impact the baseline calculation. Since the partial rebalance won't take the change detector state as an input. Then since the baseline is not directly used to generate the IdealStates, it might be OK that even the baseline is changed slightly after leadership switch. We would keep this issue open and test this case.

i3wangyi:
Just see this ticket, actually, it can be a pretty big change. Not sure about #1 that we need to modify the new session handling for waged rebalancer, if there's another option existence, I will be against the idea of handling session to fit one specific rebalance algorithm;

#2, cannot we just let the pipeline continue to finish BUT when it tries to write to ZK, you could check if the leader is still the leader of the cluster and let it fail if not

jiajunwang:
@i3wangyi I agree with you. Let's figure out a generic way instead of just fix it for the WAGED rebalancer. Then only concern not is the baseline is potentially changed unexpectedly on leadership switch. I'm not sure about the impact. Do you think we can stress test it?

i3wangyi:
To simulate mastership hand-off in CHO can be pretty tricky, I would say it's possible but requires some clever designs (maybe manually create messages to the controller_participants).

jiajunwang:
What's in my mind, find out the super cluster resource for the tested cluster. Then disable this resource partition in the current leader. And the leadership will be moved to another controller. Then we can remove/adding that disabled partition from the resource config again and again. This will trigger frequent leadership switch.

jiajunwang:
Just to record an idea from @narendly.
1. reset the change detector on leadership acquired.
2. if the change detector has no previous status recorded, trigger partial rebalance only. Then schedule a full baseline calculation after to catch up if any newly added resources are not assigned.

[...]

PR #690, opened 2020-01-18
Reset the WAGED rebalancer once the controller newly acquires leadership.

jiajunwang:
### Issues

#660 

### Description

This is to prevent any cached assignment information which is recorded during the previous session from impacting the rebalance result.
Detailed change list:
1. Move the stateful WAGED rebalancer to the GenericHelixController object instead of the rebalance stage. This is for resolving the possible race condition between the event processing thread and leader switch handling thread.
2. Adding a new test regarding leadership switch to verify that the WAGED rebalancer has been reset after the processing.

### Tests

TestWagedRebalance.testRebalancerReset

[ERROR] Failures: 
[ERROR]   TestZKUtil.testChildrenOperations:80 expected:<2> but was:<4>
[INFO] 
[ERROR] Tests run: 1067, Failures: 2, Errors: 0, Skipped: 0

Re-run

[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 14.357 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -191,6 +191,7 @@ public class GenericHelixController implements IdealStateChangeListener,
   // Since the stateful rebalancer needs to be lazily constructed when the HelixManager instance is
   // ready, the GenericHelixController is not constructed with a stateful rebalancer. This wrapper
   // is to avoid the complexity of handling a nullable value in the event handling process.
+  // TODO Create the required stateful rebalancer only when it is used by any resource.
   private final StatefulRebalancerRef _rebalancerRef = new StatefulRebalancerRef() {
     @Override
     protected StatefulRebalancer createRebalancer(HelixManager helixManager) {
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -1319,6 +1319,8 @@ public class GenericHelixController implements IdealStateChangeListener,
     /**
      * @return A valid rebalancer object.
      *         If the rebalancer is no longer valid, it will be reset before returning.
+     * TODO: Make rebalancer volatile or make it singleton, if this method is called in multiple
+     * TODO: threads outside the controller object.
      */
     synchronized T getRebalancer(HelixManager helixManager) {
       // Lazily initialize the stateful rebalancer instance since the GenericHelixController
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -62,6 +62,7 @@ import org.apache.helix.controller.dataproviders.WorkflowControllerDataProvider;
 import org.apache.helix.controller.pipeline.AsyncWorkerType;
 import org.apache.helix.controller.pipeline.Pipeline;
 import org.apache.helix.controller.pipeline.PipelineRegistry;
+import org.apache.helix.controller.rebalancer.StatefulRebalancer;
 import org.apache.helix.controller.rebalancer.waged.WagedRebalancer;
 import org.apache.helix.controller.stages.AttributeName;
 import org.apache.helix.controller.stages.BestPossibleStateCalcStage;
@@ -187,10 +188,15 @@ public class GenericHelixController implements IdealStateChangeListener,
 
   private HelixManager _helixManager;
 
-  // Since the WAGED rebalancer needs to be lazily constructed, the GenericHelixController will not
-  // be constructed with a WAGED rebalancer. This wrapper is to avoid the complexity of handling a
-  // nullable value in the GenericHelixController logic.
-  private final WagedRebalancerRef _wagedRebalancerRef = new WagedRebalancerRef();
+  // Since the stateful rebalancer needs to be lazily constructed when the HelixManager instance is
+  // ready, the GenericHelixController is not constructed with a stateful rebalancer. This wrapper
+  // is to avoid the complexity of handling a nullable value in the event handling process.
+  private final StatefulRebalancerRef _rebalancerRef = new StatefulRebalancerRef() {
+    @Override
+    protected StatefulRebalancer createRebalancer(HelixManager helixManager) {
+      return new WagedRebalancer(helixManager);
+    }
+  };
 
   /**
    * TODO: We should get rid of this once we move to:
@@ -645,7 +651,7 @@ public class GenericHelixController implements IdealStateChangeListener,
     // the pipeline will be triggered. So the reset rebalancer won't be used before the controller
     // regains leadership.
     event.addAttribute(AttributeName.STATEFUL_REBALANCER.name(),
-        _wagedRebalancerRef.getWagedRebalancer(manager));
+        _rebalancerRef.getRebalancer(manager));
 
     if (!manager.isLeader()) {
       logger.error("Cluster manager: " + manager.getInstanceName() + " is not leader for " + manager
@@ -1048,7 +1054,7 @@ public class GenericHelixController implements IdealStateChangeListener,
       // mark the rebalancer invalid only, instead of closing it here.
       // This to-be-closed WAGED rebalancer will be reset later on a later event processing if
       // the controller becomes leader again.
-      _wagedRebalancerRef.invalidRebalancer();
+      _rebalancerRef.invalidateRebalancer();
     }
 
     logger.info("END: GenericClusterController.onControllerChange() for cluster " + _clusterName);
@@ -1152,7 +1158,7 @@ public class GenericHelixController implements IdealStateChangeListener,
 
     enableClusterStatusMonitor(false);
 
-    _wagedRebalancerRef.closeWagedRebalancer();
+    _rebalancerRef.closeRebalancer();
 
     // TODO controller shouldn't be used in anyway after shutdown.
     // Need to record shutdown and throw Exception if the controller is used again.
@@ -1286,55 +1292,57 @@ public class GenericHelixController implements IdealStateChangeListener,
     eventThread.setDaemon(true);
     eventThread.start();
   }
-}
 
-/**
- * A wrapper class for the WAGED rebalancer instance.
- */
-class WagedRebalancerRef {
-  private WagedRebalancer _rebalancer = null;
-  private boolean _isRebalancerValid = true;
+  /**
+   * A wrapper class for the stateful rebalancer instance that will be tracked in the
+   * GenericHelixController.
+   */
+  private abstract class StatefulRebalancerRef<T extends StatefulRebalancer> {
+    private T _rebalancer = null;
+    private boolean _isRebalancerValid = true;
 
-  private void createWagedRebalancer(HelixManager helixManager) {
-    // Create WagedRebalancer instance if it hasn't been already initialized
-    if (_rebalancer == null) {
-      _rebalancer = new WagedRebalancer(helixManager);
-      _isRebalancerValid = true;
+    /**
+     * @param helixManager
+     * @return A new stateful rebalancer instance with initial state.
+     */
+    protected abstract T createRebalancer(HelixManager helixManager);
+
+    /**
+     * Mark the current rebalancer object to be invalid, which indicates it needs to be reset before
+     * the next usage.
+     */
+    synchronized void invalidateRebalancer() {
+      _isRebalancerValid = false;
     }
-  }
 
-  /**
-   * Mark the current rebalancer object to be invalid, which indicates it needs to be reset before
-   * the next usage.
-   */
-  synchronized void invalidRebalancer() {
-    _isRebalancerValid = false;
-  }
-
-  /**
-   * @return A valid rebalancer object.
-   *         If the rebalancer is no longer valid, it will be reset before returning.
-   */
-  synchronized WagedRebalancer getWagedRebalancer(HelixManager helixManager) {
-    // Lazily initialize the WAGED rebalancer instance since the GenericHelixController instance is
-    // instantiated without the HelixManager information that is required.
-    createWagedRebalancer(helixManager);
-    // If the rebalance exists but has been marked as invalid (due to leadership switch), it needs
-    // to be reset before return.
-    if (!_isRebalancerValid) {
-      _rebalancer.reset();
-      _isRebalancerValid = true;
+    /**
+     * @return A valid rebalancer object.
+     *         If the rebalancer is no longer valid, it will be reset before returning.
+     */
+    synchronized T getRebalancer(HelixManager helixManager) {
+      // Lazily initialize the stateful rebalancer instance since the GenericHelixController
+      // instance is instantiated without the HelixManager information that is required.
+      if (_rebalancer == null) {
+        _rebalancer = createRebalancer(helixManager);
+        _isRebalancerValid = true;
+      }
+      // If the rebalance exists but has been marked as invalid (due to leadership switch), it needs
+      // to be reset before return.
+      if (!_isRebalancerValid) {
+        _rebalancer.reset();
+        _isRebalancerValid = true;
+      }
+      return _rebalancer;
     }
-    return _rebalancer;
-  }
 
-  /**
-   * Proactively close the rebalance object to release the resources.
-   */
-  synchronized void closeWagedRebalancer() {
-    if (_rebalancer != null) {
-      _rebalancer.close();
-      _rebalancer = null;
+    /**
+     * Proactively close the rebalance object to release the resources.
+     */
+    synchronized void closeRebalancer() {
+      if (_rebalancer != null) {
+        _rebalancer.close();
+        _rebalancer = null;
+      }
     }
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -62,6 +62,7 @@ import org.apache.helix.controller.dataproviders.WorkflowControllerDataProvider;
 import org.apache.helix.controller.pipeline.AsyncWorkerType;
 import org.apache.helix.controller.pipeline.Pipeline;
 import org.apache.helix.controller.pipeline.PipelineRegistry;
+import org.apache.helix.controller.rebalancer.waged.WagedRebalancer;
 import org.apache.helix.controller.stages.AttributeName;
 import org.apache.helix.controller.stages.BestPossibleStateCalcStage;
 import org.apache.helix.controller.stages.ClusterEvent;
@@ -169,7 +170,6 @@ public class GenericHelixController implements IdealStateChangeListener,
   Timer _onDemandRebalanceTimer = null;
   AtomicReference<RebalanceTask> _nextRebalanceTask = new AtomicReference<>();
 
-
   /**
    * A cache maintained across pipelines
    */
@@ -186,6 +186,11 @@ public class GenericHelixController implements IdealStateChangeListener,
   private final Set<Pipeline.Type> _enabledPipelineTypes;
 
   private HelixManager _helixManager;
+
+  // Since the WAGED rebalancer needs to be lazily constructed, the GenericHelixController will not
+  // be constructed with a WAGED rebalancer. This wrapper is to avoid the complexity of handling a
+  // nullable value in the GenericHelixController logic.
+  private final WagedRebalancerRef _wagedRebalancerRef = new WagedRebalancerRef();
 
   /**
    * TODO: We should get rid of this once we move to:
@@ -626,6 +631,22 @@ public class GenericHelixController implements IdealStateChangeListener,
       return;
     }
 
+    // Event handling happens in a different thread from the onControllerChange processing thread.
+    // Thus, there are several possible conditions.
+    // 1. Event handled after leadership acquired. So we will have a valid rebalancer for the
+    // event processing.
+    // 2. Event handled shortly after leadership relinquished. And the rebalancer has not been
+    // marked as invalid yet. So the event will be processed the same as case one.
+    // 3. Event is leftover from the previous session, and it is handled when the controller
+    // regains the leadership. The rebalancer will be reset before being used. That is the
+    // expected behavior so as to avoid inconsistent rebalance result.
+    // 4. Event handled shortly after leadership relinquished. And the rebalancer has been marked
+    // as invalid. So we reset the rebalancer. But the later isLeader() check will return false and
+    // the pipeline will be triggered. So the reset rebalancer won't be used before the controller
+    // regains leadership.
+    event.addAttribute(AttributeName.STATEFUL_REBALANCER.name(),
+        _wagedRebalancerRef.getWagedRebalancer(manager));
+
     if (!manager.isLeader()) {
       logger.error("Cluster manager: " + manager.getInstanceName() + " is not leader for " + manager
           .getClusterName() + ". Pipeline will not be invoked");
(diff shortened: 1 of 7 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
```diff
@@ -38,5 +38,6 @@
   AsyncFIFOWorkerPool,
   PipelineType,
   LastRebalanceFinishTimeStamp,
-  ControllerDataProvider
+  ControllerDataProvider,
+  STATEFUL_REBALANCER
```
jiajunwang: Please comment if anyone of you has a better name. This one is not good.

narendly:
@jiajunwang Could we separate the scope for this change?

1. Is it possible to create a PR that includes the change on resetting the cluster data change detector?
2. The next PR would be the movement of the rebalancer to Generic Helix Controller. It's not too clear why this is necessary 

> Move the stateful WAGED rebalancer to the GenericHelixController object instead of the rebalance stage. This is for resolving the possible race condition between the event processing thread and leader switch handling thread.

Could you explain this further?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -626,6 +631,22 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
       return;
     }
 
+    // Event handling happens in a different thread from the onControllerChange processing thread.
+    // Thus, there are several possible conditions.
+    // 1. Event handled after leadership acquired. So we will have a valid rebalancer for the
+    // event processing.
+    // 2. Event handled shortly after leadership relinquished. And the rebalancer has not been
```
narendly: Do you mean that the old controller will continue to process events even after having lost the leadership using the in-memory rebalancer/cache state?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -626,6 +631,22 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
...
+    // marked as invalid yet. So the event will be processed the same as case one.
+    // 3. Event is leftover from the previous session, and it is handled when the controller
```
narendly: Shouldn't we just discard all events from stale sessions?

Resetting the cache will trigger a fresh, up-to-date global baseline calculation anyways?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -626,6 +631,22 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
...
+    // regains the leadership. The rebalancer will be reset before being used. That is the
+    // expected behavior so as to avoid inconsistent rebalance result.
+    // 4. Event handled shortly after leadership relinquished. And the rebalancer has been marked
+    // as invalid. So we reset the rebalancer. But the later isLeader() check will return false and
```
narendly: Could you please explain why another pipeline is being triggered when the controller is no longer a leader?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: The one that is being processed will continue.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: We should, but there is no good way to do it for now. And even we discard the queued events, the concerning one is actually the one that is being processed. Force canceling that could be risky. So I don't think to do it without deeply improve the pipeline logic is a good idea.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: 1. in event thread, check if is leader, return true, so event thread continues
2. in the ZK thread, session expired, so handling triggered, and leadership switch is done.
3. in the event thread, keep triggering the pipeline.

jiajunwang:
> @jiajunwang Could we separate the scope for this change?
> 
> 1. Is it possible to create a PR that includes the change on resetting the cluster data change detector?
> 2. The next PR would be the movement of the rebalancer to Generic Helix Controller. It's not too clear why this is necessary
> 
> > Move the stateful WAGED rebalancer to the GenericHelixController object instead of the rebalance stage. This is for resolving the possible race condition between the event processing thread and leader switch handling thread.
> 
> Could you explain this further? I am afraid in terms of design, we might be introducing a new dependency (between GenericHelixController and the rebalancer/stages). This doesn't seem too great to me because Helix's design I believe aims to keep them independent.

Separate change is a good idea, will do so. So I will create another PR before this one. Just adding reset methods.
As for the additional dependency that you mentioned,
1. Another option is to keep the assignment metadata store and change detector in the controller. But that actually makes the story more complicated and introduce some skip level dependencies. I tried, this design is not looking nice.
2. Even with this change, GenericHelixController does not depend on the stage.
3. GenericHelixController will be bound with the WAGED rebalancer. Since it is stateful, tracking it in the controller makes more sense than a certain stage as we are doing now. Note that the stages are better to be stateless. But I agree it is not perfect.
Like what we did for the several other stateful objects in the controller instance, I don't have a better option for now. Please let me know if you have a good idea.

jiajunwang:
Split the reset logic to https://github.com/apache/helix/pull/696 as suggested. @narendly Please review that one first.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1259,3 +1287,54 @@ private void initPipeline(Thread eventThread, BaseControllerDataProvider cache)
...
+      _rebalancer = new WagedRebalancer(helixManager);
+      _isRebalancerValid = true;
+    }
+  }
+
+  /**
+   * Mark the current rebalancer object to be invalid, which indicates it needs to be reset before
+   * the next usage.
+   */
+  synchronized void invalidRebalancer() {
```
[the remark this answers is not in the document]
jiajunwang: Sure, I will update after https://github.com/apache/helix/pull/696.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1259,3 +1287,54 @@ private void initPipeline(Thread eventThread, BaseControllerDataProvider cache)
     eventThread.start();
   }
 }
+
+/**
+ * A wrapper class for the WAGED rebalancer instance.
+ */
+class WagedRebalancerRef {
```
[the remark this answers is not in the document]
jiajunwang: I think it's a good idea. Let me try to implement a simple version as a starting point. Will update later.

jiajunwang:
@narendly Thanks for the review, I took a quick look. The comments will be addressed in this PR after we finished https://github.com/apache/helix/pull/696. Please move to there and review that PR first : )

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -187,6 +187,11 @@
 
   private HelixManager _helixManager;
 
+  // Since the WAGED rebalancer needs to be lazily constructed, the GenericHelixController will not
+  // be constructed with a WAGED rebalancer. This wrapper is to avoid the complexity of handling a
+  // nullable value in the GenericHelixController logic.
+  private final WagedRebalancerRef _wagedRebalancerRef = new WagedRebalancerRef();
```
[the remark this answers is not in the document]
jiajunwang: I've tried to make the class more generic. Now it is an abstract class. But the configuration part seems to be overdesign for now. Let's consider that part once we have multiple ones.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1259,3 +1287,54 @@ private void initPipeline(Thread eventThread, BaseControllerDataProvider cache)
...
 }
+
+/**
+ * A wrapper class for the WAGED rebalancer instance.
+ */
+class WagedRebalancerRef {
+  private WagedRebalancer _rebalancer = null;
+  private boolean _isRebalancerValid = true;
+
+  private void createWagedRebalancer(HelixManager helixManager) {
```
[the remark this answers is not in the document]
jiajunwang: That's the first thing I tried, if we don't need the _isRebalancerValid flag, it works fine and we won't need to create StatefulRebalancerRef at all. Basically, in this case, we make the WAGED rebalancer singleton.
However, since we have a flag that is not quite suitable for the WAGED class, it would be easier for me to put both the flag and the rebalancer in a StatefulRebalancerRef object. And the Ref class is for the GenericHelixController only. Given that done, the singleton pattern becomes optional.

Moreover, it is concerning to make the WAGED rebalancer singleton. If we want to use it for both task framework and the resource rebalancer, or we want to manage different resources with different rebalancer configuration, singleton won't work. That's why I didn't pick up that design.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -187,6 +188,16 @@
...
   private HelixManager _helixManager;
 
+  // Since the stateful rebalancer needs to be lazily constructed when the HelixManager instance is
+  // ready, the GenericHelixController is not constructed with a stateful rebalancer. This wrapper
+  // is to avoid the complexity of handling a nullable value in the event handling process.
+  private final StatefulRebalancerRef _rebalancerRef = new StatefulRebalancerRef() {
+    @Override
+    protected StatefulRebalancer createRebalancer(HelixManager helixManager) {
+      return new WagedRebalancer(helixManager);
+    }
```
narendly: Could we add a TODO here for making it configurable (based on ClusterConfig for example) in the future as we have different versions of stateful rebalancers?

Ideally we won't have to import WagedRebalancer explicitly (which seems like we're creating a dependency between **Generic**HelixController and a specific rebalancer class). I'm fine with creating an issue and getting to it later.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1258,4 +1292,57 @@ private void initPipeline(Thread eventThread, BaseControllerDataProvider cache)
...
+   */
+  private abstract class StatefulRebalancerRef<T extends StatefulRebalancer> {
+    private T _rebalancer = null;
+    private boolean _isRebalancerValid = true;
+
+    /**
+     * @param helixManager
+     * @return A new stateful rebalancer instance with initial state.
+     */
+    protected abstract T createRebalancer(HelixManager helixManager);
```
narendly: How do we prevent this from being called multiple times?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1258,4 +1292,57 @@ private void initPipeline(Thread eventThread, BaseControllerDataProvider cache)
     eventThread.setDaemon(true);
     eventThread.start();
   }
+
+  /**
+   * A wrapper class for the stateful rebalancer instance that will be tracked in the
+   * GenericHelixController.
+   */
+  private abstract class StatefulRebalancerRef<T extends StatefulRebalancer> {
```
narendly: 1. Should we think about which methods in this class should be idempotent?

2. Why an abstract class vs an interface?

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
narendly: I think now, with the introduction of a StatefulRebalancer interface, having this name makes sense and pretty clear. What do you think?

Also, could we follow the enum convention? Use `StatefulRebalancer` instead?

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -249,23 +220,33 @@ private boolean validateOfflineInstancesLimit(final ResourceControllerDataProvid
     return true;
   }
 
+  private void updateWagedRebalancer(WagedRebalancer wagedRebalancer, ClusterConfig clusterConfig) {
```
narendly: Should we check things on the WagedRebalancer too? is closed?

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -282,22 +263,22 @@ private boolean validateOfflineInstancesLimit(final ResourceControllerDataProvid
...
-        getWagedRebalancer(helixManager, clusterConfig.getGlobalRebalancePreference(),
-            clusterConfig.isGlobalRebalanceAsyncModeEnabled());
-    try {
-      newIdealStates.putAll(wagedRebalancer.computeNewIdealStates(cache, wagedRebalancedResourceMap,
-          currentStateOutput));
-    } catch (HelixRebalanceException ex) {
-      // Note that unlike the legacy rebalancer, the WAGED rebalance won't return partial result.
-      // Since it calculates for all the eligible resources globally, a partial result is invalid.
-      // TODO propagate the rebalancer failure information to updateRebalanceStatus for monitoring.
+    if (wagedRebalancer != null) {
```
narendly: Perhaps we should check whether the rebalancer has been closed here?

helix-core/src/test/java/org/apache/helix/integration/rebalancer/WagedRebalancer/TestWagedRebalance.java
```diff
@@ -473,6 +473,58 @@ public void testNewInstances()
...
+
+    int i = 0;
+    for (String stateModel : _testModels) {
+      String db = "Test-DB-" + TestHelper.getTestMethodName() + i++;
+      createResourceWithWagedRebalance(CLUSTER_NAME, db, stateModel, PARTITIONS, _replica,
+          _replica);
+      _gSetupTool.rebalanceStorageCluster(CLUSTER_NAME, db, _replica);
+      _allDBs.add(db);
+    }
+    Thread.sleep(300);
```
narendly: We should add a TODO here for all the sleep() calls?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: It is prevented in the get method? This method itself is a protected method, with no external callers.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: Since it is a private class, I don't think we need to overdesign it. As for idempotent, could you be more specific, please? What's the concern that you have?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: The vision is not clear to me. I don't think we have a solid idea for the TODO. For example, relying on the cluster config to set up the rebalancer type may be invalid. If we have multiple optional stateful rebalancers, they will be configured in the ideal state for each resource, instead of cluster config.

More likely, we will need to init all the stateful rebalancers here. It is just too far even for planning. For now, please assume we will only have one stateful rebalancer.

However, making the code generic enough for the future extension is always a good idea. That why I added the interface according to your suggestion : )

jiajunwang: In addition, about your 2nd paragraph comment, I guess what you concerned about is that the rebalancer class should be only configurable in the IS. So the GenericHelixController should not be aware of the resource level thing. However, please note that even there is no resource using the WAGED rebalancer, the controller will still need to init the stateful rebalancer instance.

I agree this design can be improved. But I would prefer to keep it simple before the rebalancer logic and the algorithm are stable enough.

For now, let me add a TODO.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
jiajunwang: That's a good point. For now, an explicit check would be optional. If the rebalancer has been closed, the computing will fail. One concern might be metric. It will be unregistered after the close.

Let me add a TODO to the close() method. I would prefer to keep this PR simple for now.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
jiajunwang: As mentioned above, I would prefer to reject call (by throwing exceptions or so) inside the rebalancer. For the calc stage, there is nothing it can do.
Moreover, checking inside the rebalancer can help to keep the close status private.
TODO added.

helix-core/src/test/java/org/apache/helix/integration/rebalancer/WagedRebalancer/TestWagedRebalance.java
jiajunwang: You are right, but there are too many...
Let me add to the newly created method. After we fixed the issue  https://github.com/apache/helix/issues/526 we will be able to remove all sleep anyway.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
jiajunwang: Thanks for the feedback.

As for the convention, it is not really clear to me. So I did some research. There are different opinions. The one that seems to be more reasonable to me is that enum items should be treated as constants, so they should follow constants name convention. Given this, uppercase letters make more sense here. I intend to enforce this as our enum convention unless a different rule has been documented somewhere else that I don't know : )

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
jiajunwang: One more comment, for now, the close is only called in 2 conditions:
1. the instance is GCed. So there should be no one calling any methods.
2. The controller is shutdown. There won't be a pipeline running after. So there should be no one calling any methods.

That's why I don't think we need to do this check immediately.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
narendly: My concern is that we should probably make the behavior more explicit around create() or close() being multiple times. It wasn't very clear from the code?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1258,4 +1293,57 @@ private void initPipeline(Thread eventThread, BaseControllerDataProvider cache)
...
+        _rebalancer = createRebalancer(helixManager);
+        _isRebalancerValid = true;
+      }
+      // If the rebalance exists but has been marked as invalid (due to leadership switch), it needs
+      // to be reset before return.
+      if (!_isRebalancerValid) {
+        _rebalancer.reset();
+        _isRebalancerValid = true;
+      }
+      return _rebalancer;
```
narendly: Should we follow the lazy-initialization idiom (using volatile keyword on the singleton objects)?

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
narendly: Either way is fine - I realized there is no convention for enum.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
narendly: True, but adding a check is safer and we could add a TODO to remove with better design?

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
narendly: Where is the reject call? Are you referring to Line 271?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: My point is that we don't restrict the behavior in the Ref class. The source of truth is in the rebalancer object itself. So it won't be explicitly mentioned here.
Moreover, as I commented in the other thread, it is not a real concern for now. But I have added a TODO in the rebalancer class for rejecting requests after close() called.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
jiajunwang: No reject call yet. TODO is for this task. It is not a must for now since all the callers are private.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
jiajunwang: I'd prefer adding a TODO to add this check properly. Since that is not one or two lines of change, let's not squeeze it into this PR.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: I still think it is not necessary since we want to maximize the event handling performance here.

1. getRebalancer is called in a single thread, there is no concern that multiple getRebalancer will cause the rebalancer to be created multiple times.
2. the possible concern is between getRebalancer() and closeRebalancer().
2.1. it is not possible that getRebalancer is called after closeRebalancer. Since the event thread has been stopped by the controller before closeRebalancer is triggered.
2.2. if the closeRebalancer is called after getRebalancer, and the reference has not been safed in the main memory, closeRebalancer does nothing. The clean up will be done a little bit late while the rebalancer object is GCed. This is fine. Since the controller has been close, the rebalancer will not be used for sure.
3. On the other hand, when adding volatile to the _rebalancer object, each getRebalancer will access the main memory. It's a minor delay increament, but if it happens on every event, these latencies together might count something. So I prefer not adding volatile unless it help to resolve a real concern.

Moreover, as I mentioned, the rebalancer should not be a singleton by design.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
narendly: I still don't see a TODO stating that we need to check on the rebalancer status (whether it's closed, null, etc.)

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
narendly: Will we be tracking this work with an issue?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
narendly: Adding volatile is for correctness and I don't think we should sacrifice correctness in favor of a minor performance improvement. And I am not positive if the performance is a valid concern at this juncture with so many synchronization primitives that are already in use. 

Now is Helix Controller single-thread? Yes. Will it remain that way? Not necessarily. If we do make this change (which should happen soon, especially due to the memory pressure Controller is facing), we should revisit this code. Could we add a TODO here as well, marking it not thread-safe?

jiajunwang: I understand the concern. And I also thought about that. Actually, even the controller has been changed and split, there probably still be one single thread calculating for resource rebalance. Given that, in the foreseeable future, my assumption will still be valid.
I agree with you that the performance impact is minor. But since it won't impact our correctness, I'd prefer to keep the design simple and not design for the unknown future.

jiajunwang:

I have added TODOs regarding the potential concerns. Since they are related to the possible future improvements or new features, we will revisit those logics by then.

Issue #689, opened 2020-01-17
TestZKUtil is not stable due to the dependency between test methods.


jiajunwang:
The test verifies the expected node count. However, since every test method creates some znodes without removing them, if the methods are executed in a different order as they are listed in the file, the result might be a failure.

mgao0:
@jiajunwang

jiajunwang:
@mgao0 Thanks.

jiajunwang:
@mgao0, please confirm if this issue has been fixed and close if no more work required.

mgao0:
@jiajunwang This issue has been fixed. I don't own this issue so could not close it. Could you help me close the issue please? Thank you!

PR #693, opened 2020-01-21
Removed useless znodes to resolve dependency between test methods in TestZKUtil

mgao0:
### Issues

#689

### Description

There is sometimes test failing because some znodes created in a test method didn't get deleted and it makes another test method's result false because we use one Zk cluster for all test methods in the same test class.
I added code to delete the created znodes at the end of each test method so each test method is truly independent.

### Tests

[INFO] Tests run: 897, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3,521.447 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 897, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  58:54 min
[INFO] Finished at: 2020-01-22T17:53:32-08:00

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestZKUtil.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestZKUtil.java
@@ -37,6 +37,7 @@ import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
+
 public class TestZKUtil extends ZkUnitTestBase {
   private static Logger LOG = LoggerFactory.getLogger(TestZKUtil.class);
 
@@ -98,6 +99,7 @@ public class TestZKUtil extends ZkUnitTestBase {
     AssertJUnit.assertTrue(_gZkClient.exists(path));
     record = _gZkClient.readData(path);
     AssertJUnit.assertEquals("id4", record.getId());
+    _gZkClient.delete(path);
   }
 
   @Test()
@@ -109,12 +111,14 @@ public class TestZKUtil extends ZkUnitTestBase {
     ZKUtil.subtract(_gZkClient, path, record);
     record = _gZkClient.readData(path);
     AssertJUnit.assertNull(record.getSimpleField("key1"));
+    _gZkClient.delete(path);
   }
 
   @Test()
   public void testNullChildren() {
     String path = PropertyPathBuilder.instanceConfig(clusterName, "id6");
     ZKUtil.createChildren(_gZkClient, path, (List<ZNRecord>) null);
+    _gZkClient.delete(path);
   }
 
   @Test()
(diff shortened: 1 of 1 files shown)


jiajunwang:
Looks good. But please rebase the PR so we are seeing your change only.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZKUtil.java
```diff
@@ -37,13 +37,15 @@
...
+
 public class TestZKUtil extends ZkUnitTestBase {
   private static Logger LOG = LoggerFactory.getLogger(TestZKUtil.class);
 
   String clusterName = CLUSTER_PREFIX + "_" + getShortClassName();
 
   @BeforeClass()
-  public void beforeClass() throws Exception {
+  public void beforeClass()
+      throws Exception {
```
pkuwm: Nit, this line is unnecessary. Style format file is updated in master branch. You may want to re-import the file in your Intellij so this line won't be formatted.

mgao0:
This PR is ready to be merged, approved by @pkuwm @alirezazamani

jiajunwang:
It would be easier if we have beforeTest() and afterTest() methods which create and clean up the root path before and after each tests.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZKUtil.java
```diff
@@ -63,6 +65,13 @@ public void afterClass() {
     deleteCluster(clusterName);
   }
 
+  @AfterMethod()
+  public void afterMethod() {
+    String path = PropertyPathBuilder.instanceConfig(clusterName);
+    _gZkClient.deleteRecursively(path);
+    _gZkClient.createPersistent(path);
```
pkuwm: - This is like delete and re-create the cluster, which I don't believe is a better way than deleting the path in the required methods.
- Not every method needs deleting. Ex. `testIsClusterSetup`.
- In my opinion, if a method creates the path, it is the method's responsibility to clean up.

mgao0: I agree it can cause unnecessary delete operation calls such as in `testIsClusterSetup`, but it doesn't add too much latency since what it does is deleting the nodes, not recreating the cluster . It is also safer than leaving the clean up responsibility to each method, as it can make sure the cluster is clean for each test method. I think both way work and it's barely a trade off.

jiajunwang: I checked the test cases one by one, there is no test which is testing to create the root path (please correct me if I am wrong). Given so, re-create the root path is easier and safer.
If we really have a case that requires the root path non-exists, feel free to change that test.

mgao0:
Final commit message:
Removed useless znodes to resolve dependency between test methods in TestZKUtil

Issue #694, opened 2020-01-22
Implement cluster auto registration logic to participant manager


zhangmeng916:
Helix is adding a new feature in which a participant may auto register itself to the cluster after connects. Different from current auto join function, auto registration will not only set up all the config information for the instance, but also retrieve the fault domain information for the participant, and register all this information to the cluster. We need to add this new logic to current participant manager with backward compatibility.

PR #695, opened 2020-01-22
Modify participant manager to add cluster auto registration logic

zhangmeng916:
### Issues

(#694 )

### Description

Helix is adding a new feature in which a participant may auto register itself to the cluster after connects. Different from current auto join function, auto registration will not only set up all the config information for the instance, but also retrieve the fault domain information for the participant, and register all this information to the cluster. This PR changes the logic in ParticipantManager by adding the auto registration logic. Two flags are used to denote auto join and auto registration. AutoJoin flag must be true if we would like to allow autoRegstration. 

### Tests

[INFO] Tests run: 904, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3,221.481 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 904, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  53:45 min
[INFO] Finished at: 2020-01-22T09:57:03-08:00

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -20,6 +20,8 @@ package org.apache.helix.manager.zk;
  */
 
 import java.lang.management.ManagementFactory;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -31,6 +33,7 @@ import org.apache.helix.AccessOption;
 import org.apache.helix.BaseDataAccessor;
 import org.apache.helix.ConfigAccessor;
 import org.apache.helix.HelixAdmin;
+import org.apache.helix.HelixCloudProperty;
 import org.apache.helix.HelixException;
 import org.apache.helix.HelixManager;
 import org.apache.helix.HelixManagerProperty;
@@ -40,6 +43,8 @@ import org.apache.helix.PreConnectCallback;
 import org.apache.helix.PropertyKey;
 import org.apache.helix.ZNRecord;
 import org.apache.helix.ZNRecordBucketizer;
+import org.apache.helix.api.cloud.CloudInstanceInformation;
+import org.apache.helix.api.cloud.CloudInstanceInformationProcessor;
 import org.apache.helix.manager.zk.client.HelixZkClient;
 import org.apache.helix.messaging.DefaultMessagingService;
 import org.apache.helix.model.CurrentState;
@@ -52,6 +57,7 @@ import org.apache.helix.model.StateModelDefinition;
 import org.apache.helix.model.builder.HelixConfigScopeBuilder;
 import org.apache.helix.participant.StateMachineEngine;
 import org.apache.helix.participant.statemachine.ScheduledTaskStateModelFactory;
+import org.apache.helix.util.HelixUtil;
 import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -61,6 +67,7 @@ import org.slf4j.LoggerFactory;
  */
 public class ParticipantManager {
   private static Logger LOG = LoggerFactory.getLogger(ParticipantManager.class);
+  private static String CLOUD_PROCESSOR_PATH_PREFIX = "org.apache.helix.cloud.";
 
   final HelixZkClient _zkclient;
   final HelixManager _manager;
@@ -107,7 +114,7 @@ public class ParticipantManager {
   }
 
   /**
-   * Handle new session for a participang.
+   * Handle new session for a participant.
    * @throws Exception
    */
   public void handleNewSession() throws Exception {
@@ -132,18 +139,27 @@ public class ParticipantManager {
   }
 
   private void joinCluster() {
-    // Read cluster config and see if instance can auto join the cluster
+    // Read cluster config and see if an instance can auto join or auto register to the cluster
     boolean autoJoin = false;
+    boolean autoRegistration = false;
     try {
-      HelixConfigScope scope =
-          new HelixConfigScopeBuilder(ConfigScopeProperty.CLUSTER).forCluster(
-              _manager.getClusterName()).build();
-      autoJoin =
-          Boolean.parseBoolean(_configAccessor.get(scope,
-              ZKHelixManager.ALLOW_PARTICIPANT_AUTO_JOIN));
+      HelixConfigScope scope = new HelixConfigScopeBuilder(ConfigScopeProperty.CLUSTER)
+          .forCluster(_manager.getClusterName()).build();
+      autoJoin = Boolean
+          .parseBoolean(_configAccessor.get(scope, ZKHelixManager.ALLOW_PARTICIPANT_AUTO_JOIN));
       LOG.info("instance: " + _instanceName + " auto-joining " + _clusterName + " is " + autoJoin);
     } catch (Exception e) {
       // autoJoin is false
+    }
+
+    // Read cloud config and see if an instance can auto register to the cluster
+    try {
+      autoRegistration =
+          Boolean.valueOf(_helixManagerProperty.getHelixCloudProperty().getCloudEnabled());
+      LOG.info("instance: " + _instanceName + " auto-register " + _clusterName + " is "
+          + autoRegistration);
+    } catch (Exception e) {
+      // autoRegistration is false
     }
 
     if (!ZKUtil.isInstanceSetup(_zkclient, _clusterName, _instanceName, _instanceType)) {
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -132,42 +139,102 @@ public void handleNewSession() throws Exception {
...
     }
 
+    // Read cloud config and see if an instance can auto register to the cluster
+    try {
+      autoRegistration =
+          Boolean.valueOf(_helixManagerProperty.getHelixCloudProperty().getCloudEnabled());
+      LOG.info("instance: " + _instanceName + " auto-register " + _clusterName + " is "
+          + autoRegistration);
+    } catch (Exception e) {
+      // autoRegistration is false
```
alirezazamani: Can you add a log for this exception?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -132,42 +139,102 @@ public void handleNewSession() throws Exception {
...
-        InstanceConfig instanceConfig = new InstanceConfig(_instanceName);
-        String hostName = _instanceName;
-        String port = "";
-        int lastPos = _instanceName.lastIndexOf("_");
-        if (lastPos > 0) {
-          hostName = _instanceName.substring(0, lastPos);
-          port = _instanceName.substring(lastPos + 1);
+        if (!autoRegistration) {
+          LOG.info(_instanceName + " is auto-joining cluster: " + _clusterName);
+          addInstanceConfig(null);
```
alirezazamani: Question: Why do we need to add instance config here?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -132,42 +139,102 @@ public void handleNewSession() throws Exception {
...
+        } else {
```
dasahcc: I would suggest to refactor code to be:

String domainInfo = null;
if (autoRegistration) { 
    ..... do the logic as following
}
addInstanceConfig(domainInfo);

Passing null explicit looks awkward to me.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -132,42 +139,102 @@ public void handleNewSession() throws Exception {
...
+            addInstanceConfig(domain);
+          } catch (ClassNotFoundException ex) {
+            throw new HelixException(
+                "Passed cloud instance information processor class is not found: "
+                    + cloudInstanceInformationProcessorName, ex);
+          } catch (NoSuchMethodException ex) {
+            throw new HelixException("Failed to get the constructor for the class: "
+                + cloudInstanceInformationProcessorName, ex);
+          } catch (InstantiationException | IllegalAccessException | InvocationTargetException ex) {
+            throw new HelixException("Failed to create a new instance for the class: "
```
dasahcc: Any reason to differentiate these exceptions? They looked like instantiation failure.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -61,6 +67,7 @@
  */
 public class ParticipantManager {
   private static Logger LOG = LoggerFactory.getLogger(ParticipantManager.class);
+  private static String CLOUD_PROCESSOR_PATH_PREFIX = "org.apache.helix.cloud.";
```
jiajunwang: nit, final?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -132,42 +139,102 @@ public void handleNewSession() throws Exception {
...
+      autoJoin = Boolean
+          .parseBoolean(_configAccessor.get(scope, ZKHelixManager.ALLOW_PARTICIPANT_AUTO_JOIN));
       LOG.info("instance: " + _instanceName + " auto-joining " + _clusterName + " is " + autoJoin);
     } catch (Exception e) {
       // autoJoin is false
     }
 
+    // Read cloud config and see if an instance can auto register to the cluster
+    try {
+      autoRegistration =
```
jiajunwang: So cloud enabled == auto join?
If so, let's document it more clear.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -132,42 +139,102 @@ public void handleNewSession() throws Exception {
...
+          + autoRegistration);
+    } catch (Exception e) {
+      // autoRegistration is false
+    }
+
     if (!ZKUtil.isInstanceSetup(_zkclient, _clusterName, _instanceName, _instanceType)) {
       if (!autoJoin) {
         throw new HelixException("Initial cluster structure is not set up for instance: "
             + _instanceName + ", instanceType: " + _instanceType);
       } else {
```
jiajunwang: I think you can have more flat condition. elseif (..) will also work.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -132,42 +139,102 @@ public void handleNewSession() throws Exception {
...
-        if (lastPos > 0) {
-          hostName = _instanceName.substring(0, lastPos);
-          port = _instanceName.substring(lastPos + 1);
+        if (!autoRegistration) {
+          LOG.info(_instanceName + " is auto-joining cluster: " + _clusterName);
+          addInstanceConfig(null);
+        } else {
+          String cloudInstanceInformationProcessorName =
+              _helixManagerProperty.getHelixCloudProperty().getCloudInfoProcessorName();
+          try {
```
jiajunwang: This is a long and standalone logic just for cloud, I guess. Is it possible to hide them into the processor? I remember that we discussed if we can have a factory or builder for the property. It also applied to the processor. If we have one, we can just put all these logic there. And whoever reading participant manager code won't be distracted from the participant logic.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -132,42 +139,102 @@ public void handleNewSession() throws Exception {
...
-        instanceConfig.setPort(port);
-        instanceConfig.setInstanceEnabled(true);
-        _helixAdmin.addInstance(_clusterName, instanceConfig);
       }
     }
   }
 
+  private void addInstanceConfig(String domainInfo) {
+    InstanceConfig instanceConfig = new InstanceConfig(_instanceName);
+    String hostName = _instanceName;
```
jiajunwang: I feel there has been such a logic. If not, it should be in util anyway.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
zhangmeng916: This is for backward compatibility with current "auto-join" logic. For "auto-join", it will populate all instance config to Zookeeper except "domain" information.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
zhangmeng916: when auto join is true, cloudEnabled == auto registration

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
zhangmeng916: The logic is only when an instance is not setup, we will check auto join, if auto join is false, an exception will be thrown since no one will add instance, if auto join is true, we will continue to check auto registration, and decide whether we do auto join or auto registration. This is also for the backward compatibility.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
zhangmeng916: I restructured the code a bit by separating those processor specific logic out. I keep the logic in this file is because HelixManagerProperty is only available there and is not a global variable

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
dasahcc: Was this updated?

zhangmeng916: Yeah, I updated it as something below:
 InstanceConfig instanceConfig = HelixUtil.composeInstanceConfig(_instanceName);
          instanceConfig.setDomain(domain);
          _helixAdmin.addInstance(_clusterName, instanceConfig);

Can you check whether you can see the change?

zhangmeng916:
Final commit message:
Modify participant logic to add logic for instance auto registration to a Helix cluster.
Auto registration differs with existing auto join in that auto registration will retrieve the fault domain information for the instance and populate it under instanceConfig. With auto registration on, user does not need to manually populate instance information in Zookeeper.

Issue #697, opened 2020-01-23
Implement Azure cloud instance information processor


zhangmeng916:
Helix will provide a function for participant to auto register to the cluster so that no manual efforts are needed for adding instance to a cluster. To realize it, a cloud instance information processor will take the responsibility for getting and parsing the instance information from cloud. Different cloud providers will have different implementations for the processor. We will implement the processor for Azure as a first use case.

PR #698, opened 2020-01-23
Implement Azure cloud instance information processor

zhangmeng916:
### Issues

(#697 )

### Description

Helix will provide a function for participant to auto register to the cluster so that no manual efforts are needed for adding instance to a cluster. To realize it, a cloud instance information processor will take the responsibility for getting and parsing the instance information from cloud. Different cloud providers will have different implementations for the processor. This PR implement the processor for Azure as a first use case.

### Tests

TestAzureCloudInstanceInformationProcessor

[INFO] Tests run: 904, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3,221.481 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 904, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  53:45 min
[INFO] Finished at: 2020-01-22T09:57:03-08:00
[mnzhang@mnzhang-ld1 helix-core]$ git status

### Commits

### Code Quality


helix-core/src/test/resources/AzureResponse.json
```diff
@@ -0,0 +1,104 @@
+{
+  "compute": {
+    "azEnvironment": "AzurePublicCloud",
+    "customData": "",
+    "location": "southcentralus",
+    "name": "ei-lid-vmss-kafka_1",
```
alirezazamani: I would prefer to use other names such as "Test-Helix" in our tests.

helix-core/src/test/resources/AzureResponse.json
```diff
@@ -0,0 +1,104 @@
...
+          "enabled": "false"
+        },
+        "image": {
+          "uri": ""
+        },
+        "managedDisk": {
+          "id": "/subscriptions/c9a251d8-1272-4c0f-8055-8271bbc1d677/resourceGroups/scus-lpsazureei1-app-rg/providers/Microsoft.Compute/disks/ei-lid-vmss-kafka_ei-lid-vmss-kafka_2_OsDisk_1_124c3534b8e848e296ec22b24d44c027",
+          "storageAccountType": "Standard_LRS"
+        },
+        "name": "ei-lid-vmss-kafka_ei-lid-vmss-kafka_2_OsDisk_1_124c3534b8e848e296ec22b24d44c027",
```
alirezazamani: Same as here.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureHttpUtil.java
```diff
@@ -0,0 +1,45 @@
...
+import java.io.InterruptedIOException;
+import java.net.UnknownHostException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A dedicated http client for retrieving information from Azure Instance Metadata Service
+ */
+class AzureHttpUtil {
```
jiajunwang: On general question, Yi has added one httpclient in the Helix-rest for instance health check. Can we just leverage that one?
In that case, just move the code to helix-core, or better to helix-common (I hope we have one), then leverage that client.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
```diff
@@ -19,38 +19,111 @@
...
-   * fetch the raw Azure cloud instance information
+   * Fetch raw Azure cloud instance information based on the urls provided
    * @return raw Azure cloud instance information
    */
   @Override
   public List<String> fetchCloudInstanceInformation() {
     List<String> response = new ArrayList<>();
-    //TODO: implement the fetching logic
+    for (String url : _helixCloudProperty.getCloudInfoSources()) {
+      response.add(getAzureCloudInformationFromUrl(url));
```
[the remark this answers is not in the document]
zhangmeng916: Cloud information is considered as one entity, if any part of it is missing/failed, we would consider it as incomplete, and fail the operation. As we cannot guarantee the fields we need later is already populated, if we do not fail here, we will need to fail in the participant manager.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureHttpUtil.java
zhangmeng916: I've checked the current custom rest client. It's not easy to adapt it to our use case, specifically for those user defined parameter input, such as max retry, timeout. Also the separate http client I have is easier to mock in the unit test.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
```diff
@@ -19,38 +19,111 @@
...
+    _helixCloudProperty = helixCloudProperty;
+  }
 
-  public AzureCloudInstanceInformationProcessor() {
+  /**
+   * This constructor is for unit test purpose only.
+   * User could provide helixCloudProperty and a mocked http client to test the functionality of
+   * this class.
+   */
+  public AzureCloudInstanceInformationProcessor(HelixCloudProperty helixCloudProperty,
```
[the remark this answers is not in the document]
zhangmeng916: I tried, compilation breaks due to testng not in the package. Meanwhile, I'd like to keep the consistency with the current tests. Let me know if you feel protected is a must.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
```diff
@@ -19,38 +19,110 @@
...
    */
   @Override
   public AzureCloudInstanceInformation parseCloudInstanceInformation(List<String> responses) {
     AzureCloudInstanceInformation azureCloudInstanceInformation = null;
-    //TODO: implement the parsing logic
+    for (String response : responses) {
+      ObjectMapper mapper = new ObjectMapper();
+      try {
+        JsonNode jsonNode = mapper.readTree(response);
+        JsonNode computeNode = jsonNode.path("compute");
```
dasahcc: Let's make these as variables instead of hard code.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
```diff
@@ -19,38 +19,110 @@
...
+        if (!computeNode.isMissingNode()) {
+          String vmName = computeNode.path("vmId").getTextValue();
+          String platformFaultDomain = computeNode.path("platformFaultDomain").getTextValue();
+          String vmssName = computeNode.path("vmScaleSetName").getValueAsText();
```
dasahcc: Is VMSS name necessary? What if this instance joined from container?

zhangmeng916: If it's not inside a VMSS, the field will be empty. Originally VMSS is for validation purpose. Based on our current plan, it's more likely we will not use this field. However, I prefer to keep it here just in case we can still use it for validation.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureHttpUtil.java
jiajunwang: httpclient should be a fundamental component. If we have 2 of them in our codebase, it would be confusing and harder for us to maintain.
How difficult will it be to merge 2 use cases? Please feel free to refactor if necessary.

@i3wangyi Can you please comment as well?

i3wangyi: Yes. The current CustomRestClient has one instance of `HttpClient`, a parent class of the `CloseableHttpClient` the method tries to return. However, I see the AzureHttpUtil is in helix-core while that factory is only visible in helix-rest only (because previously there's no need for helix-core to perform HTTP request).

Essentially, the factory pattern is similar and we can extract the core part of it to helix-core and leave the difference in the separate module. For example, in helix-core, create one HttpClientFactory which accepts a set of general parameters (timeout, ssl, retry..) without respecting HelixCloudProperty; And use it in rest as well as rest of places in helix-core

jiajunwang: It is OK if we move the current client class to the helix-core now. Since there is real usage of it.

zhangmeng916: Actually I double checked and found that I can just remove the util function. The processor itself can just directly make the http call, and passing the information it needs to pass. So the future usage should be, any class that needs to make the call would directly call getHttpClient, providing whatever config they would like to pass in.

helix-core/pom.xml
```diff
@@ -159,6 +159,11 @@ under the License.
       <artifactId>metrics-core</artifactId>
       <version>3.2.3</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>4.5.8</version>
```
jiajunwang: nit, can you have a try to remove the corresponding entry in the helix-rest pom? I think that would be unnecessary with this.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformation.java
```diff
@@ -42,7 +42,8 @@ public String get(String key) {
   }
 
   public static class Builder {
-    private Map<String, String> _cloudInstanceInfoMap = null;
+
+    private final Map<String, String> _cloudInstanceInfoMap = new HashMap<>();
 
     public AzureCloudInstanceInformation build() {
       return new AzureCloudInstanceInformation(_cloudInstanceInfoMap);
```
jiajunwang: Safer to clone construct a new map so the newly built instances do not share anything.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
```diff
@@ -19,38 +19,135 @@
...
+
+    HttpRequestRetryHandler httpRequestRetryHandler =
+        (IOException exception, int executionCount, HttpContext context) -> {
+          LOG.warn("Execution count: " + executionCount + ".", exception);
+          return !(executionCount >= helixCloudProperty.getCloudMaxRetry()
+              || exception instanceof InterruptedIOException
+              || exception instanceof UnknownHostException || exception instanceof SSLException);
+        };
+
+    _closeableHttpClient = HttpClients.custom().setDefaultRequestConfig(requestConifg)
```
jiajunwang: As we discussed, let's add a TODO here for generalizing all the httpclients usage.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
```diff
@@ -19,38 +19,135 @@
...
+
   /**
    * Parse raw Azure cloud instance information.
    * @return required azure cloud instance information
    */
   @Override
   public AzureCloudInstanceInformation parseCloudInstanceInformation(List<String> responses) {
     AzureCloudInstanceInformation azureCloudInstanceInformation = null;
-    //TODO: implement the parsing logic
+    for (String response : responses) {
```
jiajunwang: It confused me. If there are multiple responses, which information you return? It seems that the last one will override the previous ones.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
```diff
@@ -19,38 +19,135 @@
...
+    _closeableHttpClient = closeableHttpClient;
+  }
+
+  /**
+   * Fetch raw Azure cloud instance information based on the urls provided
    * @return raw Azure cloud instance information
    */
   @Override
   public List<String> fetchCloudInstanceInformation() {
     List<String> response = new ArrayList<>();
```
jiajunwang: nit, responseList or responses

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
zhangmeng916: If there are multiple responses, they should come from different sources, e.g. some may come from local disk, and they should not have common fields. If they do, we will need to define priority among these sources.

zhangmeng916: Discussed offline. Updated Azure implementation to restrict the response size to 1, as we do not have any source to query for now.

zhangmeng916:
This PR is ready to be merged, approved by @dasahcc and @alirezazamani 
Final Commit Message:
Implement Azure cloud instance information processor. The processor performs the functions of fetching and parsing instance information from Azure cloud environment. The endpoint that the processor queries is a globally standard url, called Azure Instance Metadata Service.

PR #700, opened 2020-01-23
Call session aware createEphemeral to create live instance.

### LINKED ISSUE #699: Call session aware createEphemeral to create live instance
We implemented a new API ZkClient.createEphemeral(path, data, sessionId) to validate the zk session and create a live instance. Currently, live instance creation is still using the old API createEphemeral(path, data) without the session ID. So zk session race condition is not completely fixed. And we would like to fix this.

pkuwm:
### Issues

Fixes #699 

### Description

We implemented a new API `ZkClient.createEphemeral(path, data, sessionId)` to validate the zk session and create a live instance. Currently, live instance creation is still using the old API `createEphemeral(path, data)` without the session ID. So zk session race condition is not completely fixed.

This PR changes the api to use `createEphemeral(path, data, sessionId)` for live instance creation. A new test is added to cover the case of using the old API. If the old API is used, the test will fail like below:
```
java.lang.AssertionError: Live instance should not be created because zk session is expired!
Expected :true
Actual   :false
```

### Tests

- TestParticipantManager.testSessionExpiryCreateLiveInstance

```
[INFO] Tests run: 898, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3,370.606 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 898, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  56:15 min
[INFO] Finished at: 2020-01-22T21:35:30-08:00
```

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -201,6 +201,7 @@ public class ParticipantManager {
     do {
       retry = false;
       try {
+        // Zk session ID will be validated in createEphemeral.
         _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), _sessionId);
         LOG.info("LiveInstance created, path: {}, sessionId: {}", liveInstancePath,
             liveInstance.getEphemeralOwner());
@@ -243,6 +244,7 @@ public class ParticipantManager {
      */
     if (retry) {
       try {
+        // Zk session ID will be validated in createEphemeral.
         _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), _sessionId);
         LOG.info("LiveInstance created, path: {}, sessionId: {}", liveInstancePath,
             liveInstance.getEphemeralOwner());
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -201,7 +201,7 @@ public class ParticipantManager {
     do {
       retry = false;
       try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord());
+        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), _sessionId);
         LOG.info("LiveInstance created, path: {}, sessionId: {}", liveInstancePath,
             liveInstance.getEphemeralOwner());
       } catch (ZkSessionMismatchedException e) {
@@ -243,7 +243,7 @@ public class ParticipantManager {
      */
     if (retry) {
       try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord());
+        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), _sessionId);
         LOG.info("LiveInstance created, path: {}, sessionId: {}", liveInstancePath,
             liveInstance.getEphemeralOwner());
       } catch (ZkSessionMismatchedException e) {
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
@@ -110,7 +110,7 @@ public class TestParticipantManager extends ZkTestBase {
 
     TestHelper.verify(
         () -> !(ZKUtil.toHexSessionId(manager.getZkClient().getSessionId()).equals(lastSessionId)),
-        3000L);
+        TestHelper.WAIT_DURATION);
 
     // New session S2 should not be equal to S1.
     final String latestSessionId = ZKUtil.toHexSessionId(manager.getZkClient().getSessionId());
@@ -119,7 +119,7 @@ public class TestParticipantManager extends ZkTestBase {
     // Proceed S1 to create live instance, which will fail.
     startCountdown.countDown();
 
-    // Wait until S1's handling new session is completed.
+    // Wait until S2 starts onPreConnect, which indicates S1's handling new session is completed.
     semaphore.acquire();
 
     // Live instance should not be created because zk session is expired.
@@ -137,7 +137,7 @@ public class TestParticipantManager extends ZkTestBase {
       return newLiveInstance != null
           && newLiveInstance.getStat().getCreationTime() != originalCreationTime
           && newLiveInstance.getEphemeralOwner().equals(latestSessionId);
-    }, 2000L);
+    }, TestHelper.WAIT_DURATION);
 
     // Clean up.
     manager.syncStop();
@@ -153,7 +153,7 @@ public class TestParticipantManager extends ZkTestBase {
     private final CountDownLatch endCountDown;
     private final Semaphore semaphore;
 
-    private int count;
+    private boolean canCreateLiveInstance;
 
     BlockingPreConnectCallback(String instanceName, CountDownLatch startCountdown,
         CountDownLatch endCountdown, Semaphore semaphore) {
@@ -170,13 +170,14 @@ public class TestParticipantManager extends ZkTestBase {
       try {
         LOG.info("Waiting session expiry to happen.");
         startCountDown.await();
-        if (count++ == 1) {
+        if (canCreateLiveInstance) {
           LOG.info("Waiting to continue creating live instance.");
           endCountDown.await();
         }
       } catch (InterruptedException ex) {
         LOG.error("Interrupted in waiting", ex);
       }
+      canCreateLiveInstance = true;
     }
   }
 }
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
@@ -40,7 +40,8 @@ public class TestParticipantManager extends ZkTestBase {
 
   /*
    * Simulates zk session expiry before creating live instance in participant manager. This test
-   * makes sure the session aware create ephemeral API is called.
+   * makes sure the session aware create ephemeral API is called, which validates the expected zk
+   * session.
    * What this test does is:
    * 1. Sets up live instance with session S0
    * 2. Expires S0 and gets new session S1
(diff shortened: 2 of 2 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -201,7 +201,7 @@ private void createLiveInstance() {
     do {
       retry = false;
       try {
-        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord());
+        _zkclient.createEphemeral(liveInstancePath, liveInstance.getRecord(), _sessionId);
```
narendly: Question: this `createLiveInstance` method is supposed to check if the `_sessionId` is still valid or not. Could you point out where we're doing the check (besides the one happening when we catch `ZkNodeExistsException`)?

Also, is it possible that `_sessionId` is stale? For example, while the ParticipantManager is in the middle of `handleNewSession`, session expired?

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
```diff
@@ -0,0 +1,182 @@
...
+   * Simulates zk session expiry before creating live instance in participant manager. This test
+   * makes sure the session aware create ephemeral API is called.
+   * What this test does is:
+   * 1. Sets up live instance with session S0
+   * 2. Expires S0 and gets new session S1
+   * 3. S1 is blocked before creating live instance in participant manager
+   * 4. Expires S1 and gets new session S2
+   * 5. Proceeds S1 to create live instance, which will fail because session S1 is expired
+   * 6. Proceeds S2 to create live instance, which will succeed
+   */
```
narendly: I just wanted to say thank you for being so great with JavaDoc and the PR description! Let's keep it up :)

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
```diff
@@ -0,0 +1,182 @@
...
+  /*
+   * Mocks PreConnectCallback to insert session expiry during ParticipantManager#handleNewSession()
+   */
+  static class BlockingPreConnectCallback implements PreConnectCallback {
+    private final String instanceName;
+    private final CountDownLatch startCountDown;
+    private final CountDownLatch endCountDown;
+    private final Semaphore semaphore;
+
+    private int count;
```
narendly: Nit: are we using this count as a flag? If so, could we just use a boolean flag?

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
```diff
@@ -0,0 +1,182 @@
...
+
+    TestHelper.verify(() -> {
+      // Newly created live instance should be created by the latest session S2
+      // and have a new creation time.
+      LiveInstance newLiveInstance =
+          accessor.getProperty(keyBuilder.liveInstance(instanceName));
+      return newLiveInstance != null
+          && newLiveInstance.getStat().getCreationTime() != originalCreationTime
+          && newLiveInstance.getEphemeralOwner().equals(latestSessionId);
+    }, 2000L);
```
narendly: Nit: we seem to be being inconsistent with wait times - could we use TestHelper's default wait time (if there is a constant defined already)?

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
```diff
@@ -0,0 +1,182 @@
...
+
+    // New session S2 should not be equal to S1.
+    final String latestSessionId = ZKUtil.toHexSessionId(manager.getZkClient().getSessionId());
+    Assert.assertFalse(lastSessionId.equals(latestSessionId));
+
+    // Proceed S1 to create live instance, which will fail.
+    startCountdown.countDown();
+
+    // Wait until S1's handling new session is completed.
+    semaphore.acquire();
```
narendly: Does this really wait? At this point isn't `release()` called by S2?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
pkuwm: Good question. 
- The validation is put in `create(path, dataObject, acl, mode, expectedSessionId)` (which is called by `createEphemeral(path, data, session)`):
https://github.com/apache/helix/blob/662dc0f14ed1db49f2c1a41bf55c7cfcecbc7241/helix-core/src/main/java/org/apache/helix/manager/zk/zookeeper/ZkClient.java#L662-L670

   And this is the reason of current PR.

- _sessionId is possibly stale. So we want to fix it in this PR by calling the new API with session id. The new API does final check and create ephemeral node.

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
pkuwm: New sessions are sequentially handled by a single thread: Handling S1 -> S1 completes -> handling S2 -> release by S2. So if `release()` is called by S2, it means S1 is completed which is what we want to wait for.

I realized that my comment was not that clear. I have modified the comment: `Wait until S2 start onPreConnect, which indicates S1's handling new session is completed.`

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
pkuwm: There is a constant wait time (20 seconds). 20s is kind of too long for this test because if waiting is longer than 2s, the test should definitely fail. But anyway, I will adopt your suggestion to use the constant. Maybe we can add at least one more constant to TestHelper like `SHORT_WAIT_DURATION`

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
pkuwm: Changed.

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
```diff
@@ -0,0 +1,183 @@
+package org.apache.helix.manager.zk;
+
```
lei-xia: Thanks for adding the test!

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
```diff
@@ -0,0 +1,183 @@
...
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+
+public class TestParticipantManager extends ZkTestBase {
+  private static Logger LOG = LoggerFactory.getLogger(TestParticipantManager.class);
+
+  /*
+   * Simulates zk session expiry before creating live instance in participant manager. This test
+   * makes sure the session aware create ephemeral API is called.
```
jiajunwang: not only is called, but also validates the session as expected.

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
```diff
@@ -0,0 +1,183 @@
...
+    // Expire S1 when S1 is blocked in onPreConnect().
+    // New session S2 will be created.
+    ZkTestHelper.asyncExpireSession(manager.getZkClient());
+
+    TestHelper.verify(
+        () -> !(ZKUtil.toHexSessionId(manager.getZkClient().getSessionId()).equals(lastSessionId)),
+        TestHelper.WAIT_DURATION);
+
+    // New session S2 should not be equal to S1.
+    final String latestSessionId = ZKUtil.toHexSessionId(manager.getZkClient().getSessionId());
```
jiajunwang: nit, let's just call them sessionOne, sessionTwo.

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
pkuwm: Validation is done in "session aware create ephemeral API"

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
pkuwm: Done.

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
jiajunwang: np, just suggest you make the doc more generic. But I don't have a strong preference.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
narendly: Great. Could we add a quick comment here:

"_sessionId will be validated in createEphemeral"

pkuwm: Done.

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
pkuwm: Sure. I should have added the test in the prior PR https://github.com/apache/helix/pull/642

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
pkuwm: Done

helix-core/src/test/java/org/apache/helix/manager/zk/TestParticipantManager.java
pkuwm: Completed. Thanks for carefully reviewing my PR.

pkuwm:
Thanks for the prompt review, @jiajunwang @narendly !
I've run the tests again and all passed.

This PR is ready to be merged, approved by @jiajunwang @narendly 

```
Live instance creation is still using the old API createEphemeral(path, data) without the session ID, resulting in zk session race condition being not completely fixed. So this commit changes the api to use createEphemeral(path, data, sessionId) for zk session validation and live instance creation. A new unit test is also added to cover session expiry in participant manager when creating live instance.
```

Issue #702, opened 2020-01-23
Create Helix distributed lock interface


mgao0:
Design a Helix distributed lock interface that is easy to understand for customers, generic to different use cases, and neutral to underlying implementations. Implementing this interface would allow Helix customer to have a distributed lock that can guide the coordination of different processes.

PR #703, opened 2020-01-23
Created Helix distributed lock interface

mgao0:
### Issues

#702 

### Description

Created a Helix distributed lock design with acquire, release, and getInfo functionalities. 

### Tests

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
+++ b/helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
@@ -22,7 +22,7 @@ package org.apache.helix.lock;
 /**
  * Generic interface for Helix distributed lock for both nonblocking and blocking calls
  */
-public interface HelixLock {
+public interface HelixLock<T> {
   /**
    * Acquire a lock
    * @return true if the lock was successfully acquired,
@@ -42,7 +42,7 @@ public interface HelixLock {
    * Retrieve the lock information, e.g. lock timeout, lock message, etc.
    * @return lock information
    */
-  LockInfo getLockInfo();
+  LockInfo<T> getLockInfo();
 
   /**
    * If the user is current lock owner
--- a/helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
+++ b/helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
@@ -20,24 +20,34 @@ package org.apache.helix.lock;
  */
 
 /**
- * Generic interface for Helix distributed lock
+ * Generic interface for Helix distributed lock for both nonblocking and blocking calls
  */
 public interface HelixLock {
   /**
    * Acquire a lock
-   * @return true if the lock was acquired, false if could not be acquired
+   * @return true if the lock was successfully acquired,
+   * false if the lock could not be acquired
    */
-  public boolean acquireLock();
+  boolean acquireLock();
 
   /**
    * Release a lock
-   * @return true if the lock was released, false if it could not be released
+   * @return true if the lock was successfully released,
+   * false if the locked is not locked or is not locked by the user,
+   * or the lock could not be released
    */
-  public boolean releaseLock();
+  boolean releaseLock();
 
   /**
    * Retrieve the lock information, e.g. lock timeout, lock message, etc.
    * @return lock information
    */
-  public Object getLockInfo();
+  LockInfo getLockInfo();
+
+  /**
+   * If the user is current lock owner
+   * @return true if the user is the lock owner,
+   * false if the user is not the lock owner or the lock doesn't have a owner
+   */
+  boolean isOwner();
 }
--- a/helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
+++ b/helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
@@ -1,0 +1,43 @@
+package org.apache.helix.lock;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Generic interface for Helix distributed lock
+ */
+public interface HelixLock {
+  /**
+   * Acquire a lock
+   * @return true if the lock was acquired, false if could not be acquired
+   */
+  public boolean acquireLock();
+
+  /**
+   * Release a lock
+   * @return true if the lock was released, false if it could not be released
+   */
+  public boolean releaseLock();
+
+  /**
+   * Retrieve the lock information, e.g. lock timeout, lock message, etc.
+   * @return lock information
+   */
+  public Object getLockInfo();
+}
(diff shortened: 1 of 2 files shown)


helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
```diff
@@ -0,0 +1,43 @@
...
+   * Release a lock
+   * @return true if the lock was released, false if it could not be released
+   */
+  public boolean releaseLock();
+
+  /**
+   * Retrieve the lock information, e.g. lock timeout, lock message, etc.
+   * @return lock information
+   */
+  public Object getLockInfo();
```
lei-xia: should this getLockInfo return a defined type (for example, LockInfo)?    Also, how does a piece of code check whether it owns the lock or not?

mgao0: Good point. I'll change the return type to LockInfo, which could these fields for now: owner ID, expiring time, lock message.
Since the lock info already have the owner ID, user can check if they are the lock owner or not by comparing their ID to the ID returned in LockInfo. If we want to make it easier for users, we can take in the user ID in the lock constructor and when they call isOwner() API, we compare the user ID with  the owner ID, and return true or false.

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Generic interface for Helix distributed lock
+ */
+public interface HelixLock {
+  /**
+   * Acquire a lock
+   * @return true if the lock was acquired, false if could not be acquired
+   */
+  public boolean acquireLock();
```
pkuwm: - Do we want `public` modifier here because the interface is already `public`? I don't think it is necessary. All abstract, default, and static methods in an interface are implicitly public, so you can omit the public modifier.
- Do we really need to return a boolean? I think it makes sense to return a boolean for a `tryAcquire`, but for `acquire`, I think void is enough and clean.
- How about just naming `acquire`
- Do we want a `tryAcquire` with timeout?

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
zhangmeng916: Please try to make LockInfo very generic, as we will have various implementations, and we cannot force all of them to have certain fields.

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
mgao0: - Good catch, will remove `public`.

- This interface is a generic interface for both blocking and nonblocking blocks. In the design, if a false is return for `acquireLock`, it means the user failed to acquire the lock; a returned true means the lock is successfully acquired. A user cannot work on the resource without holding the lock.

- I personally think `acquireLock` is more clear that the user is using a lock, even if the variable name does not show it's a lock

- For blocking lock, we will not expose the `tryAcquire` to our users, `tryAcquire` and `retry` will be wrapped inside the `acquireLock` implementation.

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
```diff
@@ -0,0 +1,43 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Generic interface for Helix distributed lock
+ */
+public interface HelixLock {
+  /**
+   * Acquire a lock
```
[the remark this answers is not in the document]
mgao0: The interface is for both blocking or nonblocking calls. I added into the comment.

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
```diff
@@ -0,0 +1,43 @@
...
+   * @return true if the lock was acquired, false if could not be acquired
+   */
+  public boolean acquireLock();
+
+  /**
+   * Release a lock
+   * @return true if the lock was released, false if it could not be released
```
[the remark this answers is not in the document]
mgao0: False will be returned. I added it into the method comment.

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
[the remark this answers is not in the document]
dasahcc: If we design the method for both types, it may end up with only one implementation at a time. What if user may have both use cases for same lock object?

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
```diff
@@ -0,0 +1,53 @@
...
+   */
+  boolean acquireLock();
+
+  /**
+   * Release a lock
+   * @return true if the lock was successfully released,
+   * false if the locked is not locked or is not locked by the user,
+   * or the lock could not be released
+   */
+  boolean releaseLock();
```
dasahcc: Do we consider privilege override or something else? What is the scope of locking

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -0,0 +1,44 @@
...
+ * @param <T> The type of the LockInfo value
+ */
+public interface LockInfo<T> {
+
+  /**
+   * Create a single filed of LockInfo, or update the value of the field if it already exists
+   * @param key the key of the LockInfo field
+   * @param value the value of the LockInfo field
+   */
+  void setValue(String key, T value);
```
dasahcc: the method name is too generic.

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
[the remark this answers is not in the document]
mgao0: @dasahcc Not sure I fully understand the scenario you described. Can you provide a specific example? What we want to achieve is to hide the implementation from our customers. So they choose to use nonblocking or blocking lock by using different HelixLock constructors, then they use the lock by calling `acquire` or `release`, we handle the difference of nonblocking and blocking in implementations.

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
mgao0: We do not consider privilege override at this moment. But it may be considered later.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -0,0 +1,46 @@
...
+package org.apache.helix.lock;
+
+import java.util.Map;
+
+
+/**
+ * Generic interface for a map contains the Helix lock information
+ * @param <T> The type of the LockInfo value
+ */
+public interface LockInfo<T> {
```
jiajunwang: Optional for this PR. But do we consider some mandatory fields to be here? For example, timeout, owner ID, etc.

jiajunwang:
Please ensure you addressed the other's comments as well.

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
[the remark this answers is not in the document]
mgao0: Currently this interface only includes blocking calls. Nonblocking calls can be added later if there is a need from our customers.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
mgao0: I added a TODO to add specific setter and getter for fields that we identify that are mandatory as LockInfo, so the method names can make more sense to our customers.

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
mgao0: Added LockInfo interface as the type contains lock metadata. And I also added a isOwner method for customers to check if they are the owner of the lock.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
mgao0: We will add some mandatory fields to the LockInfo once we identify what are needed for both blocking lock and nonblocking lock.

mgao0:

PR #706, opened 2020-01-29
Implement trie routing data

### LINKED ISSUE #701: Create TrieRoutingData - a trie structure representing Zk path sharding keys and associated realm addresses

NealSun96:
### Issues

Fixes #701 

### Description

This PR is a successor to PR #682 . 

As mentioned in the github issue, the original design of `ShardingKeyTrieNode` had numerous issues as revealed by the previous PR's comments, and therefore a new design is purposed and implemented. The previous design's has the following issues:

1. Since the trie node is exposed to outside users, extra effort is necessary to prevent misuse, such as making the children unmodifiable and make zkRealmAddress getter conditionally inaccessible. 
2. Implementation details of the trie node are not properly obscured. For any user to iterate through the trie, the user needs to get the children mapping of a node and continue forward. 
3. Traversal logic is not offered as at any common place and therefore can easily lead to repetition among different users.

With the new `TrieRoutingData` class, the trie node is encapsulated as a private class, therefore preventing any misuse. The class offers a sufficient but restrictive set of operations that outsider users may use, and can be expanded upon if necessary. The class implements the `RoutingData` interface, meaning that other forms of routing data can be implemented later to substitute `TrieRoutingData` if necessary. 

NOTE: this PR does not complete the `TrieRoutingData` class. The current constructor is a placeholder (as noted in code comments) because the actual constructor is dependent on a `RoutingDataAccessor` class which is yet to be implemented (but is second in order). Once the accessor is complemented the constructor will be written in a separate PR. 

### Tests

[INFO] Tests run: 97, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 19.733 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 97, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  23.970 s
[INFO] Finished at: 2020-01-29T13:30:43-08:00

### Commits

### Code Quality


helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * Using the root node, find the leaf node along the provided zkPath. Raise
+   * IllegalArgumentException if the zkPath does not point to any node or if there is no leaf node
+   * along the path.
+   * @param zkPath - the zkPath where the search is conducted
+   * @return the leaf node along the provided zkPath
+   * @throws IllegalArgumentException - when the zkPath points to nothing or when there is no leaf
+   *           node along the path.
+   */
+  private TrieNode findLeafTrieNodeAlongPath(String zkPath) throws IllegalArgumentException {
```
jiajunwang: This method is 90% the same as the findTrieNode, can we please merge them?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/RoutingData.java
```diff
@@ -0,0 +1,45 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.Map;
+
+public interface RoutingData {
```
jiajunwang: Shall we just call it ZkRealmData?
It might be confusing because we have routing table provider for completely different usage.

NealSun96: I don't have the context for the routing table provider, therefore I don't have any objection to that. Tagging @narendly for reference.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: I could set a flag to reuse the same logic, but I was concerned about the code being a bit messy. I'll write it up and see how it turns out.

NealSun96: I combined the two methods. It seems a bit long to me, but it does work.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/RoutingData.java
narendly: As we discussed, "MetadataStoreRoutingData" might be more appropriate here.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/RoutingData.java
```diff
@@ -0,0 +1,45 @@
...
+  Map<String, String> getAllMappingUnderPath(String zkPath);
+
+  /**
+   * Given a zkPath, return the zkRealmAddress corresponding to the sharding key contained in the
+   * zkPath. If the zkPath doesn't contain a sharding key, throw IllegalArgumentException.
+   * @param zkPath - the zkPath where the search is conducted
+   * @return the zkRealmAddress corresponding to the sharding key contained in the zkPath
+   * @throws IllegalArgumentException - when the zkPath doesn't contain a sharding key
+   */
+  String getZkRealm(String zkPath) throws IllegalArgumentException;
```
narendly: We could rename this to getMetadataStoreRealm to keep the interface generic.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/RoutingData.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+import java.util.Map;
+
+public interface RoutingData {
+  /**
+   * Given a zkPath, return all the "sharding key-zkRealmAddress" pairs where the sharding keys
+   * contain the given zkPath. For example, given "/ESPRESSO_MT_LD-1/schemata", return
+   * {"/ESPRESSO_MT_LD-1/schemata/ESPRESSO_MT_LD-1": "zk-ltx1-0-espresso.prod.linkedin.com:2181",
+   * "/ESPRESSO_MT_LD-1/schemata/ESPRESSO_MT_LD-2": "zk-ltx1-1-espresso.prod.linkedin.com:2181"}.
```
narendly: Please refrain from exposing internal details @nealsun-linkedin

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/RoutingData.java
```diff
@@ -0,0 +1,45 @@
+package org.apache.helix.rest.metadatastore;
```
narendly: This is an interface class; therefore, we want to keep it generic. This means that this interface class should be independent of ZK.
Also please refrain from using internal details in comments. This is an open source project.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,136 @@
...
+
+    Map<String, String> resultMap = new HashMap<>();
+    if (zkPath.substring(zkPath.length() - 1).equals("/")) {
+      zkPath = zkPath.substring(0, zkPath.length() - 1);
+    }
+    addAllAddressesToMapping(resultMap, curNode, zkPath);
+    return resultMap;
+  }
+
+  public String getZkRealm(String zkPath) throws IllegalArgumentException {
```
narendly: getMetadataStoreRealm since we want to make this metadata store-agnostic

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,136 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class TrieRoutingData implements RoutingData {
```
narendly: In general, be careful about terminology you use here. I'd say this class could be used for all metadata stores that use a filesystem API (assuming there are paths).

So your variables will have to be named something other than zkPath..

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,136 @@
...
+  // some input data. The constructor is blocked by the implementation of RoutingDataAccessor, and
+  // will therefore be implemented later.
+  public TrieRoutingData(TrieNode rootNode) {
+    _rootNode = rootNode;
+  }
+
+  static class TrieNode {
+    final Map<String, TrieNode> _children;
+    final boolean _isLeaf;
+    final String _realmAddress;
```
dasahcc: This is not complete realm address. Naming as it could be confusing.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,136 @@
...
+    } catch (IllegalArgumentException e) {
+      return Collections.emptyMap();
+    }
+
+    Map<String, String> resultMap = new HashMap<>();
+    if (path.substring(path.length() - 1).equals("/")) {
+      path = path.substring(0, path.length() - 1);
+    }
+    addAllAddressesToMapping(resultMap, curNode, path);
+    return resultMap;
```
dasahcc: To fast the processing, we can sacrifice some memory usage as storing all the sub children in current node.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,136 @@
...
+
+    if (path.substring(0, 1).equals("/")) {
+      path = path.substring(1);
+    }
+    String[] splitPath = path.split("/", 0);
+
+    TrieNode curNode = _rootNode;
+    if (findLeafAlongPath && curNode._isLeaf) {
+      return curNode;
+    }
```
dasahcc: This logic confuse me here. 

If you have trie:
     a - leaf
  b       c  -leaf
d   e - leaf

And you are looking for  /a/b with leaf flag as false. Then you will return root as /a. right?  But for searching purpose. I think it is correct to return b node. Can you explain what is expected scenario?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,136 @@
...
+   */
+  private static void addAllAddressesToMapping(Map<String, String> mapping, TrieNode curNode,
+      String curPath) {
+    if (curNode._isLeaf) {
+      mapping.put(curPath, curNode._realmAddress);
+      return;
+    }
+
+    for (Map.Entry<String, TrieNode> entry : curNode._children.entrySet()) {
+      addAllAddressesToMapping(mapping, entry.getValue(), curPath + "/" + entry.getKey());
```
dasahcc: Let's try to use iterative way instead of recursive call. You may take more memory in memory stack for function recursive call.

dasahcc: What if the path does not exist? Why we dont use a complete path to add to trie but with partial info trie node to add to the path? Logically, it is not correct.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Do you have any suggestion on the naming?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Github formatting might have ruined the trie you drew. 
Either way, if the leaf flag is false, then this if statement won't be triggered, and the method will search for `b` as it continues to the loop; if the leaf flag is true, it checks to see if `a` is a leaf, if yes then return. 
If this is not clear, feel free to discuss with me in person. The comment formatting is not helping.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: What I assume you mean is to store all descendant leaf nodes of a trie node as the children of a trie node. If the purposed change is a structural design change, then I'm afraid I have to disagree. 
The purposed structural change does allow a faster operation for `getAllMappingUnderPath`, however, it suffers from the following:
1. The trie will take longer to construct. Bear in mind that the trie is created on the fly by the major user, therefore construction cost is a realistic concern.
2. With the cost in mind, this change doesn't help with the other operation `getMetadataStoreRealm` at all.
3. A structural chagne like this makes the data structure no longer a trie. This is intended to be a trie class.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Two points are raised here which I'll try to break down: 
1. What if the path does not exist?
The path has to exist because it's traversing the previously constructed trie; whatever path is defined will be in the trie - this method is not creating arbitrary paths. I believe logically this method is correct. I have several unit tests to prove that it's working.
2. Why don't we store the complete path in the trie node?
I have considered this design before, and I believe storing complete paths also work. However, partial paths is working without an issue, therefore I see no significant benefit changing it. If you feel strongly about this, I'm happy to discuss further.

NealSun96: The recursive calls won't create many call stacks because the depth of the trie is typically limited: based on the current usage, the maximum is 3. The number of call stacks is bounded by the depth of the trie.
In terms of scaling in the future, I do agree that an iterative method is better, however, constructing the full paths for each node is more difficult than necessary. (This is related to your next comment - storing full paths resolves this problem. If you feel strongly about the full-path design, I'm happy to discuss further.)

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,136 @@
...
+   * @param curPath - the complete path of the current trie node
+   */
+  private static void addAllAddressesToMapping(Map<String, String> mapping, TrieNode curNode,
+      String curPath) {
+    if (curNode._isLeaf) {
+      mapping.put(curPath, curNode._realmAddress);
+      return;
+    }
+
+    for (Map.Entry<String, TrieNode> entry : curNode._children.entrySet()) {
```
narendly: Please use level-by-level traversal for this. We try to avoid recursion in production code.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,136 @@
...
+
+  // TODO: THIS IS A TEMPORARY PLACEHOLDER. A proper constructor will be created, which will not
+  // take in a TrieNode; it instead initializes the rootNode and creates a trie based on
+  // some input data. The constructor is blocked by the implementation of RoutingDataAccessor, and
+  // will therefore be implemented later.
+  public TrieRoutingData(TrieNode rootNode) {
+    _rootNode = rootNode;
+  }
+
+  static class TrieNode {
```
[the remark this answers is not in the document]
NealSun96: The class is declared as package private for testing purpose. Otherwise trie nodes cannot be created in the unit tests. 
It'll be moved to the end as suggested.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,136 @@
...
+    final String _zkRealmAddress;
+
+    TrieNode(Map<String, TrieNode> children, String name, boolean isLeaf, String zkRealmAddress) {
+      _children = children;
+      _isLeaf = isLeaf;
+      _zkRealmAddress = zkRealmAddress;
+    }
+  }
+
+  public Map<String, String> getAllMappingUnderPath(String zkPath) {
```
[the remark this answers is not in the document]
NealSun96: Please see the interface for JavaDoc of `getAllMappingUnderPath`. I looked at `CustomRestClient` and `CustomRestClientImpl` for reference and mimicked its style of having JavaDoc in interface but not implementation. If this isn't the correct convention, please feel free to correct me. 
The `zkPath` comment is on an outdated version of the files. The parameter has already been renamed.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: I experimented with "nodes storing full path" and "node storing partial path", and turns out the full path approach makes this method marginally better. I'm therefore adding a field that represents the full path of a trie node.

NealSun96: Update on point 2 (also replied on the previous comment): the "complete path" is marginally better for an iterative `getAllMappingUnderPath` therefore I added the full path field to `TrieNode`. Thanks for the suggestion @dasahcc . Leaving this conversation unresolved for point 1, point 2 has been resolved.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: The recursion code is now replaced with an iterative method based on the feedback from both @dasahcc and @narendly. Because of such a change, I'm storing the complete path of a trie node to the `name` field of a `TrieNode`, the reasons are as follows:
1. Storing complete paths in trie nodes makes `getAllMappingUnderPath` marginally better than storing partial paths. Without complete paths, the function has to either add/remove partial paths from a list as it operates a trie node stack, or store paths for all nodes as it operates the stack and waste memory; storing compete paths are faster, less wasteful, and simpler than both of these approaches. 
2. Trie nodes are usually seen to be associated with the complete prefix/strings they represent. Storing complete paths aligns with the common perception. 
Happy to discuss this further, but closing this conversion for now as the code is now iterative.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Per offline meeting, the confusion has been cleared up. Action point: adding javadoc to explain what this field actually means, but the field will remain.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Per offline meeting, the agreement is to keep the implementation as is for the current PR, and based on future usage statistics, potentially make optimization updates as suggested.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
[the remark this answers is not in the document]
NealSun96: I have reconsidered it and I find package private to be not secure enough. I will add a TODO on `TrieNode` because the unit tests are still dependent on it being package private; once the constructor is created, it will be changed to private.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,130 @@
...
+ * under the License.
+ */
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Stack;
+
+public class TrieRoutingData implements MetadataStoreRoutingData {
+  private final String DELIMITER = "/";
```
jiajunwang: private static final

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,130 @@
...
+      return Collections.emptyMap();
+    }
+
+    Map<String, String> resultMap = new HashMap<>();
+    Stack<TrieNode> nodeStack = new Stack<>();
+    nodeStack.push(curNode);
+    while (!nodeStack.isEmpty()) {
+      if (nodeStack.peek()._isLeaf) {
+        resultMap.put(nodeStack.peek()._name, nodeStack.peek()._realmAddress);
+        nodeStack.pop();
```
jiajunwang: nit, nodeStack.pop() can be done before check isLeaf()

NealSun96: Agreed.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Oops, thank you for catching it.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Per offline discussion, the confusion is led by the variable `_isLeaf`. `_isLeaf` is not using the trie sense, but the regular tree sense - it means if the node is a node without children. In a separate discussion, @narendly insisted that one variable shouldn't have two functionalities and we should not use children to dictate the node status, therefore `_isLeaf` is necessary. I have added javadoc to the variable to clarify the confusion.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreRoutingData.java
```diff
@@ -0,0 +1,44 @@
...
+  Map<String, String> getAllMappingUnderPath(String path);
+
+  /**
+   * Given a path, return the realm address corresponding to the sharding key contained in the
+   * path. If the path doesn't contain a sharding key, throw IllegalArgumentException.
+   * @param path - the path where the search is conducted
+   * @return the realm address corresponding to the sharding key contained in the path
+   * @throws IllegalArgumentException - when the path doesn't contain a sharding key
+   */
+  String getMetadataStoreRealm(String path) throws IllegalArgumentException;
```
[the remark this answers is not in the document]
NealSun96: I'll use `javassist.NotFoundException`, then.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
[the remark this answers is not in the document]
NealSun96: This has been addressed by @jiajunwang and is fixed.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,130 @@
...
+  public Map<String, String> getAllMappingUnderPath(String path) {
+    TrieNode curNode;
+    try {
+      curNode = findTrieNode(path, false);
+    } catch (IllegalArgumentException e) {
+      return Collections.emptyMap();
+    }
+
+    Map<String, String> resultMap = new HashMap<>();
+    Stack<TrieNode> nodeStack = new Stack<>();
```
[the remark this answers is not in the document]
NealSun96: I didn't realize that Stack is Vector based. Thanks for the suggestion, I'll switch it over.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreRoutingData.java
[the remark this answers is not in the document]
narendly: @NealSun96 Let's use https://docs.oracle.com/javase/8/docs/api/java/util/NoSuchElementException.html NoSuchElementException. This way, you don't have to import another library.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,130 @@
...
+   * not point to any node. If findLeafAlongPath is true, then starting from the root node, find the
+   * leaf node along the provided path; raise IllegalArgumentException if the path does not
+   * point to any node or if there is no leaf node along the path.
+   * @param path - the path where the search is conducted
+   * @param findLeafAlongPath - whether the search is for a leaf node on the path
+   * @return the node pointed by the path or a leaf node along the path
+   * @throws IllegalArgumentException - when the path points to nothing or when no leaf node is
+   *           found
+   */
+  private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
```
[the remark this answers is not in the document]
NealSun96: The two methods you mentioned were originally split out in this PR. It was @jiajunwang 's earlier comment that led to the merging of two methods (https://github.com/apache/helix/pull/706#discussion_r372655452). I don't have a strong opinion on either approach, but a general agreement should be made before I make any further changes. 
In terms of extracting common logic, the way these methods work prevent that: leaf checking logic happens on every level of iteration - if common code is extracted, the common code will each be two lines long and is not reasonable as standalone code.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreRoutingData.java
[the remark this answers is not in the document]
NealSun96: I agree that this is a good idea, but for simplicity let's keep it as string for this PR.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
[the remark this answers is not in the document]
narendly: If it's not easy, I'm okay with keeping it as one for now.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreRoutingData.java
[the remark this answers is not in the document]
NealSun96: Thank you for the suggestion. Updated.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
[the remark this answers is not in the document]
NealSun96: Both approaches are fine to me. Keeping it as is for now.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,157 @@
...
+   * @param findLeafAlongPath - whether the search is for a leaf node on the path
+   * @return the node pointed by the path or a leaf node along the path
+   * @throws NoSuchElementException - when the path points to nothing or when no leaf node is
+   *           found
+   */
+  private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
+      throws NoSuchElementException {
+    if (path.equals(DELIMITER) || path.equals("")) {
+      if (findLeafAlongPath && !_rootNode._isLeaf) {
+        throw new NoSuchElementException("no leaf node found along the path");
```
narendly: Nit: Try to make the error msg more helpful by adding the arguments (path and findLeafAlongPath)
example:
"No terminal node found along the path given! Path: " + path

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,157 @@
...
+     * This field is a mapping between trie key and children nodes. For example, node "a" has
+     * children "ab" and "ac", therefore the keys are "b" and "c" respectively.
+     */
+    Map<String, TrieNode> _children;
+    /**
+     * This field means if the node is a terminal node in the tree sense, not the trie sense. Any
+     * node that has children cannot possibly be a leaf node because only the node without children
+     * can store information.
+     */
+    final boolean _isLeaf;
```
narendly: Nit: did we agree to rename this to _isTerminalNode?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,157 @@
...
+
+    TrieNode curNode = _rootNode;
+    if (findLeafAlongPath && curNode._isLeaf) {
+      return curNode;
+    }
+    Map<String, TrieNode> curChildren = curNode._children;
+    for (String pathSection : splitPath) {
+      curNode = curChildren.get(pathSection);
+      if (curNode == null) {
+        throw new NoSuchElementException("the provided path is missing from the trie");
```
narendly: Let's make this errorMsg more helpful. See my comment above.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,157 @@
...
+      }
+      if (findLeafAlongPath && curNode._isLeaf) {
+        return curNode;
+      }
+      curChildren = curNode._children;
+    }
+    if (findLeafAlongPath) {
+      throw new NoSuchElementException("no leaf node found along the path");
```
narendly: See above.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+    return curNode;
+  }
+
+  // TODO: THE CLASS WILL BE CHANGED TO PRIVATE ONCE THE CONSTRUCTOR IS CREATED.
+  static class TrieNode {
+    /**
+     * This field is a mapping between trie key and children nodes. For example, node "a" has
+     * children "ab" and "ac", therefore the keys are "b" and "c" respectively.
```
narendly: This is not very clear?
Shouldn't it be : If a node "a" has children "b", and "c", then ....?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
narendly: Add here that if a node is terminal, it cannot have any children. This should be checked in this class as well.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,157 @@
...
+     * node that has children cannot possibly be a leaf node because only the node without children
+     * can store information.
+     */
+    final boolean _isLeaf;
+    /**
+     * This field aligns the traditional trie design: it entails the complete path/prefix leading to
+     * the current node. For example, the name of root node is "/", then the name of its child node
+     * is "/a", and the name of the child's child node is "/a/b".
+     */
+    final String _name;
```
narendly: Didn't we agree to this:

TrieNode : _name
root node: "/"
"a": "a"
"b": "b"

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Good point. Modifying test cases as well.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Resolved offline. See https://github.com/apache/helix/pull/706#discussion_r373123330 for additional reference.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Resolved offline. See https://github.com/apache/helix/pull/706#discussion_r373123330 for additional reference.

NealSun96:
Please remember to create the new branch for merging as you mentioned. 

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -0,0 +1,159 @@
...
+      }
+      return _rootNode;
+    }
+
+    String[] splitPath;
+    if (path.substring(0, 1).equals(DELIMITER)) {
+      splitPath = path.substring(1).split(DELIMITER, 0);
+    } else {
+      splitPath = path.substring(1).split(DELIMITER, 0);
+    }
```
narendly: Why is this if condition required if we're going to the same thing either way?

NealSun96: That was a mistake, thank you for catching it. I also added two more tests to ensure all if-paths are covered.

NealSun96:

NealSun96:
Final commit message:

Add MetadataStoreRoutingData interface and TrieRoutingData class to helix-rest

The TrieRoutingData class constructs a trie structure for "metadata store sharding keys - metadata store realm addressses" pairs to allow faster lookup operations. The MetadataStoreRoutingData interface allows generalized access to routing data.

PR #718, opened 2020-02-03
Implement Helix nonblocking lock

### LINKED ISSUE #708: Implement basic functions of Helix nonblocking lock based on ZooKeeper
Helix customer request a distributed lock that can help coordinate their work on same resource, so they won't step on each other. This issue is for request of nonblocking lock, that can immediate let customer know if a resource is involved in the process of a work. The first phase of the lock supports acquire, release, get info.

mgao0:
### Issues

Fixes #708 

### Description

This is a part of the Helix lock manager service. A HelixLock interface has been added and this PR is the implementation of nonblocking lock according to the interface. This nonblocking lock uses a single persistent znode to represent a lock on a resource. Upon calling acquireLock, the user either gets a "true" if they successfully acquired the lock meaning they can start their work on the resource, or a "false" if someone else currently has the lock meaning that they cannot work on this resource at this time. User can release the lock when they are done with their work, as well as get the lock information such as the current owner of the lock, any messages left by the lock owner, or the timeout of the lock.

### Tests

TestZKHelixNonblockingLock

[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 70.838 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:13 min
[INFO] Finished at: 2020-02-04T17:58:59-08:00

### Commits

### Documentation

### Code Quality


helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,159 @@
...
+  private final String _lockMsg;
+  private final BaseDataAccessor<ZNRecord> _baseDataAccessor;
+
+  /**
+   * Initialize the lock with user provided information, e.g.,cluster, scope, etc.
+   * @param clusterName the cluster under which the lock will live
+   * @param scope the scope to lock
+   * @param zkAddress the zk address the cluster connects to
+   * @param timeout the timeout period of the lcok
+   * @param lockMsg the reason for having this lock
```
zhangmeng916: suggest to change "userID" to "userId", and add the param of "userId" here.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,159 @@
...
+      Long timeout, String lockMsg, String userID) {
+    this("/" + clusterName + '/' + LOCK_ROOT + '/' + scope, zkAddress, timeout, lockMsg, userID);
+  }
+
+  /**
+   * Initialize the lock with user provided information, e.g., lock path under zookeeper, etc.
+   * @param lockPath the path of the lock under Zookeeper
+   * @param zkAddress the zk address of the cluster
+   * @param timeout the timeout period of the lcok
+   * @param lockMsg the reason for having this lock
```
zhangmeng916: same here, add "userId"

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,159 @@
...
+    _userID = userID;
+    _baseDataAccessor = new ZkBaseDataAccessor<ZNRecord>(zkClient.getServers());
+  }
+
+  @Override
+  /**
+   * Blocking call to acquire a lock
+   * @return true if the lock was successfully acquired,
+   * false if the lock could not be acquired
+   */ public boolean acquireLock() {
```
zhangmeng916: formatting?

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,159 @@
...
+
+    // If fail to create the lock node (acquire the lock), compare the timeout timestamp of current lock node with current time, if already passes the timeout, release current lock and try to acquire the lock again
+    if (!success) {
+      Stat stat = new Stat();
+      ZNRecord curLock = _baseDataAccessor.get(_lockPath, stat, AccessOption.PERSISTENT);
+      long curTimeout =
+          Long.parseLong(curLock.getSimpleField(ZKHelixNonblockingLockInfo.InfoKey.TIMEOUT.name()));
+      if (System.currentTimeMillis() >= curTimeout) {
+        success =
+            _baseDataAccessor.set(_lockPath, lockInfo, stat.getVersion(), AccessOption.PERSISTENT);
```
zhangmeng916: Better to add comment for this "set" operation, and when it may fail.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,159 @@
...
+   * Check if the user is current lock owner
+   * @return true if the user is the lock owner,
+   * false if the user is not the lock owner or the lock doesn't have a owner
+   */ public boolean isOwner() {
+    ZNRecord curLockInfo = _baseDataAccessor.get(_lockPath, null, AccessOption.PERSISTENT);
+    if (curLockInfo == null) {
+      return false;
+    }
+    String ownerID = curLockInfo.getSimpleField(ZKHelixNonblockingLockInfo.InfoKey.OWNER.name());
+    if (ownerID == null) {
```
zhangmeng916: Is ownerId null a valid scenario? In this case, the lock has no owner. You mean want to add some validation logic for lock info.
Also, suggest to change ownerID to ownerId.

helix-lock/src/test/java/org/apache/helix/lock/TestZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,176 @@
+/*
```
zhangmeng916: Could you also test the case when two threads both try to acquire the lock, independently, and one of them succeed, and the other fail?

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  /**
+   * Initialize the lock with user provided information, e.g., lock path under zookeeper, etc.
+   * @param lockPath the path of the lock under Zookeeper
+   * @param zkAddress the zk address of the cluster
+   * @param timeout the timeout period of the lcok
+   * @param lockMsg the reason for having this lock
+   */
+  public ZKHelixNonblockingLock(String lockPath, String zkAddress, Long timeout, String lockMsg,
```
jiajunwang: Discussed offline, if this one is not for our customer to use immediately, please use private.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,159 @@
...
+    return success;
+  }
+
+  @Override
+  /**
+   * Blocking call to release a lock
+   * @return true if the lock was successfully released,
+   * false if the locked is not locked or is not locked by the user,
+   * or the lock could not be released
+   */ public boolean releaseLock() {
```
jiajunwang: Format

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,159 @@
...
+    if (isOwner()) {
```
jiajunwang: It could be a race condition. We should never remove the node while releasing it. Simply update the timeout to -1 or so using the update(). So there won't be a chance someone releases the lock at the same time, and the lock is removed unexpectedly.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Helix nonblocking lock implementation based on Zookeeper
+ */
+public class ZKHelixNonblockingLock implements HelixLock {
+
+  private static final Logger LOG = Logger.getLogger(ZKHelixNonblockingLock.class);
+
+  private static final String LOCK_ROOT = "LOCKS";
+  private final String _lockPath;
+  private final String _userID;
```
jiajunwang: How we prevent 2 clients using the same ID?
I thing GUID is easier and they should not be specified by our user.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
```diff
@@ -0,0 +1,69 @@
...
+  public ZKHelixNonblockingLockInfo() {
+    lockInfo = new HashMap<>();
+  }
+
+  @Override
+  /**
+   * Create a single filed of LockInfo, or update the value of the field if it already exists
+   * @param infoKey the key of the LockInfo field
+   * @param infoValue the value of the LockInfo field
+   */ public void setInfoValue(String infoKey, String infoValue) {
```
jiajunwang: format

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
```diff
@@ -0,0 +1,69 @@
...
+    lockInfo.put(infoKey, infoValue);
+  }
+
+  @Override
+  /**
+   * Get the value of a field in LockInfo
+   * @param infoKey the key of the LockInfo field
+   * @return the value of the field or null if this key does not exist
+   */ public T getInfoValue(String infoKey) {
```
jiajunwang: getInforValue (InfoKey key)

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
jiajunwang: setInforValue (InfoKey key,  String infoValue)

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.helix.ZNRecord;
+
+
+public class ZKHelixNonblockingLockInfo<T extends String> implements LockInfo<T> {
+
+  private Map<String, String> lockInfo;
```
jiajunwang: Map<InfoKey, String>

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
```diff
@@ -0,0 +1,69 @@
...
+
+  enum InfoKey {
```
jiajunwang: Should be public if you want strong control on the types.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
```diff
@@ -0,0 +1,69 @@
...
+   * @return the value of the field or null if this key does not exist
+   */ public T getInfoValue(String infoKey) {
+    return (T) lockInfo.get(infoKey);
+  }
+
+  /**
+   * Update the lock info with information in a ZNRecord
+   * @param record Information about the lock that stored as ZNRecord format
+   */
+  public void setLockInfoFields(ZNRecord record) {
```
jiajunwang: What's the usage? The Helix convention is creating a copy constructor for this scenario.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,151 @@
...
+   * @param clusterName the cluster under which the lock will live
+   * @param scope the scope to lock
+   * @param zkAddress the zk address the cluster connects to
+   * @param timeout the timeout period of the lcok
+   * @param lockMsg the reason for having this lock
+   * @param userId a universal unique userId for lock owner identity
+   */
+  public ZKHelixNonblockingLock(String clusterName, HelixConfigScope scope, String zkAddress,
+      Long timeout, String lockMsg, String userId) {
+    this("/" + clusterName + '/' + LOCK_ROOT + '/' + scope, zkAddress, timeout, lockMsg, userId);
```
dasahcc: Let's make separator to be consistent either "/" or '/'. And it is better to have a constant for it.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,151 @@
...
+   * @param lockMsg the reason for having this lock
+   * @param userId a universal unique userId for lock owner identity
+   */
+  public ZKHelixNonblockingLock(String lockPath, String zkAddress, Long timeout, String lockMsg,
+      String userId) {
+    HelixZkClient zkClient = new ZkClient(zkAddress);
+    _lockPath = lockPath;
+    _timeout = timeout;
+    _lockMsg = lockMsg;
+    if (userId.matches("[a-f0-9]{8}-[a-f0-9]{4}-4[0-9]{3}-[89ab][a-f0-9]{3}-[0-9a-f]{12}")) {
```
dasahcc: Do not hard code it. Make it as a constant.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,151 @@
...
+      _userId = userId;
+    } else {
+      throw new IllegalArgumentException();
+    }
```
dasahcc: you can refactor code like:

if (!userId.matches(...)) {
   throw new IllegalArgumentException();
}
_userId = userId;

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,151 @@
...
+
+    // Try to create the lock node
+    boolean success = _baseDataAccessor.create(_lockPath, lockInfo, AccessOption.PERSISTENT);
+
+    // If fail to create the lock node (acquire the lock), compare the timeout timestamp of current lock node with current time, if already passes the timeout, release current lock and try to acquire the lock again
+    if (!success) {
+      Stat stat = new Stat();
+      ZNRecord curLock = _baseDataAccessor.get(_lockPath, stat, AccessOption.PERSISTENT);
+      long curTimeout =
+          Long.parseLong(curLock.getSimpleField(ZKHelixNonblockingLockInfo.InfoKey.TIMEOUT.name()));
```
dasahcc: Is this formatted?

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,151 @@
...
+    lockInfo
+        .setSimpleField(ZKHelixNonblockingLockInfo.InfoKey.TIMEOUT.name(), String.valueOf(timeout));
+
+    // Try to create the lock node
+    boolean success = _baseDataAccessor.create(_lockPath, lockInfo, AccessOption.PERSISTENT);
+
+    // If fail to create the lock node (acquire the lock), compare the timeout timestamp of current lock node with current time, if already passes the timeout, release current lock and try to acquire the lock again
+    if (!success) {
+      Stat stat = new Stat();
+      ZNRecord curLock = _baseDataAccessor.get(_lockPath, stat, AccessOption.PERSISTENT);
```
dasahcc: If you dont need to return Stat data, then you can pass null as argument. If you need it, set it in the ZNRecord.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: It is formatted

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: This is dangling comment, I deleted it since there is already a same comment in the interface.

helix-lock/src/test/java/org/apache/helix/lock/TestZKHelixNonblockingLock.java
mgao0: Test added. Please see testSimultaneousAcquire().

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: I changed the logic and now the user Id is validated in the constructor so there is no need to check if the ownerId is null, even if there is no current owner of the lock, the ownerId would be DEFAULT_OWNER_TEXT which is an empty string.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: Deleted the comment since it is duplicated with comments in interface

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: The purpose of this is to let user save a user Id locally, so even they lost session with Zk, after reconnect they will have the same user Id, so they can still check the lock owner and know if the lock belongs to them, as well as maintain authorizations to release their own lock.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
mgao0: I deleted this method, and created a copy constructor in ZKHelixNonblockingLockInfo to copy fields from ZNRecord.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: I created a constant for path delimiter. Please see updated version.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: Done.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: I use this Stat data to get the version number for the set operation later. It is one time use so i didn't set it in the ZNRecord.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: I changed the logic of release according to your suggestion to avoid the race condition. I created a LockUpdater which checks if the current user is the lock owner (so here I'm keeping the design of only lock owner can release a lock). If a non-owner tries to delete the lock, an exception will be thrown from the LockUpdater so when data accessor does update, it can return false.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,200 @@
...
+  public ZKHelixNonblockingLockInfo getLockInfo() {
+    if (!_baseDataAccessor.exists(_lockPath, AccessOption.PERSISTENT)) {
+      return new ZKHelixNonblockingLockInfo();
+    }
+    ZNRecord curLockInfo = _baseDataAccessor.get(_lockPath, null, AccessOption.PERSISTENT);
+    return new ZKHelixNonblockingLockInfo(curLockInfo);
+  }
+
+  @Override
+  public boolean isOwner() {
```
jiajunwang: I don't have a strong preference. But it might be easier for our customers if we return the remaining time instead of a boolean. If the lock owner changed, we just return -1.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,200 @@
...
+ * Helix nonblocking lock implementation based on Zookeeper
+ */
+public class ZKHelixNonblockingLock implements HelixLock {
+
+  private static final Logger LOG = Logger.getLogger(ZKHelixNonblockingLock.class);
+
+  private static final String LOCK_ROOT = "LOCKS";
+  private static final String PATH_DELIMITER = "/";
+  private static final String UUID_FORMAT_REGEX =
+      "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}";
```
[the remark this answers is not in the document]
mgao0: You are right, as long as we ask caller to provide user id themselves, we don't have control over it. But it works for our current use cases. So I'll remove the user id validation for now, since we don't want to make it harder for users to use.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: I personally prefer to have it return a boolean because it is more intuitive with the method name. But what you mentioned is a valid point, for now customers can call getLockInfo and get the expire time of the lock. If we see any future needs from customer, we can add a method to get the remaining time of the lock.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,200 @@
...
+    LockUpdater updater = new LockUpdater(newLockInfo);
+    return _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
+  }
+
+  @Override
+  public ZKHelixNonblockingLockInfo getLockInfo() {
+    if (!_baseDataAccessor.exists(_lockPath, AccessOption.PERSISTENT)) {
+      return new ZKHelixNonblockingLockInfo();
+    }
+    ZNRecord curLockInfo = _baseDataAccessor.get(_lockPath, null, AccessOption.PERSISTENT);
```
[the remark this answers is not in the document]
mgao0: I changed it to only call get once, and didn't use try catch. Instead, if the node doesn't exist, curLockInfo would be null, and I chose to handle it in the constructor of ZKHelixNonblockingLockInfo(ZNRecord).

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,200 @@
...
+  public boolean releaseLock() {
+    ZNRecord newLockInfo = new ZNRecord(_userId);
+    newLockInfo.setSimpleField(ZKHelixNonblockingLockInfo.InfoKey.OWNER.name(),
+        ZKHelixNonblockingLockInfo.DEFAULT_OWNER_TEXT);
+    newLockInfo.setSimpleField(ZKHelixNonblockingLockInfo.InfoKey.MESSAGE.name(),
+        ZKHelixNonblockingLockInfo.DEFAULT_MESSAGE_TEXT);
+    newLockInfo.setSimpleField(ZKHelixNonblockingLockInfo.InfoKey.TIMEOUT.name(),
+        ZKHelixNonblockingLockInfo.DEFAULT_TIMEOUT_TEXT);
+    LockUpdater updater = new LockUpdater(newLockInfo);
+    return _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
```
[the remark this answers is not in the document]
mgao0: Let's discuss about this further. 
In my current implementation, we allow a user to update a lock that currently has no owner or currently belongs to this person with the usage of the updater, data accessor will handle the NoNode or WrongVersion situations, and retry until the node exists and version is correct. And the release call will return a true.
However, after comparison we determine it is the correct version, and find out that a lock user try to update a lock that is currently held by someone else, the updater throws an exception which is to be caught by the data accessor, so data accessor will return an error retCode and return false immediately (no more retries) for the update call, but no exception will be thrown from the update call. 
So there shouldn't be any exceptions I could catch when calling the update with data accessor.

helix-lock/src/main/java/org/apache/helix/lock/HelixLockScope.java
```diff
@@ -0,0 +1,179 @@
...
+
+import java.util.List;
+
+import org.apache.helix.util.StringTemplate;
+
+
+/**
+ *  Defines the various scopes of Helix locks, and how they are represented on Zookeeper
+ */
+public class HelixLockScope {
```
jiajunwang: It could be cleaner if we have an abstract HelixScope class. Then there should be no duplicate code at all.
Then things we can put into the parent class:
- StringTemplate template
- clusterName
- zkPath
- type
- their get methods
- participant name is also something in common. But I think this one is not generic enough for the parent abstract class.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,198 @@
...
+
+
+/**
+ * Helix nonblocking lock implementation based on Zookeeper
+ */
+public class ZKHelixNonblockingLock implements HelixLock {
+
+  private static final Logger LOG = Logger.getLogger(ZKHelixNonblockingLock.class);
+
+  private static final String LOCK_ROOT = "/LOCK";
```
jiajunwang: Should this be part of the lock scope?

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+    lockInfo.setLongField(ZKHelixNonblockingLockInfo.InfoKey.TIMEOUT.name(), deadline);
+
+    LockUpdater updater = new LockUpdater(lockInfo);
+    return _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
+  }
+
+  @Override
+  public boolean releaseLock() {
+    ZNRecord newLockInfo = new ZKHelixNonblockingLockInfo<>().toZNRecord();
```
jiajunwang: Comment that the default constructor creates an unlocked lock.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /**
+   * Check if a lock has timed out
+   * @param record the current lock information in ZNRecord format
+   * @return return true if the lock has timed out, otherwise return false.
+   */
+  private boolean hasTimedOut(ZNRecord record) {
+    if (record == null) {
+      return false;
```
jiajunwang: Should be true? If there is no such lock node, we can assume timeout == true, right?

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,198 @@
...
+    _timeout = timeout;
+    _lockMsg = lockMsg;
+    _userId = userId;
+    _baseDataAccessor = new ZkBaseDataAccessor<>(zkAddress);
+  }
+
+  @Override
+  public boolean acquireLock() {
+
+    // Set lock information fields
```
jiajunwang: This logic should belong to the ZKHelixNonblockingLockInfo class. Considering in the near future, we want to add several more fields into ZKHelixNonblockingLockInfo. You will find we need to update ZKHelixNonblockingLockInfo class and this section.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
```diff
@@ -0,0 +1,92 @@
...
+    lockInfo.put(InfoKey.OWNER, DEFAULT_OWNER_TEXT);
+    lockInfo.put(InfoKey.MESSAGE, DEFAULT_MESSAGE_TEXT);
+    lockInfo.put(InfoKey.TIMEOUT, DEFAULT_TIMEOUT_TEXT);
+  }
+
+  /**
+   * Construct a ZKHelixNonblockingLockInfo using a ZNRecord format of data
+   * @param znRecord A ZNRecord that contains lock information in its simple fields
+   */
+  public ZKHelixNonblockingLockInfo(ZNRecord znRecord) {
```
jiajunwang: As I commented above, there should be a default constructor that builds instance with strongly typed parameters. And the other 2 constructors can call that one inside.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
```diff
@@ -0,0 +1,92 @@
...
+  private Map<InfoKey, String> lockInfo;
+
+  public enum InfoKey {
+    OWNER, MESSAGE, TIMEOUT
+  }
+
+  /**
+   * Constructor of ZKHelixNonblockingLockInfo that set each field to default data
+   */
+  public ZKHelixNonblockingLockInfo() {
```
jiajunwang: This could be confusing. I suggest having a static method for building unlocked lock. And you can keep an static final object internal, and we don't need to create new object every time.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
```diff
@@ -0,0 +1,92 @@
...
+
+package org.apache.helix.lock;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.helix.ZNRecord;
+
+
+public class ZKHelixNonblockingLockInfo<K extends ZKHelixNonblockingLockInfo.InfoKey, V extends String> implements LockInfo<K, V> {
```
jiajunwang: This class should extend HelixProperty? In that case, you have everything you need regarding ZnRecord.

helix-lock/src/main/java/org/apache/helix/lock/HelixLockScope.java
mgao0: You are right, it would be cleaner if we have an abstract class. But I decided not to do it at this time because 
1. HelixLockScope and HelixConfigScope don't share the same logic for most part, for example: HelixConfigScope takes a mapKey as a variable in the constructor while HelixLockScope does not, and these two classes don't define same zkPath argument number for each level so the getZkPath logic is different too. 
2. We want to decouple Helix Lock module from Helix core module, so in the future if needed we can freely extend the lock scope for other usages of the lock module.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLock.java
mgao0: Very good suggestion. I moved all the set lock info logic to LockInfo class.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
mgao0: I added the strongly typed parameter constructor. But currently I'm not calling this method using the other 2 constructors since I made the LockInfo class extend HelixProperty, and I used the super(). With super(), a ZNRecord is already created so we only need to update the values in the ZNRecord.

helix-lock/src/main/java/org/apache/helix/lock/ZKHelixNonblockingLockInfo.java
mgao0: Made it extend HelixPropery. I also handled it to ensure there are no null values in the LockInfo, so there are some methods or logics to set the null values in ZNRecord to the default values I defined for the LockInfo.

helix-lock/src/main/java/org/apache/helix/lock/HelixLockScope.java
mgao0: Discussed offline. Will make this LockScope more generic so user can define the properties themselves.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -19,28 +19,162 @@
...
+   */
+  public enum LockInfoAttribute {
+    OWNER, MESSAGE, TIMEOUT
+  }
+
+  /**
+   * Initialize a LockInfo with a ZNRecord id, set all info fields to default data
+   */
+  public LockInfo(String id) {
+    super(id);
```
jiajunwang: Should this one be the ownerId? Or what is the use case of this default constructor?

jiajunwang: Another option, just keep all lock node's ZNRecord id to be "LOCK"

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -19,28 +19,162 @@
...
+   */
+  public Long getTimeout() {
+    return _record.getLongField(LockInfoAttribute.TIMEOUT.name(), DEFAULT_TIMEOUT_LONG);
+  }
+
+  /**
+   * Get the value for OWNER attribute of the lock from a ZNRecord
+   * @return the owner id of the lock, empty string if there is no owner id set
+   */
+  public static String getOwner(ZNRecord znRecord) {
```
jiajunwang: Should accept LockInfo object.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -19,28 +19,162 @@
...
+    }
+    String owner = znRecord.getSimpleField(LockInfoAttribute.OWNER.name());
+    return owner == null ? DEFAULT_OWNER_TEXT : owner;
+  }
+
+  /**
+   * Get the value for MESSAGE attribute of the lock from a ZNRecord
+   * @return the message of the lock, empty string if there is no message set
+   */
+  public static String getMessage(ZNRecord znRecord) {
```
jiajunwang: Should accept LockInfo object.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -19,28 +19,162 @@
...
 
   /**
-   * Create a single filed of LockInfo, or update the value of the field if it already exists
-   * @param infoKey the key of the LockInfo field
-   * @param infoValue the value of the LockInfo field
+   * Get the value for TIMEOUT attribute of the lock from a ZNRecord
+   * @return the expiring time of the lock, -1 if there is no timeout set
    */
-  void setInfoValue(String infoKey, T infoValue);
+  public static long getTimeout(ZNRecord znRecord) {
```
jiajunwang: Should accept LockInfo object.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
jiajunwang: And then you can avoid the duplicate code by calling the instance method.

helix-lock/src/main/java/org/apache/helix/lock/helix/HelixLockScope.java
```diff
@@ -0,0 +1,183 @@
...
+   */
+  public HelixLockScope.LockScopeProperty getType() {
+    return _type;
+  }
+
+  /**
+   * Get the cluster name if it exists
+   * @return the cluster name
+   */
+  public String getClusterName() {
```
jiajunwang: If not necessary, let's don't adding these get methods for now.

helix-lock/src/main/java/org/apache/helix/lock/helix/HelixLockScope.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+
+  /**
+   * string templates to generate znode path
+   */
+  private static final StringTemplate template = new StringTemplate();
+
+  static {
+    template.addEntry(LockScopeProperty.CLUSTER, 1, "/{clusterName}/LOCK");
```
jiajunwang: This will be a shared path with the other scope types. I think it might cause some conflicts.

helix-lock/src/main/java/org/apache/helix/lock/helix/HelixLockScope.java
```diff
@@ -0,0 +1,183 @@
...
+    template.addEntry(HelixLockScope.LockScopeProperty.PARTICIPANT, 2,
+        "/{clusterName}/LOCK/{participantName}");
+    template.addEntry(HelixLockScope.LockScopeProperty.RESOURCE, 3,
+        "/{clusterName}/LOCK/{participantName}/{resourceName}");
```
jiajunwang: Can I lock the whole resource regardless of which participant holding it?

helix-lock/src/main/java/org/apache/helix/lock/helix/HelixLockScope.java
```diff
@@ -0,0 +1,183 @@
...
+    template.addEntry(HelixLockScope.LockScopeProperty.PARTITION, 4,
+        "/{clusterName}/LOCK/{participantName}/{resourceName}/{partitionName}");
```
jiajunwang: Same here, the participant should not be part of the path.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,186 @@
...
+    _baseDataAccessor = new ZkBaseDataAccessor<>(zkAddress);
+  }
+
+  @Override
+  public boolean acquireLock() {
+
+    // Set lock information fields
+    long deadline;
+    // Prevent value overflow
+    if (_timeout > Long.MAX_VALUE - System.currentTimeMillis()) {
```
jiajunwang: This is fine. But if timeout cannot be negative, we should check this in the constructor.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -19,28 +19,162 @@
...
+public class LockInfo extends HelixProperty {
 
-  //TODO: add specific setter and getter for any field that is determined to be universal for all implementations of HelixLock
+  // Default values for each attribute if there are no current values set by user
+  public static final String DEFAULT_OWNER_TEXT = "";
+  public static final String DEFAULT_MESSAGE_TEXT = "";
+  public static final long DEFAULT_TIMEOUT_LONG = -1L;
+
+  // default lock info represents the status of a unlocked lock
+  public static final LockInfo defaultLockInfo = new LockInfo("");
```
jiajunwang: This can be a private constructor if this is the only caller.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,186 @@
...
+
+  @Override
+  public boolean releaseLock() {
+    // Initialize the lock updater with a default lock info represents the state of a unlocked lock
+    LockUpdater updater = new LockUpdater(LockInfo.defaultLockInfo);
+    return _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
+  }
+
+  @Override
+  public LockInfo getLockInfo() {
```
jiajunwang: GetCurrentLockInfo? Since this method is not returning this instance's lock info.
Actually, keep it a static method could be cleaner.
Also for the following method isOwner().

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,186 @@
...
+  /**
+   * Check if a lock has timed out with lock information stored in a ZNRecord
+   * @return return true if the lock has timed out, otherwise return false.
+   */
+  private boolean hasTimedOut(ZNRecord znRecord) {
+    return System.currentTimeMillis() >= LockInfo.getTimeout(znRecord);
+  }
+
+  /**
+   * Check if the current user Id matches with the owner Id in a lock info
```
jiajunwang: Humm... These two are really confusing.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
mgao0: I has changed the ZNRecord id to "LOCK".

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKHelixNonblockingLock.java
mgao0: Added the check in the constructor.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKHelixNonblockingLock.java
mgao0: I tried to split methods into smaller logic so that it can be reused. But since now I'm only using once for each method, I guess this is not necessary to do that. I've rearranged the methods so it is less confusing.

helix-lock/src/main/java/org/apache/helix/lock/helix/HelixLockScope.java
mgao0: Yes, as discussed, I'll make it flat path for now.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKHelixNonblockingLock.java
mgao0: To call getCurrentLockInfo would need a lockPath defined. So here I will not change it to a static method.

helix-lock/src/main/java/org/apache/helix/lock/helix/HelixLockScope.java
mgao0: Changed it to the same format as path for cluster in HelixConfigScope, should not have conflicts now.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -19,28 +19,109 @@
...
+import org.apache.helix.ZNRecord;
 
 
 /**
- * Generic interface for a map contains the Helix lock information
- * @param <T> The type of the LockInfo value
+ * Structure represents a lock node information, implemented using ZNRecord
  */
-public interface LockInfo<T> {
+public class LockInfo extends HelixProperty {
```
dasahcc: Put a TODO here. This Helix Property will be replaced since distributed lock will be a module independent from Helix.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+    LockUpdater updater = new LockUpdater(new LockInfo(_userId, _lockMsg, deadline));
+    return _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
+  }
+
+  @Override
+  public boolean releaseLock() {
+    // Initialize the lock updater with a default lock info represents the state of a unlocked lock
+    LockUpdater updater = new LockUpdater(LockInfo.defaultLockInfo);
+    return _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
```
dasahcc: Why we dont delete the node instead of updating it?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKHelixNonblockingLock.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  @Override
+  public LockInfo getCurrentLockInfo() {
+    ZNRecord curLockInfo = _baseDataAccessor.get(_lockPath, null, AccessOption.PERSISTENT);
+    return new LockInfo(curLockInfo);
```
dasahcc: What if the path does not exists?

mgao0: Data accessor returns a null, and when converting it to the LockInfo, a LockInfo instance with all the fields filled with default values (which is also the value we use to represent an unlocked lock node) will be returned.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKHelixNonblockingLock.java
mgao0: Directly deleting it can cause race conditions. For example, when we call delete to delete lock node A, but during the process someone else acquired the lock or the current lock owner updated the lock, so currently it's lock node B, as a result the delete call deletes lock node B instead of lock node A, and both the caller and the lock owner wouldn't be aware of this mistake. The update method checks the version of the znode, so it can prevent this situation.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
mgao0: In this case I can remove the "extends HelixProperty" to make it independent from Helix. The LockInfo is barely a ZNRecord, and currently I'm not using any properties and methods exclusively provided by HelixProperty.

dasahcc: Sounds good.

helix-lock/src/main/java/org/apache/helix/lock/HelixLock.java
```diff
@@ -22,7 +22,7 @@
 /**
  * Generic interface for Helix distributed lock
  */
-public interface HelixLock<T> {
+public interface HelixLock {
```
dasahcc: Let's have the interface as DistributedLock or something. Name it HelixLock still seemed to be sticking to Helix.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -19,28 +19,106 @@
...
+  public static final LockInfo defaultLockInfo =
+      new LockInfo(DEFAULT_OWNER_TEXT, DEFAULT_MESSAGE_TEXT, DEFAULT_TIMEOUT_LONG);
+
+  private ZNRecord record;
+
+  /**
+   * The keys to lock information
+   */
+  public enum LockInfoAttribute {
+    OWNER, MESSAGE, TIMEOUT
```
dasahcc: Make one element per line.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -19,28 +19,106 @@
...
+  }
+
+  /**
+   * Initialize a default LockInfo instance
+   */
+  private LockInfo() {
+    record = new ZNRecord("LOCK");
```
dasahcc: Make "LOCK" to be a constant instead of hard coding here.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -19,28 +19,106 @@
...
+  // Default values for each attribute if there are no current values set by user
+  public static final String DEFAULT_OWNER_TEXT = "";
+  public static final String DEFAULT_MESSAGE_TEXT = "";
+  public static final long DEFAULT_TIMEOUT_LONG = -1L;
+
+  // default lock info represents the status of a unlocked lock
+  public static final LockInfo defaultLockInfo =
+      new LockInfo(DEFAULT_OWNER_TEXT, DEFAULT_MESSAGE_TEXT, DEFAULT_TIMEOUT_LONG);
+
+  private ZNRecord record;
```
dasahcc: naming convention: _record;

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKHelixNonblockingLock.java
mgao0: Keeping this way for now. Added TODO to change it in the future.

mgao0:
This PR is ready to be merged, approved by @dasahcc.
Final commit message: Implemented distributed nonblocking lock with acquire lock, release lock, get current lock info, and check is current user functionalities.

PR #714, opened 2020-02-01
Create data accessor interface and ZkRoutingDataAccessor

NealSun96:
### Issues

- [x] My PR addresses the following Helix issues and references them in the PR description:

Fixes #713  

### Description

- [x] Here are some details about my PR, including screenshots of any UI changes:

This PR adds an interface and an implementation for metadata store routing data accessing. The interface works directly with `MetadataStoreRoutingData`, acting as an abstract layer that allows the data to be accessed from different sources. The implementation focuses on accessing data from ZooKeeper. 

### Tests
- [x] The following is the result of the "mvn test" command on the appropriate module:

[INFO] Tests run: 100, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 20.581 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 100, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  25.231 s
[INFO] Finished at: 2020-01-31T17:53:37-08:00
[INFO] ------------------------------------------------------------------------

### Commits

- [x] My commits all reference appropriate Apache Helix GitHub issues in their subject lines, and I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Code Quality

- [x] My diff has been formatted using helix-style.xml

PR #720, opened 2020-02-04
Add MetadataStoreDirectory and ZkMetadataStoreDirectory

### LINKED ISSUE #721: Add MetadataStoreDirectory
MetadataStoreDirectory is an object that provides Metadata Store Directory APIs (routing APIs, CRUD of routing data, etc.). Helix REST will use this object to serve Metadata Store Directory Service REST endpoints.

narendly:
### Issues

Fixes #721

### Description

MetadataStoreDirectory is an object that provides Metadata Store Directory APIs (routing APIs, CRUD of routing data, etc.). Helix REST will use this object to serve Metadata Store Directory Service REST endpoints.
Also, it will make appropriate changes to the ZK access layer to listen on changes on the routing data.

Changelist:
- Refactor AbstractTestClass to make multi-ZK setup work
- Add implementation of MetadataStoreDirectory
- Add TestZkMetadataStoreDirectory

### Tests

TestZkMetadataStoreDirectory

helix-rest:

[INFO] BUILD SUCCESS
[INFO] Total time:  01:04 min
[INFO] Finished at: 2020-02-05T10:31:41-08:00

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
@@ -119,4 +119,9 @@ public interface MetadataStoreDirectory {
    * @return false if failed; true if the deletion is successful or the key does not exist.
    */
   boolean deleteShardingKey(String namespace, String realm, String shardingKey);
+
+  /**
+   * Updates the internally-cached routing data if available.
+   */
+  void updateRoutingData();
 }
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
@@ -21,6 +21,7 @@ package org.apache.helix.rest.metadatastore;
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.NoSuchElementException;
 
 
 /**
@@ -30,7 +31,7 @@ import java.util.Map;
  * realm: a metadata store deployable/ensemble. for example, if an application wishes to use 3 ZK quorums, then each ZK quorum would be considered a realm (ZK realm)
  * metadata store path sharding key: assuming the metadata store uses a file system APIs, this sharding key denotes the key that maps to a particular metadata store realm. an example of a key is a cluster name mapping to a particular ZK realm (ZK address)
  */
-public interface MetadataStoreDirectory {
+public interface MetadataStoreDirectory extends AutoCloseable {
 
   /**
    * Retrieves all existing namespaces in the routing metadata store.
@@ -74,24 +75,8 @@ public interface MetadataStoreDirectory {
    * @param shardingKey
    * @return
    */
-  String getMetadataStoreRealm(String namespace, String shardingKey);
-
-  /**
-   * NOTE: The following CRUD methods are idempotent.
-   */
-  /**
-   * Creates a namespace in the routing metadata store.
-   * @param namespace
-   * @return true if successful or if the namespace already exists. false if not successful.
-   */
-  boolean addNamespace(String namespace);
-
-  /**
-   * Deletes a namespace in the routing metadata store.
-   * @param namespace
-   * @return true if successful or if the namespace does not exist. false otherwise.
-   */
-  boolean deleteNamespace(String namespace);
+  String getMetadataStoreRealm(String namespace, String shardingKey)
+      throws NoSuchElementException;
 
   /**
    * Creates a realm. If the namespace does not exist, it creates one.
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
@@ -19,29 +19,36 @@ package org.apache.helix.rest.metadatastore;
  * under the License.
  */
 
-import java.util.List;
+import java.util.Collection;
 import java.util.Map;
 
 
+/**
+ * MetadataStoreDirectory interface that provides methods that are used to route requests to appropriate metadata store realm.
+ *
+ * namespace: tied to a namespace used in Helix REST (Metadata Store Directory Service endpoints will be served by Helix REST deployables)
+ * realm: a metadata store deployable/ensemble. for example, if an application wishes to use 3 ZK quorums, then each ZK quorum would be considered a realm (ZK realm)
+ * metadata store path sharding key: assuming the metadata store uses a file system APIs, this sharding key denotes the key that maps to a particular metadata store realm. an example of a key is a cluster name mapping to a particular ZK realm (ZK address)
+ */
 public interface MetadataStoreDirectory {
 
   /**
    * Retrieves all existing namespaces in the routing metadata store.
    * @return
    */
-  List<String> getAllNamespaces();
+  Collection<String> getAllNamespaces();
 
   /**
    * Returns all metadata store realms in the given namespace.
    * @return
    */
-  List<String> getAllMetadataStoreRealms(String namespace);
+  Collection<String> getAllMetadataStoreRealms(String namespace);
 
   /**
    * Returns all path-based sharding keys in the given namespace.
    * @return
    */
-  List<String> getAllShardingKeys(String namespace);
+  Collection<String> getAllShardingKeys(String namespace);
 
   /**
    * Returns all path-based sharding keys in the given namespace and the realm.
@@ -49,17 +56,17 @@ public interface MetadataStoreDirectory {
    * @param realm
    * @return
    */
-  List<String> getAllShardingKeysInRealm(String namespace, String realm);
+  Collection<String> getAllShardingKeysInRealm(String namespace, String realm);
 
   /**
    * Returns all sharding keys that have the given path as the prefix substring.
    * E.g) Given that there are sharding keys: /a/b/c, /a/b/d, /a/e,
-   * getAllShardingKeysUnderPath(namespace, "/a/b") returns ["/a/b/c", "/a/b/d"].
+   * getAllShardingKeysUnderPath(namespace, "/a/b") returns ["/a/b/c": "realm", "/a/b/d": "realm].
    * @param namespace
    * @param path
    * @return
    */
-  Map<String, String> getAllShardingKeysUnderPath(String namespace, String path);
+  Map<String, String> getAllMappingUnderPath(String namespace, String path);
 
   /**
    * Returns the name of the metadata store realm based on the namespace and the sharding key given.
@@ -121,7 +128,7 @@ public interface MetadataStoreDirectory {
   boolean deleteShardingKey(String namespace, String realm, String shardingKey);
 
   /**
-   * Updates the internally-cached routing data if available.
+   * Close MetadataStoreDirectory.
    */
-  void updateRoutingData();
+  void close();
 }
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
@@ -1,0 +1,122 @@
+package org.apache.helix.rest.metadatastore;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.List;
+import java.util.Map;
+
+
+public interface MetadataStoreDirectory {
+
+  /**
+   * Retrieves all existing namespaces in the routing metadata store.
+   * @return
+   */
+  List<String> getAllNamespaces();
+
+  /**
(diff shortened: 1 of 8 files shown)


helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
```diff
@@ -0,0 +1,127 @@
...
+   */
+  List<String> getAllShardingKeys(String namespace);
+
+  /**
+   * Returns all path-based sharding keys in the given namespace and the realm.
+   * @param namespace
+   * @param realm
+   * @return
+   */
+  List<String> getAllShardingKeysInRealm(String namespace, String realm);
```
NealSun96: Responding to a point raised by @pkuwm earlier during the design phase, the naming of this method should be changed to `getAllMappingUnderPath`. See https://github.com/apache/helix/pull/706/files#diff-184698fcbd1600d535a7b9f7996ca9f6R37 for reference as it has been changed for the underlying class. 
The reason behind this proposal is that `getAllShardingKeysInRealm` returns a list of sharding keys, while this method returns a mapping of sharding keys to realm addresses. If this method has the same prefix of "getAllShardkingKeys" it's easy to cause confusion on the usage.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -0,0 +1,125 @@
...
+    return null;
+  }
+
+  @Override
+  public List<String> getAllShardingKeysInRealm(String namespace, String realm) {
+    return null;
+  }
+
+  @Override
+  public Map<String, String> getAllShardingKeysUnderPath(String namespace, String path) {
```
NealSun96: Ditto the point above.

NealSun96:
As I recall from the design discussion, there should be a method that returns all the "sharding key - realm address" mapping. This method will be used to transfer all routing data across REST so that the routing data cache can be constructed from it (without accessing ZK for data).
For this purpose, we can use the "get mapping under path" method and pass in "/" for the path, but that will be using one method for two purposes.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+   * Returns all sharding keys that have the given path as the prefix substring.
+   * E.g) Given that there are sharding keys: /a/b/c, /a/b/d, /a/e,
+   * getAllShardingKeysUnderPath(namespace, "/a/b") returns ["/a/b/c", "/a/b/d"].
+   * @param namespace
+   * @param path
+   * @return
+   */
+  Map<String, String> getAllShardingKeysUnderPath(String namespace, String path);
```
NealSun96: Responding to a point raised by @pkuwm earlier during the design phase, the naming of this method should be changed to `getAllMappingUnderPath`. See https://github.com/apache/helix/pull/706/files#diff-184698fcbd1600d535a7b9f7996ca9f6R37 for reference as it has been changed for the underlying class.
The reason behind this proposal is that `getAllShardingKeysInRealm` returns a list of sharding keys, while this method returns a mapping of sharding keys to realm addresses. If this method has the same prefix of "getAllShardkingKeys" it's easy to cause confusion on the usage.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -0,0 +1,125 @@
...
+//   */
+//  private MetadataStoreRoutingData constructRoutingData(MetadataStoreRoutingDataAccessor routingDataAccessor) {
+//    /**
+//     * Trie construction logic
+//     */
+//
+//    // 1. Construct
+//    // 2. Update the in-memory reference when complete
+//    _routingData = newRoutingData;
+//  }
```
dasahcc: Do we need this?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -0,0 +1,125 @@
...
+
+public class ZkMetadataStoreDirectory implements MetadataStoreDirectory {
+  // TODO: enable the line below when implementation is complete
+//  private final MetadataStoreRoutingDataAccessor _routingDataAccessor;
+  private MetadataStoreRoutingData _routingData;
+
+  public ZkMetadataStoreDirectory(String zkAddress) {
+//    _routingDataAccessor = new ZkRoutingDataAccessor(zkAddress, this); // dependency injection
+    // 1. Create a RoutingDataTrieNode using routingDataAccessor
+    // _routingData = constructRoutingData(_routingDataAccessor);
```
dasahcc: If the code is not used, let's remove it.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
```diff
@@ -0,0 +1,127 @@
...
+   * Returns all metadata store realms in the given namespace.
+   * @return
+   */
+  List<String> getAllMetadataStoreRealms(String namespace);
+
+  /**
+   * Returns all path-based sharding keys in the given namespace.
+   * @return
+   */
+  List<String> getAllShardingKeys(String namespace);
```
dasahcc: Will this return complete path of keys?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
```diff
@@ -0,0 +1,127 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.List;
+import java.util.Map;
+
+
+public interface MetadataStoreDirectory {
```
dasahcc: It is better in the Interface level to explain the namespace, realms and sharding keys. Reviewers may not have context on what are they.

narendly:
@all

This PR is a draft and is not ready for review. This depends on https://github.com/apache/helix/pull/714. #714

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
narendly: Added more explanation.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
narendly: Yes. All sharding keys in the namespace.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
narendly: Modified.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
```diff
@@ -0,0 +1,134 @@
...
+   */
+  Map<String, String> getAllMappingUnderPath(String namespace, String path);
+
+  /**
+   * Returns the name of the metadata store realm based on the namespace and the sharding key given.
+   * @param namespace
+   * @param shardingKey
+   * @return
+   */
+  String getMetadataStoreRealm(String namespace, String shardingKey);
```
dasahcc: What if it does not exist? Shall we return null String or throw exception? If throw exception, we should have signature here.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * NOTE: The following CRUD methods are idempotent.
+   */
+  /**
+   * Creates a namespace in the routing metadata store.
+   * @param namespace
+   * @return true if successful or if the namespace already exists. false if not successful.
+   */
+  boolean addNamespace(String namespace);
```
dasahcc: This is not like to be a REST level operation, right? Does REST support this operation now? It would be dangerous that MDSD support add/delete namespaces.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
```diff
@@ -0,0 +1,134 @@
...
+   * @param realm
+   * @param shardingKey
+   * @return false if failed; true if the deletion is successful or the key does not exist.
+   */
+  boolean deleteShardingKey(String namespace, String realm, String shardingKey);
+
+  /**
+   * Close MetadataStoreDirectory.
+   */
+  void close();
```
dasahcc: Do we consider close operation is idempotent? If yes, we can think about extend the interface AutoCloseable. Then user can use try-with-resources.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
narendly: I'd like to understand the context of your questions -
1. Why do you think supporting write operations is dangerous?
2. This will be supported; otherwise, users won't have a safe way to register namespaces and realms. We should add a layer of security (access control), but that would be the roadmap for the future.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
narendly: Using AutoCloseable could be a good idea. I'll consider adding it.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
narendly: If it doesn't exist, yes, we should alert the user with an exception. I'll consider adding that.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -0,0 +1,199 @@
...
+   * The global consistency of the in-memory routing data is not a requirement (eventual consistency is enough).
+   * @param namespace
+   */
+  @Override
+  public void updateRoutingData(String namespace) {
+    // Safe to ignore the callback if routingDataMap is null.
+    // If routingDataMap is null, then it will be populated by the constructor anyway
+    // If routingDataMap is not null, then it's safe for the callback function to update it
+    try {
+      _realmToShardingKeysMap.put(namespace, _routingDataReaderMap.get(namespace).getRoutingData());
```
dasahcc: What if this namespace key does not exists? Could NPE happens?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  @Override
+  public boolean deleteShardingKey(String namespace, String realm, String shardingKey) {
+    // TODO implement when MetadataStoreRoutingDataWriter is ready
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Callback for updating the cached routing data.
```
dasahcc: This is more like an internal method to refresh internal cached data? Then refreshRoutingData could be a better name. Otherwise, it will be confused with users operation.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkRoutingDataReader.java
```diff
@@ -22,54 +22,161 @@
...
+    _routingDataListener.updateRoutingData(_namespace);
+  }
+
+  @Override
+  public synchronized void handleDataDeleted(String s)
+      throws Exception {
+    if (_zkClient.isClosed()) {
+      return;
+    }
+    _zkClient.unsubscribeAll();
```
dasahcc: Can you share why we need to unsubscribe all here?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
dasahcc: I think I was confused with existing REST namespace? Because I saw you are subscribing to namespace. Will that be a ZNode?

narendly: @dasahcc I think I see what you're referring to. This was an old design that was copied and pasted. Let me remove these APIs around namespaces. Namespaces will be registered via application-supplied configs.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
narendly: I can change the method name.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
narendly: Adding the exists check with a log.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkRoutingDataReader.java
narendly: This is meant to go under handleChildChange. Good catch!

narendly:
This PR is ready to be merged, approved by @dasahcc .

Add MetadataStoreDirectory and ZkMetadataStoreDirectory
---
MetadataStoreDirectory is an object that provides Metadata Store Directory APIs (routing APIs, CRUD of routing data, etc.). Helix REST will use this object to serve Metadata Store Directory Service REST endpoints.
Also, it will make appropriate changes to the ZK access layer to listen on changes on the routing data.

Changelist:
Refactor AbstractTestClass to make multi-ZK setup work
Add implementation of MetadataStoreDirectory
Add TestZkMetadataStoreDirectory

PR #727, opened 2020-02-06
Add MetadataStoreRoutingDataWriter with DistributedLeaderElection

### LINKED ISSUE #725: Add MetadataStoreRoutingDataWriter
MetadataStoreDirectory in Helix API Service(helix-rest) will not only be providing read functionality but it will also provide write APIs. Because of the way it interacts with ZooKeeper and in order to keep the data consistent with what's in the metadata store, we need to have a separate MetadataStoreRoutingDataWriter in addition to Reader. This issue tracks this work.

narendly:
### Issues

Resolves #725 

### Description

We need a separate ZkClient-based writer that could allow users to write routing data to ZK. This diff adds such an interface, an implementation, and a distributed lock implementation that could help users to manipulate the routing data.

Changelist:
1. Add ZkRoutingDataWriter (+ interface)
2. Add ZkDistributedLock (+ interface) to guarantee that there's at most one active writer at a time (where there are multiple Helix REST deployables)
3. Add a test for ZkRoutingDataWriter
4. Integrate ZkRoutingDataWriter with ZkMetadataStoreDirectory
5. Add test methods to TestZkMetadataStoreDirectory
6. Add ZkDistributedElection to replace ZkDistributedLock (and move ZkDistributedLock to a separate PR)

### Tests

TestZkRoutingDataWriter

> Shut down zookeeper at port 2124 in thread main
> Shut down zookeeper at port 2124 in thread main
> Shut down zookeeper at port 2123 in thread main
> [INFO] Tests run: 117, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 21.253 s - in TestSuite
> [INFO] 
> [INFO] Results:
> [INFO] 
> [INFO] Tests run: 117, Failures: 0, Errors: 0, Skipped: 0
> [INFO] 
> [INFO] 
> [INFO] --- maven-surefire-plugin:3.0.0-M3:test (multi-zk) @ helix-rest ---
> [INFO] 
> [INFO] -------------------------------------------------------
...

> End test :testInvalidHelixRestServerInitialization
> Shut down zookeeper at port 2123 in thread main
> Shut down zookeeper at port 2124 in thread main
> Shut down zookeeper at port 2124 in thread main
> Shut down zookeeper at port 2125 in thread main
> Shut down zookeeper at port 2123 in thread main
> [INFO] Tests run: 117, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 24.038 s - in TestSuite
> [INFO] 
> [INFO] Results:
> [INFO] 
> [INFO] Tests run: 117, Failures: 0, Errors: 0, Skipped: 0
> [INFO] 
> [INFO] BUILD SUCCESS
> [INFO] Total time:  51.877 s
> [INFO] Finished at: 2020-02-05T21:14:05-08:00

----
helix-rest run:

> [INFO] Tests run: 117, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 20.951 s - in TestSuite
> [INFO] 
> [INFO] Results:
> [INFO] 
> [INFO] Tests run: 117, Failures: 0, Errors: 0, Skipped: 0
> [INFO] 
> [INFO] BUILD SUCCESS
> [INFO] Total time:  47.479 s
> [INFO] Finished at: 2020-02-09T16:28:51-08:00
> [INFO] ----------------------------

Newest run:

> Shut down zookeeper at port 2124 in thread main
> Shut down zookeeper at port 2125 in thread main
> Shut down zookeeper at port 2123 in thread main
> [INFO] Tests run: 127, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 20.768 s - in TestSuite
> [INFO] 
> [INFO] Results:
> [INFO] 
> [INFO] Tests run: 127, Failures: 0, Errors: 0, Skipped: 0
> [INFO] 
> [INFO] BUILD SUCCESS
> [INFO] Total time:  47.168 s
> [INFO] Finished at: 2020-02-10T16:56:16-08:00

### Commits

### Code Quality


helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -0,0 +1,265 @@
...
+import org.slf4j.LoggerFactory;
+
+
+public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
+  private static final Logger LOG = LoggerFactory.getLogger(ZkBaseDataAccessor.class);
+
+  private final String _namespace;
+  private final String _zkAddress;
+  private final HelixZkClient _zkClient;
+  private final DistributedLock _routingDataLock;
```
dasahcc: Fundamentally, I think this is not an efficient way for purpose of control of centralized work. This may increase the traffic to ZK. 

It will be better to work as Helix leader controller. One of the REST server become write REST server (acquire the lock once). Write request to read-only server forward the request to write server for write purpose.

narendly: @dasahcc 
I agree that making Helix API service (Helix REST) an Helix application is a clever solution, but I have some concerns about that approach - wouldn't we be creating a circular dependency?

Also, there was an agreement on this approach because we have an assumption that write operations are rare. Ideally, all operations provided by the writer would be ACL-ed and will only be done by admin. In that sense, I believe this is an appropriate approach since it is simple, self-contained without any other dependency (say Helix controller), and it follows a distributed lock recipe that ZooKeeper is very well-known and tested for.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -0,0 +1,256 @@
...
+    // Ensure that ROUTING_DATA_PATH exists in ZK. If not, create
+    // create() semantic will fail if it already exists
+    try {
+      _zkClient.createPersistent(MetadataStoreRoutingConstants.ROUTING_DATA_PATH);
+    } catch (ZkNodeExistsException e) {
+      // This is okay
+    }
+
+    // Get the hostname (REST endpoint) from System property
+    ZNRecord myServerInfo = new ZNRecord("dummy hostname");
```
mgao0: Did you mean to get the hostname here instead of using the dummy hostname?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  @Override
+  public synchronized boolean addShardingKey(String realm, String shardingKey) {
+    if (_leaderElection.isLeader()) {
+      if (_zkClient.isClosed()) {
+        throw new IllegalStateException("ZkClient is closed!");
+      }
+      // If the realm does not exist already, then create the realm
+      if (!_zkClient.exists(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm)) {
```
mgao0: Can you create a variable for the path so you don't need to repeatedly type it out?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: This item is a TODO: I'll have to get this from Systems property. Dummy is good for now.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: Will do.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -0,0 +1,256 @@
...
+      // For each ZkRealm, write the given routing data to ZooKeeper
+      for (Map.Entry<String, List<String>> routingDataEntry : routingData.entrySet()) {
+        String zkRealm = routingDataEntry.getKey();
+        List<String> shardingKeyList = routingDataEntry.getValue();
+
+        ZNRecord znRecord = new ZNRecord(zkRealm);
+        znRecord
+            .setListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY, shardingKeyList);
+        try {
+          if (!_zkClient.exists(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + zkRealm)) {
```
mgao0: try catch so no need to check if the path exists?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -0,0 +1,256 @@
...
+      if (_zkClient.isClosed()) {
+        throw new IllegalStateException("ZkClient is closed!");
+      }
+      if (routingData == null) {
+        throw new IllegalArgumentException("routingData given is null!");
+      }
+
+      // Remove existing routing data
+      for (String zkRealm : _zkClient
+          .getChildren(MetadataStoreRoutingConstants.ROUTING_DATA_PATH)) {
```
NealSun96: Nit: can we recursively delete `MetadataStoreRoutingConstants.ROUTING_DATA_PATH` here?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: I believe @narendly 's logic here is to not write on top of an existing node, therefore the if statement is necessary even with try catch. 
However, this check is unnecessary for another reason: this is a synchronized function and each loop of `routingData` guarantees a different `zkRealm`, therefore the node will never already exist here. Correct me if I'm wrong, please.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.helix.ZNRecord;
+import org.apache.helix.rest.metadatastore.constant.MetadataStoreRoutingConstants;
+import org.apache.helix.rest.server.AbstractTestClass;
+import org.junit.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+
+public class TestZkRoutingDataWriter extends AbstractTestClass {
```
NealSun96: IMO the test cases are not complete. Each of the test cases should at least cover both true and false situations. The test for `setRoutingData` needs to make sure it overwrites.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: Deletion of MetadataStoreRoutingConstants.ROUTING_DATA_PATH is not necessary, and it could cause ZkNoNodeException.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: It's not completely necessary, but it's better to be safe since exists() calls are relatively cheap. I'd rather have an exist() call here than risk getting a ZkNoNodeException.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
narendly: setRoutingData now checks for strict equality.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: Update: using distributed leader election.

NealSun96:
A generic comment on all writer logics: some deletion operations can lead to a data state that is deemed "invalid" by `ZkRoutingDataReader`. For example, removing the last realmAddress, or removing the last sharding key for a realmAddress.
Questions:
1. Do we allow that to happen? 
2. If we don't allow that to happen, should the logic for checking these situations be a part of this PR, or a part of the validation logic PR?

narendly:
> A generic comment on all writer logics: some deletion operations can lead to a data state that is deemed "invalid" by `ZkRoutingDataReader`. For example, removing the last realmAddress, or removing the last sharding key for a realmAddress.
> Questions:
> 
> 1. Do we allow that to happen?
> 2. If we don't allow that to happen, should the logic for checking these situations be a part of this PR, or a part of the validation logic PR?

Could you include some specific examples? We could do either, but since other changes might be dependent on this diff, we could create an issue with your specific TODOs so that we could include that change in other PRs. @NealSun96

NealSun96:
> > A generic comment on all writer logics: some deletion operations can lead to a data state that is deemed "invalid" by `ZkRoutingDataReader`. For example, removing the last realmAddress, or removing the last sharding key for a realmAddress.
> > Questions:
> > 
> > 1. Do we allow that to happen?
> > 2. If we don't allow that to happen, should the logic for checking these situations be a part of this PR, or a part of the validation logic PR?
> 
> Could you include some specific examples? We could do either, but since other changes might be dependent on this diff, we could create an issue with your specific TODOs so that we could include that change in other PRs. @NealSun96

Examples: 
1. "realmAddress1" is the only realm address for namespace "x", and the user wants to call `deleteMetadataStoreRealm("realmAddress1")`.
2. "/a/b/c" is the only sharding key for "realmAddressX", and the user wants to call `deleteShardingKey("realmAddressX", "/a/b/c")` which will leave the list field with 0 sharding key. 
We still need to think about the question 1 I posted in my first comment: do we allow this to happen? From the operations alone, they don't seem "illegal" - users should be allowed to delete things regardless if they are the last elements. But if they are allowed to do that, `ZkRoutingDataReader` will throw exceptions as it's attempting to read because it will complain "There are no metadata store realms defined" for case 1 and "Realm address does not have a value" for case 2. 
In my opinion, users should be allowed to do that. When the last element is removed, the deletion should propagate up.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -0,0 +1,253 @@
...
+    } else {
+      try {
+        _zkClient.createPersistent(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm);
+        _zkClient.writeData(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm,
+            new ZNRecord(realm));
+      } catch (Exception e) {
+        LOG.error("Failed to create ZkRealm: {}, Namespace: ", realm, _namespace);
+        return false;
+      }
+    }
```
dasahcc: else clause may not need for this case.

narendly: Will remove.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -0,0 +1,253 @@
...
+      znRecord.setListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY,
+          Collections.singletonList(shardingKey));
+      try {
+        _zkClient.writeData(realmPath, znRecord);
+      } catch (Exception e) {
+        LOG.error(
+            "Failed to write the realm ZNRecord in addShardingKey()! Namespace: {}, Realm: {}, ShardingKey: {}",
+            _namespace, realm, shardingKey, e);
+        return false;
+      }
```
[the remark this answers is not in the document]
narendly: This was considered, but raw zookeeper is not very clear about the kind of exceptions it throws (it throws Exception).
Note that as soon as we encounter an exception, we need to return and should not proceed. Otherwise, we might risk getting an NPE.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/concurrency/ZkDistributedLeaderElection.java
```diff
@@ -0,0 +1,142 @@
...
+
+    if (leaderName.equals(myName)) {
+      // My turn for leadership
+      _isLeader = true;
+      _currentLeaderInfo = leaderInfo;
+      LOG.info("{} acquired leadership! Info: {}", myName, leaderInfo);
+    } else {
+      // Watch the ephemeral ZNode before me for a deletion event
+      String beforeMe = children.get(children.indexOf(myName) - 1);
+      _zkClient.subscribeDataChanges(_basePath + "/" + beforeMe, this);
```
dasahcc: Should this be state change? There should not be any case of ZNode data change, right?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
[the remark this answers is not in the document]
dasahcc: Then let's have a todo here. Since we are working the ZKClient module. Let's have clear different exceptions for different errors. It will benefit the coding convention with that.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/concurrency/ZkDistributedLeaderElection.java
```diff
@@ -0,0 +1,142 @@
...
+    return;
+  }
+
+  @Override
+  public void handleDataChange(String s, Object o) {
+    return;
+  }
+
+  @Override
+  public void handleDataDeleted(String s) {
```
dasahcc: Was data delete referring the content of ZNode instead of ZNode it self?

narendly: I'm not sure what you're asking about exactly...

>   private void fireDataChangedEvents(final String path, Set<IZkDataListenerEntry> listeners,
>       final OptionalLong notificationTime) {
>     try {
>       final ZkPathStatRecord pathStatRecord = new ZkPathStatRecord(path);
>       // Trigger listener callbacks
>       for (final IZkDataListenerEntry listener : listeners) {
>         _eventThread.send(new ZkEvent("Data of " + path + " changed sent to "
>             + listener.getDataListener() + " prefetch data: " + listener.isPrefetchData()) {
>           @Override
>           public void run() throws Exception {
>             // Reinstall watch before listener callbacks to check the znode status
>             if (!pathStatRecord.pathChecked()) {
>               pathStatRecord.recordPathStat(getStat(path, true), notificationTime);
>             }
>             if (!pathStatRecord.pathExists()) {
>               // no znode found at the path, trigger data deleted handler.
>               listener.getDataListener().handleDataDeleted(path);
>             } else {
>               Object data = null;
>               if (listener.isPrefetchData()) {
>                 if (LOG.isDebugEnabled()) {
>                   LOG.debug("Prefetch data for path: {}", path);
>                 }
>                 try {
>                   data = readData(path, null, true);
>                 } catch (ZkNoNodeException e) {
>                   LOG.warn("Prefetch data for path: {} failed.", path, e);
>                   listener.getDataListener().handleDataDeleted(path);
>                   return;
>                 }
>               }
>               listener.getDataListener().handleDataChange(path, data);
>             }
>           }
>         });
>       }

Does this help?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/concurrency/ZkDistributedLeaderElection.java
```diff
@@ -0,0 +1,142 @@
...
+   */
+  private void init() {
+    try {
+      _zkClient.createPersistent(_basePath, true);
+    } catch (ZkNodeExistsException e) {
+      // Okay if it exists already
+    }
+
+    // Create my ephemeral sequential node with my information
+    _myEphemeralSequentialPath = _zkClient
```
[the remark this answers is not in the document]
narendly: What do you mean by storing structure?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/concurrency/ZkDistributedLeaderElection.java
```diff
@@ -0,0 +1,142 @@
...
+
+  private final HelixZkClient _zkClient;
+  private final String _basePath;
+  private final ZNRecord _participantInfo;
+  private ZNRecord _currentLeaderInfo;
+
+  private String _myEphemeralSequentialPath;
+  private volatile boolean _isLeader;
+
+  public ZkDistributedLeaderElection(HelixZkClient zkClient, String basePath,
```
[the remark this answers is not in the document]
narendly: Discussed offline. This class doesn't need to be a singleton because we could use it elsewhere.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/concurrency/ZkDistributedLeaderElection.java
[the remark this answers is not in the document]
narendly: Clarified offline. This is needed.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/concurrency/ZkDistributedLeaderElection.java
narendly: DataChange includes deletion event as noted in the comment below.

dasahcc:
Remember to resolve the conflict

narendly:
This PR is ready to be merged, approved by @dasahcc .

Add MetadataStoreRoutingDataWriter with DistributedLeaderElection

We need a separate ZkClient-based writer that could allow users to write routing data to ZK. This diff adds such an interface, an implementation, and a distributed lock implementation that could help users to manipulate the routing data.

Changelist:

Add ZkRoutingDataWriter (+ interface)
Add ZkDistributedLock (+ interface) to guarantee that there's at most one active writer at a time (where there are multiple Helix REST deployables)
Add a test for ZkRoutingDataWriter
Integrate ZkRoutingDataWriter with ZkMetadataStoreDirectory
Add test methods to TestZkMetadataStoreDirectory
Add ZkDistributedElection to replace ZkDistributedLock (and move ZkDistributedLock to a separate PR)

PR #731, opened 2020-02-06
Add TrieRoutingData constructor

### LINKED ISSUE #730: Add parsing logic to TrieRoutingData
The previously added TrieRoutingData lacks a constructor which was dependent on a DAO object. That design has been changed, and therefore the constructor is free to be implemented. The constructor takes in a mapping of sharding keys to realm addresses and parses it into a trie.

NealSun96:
### Issues

Fixes #730   

### Description

This PR adds a constructor to TrieRoutingData. The constructor takes in a mapping of sharding keys to realm addresses and parses it into a trie.

### Tests

[INFO] Tests run: 117, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 20.509 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 117, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  46.303 s
[INFO] Finished at: 2020-02-06T11:56:22-08:00
[INFO] -----------------------------------------------------------------------

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
@@ -71,8 +71,8 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
     nodeStack.push(curNode);
     while (!nodeStack.isEmpty()) {
       curNode = nodeStack.pop();
-      if (curNode.isLeaf()) {
-        resultMap.put(curNode.getName(), curNode.getRealmAddress());
+      if (curNode.isShardingKey()) {
+        resultMap.put(curNode.getPath(), curNode.getRealmAddress());
       } else {
         for (TrieNode child : curNode.getChildren().values()) {
           nodeStack.push(child);
@@ -108,14 +108,14 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
   private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
       throws NoSuchElementException {
     if (path.equals(DELIMITER)) {
-      if (findLeafAlongPath && !_rootNode.isLeaf()) {
+      if (findLeafAlongPath && !_rootNode.isShardingKey()) {
         throw new NoSuchElementException("No leaf node found along the path. Path: " + path);
       }
       return _rootNode;
     }
 
     TrieNode curNode = _rootNode;
-    if (findLeafAlongPath && curNode.isLeaf()) {
+    if (findLeafAlongPath && curNode.isShardingKey()) {
       return curNode;
     }
     Map<String, TrieNode> curChildren = curNode.getChildren();
@@ -125,7 +125,7 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
         throw new NoSuchElementException(
             "The provided path is missing from the trie. Path: " + path);
       }
-      if (findLeafAlongPath && curNode.isLeaf()) {
+      if (findLeafAlongPath && curNode.isShardingKey()) {
         return curNode;
       }
       curChildren = curNode.getChildren();
@@ -198,7 +198,7 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
         while (nextDelimiterIndex > 0) {
           // If the node is already a leaf node, the current sharding key is invalid; if the node
           // doesn't exist, construct a node and continue
-          if (nextNode != null && nextNode.isLeaf()) {
+          if (nextNode != null && nextNode.isShardingKey()) {
             throw new InvalidRoutingDataException(shardingKey + " cannot be a sharding key because "
                 + shardingKey.substring(0, nextDelimiterIndex)
                 + " is its parent key and is also a sharding key.");
@@ -234,27 +234,26 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
      */
     private Map<String, TrieNode> _children;
     /**
-     * This field means if the node is a terminal node in the tree sense, not the trie sense. Any
-     * node that has children cannot possibly be a leaf node because only the node without children
-     * can store information. If a node is leaf, then it shouldn't have any children.
+     * This field states whether the path represented by the node is a sharding key
      */
-    private final boolean _isLeaf;
+    private final boolean _isShardingKey;
     /**
-     * This field aligns the traditional trie design: it entails the complete path/prefix leading to
-     * the current node. For example, the name of root node is "/", then the name of its child node
+     * This field contains the complete path/prefix leading to the current node. For example, the
+     * name of root node is "/", then the name of its child node
      * is "/a", and the name of the child's child node is "/a/b".
      */
-    private final String _name;
+    private final String _path;
     /**
      * This field represents the data contained in a node(which represents a path), and is only
      * available to the terminal nodes.
      */
     private final String _realmAddress;
 
-    TrieNode(Map<String, TrieNode> children, String name, boolean isLeaf, String realmAddress) {
+    TrieNode(Map<String, TrieNode> children, String path, boolean isShardingKey,
+        String realmAddress) {
       _children = children;
-      _isLeaf = isLeaf;
-      _name = name;
+      _isShardingKey = isShardingKey;
+      _path = path;
       _realmAddress = realmAddress;
     }
 
@@ -262,12 +261,12 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
       return _children;
     }
 
-    public boolean isLeaf() {
-      return _isLeaf;
+    public boolean isShardingKey() {
+      return _isShardingKey;
     }
 
-    public String getName() {
-      return _name;
+    public String getPath() {
+      return _path;
     }
 
     public String getRealmAddress() {
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
@@ -23,8 +23,10 @@ import java.util.ArrayDeque;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import org.apache.helix.rest.metadatastore.exceptions.InvalidRoutingDataException;
 
 /**
  * This is a class that uses a data structure similar to trie to represent metadata store routing
@@ -37,12 +39,14 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
 
   private final TrieNode _rootNode;
 
-  // TODO: THIS IS A TEMPORARY PLACEHOLDER. A proper constructor will be created, which will not
-  // take in a TrieNode; it instead initializes the rootNode and creates a trie based on
-  // some input data. The constructor is blocked by the implementation of RoutingDataAccessor, and
-  // will therefore be implemented later.
-  public TrieRoutingData(TrieNode rootNode) {
-    _rootNode = rootNode;
+  public TrieRoutingData(Map<String, List<String>> routingData) throws InvalidRoutingDataException {
+    if (isRootShardingKey(routingData)) {
+      Map.Entry<String, List<String>> entry = routingData.entrySet().iterator().next();
+      _rootNode = new TrieNode(Collections.emptyMap(), "/", true, entry.getKey());
+    } else {
+      _rootNode = new TrieNode(new HashMap<>(), "/", false, "");
+      constructTrie(routingData);
+    }
   }
 
   public Map<String, String> getAllMappingUnderPath(String path) {
@@ -124,8 +128,91 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
     return curNode;
   }
 
-  // TODO: THE CLASS WILL BE CHANGED TO PRIVATE ONCE THE CONSTRUCTOR IS CREATED.
-  static class TrieNode {
+  /**
+   * Checks for the edge case when the only sharding key in provided routing data is the delimiter
+   * or an empty string. When this is the case, the trie is valid and contains only one node, which
+   * is the root node, and the root node is a leaf node with a realm address associated with it.
+   * @param routingData - a mapping from "sharding keys" to "realm addresses" to be parsed into a
+   *          trie
+   * @return whether the edge case is true
+   */
+  private boolean isRootShardingKey(Map<String, List<String>> routingData) {
+    if (routingData.values().size() == 1) {
+      for (List<String> shardingKeys : routingData.values()) {
+        return shardingKeys.size() == 1
+            && (shardingKeys.get(0).equals(DELIMITER) || shardingKeys.get(0).equals(""));
+      }
+    }
+
+    return false;
+  }
+
+  /**
+   * Constructs a trie based on the provided routing data. It loops through all sharding keys and
+   * construct the trie in a top down manner.
+   * @param routingData- a mapping from "sharding keys" to "realm addresses" to be parsed into a
+   *          * trie
+   * @throws InvalidRoutingDataException - when there is an empty sharding key (edge case that
+   *           always renders the routing data invalid); when there is a sharding key which already
+   *           contains a sharding key (invalid); when there is a sharding key that is a part of
+   *           another sharding key (invalid)
+   */
+  private void constructTrie(Map<String, List<String>> routingData)
+      throws InvalidRoutingDataException {
+    for (Map.Entry<String, List<String>> entry : routingData.entrySet()) {
+      for (String shardingKey : entry.getValue()) {
+        // Add a leading delimiter if there isn't any
+        if (!shardingKey.substring(0, 1).equals(DELIMITER)) {
+          shardingKey = DELIMITER + shardingKey;
+        }
+
+        // Root can only be a sharding key if it's the only sharding key. Since this method is
+        // running, the special case has already been checked, therefore it's definitely invalid
+        if (shardingKey.equals(DELIMITER)) {
+          throw new InvalidRoutingDataException(
+              "There exists other sharding keys. Root cannot be a sharding key.");
+        }
+
+        // Locate the next delimiter
+        int nextDelimiterIndex = shardingKey.indexOf(DELIMITER, 1);
+        int prevDelimiterIndex = 0;
+        String keySection = shardingKey.substring(prevDelimiterIndex + 1,
+            nextDelimiterIndex > 0 ? nextDelimiterIndex : shardingKey.length());
+        TrieNode curNode = _rootNode;
+        TrieNode nextNode = curNode._children.get(keySection);
+
+        // If the key section is not the last section yet, perform the following
+        while (nextDelimiterIndex > 0) {
+          // If the node is already a leaf node, the current sharding key is invalid
+          if (nextNode != null && nextNode._isLeaf) {
+            throw new InvalidRoutingDataException(shardingKey.substring(0, nextDelimiterIndex)
+                + " is already a sharding key. " + shardingKey + " cannot be a sharding key.");
+          } else if (nextNode == null) {
+            nextNode = new TrieNode(new HashMap<>(), shardingKey.substring(0, nextDelimiterIndex),
+                false, "");
+            curNode._children.put(keySection, nextNode);
+          }
+          prevDelimiterIndex = nextDelimiterIndex;
+          nextDelimiterIndex = shardingKey.indexOf(DELIMITER, prevDelimiterIndex + 1);
+          keySection = shardingKey.substring(prevDelimiterIndex + 1,
+              nextDelimiterIndex > 0 ? nextDelimiterIndex : shardingKey.length());
+          curNode = nextNode;
+          nextNode = curNode._children.get(keySection);
+        }
+
+        // If the last node already exists, it's a part of another sharding key, making the current
+        // sharding key invalid
+        if (nextNode != null) {
+          throw new InvalidRoutingDataException(shardingKey
+              + " is a part of another sharding key, therefore it cannot be a sharding key.");
+        }
+        nextNode = new TrieNode(new HashMap<>(), shardingKey, true, entry.getKey());
+        curNode._children.put(keySection, nextNode);
+      }
+    }
+  }
+
+  private static class TrieNode {
     /**
      * This field is a mapping between trie key and children nodes. For example, node "a" has
      * children "ab" and "ac", therefore the keys are "b" and "c" respectively.
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
@@ -55,8 +55,8 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
 
   public Map<String, String> getAllMappingUnderPath(String path) throws IllegalArgumentException {
     if (path.isEmpty() || !path.substring(0, 1).equals(DELIMITER)) {
-      throw new IllegalArgumentException(
-          "Provided path is empty or does not have a leading \"" + DELIMITER + "\" character: " + path);
+      throw new IllegalArgumentException("Provided path is empty or does not have a leading \""
+          + DELIMITER + "\" character: " + path);
     }
 
     TrieNode curNode;
@@ -85,8 +85,8 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
   public String getMetadataStoreRealm(String path)
       throws IllegalArgumentException, NoSuchElementException {
     if (path.isEmpty() || !path.substring(0, 1).equals(DELIMITER)) {
-      throw new IllegalArgumentException(
-          "Provided path is empty or does not have a leading \"" + DELIMITER + "\" character: " + path);
+      throw new IllegalArgumentException("Provided path is empty or does not have a leading \""
+          + DELIMITER + "\" character: " + path);
     }
 
     TrieNode leafNode = findTrieNode(path, true);
@@ -174,8 +174,8 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
       for (String shardingKey : entry.getValue()) {
         // Missing leading delimiter is invalid
         if (shardingKey.isEmpty() || !shardingKey.substring(0, 1).equals(DELIMITER)) {
-          throw new InvalidRoutingDataException(
-              "Sharding key does not have a leading \"" + DELIMITER + "\" character: " + shardingKey);
+          throw new InvalidRoutingDataException("Sharding key does not have a leading \""
+              + DELIMITER + "\" character: " + shardingKey);
         }
 
         // Root can only be a sharding key if it's the only sharding key. Since this method is
@@ -199,8 +199,9 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
           // If the node is already a leaf node, the current sharding key is invalid; if the node
           // doesn't exist, construct a node and continue
           if (nextNode != null && nextNode.isLeaf()) {
-            throw new InvalidRoutingDataException(shardingKey.substring(0, nextDelimiterIndex)
-                + " is already a sharding key. " + shardingKey + " cannot be a sharding key.");
+            throw new InvalidRoutingDataException(shardingKey + " cannot be a sharding key because "
+                + shardingKey.substring(0, nextDelimiterIndex)
+                + " is its parent key and is also a sharding key.");
           } else if (nextNode == null) {
             nextNode = new TrieNode(new HashMap<>(), shardingKey.substring(0, nextDelimiterIndex),
                 false, "");
@@ -218,7 +219,7 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
         // sharding key invalid
         if (nextNode != null) {
           throw new InvalidRoutingDataException(shardingKey
-              + " is a part of another sharding key, therefore it cannot be a sharding key.");
+              + " cannot be a sharding key because it is a parent key to another sharding key.");
         }
         nextNode = new TrieNode(new HashMap<>(), shardingKey, true, entry.getKey());
         curNode.addChild(keySection, nextNode);
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
@@ -40,6 +40,10 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
   private final TrieNode _rootNode;
 
   public TrieRoutingData(Map<String, List<String>> routingData) throws InvalidRoutingDataException {
+    if (routingData == null || routingData.isEmpty()) {
+      throw new InvalidRoutingDataException("Missing routing data");
+    }
+
     if (isRootShardingKey(routingData)) {
       Map.Entry<String, List<String>> entry = routingData.entrySet().iterator().next();
       _rootNode = new TrieNode(Collections.emptyMap(), "/", true, entry.getKey());
@@ -49,7 +53,12 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
     }
   }
 
-  public Map<String, String> getAllMappingUnderPath(String path) {
+  public Map<String, String> getAllMappingUnderPath(String path) throws IllegalArgumentException {
+    if (path.isEmpty() || !path.substring(0, 1).equals(DELIMITER)) {
+      throw new IllegalArgumentException(
+          "Provided path is empty or does not have a leading delimiter: " + path);
+    }
+
     TrieNode curNode;
     try {
       curNode = findTrieNode(path, false);
@@ -62,10 +71,10 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
     nodeStack.push(curNode);
     while (!nodeStack.isEmpty()) {
       curNode = nodeStack.pop();
-      if (curNode._isLeaf) {
-        resultMap.put(curNode._name, curNode._realmAddress);
+      if (curNode.isLeaf()) {
+        resultMap.put(curNode.getName(), curNode.getRealmAddress());
       } else {
-        for (TrieNode child : curNode._children.values()) {
+        for (TrieNode child : curNode.getChildren().values()) {
           nodeStack.push(child);
         }
       }
@@ -73,9 +82,15 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
     return resultMap;
   }
 
-  public String getMetadataStoreRealm(String path) throws NoSuchElementException {
+  public String getMetadataStoreRealm(String path)
+      throws IllegalArgumentException, NoSuchElementException {
+    if (path.isEmpty() || !path.substring(0, 1).equals(DELIMITER)) {
+      throw new IllegalArgumentException(
+          "Provided path is empty or does not have a leading delimiter: " + path);
+    }
+
     TrieNode leafNode = findTrieNode(path, true);
-    return leafNode._realmAddress;
+    return leafNode.getRealmAddress();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -179,10 +179,10 @@ public void refreshRoutingData(String namespace) {
     }
 
     if (_routingDataMap != null) {
-      MetadataStoreRoutingData newRoutingData =
-          new TrieRoutingData(new TrieRoutingData.TrieNode(null, null, false, null));
+//      MetadataStoreRoutingData newRoutingData =
+//          new TrieRoutingData(new TrieRoutingData.TrieNode(null, null, false, null));
       // TODO call constructRoutingData() here.
-      _routingDataMap.put(namespace, newRoutingData);
+//      _routingDataMap.put(namespace, newRoutingData);
```
dasahcc: If the code is not necessary, let's remove it.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -37,12 +39,14 @@
...
   private final TrieNode _rootNode;
 
-  // TODO: THIS IS A TEMPORARY PLACEHOLDER. A proper constructor will be created, which will not
-  // take in a TrieNode; it instead initializes the rootNode and creates a trie based on
-  // some input data. The constructor is blocked by the implementation of RoutingDataAccessor, and
-  // will therefore be implemented later.
-  public TrieRoutingData(TrieNode rootNode) {
-    _rootNode = rootNode;
+  public TrieRoutingData(Map<String, List<String>> routingData) throws InvalidRoutingDataException {
+    if (isRootShardingKey(routingData)) {
```
dasahcc: Is this necessary to special handling? following constructTrie logic will not handle the root sharding key?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+              "There exists other sharding keys. Root cannot be a sharding key.");
+        }
+
+        // Locate the next delimiter
+        int nextDelimiterIndex = shardingKey.indexOf(DELIMITER, 1);
+        int prevDelimiterIndex = 0;
+        String keySection = shardingKey.substring(prevDelimiterIndex + 1,
+            nextDelimiterIndex > 0 ? nextDelimiterIndex : shardingKey.length());
+        TrieNode curNode = _rootNode;
+        TrieNode nextNode = curNode._children.get(keySection);
```
dasahcc: Better not call variables directly.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+    for (Map.Entry<String, List<String>> entry : routingData.entrySet()) {
+      for (String shardingKey : entry.getValue()) {
+        // Add a leading delimiter if there isn't any
+        if (!shardingKey.substring(0, 1).equals(DELIMITER)) {
+          shardingKey = DELIMITER + shardingKey;
+        }
+
+        // Root can only be a sharding key if it's the only sharding key. Since this method is
+        // running, the special case has already been checked, therefore it's definitely invalid
+        if (shardingKey.equals(DELIMITER)) {
```
pkuwm: You can check this first `isEmpty(), throw exception` before you add DELIMITER to the shardingKey. Could've saved one string object for this loop.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+          throw new InvalidRoutingDataException(
+              "There exists other sharding keys. Root cannot be a sharding key.");
+        }
+
+        // Locate the next delimiter
+        int nextDelimiterIndex = shardingKey.indexOf(DELIMITER, 1);
```
pkuwm: For this part, it seems you can just get a string array shardingKey.split(DELIMITER)  and use the array? This would reduce time complexity. Notice that string.substring() takes O(N) time.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
narendly: With the complete implementation of TrieRoutingData, you should be able to implement the logic here, correct?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
narendly: Null check on routingData?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: It's not unnecessary, it just doesn't work anymore after the changes I made in the PR. @narendly Letting you know that this no longer compiles because TrieNode is converted to private as planned. I'm not entirely sure on how to change it so I'm leaving it up to you.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Yes it is necessary. TrieNodes' fields are mostly final, such as `_isLeaf`. Therefore, we always need to determine if the root node is leaf or not (and construct it) before any of the `constructTrie` code runs. Since it's always the first step, performing a check here is the cleanest way in my opinion.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+   * is the root node, and the root node is a leaf node with a realm address associated with it.
+   * @param routingData - a mapping from "sharding keys" to "realm addresses" to be parsed into a
+   *          trie
+   * @return whether the edge case is true
+   */
+  private boolean isRootShardingKey(Map<String, List<String>> routingData) {
+    if (routingData.values().size() == 1) {
+      for (List<String> shardingKeys : routingData.values()) {
+        return shardingKeys.size() == 1
+            && (shardingKeys.get(0).equals(DELIMITER) || shardingKeys.get(0).equals(""));
```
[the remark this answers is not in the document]
NealSun96: Sure.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+   *           always renders the routing data invalid); when there is a sharding key which already
+   *           contains a sharding key (invalid); when there is a sharding key that is a part of
+   *           another sharding key (invalid)
+   */
+  private void constructTrie(Map<String, List<String>> routingData)
+      throws InvalidRoutingDataException {
+    for (Map.Entry<String, List<String>> entry : routingData.entrySet()) {
+      for (String shardingKey : entry.getValue()) {
+        // Add a leading delimiter if there isn't any
+        if (!shardingKey.substring(0, 1).equals(DELIMITER)) {
```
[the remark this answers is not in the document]
NealSun96: But if I want to compare it with a String, I'd have to cast it back into a string. I'm not sure if this is worth sacrificing cleanness of the code.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
narendly: The code was just left in there for demonstration purposes. Let's implement this part of the logic to complete ZkMetadataStoreDirectory.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+          nextNode = curNode._children.get(keySection);
+        }
+
+        // If the last node already exists, it's a part of another sharding key, making the current
+        // sharding key invalid
+        if (nextNode != null) {
+          throw new InvalidRoutingDataException(shardingKey
+              + " is a part of another sharding key, therefore it cannot be a sharding key.");
+        }
+        nextNode = new TrieNode(new HashMap<>(), shardingKey, true, entry.getKey());
```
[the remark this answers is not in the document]
NealSun96: I think it's more flexible to allow the direct assignment of children. If somehow we need to assign existing children in the future as we are constructing, taking the parameter out now means I have to modify the constructor later.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+        TrieNode nextNode = curNode._children.get(keySection);
+
+        // If the key section is not the last section yet, perform the following
+        while (nextDelimiterIndex > 0) {
+          // If the node is already a leaf node, the current sharding key is invalid
+          if (nextNode != null && nextNode._isLeaf) {
+            throw new InvalidRoutingDataException(shardingKey.substring(0, nextDelimiterIndex)
+                + " is already a sharding key. " + shardingKey + " cannot be a sharding key.");
+          } else if (nextNode == null) {
+            nextNode = new TrieNode(new HashMap<>(), shardingKey.substring(0, nextDelimiterIndex),
```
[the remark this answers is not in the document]
NealSun96: There was a discussion I had with @narendly before about this. I thought that we didn't need `_isLeaf` here also, but `_isLeaf` is a meaningful field. For any trie node, it's either a leaf or not; for regular trie nodes, whether it's a leaf is unrelated to whether it has children. For our trie, it just happens that `_isLeaf == children.isEmpty()`, but it doesn't render the field useless - it's still a property of a trie node.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: That is untrue, because `shardingKey` could be either empty or "/". Adding delimiter first makes the logic easy to understand - after adding the delimiter, all logic applies to sharding keys with leading delimiters.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: That was the original design but note on line 191 that I'll need to obtain the "prefix sharding key", or in other words "the section of sharding key that has been processed". If I do the split approach, I'll need to reconstruct the "prefix sharding key", which costs more time.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: @narendly I didn't entirely understand the refresh logic here, actually. That's why I didn't make the change. Let's sync up so I know what to do here.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Good point, I should do null check and also empty list check.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
narendly: @NealSun96 Let's work together offline :)

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
narendly: @NealSun96 
Can you do a quick analysis here - one method would require the reader/traverser to keep track of the nodes it's traversing and construct a path, and the implementation you're providing here is "caching" the path in the name. Which one makes more sense? Think in the context of what kind of operations would be called more frequently - as in is there a real benefit to caching?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+        // Add a leading delimiter if there isn't any
+        if (!shardingKey.substring(0, 1).equals(DELIMITER)) {
+          shardingKey = DELIMITER + shardingKey;
+        }
+
+        // Root can only be a sharding key if it's the only sharding key. Since this method is
+        // running, the special case has already been checked, therefore it's definitely invalid
+        if (shardingKey.equals(DELIMITER)) {
+          throw new InvalidRoutingDataException(
+              "There exists other sharding keys. Root cannot be a sharding key.");
```
narendly: "There exist"

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+            && (shardingKeys.get(0).equals(DELIMITER) || shardingKeys.get(0).equals(""));
+      }
+    }
+
+    return false;
+  }
+
+  /**
+   * Constructs a trie based on the provided routing data. It loops through all sharding keys and
+   * construct the trie in a top down manner.
```
narendly: constructs

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
narendly: Why is it possible to have two different cases? What does it mean to have just "/", and what does it mean to have an empty string?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+
+        // Locate the next delimiter
+        int nextDelimiterIndex = shardingKey.indexOf(DELIMITER, 1);
+        int prevDelimiterIndex = 0;
+        String keySection = shardingKey.substring(prevDelimiterIndex + 1,
+            nextDelimiterIndex > 0 ? nextDelimiterIndex : shardingKey.length());
+        TrieNode curNode = _rootNode;
+        TrieNode nextNode = curNode._children.get(keySection);
+
+        // If the key section is not the last section yet, perform the following
```
narendly: "the following" ? Can you spell it out?

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/TestTrieRoutingData.java
```diff
@@ -19,146 +19,180 @@
...
   @Test
-  public void testGetAllMappingUnderPathFromRoot() {
-    TrieRoutingData trie = constructTestTrie();
-    Map<String, String> result = trie.getAllMappingUnderPath("/");
-    Assert.assertEquals(result.size(), 4);
-    Assert.assertEquals(result.get("/b/c/d"), "realmAddressD");
-    Assert.assertEquals(result.get("/b/c/e"), "realmAddressE");
-    Assert.assertEquals(result.get("/b/f"), "realmAddressF");
-    Assert.assertEquals(result.get("/g"), "realmAddressG");
+  public void testConstructionSpecialCase() {
```
narendly: Do you think you could add more descriptions about each test case?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Ack

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Ack

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: This comment is really trying to explain the while condition only. The detailed logic of the loop is explained further inside the loop. Modifying this comment to not make it seem incomplete.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: After an offline discussion with  @pkuwm , I've realized that the "storing sharding keys prefixes in nodes" approach is not as good as I used to think, and therefore I'm opening up a discussion on this. A quick analysis of two approaches: 
|                                                                     | Store sharding key prefixes in nodes ("/a", "/a/b", "/a/b/c") | Store sharding key sections in nodes ("/a", "/b", "/c")                                   |
|---------------------------------------------------------------------|--------------------------------------------------------------|------------------------------------------------------------------------------------------|
| Efficiency to return full sharding keys in `getAllMappingUnderPath` | O(1) accessing operations                                    | O(N) joining operations; O(1) to add/remove key sections while traversing the trie; O(N) size map used to keep track whether a trie node has been visited |
| Memory consumption to store the keys in nodes                       | Strictly larger than the combined sizes of all keys (usually several times larger)          | Equal or lower than the combined sizes of all keys                                       |
| Efficiency to construct the trie                                    | O(N) to get the prefix substring of sharding keys            | O(1) since we already have the key sections                                              |

Several additional points:

1. The efficiency of `getAllMappingUnderPath` is more important than the efficiency of construction for two reasons: 1. the frequency of `getAllMappingUnderPath` is likely higher than construction; 2. the efficiency of `getAllMappingUnderPath` matters more because it's serving a REST endpoint. 
2. In the discussion with @pkuwm , we came to a conclusion: if we are storing sharding keys prefixes, why don't we just use a hashmap instead, because a hashmap consumes less memory? This is false because hashmap provides slower string searching than trie. Even if sharding key prefixs are stored, a trie is still better. 

Please provide your opinion and I'll make a decision based on the discussion. Thanks! @dasahcc @narendly @pkuwm

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: Resolved by offline discussion.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: As discussed, the "missing leading slash" case is treated as invalid. Similar other checks are put in place.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Thank you all for the discussion. A rare conclusion: no one has a strong opinion. Therefore it'll be kept as is.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
pkuwm: Agree with @dasahcc . Should have an api getChildren().

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Behaviour has changed. No leading slash is an exception now. Closing conversation.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Adding accessors.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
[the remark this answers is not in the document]
NealSun96: I really don't think this justifies it. There's barely any cost. Thanks for the suggestion, though. Resolving the conversation.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/TestTrieRoutingData.java
NealSun96: Ack

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -124,8 +128,91 @@ private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
...
+  /**
+   * Checks for the edge case when the only sharding key in provided routing data is the delimiter
+   * or an empty string. When this is the case, the trie is valid and contains only one node, which
+   * is the root node, and the root node is a leaf node with a realm address associated with it.
+   * @param routingData - a mapping from "sharding keys" to "realm addresses" to be parsed into a
+   *          trie
+   * @return whether the edge case is true
+   */
+  private boolean isRootShardingKey(Map<String, List<String>> routingData) {
+    if (routingData.values().size() == 1) {
```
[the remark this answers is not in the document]
NealSun96: Ack

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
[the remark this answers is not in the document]
NealSun96: `_isLeaf` is a key field for any trie node. If we are treating this data structure as a trie, a trie node should have such a property. Also, it just happens that `_isLeaf == children.isEmpty()` for now - this might change in the future.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -170,20 +170,21 @@ public void refreshRoutingData(String namespace) {
     // Check if namespace exists; otherwise, return as a NOP and log it
     if (!_routingZkAddressMap.containsKey(namespace)) {
       LOG.error("Failed to refresh internally-cached routing data! Namespace not found: " + namespace);
+      return;
     }
 
     try {
-      _realmToShardingKeysMap.put(namespace, _routingDataReaderMap.get(namespace).getRoutingData());
+      Map<String, List<String>> routingData = _routingDataReaderMap.get(namespace).getRoutingData();
+      _realmToShardingKeysMap.put(namespace, routingData);
```
narendly: Null check  on _realmToShardingKeysMap since method is not necessarily threadsafe.

narendly: As a matter of fact, let's do this check right in the beginning of this method.

If any of the maps (_routingZkAddressMap, _routingZkAddressMap, _routingDataMap) are null, just return.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -170,20 +170,21 @@ public void refreshRoutingData(String namespace) {
...
+
+      if (_routingDataMap != null) {
+        MetadataStoreRoutingData newRoutingData = new TrieRoutingData(routingData);
+        _routingDataMap.put(namespace, newRoutingData);
+      }
     } catch (InvalidRoutingDataException e) {
-      LOG.error("Failed to get routing data for namespace: " + namespace + "!");
+      LOG.error("Routing data construction has failed for namespace: " + namespace + "!");
```
narendly: Not necessarily. There are a few different places where you could get this exception, so your log could be misleading.

You should do `LOG.error("Failed to refresh cached routing data for namespace {}, namespace, e);` This propagates the correct exception message.

narendly:
@NealSun96  
We also discussed adding validation at different stages of I/O for routing data. If that is not going to be in this PR, that's fine, but please create a GitHub issue that contains 1) what the validation will be 2) how you'll validate 3) where the validation logic should be added.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -37,15 +39,26 @@
...
 
-  // TODO: THIS IS A TEMPORARY PLACEHOLDER. A proper constructor will be created, which will not
-  // take in a TrieNode; it instead initializes the rootNode and creates a trie based on
-  // some input data. The constructor is blocked by the implementation of RoutingDataAccessor, and
-  // will therefore be implemented later.
-  public TrieRoutingData(TrieNode rootNode) {
-    _rootNode = rootNode;
+  public TrieRoutingData(Map<String, List<String>> routingData) throws InvalidRoutingDataException {
+    if (routingData == null || routingData.isEmpty()) {
+      throw new InvalidRoutingDataException("Missing routing data");
```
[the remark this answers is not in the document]
NealSun96: Ack

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -37,15 +39,26 @@
...
+      _rootNode = new TrieNode(new HashMap<>(), "/", false, "");
+      constructTrie(routingData);
+    }
   }
 
-  public Map<String, String> getAllMappingUnderPath(String path) {
+  public Map<String, String> getAllMappingUnderPath(String path) throws IllegalArgumentException {
+    if (path.isEmpty() || !path.substring(0, 1).equals(DELIMITER)) {
+      throw new IllegalArgumentException(
+          "Provided path is empty or does not have a leading delimiter: " + path);
```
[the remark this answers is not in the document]
NealSun96: I'll replace the "delimiter" text with the actual delimiter character to make it more clear.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -58,20 +71,26 @@ public TrieRoutingData(TrieNode rootNode) {
...
     }
     return resultMap;
   }
 
-  public String getMetadataStoreRealm(String path) throws NoSuchElementException {
+  public String getMetadataStoreRealm(String path)
+      throws IllegalArgumentException, NoSuchElementException {
+    if (path.isEmpty() || !path.substring(0, 1).equals(DELIMITER)) {
+      throw new IllegalArgumentException(
+          "Provided path is empty or does not have a leading delimiter: " + path);
```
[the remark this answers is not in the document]
NealSun96: ditto

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -88,72 +107,174 @@ public String getMetadataStoreRealm(String path) throws NoSuchElementException {
...
+        TrieNode nextNode = curNode.getChildren().get(keySection);
+
+        // If the key section is not the last section yet, go in the loop; if the key section is the
+        // last section, exit
+        while (nextDelimiterIndex > 0) {
+          // If the node is already a leaf node, the current sharding key is invalid; if the node
+          // doesn't exist, construct a node and continue
+          if (nextNode != null && nextNode.isLeaf()) {
+            throw new InvalidRoutingDataException(shardingKey.substring(0, nextDelimiterIndex)
+                + " is already a sharding key. " + shardingKey + " cannot be a sharding key.");
```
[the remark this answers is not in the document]
NealSun96: I think this is actually less clear for users that do not know the implementation details of `TrieRoutingData`. Saying "parent nodes" does not directly state the problem with "X" being a sharding key; saying "a substring of X is already a sharding key" is more clear.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -88,72 +107,174 @@ public String getMetadataStoreRealm(String path) throws NoSuchElementException {
...
+              nextDelimiterIndex > 0 ? nextDelimiterIndex : shardingKey.length());
+          curNode = nextNode;
+          nextNode = curNode.getChildren().get(keySection);
+        }
+
+        // If the last node already exists, it's a part of another sharding key, making the current
+        // sharding key invalid
+        if (nextNode != null) {
+          throw new InvalidRoutingDataException(shardingKey
+              + " is a part of another sharding key, therefore it cannot be a sharding key.");
```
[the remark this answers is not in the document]
NealSun96: Ditto above.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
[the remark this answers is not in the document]
NealSun96: I can explain further in that direction to make it clearer, such as "having a substring being a sharding key makes the path ambiguous".

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -170,20 +170,21 @@ public void refreshRoutingData(String namespace) {
     // Check if namespace exists; otherwise, return as a NOP and log it
     if (!_routingZkAddressMap.containsKey(namespace)) {
       LOG.error("Failed to refresh internally-cached routing data! Namespace not found: " + namespace);
+      return;
     }
 
     try {
-      _realmToShardingKeysMap.put(namespace, _routingDataReaderMap.get(namespace).getRoutingData());
+      Map<String, List<String>> routingData = _routingDataReaderMap.get(namespace).getRoutingData();
```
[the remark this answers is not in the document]
NealSun96: Right, changing it.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -170,20 +170,21 @@ public void refreshRoutingData(String namespace) {
     // Check if namespace exists; otherwise, return as a NOP and log it
     if (!_routingZkAddressMap.containsKey(namespace)) {
       LOG.error("Failed to refresh internally-cached routing data! Namespace not found: " + namespace);
```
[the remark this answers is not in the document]
NealSun96: Ok.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: Question: if the method is not threadsafe, is a null check at beginning really enough? For example, by the time it gets to `_routingDataMap.put()`, `_routingDataMap` can still be null, same thing applies to `_realmToShardingKeysMap`. 
On the other hand, all map fields are final and they are initialized to be not null. In what situations will they actually become null?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: Good point. I wasn't sure how to phrase this message because there are multiple reasons. I'll use your advised one.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: Interesting Java gotcha! Thank you for explaining offline @narendly !

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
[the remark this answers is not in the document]
NealSun96: Using "parent key" as the clearer terminology for exception messages.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -156,34 +156,42 @@ public boolean deleteShardingKey(String namespace, String realm, String sharding
...
+      Map<String, List<String>> rawRoutingData =
+          _routingDataReaderMap.get(namespace).getRoutingData();
+      _realmToShardingKeysMap.put(namespace, rawRoutingData);
+
+      MetadataStoreRoutingData routingData = new TrieRoutingData(rawRoutingData);
+      _routingDataMap.put(namespace, routingData);
     } catch (InvalidRoutingDataException e) {
-      LOG.error("Failed to get routing data for namespace: " + namespace + "!");
+      LOG.error("Failed to refresh cached routing data for namespace {}, exception: {}", namespace,
+          e.getMessage());
```
narendly: Nit: usually passing the error itself is enough:

`LOG.error("Failed to refresh cached routing data for namespace {}", namespace, e);`

NealSun96:
> Please address pending comments before marking it as ready for check-in.

The validation logic will be on `TrieRoutingDataWriter` and is therefore not a part of this PR. The issue has been created: #737

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: Changed.

NealSun96:
Final commit message:
## Add TrieRoutingData constructor ##
This PR adds a constructor to TrieRoutingData. The constructor takes in a mapping of sharding keys to realm addresses and parses it into a trie. Also, TrieRoutingData now treats sharding keys without leading slashes as invalid. Related changes are made due to the addition of the constructor: TrieNode is made private; missing logic is added for refreshRoutingData in ZkMetadataStoreDirectory.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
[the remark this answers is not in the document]
NealSun96: Per offline discussion, I renamed the fields to reduce confusion on their functionalities.

NealSun96:
Final commit message:
## Add TrieRoutingData constructor ##
This PR adds a constructor to TrieRoutingData. The constructor takes in a mapping of sharding keys to realm addresses and parses it into a trie. Also, TrieRoutingData now treats sharding keys without leading slashes as invalid. Related changes are made due to the addition of the constructor: TrieNode is made private; missing logic is added for refreshRoutingData in ZkMetadataStoreDirectory.

PR #741, opened 2020-02-10
Fix ConcurrentModification exception in Workflow Garbage Collection

### LINKED ISSUE #738: ConcurrentModificationException for the TaskGarbageCollectionStage
In the async TaskGarbageCollectionStage, since the workflow context can be changes in other threads (main pipeline), while looping through all of the contexts we might get the exception below: 

377558 [TaskJobPurgeWorker-CLUSTER_TestTaskRebalancerParallel] ERROR org.apache.helix.controller.pipeline.AbstractAsyncBaseStage  - Failed to process TASK::TaskGarbageCollectionStage asynchronously
java.util.ConcurrentModificationException
at java.util.HashMap$HashIterator.nextNode(HashMap.java:1442)
at java.util.HashMap$EntryIterator.next(HashMap.java:1476)
at java.util.HashMap$EntryIterator.next(HashMap.java:1474)
at org.apache.helix.task.TaskUtil.workflowGarbageCollection(TaskUtil.java:1054)
at org.apache.helix.controller.stages.TaskGarbageCollectionStage.execute(TaskGarbageCollectionStage.java:55)
at org.apache.helix.controller.pipeline.AbstractAsyncBaseStage$1.run(AbstractAsyncBaseStage.java:47)
at org.apache.helix.controller.GenericHelixController$3.handleEvent(GenericHelixController.java:605)
at org.apache.helix.controller.GenericHelixController$3.handleEvent(GenericHelixController.java:601)
at org.apache.helix.common.DedupEventProcessor.run(DedupEventProcessor.java:42)

alirezazamani:
### Issues
Fixes #738 

### Description
In workflow Garbage collection, there is possibility that we see ConcurrentMod exception
while looping through the contexts. The reason behind this exception is that the contexts can be modified using other thread (i.e. controller's main thread or other Garbage Collection threads). This commit fixes this issue by first making a copy of the contexts and then leveraging try catch to avoid such exceptions. 

[ERROR] Tests run: 1083, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 4,194.441 s <<< FAILURE! - in TestSuite
[ERROR] testSubscribeDataChange(org.apache.helix.manager.zk.TestZKWatch)  Time elapsed: 0.023 s  <<< FAILURE!
java.lang.AssertionError: expected:<0> but was:<1>
	at org.apache.helix.manager.zk.TestZKWatch.testSubscribeDataChange(TestZKWatch.java:81)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestZKWatch.testSubscribeDataChange:81 expected:<0> but was:<1>
[INFO] 
[ERROR] Tests run: 1083, Failures: 1, Errors: 0, Skipped: 2
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:09 h
[INFO] Finished at: 2020-02-10T13:27:31-08:00

mvn test -Dtest="TestZKWatch" result:
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.918 s - in org.apache.helix.manager.zk.TestZKWatch
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  6.069 s
[INFO] Finished at: 2020-02-10T13:27:54-08:00

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
@@ -1056,17 +1056,15 @@ public class TaskUtil {
         if (entry != null) {
           WorkflowConfig cfg = dataProvider.getWorkflowConfig(entry);
           WorkflowContext ctx = dataProvider.getWorkflowContext(entry);
-          if (ctx != null && ctx.getId().equals(TaskUtil.WORKFLOW_CONTEXT_KW)) {
-            if (cfg == null) {
-              toBeDeletedWorkflows.add(entry);
-            }
+          if (ctx != null && ctx.getId().equals(TaskUtil.WORKFLOW_CONTEXT_KW) && cfg == null) {
+            toBeDeletedWorkflows.add(entry);
           }
         }
       }
     } catch (Exception e) {
-      LOG.warn(String.format(
-          "Exception occurred while creating a list of all existing contexts with missing config!! Reason: %s"),
-          e.getMessage());
+      LOG.warn(
+          "Exception occurred while creating a list of all existing contexts with missing config!",
+          e);
     }
 
     HelixDataAccessor accessor = manager.getHelixDataAccessor();
--- a/helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
@@ -1040,8 +1040,6 @@ public class TaskUtil {
   /**
    * The function that loops through the all existing workflow contexts and removes IdealState and
    * workflow context of the workflow whose workflow config does not exist.
-   * Try-catch has been used to avoid concurrent modification exception while doing deep copy. Since
-   * Map.keySet() can produce concurrent modification exception.
    * @param dataProvider
    * @param manager
    */
@@ -1049,12 +1047,18 @@ public class TaskUtil {
       final HelixManager manager) {
     // Garbage collections for conditions where workflow context exists but config is missing.
 
-    Set<String> existingWorkflowContexts;
+    Set<String> existingContexts;
+    /*
+     * Here try-catch is used to avoid concurrent modification exception while doing deep copy.
+     * Map.keySet() can produce concurrent modification exception.
+     * Reason: If the map is modified while an iteration over the set is in progress, concurrent
+     * modification exception will be thrown.
+     */
     try {
-      existingWorkflowContexts = new HashSet<>(dataProvider.getContexts().keySet());
+      existingContexts = new HashSet<>(dataProvider.getContexts().keySet());
     } catch (Exception e) {
       LOG.warn(
-          "Exception occurred while creating a list of all existing contexts with missing config!",
+          "Exception occurred while creating a list of all workflow/job context names!",
           e);
       return;
     }
@@ -1062,7 +1066,7 @@ public class TaskUtil {
     // toBeDeletedWorkflows is a set that contains the name of the workflows that their contexts
     // should be deleted.
     Set<String> toBeDeletedWorkflows = new HashSet<>();
-    for (String entry : existingWorkflowContexts) {
+    for (String entry : existingContexts) {
       WorkflowConfig cfg = dataProvider.getWorkflowConfig(entry);
       WorkflowContext ctx = dataProvider.getWorkflowContext(entry);
       if (ctx != null && ctx.getId().equals(TaskUtil.WORKFLOW_CONTEXT_KW) && cfg == null) {
--- a/helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
@@ -1043,22 +1043,34 @@ public class TaskUtil {
    * @param dataProvider
    * @param manager
    */
-  public static void workflowGarbageCollection(WorkflowControllerDataProvider dataProvider,
+  public static void workflowGarbageCollection(final WorkflowControllerDataProvider dataProvider,
       final HelixManager manager) {
     // Garbage collections for conditions where workflow context exists but config is missing.
-    Map<String, ZNRecord> contexts = dataProvider.getContexts();
-    HelixDataAccessor accessor = manager.getHelixDataAccessor();
-    HelixPropertyStore<ZNRecord> propertyStore = manager.getHelixPropertyStore();
 
+    // toBeDeletedWorkflows is a set that contains the name of the workflows that their contexts
+    // should be deleted.
     Set<String> toBeDeletedWorkflows = new HashSet<>();
-    for (Map.Entry<String, ZNRecord> entry : contexts.entrySet()) {
-      if (entry.getValue() != null
-          && entry.getValue().getId().equals(TaskUtil.WORKFLOW_CONTEXT_KW)) {
-        if (dataProvider.getWorkflowConfig(entry.getKey()) == null) {
-          toBeDeletedWorkflows.add(entry.getKey());
+    try {
+      Set<String> existingWorkflowContexts = new HashSet<>(dataProvider.getContexts().keySet());
+      for (String entry : existingWorkflowContexts) {
+        if (entry != null) {
+          WorkflowConfig cfg = dataProvider.getWorkflowConfig(entry);
+          WorkflowContext ctx = dataProvider.getWorkflowContext(entry);
+          if (ctx != null && ctx.getId().equals(TaskUtil.WORKFLOW_CONTEXT_KW)) {
+            if (cfg == null) {
+              toBeDeletedWorkflows.add(entry);
+            }
+          }
         }
       }
+    } catch (Exception e) {
+      LOG.warn(String.format(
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -1043,23 +1043,35 @@ public static void purgeExpiredJobs(String workflow, WorkflowConfig workflowConf
...
-        if (dataProvider.getWorkflowConfig(entry.getKey()) == null) {
-          toBeDeletedWorkflows.add(entry.getKey());
+    try {
+      Set<String> existingWorkflowContexts = new HashSet<>(dataProvider.getContexts().keySet());
+      for (String entry : existingWorkflowContexts) {
+        if (entry != null) {
+          WorkflowConfig cfg = dataProvider.getWorkflowConfig(entry);
+          WorkflowContext ctx = dataProvider.getWorkflowContext(entry);
+          if (ctx != null && ctx.getId().equals(TaskUtil.WORKFLOW_CONTEXT_KW)) {
+            if (cfg == null) {
```
narendly: Can you merge these two if's?

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -1043,23 +1043,35 @@ public static void purgeExpiredJobs(String workflow, WorkflowConfig workflowConf
...
+              toBeDeletedWorkflows.add(entry);
+            }
+          }
         }
       }
+    } catch (Exception e) {
+      LOG.warn(String.format(
+          "Exception occurred while creating a list of all existing contexts with missing config!! Reason: %s"),
```
narendly: 1. Use {}
2. You can input the exception as the second parameter. You don't have to do e.getMessage().

narendly: Use one "!'.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
alirezazamani: Done.

alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -1043,23 +1043,33 @@ public static void purgeExpiredJobs(String workflow, WorkflowConfig workflowConf
...
+          if (ctx != null && ctx.getId().equals(TaskUtil.WORKFLOW_CONTEXT_KW) && cfg == null) {
+            toBeDeletedWorkflows.add(entry);
+          }
         }
       }
+    } catch (Exception e) {
+      LOG.warn(
+          "Exception occurred while creating a list of all existing contexts with missing config!",
+          e);
     }
```
narendly: Do you clearly understand why and where ConcurrentModificationException is happening? This looks like we're just masking it with a try-catch. I don't think that's a fix. Could we identify why and where clearly first and could you please update the description with your findings? Then this try-catch shouldn't be necessary.

alirezazamani: Yes. I tried several scenarios and for each scenario I used Jiajun's scripts which runs the test for 50 time. The most effective solution is the one that I proposed in this PR. Please have a look at this line: 

- Set<String> existingWorkflowContexts = new HashSet<>(dataProvider.getContexts().keySet());

In some minor cases (about 2 out of 50 runs) above line is the only line that can generate concurrent modification exception which I eliminated it with try-catch. The reason behind this is because while we want to get all of the existing contexts, the contextMap can be modified in the cache by other threads. As a result we will get concurrentMod exception. Please note that this part of the code runs asynchronously.
@narendly I don't have strong preference about this method and I would be happy if you can propose new way to get all of the context without hitting concurrent modification exception.

narendly: If you make a deep copy of the set, you shouldn't need to wrap it with a try-catch since your async thread would be the only thread iterating over the data structure, correct? What I'm not understanding is why we need the try-catch.

alirezazamani: I don't believe deep copy will help in this case (it would help in the single threaded case where you want to remove elements from the map or list). The code generate concurrent modification in this function: map.keySet(). We cannot avoid this. How you are proposing to get the keys in the map in this scenario? We need to get the keys no matter what. Even the implementation of deep copying involves to loop over the elements and copy them one by one to the new map. Right? In this case if original map has been changes while copy operation is happening, we might still get concurrent modification exception.

narendly: @alirezazamani  thanks for the explanation :) Now that makes a lot more sense.

One structural improvement I'd suggest:
1. Please only wrap the copy constructor with a the try-catch to make your intentions clear.
2. It would be good to explain this in the code as a javadoc comment (as to why it needs to be wrapped with a try-catch).
3. Your catch should return the function early.

alirezazamani: Done. Thanks @narendly for the comments.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -1040,26 +1040,39 @@ public static void purgeExpiredJobs(String workflow, WorkflowConfig workflowConf
   /**
    * The function that loops through the all existing workflow contexts and removes IdealState and
    * workflow context of the workflow whose workflow config does not exist.
+   * Try-catch has been used to avoid concurrent modification exception while doing deep copy. Since
+   * Map.keySet() can produce concurrent modification exception.
```
narendly: 1. move this to where try-catch is?
2. Explain why Map.keySet() throws CME w regards to copy constructor?

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -1040,26 +1040,39 @@ public static void purgeExpiredJobs(String workflow, WorkflowConfig workflowConf
...
-    HelixDataAccessor accessor = manager.getHelixDataAccessor();
-    HelixPropertyStore<ZNRecord> propertyStore = manager.getHelixPropertyStore();
 
+    Set<String> existingWorkflowContexts;
+    try {
+      existingWorkflowContexts = new HashSet<>(dataProvider.getContexts().keySet());
+    } catch (Exception e) {
+      LOG.warn(
+          "Exception occurred while creating a list of all existing contexts with missing config!",
+          e);
```
narendly: "workflow context names"

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
alirezazamani: Done. Please let me know if it is still not clear.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
alirezazamani: Done.

alirezazamani:

Final commit message:
Title: 
Fix ConcurrentModification exception in Workflow Garbage Collection

Body:
In workflow Garbage collection, there is possibility that we encounter
ConcurrentMod exception while looping through the workflow contexts.
This commit fixes this issue by looping through a deep copy of the 
contexts list.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -1043,23 +1043,40 @@ public static void purgeExpiredJobs(String workflow, WorkflowConfig workflowConf
...
 
+    Set<String> existingContexts;
+    /*
+     * Here try-catch is used to avoid concurrent modification exception while doing deep copy.
+     * Map.keySet() can produce concurrent modification exception.
+     * Reason: If the map is modified while an iteration over the set is in progress, concurrent
+     * modification exception will be thrown.
+     */
+    try {
+      existingContexts = new HashSet<>(dataProvider.getContexts().keySet());
```
pkuwm: @alirezazamani Is there a specific reason why we don't change the contexts map below to a concurrent hash map and make it thread safe? 
```
  private Map<String, ZNRecord> _contextMap = new HashMap<>();
``` 
What I can only see is performance concern when contexts modification is blocked by this copying. If performance is not that bad, I believe changing the contextMap to a concurrent hash map is the right fix.

alirezazamani: dataProvider.getContexts() is outside the scope of this class and perhaps there are performance concerns regarding this. Maybe we can consider this in next design of TF.

PR #745, opened 2020-02-11
Add RealmAwareZkClient and RealmAwareZkClientFactory interfaces

### LINKED ISSUE #760: Add RealmAwareZkClient and RealmAwareZkClientFactory
RealmAwareZkClient and RealmAwareZkClientFactory are going to be the top-level interfaces for other realm-aware ZkClient APIs (think FederatedZkClient, DedicatedZkClient, etc.). 

RealmAwareZkClient will support all the existing interface methods that HelixZkClient supports.
RealmAwareZkClientFactory will be the interface implemented by SharedZkClientFactory and DedicatedZkClientFactory.

narendly:
### Issues

Resolves #760 

### Description

RealmAwareZkClient and RealmAwareZkClientFactory are going to be the top-level interfaces for other realm-aware ZkClient APIs (think FederatedZkClient, DedicatedZkClient, etc.).

RealmAwareZkClient will support all the existing interface methods that HelixZkClient supports.
RealmAwareZkClientFactory will be the interface implemented by SharedZkClientFactory and DedicatedZkClientFactory.

### Tests

Interface methods without logic - no tests added.

build passed.

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
@@ -89,6 +89,9 @@ public interface HelixZkClient extends RealmAwareZkClient {
    * Deprecated - please use RealmAwareZkClient and RealmAwareZkClientConfig instead.
    *
    * Configuration for creating a new HelixZkClient with serializer and monitor.
+   *
+   * TODO: If possible, try to merge with RealmAwareZkClient's RealmAwareZkClientConfig to reduce duplicate logic/code (without breaking backward-compatibility).
+   * Simply making this a subclass of RealmAwareZkClientConfig will break backward-compatiblity.
    */
   @Deprecated
   class ZkClientConfig {
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
@@ -19,8 +19,178 @@ package org.apache.helix.zookeeper.api.client;
  * under the License.
  */
 
+import org.apache.helix.zookeeper.zkclient.serialize.BasicZkSerializer;
+import org.apache.helix.zookeeper.zkclient.serialize.PathBasedZkSerializer;
+import org.apache.helix.zookeeper.zkclient.serialize.SerializableSerializer;
+import org.apache.helix.zookeeper.zkclient.serialize.ZkSerializer;
+
+
 /**
+ * Deprecated - please use RealmAwareZkClient instead.
+ *
  * HelixZkClient interface that follows the supported API structure of RealmAwareZkClient.
  */
+@Deprecated
 public interface HelixZkClient extends RealmAwareZkClient {
+
+  /**
+   * Deprecated - please use RealmAwareZkClient and RealmAwareZkConnectionConfig instead.
+   *
+   * Configuration for creating a new ZkConnection.
+   */
+  @Deprecated
+  class ZkConnectionConfig {
+    // Connection configs
+    private final String _zkServers;
+    private int _sessionTimeout = HelixZkClient.DEFAULT_SESSION_TIMEOUT;
+
+    public ZkConnectionConfig(String zkServers) {
+      _zkServers = zkServers;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (obj == this) {
+        return true;
+      }
+      if (!(obj instanceof HelixZkClient.ZkConnectionConfig)) {
+        return false;
+      }
+      HelixZkClient.ZkConnectionConfig configObj = (HelixZkClient.ZkConnectionConfig) obj;
+      return (_zkServers == null && configObj._zkServers == null || _zkServers != null && _zkServers
+          .equals(configObj._zkServers)) && _sessionTimeout == configObj._sessionTimeout;
+    }
+
+    @Override
+    public int hashCode() {
+      return _sessionTimeout * 31 + _zkServers.hashCode();
+    }
+
+    @Override
+    public String toString() {
+      return (_zkServers + "_" + _sessionTimeout).replaceAll("[\\W]", "_");
+    }
+
+    public HelixZkClient.ZkConnectionConfig setSessionTimeout(Integer sessionTimeout) {
+      this._sessionTimeout = sessionTimeout;
+      return this;
+    }
+
+    public String getZkServers() {
+      return _zkServers;
+    }
+
+    public int getSessionTimeout() {
+      return _sessionTimeout;
+    }
+  }
+
+  /**
+   * Deprecated - please use RealmAwareZkClient and RealmAwareZkClientConfig instead.
+   *
+   * Configuration for creating a new HelixZkClient with serializer and monitor.
+   */
+  @Deprecated
+  class ZkClientConfig extends RealmAwareZkClientConfig {
+    // For client to init the connection
+    private long _connectInitTimeout = DEFAULT_CONNECTION_TIMEOUT;
+
+    // Data access configs
+    private long _operationRetryTimeout = DEFAULT_OPERATION_TIMEOUT;
+
+    // Others
+    private PathBasedZkSerializer _zkSerializer;
+
+    // Monitoring
+    private String _monitorType;
+    private String _monitorKey;
+    private String _monitorInstanceName = null;
+    private boolean _monitorRootPathOnly = true;
+
+    public ZkClientConfig setZkSerializer(PathBasedZkSerializer zkSerializer) {
+      this._zkSerializer = zkSerializer;
+      return this;
+    }
+
+    public ZkClientConfig setZkSerializer(ZkSerializer zkSerializer) {
+      this._zkSerializer = new BasicZkSerializer(zkSerializer);
+      return this;
+    }
+
+    /**
+     * Used as part of the MBean ObjectName. This item is required for enabling monitoring.
+     *
+     * @param monitorType
+     */
+    public ZkClientConfig setMonitorType(String monitorType) {
+      this._monitorType = monitorType;
+      return this;
+    }
+
+    /**
+     * Used as part of the MBean ObjectName. This item is required for enabling monitoring.
+     *
+     * @param monitorKey
+     */
+    public ZkClientConfig setMonitorKey(String monitorKey) {
+      this._monitorKey = monitorKey;
+      return this;
+    }
+
+    /**
+     * Used as part of the MBean ObjectName. This item is optional.
+     *
+     * @param instanceName
+     */
+    public ZkClientConfig setMonitorInstanceName(String instanceName) {
+      this._monitorInstanceName = instanceName;
+      return this;
+    }
+
+    public ZkClientConfig setMonitorRootPathOnly(Boolean monitorRootPathOnly) {
+      this._monitorRootPathOnly = monitorRootPathOnly;
+      return this;
+    }
+
+    public ZkClientConfig setOperationRetryTimeout(Long operationRetryTimeout) {
+      this._operationRetryTimeout = operationRetryTimeout;
+      return this;
+    }
+
+    public ZkClientConfig setConnectInitTimeout(long _connectInitTimeout) {
+      this._connectInitTimeout = _connectInitTimeout;
+      return this;
+    }
+
+    public PathBasedZkSerializer getZkSerializer() {
+      if (_zkSerializer == null) {
+        _zkSerializer = new BasicZkSerializer(new SerializableSerializer());
+      }
+      return _zkSerializer;
+    }
+
+    public long getOperationRetryTimeout() {
+      return _operationRetryTimeout;
+    }
+
+    public String getMonitorType() {
+      return _monitorType;
+    }
+
+    public String getMonitorKey() {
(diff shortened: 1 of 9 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/api/factory/RealmAwareZkClientFactory.java
```diff
@@ -19,5 +19,14 @@
...
+import org.apache.helix.zookeeper.api.api.client.RealmAwareZkClient;
+
+
+/**
+ * Creates an instance of RealmAwareZkClient.
+ */
 public interface RealmAwareZkClientFactory {
+
+  RealmAwareZkClient buildZkClient(RealmAwareZkClient.MODE mode,
+      RealmAwareZkClient.RealmAwareZkClientConfig config);
```
kaisun2000: Just a note, this means that we need to use the RealmAwareZkClientConfig to genereate HelixZkClient.ZkConnectionConfig. Needs validate.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/api/factory/RealmAwareZkClientFactory.java
```diff
@@ -19,5 +19,14 @@
...
 
+import org.apache.helix.zookeeper.api.api.client.RealmAwareZkClient;
+
+
+/**
+ * Creates an instance of RealmAwareZkClient.
+ */
 public interface RealmAwareZkClientFactory {
+
+  RealmAwareZkClient buildZkClient(RealmAwareZkClient.MODE mode,
```
kaisun2000: mode is redundant.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/impl/factory/HelixZkClientFactory.java
```diff
@@ -28,7 +29,7 @@
 /**
  * Abstract class of the ZkClient factory.
  */
-abstract class HelixZkClientFactory {
+abstract class HelixZkClientFactory implements RealmAwareZkClientFactory {
```
kaisun2000: Do we want to deprecate the following?
public abstract HelixZkClient buildZkClient(HelixZkClient.ZkConnectionConfig connectionConfig,
      HelixZkClient.ZkClientConfig clientConfig);

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/api/factory/RealmAwareZkClientFactory.java
kaisun2000: Another high level question, in the scope of dedicatedZkClient or sharedZkClient discussion, all the create ZkClient already are RealAware, right? we don't really need to have a mode for user to specify.

The mode can be kept internally for the client to know. In fact, even this (internally keep the type) may not be needed as we have new dedicatedZkClient and new sharedZkClient which wraps sharedZkClient and zkclient as of now. The type information is already encode in the class design. The beauty of OO inheritance.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/impl/factory/HelixZkClientFactory.java
kaisun2000: In fact, after some thought, it seems to me that you only need a realm key and keep the rest of HelixZkClientFactory intact would be fine. Let me draw some psudeo-code to illustrate why this is true? I will share a doc a little bit later.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/api/factory/RealmAwareZkClientFactory.java
narendly: @kaisun2000 
1. Yes, we should add the sessionTimeout field to this. I will add it for completeness (this was a draft, so I just wanted to first give you an idea). :)
2. Yes and no.
- RealmAwareZkClients produced by RealmAwareZkClientFactory are all single-realm mode. So yes, mode might not be required as a method parameter for `buildZkClient()`.
- We should still keep this field internally because not all RealmAwareZkClients are single-realm mode. For example, if you create a FederatedZkClient instance, it should still have a `mode = MULTI_REALM`.
So yes, let's follow your suggestion to remove the mode parameter from here, but we should keep it as an internal variable that we could expose to the user (for instance, RealmAwareZkClient should support an interface method like `getMode()` that returns either `SINGLE_REALM` or `MULTI_REALM`)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/api/factory/RealmAwareZkClientFactory.java
narendly: Yes, I will remove :)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/impl/factory/HelixZkClientFactory.java
narendly: Yes, let us deprecate all HelixZkClientFactory/HelixZkClient APIs.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/api/client/RealmAwareZkClient.java
```diff
@@ -305,159 +314,12 @@ public int hashCode() {
...
-      return _monitorInstanceName;
-    }
-
-    public boolean isMonitorRootPathOnly() {
-      return _monitorRootPathOnly;
-    }
-
-    public long getConnectInitTimeout() {
-      return _connectInitTimeout;
+  class RealmAwareZkClientConfig {
```
kaisun2000: Shall we change it ConnectionConfig? Conceptually this is the same as HelixZkClient.ZkConnectionConfig. In HelixZkClient the ClientConfig is about the monitor key, serializer etc. Let us keep the naming to concept consistent, otherwise, it can be confusing.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
```diff
@@ -19,8 +19,178 @@
...
+    }
+  }
+
+  /**
+   * Deprecated - please use RealmAwareZkClient and RealmAwareZkClientConfig instead.
+   *
+   * Configuration for creating a new HelixZkClient with serializer and monitor.
+   */
+  @Deprecated
+  class ZkClientConfig {
```
kaisun2000: Another choice is to let ZkClientConfig inherit RealmAwareZkClient.RealmAwareZkClientConfig as the implementation of this two class are implementation are exactly the same. This way, no duplicate code. What is your take?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/factory/RealmAwareZkClientFactory.java
```diff
@@ -19,5 +19,13 @@
...
  */
 
+import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
+
+
+/**
+ * Creates an instance of RealmAwareZkClient.
+ */
 public interface RealmAwareZkClientFactory {
+  RealmAwareZkClient buildZkClient(RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig,
```
kaisun2000: To achieve feature parity of HelixZkClientInterface, shall we add this constructor too?

`   public RealmAwareZkClient buildZkClient(
      RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig) {
     return buildZkClient(connectConfig, new RealmAwareZkClient.RealmAwareZkClientConfig());
   }`

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
narendly: @kaisun2000 
I tried that but that is going to break backward-compatibility because of different return types in the setter methods.

To illustrate:
`public ZkClientConfig setZkSerializer(PathBasedZkSerializer zkSerializer)` is what we have in HelixZkClient. But if this class extended RealmAwareZkClientConfig, then it will be returning an instance of RealmAwareZkClientConfig, which would break the existing methods in Shared and DedicatedZkClientFactories.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/factory/RealmAwareZkClientFactory.java
narendly: Sure, that's a good idea. Let me add it as well.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
```diff
@@ -306,48 +326,63 @@ public int hashCode() {
...
+    /**
+     * zkRealmShardingKey: used to deduce which ZK realm this RealmAwareZkClientConfig should connect to.
+     * NOTE: this field will be ignored if MODE is MULTI_REALM!
+     */
+    private final String _zkRealmShardingKey;
+    private int _sessionTimeout = DEFAULT_SESSION_TIMEOUT;
 
-    public ZkConnectionConfig(String zkServers) {
-      _zkServers = zkServers;
+    public RealmAwareZkConnectionConfig(MODE mode, String zkRealmShardingKey) {
```
kaisun2000: If we think MODE is already encoded in the Type (class hierarchy), we don't need this in the constructor.  Based on looking at the psuedo code I drew, I don't fee we really need to pass this in.

narendly: I am thinking that we would use this for FederatedZkClient as well since FederatedZkClient is an implementation of RealmAwareZkClient. With that said, this is necessary because the mode will have to be different.

So mode is not necessarily encoded in the type.

kaisun2000: FederatedZkClient is always multi. The implementation of FederateZkClient can assume this without you passing a mode. The problem is that if you pass a non-multi, it need to handle it then.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
```diff
@@ -250,21 +268,22 @@ default long waitForEstablishedSession(long timeout, TimeUnit timeUnit) {
    * This is for backward compatibility and to avoid breaking the original implementation of
    * {@link org.apache.helix.zookeeper.zkclient.deprecated.IZkStateListener}.
    */
-  class I0ItecIZkStateListenerHelixImpl
-      implements org.apache.helix.zookeeper.zkclient.deprecated.IZkStateListener {
+  class I0ItecIZkStateListenerHelixImpl implements org.apache.helix.zookeeper.zkclient.deprecated.IZkStateListener {
```
dasahcc: Shall we remove Helix word?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
```diff
@@ -306,48 +326,63 @@ public int hashCode() {
...
+    private int _sessionTimeout = DEFAULT_SESSION_TIMEOUT;
 
-    public ZkConnectionConfig(String zkServers) {
-      _zkServers = zkServers;
+    public RealmAwareZkConnectionConfig(MODE mode, String zkRealmShardingKey) {
+      if (mode == null) {
+        throw new ZkClientException("Mode cannot be null!");
+      }
+      _mode = mode;
+      _zkRealmShardingKey = zkRealmShardingKey;
```
dasahcc: Seemed to me the connection is per zk realm. Then the routing logic should be at Factory layer, right?

narendly: That's exactly right.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
narendly: Removed

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
```diff
@@ -40,7 +41,24 @@
...
 public interface RealmAwareZkClient {
+
+  /**
+   * Specifies which mode to run this RealmAwareZkClient on.
+   *
+   * SINGLE_REALM: CRUD, change subscription, and EPHEMERAL CreateMode are supported.
+   * MULTI_REALM: CRUD and change subscription are supported. Operations involving EPHEMERAL CreateMode will throw an UnsupportedOperationException.
+   */
+  enum MODE {
+    SINGLE_REALM, MULTI_REALM
```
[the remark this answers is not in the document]
narendly: updated!

PR #759, opened 2020-02-13
Add validation logic to MSD write operations

### LINKED ISSUE #758: Add validation logic to MetadataStoreDirectory write methods
The metadata storage routing data stored in Zookeeper is the source of truth for all related services, and therefore, it needs to stay well maintained and should always be valid. As a result, validation logic is implemented on all writing operations towards the routing data - when a user attempts to add a sharding key to some existing routing data, validation logic is used to confirm that the addition will leave the resulting routing data in a valid state.

NealSun96:
### Issues

Fixes #758   

### Description

This PR add routing data validation logic. When a user attempts to add a sharding key to some existing routing data, validation logic is used to confirm that the addition will leave the resulting routing data in a valid state. We make use of the existing trie structure to efficiently check if a sharding key results in an ambiguity case. Some old code in `TrieRoutingData` is modified to minimize duplicate code and make everything a lot cleaner. 

### Tests

[INFO] Tests run: 143, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 21.642 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 143, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  47.466 s
[INFO] Finished at: 2020-02-13T11:13:56-08:00

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
@@ -26,7 +26,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
+
 import org.apache.helix.rest.metadatastore.exceptions.InvalidRoutingDataException;
+
 
 /**
  * This is a class that uses a data structure similar to trie to represent metadata store routing
@@ -55,14 +57,13 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
 
   public Map<String, String> getAllMappingUnderPath(String path) throws IllegalArgumentException {
     if (path.isEmpty() || !path.substring(0, 1).equals(DELIMITER)) {
-      throw new IllegalArgumentException("Provided path is empty or does not have a leading \""
-          + DELIMITER + "\" character: " + path);
+      throw new IllegalArgumentException(
+          "Provided path is empty or does not have a leading \"" + DELIMITER + "\" character: "
+              + path);
     }
 
-    TrieNode curNode;
-    try {
-      curNode = findTrieNode(path, false);
-    } catch (NoSuchElementException e) {
+    TrieNode curNode = findNearestNodeToPath(path);
+    if (!curNode.getPath().equals(path)) {
       return Collections.emptyMap();
     }
 
@@ -85,53 +86,57 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
   public String getMetadataStoreRealm(String path)
       throws IllegalArgumentException, NoSuchElementException {
     if (path.isEmpty() || !path.substring(0, 1).equals(DELIMITER)) {
-      throw new IllegalArgumentException("Provided path is empty or does not have a leading \""
-          + DELIMITER + "\" character: " + path);
+      throw new IllegalArgumentException(
+          "Provided path is empty or does not have a leading \"" + DELIMITER + "\" character: "
+              + path);
     }
 
-    TrieNode leafNode = findTrieNode(path, true);
-    return leafNode.getRealmAddress();
+    TrieNode node = findNearestNodeToPath(path);
+    if (!node.isShardingKey()) {
+      throw new NoSuchElementException(
+          "No sharding key found within the provided path. Path: " + path);
+    }
+    return node.getRealmAddress();
   }
 
   /**
-   * If findLeafAlongPath is false, then starting from the root node, find the trie node that the
-   * given path is pointing to and return it; raise NoSuchElementException if the path does
-   * not point to any node. If findLeafAlongPath is true, then starting from the root node, find the
-   * leaf node along the provided path; raise NoSuchElementException if the path does not
-   * point to any node or if there is no leaf node along the path.
-   * @param path - the path where the search is conducted
-   * @param findLeafAlongPath - whether the search is for a leaf node on the path
-   * @return the node pointed by the path or a leaf node along the path
-   * @throws NoSuchElementException - when the path points to nothing or when no leaf node is
-   *           found
+   * Check if the provided sharding key can be inserted to the existing trie. The insertion is
+   * invalid if: 1. the sharding key is a parent key to an existing sharding key; 2. the sharding
+   * key has a parent key that is an existing sharding key; 3. the sharding key already exists
+   * @param shardingKey - the sharding key to be inserted
+   * @return true if the sharding key could be inserted, false otherwise
    */
-  private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
-      throws NoSuchElementException {
+  public boolean isShardingKeyInsertionValid(String shardingKey) {
+    if (shardingKey.isEmpty() || !shardingKey.substring(0, 1).equals(DELIMITER)) {
+      throw new IllegalArgumentException(
+          "Provided shardingKey is empty or does not have a leading \"" + DELIMITER
+              + "\" character: " + shardingKey);
+    }
+
+    TrieNode node = findNearestNodeToPath(shardingKey);
+    return !node.isShardingKey() && !node.getPath().equals(shardingKey);
+  }
+
+  /**
+   * Given a path, find a trie node that is the nearest to the given path. For example, given
+   * "/a/b/c", the TrieNode that represents "/a/b/c" is the nearest node to path. If "/a/b/c"
+   * doesn't exist, then "/a/b" is the nearest node, then "/a", and lastly "/"
+   * @param path - the path where the search is conducted
+   * @return a TrieNode that is the "nearest" according to the definition here
+   */
+  private TrieNode findNearestNodeToPath(String path) {
     if (path.equals(DELIMITER)) {
-      if (findLeafAlongPath && !_rootNode.isShardingKey()) {
-        throw new NoSuchElementException("No leaf node found along the path. Path: " + path);
-      }
       return _rootNode;
     }
 
     TrieNode curNode = _rootNode;
-    if (findLeafAlongPath && curNode.isShardingKey()) {
-      return curNode;
-    }
-    Map<String, TrieNode> curChildren = curNode.getChildren();
+    TrieNode nextNode;
     for (String pathSection : path.substring(1).split(DELIMITER, 0)) {
-      curNode = curChildren.get(pathSection);
-      if (curNode == null) {
-        throw new NoSuchElementException(
-            "The provided path is missing from the trie. Path: " + path);
-      }
-      if (findLeafAlongPath && curNode.isShardingKey()) {
+      nextNode = curNode.getChildren().get(pathSection);
+      if (nextNode == null) {
         return curNode;
       }
-      curChildren = curNode.getChildren();
-    }
-    if (findLeafAlongPath) {
-      throw new NoSuchElementException("No leaf node found along the path. Path: " + path);
+      curNode = nextNode;
     }
     return curNode;
   }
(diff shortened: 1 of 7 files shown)


NealSun96:
**To all:** As I attempt to format the relevant files, it turns out that the formatting on my IDE disagrees with some of the existing files (using the latest `helix-style-intellij.xml`). I'm not sure if it's due to the files not formatted correctly previously or misconfiguration on my end. 
To reduce noise in diff, formatting has been temporarily held off. Once the logic has been reviewed, I'll create a commit for formatting only, where the formatting  can be reviewed.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -89,49 +87,55 @@ public String getMetadataStoreRealm(String path)
...
    * @param path - the path where the search is conducted
-   * @param findLeafAlongPath - whether the search is for a leaf node on the path
-   * @return the node pointed by the path or a leaf node along the path
-   * @throws NoSuchElementException - when the path points to nothing or when no leaf node is
-   *           found
+   * @return a TrieNode that represents the longest prefix of the path
    */
-  private TrieNode findTrieNode(String path, boolean findLeafAlongPath)
-      throws NoSuchElementException {
+  private TrieNode getLongestPrefixNodeAlongPath(String path) {
```
narendly: The name for this method is a little confusing. Do you think something like findClosestParentNode or findClosetShardingKeyNode or findClosestTerminalNode would be easier to understand?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -89,49 +87,55 @@ public String getMetadataStoreRealm(String path)
...
+    return !node.isShardingKey() && !node.getPath().equals(shardingKey);
   }
 
   /**
-   * If findLeafAlongPath is false, then starting from the root node, find the trie node that the
-   * given path is pointing to and return it; raise NoSuchElementException if the path does
-   * not point to any node. If findLeafAlongPath is true, then starting from the root node, find the
-   * leaf node along the provided path; raise NoSuchElementException if the path does not
-   * point to any node or if there is no leaf node along the path.
+   * Given a path, find a trie node that represents the longest prefix of the path. For example,
```
narendly: "longest prefix"? Could we just call it the closest existing sharding key node?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -89,49 +87,55 @@ public String getMetadataStoreRealm(String path)
...
+   * given "/a/b/c", the method starts at "/", and attempts to reach "/a", then attempts to reach
+   * "/a/b", then ends on "/a/b/c"; if any of the node doesn't exist, the traversal terminates and
+   * the last existing node is returned.
+   * Note:
+   * 1. When the returned TrieNode is a sharding key, it is the only sharding key along the
+   * provided path (the path points to this sharding key);
+   * 2. When the returned TrieNode is not a sharding key but it represents the provided path, the
+   * provided path is a prefix(parent) to a sharding key;
+   * 3. When the returned TrieNode is not a sharding key and it does not represent the provided
```
narendly: Is this scenario possible (assuming that the existing routing data is valid)?

narendly: Oh, never mind. This is the case where it's "successful".

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Appreciate the feedback on naming - it's something that I wish to discuss. 
My original naming is along the line of "findClosestNode", however I find myself have to explain the definition of "closest". For example, given "/a/b/c", why is "/a" the closer than "/a/x/c" (if "/a/b" doesn't exist)?  Also, saying "closest node" when `path` sometimes doesn't point to a node is strange.
I found that "longest prefix" is the most accurate definition of what this method is doing - it finds the longest prefix of `path` that exists in this trie.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Ditto my explanation above.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Yes it is. This scenario is when `path` is a sharding key that could be added to the trie. Consider a trie that only has "/a/b/c" in it:
1. `getLongestPrefixNodeAlongPath("/a/b/c/x")` is the first case (typically used during realm lookups);
2. `getLongestPrefixNodeAlongPath("/a/b")` is the second case (typically used during key scanning); 
3. `getLongestPrefixNodeAlongPath("/a/b/d")` is the last case. Note that "/a/b/d" is a valid sharding key to add.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: To add to my last comment, "longest prefix" is very fitting of an explanation. Given "/a/b/c", this method will attempt to return a node that represents "/a/b/c"; if that doesn't exist, it tries "/a/b"; if not, "/a"; if everything else fails, return "/". 
IMO it's easier to say "longest prefix" then explaining "node that is closest to `path` and is a parent to `path`".

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -89,49 +87,55 @@ public String getMetadataStoreRealm(String path)
...
   /**
-   * If findLeafAlongPath is false, then starting from the root node, find the trie node that the
-   * given path is pointing to and return it; raise NoSuchElementException if the path does
-   * not point to any node. If findLeafAlongPath is true, then starting from the root node, find the
-   * leaf node along the provided path; raise NoSuchElementException if the path does not
-   * point to any node or if there is no leaf node along the path.
+   * Given a path, find a trie node that represents the longest prefix of the path. For example,
+   * given "/a/b/c", the method starts at "/", and attempts to reach "/a", then attempts to reach
+   * "/a/b", then ends on "/a/b/c"; if any of the node doesn't exist, the traversal terminates and
+   * the last existing node is returned.
```
narendly: last-seen existing node?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
narendly: Explain what the longest prefix is: The longest prefix means the closest existing terminal node (sharding key).

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
narendly: Longest prefix is still confusing. When I read the code, I had to think about what you meant by "longest prefix". Because the word "prefix' could mean a host of different things depending on the context.
I believe I'm pretty familiar with the logic and if I found it a little confusing even with the understanding, wouldn't that be an indication that perhaps you could find a better, more intuitive name? :)

NealSun96: Thank you for the feedback. If you know the logic but still find it confusing, then that definitely is an indication. I coded this so I can't possibly have an objective view on it. 
While I agree that the current naming isn't the best, I'm not very keen about "closest node" either. Let me keep the discussion open so that there may be some second opinions coming in. I'll make sure to update the name before the PR is checked in. Thanks!

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Related to the previous conversation and will be updated along the previous part.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: Sure.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
narendly: @NealSun96 Let's add this in the Javadoc :)

> Yes it is. This scenario is when `path` is a sharding key that could be added to the trie. Consider a trie that only has "/a/b/c" in it:
> 
> 1. `getLongestPrefixNodeAlongPath("/a/b/c/x")` is the first case (typically used during realm lookups), which returns "/a/b/c";
> 2. `getLongestPrefixNodeAlongPath("/a/b")` is the second case (typically used during key scanning), which returns "/a/b";
> 3. `getLongestPrefixNodeAlongPath("/a/b/d")` is the last case, which returns "/a/b". Note that "/a/b/d" is a valid sharding key to add.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
NealSun96: As @pkuwm mentioned to me privately, he's able to understand the logic from the name. The method name will be kept as is without strong objections.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/TrieRoutingData.java
```diff
@@ -89,49 +87,55 @@ public String getMetadataStoreRealm(String path)
...
+      throw new IllegalArgumentException(
+          "Provided shardingKey is empty or does not have a leading \"" + DELIMITER
+              + "\" character: " + shardingKey);
+    }
+
+    TrieNode node = getLongestPrefixNodeAlongPath(shardingKey);
+    return !node.isShardingKey() && !node.getPath().equals(shardingKey);
   }
 
   /**
```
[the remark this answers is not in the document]
NealSun96: Interesting point. I have converted the javadoc to block comments.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/TestTrieRoutingData.java
```diff
@@ -249,7 +251,44 @@ public void testGetMetadataStoreRealmNoLeaf() {
...
+    }
+  }
+
+  @Test(dependsOnMethods = "testConstructionNormal")
+  public void testIsShardingKeyInsertionValidSlashOnly() {
+    Assert.assertFalse(_trie.isShardingKeyInsertionValid("/"));
+  }
+
+  @Test(dependsOnMethods = "testConstructionNormal")
+  public void testIsShardingKeyInsertionValidNormal() {
```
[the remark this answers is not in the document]
NealSun96: Good catch, I'll add validation for sharding key formats.

NealSun96:
Final commit message:
## Add validation logic to MSD write operations ##
This PR add routing data validation logic MSD writing operations to ensure the writes leave routing data in a valid state. TrieRoutingData logic is modified to reduce duplicate code.

Issue #752, opened 2020-02-11
Add left-over read MSDS endpoints to helix rest


pkuwm:
Related to #744  
There are some operations that are unsupported like 
getAllNamespaces
getAllMappingUnderPath
getMetadataStoreRealm
etc..
Note: endpoint of getAllNamespaces seems same as "/admin/v2/namespaces" but it is disabled when there is no default servlet/namespace in rest server. We need to figure out a way for this endpoint.

PR #761, opened 2020-02-14
Add REST read endpoints to helix-rest for metadata store directory

pkuwm:
### Issues

Implements #752 

### Description

We have metadata store directory service to help scale out zookeeper. Metadata store directory service provides REST APIs to access.
This PR adds remaining MSDS read endpoints to Helix REST.

### Tests

 - testGetRealmShardingKeysUnderPath
 - testGetShardingKeysUnderPath
 - testGetMetadataStoreRealmWithShardingKey
 - testGetAllNamespaces

```
[INFO] Tests run: 127, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 26.348 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 127, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:01 min
[INFO] Finished at: 2020-02-21T17:08:04-08:00
```
### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
@@ -143,7 +143,6 @@ public class MetadataStoreDirectoryAccessor extends AbstractResource {
   /**
    * Gets all sharding keys for following requests:
    * - "HTTP GET /sharding-keys" which returns all sharding keys in a namespace.
-   * - "HTTP GET /sharding-keys?realm={realm}" which returns sharding keys in the realm.
    * - "HTTP GET /sharding-keys?prefix={prefix}" which returns sharding keys that have the prefix.
    * -- JSON response example for this path:
    * {
@@ -163,31 +162,19 @@ public class MetadataStoreDirectoryAccessor extends AbstractResource {
    *  }]
    * }
    *
-   * - "HTTP GET /sharding-keys?realm={realm}&prefix={prefix}" which returns sharding keys in the
-   * realm and that have the given path as the prefix substring.
-   *
-   * @param realm Query param in endpoint path: metadata store realm.
    * @param prefix Query param in endpoint path: prefix substring of sharding key.
    * @return Json representation for the sharding keys.
    */
   @GET
   @Path("/sharding-keys")
-  public Response getShardingKeys(@QueryParam("realm") String realm,
-      @QueryParam("prefix") String prefix) {
+  public Response getShardingKeys(@QueryParam("prefix") String prefix) {
     try {
-      if (realm == null && prefix == null) {
+      if (prefix == null) {
         // For endpoint: "/sharding-keys" to get all sharding keys in a namespace.
         return getAllShardingKeys();
-      } else if (prefix == null) {
-        // For endpoint: "/sharding-keys?realm={realm}"
-        return getAllShardingKeysInRealm(realm);
-      } else if (realm == null) {
-        // For endpoint: "/sharding-keys?prefix={prefix}"
-        return getAllShardingKeysUnderPath(prefix);
-      } else {
-        // For endpoint: "/sharding-keys?realm={realm}&prefix={prefix}"
-        return getRealmShardingKeysUnderPath(realm, prefix);
       }
+      // For endpoint: "/sharding-keys?prefix={prefix}"
+      return getAllShardingKeysUnderPath(prefix);
     } catch (NoSuchElementException ex) {
       return notFound(ex.getMessage());
     }
@@ -195,17 +182,26 @@ public class MetadataStoreDirectoryAccessor extends AbstractResource {
 
   /**
    * Gets all path-based sharding keys for a queried realm at endpoint:
-   * "GET /metadata-store-realms/{realm}/sharding-keys". This endpoint is equivalent to
-   * the endpoint: "GET /sharding-keys?realm={realm}".
+   * "GET /metadata-store-realms/{realm}/sharding-keys"
+   * <p>
+   * "GET /metadata-store-realms/{realm}/sharding-keys?prefix={prefix}" is also supported,
+   * which is helpful when you want to check what sharding keys have the prefix substring.
    *
    * @param realm Queried metadata store realm to get sharding keys.
+   * @param prefix Query param in endpoint path: prefix substring of sharding key.
    * @return All path-based sharding keys in the queried realm.
    */
   @GET
   @Path("/metadata-store-realms/{realm}/sharding-keys")
-  public Response getRealmShardingKeys(@PathParam("realm") String realm) {
+  public Response getRealmShardingKeys(@PathParam("realm") String realm,
+      @QueryParam("prefix") String prefix) {
     try {
-      return getAllShardingKeysInRealm(realm);
+      if (prefix == null) {
+        return getAllShardingKeysInRealm(realm);
+      }
+
+      // For "GET /metadata-store-realms/{realm}/sharding-keys?prefix={prefix}"
+      return getRealmShardingKeysUnderPath(realm, prefix);
     } catch (NoSuchElementException ex) {
       return notFound(ex.getMessage());
     }
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
@@ -135,7 +135,7 @@ public class MetadataStoreDirectoryAccessor extends AbstractResource {
   }
 
   /**
-   * Gets all sharding keys mapped at paths:
+   * Gets all sharding keys for following requests:
    * - "HTTP GET /sharding-keys" which returns all sharding keys in a namespace.
    * - "HTTP GET /sharding-keys?realm={realm}" which returns sharding keys in the realm.
    * - "HTTP GET /sharding-keys?prefix={prefix}" which returns sharding keys that have the prefix.
@@ -158,7 +158,7 @@ public class MetadataStoreDirectoryAccessor extends AbstractResource {
    * }
    *
    * - "HTTP GET /sharding-keys?realm={realm}&prefix={prefix}" which returns sharding keys in the
-   * realm and that have the prefix.
+   * realm and that have the given path as the prefix substring.
    *
    * @param realm Query param in endpoint path: metadata store realm.
    * @param prefix Query param in endpoint path: prefix substring of sharding key.
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
@@ -94,14 +94,17 @@ public class MetadataStoreDirectoryAccessor extends AbstractResource {
   public Response getAllMetadataStoreRealms(@QueryParam("sharding-key") String shardingKey) {
     try {
       if (shardingKey == null) {
+        // Get all realms: "GET /metadata-store-realms"
         Collection<String> realms = _metadataStoreDirectory.getAllMetadataStoreRealms(_namespace);
         Map<String, Collection<String>> responseMap =
             ImmutableMap.of(MetadataStoreRoutingConstants.METADATA_STORE_REALMS, realms);
         return JSONRepresentation(responseMap);
-      } else {
-        String realm = _metadataStoreDirectory.getMetadataStoreRealm(_namespace, shardingKey);
-        return JSONRepresentation(new MetadataStoreShardingKey(shardingKey, realm));
       }
+
+      // Get a single realm filtered by sharding key:
+      // "GET /metadata-store-realms?sharding-key={sharding-key}"
+      String realm = _metadataStoreDirectory.getMetadataStoreRealm(_namespace, shardingKey);
+      return JSONRepresentation(new MetadataStoreShardingKey(shardingKey, realm));
     } catch (NoSuchElementException ex) {
       return notFound(ex.getMessage());
     }
@@ -179,6 +182,24 @@ public class MetadataStoreDirectoryAccessor extends AbstractResource {
         // For endpoint: "/sharding-keys?realm={realm}&prefix={prefix}"
         return getRealmShardingKeysUnderPath(realm, prefix);
       }
+    } catch (NoSuchElementException ex) {
+      return notFound(ex.getMessage());
+    }
+  }
+
+  /**
+   * Gets all path-based sharding keys for a queried realm at endpoint:
+   * "GET /metadata-store-realms/{realm}/sharding-keys". This endpoint is equivalent to
+   * the endpoint: "GET /sharding-keys?realm={realm}".
+   *
+   * @param realm Queried metadata store realm to get sharding keys.
+   * @return All path-based sharding keys in the queried realm.
+   */
+  @GET
+  @Path("/metadata-store-realms/{realm}/sharding-keys")
+  public Response getRealmShardingKeys(@PathParam("realm") String realm) {
+    try {
+      return getAllShardingKeysInRealm(realm);
     } catch (NoSuchElementException ex) {
       return notFound(ex.getMessage());
     }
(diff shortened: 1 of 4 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -66,24 +67,44 @@ private void postConstruct() {
...
-      Collection<String> realms = _metadataStoreDirectory.getAllMetadataStoreRealms(_namespace);
-
-      responseMap = new HashMap<>(1);
-      responseMap.put(MetadataStoreRoutingConstants.METADATA_STORE_REALMS, realms);
+      if (shardingKey == null) {
+        Collection<String> realms = _metadataStoreDirectory.getAllMetadataStoreRealms(_namespace);
+        Map<String, Collection<String>> responseMap =
+            ImmutableMap.of(MetadataStoreRoutingConstants.METADATA_STORE_REALMS, realms);
+        return JSONRepresentation(responseMap);
+      } else {
```
dasahcc: Remove else clause.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -66,24 +67,44 @@ private void postConstruct() {
...
+   * or gets a metadata store realm with the sharding key at path:
+   * "GET /metadata-store-realms?sharding-key={sharding-key}"
    *
    * @return Json representation of all realms.
    */
   @GET
   @Path("/metadata-store-realms")
-  public Response getAllMetadataStoreRealms() {
-    Map<String, Collection<String>> responseMap;
+  public Response getAllMetadataStoreRealms(@QueryParam("sharding-key") String shardingKey) {
```
dasahcc: Shall we define different response object instead of using plain Response.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -111,41 +132,56 @@ public Response deleteMetadataStoreRealm(@PathParam("realm") String realm) {
...
+        return getAllShardingKeysUnderPath(prefix);
       } else {
-        // For endpoint: "/sharding-keys?realm={realmName}"
-        shardingKeys = _metadataStoreDirectory.getAllShardingKeysInRealm(_namespace, realm);
-        // To avoid allocating unnecessary resource, limit the map's capacity only for
-        // SHARDING_KEYS and SINGLE_METADATA_STORE_REALM.
-        responseMap = new HashMap<>(2);
-        responseMap.put(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM, realm);
+        // For endpoint: "/sharding-keys?realm={realm}&prefix={prefix}"
+        return getRealmShardingKeysUnderPath(realm, prefix);
```
dasahcc: This logic is very complex. Can we leverage the default value to simplify the logic?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/constant/MetadataStoreRoutingConstants.java
```diff
@@ -28,14 +28,21 @@
...
+  public static final String SINGLE_METADATA_STORE_REALM = "realm";
 
   // Field name in JSON REST response of getting sharding keys.
   public static final String SHARDING_KEYS = "shardingKeys";
 
+  // Field name in JSON REST response related to one single sharding key.
+  public static final String SINGLE_SHARDING_KEY = "shardingKey";
 
+  // Field name in JSON REST response of getting sharding keys with prefix.
+  public static final String SHARDING_KEY_PREFIX = "prefix";
```
narendly: `SHARDING_KEY_PREFIX`'s naming got me thinking for a little bit and I couldn't figure it out. Could we rename this?

I think something like `SHARDING_KEY_PATH_PREFIX` would be more descriptive. To also help with understanding, could we please add where it's going to be used? (like getAllMappingUnderPath() for example).

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
narendly: Is this else needed?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -66,24 +67,44 @@ private void postConstruct() {
...
+    Map<String, Collection<String>> responseMap =
+        ImmutableMap.of(MetadataStoreRoutingConstants.METADATA_STORE_NAMESPACES, namespaces);
+
+    return JSONRepresentation(responseMap);
+  }
+
+  /**
+   * Gets all metadata store realms in a namespace at path: "GET /metadata-store-realms",
+   * or gets a metadata store realm with the sharding key at path:
+   * "GET /metadata-store-realms?sharding-key={sharding-key}"
```
narendly: could we also document the difference in behavior depending on whether the query param is given?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: I've thought about this carefully. As it is necessary, I've defined a POJO class `MetadataStoreShardingKey` to represent the object mapping. The response will look like:
```
{
	"metadataStoreRealms": ["realm-0", "realm-1"]
}
```
or  single ream with sharding key:
```
{
	"shardingKey": "/a/b/c",
	"realm": "realm-0"
}
```
I definitely agree that if the response of object is complex (has more fields), we would need POJO classes to represent the responses.
For the time being, since the most of the responses are quite simple: a list, I don't think we want to make it complex .  I've tried to make all json responses to have key-value mappings.

narendly: Since this path resource seems to be doing two things, should we rename this to `getMetadataStoreRealm()`?

narendly: Or better yet, should we consider breaking this up into two different path resources to eliminate if-elses?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -111,41 +132,56 @@ public Response deleteMetadataStoreRealm(@PathParam("realm") String realm) {
...
+   * @param realm Query param in endpoint path: metadata store realm.
+   * @param prefix Query param in endpoint path: prefix substring of sharding key.
+   * @return Json representation for the sharding keys.
    */
   @GET
   @Path("/sharding-keys")
-  public Response getShardingKeys(@QueryParam("realm") String realm) {
-    Map<String, Object> responseMap;
-    Collection<String> shardingKeys;
+  public Response getShardingKeys(@QueryParam("realm") String realm,
```
narendly: I have a feeling that this path resource is again doing too many things. Do you think we should consider breaking this up into different path resources?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: I think the endpoint already tells the behavior. That's why we need to follow the REST API design principal.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: `?sharding-key=` is a query param to filter the result. In this case, a single object being returned is acceptable.
It is like
`/employees` returns all employees
`/employees?name=david` could returns a single employee (assume name is unique).

pkuwm: My explanation above. It is OK to have if-else because we accept query params. A better way to process query param is like we construct a query for DB:
```
query += param1,
query += param2
```
But we don't have a query layer here so if-else acts as this query layer.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: A new endpoint which is equivalent to `GET /sharding-keys?realm={realm}`  is added: `GET /metadata-store-realms/{realm}/sharding-keys`

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: A new endpoint which is equivalent to GET /sharding-keys?realm={realm} is added: GET /metadata-store-realms/{realm}/sharding-keys

In my opinion, these 2 filters are valid for this resource. If remove one query param, this method would only have 1 query param `prefix` and 1 if-else: all sharding keys or sharding keys with prefix. 
But again, I would like to offer this query param option as they are valid, and it is also implemented now.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -111,41 +135,74 @@ public Response deleteMetadataStoreRealm(@PathParam("realm") String realm) {
...
+   * @param realm Queried metadata store realm to get sharding keys.
+   * @return All path-based sharding keys in the queried realm.
+   */
+  @GET
+  @Path("/metadata-store-realms/{realm}/sharding-keys")
+  public Response getRealmShardingKeys(@PathParam("realm") String realm) {
+    try {
+      return getAllShardingKeysInRealm(realm);
+    } catch (NoSuchElementException ex) {
+      return notFound(ex.getMessage());
```
NealSun96: Given my understanding to the status code, `badRequest` is more suitable than `notFound` in this situation. What `notFound` entails is that this url is not matched to anything, however in this situation, the url is correctly connecting to the endpoint. Returning `notFound` might cause confusion. 
I understand that `badRequest` is not perfect in this case either, but it should be more suitable than `notFound`. If you have a better idea, I'm happy to discuss. (Ref: https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html)

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: There is a good way to simplify the logic here utilizing the default values as @dasahcc has mentioned. If the prefix is not provided, consider assigning "/" to the prefix variable and use the same logic as the situation where prefix exists.

helix-rest/src/test/java/org/apache/helix/rest/server/resources/zookeeper/TestMetadataStoreDirectoryAccessor.java
```diff
@@ -110,7 +110,33 @@ public void beforeClass() throws Exception {
...
+    String responseBody = get(TEST_NAMESPACE_URI_PREFIX + "/metadata-store-namespaces", null,
+        Response.Status.OK.getStatusCode(), true);
+
+    // It is safe to cast the object and suppress warnings.
+    @SuppressWarnings("unchecked")
+    Map<String, Collection<String>> queriedNamespacesMap =
+        OBJECT_MAPPER.readValue(responseBody, Map.class);
+
+    Assert.assertTrue(
+        queriedNamespacesMap.containsKey(MetadataStoreRoutingConstants.METADATA_STORE_NAMESPACES));
```
NealSun96: When verifying a map, I'd make sure that "it contains the right elements and **only** the right elements". Therefore I usually perform a check on map length - without it if the map returns something extra and unexpected, the test cases would still incorrectly pass.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: @NealSun96 Thanks for reviewing.
As you referred in the link, `400 Bad Request`: The request could not be understood by the server due to **malformed syntax**. If you request a POST/PUT with json payload which has syntax error, I believe `Bad Request` is more suitable.
`404 Not Found` means the endpoint is correct but resource is not found in server, which matches exactly our case here `NoSuchElementException`.  (refs as an examples: 
- https://stackoverflow.com/questions/26968479/404-not-found-or-bad-request;
- https://stackoverflow.com/questions/19671317/400-bad-request-http-error-code-meaning)

What do you think?

helix-rest/src/test/java/org/apache/helix/rest/server/resources/zookeeper/TestMetadataStoreDirectoryAccessor.java
pkuwm: Sure. I was checking the response has this field. It is definitely more accurate to check that the response has only expected fields.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: > There is a good way to simplify the logic here utilizing the default values as @dasahcc has mentioned. If the prefix is not provided, consider assigning "/" to the prefix variable and use the same logic as the situation where prefix exists.

@NealSun96 I don't think adding "/" to the prefix var is a better idea because:
- The Json responses are different for each case here. If you assign "/" to a method, the response could have `{ "prefix": "/" }`
- I've tried to find a good default value for this prefix. If I assign "/" as a default value, I've no idea how to differentiate if prefix is not provide or it is "/".
- The code logic to generate response map is different.

If I don't understand your idea, feel free to clarify it. Thanks for the tip.

NealSun96: When the prefix parameter is not provided, it's treated as the caller doesn't care about the prefix - this method is doing "return all keys regardless of any prefix". This behavior is functionally the same as if the caller provides "/" as the prefix. I think it's reasonable to use "/" as a default value.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: `404` does not mean the endpoint is correct. If I enter an end point "/metadata-store-realms/test-realm/sharding-keys" and I receive a `404` in return, my first reaction would be to change "sharding-keys" to "sharding-key", then "shardingkeys". 
On a second thought, `404` with appropriate message does work better than `400`. `400` is not very fitting.

pkuwm: In your example, the **endpoint** is `/metadata-store-realms/{realm}/sharding-keys` is already correct. But the resource `test-realm` does not exist, and 404 is a **correct** response code. If endpoint is incorrect, you would not get 404 but 400 (bad URI) instead.
So if you get 404, the first reaction is definitely checking the requested resource, not changing the endpoint.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
narendly: public static final String ROUTING_DATA_PATH = "/METADATA_STORE_ROUTING_DATA";

This should be the default value.

pkuwm:

```
We need restful metadata store directory service to help scale out zookeeper.

This commit adds REST read endpoints to helix-rest to get sharding keys, realms and namespaces.
```

PR #765, opened 2020-02-14
Add DedicatedZkClient and update DedicatedZkClientFactory

### LINKED ISSUE #766: Add DedicatedZkClient and update DedicatedZkClientFactory
As part of ZkClient API enhancement, we wish to add DedicatedZkClient, which is a wrapper of the raw ZkClient, that provides realm-aware access to ZooKeeper.

Realm-aware in that it only performs requests whose path's Zk path sharding key belongs to the ZK realm it's connected to.

Also, we need to modify DedicatedZkClientFactory so that users could use this factory to generate instances of DedicatedZkClient.

narendly:
### Issues

Fixes #766 

### Description

As part of ZkClient API enhancement, we wish to add DedicatedZkClient, which is a wrapper of the raw ZkClient, that provides realm-aware access to ZooKeeper.

Realm-aware in that it only performs requests whose path's Zk path sharding key belongs to the ZK realm it's connected to.

Also, we need to modify DedicatedZkClientFactory so that users could use this factory to generate instances of DedicatedZkClient.

### Tests

RealmAwareZkClientTestBase
TestDedicatedZkClient

> [INFO] Results:
> [INFO] 
> [INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
> [INFO] 
> [INFO] BUILD SUCCESS
> [INFO] Total time:  5.256 s
> [INFO] Finished at: 2020-02-20T16:00:28-08:00

### Commits

### Code Quality


zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/SharedZkClientFactory.java
```diff
@@ -44,15 +44,17 @@ protected SharedZkClientFactory() {
   @Override
   public RealmAwareZkClient buildZkClient(
       RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig,
-      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig) {
+      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig,
```
dasahcc: This is PR for dedicated zkclient factory, right?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/TrieRoutingData.java
```diff
@@ -0,0 +1,277 @@
...
+
+
+/**
+ * TODO: remove when there's a separate module.
+ * This is a class that uses a data structure similar to trie to represent metadata store routing
+ * data. It is not exactly a trie because it in essence stores a mapping (from sharding keys to
+ * realm addresses) instead of pure text information; also, only the terminal nodes store meaningful
+ * information (realm addresses).
+ */
+public class TrieRoutingData implements MetadataStoreRoutingData {
```
dasahcc: Is this new code or refactor? If refactor, we should remove the old code in REST module, right?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/SharedZkClientFactory.java
narendly: There was a change in the interface. SharedZkClient implements that interface. So this change is just a signature change, not logical change. 

This change is necessary.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/TrieRoutingData.java
narendly: This class will be part of metadata-store-directory-common module. 
Once https://github.com/apache/helix/pull/771 is checked in, this class will be removed.

We cannot directly use this class because that would create a circular dependency.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,592 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * NOTE: DO NOT USE THIS CLASS DIRECTLY. Use DedicatedZkClientFactory to create instances of DedicatedZkClient.
+ *
+ * An implementation of the RealmAwareZkClient interface.
+ * Supports CRUD, data change subscription, and ephemeral mode operations.
+ */
+public class DedicatedZkClient implements RealmAwareZkClient {
```
pkuwm: If we don't want users to directly use this class, shall we consider refactoring the code structure so we can make this class `protected`? I think that may be better than just commenting not to use it directly.

narendly: Good thinking. But we agreed to make the code structure so that factory classes and implementation classes in separate packages, so we cannot make this protected. This was done to better organize the code.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,592 @@
...
+    _rawZkClient.asyncDelete(path, cb);
+  }
+
+  @Override
+  public void watchForData(String path) {
+    if (!checkIfPathBelongsToZkRealm(path)) {
+      throw new IllegalArgumentException(
+          "The given path does not map to the ZK realm for this DedicatedZkClient! Path: " + path
+              + " ZK realm sharding key: " + _zkRealmShardingKey);
+    }
```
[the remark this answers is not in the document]
narendly: 1. We do not modify the raw ZkClient, even though it might save some duplicate code.
2. We have different logic for checking whether a request is valid, so there is less duplication now.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -636,7 +636,7 @@ public String create(final String path, Object datat, final List<ACL> acl, final
    * @throws IllegalArgumentException if called from anything else except the ZooKeeper event thread
    * @throws ZkException if any zookeeper exception occurs
    */
-  private String create(final String path, final Object dataObject, final List<ACL> acl,
+   public String create(final String path, final Object dataObject, final List<ACL> acl,
```
kaisun2000: Why change this implementation? This is interface change.

pkuwm: @kaisun2000 This is not an interface change. This method is the base method for node creation. When I implemented it, I intentionally made it private not to allow direct call of `create(CreateMode.persistent, expectedSessionId)`.
I also had the question for changing it public. But here, the implementation needs to call `rawZkClient.create()` in `DedicatedZkClient` which is put in a different package. And this needs `public`. If we could put `DedicatedZkClient` in a same package with `ZkClient`, we don't need it public.

narendly: @pkuwm Said it.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,488 @@
...
+    _zkRealmShardingKey = connectionConfig.getZkRealmShardingKey();
+
+    if (metadataStoreRoutingData == null) {
+      throw new IllegalArgumentException("MetadataStoreRoutingData cannot be null!");
+    }
+    _metadataStoreRoutingData = metadataStoreRoutingData;
+
+    // Get the ZkRealm address based on the ZK path sharding key
+    String zkRealmAddress = _metadataStoreRoutingData.getMetadataStoreRealm(_zkRealmShardingKey);
```
NealSun96: Nit: could assign to _zkRealmAddress first and then perform the check

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/SharedZkClientFactory.java
```diff
@@ -44,15 +44,17 @@ protected SharedZkClientFactory() {
...
     // TODO: Implement the logic
     // Return an instance of SharedZkClient
     return null;
   }
 
   @Override
   public RealmAwareZkClient buildZkClient(
-      RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig) {
+      RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig,
+      MetadataStoreRoutingData metadataStoreRoutingData) {
```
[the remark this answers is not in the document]
narendly: @kaisun2000 Good point. As we discussed in person, this is only put in here to make the code build. We will remove this from the constructor in a future PR when the global routing data is available.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/TrieRoutingData.java
NealSun96: This is no longer the latest `TrieRoutingdata` after the changes introduced in: https://github.com/apache/helix/commit/8c050b3928af3e4076311744076dcc75cc21e602 .

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestDedicatedZkClient.java
```diff
@@ -0,0 +1,154 @@
...
+    _realmAwareZkClient.writeData(TEST_VALID_PATH, znRecord);
+    Assert.assertEquals(_realmAwareZkClient.readData(TEST_VALID_PATH), znRecord);
+
+    // Test writing and reading against the invalid path
+    try {
+      _realmAwareZkClient.createPersistent(TEST_INVALID_PATH, true);
+      _realmAwareZkClient.writeData(TEST_INVALID_PATH, znRecord);
+      Assert.fail("Create() should not succeed on an invalid path!");
+    } catch (IllegalArgumentException e) {
+      // Okay - expected
```
pkuwm: It seems `IllegalArgumentException` will be thrown for `createPersistent()`, and there is no check for `writeData()`. If we want to test `writeData()` as well, shall we have another validation for `writeData()` to be more accurate?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/TrieRoutingData.java
narendly: This class will be removed after the module separation. Please see the TODO.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestDedicatedZkClient.java
narendly: This is a test for createPersistent(). I added `_realmAwareZkClient.writeData(TEST_INVALID_PATH, znRecord);` for parity with the test I did for the valid path.
_realmAwareZkClient.writeData(TEST_INVALID_PATH, znRecord); doesn't make sense anyway because the ZNode should never be created at that path.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,480 @@
...
+  public <T> T readData(String path, boolean returnNullIfPathNotExists) {
+    T data = null;
+    try {
+      return readData(path, null);
+    } catch (ZkNoNodeException e) {
+      if (!returnNullIfPathNotExists) {
+        throw e;
+      }
+    }
+    return data;
```
pkuwm: Shall we just call `_rawZkClient.readData(path, returnNullIfPathNotExists)`, so it is easier to maintain the code because it relies on raw ZkClient, which is our initial purpose?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,480 @@
...
+  public <T> T readDataAndStat(String path, Stat stat, boolean returnNullIfPathNotExists) {
+    T data = null;
+    try {
+      data = readData(path, stat);
+    } catch (ZkNoNodeException e) {
+      if (!returnNullIfPathNotExists) {
+        throw e;
+      }
+    }
+    return data;
```
pkuwm: Same, to call raw zkClient?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,480 @@
...
+  @Override
+  public List<String> getChildren(String path) {
+    checkIfPathContainsShardingKey(path);
+    return _rawZkClient.getChildren(path);
+  }
+
+  @Override
+  public int countChildren(String path) {
+    checkIfPathContainsShardingKey(path);
+    return countChildren(path);
```
pkuwm: Typo? `_rawZkClient.countChildren()` Otherwise, a stack overflow :)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,480 @@
...
+
+  @Override
+  public void createEphemeral(String path, List<ACL> acl) {
+    create(path, null, acl, CreateMode.EPHEMERAL);
+  }
+
+  @Override
+  public void createEphemeral(String path, List<ACL> acl, String sessionId) {
+    checkIfPathContainsShardingKey(path);
+    _rawZkClient.createEphemeral(path, acl, sessionId);
```
pkuwm: `create(path, null, acl, CreateMode.EPHEMERAL, sessionId);` to re-use the code and minimize the code branches to reduce duplicate code and maintenance effort? For example, if we want to change the behavior, we just need to change `create()`.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,480 @@
...
+
+  @Override
+  public void createEphemeral(String path, Object data) {
+    create(path, data, CreateMode.EPHEMERAL);
+  }
+
+  @Override
+  public void createEphemeral(String path, Object data, String sessionId) {
+    checkIfPathContainsShardingKey(path);
+    _rawZkClient.createEphemeral(path, data, sessionId);
```
pkuwm: Same: call its own`create()`?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,480 @@
...
+
+  @Override
+  public void createEphemeral(String path, Object data, List<ACL> acl) {
+    create(path, data, acl, CreateMode.EPHEMERAL);
+  }
+
+  @Override
+  public void createEphemeral(String path, Object data, List<ACL> acl, String sessionId) {
+    checkIfPathContainsShardingKey(path);
+    _rawZkClient.createEphemeral(path, data, acl, sessionId);
```
pkuwm: Same: call its owncreate()?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,480 @@
...
+
+  @Override
+  public String createEphemeralSequential(String path, Object data, List<ACL> acl) {
+    return create(path, data, acl, CreateMode.EPHEMERAL_SEQUENTIAL);
+  }
+
+  @Override
+  public String createEphemeralSequential(String path, Object data, String sessionId) {
+    checkIfPathContainsShardingKey(path);
+    return _rawZkClient.createEphemeralSequential(path, data, sessionId);
```
pkuwm: Same: call its `owncreate()`?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestDedicatedZkClient.java
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.helix.zookeeper.impl.factory.DedicatedZkClientFactory;
+import org.apache.helix.zookeeper.impl.factory.MetadataStoreRoutingData;
+import org.apache.helix.zookeeper.impl.factory.TrieRoutingData;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+
+public class TestDedicatedZkClient extends ZkTestBase {
+  private static final String ZK_SHARDING_KEY_PREFIX = "/TEST_SHARDING_KEY_";
```
pkuwm: Nit, it seems each time concatenating the path, underscore is added and so the paths would be `/TEST_SHARDING_KEY__0/a/b`, which does not affect the results but it may not be what you wanted to have in comment?
```   
String validShardingKey = ZK_SHARDING_KEY_PREFIX + "_" + 0; // Use TEST_SHARDING_KEY_0
```

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/ZkTestBase.java
```diff
@@ -0,0 +1,148 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(ZkTestBase.class);
+  private static final MBeanServerConnection MBEAN_SERVER =
+      ManagementFactory.getPlatformMBeanServer();
+
+  // maven surefire-plugin's multiple ZK config keys
+  private static final String MULTI_ZK_PROPERTY_KEY = "multiZk";
+  private static final String NUM_ZK_PROPERTY_KEY = "numZk";
+
+  protected static final String ZK_PREFIX = "localhost:";
+  protected static final int ZK_START_PORT = 2183;
```
pkuwm: Current ZkTestBase in helix-core is occupying this port. If we run the tests in the whole project and port 2183 is already occupied by ZK started by helix-core, can we still start a ZK at this port 2183? If not, we may want to use another port.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -636,7 +636,7 @@ public String create(final String path, Object datat, final List<ACL> acl, final
    * @throws IllegalArgumentException if called from anything else except the ZooKeeper event thread
    * @throws ZkException if any zookeeper exception occurs
    */
-  private String create(final String path, final Object dataObject, final List<ACL> acl,
+   private String create(final String path, final Object dataObject, final List<ACL> acl,
```
pkuwm: Nit, Do you want it to be `public` or `private`? If private, no need to add the space.

narendly: Fixed

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestDedicatedZkClient.java
narendly: Fixed.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/ZkTestBase.java
narendly: Changed it to a different port.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
narendly: This will be removed as noted in the TODO

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
narendly: As discussed offline, we'll just leave it as is. This is to completely mimic what the raw ZkClient is doing.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
narendly: Replied above

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
narendly: Replied above

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
narendly: Good catch!

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestDedicatedZkClient.java
```diff
@@ -0,0 +1,154 @@
...
+          .buildZkClient(connectionConfig, clientConfig, _metadataStoreRoutingData);
+      Assert.fail("Should not succeed with an invalid sharding key!");
+    } catch (IllegalArgumentException e) {
+      // Expected
+    }
+
+    // Use a valid sharding key this time around
+    String validShardingKey = ZK_SHARDING_KEY_PREFIX + "_" + 0; // Use TEST_SHARDING_KEY_0
+    connectionConfig = new RealmAwareZkClient.RealmAwareZkConnectionConfig(validShardingKey);
+    _realmAwareZkClient = DedicatedZkClientFactory.getInstance()
```
pkuwm: Is this client supposed to be close after class?

narendly: Yes. Will close.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/ZkTestBase.java
pkuwm: Double checked to verify my concern is valid. I think if we don't change the port, this zk could not be started at this port, because testng.xml in helix-core includes all helix packages: `<package name="org.apache.helix.*"/>` 

I think there are 3 options to solve this:
1. change the port to use a different one -- OK. but could create more zk servers during the tests.
2. if the port is already occupied by a running zk, we don't assert fail it but use current zk. -- Not that good. Could not get the zk server reference.
3. exclude `zookeeper` and also `rest` module in helix-core's testng.xml so all ZK servers would be shut down after all helix-core's tests. So zookeeper module can create its own ZK servers on the port 2183. -- I prefer this one so we don't create more zk servers during the tests. Cons: it takes time when restarting zk servers.

narendly: Changed the port so this is no longer a concern.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestDedicatedZkClient.java
```diff
@@ -0,0 +1,161 @@
...
+import org.apache.helix.zookeeper.impl.factory.DedicatedZkClientFactory;
+import org.apache.helix.zookeeper.impl.factory.MetadataStoreRoutingData;
+import org.apache.helix.zookeeper.impl.factory.TrieRoutingData;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+
+public class TestDedicatedZkClient extends ZkTestBase {
```
pkuwm: Since `checkIfPathContainsShardingKey()` is critical, can we add a test to cover this: valid path and all invalid path cases?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,465 @@
...
+    return _rawZkClient.getZkSerializer();
+  }
+
+  /**
+   * Checks whether the given path belongs matches the ZK path sharding key this DedicatedZkClient is designated to at initialization.
+   * @param path
+   * @return
+   */
+  private void checkIfPathContainsShardingKey(String path) {
+    if (!path.startsWith(_zkRealmShardingKey)) {
```
pkuwm: If sharding key is `/abcde`, are these paths expected to pass this check and supposed to use this zkclient:
- `/abcde_0/xxx`
- `/abcde/abcde-0/xxx`

narendly: Good catch! Let me fix this up.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestDedicatedZkClient.java
narendly: We will be using MetadataStoreRoutingData (the trie-like data structure) for that check. We have test coverage on this component already, so we won't need to test here again.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/factory/RealmAwareZkClientFactory.java
```diff
@@ -30,16 +31,21 @@
...
   /**
    * Builds a RealmAwareZkClient using specified connection config and default client config.
    * @param connectionConfig
+   * @param metadataStoreRoutingData
    * @return RealmAwareZkClient
    */
-  RealmAwareZkClient buildZkClient(
-      RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig);
+  // TODO: remove MetadataStoreRoutingData
+  RealmAwareZkClient buildZkClient(RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig,
```
kaisun2000: Nit. We can provide a default implementation of this one. The default implementation would call the above one by constructing an default  RealmAwareZkClient.RealmAwareZkClientConfig.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
kaisun2000: There are other possible way to fix it and reason it would work.

For example the prefix must be normalized to end with "/" before the match against input. 

Either way, this implementation detail can be fix later.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/ZkTestBase.java
```diff
@@ -0,0 +1,148 @@
+package org.apache.helix.zookeeper.impl;
```
kaisun2000: There is ZkTestBase in Helix common before. So what is the strategy here. 
1/ Are we going to remove that ZkTestBase from Helix common and associated test?
2/ Move ZkTestBase in Helix common and associated test here?
3/ Have two ZKTestBase s and their related test in parallel?

narendly: 1. No, this is a ZkTestBase for zookeeper-api module. The name of the class is the same, but it does different things.
2. No.
3. Yes. zookeeper-api needs to have tests and this ZkTestBase contains the ZK environment setup logic.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
narendly: We don't have all the components implemented to make this work yet. 

@kaisun2000 

> For example the prefix must be normalized to end with "/" before the match against input.

This may not be necessary since we'll have the trie-like routing data already. That way, we don't implement another line of the check logic and we are able to re-use a tested component.

At any rate, I'll put a TODO and let's revisit this method when we have all the components ready.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/RealmAwareZkClientTestBase.java
```diff
@@ -0,0 +1,164 @@
+package org.apache.helix.zookeeper.impl.client;
```
kaisun2000: As a general comment. As we took the approach to composition (aka create new class DedicatedZkClient), theoretically, it would be idea to still test the new method even it is a thin wrapper. 

If we choose the approach of modifying the ZKClient, since the code is not change, we can reason that limiting our testing to only the modified part, namely the checkingPath validity part would be enough.

Considering composition is only a thin wrapper, I don't see the necessity to do it right now. However, later if we change the code of DedicatedZkClient implementation more, we should add more test then.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/ZkTestBase.java
```diff
@@ -0,0 +1,148 @@
...
+import org.slf4j.LoggerFactory;
+import org.testng.Assert;
+import org.testng.annotations.AfterSuite;
+import org.testng.annotations.BeforeSuite;
+
+
+/**
+ * Test base class for various integration tests with an in-memory ZooKeeper.
+ */
+public class ZkTestBase {
```
dasahcc: This is same class as helix-core, right? Since helix-core refers this module, can we remove the test base in the core?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/RealmAwareZkClientTestBase.java
narendly: Do you think you'd be able to enhance the tests to our `RealmAwareZkClientTestBase.java` after SharedZkClient is complete? @kaisun2000

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/ZkTestBase.java
narendly: @dasahcc 
their names are the same but these are two different test bases.

helix-core's TestBase contains a lot of Helix logic. this ZkTestBase is only for zookeeper-related tests and therefore has no Helix-related logic.

In short, we shouldn't remove the test base in the core.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/RealmAwareZkClientTestBase.java
narendly: Discussed offline; @kaisun2000  to add more test cases for SharedZkClient (regarding ephemeral node cases).

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -0,0 +1,473 @@
...
+  }
+
+  @Override
+  public void createPersistent(String path) {
+    createPersistent(path, false);
+  }
+
+  @Override
+  public void createPersistent(String path, boolean createParents) {
+    createPersistent(path, createParents, ZooDefs.Ids.OPEN_ACL_UNSAFE);
```
kaisun2000: `checkIfPathContainsShardingKey(path);
_rawZkClient.createPersistent(path, createParents);
`
Delegate to other public API is anti-pattern of composition and delegation or sometime called PIMPL in C++ world. It make harder to reason if this is keeping the _rawZkClient's behavior or not. In fact, if later _rawZkClient implementation is changed, you may have unexpected behavior.

PR #770, opened 2020-02-16
WIP: Add SharedZkClient and update SharedZkClientFactory

### LINKED ISSUE #762: Add  RealmAware SharedZkFactory and SharedZkClient implementation
Implement the functionality to create realm aware dedicated/shared ZKClient using SharedZKClientFactory and DedicatedZKClientFactory which already exists.

narendly:
**Note: this will be continued at https://github.com/apache/helix/pull/796 **

### Issues

Fixes #762 

### Description

As part of ZkClient API enhancement, we wish to add SharedZkClient, which is a wrapper of the raw ZkClient, that provides realm-aware access to ZooKeeper.

Realm-aware in that it only performs requests whose path's Zk path sharding key belongs to the ZK realm it's connected to.

Also, we need to modify SharedZkClientFactory so that users could use this factory to generate instances of SharedZkClient.

### Tests

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
@@ -21,7 +21,6 @@ package org.apache.helix.zookeeper.api.client;
 
 import org.apache.helix.zookeeper.zkclient.serialize.BasicZkSerializer;
 import org.apache.helix.zookeeper.zkclient.serialize.PathBasedZkSerializer;
-import org.apache.helix.zookeeper.zkclient.serialize.SerializableSerializer;
 import org.apache.helix.zookeeper.zkclient.serialize.ZkSerializer;
 
 
@@ -42,7 +41,7 @@ public interface HelixZkClient extends RealmAwareZkClient {
   class ZkConnectionConfig {
     // Connection configs
     private final String _zkServers;
-    private int _sessionTimeout = HelixZkClient.DEFAULT_SESSION_TIMEOUT;
+    private int _sessionTimeout = DEFAULT_SESSION_TIMEOUT;
 
     public ZkConnectionConfig(String zkServers) {
       _zkServers = zkServers;
@@ -53,10 +52,10 @@ public interface HelixZkClient extends RealmAwareZkClient {
       if (obj == this) {
         return true;
       }
-      if (!(obj instanceof HelixZkClient.ZkConnectionConfig)) {
+      if (!(obj instanceof ZkConnectionConfig)) {
         return false;
       }
-      HelixZkClient.ZkConnectionConfig configObj = (HelixZkClient.ZkConnectionConfig) obj;
+      ZkConnectionConfig configObj = (ZkConnectionConfig) obj;
       return (_zkServers == null && configObj._zkServers == null || _zkServers != null && _zkServers
           .equals(configObj._zkServers)) && _sessionTimeout == configObj._sessionTimeout;
     }
@@ -71,7 +70,7 @@ public interface HelixZkClient extends RealmAwareZkClient {
       return (_zkServers + "_" + _sessionTimeout).replaceAll("[\\W]", "_");
     }
 
-    public HelixZkClient.ZkConnectionConfig setSessionTimeout(Integer sessionTimeout) {
+    public ZkConnectionConfig setSessionTimeout(Integer sessionTimeout) {
       this._sessionTimeout = sessionTimeout;
       return this;
     }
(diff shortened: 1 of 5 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
```diff
@@ -89,32 +88,17 @@ public int getSessionTimeout() {
...
-
-    // Others
-    private PathBasedZkSerializer _zkSerializer;
-
-    // Monitoring
-    private String _monitorType;
-    private String _monitorKey;
-    private String _monitorInstanceName = null;
-    private boolean _monitorRootPathOnly = true;
+  class ZkClientConfig extends RealmAwareZkClientConfig {
```
kaisun2000: This is the addressing the previous ZkClientConfig "duplicate code" issue, right? What is the reason that previous it does not work.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
```diff
@@ -89,32 +88,17 @@ public int getSessionTimeout() {
...
 
+    @Override
```
kaisun2000: Why do we need to re-implement these methods if you inherit from the RealmAwareZkClientConfig?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
narendly: As I explained, the issue was trying to

> cast superclass to subclass which is not allowed

I had to override so that these setters return an instance of HelixZkClient.ZkClientConfig instead of RealmAwareZkClientConfig. I suggest thinking more carefully about the return types and their inheritance relationships.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/HelixZkClient.java
narendly: As I explained, the issue was trying to

> cast superclass to subclass which is not allowed

I had to override so that these setters return an instance of HelixZkClient.ZkClientConfig instead of RealmAwareZkClientConfig. I suggest thinking more carefully about the return types and their inheritance relationships.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/SharedZkClientFactory.java
```diff
@@ -125,4 +132,74 @@ public int getActiveConnectionCount() {
...
+     * Triggered after the SharedZkClient is closed.
+     */
+    void onClose();
+  }
+
+  /**
+   * NOTE: do NOT use this class directly. Please use SharedZkClientFactory to create an instance of SharedZkClient.
+   * InnerSharedZkClient is a ZkClient used by SharedZkClient to power ZK operations against a single ZK realm.
+   */
+  public static class InnerSharedZkClient extends ZkClient implements HelixZkClient {
```
kaisun2000: So basically InnerSharedZkClient is exactly the original SharedZkClient. Is the code exactly the same?

kaisun2000: This is actually what I worried, the code here changed. How do we reason that the new implementation of InnerSharedZkClient keeps the old InnerSharedZkClient original behavior? 
@jiajunwang can you have a look?

narendly: @kaisun2000 
InnerSharedZkClient is meant to preserve the behavior/logic of the original SharedZkClient.

1. Clearly the code is not "exactly" the same. You have access to the original SharedZkClient - please do a comparison.
2. What is the exact concern here? I'm not exactly sure what your question is.

kaisun2000: I had a careful look, this seems to be OK to maintain the original behavior.  Assertion 1.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/SharedZkClient.java
```diff
@@ -37,79 +50,541 @@
...
+
+    // Get the ZkRealm address based on the ZK path sharding key
+    String zkRealmAddress = _routingDataCache.get(_zkRealmShardingKey);
+
+    // Create an InnerSharedZkClient to actually serve ZK requests
+    // TODO: Rename HelixZkClient in the future or remove it entirely - this will be a backward-compatibility breaking change because HelixZkClient is being used by Helix users.
+    HelixZkClient.ZkConnectionConfig zkConnectionConfig =
+        new HelixZkClient.ZkConnectionConfig(zkRealmAddress)
+            .setSessionTimeout(connectionConfig.getSessionTimeout());
+    _innerSharedZkClient = SharedZkClientFactory.getInstance()
```
kaisun2000: This actually implies the ZKRealmAware SharedZKClient are mixed with the original SharedZKCLient (now called innerSharedZKClient). We should note this behavior.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/SharedZkClient.java
```diff
@@ -37,79 +50,541 @@
...
+  }
+
+  @Override
+  public long getSessionId() {
+    return _innerSharedZkClient.getSessionId();
+  }
+
+  @Override
+  public void close() {
+    _innerSharedZkClient.close();
```
kaisun2000: This should be fine assuming assertion 1, innerSharedZKClient keeps the original behavior. Namely the last closed one will close the underlying ZkConnectionManager.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/SharedZkClientFactory.java
```diff
@@ -82,13 +94,8 @@ public HelixZkClient buildZkClient(HelixZkClient.ZkConnectionConfig connectionCo
...
       }
       LOG.info("Sharing ZkConnection {} to a new SharedZkClient.", connectionConfig.toString());
-      return new SharedZkClient(zkConnectionManager, clientConfig,
-          new SharedZkClient.OnCloseCallback() {
-            @Override
-            public void onClose() {
-              cleanupConnectionManager(zkConnectionManager);
-            }
-          });
+      return new InnerSharedZkClient(zkConnectionManager, clientConfig,
```
kaisun2000: nit: the intention is to keeper old sharedZkClient the same, but use new name innerSharedZkClient.  let us don't use the new Java feature to rewrite the same logic. Just leave it as what it were. It would be easier to people to look at the history to understand the intention.

PR #775, opened 2020-02-18
Add HttpRoutingDataReader

### LINKED ISSUE #774: Add HttpRoutingDataReader
HttpRoutingDataReader is a component used by new ZkClient APIs to make an HTTP read request to the metadata store directory service to retrieve routing data. ZkClient APIs will construct an internal MetadataStoreRoutingData instance based on the raw routing data retrieved from MSDS.

narendly:
### Issues

Resolves #774 

### Description

HttpRoutingDataReader is a component used by new ZkClient APIs to make an HTTP read request to the metadata store directory service to retrieve routing data. ZkClient APIs will construct an internal MetadataStoreRoutingData instance based on the raw routing data retrieved from MSDS.

Note: this change contains modifications to MockMSDS because the actual endpoint names changed. The methods and http server contexts have been updated.

### Tests

TestHttpRoutingDataReader

zookeeper-api

> [INFO]  T E S T S
> [INFO] -------------------------------------------------------
> [INFO] Running TestSuite
> [INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.429 s - in TestSuite
> [INFO] 
> [INFO] Results:
> [INFO] 
> [INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0
> [INFO] 
> [INFO] BUILD SUCCESS
> [INFO] Total time:  7.497 s
> [INFO] Finished at: 2020-02-22T17:59:33-08:00

### Commits

### Code Quality

### DIFF
--- a/metadata-store-directory-common/pom.xml
+++ b/metadata-store-directory-common/pom.xml
@@ -44,6 +44,11 @@ under the License.
 
   <dependencies>
     <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>4.5.8</version>
+    </dependency>
+    <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
       <version>2.10.2</version>
--- a/metadata-store-directory-common/src/test/java/org/apache/helix/msdcommon/mock/TestMockMetadataStoreDirectoryServer.java
+++ b/metadata-store-directory-common/src/test/java/org/apache/helix/msdcommon/mock/TestMockMetadataStoreDirectoryServer.java
@@ -59,74 +59,57 @@ public class TestMockMetadataStoreDirectoryServer {
 
     // Send a GET request for all routing data
     HttpGet getRequest = new HttpGet(
-        endpoint + MockMetadataStoreDirectoryServer.REST_PREFIX + namespace + "/"
-            + MetadataStoreRoutingConstants.ROUTING_DATA);
-    try {
-      CloseableHttpResponse getResponse = httpClient.execute(getRequest);
-      Map<String, Object> resultMap = MockMetadataStoreDirectoryServer.OBJECT_MAPPER
-          .readValue(getResponse.getEntity().getContent(), Map.class);
-      List<Map<String, Object>> routingDataList =
-          (List<Map<String, Object>>) resultMap.get(MetadataStoreRoutingConstants.ROUTING_DATA);
-      Collection<String> allRealms = routingDataList.stream().map(mapEntry -> (String) mapEntry
-          .get(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM))
-          .collect(Collectors.toSet());
-      Assert.assertEquals(allRealms, routingData.keySet());
-      Map<String, List<String>> retrievedRoutingData = routingDataList.stream().collect(Collectors
-          .toMap(mapEntry -> (String) mapEntry
-                  .get(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM),
-              mapEntry -> (List<String>) mapEntry
-                  .get(MetadataStoreRoutingConstants.SHARDING_KEYS)));
-      Assert.assertEquals(retrievedRoutingData, routingData);
-    } catch (IOException e) {
-      e.printStackTrace();
-    }
+        endpoint + MockMetadataStoreDirectoryServer.REST_PREFIX + namespace
+            + MetadataStoreRoutingConstants.MSDS_GET_ALL_ROUTING_DATA_ENDPOINT);
+
+    CloseableHttpResponse getResponse = httpClient.execute(getRequest);
+    Map<String, Object> resultMap = MockMetadataStoreDirectoryServer.OBJECT_MAPPER
+        .readValue(getResponse.getEntity().getContent(), Map.class);
+    List<Map<String, Object>> routingDataList =
+        (List<Map<String, Object>>) resultMap.get(MetadataStoreRoutingConstants.ROUTING_DATA);
+    Collection<String> allRealms = routingDataList.stream().map(mapEntry -> (String) mapEntry
+        .get(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM))
+        .collect(Collectors.toSet());
+    Assert.assertEquals(allRealms, routingData.keySet());
+    Map<String, List<String>> retrievedRoutingData = routingDataList.stream().collect(Collectors
+        .toMap(mapEntry -> (String) mapEntry
+                .get(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM),
+            mapEntry -> (List<String>) mapEntry.get(MetadataStoreRoutingConstants.SHARDING_KEYS)));
+    Assert.assertEquals(retrievedRoutingData, routingData);
 
     // Send a GET request for all realms
     getRequest = new HttpGet(endpoint + MockMetadataStoreDirectoryServer.REST_PREFIX + namespace
         + MockMetadataStoreDirectoryServer.ZK_REALM_ENDPOINT);
-    try {
-      CloseableHttpResponse getResponse = httpClient.execute(getRequest);
-      Map<String, Collection<String>> allRealmsMap = MockMetadataStoreDirectoryServer.OBJECT_MAPPER
-          .readValue(getResponse.getEntity().getContent(), Map.class);
-      Assert.assertTrue(
-          allRealmsMap.containsKey(MetadataStoreRoutingConstants.METADATA_STORE_REALMS));
-      Collection<String> allRealms =
-          allRealmsMap.get(MetadataStoreRoutingConstants.METADATA_STORE_REALMS);
-      Assert.assertEquals(allRealms, routingData.keySet());
-    } catch (IOException e) {
-      e.printStackTrace();
-    }
+    getResponse = httpClient.execute(getRequest);
+    Map<String, Collection<String>> allRealmsMap = MockMetadataStoreDirectoryServer.OBJECT_MAPPER
+        .readValue(getResponse.getEntity().getContent(), Map.class);
+    Assert
+        .assertTrue(allRealmsMap.containsKey(MetadataStoreRoutingConstants.METADATA_STORE_REALMS));
+    allRealms = allRealmsMap.get(MetadataStoreRoutingConstants.METADATA_STORE_REALMS);
+    Assert.assertEquals(allRealms, routingData.keySet());
 
     // Send a GET request for testZkRealm
     String testZkRealm = "zk-0";
     getRequest = new HttpGet(endpoint + MockMetadataStoreDirectoryServer.REST_PREFIX + namespace
         + MockMetadataStoreDirectoryServer.ZK_REALM_ENDPOINT + "/" + testZkRealm);
-    try {
-      CloseableHttpResponse getResponse = httpClient.execute(getRequest);
-      Map<String, Object> shardingKeysMap = MockMetadataStoreDirectoryServer.OBJECT_MAPPER
-          .readValue(getResponse.getEntity().getContent(), Map.class);
-      Assert.assertTrue(
-          shardingKeysMap.containsKey(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM));
-      Assert.assertTrue(shardingKeysMap.containsKey(MetadataStoreRoutingConstants.SHARDING_KEYS));
-      String zkRealm =
-          (String) shardingKeysMap.get(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM);
-      Collection<String> shardingKeyList =
-          (Collection) shardingKeysMap.get(MetadataStoreRoutingConstants.SHARDING_KEYS);
-      Assert.assertEquals(zkRealm, testZkRealm);
-      Assert.assertEquals(shardingKeyList, routingData.get(testZkRealm));
-    } catch (IOException e) {
-      e.printStackTrace();
-    }
+    getResponse = httpClient.execute(getRequest);
+    Map<String, Object> shardingKeysMap = MockMetadataStoreDirectoryServer.OBJECT_MAPPER
+        .readValue(getResponse.getEntity().getContent(), Map.class);
+    Assert.assertTrue(
+        shardingKeysMap.containsKey(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM));
+    Assert.assertTrue(shardingKeysMap.containsKey(MetadataStoreRoutingConstants.SHARDING_KEYS));
+    String zkRealm =
+        (String) shardingKeysMap.get(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM);
+    Collection<String> shardingKeyList =
+        (Collection) shardingKeysMap.get(MetadataStoreRoutingConstants.SHARDING_KEYS);
+    Assert.assertEquals(zkRealm, testZkRealm);
+    Assert.assertEquals(shardingKeyList, routingData.get(testZkRealm));
 
     // Try sending a POST request (not supported)
     HttpPost postRequest = new HttpPost(
         endpoint + MockMetadataStoreDirectoryServer.REST_PREFIX + namespace
             + MockMetadataStoreDirectoryServer.ZK_REALM_ENDPOINT + "/" + testZkRealm);
-    try {
-      CloseableHttpResponse postResponse = httpClient.execute(postRequest);
-    } catch (IOException e) {
-      e.printStackTrace();
-    }
+    CloseableHttpResponse postResponse = httpClient.execute(postRequest);
 
     // Shutdown
     server.stopServer();
--- a/metadata-store-directory-common/src/test/java/org/apache/helix/msdcommon/mock/TestMockMetadataStoreDirectoryServer.java
+++ b/metadata-store-directory-common/src/test/java/org/apache/helix/msdcommon/mock/TestMockMetadataStoreDirectoryServer.java
@@ -1,4 +1,4 @@
-package org.apache.helix.rest.metadatastore.mock;
+package org.apache.helix.msdcommon.mock;
 
 /*
  * Licensed to the Apache Software Foundation (ASF) under one
@@ -20,11 +20,12 @@ package org.apache.helix.rest.metadatastore.mock;
  */
 
 import java.io.IOException;
+import java.util.Collection;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
 import com.google.common.collect.ImmutableList;
+import org.apache.helix.msdcommon.constant.MetadataStoreRoutingConstants;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
@@ -39,7 +40,7 @@ public class TestMockMetadataStoreDirectoryServer {
   public void testMockMetadataStoreDirectoryServer()
       throws IOException {
     // Create fake routing data
-    Map<String, List<String>> routingData = new HashMap<>();
+    Map<String, Collection<String>> routingData = new HashMap<>();
     routingData.put("zk-0", ImmutableList.of("sharding-key-0", "sharding-key-1", "sharding-key-2"));
     routingData.put("zk-1", ImmutableList.of("sharding-key-3", "sharding-key-4", "sharding-key-5"));
     routingData.put("zk-2", ImmutableList.of("sharding-key-6", "sharding-key-7", "sharding-key-8"));
@@ -54,15 +55,39 @@ public class TestMockMetadataStoreDirectoryServer {
     server.startServer();
     CloseableHttpClient httpClient = HttpClients.createDefault();
 
-    // Send a GET request
-    String testZkRealm = "zk-0";
+    // Send a GET request for all realms
     HttpGet getRequest = new HttpGet(
         endpoint + MockMetadataStoreDirectoryServer.REST_PREFIX + namespace
-            + MockMetadataStoreDirectoryServer.ZK_REALM_ENDPOINT + testZkRealm);
+            + MockMetadataStoreDirectoryServer.ZK_REALM_ENDPOINT);
     try {
       CloseableHttpResponse getResponse = httpClient.execute(getRequest);
-      List<String> shardingKeyList = MockMetadataStoreDirectoryServer.OBJECT_MAPPER
-          .readValue(getResponse.getEntity().getContent(), List.class);
+      Map<String, Collection<String>> allRealmsMap = MockMetadataStoreDirectoryServer.OBJECT_MAPPER
+          .readValue(getResponse.getEntity().getContent(), Map.class);
+      Assert.assertTrue(
+          allRealmsMap.containsKey(MetadataStoreRoutingConstants.METADATA_STORE_REALMS));
+      Collection<String> allRealms =
+          allRealmsMap.get(MetadataStoreRoutingConstants.METADATA_STORE_REALMS);
+      Assert.assertEquals(allRealms, routingData.keySet());
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+
+    // Send a GET request for testZkRealm
+    String testZkRealm = "zk-0";
+    getRequest = new HttpGet(endpoint + MockMetadataStoreDirectoryServer.REST_PREFIX + namespace
+        + MockMetadataStoreDirectoryServer.ZK_REALM_ENDPOINT + "/" + testZkRealm);
+    try {
+      CloseableHttpResponse getResponse = httpClient.execute(getRequest);
+      Map<String, Object> shardingKeysMap = MockMetadataStoreDirectoryServer.OBJECT_MAPPER
+          .readValue(getResponse.getEntity().getContent(), Map.class);
+      Assert.assertTrue(
+          shardingKeysMap.containsKey(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM));
+      Assert.assertTrue(shardingKeysMap.containsKey(MetadataStoreRoutingConstants.SHARDING_KEYS));
+      String zkRealm =
+          (String) shardingKeysMap.get(MetadataStoreRoutingConstants.SINGLE_METADATA_STORE_REALM);
+      Collection<String> shardingKeyList =
+          (Collection) shardingKeysMap.get(MetadataStoreRoutingConstants.SHARDING_KEYS);
+      Assert.assertEquals(zkRealm, testZkRealm);
       Assert.assertEquals(shardingKeyList, routingData.get(testZkRealm));
     } catch (IOException e) {
       e.printStackTrace();
@@ -71,7 +96,7 @@ public class TestMockMetadataStoreDirectoryServer {
     // Try sending a POST request (not supported)
     HttpPost postRequest = new HttpPost(
         endpoint + MockMetadataStoreDirectoryServer.REST_PREFIX + namespace
-            + MockMetadataStoreDirectoryServer.ZK_REALM_ENDPOINT + testZkRealm);
+            + MockMetadataStoreDirectoryServer.ZK_REALM_ENDPOINT + "/" + testZkRealm);
     try {
       CloseableHttpResponse postResponse = httpClient.execute(postRequest);
     } catch (IOException e) {
@@ -80,5 +105,6 @@ public class TestMockMetadataStoreDirectoryServer {
 
     // Shutdown
     server.stopServer();
+    httpClient.close();
   }
 }
--- a/metadata-store-directory-common/src/test/java/org/apache/helix/msdcommon/mock/TestMockMetadataStoreDirectoryServer.java
+++ b/metadata-store-directory-common/src/test/java/org/apache/helix/msdcommon/mock/TestMockMetadataStoreDirectoryServer.java
@@ -22,7 +22,9 @@ package org.apache.helix.msdcommon.mock;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import com.google.common.collect.ImmutableList;
 import org.apache.helix.msdcommon.constant.MetadataStoreRoutingConstants;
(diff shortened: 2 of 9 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * This class is a Singleton.
+   */
+  private HttpRoutingDataReader() {
+  }
+
+  /**
+   * Fetches routing data from the data source via HTTP.
+   * @return a mapping from "metadata store realm addresses" to lists of "metadata store sharding keys", where the sharding keys in a value list all route to the realm address in the key disallows a meaningful mapping to be returned
```
pkuwm: Nit, wrap long line.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,81 @@
...
+   */
+  public static Map<String, List<String>> getRoutingData(String msdsEndpoint) {
+
+    Map<String, List<String>> rawRoutingData = new HashMap<>();
+
+    HttpGet requestAllRealmNames =
+        new HttpGet(msdsEndpoint); //TODO: construct an endpoint once REST endpoint is finalized
+    try (CloseableHttpClient httpClient = HttpClients.createDefault();
+        CloseableHttpResponse response = httpClient.execute(requestAllRealmNames)) {
```
pkuwm: 1. Since this is a blocking IO, my thought is it may add more time for zkclient init/creation. Would it help if the request is async while creating a new zkclient?
2. Do we need to consider timeout for this request?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
pkuwm: Nit, wrap long line?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.http.client.ClientProtocolException;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.util.EntityUtils;
+import org.codehaus.jackson.map.ObjectMapper;
+
+
+public class HttpRoutingDataReader {
```
dasahcc: Is that possible to have comment module to hold these http related functions in general? I do see these kind of operations here and there in different modules.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
kaisun2000: Does not matter, right? If you don't have the data, how can you proceeds?

kaisun2000: My thinking is that do we have a singleton class that would use this class of HttpRoutingDataReader upon application initialization to read and cache the routing data. So that later zkclient (shared, dedicated) will all use this cached data? or we will just use this class itself to do the cache?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * Fetches routing data from the data source via HTTP.
+   * @return a mapping from "metadata store realm addresses" to lists of "metadata store sharding keys", where the sharding keys in a value list all route to the realm address in the key disallows a meaningful mapping to be returned
+   */
+  public static Map<String, List<String>> getRoutingData(String msdsEndpoint) {
+
+    Map<String, List<String>> rawRoutingData = new HashMap<>();
+
+    HttpGet requestAllRealmNames =
+        new HttpGet(msdsEndpoint); //TODO: construct an endpoint once REST endpoint is finalized
```
kaisun2000: The application config should provide (via system property) the namespace of endpoint (such as espresso or whatever application). This is the input to construct msdsEndpoint. 

This construction logic should be outside this class?  Or we want to put it inside this class?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: Fixed

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,134 @@
...
+   * @param httpClient
+   * @param msdsEndpoint
+   * @return
+   * @throws IOException
+   */
+  private static Map<String, List<String>> getAllRoutingData(CloseableHttpClient httpClient,
+      String msdsEndpoint)
+      throws IOException {
+    HttpGet requestAllData =
+        new HttpGet(msdsEndpoint + "/" + MetadataStoreRoutingConstants.ROUTING_DATA);
```
pkuwm: Endpoint path is incorrect. You may want “routing-data”

narendly: Good catch - updated.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,134 @@
...
+  private HttpRoutingDataReader() {
+  }
+
+  /**
+   * Fetches routing data from the data source via HTTP.
+   * @return a mapping from "metadata store realm addresses" to lists of
+   * "metadata store sharding keys", where the sharding keys in a value list all route to
+   * the realm address in the key disallows a meaningful mapping to be returned
+   */
+  @RetryOnFailure(attempts = 3, delay = 1, unit = TimeUnit.SECONDS)
```
pkuwm: What is the benefit of introducing new lib: jcabi. I think HttpClients' already provides retry mechanism: `HttpRequestInterceptor` or `setRetryHandler(new DefaultHttpRequestRetryHandler())`

kaisun2000: What is this RetryOnFailture notation? My understanding of this part is kind of weak

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,134 @@
...
+   * @param msdsEndpoint
+   * @return
+   * @throws IOException
+   */
+  private static Map<String, List<String>> getAllRoutingData(CloseableHttpClient httpClient,
+      String msdsEndpoint)
+      throws IOException {
+    HttpGet requestAllData = new HttpGet(
+        msdsEndpoint + MetadataStoreRoutingConstants.MSDS_GET_ALL_ROUTING_DATA_ENDPOINT);
+    CloseableHttpResponse response = httpClient.execute(requestAllData);
```
kaisun2000: will httpCient.execute throw other exceptions aside from IOException?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /**
+   * Fetches routing data from the data source via HTTP.
+   * @return a mapping from "metadata store realm addresses" to lists of
+   * "metadata store sharding keys", where the sharding keys in a value list all route to
+   * the realm address in the key disallows a meaningful mapping to be returned
+   */
+  @RetryOnFailure(attempts = 3, delay = 1, unit = TimeUnit.SECONDS)
+  public static Map<String, List<String>> getRawRoutingData()
```
kaisun2000: Why this is a public API? Do we intend to provide this one outside users? Is there such a case?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,134 @@
...
+      throws IOException {
+    if (_rawRoutingData == null) {
+      synchronized (HttpRoutingDataReader.class) {
+        if (_rawRoutingData == null) {
+          String msdsEndpoint =
+              System.getProperty(MetadataStoreRoutingConstants.MSDS_SERVER_ENDPOINT_KEY);
```
pkuwm: I am considering setting this in constructor. At least we can move it out of synchronized block.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,134 @@
...
+   * @return
+   * @throws IOException
+   * @throws InvalidRoutingDataException
+   */
+  public static MetadataStoreRoutingData getMetadataStoreRoutingData()
+      throws IOException, InvalidRoutingDataException {
+    if (_metadataStoreRoutingData == null) {
+      synchronized (HttpRoutingDataReader.class) {
+        if (_metadataStoreRoutingData == null) {
+          _metadataStoreRoutingData = new TrieRoutingData(getRawRoutingData());
```
kaisun2000: Here, getRawRoutingData would throw illegalStateException which is unchecked. Think you are a user, how do you plan to handle this one? Also, IOException, and InvalidRoutingDataException?

Or do we really want user to handle these exceptions?

It seems to me the user either needs to understand that 1/ configuration issue 2/ MSDS for some reason not working.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,134 @@
...
+            throw new IllegalStateException(
+                "HttpRoutingDataReader was unable to find a valid MSDS endpoint String in System Properties!");
+          }
+
+          // Note: HttpClient's timeout settings are system timeout settings by default
+          CloseableHttpClient httpClient = HttpClients.createDefault();
+          // Update the reference
+          _rawRoutingData = getAllRoutingData(httpClient, msdsEndpoint);
+          // Close any open resources
+          httpClient.close();
```
kaisun2000: This close can be leaked. use try, catch?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,134 @@
...
+        if (_rawRoutingData == null) {
+          String msdsEndpoint =
+              System.getProperty(MetadataStoreRoutingConstants.MSDS_SERVER_ENDPOINT_KEY);
+          if (msdsEndpoint == null || msdsEndpoint.isEmpty()) {
+            throw new IllegalStateException(
+                "HttpRoutingDataReader was unable to find a valid MSDS endpoint String in System Properties!");
+          }
+
+          // Note: HttpClient's timeout settings are system timeout settings by default
+          CloseableHttpClient httpClient = HttpClients.createDefault();
```
pkuwm: - `@RetryOnFailure` is applied to this whole method. It seems we are going to call this reader each time we create a RealmAwareZkClient? The heavy work is one the first time we call to get routing data. 
I would use HttpClients' retry so we don't have to acquire lock and create HttpClient 3 times (a bit heavier) if attempting. Instead, http client could just retry on exception within the locking block, which I think is easier. 
What do you think?
- When is reader supposed to be created, each time when creating zkClient, or a singleton is maintained ?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.util.EntityUtils;
+import org.codehaus.jackson.map.ObjectMapper;
+
+
+public class HttpRoutingDataReader {
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static volatile Map<String, List<String>> _rawRoutingData;
```
kaisun2000: Suggest you add a comment as why use volatile here. https://en.wikipedia.org/wiki/Double-checked_locking. Otherwise, later people changing the code may remove them.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /**
+   * Makes an HTTP call to fetch all routing data.
+   * @param httpClient
+   * @param msdsEndpoint
+   * @return
+   * @throws IOException
+   */
+  private static Map<String, List<String>> getAllRoutingData(CloseableHttpClient httpClient,
```
pkuwm: Maybe I would decouple fetch http json response and parse json to map:
- response = httpClient.execute(routing-data-endpoint) // Easier to just retry and timeout http request.
- Map<String, List<>> parseRoutingData(response) {}

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: Sure.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
kaisun2000: The other more important thing is that I assume that the user needs to configure multiple host to msdsEndPoint? Similar like DNS return IP address set?
Otherwise, this is a single point of failure. 
How doe it work currently in rest client? We can leverage the same mechanism to avoid SPOF.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: That's correct. I believe we still want to float these exceptions up as you said.

IOException means MSDS isn't working for some reason.
InvalidRoutingDataException means their routing data is not configured correctly.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: I agree. Let me split the logic into two.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: I don't think so.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: The benefit of using jcabi is that it allows us to retry in a more succinct way. Defining a retry handler every time would incur repeated code.

I want to use https://http.jcabi.com/ (Fluent HttpClient) over Apache's at some point once we refactor all HttpClients out into a new module because it provides useful annotations for caching and logging, but this is more like an ad-hoc retry here. Let me remove this library import. I'll just use the retry handler. 

Thanks for the suggestion!

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: Not yet. I could make it the access scope private for the time being. Good call.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: I agree - I want to pull this out of the block and fail early if it doesn't exist.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: Let me refactor the code here so that httpClient gets closed properly.

HttpRoutingDataReader is a singleton and the data is read lazily when it's called the first time, hence the double checked locking.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: Good point. Let me do some refactoring here.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: It's a VIP so we have different hosts that are mapped to the single endpoint. But this is an implementation detail for Helix applications/enterprises, so let us now discuss those details here.

kaisun2000: Either I did not get what you mean by VIP or you did not get what I mean here. What really concerns me is that some rest servers may not be available for whatever reason, say network partition. Then we should retry for another rest server. Otherwise, mission critical application may not be able to start. 

So the retry logic is very critical. We don't have this retry logic. You can think of this retry logic similar to Zookeeper native client retry different quorum node.

narendly: Discussed offline. VIP is just an example - different network vendors will provide different abstraction layers.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,141 @@
...
+   * @return
+   * @throws IOException
+   */
+  private static CloseableHttpResponse getAllRoutingData()
+      throws IOException {
+    // Retry count is 3 by default
+    HttpGet requestAllData = new HttpGet(
+        MSDS_ENDPOINT + MetadataStoreRoutingConstants.MSDS_GET_ALL_ROUTING_DATA_ENDPOINT);
+    try (CloseableHttpClient httpClient = HttpClients.custom()
+        .setConnectionBackoffStrategy(new DefaultBackoffStrategy())
```
pkuwm: I don't see a timeout setting. If we don't set timeout, it may use system defined timeout or infinite timeout. To make the system responsive and stable, are we considering timeout: connectTimeout, socketTimeout, etc.?

metadata-store-directory-common/pom.xml
```diff
@@ -43,6 +43,11 @@ under the License.
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>4.5.8</version>
```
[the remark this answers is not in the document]
narendly: Updated to 4.5.8 all throughout

zookeeper-api/pom.xml
```diff
@@ -59,6 +59,11 @@ under the License.
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>4.5.11</version>
```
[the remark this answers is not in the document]
narendly: Updated to 4.5.8

metadata-store-directory-common/src/test/java/org/apache/helix/msdcommon/mock/TestMockMetadataStoreDirectoryServer.java
```diff
@@ -0,0 +1,118 @@
...
+
+    // Try sending a POST request (not supported)
+    HttpPost postRequest = new HttpPost(
+        endpoint + MockMetadataStoreDirectoryServer.REST_PREFIX + namespace
+            + MockMetadataStoreDirectoryServer.ZK_REALM_ENDPOINT + "/" + testZkRealm);
+    CloseableHttpResponse postResponse = httpClient.execute(postRequest);
+
+    // Shutdown
+    server.stopServer();
+    httpClient.close();
```
pkuwm: Minor as it is a test: maybe we would use try...finally to close the server and client, in case exception or assertion fails and server is still running?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,141 @@
...
+   * Returns the raw routing data in a Map< ZkRealm, List of shardingKeys > format.
+   * @param routingDataResponse
+   * @return
+   */
+  private static Map<String, List<String>> parseRoutingData(
+      CloseableHttpResponse routingDataResponse)
+      throws IOException {
+    HttpEntity entity = routingDataResponse.getEntity();
+    if (entity != null) {
+      String resultStr = EntityUtils.toString(entity);
```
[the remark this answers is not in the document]
narendly: Either way seems to work. But setting it as UTF-8 just in case.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: Set.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/util/TestHttpRoutingDataReader.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  @AfterClass
+  public void afterClass() {
+    _msdsServer.stopServer();
+  }
+
+  @Test
+  public void testGetRawRoutingData()
+      throws IOException {
```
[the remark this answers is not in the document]
narendly: Updated

metadata-store-directory-common/src/test/java/org/apache/helix/msdcommon/mock/TestMockMetadataStoreDirectoryServer.java
narendly: Sure. updated.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/util/TestHttpRoutingDataReader.java
```diff
@@ -0,0 +1,124 @@
...
+  @AfterClass
+  public void afterClass() {
+    _msdsServer.stopServer();
+  }
+
+  @Test
+  public void testGetRawRoutingData() throws IOException {
+    Map<String, List<String>> rawRoutingData = HttpRoutingDataReader.getRawRoutingData();
+    _testRawRoutingData
+        .forEach((realm, keys) -> Assert.assertEquals(rawRoutingData.get(realm), keys));
```
pkuwm: Can Assert correctly check a list and a collection (a set instance) ? I doubt it.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/util/TestHttpRoutingDataReader.java
```diff
@@ -0,0 +1,124 @@
...
+
+  @Test(dependsOnMethods = "testGetRawRoutingData")
+  public void testGetMetadataStoreRoutingData() throws IOException, InvalidRoutingDataException {
+    MetadataStoreRoutingData data = HttpRoutingDataReader.getMetadataStoreRoutingData();
+    Map<String, String> allMappings = data.getAllMappingUnderPath("/");
+    Map<String, Set<String>> groupedMappings = allMappings.entrySet().stream().collect(Collectors
+        .groupingBy(Map.Entry::getValue,
+            Collectors.mapping(Map.Entry::getKey, Collectors.toSet())));
+    _testRawRoutingData.forEach((realm, keys) -> {
+      // Two way containsAll because AssertEquals on two set collections is buggy in that
```
pkuwm: What do you mean here? My understanding is, if neither set is null, Assert could check them, regardless ordering of elements: set["a", "b", "c"], set["c", "b", "a"] should be equal. What is the case they are not equal: "**if the ordering of elements is not equal**"?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/util/TestHttpRoutingDataReader.java
```diff
@@ -0,0 +1,124 @@
...
+    // Kill MSDS and restart with a new mapping
+    _msdsServer.stopServer();
+    _msdsServer =
+        new MockMetadataStoreDirectoryServer(_host, _port, _namespace, _testRawRoutingData);
+    _msdsServer.startServer();
+
+    // HttpRoutingDataReader should still return old data because it's static
+    // Make sure the results don't contain the new realm
+    Map<String, List<String>> rawRoutingData = HttpRoutingDataReader.getRawRoutingData();
+    Assert.assertFalse(rawRoutingData.containsKey(newRealm));
```
pkuwm: If you say "should still return old data because it's static", I think it is important to check we could still check the old data is actually returned. assert `rawRoutingData` equals to `_rawRoutingData`?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,146 @@
...
+    return _metadataStoreRoutingData;
+  }
+
+  /**
+   * Makes an HTTP call to fetch all routing data.
+   * @return
+   * @throws IOException
+   */
+  private static CloseableHttpResponse getAllRoutingData() throws IOException {
+    // Retry count is 3 by default
```
kaisun2000: Where is retry count set? Also, shall we make the retry count configurable? I think configurable can be good idea. 3 still seems to me not very safe. If this part not working well, the whole cluster may not bootup.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,146 @@
...
+    if (MSDS_ENDPOINT == null || MSDS_ENDPOINT.isEmpty()) {
+      throw new IllegalStateException(
+          "HttpRoutingDataReader was unable to find a valid MSDS endpoint String in System Properties!");
+    }
+    if (_rawRoutingData == null) {
+      synchronized (HttpRoutingDataReader.class) {
+        if (_rawRoutingData == null) {
+          CloseableHttpResponse routingDataResponse = getAllRoutingData();
+          // Update the reference if reading routingData over HTTP is successful
+          _rawRoutingData = parseRoutingData(routingDataResponse);
```
kaisun2000: _parseRoutingData can return an empty map. Then the empty map can be return to user.  This is basically telling user we find nothing. This is error condition. 

If we make this method public, we should let user know this error condition by throwing exception.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,146 @@
...
+   * @return
+   * @throws IOException if there is an issue connecting to MSDS
+   * @throws InvalidRoutingDataException if the raw routing data is not valid
+   */
+  public static MetadataStoreRoutingData getMetadataStoreRoutingData()
+      throws IOException, InvalidRoutingDataException {
+    if (_metadataStoreRoutingData == null) {
+      synchronized (HttpRoutingDataReader.class) {
+        if (_metadataStoreRoutingData == null) {
+          _metadataStoreRoutingData = new TrieRoutingData(getRawRoutingData());
```
kaisun2000: I did not see the TrieRoutingData implementation. Can we make sure _metaDataStoreRoutingData will not be empty map after this line? 

This is important. Otherwise, with empty map, you are not going to retry getRawRoutingData. The result is disaster then.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+    return _metadataStoreRoutingData;
+  }
+
+  /**
+   * Makes an HTTP call to fetch all routing data.
+   * @return
+   * @throws IOException
+   */
+  private static CloseableHttpResponse getAllRoutingData() throws IOException {
```
kaisun2000: Let us add a comment here. 
We avoid this point to be SPOF by relying on VIP LB and retry mechanism. This dependency is important. We may revisit this dependency assumption later.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/util/TestHttpRoutingDataReader.java
narendly: I'm not quite sure I understand what you're asking. Two-way containsAll is basically an equality check. .

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: @kaisun2000 Please create an issue for this - retry is set as the default HttpClient behavior.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: OK.

metadata-store-directory-common/src/test/java/org/apache/helix/msdcommon/mock/TestMockMetadataStoreDirectoryServer.java
```diff
@@ -0,0 +1,119 @@
...
+    // Start MockMSDS
+    String host = "localhost";
+    int port = 11000;
+    String endpoint = "http://" + host + ":" + port;
+    String namespace = "MY-HELIX-NAMESPACE";
+
+    MockMetadataStoreDirectoryServer server =
+        new MockMetadataStoreDirectoryServer(host, port, namespace, routingData);
+    try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
+      server.startServer();
```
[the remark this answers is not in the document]
narendly: OK

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: If we find nothing, we should return an empty map. Users should handle it. Helix Java APIs are the users, and they will handle this scenario as well.

In other situations (IOException, invalidRoutingDataExceptions), appropriate exceptions are thrown.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: _metaDataStoreRoutingData is not a map. It may have empty content, but if it does have an empty content, the users will fail (basically fail to instantiate).

This RoutingDataReader only reads - and from its perspective, it can read something that's empty. That's valid. Error checking should be done by the user. Separation of responsibility.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: OK. I will add clarifying comment next to MSDS endpoint.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/util/TestHttpRoutingDataReader.java
narendly: Would you feel more comfortable if I wrapped it as a HashMap? done.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/util/TestHttpRoutingDataReader.java
narendly: I could do that check too.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/util/TestHttpRoutingDataReader.java
pkuwm: I understand `Two-way containsAll is basically an equality check`.
My point is, if the 2 sets have the same string elements, assertEqual() should pass, no matter the ordering of elements is equal or not.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
kaisun2000: Better to throw if _metaDataStoreRoutingData upon instantiation successfully still have empty (invalid data). Throw early easy for user to identify the problem.

kaisun2000:
Better to throw if _metaDataStoreRoutingData upon instantiation successfully still have empty (invalid condition). Throw early easy for user to identify the problem. 

This instantiation happens only one time. Later usage would be all over the places, even it is ideally still in Helix Java API.

narendly:
----
HttpRoutingDataReader is a component used by new ZkClient APIs to make an HTTP read request to the metadata store directory service to retrieve routing data. ZkClient APIs will construct an internal MetadataStoreRoutingData instance based on the raw routing data retrieved from MSDS.

Note: this change contains modifications to MockMSDS because the actual endpoint names changed. The methods and http server contexts have been updated.

PR #785, opened 2020-02-19
Fix the ConcurrentModificationException in ClusterEvent.java

### LINKED ISSUE #784: ConcurrentModificationException during mvn test
@narendly reported the stack trace of a problematic exception

```
22723 [ZkClient-EventThread-752-localhost:2183] ERROR org.apache.helix.manager.zk.ZKExceptionHandler - exception in handling data-change. path: /TestHelixAdminCli_grand/INSTANCES/controller_9000/CURRENTSTATES/10014b003960037/TestHelixAdminCli, listener: org.apache.helix.controller.GenericHelixController@64b7b07
java.util.ConcurrentModificationException
at java.util.HashMap$HashIterator.nextNode(HashMap.java:1442)
at java.util.HashMap$KeyIterator.next(HashMap.java:1466)
at org.apache.helix.controller.stages.ClusterEvent.clone(ClusterEvent.java:119)
at org.apache.helix.controller.GenericHelixController.pushToEventQueues(GenericHelixController.java:1010)
at org.apache.helix.controller.GenericHelixController.onStateChange(GenericHelixController.java:826)
at org.apache.helix.manager.zk.CallbackHandler.invoke(CallbackHandler.java:418)
at org.apache.helix.manager.zk.CallbackHandler.enqueueTask(CallbackHandler.java:341)
at org.apache.helix.manager.zk.CallbackHandler.handleDataChange(CallbackHandler.java:625)
at org.apache.helix.zookeeper.zkclient.ZkClient$6.run(ZkClient.java:1305)
at org.apache.helix.zookeeper.zkclient.ZkEventThread.run(ZkEventThread.java:90)
true: wait 426ms, ClusterStateVerifier$BestPossAndExtViewZkVerifier(TestHelixAdminCli@localhost:2183)
true: wait 250ms, ClusterStateVerifier$BestPossAndExtViewZkVerifier(TestHelixAdminCli@localhost:2183)
END testDropAddResource at Thu Feb 13 14:50:12 PST 2020, took: 3075ms.
START testInstanceOperations at Thu Feb 13 14:50:12 PST 2020

```

The root cause is: In ClusterEvent.java, the hashmap is updated in another thread while `clone` is still in the for-loop
```
for (String attributeName : _eventAttributeMap.keySet()) {
  newEvent.addAttribute(attributeName, _eventAttributeMap.get(attributeName));
}


public void addAttribute(String attrName, Object attrValue) {
  if (logger.isTraceEnabled()) {
    logger.trace("Adding attribute:" + attrName);
    logger.trace(" attribute value:" + attrValue);
  }

  _eventAttributeMap.put(attrName, attrValue);
}
```

i3wangyi:
### Issues

(Fixes #784 )
### Description

The `clone` method is not quite clean in its implementation, readers can hardly tell which `_eventAttributeMap` is the old event versus the new one.

Also, the stack trace in #784 proves the possibility of ConcurrentModificationException. A right and straightforward way is to make the current attribute map concurrent and thread-safe

### Tests

- TestClusterEvent#testClone

END testSessionExpiredWhenResetHandlers at Wed Feb 19 19:03:44 PST 2020, took: 7722ms.
Tests run: 1083, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 4,470.368 sec <<< FAILURE! - in TestSuite
testMissingTopStateDurationMonitoring(org.apache.helix.integration.controller.TestControllerLeadershipChange)  Time elapsed: 4.353 sec  <<< FAILURE!
	at org.testng.Assert.fail(Assert.java:89)
	at org.testng.Assert.failNotEquals(Assert.java:480)
	at org.testng.Assert.assertTrue(Assert.java:37)
	at org.testng.Assert.assertTrue(Assert.java:47)
	at org.apache.helix.integration.controller.TestControllerLeadershipChange.testMissingTopStateDurationMonitoring(TestControllerLeadershipChange.java:262)

Results :

Failed tests: 
  TestControllerLeadershipChange.testMissingTopStateDurationMonitoring:262 expected:<true> but was:<false>

Tests run: 1083, Failures: 1, Errors: 0, Skipped: 0

The failed test passed in IDE

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
@@ -19,9 +19,9 @@ package org.apache.helix.controller.stages;
  * under the License.
  */
 
-import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -37,7 +37,7 @@ public class ClusterEvent {
   @Deprecated
   public ClusterEvent(ClusterEventType eventType) {
     _eventType = eventType;
-    _eventAttributeMap = new HashMap<>();
+    _eventAttributeMap = new ConcurrentHashMap<>();
     _creationTime = System.currentTimeMillis();
     _eventId = UUID.randomUUID().toString();
   }
@@ -50,21 +50,8 @@ public class ClusterEvent {
     _clusterName = clusterName;
     _eventType = eventType;
 
-    _eventAttributeMap = new HashMap<>();
+    _eventAttributeMap = new ConcurrentHashMap<>();
     _creationTime = System.currentTimeMillis();
-    _eventId = eventId;
-  }
-
-  /**
-   * A private copy constructor that allows the override of {@link #_eventId}
-   * @param clusterEvent The other cluster event object
-   * @param eventId The event id to be overridden
-   */
-  private ClusterEvent(ClusterEvent clusterEvent, String eventId) {
-    _clusterName = clusterEvent._clusterName;
-    _eventType = clusterEvent._eventType;
-    _eventAttributeMap = new HashMap<>(clusterEvent._eventAttributeMap);
-    _creationTime = clusterEvent._creationTime;
     _eventId = eventId;
   }
 
@@ -127,6 +114,11 @@ public class ClusterEvent {
   }
 
   public ClusterEvent clone(String eventId) {
-    return new ClusterEvent(this, eventId);
+    ClusterEvent newEvent = new ClusterEvent(_clusterName, _eventType, eventId);
+    newEvent.setCreationTime(_creationTime);
+    for (String attributeName : _eventAttributeMap.keySet()) {
+      newEvent.addAttribute(attributeName, _eventAttributeMap.get(attributeName));
+    }
+    return newEvent;
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
@@ -58,7 +58,7 @@ public class ClusterEvent {
   /**
    * A private copy constructor that allows the override of {@link #_eventId}
    * @param clusterEvent The other cluster event object
-   * @param eventId The event Id to be overriden
+   * @param eventId The event id to be overridden
    */
   private ClusterEvent(ClusterEvent clusterEvent, String eventId) {
     _clusterName = clusterEvent._clusterName;
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
@@ -55,6 +55,19 @@ public class ClusterEvent {
     _eventId = eventId;
   }
 
+  /**
+   * A private copy constructor that allows the override of {@link #_eventId}
+   * @param clusterEvent The other cluster event object
+   * @param eventId The event Id to be overriden
+   */
+  private ClusterEvent(ClusterEvent clusterEvent, String eventId) {
+    _clusterName = clusterEvent._clusterName;
+    _eventType = clusterEvent._eventType;
+    _eventAttributeMap = new HashMap<>(clusterEvent._eventAttributeMap);
+    _creationTime = clusterEvent._creationTime;
+    _eventId = eventId;
+  }
+
   public void addAttribute(String attrName, Object attrValue) {
     if (logger.isTraceEnabled()) {
       logger.trace("Adding attribute:" + attrName);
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
```diff
@@ -55,6 +55,19 @@ public ClusterEvent(String clusterName, ClusterEventType eventType, String event
     _eventId = eventId;
   }
 
+  /**
```
narendly: Thanks for the PR!

1. Can we just replace clone() logic with what you have in the copy constructor? If the constructor is private, then I think it'd just be simpler to put this "copy" logic in clone() (it seems pretty simple anyways).

2. With that said, do we still need addAttribute() anymore?

pkuwm:
@i3wangyi I am a bit confused and not convinced by the root cause: "In ClusterEvent.java, the hashmap update itself while in its for loop".

The maps `_eventAttributeMap` are different objects. So addAttribute() does not add the attributes to the  original event's map, but the new event's different map.

So I think the root cause could be another place (eg. another thread) updating the original event's map while in `for` loop.

helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
```diff
@@ -114,11 +127,6 @@ public String toString() {
   }
 
   public ClusterEvent clone(String eventId) {
-    ClusterEvent newEvent = new ClusterEvent(_clusterName, _eventType, eventId);
-    newEvent.setCreationTime(_creationTime);
-    for (String attributeName : _eventAttributeMap.keySet()) {
```
pkuwm: @i3wangyi I am a bit confused and not convinced by the root cause: "In ClusterEvent.java, the hashmap update itself while in its for loop".

The maps _eventAttributeMap are different objects. So addAttribute() does not add the attributes to the original event's map, but the new event's different map.

So I think the root cause could be another place (eg. another thread) updating the original event's map while in for loop.

helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
i3wangyi: That's really what I thought initially. But it turns out the variable `_eventAttributeMap` is `final`, which means I cannot mutate the value. 
Are you also referring to this
```
ClusterEvent newEvent = new ClusterEvent(_clusterName, _eventType, eventId);	    return new ClusterEvent(this, eventId);
newEvent.setCreationTime(_creationTime);
newEvent._eventAttributeMap = new HashMap(__eventAttributeMap);
```

helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
i3wangyi: You're right! I'll update the root cause in #785. I was initially unaware of the different objects of `_eventAttributeMap`, cause the old code is readable at first sight and it's not thread-safe. 
This fix is still valid under multi-threads. Added one test but making it multi-threading and produce the same issue is non-trivial.

pkuwm: Can you explain why the copy constructor is different from the original for-loop? Maybe I miss something, but I see that `new HashMap<>(map)` is basically also a for-loop to copy the original map entries, which means, the entries could still change and throws concurrent mod exception for the step `new HashMap<>(map)`.
Please correct me if I am wrong. Appreciate it!

i3wangyi: @pkuwm Thanks! I was under the impression that copy over a map is a thread-safe operation in some languages(python, e.g). I just tried it in Java, it still suffers the same exception. So I believe the right fix & less code change is to make it ConcurrentHashMap. Also added a test for it.

helix-core/src/test/java/org/apache/helix/controller/stages/TestClusterEvent.java
```diff
@@ -19,16 +19,51 @@
...
+    String clusterName = "TestCluster";
+    ClusterEvent event = new ClusterEvent(clusterName, ClusterEventType.Unknown, "testId");
+    for (int i = 0; i < 100; i++) {
+      event.addAttribute(String.valueOf(i), i);
+    }
+    List<Thread> threads = new ArrayList<>();
+    for (int i = 0; i < 10; i++) {
+      threads.add(new Thread(() -> {
+        String threadName = Thread.currentThread().getName();
+        event.addAttribute(threadName, threadName);
```
pkuwm: This is a reasonable test. Maybe to make it more accurate in this test, we can use a CountDownLatch as a signal right before clone().
```
latch.countDown(); // Tell 10 threads to modify this the attributeMap.
clone();
```
And also add more attributes (say, 20) in each thread just to make the time a bit longer.
This CountDownLatch and more attributes make sure that the attribute adding is not completed yet before clone().

helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEvent.java
pkuwm: Agreed that ConcurrentHashMap could protect it.

helix-core/src/test/java/org/apache/helix/controller/stages/TestClusterEvent.java
i3wangyi: Actually, one additional thread is good enough if it's adding, say, 100 items into the map. It reduces the cost of thread consumptions by test class.

i3wangyi:
The PR is ready to be merged into Master, approved by @dasahcc

PR #788, opened 2020-02-20
Implement request forwarding for ZkRoutingDataWriter

### LINKED ISSUE #778: Add leader request forwarding for ZkRoutingDataWriter
Due to the potential concurrency issues raised by multiple writes to ZooKeeper, we have decided on the design of having one single writer for `ZkRoutingDataWriter`. More specifically, by utilizing `ZkDistributedLeaderElection`, one instance of the writer will be elected as the leader, and all other instances will forward write requests to the leader. We need to add logic in `ZkRoutingDataWriter` to support that.

NealSun96:
### Issues

Fixes #778 , #794 , #795 , #800 , #801

### Description

Because of the single-writer design for `ZkRoutingDataWriter`, we are utilizing `ZkDistributedLeaderElection` to choose a leader instance and non-leaders will forward write requests to the leader. This PR:
1. implements the logic for request forwarding; 
2. changes the current MSDS endpoints behaviors such that the endpoints now respect the return values of underlying methods (previously, if the underlying methods return `false`, the endpoints will still return `OK` which is not desirable); 
3. enables the previously disabled MSDS endpoints tests;
4. fixes `TestMetadataStoreDirectoryAccessor` to allow tests to be run in leader mode (the previous version disallows tests to be run in leader mode);
5. fixes 2 bugs that break endpoionts tests (#794, #795);
6. adds unit tests for follower situations in `TestZkRoutingDataWriter`, which uncovers a bug on leader election that causes null pointer for followers (#801, fixed);
7. makes ZkMetadataStoreDirectory singleton (#800);
8. added integration tests for the entire flow.

### Tests
[INFO] Tests run: 131, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 24.278 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 131, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  52.393 s
[INFO] Finished at: 2020-02-25T18:48:33-08:00

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -74,15 +74,7 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
 
   public static ZkMetadataStoreDirectory getInstance(String namespace, String zkAddress)
       throws InvalidRoutingDataException {
-    if (_zkMetadataStoreDirectoryInstance == null) {
-      synchronized (ZkMetadataStoreDirectory.class) {
-        if (_zkMetadataStoreDirectoryInstance == null) {
-          _zkMetadataStoreDirectoryInstance = new ZkMetadataStoreDirectory();
-        }
-      }
-    }
-    _zkMetadataStoreDirectoryInstance.init(namespace, zkAddress);
-
+    getInstance().init(namespace, zkAddress);
     return _zkMetadataStoreDirectoryInstance;
   }
 
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -28,6 +28,7 @@ import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.helix.msdcommon.callback.RoutingDataListener;
 import org.apache.helix.msdcommon.datamodel.MetadataStoreRoutingData;
 import org.apache.helix.msdcommon.datamodel.TrieRoutingData;
@@ -41,20 +42,21 @@ import org.slf4j.LoggerFactory;
 
 
 /**
+ * NOTE: This is a singleton class. DO NOT EXTEND!
  * ZK-based MetadataStoreDirectory that listens on the routing data in routing ZKs with a update
  * callback.
  */
 public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, RoutingDataListener {
   private static final Logger LOG = LoggerFactory.getLogger(ZkMetadataStoreDirectory.class);
 
-  // TODO: enable the line below when implementation is complete
   // The following maps' keys represent the namespace
-  private final Map<String, MetadataStoreRoutingDataReader> _routingDataReaderMap;
-  private final Map<String, MetadataStoreRoutingDataWriter> _routingDataWriterMap;
-  private final Map<String, MetadataStoreRoutingData> _routingDataMap;
-  private final Map<String, String> _routingZkAddressMap;
+  // NOTE: made protected for testing reasons. DO NOT MODIFY!
+  protected final Map<String, MetadataStoreRoutingDataReader> _routingDataReaderMap;
+  protected final Map<String, MetadataStoreRoutingDataWriter> _routingDataWriterMap;
+  protected final Map<String, MetadataStoreRoutingData> _routingDataMap;
+  protected final Map<String, String> _routingZkAddressMap;
   // <namespace, <realm, <list of sharding keys>> mappings
-  private final Map<String, Map<String, List<String>>> _realmToShardingKeysMap;
+  protected final Map<String, Map<String, List<String>>> _realmToShardingKeysMap;
 
   private static volatile ZkMetadataStoreDirectory _zkMetadataStoreDirectoryInstance;
 
@@ -84,7 +86,11 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
     return _zkMetadataStoreDirectoryInstance;
   }
 
-  private ZkMetadataStoreDirectory() {
+  /**
+   * Note: this is a singleton class. The constructor is made protected for testing. DO NOT EXTEND!
+   */
+  @VisibleForTesting
+  protected ZkMetadataStoreDirectory() {
     _routingDataReaderMap = new ConcurrentHashMap<>();
     _routingDataWriterMap = new ConcurrentHashMap<>();
     _routingZkAddressMap = new ConcurrentHashMap<>();
@@ -175,7 +181,6 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
 
   @Override
   public boolean addMetadataStoreRealm(String namespace, String realm) {
-    System.out.println("actual method called");
     if (!_routingDataWriterMap.containsKey(namespace)) {
       throw new IllegalArgumentException(
           "Failed to add metadata store realm: Namespace " + namespace + " is not found!");
@@ -261,5 +266,11 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
   public synchronized void close() {
     _routingDataReaderMap.values().forEach(MetadataStoreRoutingDataReader::close);
     _routingDataWriterMap.values().forEach(MetadataStoreRoutingDataWriter::close);
+    _routingDataReaderMap.clear();
+    _routingDataWriterMap.clear();
+    _routingZkAddressMap.clear();
+    _realmToShardingKeysMap.clear();
+    _routingDataMap.clear();
+    _zkMetadataStoreDirectoryInstance = null;
   }
 }
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -68,7 +68,6 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
         }
       }
     }
-
     return _zkMetadataStoreDirectoryInstance;
   }
 
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -21,7 +21,6 @@ package org.apache.helix.rest.metadatastore;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -57,34 +56,57 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
   // <namespace, <realm, <list of sharding keys>> mappings
   private final Map<String, Map<String, List<String>>> _realmToShardingKeysMap;
 
-  /**
-   * Creates a ZkMetadataStoreDirectory based on the given routing ZK addresses.
-   * @param routingZkAddressMap (namespace, routing ZK connect string)
-   * @throws InvalidRoutingDataException
-   */
-  public ZkMetadataStoreDirectory(Map<String, String> routingZkAddressMap)
-      throws InvalidRoutingDataException {
-    if (routingZkAddressMap == null || routingZkAddressMap.isEmpty()) {
-      throw new InvalidRoutingDataException("Routing ZK Addresses given are invalid!");
+  private static volatile ZkMetadataStoreDirectory _zkMetadataStoreDirectoryInstance;
+
+  public static ZkMetadataStoreDirectory getInstance() {
+    if (_zkMetadataStoreDirectoryInstance == null) {
+      synchronized (ZkMetadataStoreDirectory.class) {
+        if (_zkMetadataStoreDirectoryInstance == null) {
+          _zkMetadataStoreDirectoryInstance = new ZkMetadataStoreDirectory();
+        }
+      }
     }
-    _routingDataReaderMap = new HashMap<>();
-    _routingDataWriterMap = new HashMap<>();
-    _routingZkAddressMap = routingZkAddressMap;
+
+    return _zkMetadataStoreDirectoryInstance;
+  }
+
+  public static ZkMetadataStoreDirectory getInstance(String namespace, String zkAddress)
+      throws InvalidRoutingDataException {
+    if (_zkMetadataStoreDirectoryInstance == null) {
+      synchronized (ZkMetadataStoreDirectory.class) {
+        if (_zkMetadataStoreDirectoryInstance == null) {
+          _zkMetadataStoreDirectoryInstance = new ZkMetadataStoreDirectory();
+        }
+      }
+    }
+    _zkMetadataStoreDirectoryInstance.init(namespace, zkAddress);
+
+    return _zkMetadataStoreDirectoryInstance;
+  }
+
+  private ZkMetadataStoreDirectory() {
+    _routingDataReaderMap = new ConcurrentHashMap<>();
+    _routingDataWriterMap = new ConcurrentHashMap<>();
+    _routingZkAddressMap = new ConcurrentHashMap<>();
     _realmToShardingKeysMap = new ConcurrentHashMap<>();
     _routingDataMap = new ConcurrentHashMap<>();
+  }
 
-    // Create RoutingDataReaders and RoutingDataWriters
-    for (Map.Entry<String, String> routingEntry : _routingZkAddressMap.entrySet()) {
-      _routingDataReaderMap.put(routingEntry.getKey(),
-          new ZkRoutingDataReader(routingEntry.getKey(), routingEntry.getValue(), this));
-      _routingDataWriterMap.put(routingEntry.getKey(),
-          new ZkRoutingDataWriter(routingEntry.getKey(), routingEntry.getValue()));
+  private void init(String namespace, String zkAddress) throws InvalidRoutingDataException {
+    if (!_routingZkAddressMap.containsKey(namespace)) {
+      synchronized (_routingZkAddressMap) {
+        if (!_routingZkAddressMap.containsKey(namespace)) {
+          _routingZkAddressMap.put(namespace, zkAddress);
+          _routingDataReaderMap.put(namespace, new ZkRoutingDataReader(namespace, zkAddress, this));
+          _routingDataWriterMap.put(namespace, new ZkRoutingDataWriter(namespace, zkAddress));
 
-      // Populate realmToShardingKeys with ZkRoutingDataReader
-      _realmToShardingKeysMap.put(routingEntry.getKey(),
-          _routingDataReaderMap.get(routingEntry.getKey()).getRoutingData());
-      _routingDataMap.put(routingEntry.getKey(),
-          new TrieRoutingData(_realmToShardingKeysMap.get(routingEntry.getKey())));
+          // Populate realmToShardingKeys with ZkRoutingDataReader
+          _realmToShardingKeysMap
+              .put(namespace, _routingDataReaderMap.get(namespace).getRoutingData());
+          _routingDataMap
+              .put(namespace, new TrieRoutingData(_realmToShardingKeysMap.get(namespace)));
+        }
+      }
     }
   }
 
@@ -153,6 +175,7 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
 
   @Override
   public boolean addMetadataStoreRealm(String namespace, String realm) {
+    System.out.println("actual method called");
     if (!_routingDataWriterMap.containsKey(namespace)) {
       throw new IllegalArgumentException(
           "Failed to add metadata store realm: Namespace " + namespace + " is not found!");
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
@@ -21,6 +21,7 @@ package org.apache.helix.rest.metadatastore.accessor;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
@@ -44,8 +45,6 @@ import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.util.EntityUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import static org.apache.helix.msdcommon.constant.MetadataStoreRoutingConstants.FORWARD_HTTP_REQUEST_TIMEOUT;
 
 
 public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
@@ -89,10 +88,12 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
     _leaderElection = new ZkDistributedLeaderElection(_zkClient,
         MetadataStoreRoutingConstants.LEADER_ELECTION_ZNODE, myServerInfo);
 
-    RequestConfig config =
-        RequestConfig.custom().setConnectTimeout(FORWARD_HTTP_REQUEST_TIMEOUT * 1000)
-            .setConnectionRequestTimeout(FORWARD_HTTP_REQUEST_TIMEOUT * 1000)
-            .setSocketTimeout(FORWARD_HTTP_REQUEST_TIMEOUT * 1000).build();
+    RequestConfig config = RequestConfig.custom()
+        .setConnectTimeout(MetadataStoreRoutingConstants.HTTP_REQUEST_FORWARDING_TIMEOUT * 1000)
+        .setConnectionRequestTimeout(
+            MetadataStoreRoutingConstants.HTTP_REQUEST_FORWARDING_TIMEOUT * 1000)
+        .setSocketTimeout(MetadataStoreRoutingConstants.HTTP_REQUEST_FORWARDING_TIMEOUT * 1000)
+        .build();
     _forwardHttpClient = HttpClientBuilder.create().setDefaultRequestConfig(config).build();
   }
 
@@ -105,8 +106,10 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
       return createZkRealm(realm);
     }
 
-    String url_suffix = "namespaces/" + _namespace + "/metadata-store-realms/" + realm;
-    return forwardRequestToLeader(url_suffix, "put", "addMetadataStoreRealm",
+    String urlSuffix =
+        constructUrlSuffix(MetadataStoreRoutingConstants.MSDS_GET_ALL_REALMS_ENDPOINT, realm);
+    return forwardRequestToLeader(urlSuffix,
+        MetadataStoreRoutingConstants.HttpRequestForwardingVerbs.PUT, "addMetadataStoreRealm",
         Response.Status.CREATED.getStatusCode());
   }
 
@@ -119,8 +122,10 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
       return deleteZkRealm(realm);
     }
 
-    String url_suffix = "namespaces/" + _namespace + "/metadata-store-realms/" + realm;
-    return forwardRequestToLeader(url_suffix, "delete", "deleteMetadataStoreRealm",
+    String urlSuffix =
+        constructUrlSuffix(MetadataStoreRoutingConstants.MSDS_GET_ALL_REALMS_ENDPOINT, realm);
+    return forwardRequestToLeader(urlSuffix,
+        MetadataStoreRoutingConstants.HttpRequestForwardingVerbs.DELETE, "deleteMetadataStoreRealm",
         Response.Status.OK.getStatusCode());
   }
 
@@ -133,10 +138,11 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
       return createZkShardingKey(realm, shardingKey);
     }
 
-    String url_suffix =
-        "namespaces/" + _namespace + "/metadata-store-realms/" + realm + "/sharding-keys/"
-            + shardingKey;
-    return forwardRequestToLeader(url_suffix, "put", "addShardingKey",
+    String urlSuffix =
+        constructUrlSuffix(MetadataStoreRoutingConstants.MSDS_GET_ALL_REALMS_ENDPOINT, realm,
+            MetadataStoreRoutingConstants.MSDS_GET_ALL_SHARDING_KEYS_ENDPOINT, shardingKey);
+    return forwardRequestToLeader(urlSuffix,
+        MetadataStoreRoutingConstants.HttpRequestForwardingVerbs.PUT, "addShardingKey",
         Response.Status.CREATED.getStatusCode());
   }
 
@@ -149,10 +155,11 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
       return deleteZkShardingKey(realm, shardingKey);
     }
 
-    String url_suffix =
-        "namespaces/" + _namespace + "/metadata-store-realms/" + realm + "/sharding-keys/"
-            + shardingKey;
-    return forwardRequestToLeader(url_suffix, "delete", "deleteShardingKey",
+    String urlSuffix =
+        constructUrlSuffix(MetadataStoreRoutingConstants.MSDS_GET_ALL_REALMS_ENDPOINT, realm,
+            MetadataStoreRoutingConstants.MSDS_GET_ALL_SHARDING_KEYS_ENDPOINT, shardingKey);
+    return forwardRequestToLeader(urlSuffix,
+        MetadataStoreRoutingConstants.HttpRequestForwardingVerbs.DELETE, "deleteShardingKey",
         Response.Status.OK.getStatusCode());
   }
 
@@ -306,16 +313,25 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
     return true;
   }
 
-  private boolean forwardRequestToLeader(String url_suffix, String request_method, String endPoint,
+  private String constructUrlSuffix(String... urlParams) {
+    List<String> allUrlParameters = new ArrayList<>(
+        Arrays.asList(MetadataStoreRoutingConstants.MSDS_NAMESPACES_URL_PREFIX, _namespace));
+    allUrlParameters.addAll(Arrays.asList(urlParams));
+    // HttpUriRequest will convert all double slashes to single slashes
+    return String.join("/", allUrlParameters);
+  }
+
+  private boolean forwardRequestToLeader(String urlSuffix,
+      MetadataStoreRoutingConstants.HttpRequestForwardingVerbs request_method, String endPoint,
       int expectedResponseCode) throws IllegalArgumentException {
     String leaderHostName = _leaderElection.getCurrentLeaderInfo().getId();
-    String url = leaderHostName + url_suffix;
+    String url = leaderHostName + urlSuffix;
     HttpUriRequest request;
     switch (request_method) {
-      case "put":
+      case PUT:
         request = new HttpPut(url);
         break;
-      case "delete":
+      case DELETE:
         request = new HttpDelete(url);
         break;
       default:
@@ -333,8 +349,9 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
       if (response.getStatusLine().getStatusCode() != expectedResponseCode) {
         HttpEntity respEntity = response.getEntity();
         String errorLog =
-            "The forwarded request to leader has failed for " + endPoint + ". Current hostname: "
-                + _myHostName + " Leader hostname: " + leaderHostName;
+            "The forwarded request to leader has failed for " + endPoint + ". Error code: "
+                + response.getStatusLine().getStatusCode() + " Current hostname: " + _myHostName
+                + " Leader hostname: " + leaderHostName;
         if (respEntity != null) {
           errorLog += " Response: " + EntityUtils.toString(respEntity);
         }
@@ -347,7 +364,6 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
           endPoint, _myHostName, leaderHostName, e);
       return false;
     }
-
     return true;
   }
 }
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
@@ -20,7 +20,7 @@ package org.apache.helix.rest.metadatastore.accessor;
  */
 
 import java.io.IOException;
-import java.util.Collections;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
@@ -167,8 +167,13 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
             _namespace, realm, shardingKey, e);
         return false;
       }
-      znRecord.setListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY,
-          Collections.singletonList(shardingKey));
+      List<String> shardingKeys =
+          znRecord.getListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY);
+      if (shardingKeys == null || shardingKeys.isEmpty()) {
+        shardingKeys = new ArrayList<>();
+      }
+      shardingKeys.add(shardingKey);
+      znRecord.setListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY, shardingKeys);
       try {
         _zkClient.writeData(realmPath, znRecord);
       } catch (Exception e) {
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
@@ -56,6 +56,7 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
   private final String _myHostName;
 
   public ZkRoutingDataWriter(String namespace, String zkAddress) {
+    System.out.println("writer created namespace=" + namespace + " zkAddress=" + zkAddress);
     if (namespace == null || namespace.isEmpty()) {
       throw new IllegalArgumentException("namespace cannot be null or empty!");
     }
@@ -102,7 +103,7 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
       return createZkRealm(realm);
     }
 
-    String url_suffix = "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm;
+    String url_suffix = "namespaces/" + _namespace + "/metadata-store-realms/" + realm;
     return forwardRequestToLeader(url_suffix, "put", "addMetadataStoreRealm",
         Response.Status.CREATED.getStatusCode());
   }
@@ -113,10 +114,10 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
       if (_zkClient.isClosed()) {
         throw new IllegalStateException("ZkClient is closed!");
       }
-      return _zkClient.delete(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm);
+      return deleteZkRealm(realm);
     }
 
-    String url_suffix = "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm;
+    String url_suffix = "namespaces/" + _namespace + "/metadata-store-realms/" + realm;
     return forwardRequestToLeader(url_suffix, "delete", "deleteMetadataStoreRealm",
         Response.Status.OK.getStatusCode());
   }
@@ -127,48 +128,11 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
       if (_zkClient.isClosed()) {
         throw new IllegalStateException("ZkClient is closed!");
       }
-      // If the realm does not exist already, then create the realm
-      String realmPath = MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm;
-      if (!_zkClient.exists(realmPath)) {
-        // Create the realm
-        if (!createZkRealm(realm)) {
-          // Failed to create the realm - log and return false
-          LOG.error(
-              "Failed to add sharding key because ZkRealm creation failed! Namespace: {}, Realm: {}, Sharding key: {}",
-              _namespace, realm, shardingKey);
-          return false;
-        }
-      }
-
-      ZNRecord znRecord;
-      try {
-        znRecord = _zkClient.readData(realmPath);
-      } catch (Exception e) {
-        LOG.error(
-            "Failed to read the realm ZNRecord in addShardingKey()! Namespace: {}, Realm: {}, ShardingKey: {}",
-            _namespace, realm, shardingKey, e);
-        return false;
-      }
-      List<String> shardingKeys =
-          znRecord.getListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY);
-      if (shardingKeys == null || shardingKeys.isEmpty()) {
-        shardingKeys = new ArrayList<>();
-      }
-      shardingKeys.add(shardingKey);
-      znRecord.setListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY, shardingKeys);
-      try {
-        _zkClient.writeData(realmPath, znRecord);
-      } catch (Exception e) {
-        LOG.error(
-            "Failed to write the realm ZNRecord in addShardingKey()! Namespace: {}, Realm: {}, ShardingKey: {}",
-            _namespace, realm, shardingKey, e);
-        return false;
-      }
-      return true;
+      return createZkShardingKey(realm, shardingKey);
     }
 
     String url_suffix =
-        "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm + "/sharding-keys/"
+        "namespaces/" + _namespace + "/metadata-store-realms/" + realm + "/sharding-keys/"
             + shardingKey;
     return forwardRequestToLeader(url_suffix, "put", "addShardingKey",
         Response.Status.CREATED.getStatusCode());
@@ -180,31 +144,11 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
       if (_zkClient.isClosed()) {
         throw new IllegalStateException("ZkClient is closed!");
       }
-      ZNRecord znRecord =
-          _zkClient.readData(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm, true);
-      if (znRecord == null || !znRecord
-          .getListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY)
-          .contains(shardingKey)) {
-        // This realm does not exist or shardingKey doesn't exist. Return true!
-        return true;
-      }
-      znRecord.getListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY)
-          .remove(shardingKey);
-      // Overwrite this ZNRecord with the sharding key removed
-      try {
-        _zkClient
-            .writeData(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm, znRecord);
-      } catch (Exception e) {
-        LOG.error(
-            "Failed to write the data back in deleteShardingKey()! Namespace: {}, Realm: {}, ShardingKey: {}",
-            _namespace, realm, shardingKey, e);
-        return false;
-      }
-      return true;
+      return deleteZkShardingKey(realm, shardingKey);
     }
 
     String url_suffix =
-        "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm + "/sharding-keys/"
+        "namespaces/" + _namespace + "/metadata-store-realms/" + realm + "/sharding-keys/"
             + shardingKey;
     return forwardRequestToLeader(url_suffix, "delete", "deleteShardingKey",
         Response.Status.OK.getStatusCode());
@@ -274,7 +218,7 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
    * @param realm
    * @return
    */
-  private boolean createZkRealm(String realm) {
+  protected boolean createZkRealm(String realm) {
     if (_zkClient.exists(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm)) {
       LOG.warn("createZkRealm() called for realm: {}, but this realm already exists! Namespace: {}",
           realm, _namespace);
@@ -292,10 +236,80 @@ public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
     return true;
   }
 
+  protected boolean deleteZkRealm(String realm) {
+    return _zkClient.delete(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm);
+  }
+
+  protected boolean createZkShardingKey(String realm, String shardingKey) {
+    // If the realm does not exist already, then create the realm
+    String realmPath = MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm;
+    if (!_zkClient.exists(realmPath)) {
+      // Create the realm
+      if (!createZkRealm(realm)) {
+        // Failed to create the realm - log and return false
+        LOG.error(
+            "Failed to add sharding key because ZkRealm creation failed! Namespace: {}, Realm: {}, Sharding key: {}",
+            _namespace, realm, shardingKey);
+        return false;
+      }
+    }
+
+    ZNRecord znRecord;
+    try {
+      znRecord = _zkClient.readData(realmPath);
+    } catch (Exception e) {
+      LOG.error(
+          "Failed to read the realm ZNRecord in addShardingKey()! Namespace: {}, Realm: {}, ShardingKey: {}",
+          _namespace, realm, shardingKey, e);
+      return false;
+    }
+    List<String> shardingKeys =
+        znRecord.getListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY);
+    if (shardingKeys == null || shardingKeys.isEmpty()) {
+      shardingKeys = new ArrayList<>();
+    }
+    shardingKeys.add(shardingKey);
+    znRecord.setListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY, shardingKeys);
+    try {
+      _zkClient.writeData(realmPath, znRecord);
+    } catch (Exception e) {
+      LOG.error(
+          "Failed to write the realm ZNRecord in addShardingKey()! Namespace: {}, Realm: {}, ShardingKey: {}",
+          _namespace, realm, shardingKey, e);
+      return false;
+    }
+    return true;
+  }
+
+  protected boolean deleteZkShardingKey(String realm, String shardingKey) {
+    ZNRecord znRecord =
+        _zkClient.readData(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm, true);
+    if (znRecord == null || !znRecord
+        .getListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY)
+        .contains(shardingKey)) {
+      // This realm does not exist or shardingKey doesn't exist. Return true!
+      return true;
+    }
+    znRecord.getListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY)
+        .remove(shardingKey);
+    // Overwrite this ZNRecord with the sharding key removed
+    try {
+      _zkClient
+          .writeData(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm, znRecord);
+    } catch (Exception e) {
+      LOG.error(
+          "Failed to write the data back in deleteShardingKey()! Namespace: {}, Realm: {}, ShardingKey: {}",
+          _namespace, realm, shardingKey, e);
+      return false;
+    }
+    return true;
+  }
+
   private boolean forwardRequestToLeader(String url_suffix, String request_method, String endPoint,
       int expectedResponseCode) throws IllegalArgumentException {
     String leaderHostName = _leaderElection.getCurrentLeaderInfo().getId();
     String url = leaderHostName + url_suffix;
+    System.out.println(url);
     HttpUriRequest request;
     switch (request_method) {
       case "put":
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import javax.ws.rs.core.Response;
 
 import org.apache.helix.msdcommon.constant.MetadataStoreRoutingConstants;
+import org.apache.helix.rest.common.HttpConstants;
 import org.apache.helix.rest.metadatastore.concurrency.ZkDistributedLeaderElection;
 import org.apache.helix.zookeeper.api.client.HelixZkClient;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
@@ -48,6 +49,8 @@ import org.slf4j.LoggerFactory;
 
 
 public class ZkRoutingDataWriter implements MetadataStoreRoutingDataWriter {
+  // Time out for http requests that are forwarded to leader instances measured in milliseconds
+  private static final int HTTP_REQUEST_FORWARDING_TIMEOUT = 60 * 1000;
   private static final Logger LOG = LoggerFactory.getLogger(ZkRoutingDataWriter.class);
 
   private final String _namespace;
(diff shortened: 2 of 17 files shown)


helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -77,7 +88,23 @@ public synchronized boolean addMetadataStoreRealm(String realm) {
...
+    String leaderHostName = _leaderElection.getCurrentLeaderInfo().getId();
+    String url =
+        leaderHostName + "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm;
+    HttpPut putRequest = new HttpPut(url);
+    try {
+      HttpResponse response = _forwardHttpClient.execute(putRequest);
+      if (response.getStatusLine().getStatusCode() != 201) {
+        HttpEntity respEntity = response.getEntity();
+        LOG.error("the forwarded request to leader has failed for addMetadataStoreRealm: {}",
+            respEntity != null ? EntityUtils.toString(respEntity) : "");
```
narendly: What's the point of recording ""? Do you really need the ternary operator in this case?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -90,7 +117,24 @@ public synchronized boolean deleteMetadataStoreRealm(String realm) {
       return _zkClient.delete(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm);
     }
 
-    // TODO: Forward the request to leader
+    String leaderHostName = _leaderElection.getCurrentLeaderInfo().getId();
+    String url =
+        leaderHostName + "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm;
```
narendly: See above - refactor!

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -77,7 +88,23 @@ public synchronized boolean addMetadataStoreRealm(String realm) {
       return createZkRealm(realm);
     }
 
-    // TODO: Forward the request to leader
+    String leaderHostName = _leaderElection.getCurrentLeaderInfo().getId();
+    String url =
+        leaderHostName + "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm;
+    HttpPut putRequest = new HttpPut(url);
+    try {
+      HttpResponse response = _forwardHttpClient.execute(putRequest);
```
narendly: Could we configure a timeout for these requests? 

https://www.baeldung.com/httpclient-timeout

Something like 1 minute sounds reasonable. Make that 1 minute a constant somewhere and you could use it throughout.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -90,7 +117,24 @@ public synchronized boolean deleteMetadataStoreRealm(String realm) {
...
+    String leaderHostName = _leaderElection.getCurrentLeaderInfo().getId();
+    String url =
+        leaderHostName + "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm;
+    HttpDelete deleteRequest = new HttpDelete(url);
+    try {
+      HttpResponse response = _forwardHttpClient.execute(deleteRequest);
+      if (response.getStatusLine().getStatusCode() != 200) {
+        HttpEntity respEntity = response.getEntity();
+        LOG.error("the forwarded request to leader has failed for deleteMetadataStoreRealm: {}",
+            respEntity != null ? EntityUtils.toString(respEntity) : "");
```
narendly: Why is the ternary necessary?

Did you meant to have a log msg like:

"the forwarded request to leader has failed for deleteMetadataStoreRealm:  "?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -136,7 +180,23 @@ public synchronized boolean addShardingKey(String realm, String shardingKey) {
       return true;
     }
 
-    // TODO: Forward the request to leader
+    String leaderHostName = _leaderElection.getCurrentLeaderInfo().getId();
+    String url =
+        leaderHostName + "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm
+            + "/sharding-keys/" + shardingKey;
```
narendly: refactor?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -225,6 +302,11 @@ public synchronized boolean setRoutingData(Map<String, List<String>> routingData
   @Override
   public synchronized void close() {
     _zkClient.close();
+    try {
+      _forwardHttpClient.close();
+    } catch (IOException e) {
+      LOG.error("HttpClient fails to close. ", e);
```
narendly: Nit: "failed"

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -90,9 +90,13 @@ public Response getAllMetadataStoreRealms() {
...
   public Response addMetadataStoreRealm(@PathParam("realm") String realm) {
     try {
-      _metadataStoreDirectory.addMetadataStoreRealm(_namespace, realm);
+      if (!_metadataStoreDirectory.addMetadataStoreRealm(_namespace, realm)) {
+        return serverError();
+      }
     } catch (IllegalArgumentException ex) {
       return notFound(ex.getMessage());
+    } catch (IllegalStateException ex) {
+      return serverError();
```
narendly: Where are you throwing an IllegalStateException?

metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/constant/MetadataStoreRoutingConstants.java
```diff
@@ -37,5 +37,6 @@
   // Field name in JSON REST response of getting sharding keys.
   public static final String SHARDING_KEYS = "shardingKeys";
 
-
+  // The key for system properties that contains the hostname of an instance
+  public static final String HOSTNAME_SYSTEM_PROPERTY_KEY = "instance_hostname";
```
narendly: Nit: rename to MSDS_SERVER_HOSTNAME_KEY?

And for the actual value, let's use "msds_hostname" so it doesn't conflict with some general "instance_hostname" key.

metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/constant/MetadataStoreRoutingConstants.java
```diff
@@ -37,5 +37,6 @@
   // Field name in JSON REST response of getting sharding keys.
   public static final String SHARDING_KEYS = "shardingKeys";
 
-
+  // The key for system properties that contains the hostname of an instance
```
narendly: hostname of the MSDS server instance

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -77,7 +88,23 @@ public synchronized boolean addMetadataStoreRealm(String realm) {
       return createZkRealm(realm);
     }
 
-    // TODO: Forward the request to leader
+    String leaderHostName = _leaderElection.getCurrentLeaderInfo().getId();
+    String url =
+        leaderHostName + "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm;
```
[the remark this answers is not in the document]
NealSun96: Agreed.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: When there's no response body, it is still necessary to log "the forward request has failed". Constructing it like this reduces duplication. 
Maybe I can change "" to "No response body is returned" or something else meaningful.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Ack.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: I'll take a look. Very good point.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Ditto my reply above.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Ack.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: In `ZkRoutingDataWriter` if `_zkClient.isClosed()`.

metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/constant/MetadataStoreRoutingConstants.java
NealSun96: Sure.

metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/constant/MetadataStoreRoutingConstants.java
NealSun96: Ack.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -62,10 +71,12 @@ public ZkRoutingDataWriter(String namespace, String zkAddress) {
     }
 
     // Get the hostname (REST endpoint) from System property
-    // TODO: Fill in when Helix REST implementations are ready
-    ZNRecord myServerInfo = new ZNRecord("dummy hostname");
+    ZNRecord myServerInfo = new ZNRecord(
+        System.getProperty(MetadataStoreRoutingConstants.HOSTNAME_SYSTEM_PROPERTY_KEY));
```
pkuwm: What if the property is not set in system?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -77,7 +88,23 @@ public synchronized boolean addMetadataStoreRealm(String realm) {
...
     }
 
-    // TODO: Forward the request to leader
+    String leaderHostName = _leaderElection.getCurrentLeaderInfo().getId();
+    String url =
+        leaderHostName + "/admin/v2/namespaces/" + _namespace + "/metadata-store-realms/" + realm;
+    HttpPut putRequest = new HttpPut(url);
+    try {
+      HttpResponse response = _forwardHttpClient.execute(putRequest);
+      if (response.getStatusLine().getStatusCode() != 201) {
```
pkuwm: I would use created code rather than the hard coded 201: easier to read and maintain

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: If that’s the case, I would put it in addMetadatStoreRealm or a proper place, but not here in this endpoint logic. It is not endpoint logic’s responsibility to process such exception.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -153,9 +161,13 @@ public Response getShardingKeys(@QueryParam("realm") String realm) {
   public Response addShardingKey(@PathParam("realm") String realm,
       @PathParam("sharding-key") String shardingKey) {
     try {
-      _metadataStoreDirectory.addShardingKey(_namespace, realm, shardingKey);
+      if (!_metadataStoreDirectory.addShardingKey(_namespace, realm, shardingKey)) {
+        return serverError();
```
pkuwm: Can we have a message in the server error message why it fails or what the client should do? Server error is too plain.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: +1. We could probably throw an IllegalStateException here if System.getProperty returns null.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: +1. Don't we already have this as part of Helix REST? See `javax.ws.rs.core.Response;`?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: You could do this:

```
String errorMessage = "The forwarded request from " + (this node's hostname) + " to leader " + (leader's hostname) + " has failed for addMetadataStoreRealm()!";
if (respEntity != null) {
  errorMessage += " Response: " + EntityUtils.toString(respEntity);
}
LOG. error(errorMessage);
```

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -77,7 +88,23 @@ public synchronized boolean addMetadataStoreRealm(String realm) {
...
+        HttpEntity respEntity = response.getEntity();
+        LOG.error("the forwarded request to leader has failed for addMetadataStoreRealm: {}",
+            respEntity != null ? EntityUtils.toString(respEntity) : "");
+        return false;
+      }
+    } catch (IOException e) {
+      LOG.error("the forwarded request to leader raised an exception for addMetadataStoreRealm", e);
```
narendly: More helpful to include which host is forwarding to which leader.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: See above

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -90,7 +117,24 @@ public synchronized boolean deleteMetadataStoreRealm(String realm) {
...
+      HttpResponse response = _forwardHttpClient.execute(deleteRequest);
+      if (response.getStatusLine().getStatusCode() != 200) {
+        HttpEntity respEntity = response.getEntity();
+        LOG.error("the forwarded request to leader has failed for deleteMetadataStoreRealm: {}",
+            respEntity != null ? EntityUtils.toString(respEntity) : "");
+        return false;
+      }
+    } catch (IOException e) {
+      LOG.error("the forwarded request to leader raised an exception for deleteMetadataStoreRealm",
+          e);
```
narendly: Improve logging with host name and leader host name

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
narendly: +1. Let's make sure the MSDS server logs the errors. it would be helpful if we could include as much detail as possible including the host name, etc.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -102,9 +106,13 @@ public Response addMetadataStoreRealm(@PathParam("realm") String realm) {
...
   public Response deleteMetadataStoreRealm(@PathParam("realm") String realm) {
     try {
-      _metadataStoreDirectory.deleteMetadataStoreRealm(_namespace, realm);
+      if (!_metadataStoreDirectory.deleteMetadataStoreRealm(_namespace, realm)) {
+        return serverError();
+      }
     } catch (IllegalArgumentException ex) {
       return notFound(ex.getMessage());
+    } catch (IllegalStateException ex) {
+      return serverError();
```
narendly: Remove?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: If it's not handled, the exception will cause a server error anyways. Are you suggesting to handle it and not raise server error? IMO zkClient being closed is definitely a server error.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Good point. Fixing.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Sure.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: There are numerous different reasons why this could happen, and one single returned message is not going to cover all cases; saying it's a server error is sufficient to express that it's a server side situation. 
On the other hand, whenever `false` is returned here, there are server logs that can help us identify the cause.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: As in the hostnames? Okay.

narendly: Like this:

`String errorMessage = "The forwarded request from " + (this node's hostname) + " to leader " + (leader's hostname) + " has failed for addMetadataStoreRealm()!";
if (respEntity != null) {
  errorMessage += " Response: " + EntityUtils.toString(respEntity);
}
LOG. error(errorMessage);`

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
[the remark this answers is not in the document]
NealSun96: `_leaderHostName` is an instance variable now since it's used during multiple occasions. I'm refactoring the calling code since it can be generalized and reduce duplication.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
narendly: You could let this IllegalStateException float up all the way to the servlet..

For instance, for uncaught exceptions, the servlet container usually catch the Exception, logs it and returns http status 500.

Just make sure MetadataStoreDirectory logs the exception properly.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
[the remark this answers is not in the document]
NealSun96: On a second thought I cannot do that because `_leaderHostName` will be changing. Modified my planned code structure for that. Duplication can still be reduced.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: I see. I'll remove this section of the code. Thank you both for the insight.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: That is for server side. Think about client side: what is response for this? If you are a user, what would you like to know?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: @narendly What's your opinion on the necessity of setting a socket timeout in addition to a connection timeout?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: Here are the situations where this condition can be invoked:
1. When the request forwarded to the leader instance returns a non-ok;
2. When the request forwarded to the leader raises an IOException;
3. When writing to ZooKeeper fails. 
All of them happen on the layer of `ZkRoutingDataWriter`. I'll need some advice on how to propagate the correct information back to the REST layer - I do not see an easy way to do this. 
At the same time, I believe it's unnecessary to let clients know the exact situation among these 3 situations, because these situations are unrelated to client input and there's nothing clients can do. All these situations are strictly internal errors and are out of the hands of clients.

pkuwm: I understand the possible situations and there is no need to tell the client the real reason (and it is impossible here because the api just returns true or false). I meant, include a message suggesting what to do like “server error , please contact admin” is helpful than plain response. Imagine, what would you get if you do curl this endpoint but server error? 
Anyway, I have an issue Created to improve the response. You don’t have to do it here. Focus on the main task of the pr instead.

narendly:
Note on this PR:

1. As discussed offline, let's make ZkMetadataStoreDirectory a static singleton that is shared throughout the server's process. The benefit of this is that each REST server will truly act as one entity, which is necessary because we use distributed leader election to designate a single writer. 

2. With that said, we would need to add both unit and integration tests. With an introduction of a singleton class, testing might not be so straightforward, but as discussed in person, it might be possible by creating mock classes.

Good job and let's continue adding tests and reviewing this PR :)

NealSun96:
Current state of the PR: missing integration tests for request forwarding, which will be added to `TestMetadataStoreDirectoryAccessor`. Everything else are ready for review.

NealSun96:
Integration tests have been added.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -42,49 +42,77 @@
...
+      synchronized (ZkMetadataStoreDirectory.class) {
+        if (_zkMetadataStoreDirectoryInstance == null) {
+          _zkMetadataStoreDirectoryInstance = new ZkMetadataStoreDirectory();
+        }
+      }
     }
-    _routingDataReaderMap = new HashMap<>();
-    _routingDataWriterMap = new HashMap<>();
-    _routingZkAddressMap = routingZkAddressMap;
+    _zkMetadataStoreDirectoryInstance.init(namespace, zkAddress);
```
narendly: Can you do

getInstance().init(namespace, zkAddress) here to cut down on duplicate code?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -40,8 +52,11 @@
   private final String _namespace;
   private final HelixZkClient _zkClient;
   private final ZkDistributedLeaderElection _leaderElection;
+  private final CloseableHttpClient _forwardHttpClient;
+  private final String _myHostName;
 
   public ZkRoutingDataWriter(String namespace, String zkAddress) {
+    System.out.println("writer created namespace=" + namespace + " zkAddress=" + zkAddress);
```
narendly: Remove print statements? We want to avoid this kind of stuff in production code.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -62,10 +77,21 @@ public ZkRoutingDataWriter(String namespace, String zkAddress) {
     }
 
     // Get the hostname (REST endpoint) from System property
-    // TODO: Fill in when Helix REST implementations are ready
-    ZNRecord myServerInfo = new ZNRecord("dummy hostname");
+    _myHostName = System.getProperty(MetadataStoreRoutingConstants.MSDS_SERVER_HOSTNAME_KEY);
+    if (_myHostName == null) {
```
narendly: Also include the check: `_myHostName.isEmpty()`

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -62,10 +77,21 @@ public ZkRoutingDataWriter(String namespace, String zkAddress) {
...
+      throw new IllegalStateException(
+          "Unable to get the hostname of this server instance. System.getProperty fails to fetch "
+              + MetadataStoreRoutingConstants.MSDS_SERVER_HOSTNAME_KEY + ".");
+    }
+    ZNRecord myServerInfo = new ZNRecord(_myHostName);
+
     _leaderElection = new ZkDistributedLeaderElection(_zkClient,
         MetadataStoreRoutingConstants.LEADER_ELECTION_ZNODE, myServerInfo);
+
+    int timeout = 60; // seconds
```
narendly: Let's make this a Java constant so that we could make this configurable later.

`private static final int HTTP_REQUEST_TIMEOUT = 60;`

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -77,8 +103,9 @@ public synchronized boolean addMetadataStoreRealm(String realm) {
       return createZkRealm(realm);
     }
 
-    // TODO: Forward the request to leader
-    return true;
+    String url_suffix = "namespaces/" + _namespace + "/metadata-store-realms/" + realm;
```
narendly: 1. Generally a good idea to refactor this to a private method so that it's easier to read. For instance, create `private String getUrlSuffix()`
2. Some of these string literals have already been defined in MetadataStoreDirectoryConstants.java. Could you use the constants there please?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -77,8 +103,9 @@ public synchronized boolean addMetadataStoreRealm(String realm) {
...
+    return forwardRequestToLeader(url_suffix, "put", "addMetadataStoreRealm",
```
narendly: Same here, let's make these String constants.

I believe "addMetadataStoreRealm" should come from MetadataStoreDirectoryConstants.
"put" could be a private enum called 
```
enum RequestForwardingHttpVerbs {
    PUT,
    POST,
    DELETE
}
``` something like this.

File: helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java [hunk 23]
```diff
@@ -87,11 +114,12 @@ public synchronized boolean deleteMetadataStoreRealm(String realm) {
       if (_zkClient.isClosed()) {
         throw new IllegalStateException("ZkClient is closed!");
       }
-      return _zkClient.delete(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realm);
+      return deleteZkRealm(realm);
     }
 
-    // TODO: Forward the request to leader
-    return true;
+    String url_suffix = "namespaces/" + _namespace + "/metadata-store-realms/" + realm;
```
narendly: Refactor into a private method.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -87,11 +114,12 @@ public synchronized boolean deleteMetadataStoreRealm(String realm) {
...
+    return forwardRequestToLeader(url_suffix, "delete", "deleteMetadataStoreRealm",
```
narendly: Use constants. No magic string literals floating around the code.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -100,44 +128,14 @@ public synchronized boolean addShardingKey(String realm, String shardingKey) {
...
+      return createZkShardingKey(realm, shardingKey);
     }
 
-    // TODO: Forward the request to leader
-    return true;
+    String url_suffix =
+        "namespaces/" + _namespace + "/metadata-store-realms/" + realm + "/sharding-keys/"
+            + shardingKey;
+    return forwardRequestToLeader(url_suffix, "put", "addShardingKey",
+        Response.Status.CREATED.getStatusCode());
```
narendly: Replace string literals.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -146,31 +144,14 @@ public synchronized boolean deleteShardingKey(String realm, String shardingKey)
...
+      return deleteZkShardingKey(realm, shardingKey);
     }
 
-    // TODO: Forward the request to leader
-    return true;
+    String url_suffix =
+        "namespaces/" + _namespace + "/metadata-store-realms/" + realm + "/sharding-keys/"
+            + shardingKey;
+    return forwardRequestToLeader(url_suffix, "delete", "deleteShardingKey",
+        Response.Status.OK.getStatusCode());
```
narendly: String literals.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -249,4 +235,119 @@ private boolean createZkRealm(String realm) {
...
+    String url = leaderHostName + url_suffix;
+    System.out.println(url);
+    HttpUriRequest request;
+    switch (request_method) {
+      case "put":
+        request = new HttpPut(url);
+        break;
+      case "delete":
+        request = new HttpDelete(url);
+        break;
```
narendly: Use an enum here instead of String literals.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -249,4 +235,119 @@ private boolean createZkRealm(String realm) {
...
+      String endPoint, String leaderHostName) {
+    try {
+      HttpResponse response = _forwardHttpClient.execute(request);
+      if (response.getStatusLine().getStatusCode() != expectedResponseCode) {
+        HttpEntity respEntity = response.getEntity();
+        String errorLog =
+            "The forwarded request to leader has failed for " + endPoint + ". Current hostname: "
+                + _myHostName + " Leader hostname: " + leaderHostName;
+        if (respEntity != null) {
+          errorLog += " Response: " + EntityUtils.toString(respEntity);
```
narendly: Also a good idea to log the response code.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -249,4 +235,119 @@ private boolean createZkRealm(String realm) {
...
+        LOG.error(errorLog);
+        return false;
+      }
+    } catch (IOException e) {
+      LOG.error(
+          "The forwarded request to leader raised an exception for {}. Current hostname: {} Leader hostname: {}",
+          endPoint, _myHostName, leaderHostName, e);
+      return false;
+    }
+
```
narendly: Remove empty line?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -64,6 +66,8 @@ public ServerContext(String zkAddr) {
     // cannot be started correctly.
     _helixDataAccessorPool = new HashMap<>();
     _taskDriverPool = new HashMap<>();
+    // Initiate the singleton ZkMetadataStoreDirectory instance to allow it to be closed later
```
narendly: Nit: "Initialize" would the the term.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -152,8 +155,13 @@ public Response getShardingKeys(@QueryParam("realm") String realm) {
   @Path("/metadata-store-realms/{realm}/sharding-keys/{sharding-key: .+}")
   public Response addShardingKey(@PathParam("realm") String realm,
       @PathParam("sharding-key") String shardingKey) {
+    if (shardingKey.charAt(0) != '/') {
+      shardingKey = "/" + shardingKey;
+    }
```
narendly: Could you add some comments about the Jersey behavior here for reference as a block comment?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -165,8 +173,13 @@ public Response addShardingKey(@PathParam("realm") String realm,
   @Path("/metadata-store-realms/{realm}/sharding-keys/{sharding-key: .+}")
   public Response deleteShardingKey(@PathParam("realm") String realm,
       @PathParam("sharding-key") String shardingKey) {
+    if (shardingKey.charAt(0) != '/') {
+      shardingKey = "/" + shardingKey;
+    }
```
narendly: Now this is duplicate code. Could you refactor this out to a private method? You could explain the JAX-RS behavior around slashes there.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.helix.rest.metadatastore.accessor.MetadataStoreRoutingDataReader;
+import org.apache.helix.rest.metadatastore.accessor.MetadataStoreRoutingDataWriter;
+import org.apache.helix.rest.metadatastore.accessor.ZkRoutingDataReader;
+import org.apache.helix.rest.metadatastore.accessor.ZkRoutingDataWriter;
+import org.apache.helix.rest.server.resources.metadatastore.MetadataStoreDirectoryAccessor;
+
+
+/**
+ * An accessor that mocks the MetadataStoreDirectoryAccessor for testing purpose.
+ */
```
narendly: Also add here that the reason we are not putting this in the test package is because the Jersey framework does not recognize resources in the test package when it spins up servers.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
```diff
@@ -0,0 +1,118 @@
...
+@Path("/mock")
+public class MockMetadataStoreDirectoryAccessor extends MetadataStoreDirectoryAccessor {
+  // A flag that will be modified if the underlying MockZkRoutingDataWriter makes an operation
+  // against ZooKeeper
+  public static boolean operatedOnZk = false;
```
narendly: Static boolean -> make it all caps?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * This method is overriden so that an instance of MockZkMetadataStoreDirectory can be passed in
+   */
+  @Override
+  protected void buildMetadataStoreDirectory(String namespace, String address) {
+    try {
+      _metadataStoreDirectory = new MockZkMetadataStoreDirectory(namespace, address);
+      _mockMSDInstance = _metadataStoreDirectory;
+    } catch (InvalidRoutingDataException e) {
+      e.printStackTrace();
```
narendly: Do not leave any stacktraces here. Let's use LOG.error(...)

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  /**
+   * Used to artificially create another instance of ZkMetadataStoreDirectory.
+   * ZkMetadataStoreDirectory being a singleton makes it impossible to test it,
```
narendly: Not impossible, but difficult. You're making it happen here ;)

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.helix.rest.metadatastore.accessor.ZkRoutingDataReader;
+import org.apache.helix.rest.metadatastore.accessor.ZkRoutingDataWriter;
+import org.apache.helix.rest.server.resources.metadatastore.MetadataStoreDirectoryAccessor;
+
+
+/**
+ * An accessor that mocks the MetadataStoreDirectoryAccessor for testing purpose.
+ */
+@Path("/mock")
+public class MockMetadataStoreDirectoryAccessor extends MetadataStoreDirectoryAccessor {
```
narendly: Could you create an issue and link that issue here as a TODO? The TODO should talk about how you should make this a configurable thing where developers could just feed in configs (like # of servers and names of resources to be included) to start multiple mock server instances. The purpose is to test features that involve multiple servers (like distributed leader election, distributed locking, etc.)

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
```diff
@@ -0,0 +1,118 @@
...
+      _routingDataReaderMap.put(namespace, new ZkRoutingDataReader(namespace, zkAddress, this));
+      _routingDataWriterMap.put(namespace, new MockZkRoutingDataWriter(namespace, zkAddress));
+      _realmToShardingKeysMap.put(namespace, _routingDataReaderMap.get(namespace).getRoutingData());
+      _routingDataMap.put(namespace, new TrieRoutingData(_realmToShardingKeysMap.get(namespace)));
+    }
+
+    @Override
+    public void close() {
+      _routingDataReaderMap.values().forEach(MetadataStoreRoutingDataReader::close);
+      _routingDataWriterMap.values().forEach(MetadataStoreRoutingDataWriter::close);
```
narendly: Nit: Should you clear the maps?

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
```diff
@@ -40,15 +41,35 @@
   private static final String DUMMY_SHARDING_KEY = "SHARDING_KEY";
   private MetadataStoreRoutingDataWriter _zkRoutingDataWriter;
 
+  // MockWriter is used for testing request forwarding features in non-leader situations
+  class MockWriter extends ZkRoutingDataWriter {
```
narendly: Nit: MockZkRoutingDataWriter so there's no ambiguity?

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
```diff
@@ -40,15 +41,35 @@
...
+    HttpUriRequest calledRequest;
+
+    MockWriter(String namespace, String zkAddress) {
+      super(namespace, zkAddress);
+    }
+
+    // This method does not call super() because the http call should not be actually made
+    @Override
```
narendly: Shouldn't you make the HTTP call? Otherwise you won't be covering distributed leader election? I think we **should** make the HTTP call here to the leader. Otherwise it's not a true integration test.

Or is this meant to be a unit test?

helix-rest/src/test/java/org/apache/helix/rest/server/MetadataStoreDirectoryAccessorTestBase.java
```diff
@@ -0,0 +1,128 @@
...
+          return false;
+        }
+      }
+
+      return true;
+    }, TestHelper.WAIT_DURATION), "Routing data path should be deleted after the tests.");
+  }
+
+  protected Set<String> getAllMetadataStoreRealmsHelper() throws InvalidRoutingDataException {
+    return new HashSet<>(_routingDataReader.getRoutingData().keySet());
```
narendly: Why do you need to create a new HashSet out of a set? 

You could probably return Collections.unmodifiableSet(....keySet()) as well if you want to prevent modifications

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
```diff
@@ -0,0 +1,225 @@
...
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.glassfish.jersey.server.ResourceConfig;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import static org.apache.helix.msdcommon.constant.MetadataStoreRoutingConstants.LEADER_ELECTION_ZNODE;
```
narendly: Avoid static imports. Just qualify with "MetadataStoreRoutingConstants."

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
```diff
@@ -0,0 +1,225 @@
...
+
+
+public class TestMSDADistributedLeaderElection extends MetadataStoreDirectoryAccessorTestBase {
```
narendly: Nit: I think `TestMSDAccessorLeaderElection` sounds better.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
```diff
@@ -0,0 +1,225 @@
...
+  private final String MOCK_URL_PREFIX = "/mock";
```
narendly: `private static final String` is how you declare a constant.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
```diff
@@ -0,0 +1,225 @@
...
+    _mockBaseUri =
+        getBaseUri().getScheme() + "://" + getBaseUri().getHost() + ":" + (getBaseUri().getPort()
+            + 1);
+    try {
+      List<HelixRestNamespace> namespaces = new ArrayList<>();
+      // Add test namespace
+      namespaces.add(new HelixRestNamespace(TEST_NAMESPACE,
+          HelixRestNamespace.HelixMetadataStoreType.ZOOKEEPER, _zkAddrTestNS, false));
+      _mockHelixRestServer =
+          new MockHelixRestServer(namespaces, getBaseUri().getPort() + 1, getBaseUri().getPath(),
```
narendly: We see a lot of `getPort() + 1`'s here. Let's just create a new variable,

int newPort = getPort() + 1; and use newPort throughout? This will make your code more readable.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
```diff
@@ -0,0 +1,225 @@
...
+              Collections.singletonList(_auditLogger));
+      _mockHelixRestServer.start();
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
narendly: LOG.error instead of printing stacktrace please.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
```diff
@@ -0,0 +1,225 @@
...
+    _zkClient.close();
+  }
+
+  @Test
+  public void testAddMetadataStoreRealmRequestForwarding()
+      throws InvalidRoutingDataException, IOException {
+    Set<String> expectedRealmsSet = getAllMetadataStoreRealmsHelper();
+    Assert.assertFalse(expectedRealmsSet.contains(TEST_REALM_3),
+        "Metadata store directory should not have realm: " + TEST_REALM_3);
+    sendRequestAndValidate("/metadata-store-realms/" + TEST_REALM_3, "put",
```
narendly: "put" could be an enum.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
```diff
@@ -0,0 +1,225 @@
...
+    expectedRealmsSet.add(TEST_REALM_3);
+    Assert.assertEquals(getAllMetadataStoreRealmsHelper(), expectedRealmsSet);
+    MockMetadataStoreDirectoryAccessor._mockMSDInstance.close();
+  }
+
+  @Test(dependsOnMethods = "testAddMetadataStoreRealmRequestForwarding")
+  public void testDeleteMetadataStoreRealmRequestForwarding()
+      throws InvalidRoutingDataException, IOException {
+    Set<String> expectedRealmsSet = getAllMetadataStoreRealmsHelper();
+    sendRequestAndValidate("/metadata-store-realms/" + TEST_REALM_3, "delete",
```
narendly: Use enum for "delete"

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
```diff
@@ -0,0 +1,225 @@
...
+      int expectedResponseCode) throws IllegalArgumentException, IOException {
+    String url = _mockBaseUri + TEST_NAMESPACE_URI_PREFIX + MOCK_URL_PREFIX + url_suffix;
+    HttpUriRequest request;
+    switch (request_method) {
+      case "put":
+        request = new HttpPut(url);
+        break;
+      case "delete":
+        request = new HttpDelete(url);
+        break;
```
narendly: Enum!

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: This one escaped my own review. Fixing.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Makes sense.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Ack.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: 1. I wanted to include the url creation as a part of `forwardRequestToLeader`, but I do not see a good way to generalize url construction. The 4 endpoints require 2 different urls; refactoring it results in an if statement, which isn't generalization. On the other hand, leaving url construction logic in the methods also clearly indicates which urls each endpoints correspond to, which is good for maintenance. 
2. In `MetadataStoreRoutingConstants`, the constants are not quite the same as the strings in these urls. (`shardingKeys` vs `sharding-keys`, `metadataStoreRealms` vs `metadata-store-realms`)

narendly:
@NealSun96 Please don't forget to rebase onto apache/zooscalability as well.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: I agree with this. Let me add an enum for it.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: 1. I do see a lot of duplication here. It seems it's possible to achieve the refactoring. Let me sync with you privately.

2. If you don't have them, you can create them. That's what the constant file is for. Please clearly note what they are used for.

NealSun96: Resolved by offline discussion. A method that joins parameters will be used for url suffix building.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Added the response code on line 336 because it always exists. Thank you for the suggestion.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
NealSun96: I agree.

NealSun96: Hm, I think my eyes were caught by the "initiate" in the comment on line 63, therefore I wrote "initiate" too.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: Fixed together with the later occurrence.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: Done so.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
NealSun96: This might not be true. On Monday we assumed that Jersey isn't able to recognize it in the test package, and we moved this file into the same package as `MetadataStoreDirectoryAccessor` for it to work, hence our conclusion. However, on Tuesday, after moving it to its own package (not in the test package, but its own package), it was not working either. 
After research, I realized that its package path was not correctly added. It was working on Monday because the package for `MetadataStoreDirectoryAccessor` was added already, causing this file to be added as well. There were some major changes to `MockHelixRestServer` for it to work in this current package. Therefore, the conclusion on Monday is not true anymore - the reason it wasn't working might not be due to the test package. 
Regardless, I think it's fine to leave this file here. However, if you believe this file is better in the test package, I can experiment with that and verify our conclusion.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
NealSun96: This is not a constant variable, though. It's a flag value that's used to identify if there were zookeeper operations. Should it still be all caps?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
NealSun96: Ack.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
NealSun96: It's not necessary because MockMSD is not a singleton. A new instance is created each time.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
NealSun96: Ditto below. This mock writer is for unit testing `ZkRoutingDataWriter`, it's not the one for integration testing. Calling this `MockZkRoutingDataWriter` actually makes it ambiguous because that would create two classes of `MockZkRoutingDataWriter`.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
NealSun96: I think you got the wrong file. :) 
The `MockZkRoutingDataWriter` for integration testing is in `MockMetadataStoreDirectoryAccessor.java`, and that's where the actual HTTP calls are made to the leader. This file is for unit-testing, where only the forwarding logic is tested.

helix-rest/src/test/java/org/apache/helix/rest/server/MetadataStoreDirectoryAccessorTestBase.java
NealSun96: Because in `TestMetadataStoreDirectoryAccessor` we need to modify this set. For example: 
```
Set<String> expectedRealmsSet = getAllMetadataStoreRealmsHelper();
expectedRealmsSet.add(TEST_REALM_3);
```
`keySet()` is unmodifiable. Making a copy makes it modifiable.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
NealSun96: Ack.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
NealSun96: Sure, saying "leader election" is sufficient.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
NealSun96: Oops. Fixing it.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
NealSun96: Sure.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
NealSun96: Ack.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMSDADistributedLeaderElection.java
NealSun96: Using the same enum as the other class.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
NealSun96: https://github.com/apache/helix/issues/816 I'll link this in the code.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: That is a good point. Let me try that.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -62,10 +77,24 @@ public ZkRoutingDataWriter(String namespace, String zkAddress) {
...
+    ZNRecord myServerInfo = new ZNRecord(_myHostName);
+
     _leaderElection = new ZkDistributedLeaderElection(_zkClient,
         MetadataStoreRoutingConstants.LEADER_ELECTION_ZNODE, myServerInfo);
+
+    RequestConfig config = RequestConfig.custom()
+        .setConnectTimeout(MetadataStoreRoutingConstants.HTTP_REQUEST_FORWARDING_TIMEOUT * 1000)
+        .setConnectionRequestTimeout(
+            MetadataStoreRoutingConstants.HTTP_REQUEST_FORWARDING_TIMEOUT * 1000)
+        .setSocketTimeout(MetadataStoreRoutingConstants.HTTP_REQUEST_FORWARDING_TIMEOUT * 1000)
```
narendly: Is there a strong and compelling reason why you couldn't include * 1000 in the constant itself?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -77,8 +106,11 @@ public synchronized boolean addMetadataStoreRealm(String realm) {
       return createZkRealm(realm);
     }
 
-    // TODO: Forward the request to leader
-    return true;
+    String urlSuffix =
+        constructUrlSuffix(MetadataStoreRoutingConstants.MSDS_GET_ALL_REALMS_ENDPOINT, realm);
+    return forwardRequestToLeader(urlSuffix,
+        MetadataStoreRoutingConstants.HttpRequestForwardingVerbs.PUT, "addMetadataStoreRealm",
```
narendly: 1. If you want these HTTP verbs included in `MetadataStoreRoutingConstants `, you could use a more generalized name like `HttpVerbs`.  `HttpRequestForwardingVerbs` if you want to keep it as an enum in this class.
2. If you want to generalize and put this in MetadataStoreRoutingConstants, then I'm pretty sure you could already find the supported HTTP verb enum somewhere defined in helix-rest module. Use that instead.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -146,31 +152,15 @@ public synchronized boolean deleteShardingKey(String realm, String shardingKey)
...
+      return deleteZkShardingKey(realm, shardingKey);
     }
 
-    // TODO: Forward the request to leader
-    return true;
+    String urlSuffix =
+        constructUrlSuffix(MetadataStoreRoutingConstants.MSDS_GET_ALL_REALMS_ENDPOINT, realm,
+            MetadataStoreRoutingConstants.MSDS_GET_ALL_SHARDING_KEYS_ENDPOINT, shardingKey);
+    return forwardRequestToLeader(urlSuffix,
+        MetadataStoreRoutingConstants.HttpRequestForwardingVerbs.DELETE, "deleteShardingKey",
```
narendly: I still see a string literal here?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -249,4 +244,126 @@ private boolean createZkRealm(String realm) {
...
+      return false;
+    }
+    return true;
+  }
+
+  private String constructUrlSuffix(String... urlParams) {
+    List<String> allUrlParameters = new ArrayList<>(
+        Arrays.asList(MetadataStoreRoutingConstants.MSDS_NAMESPACES_URL_PREFIX, _namespace));
+    allUrlParameters.addAll(Arrays.asList(urlParams));
+    // HttpUriRequest will convert all double slashes to single slashes
```
narendly: Do not rely on this behavior. Let's make sure this method creates a valid URL.

Do you think making this assumption is hacky? Could we improve it?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
narendly: Why wasn't it working then? I'd like to know the reason. If you don't think putting this class in the test package causes a problem, should we move it back to the test package?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
narendly: I'm actually fine either way since this is just a global created for testing.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
narendly: OK.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
narendly: OK

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
narendly: Ok, so a unit test.

helix-rest/src/test/java/org/apache/helix/rest/server/MetadataStoreDirectoryAccessorTestBase.java
narendly: getAllMetadataStoreRealmsHelper -> this naming is a little confusing. Consider renaming or explain what it does and why it's needed.

metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/constant/MetadataStoreRoutingConstants.java
```diff
@@ -20,6 +20,11 @@
  */
 
 public class MetadataStoreRoutingConstants {
+  public enum HttpRequestForwardingVerbs {
```
narendly: If you want to do this, I think you could find HTTP verbs defined elsewhere already.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
NealSun96: It wasn't working because the file's package was never added properly. Our method of overriding silently failed. 
Sure, let me experiment with moving it to a test package and let's see if it still works.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Good point. Let me move the multiplication.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Created `HttpConstants.RestVerbs` in `org.apache.helix.rest.common` that serves the enum purpose.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Removing this method name parameter altogether because it's better to use url to log. The url is more verbose.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: We now take out all possible double slashes.

helix-rest/src/test/java/org/apache/helix/rest/server/MetadataStoreDirectoryAccessorTestBase.java
NealSun96: Renamed and added comments.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/mock/MockMetadataStoreDirectoryAccessor.java
NealSun96: The experiment was successful. It shows that it works in the test package too. Now the file sits in `test.org.apache.helix.rest.server.mock`.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -42,49 +42,69 @@
...
+
+  public static ZkMetadataStoreDirectory getInstance() {
+    if (_zkMetadataStoreDirectoryInstance == null) {
+      synchronized (ZkMetadataStoreDirectory.class) {
+        if (_zkMetadataStoreDirectoryInstance == null) {
+          _zkMetadataStoreDirectoryInstance = new ZkMetadataStoreDirectory();
+        }
+      }
+    }
+
```
narendly: Remove empty line

NealSun96:
Final commit message:
## Implement request forwarding for ZkRoutingDataWriter ##
This PR added the request forwarding feature to ZkRoutingDataWriter. It also included a lot of other work, most notably: changing ZkMetadataStoreDirectory to singleton in order to allow leader forwarding, added integration tests for the request forwarding flow, modified MetadataStoreDirectoryAccessor for it to respect underlying return values, fixed numerous behavior bugs.

Issue #779, opened 2020-02-19
Implement FederatedZkClient to Support Multi-Realm CRUD and Callback


pkuwm:
FederatedZkClient (One FederatedZkClient, multiple Zookeeper sessions inside)
- Some Helix Java APIs such as HelixAdmin can only perform CRUD operations.
- In order to enable such Java APIs to operate against multiple Zookeeper realms, we will create FederatedZkClient, which is a new version of ZkClient that could replace the existing HelixZkClient used by CRUD-only Java APIs.
- FederatedZkClient will internally maintain multiple ZooKeeper sessions connecting to different ZooKeeper realms on an as-needed basis.
- By creating and maintaining multiple ZooKeeper sessions, FederatedZkClient will route incoming ZooKeeper requests to the appropriate ZooKeeper sessions based on the ZK path sharding key.

FederatedZkClient (implements RealmAwareZkClient)
- An implementation of RealmAwareZkClient that connects to multiple ZK realms.
- Supports CRUD + change subscription.

PR #789, opened 2020-02-20
Add FederatedZkClient

pkuwm:
### Issues

Implements #779 

### Description

FederatedZkClient (One FederatedZkClient, multiple Zookeeper sessions inside)
- Some Helix Java APIs such as HelixAdmin can only perform CRUD operations.
- In order to enable such Java APIs to operate against multiple Zookeeper realms, we will create FederatedZkClient, which is a new version of ZkClient that could replace the existing HelixZkClient used by CRUD-only Java APIs.
- FederatedZkClient will internally maintain multiple ZooKeeper sessions connecting to different ZooKeeper realms on an as-needed basis.
- By creating and maintaining multiple ZooKeeper sessions, FederatedZkClient will route incoming ZooKeeper requests to the appropriate ZooKeeper sessions based on the ZK path sharding key.

FederatedZkClient (implements RealmAwareZkClient)
- An implementation of RealmAwareZkClient that connects to multiple ZK realms.
- Supports CRUD + change subscription.

### Tests

 - TestFederatedZkClient

```
[INFO] --- maven-surefire-plugin:3.0.0-M3:test (default-test) @ zookeeper-api ---
[INFO]
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.024 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO]
[INFO] --- maven-surefire-plugin:3.0.0-M3:test (multi-zk) @ zookeeper-api ---
[INFO]
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------

[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.032 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  6.370 s
[INFO] Finished at: 2020-02-22T18:01:25-08:00
```

### Commits

### Documentation

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
@@ -26,7 +26,6 @@ import java.util.NoSuchElementException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
-import com.sun.istack.internal.NotNull;
 import org.apache.helix.msdcommon.datamodel.MetadataStoreRoutingData;
 import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.apache.helix.zookeeper.impl.factory.DedicatedZkClientFactory;
@@ -478,7 +477,6 @@ public class FederatedZkClient implements RealmAwareZkClient {
     return getZkClient(path).create(path, dataObject, acl, mode);
   }
 
-  @NotNull
   private ZkClient getZkClient(String path) {
     // If FederatedZkClient is closed, should not return ZkClient.
     checkClosedState();
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
@@ -69,6 +69,7 @@ public class FederatedZkClient implements RealmAwareZkClient {
   // ZK realm -> ZkClient
   private final Map<String, ZkClient> _zkRealmToZkClientMap;
 
+  private volatile boolean _isClosed;
   private PathBasedZkSerializer _pathBasedZkSerializer;
 
   // TODO: support capacity of ZkClient number in one FederatedZkClient and do garbage collection.
@@ -81,6 +82,7 @@ public class FederatedZkClient implements RealmAwareZkClient {
       throw new IllegalArgumentException("Client config cannot be null!");
     }
 
+    _isClosed = false;
     _clientConfig = clientConfig;
     _pathBasedZkSerializer = clientConfig.getZkSerializer();
     _metadataStoreRoutingData = metadataStoreRoutingData;
@@ -405,12 +407,13 @@ public class FederatedZkClient implements RealmAwareZkClient {
     synchronized (_zkRealmToZkClientMap) {
       _zkRealmToZkClientMap.values().forEach(ZkClient::close);
       _zkRealmToZkClientMap.clear();
+      _isClosed = true;
     }
   }
 
   @Override
   public boolean isClosed() {
-    return _zkRealmToZkClientMap.isEmpty();
+    return _isClosed;
   }
 
   @Override
@@ -453,6 +456,11 @@ public class FederatedZkClient implements RealmAwareZkClient {
   }
 
   private ZkClient getZkClient(String path) {
+    // If FederatedZkClient is closed, should not return ZkClient.
+    if (isClosed()) {
+      throw new IllegalStateException(FEDERATED_ZK_CLIENT + " is closed!");
+    }
+
     String zkRealm = getZkRealm(path);
     if (!_zkRealmToZkClientMap.containsKey(zkRealm)) {
       // Synchronized to avoid creating duplicate ZkClient for the same ZkRealm.
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
@@ -25,6 +25,7 @@ import java.util.NoSuchElementException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
+import com.sun.istack.internal.NotNull;
 import org.apache.helix.msdcommon.datamodel.MetadataStoreRoutingData;
 import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.apache.helix.zookeeper.impl.factory.DedicatedZkClientFactory;
@@ -56,10 +57,11 @@ import org.slf4j.LoggerFactory;
  * requests to the corresponding ZkClient with the help of metadata store directory service.
  * It could connect to multiple ZK addresses and maintain a {@link ZkClient} for each ZK address.
  * <p>
- * Note: while the ordering of handling data/child changes listeners in one single ZK realm is
- * guaranteed, listeners from different ZK realms are NOT guaranteed in order, which means listeners
- * from different ZK realms could be handled concurrently. So the concurrency of listeners should be
- * aware of when implementing listeners for different ZK realms.
+ * Note: each Zk realm has its own event queue to handle listeners. So listeners from different ZK
+ * realms could be handled concurrently because listeners of a ZK realm are handled in its own
+ * queue. The concurrency of listeners should be aware of when implementing listeners for different
+ * ZK realms. The users should use thread-safe data structures if they wish to handle change
+ * callbacks.
  */
 public class FederatedZkClient implements RealmAwareZkClient {
   private static final Logger LOG = LoggerFactory.getLogger(FederatedZkClient.class);
@@ -460,6 +462,7 @@ public class FederatedZkClient implements RealmAwareZkClient {
     return getZkClient(path).create(path, dataObject, acl, mode);
   }
 
+  @NotNull
   private ZkClient getZkClient(String path) {
     // If FederatedZkClient is closed, should not return ZkClient.
     checkClosedState();
@@ -467,14 +470,21 @@ public class FederatedZkClient implements RealmAwareZkClient {
     String zkRealm = getZkRealm(path);
 
     // Use this zkClient reference to protect the returning zkClient from being null because of
-    // race condition.
+    // race condition. Once we get the reference, even _zkRealmToZkClientMap is cleared by closed(),
+    // this zkClient is not null which guarantees the returned value not null.
     ZkClient zkClient = _zkRealmToZkClientMap.get(zkRealm);
 
     if (zkClient == null) {
-      // Synchronized to avoid creating duplicate ZkClient for the same ZkRealm.
+      // 1. Synchronized to avoid creating duplicate ZkClient for the same ZkRealm.
+      // 2. Synchronized with close() to avoid creating new ZkClient when all ZkClients are
+      // being closed and _zkRealmToZkClientMap is being cleared.
       synchronized (_zkRealmToZkClientMap) {
-        // Check closed state again to avoid race condition and creating a new ZkClient
-        // after FederatedZkClient is already closed.
+        // Because of potential race condition: thread B to get ZkClient could be blocked by this
+        // synchronized, while thread A is executing closed() in its synchronized block. So thread B
+        // could still enter this synchronized block once A completes executing closed() and
+        // releases the synchronized lock.
+        // Check closed state again to avoid creating a new ZkClient after FederatedZkClient
+        // is already closed.
         checkClosedState();
 
         if (!_zkRealmToZkClientMap.containsKey(zkRealm)) {
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
@@ -20,344 +20,478 @@ package org.apache.helix.zookeeper.impl.client;
  */
 
 import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.helix.msdcommon.datamodel.MetadataStoreRoutingData;
 import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
+import org.apache.helix.zookeeper.impl.factory.DedicatedZkClientFactory;
 import org.apache.helix.zookeeper.zkclient.DataUpdater;
 import org.apache.helix.zookeeper.zkclient.IZkChildListener;
 import org.apache.helix.zookeeper.zkclient.IZkDataListener;
+import org.apache.helix.zookeeper.zkclient.ZkConnection;
 import org.apache.helix.zookeeper.zkclient.callback.ZkAsyncCallbacks;
-import org.apache.helix.zookeeper.zkclient.deprecated.IZkStateListener;
+import org.apache.helix.zookeeper.zkclient.IZkStateListener;
+import org.apache.helix.zookeeper.zkclient.serialize.BasicZkSerializer;
 import org.apache.helix.zookeeper.zkclient.serialize.PathBasedZkSerializer;
 import org.apache.helix.zookeeper.zkclient.serialize.ZkSerializer;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.Op;
 import org.apache.zookeeper.OpResult;
+import org.apache.zookeeper.ZooDefs;
 import org.apache.zookeeper.data.ACL;
 import org.apache.zookeeper.data.Stat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
+/**
+ * Implements and supports all ZK operations defined in interface {@link RealmAwareZkClient},
+ * except for session-aware operations such as creating ephemeral nodes, for which
+ * an {@link UnsupportedOperationException} will be thrown.
+ * <p>
+ * It acts as a single ZK client but will automatically route read/write/change subscription
+ * requests to the corresponding ZkClient with the help of metadata store directory service.
+ * It could connect to multiple ZK addresses and maintain a {@link ZkClient} for each ZK address.
+ */
 public class FederatedZkClient implements RealmAwareZkClient {
+  private static final Logger LOG = LoggerFactory.getLogger(FederatedZkClient.class);
+
+  private static final String FEDERATED_ZK_CLIENT = FederatedZkClient.class.getSimpleName();
+  private static final String DEDICATED_ZK_CLIENT_FACTORY =
+      DedicatedZkClientFactory.class.getSimpleName();
+
+  private final MetadataStoreRoutingData _metadataStoreRoutingData;
+  private final RealmAwareZkClient.RealmAwareZkClientConfig _clientConfig;
+
+  // ZK realm -> ZkClient
+  private final Map<String, ZkClient> _zkRealmToZkClientMap;
+
+  private PathBasedZkSerializer _pathBasedZkSerializer;
+
+  // TODO: support capacity of ZkClient number in one FederatedZkClient and do garbage collection.
+  public FederatedZkClient(RealmAwareZkClient.RealmAwareZkClientConfig clientConfig,
+      MetadataStoreRoutingData metadataStoreRoutingData) {
+    if (metadataStoreRoutingData == null) {
+      throw new IllegalArgumentException("MetadataStoreRoutingData cannot be null!");
+    }
+    if (clientConfig == null) {
+      throw new IllegalArgumentException("Client config cannot be null!");
+    }
+
+    _clientConfig = clientConfig;
+    _pathBasedZkSerializer = clientConfig.getZkSerializer();
+    _metadataStoreRoutingData = metadataStoreRoutingData;
+    _zkRealmToZkClientMap = new ConcurrentHashMap<>();
+  }
+
   @Override
   public List<String> subscribeChildChanges(String path, IZkChildListener listener) {
-    return null;
+    return getZkClient(path).subscribeChildChanges(path, listener);
   }
 
   @Override
   public void unsubscribeChildChanges(String path, IZkChildListener listener) {
-
+    getZkClient(path).unsubscribeChildChanges(path, listener);
   }
 
   @Override
   public void subscribeDataChanges(String path, IZkDataListener listener) {
-
+    getZkClient(path).subscribeDataChanges(path, listener);
   }
 
   @Override
   public void unsubscribeDataChanges(String path, IZkDataListener listener) {
-
+    getZkClient(path).unsubscribeDataChanges(path, listener);
   }
 
   @Override
   public void subscribeStateChanges(IZkStateListener listener) {
-
+    throwUnsupportedOperationException();
   }
 
   @Override
   public void unsubscribeStateChanges(IZkStateListener listener) {
+    throwUnsupportedOperationException();
+  }
 
+  @Override
+  public void subscribeStateChanges(
+      org.apache.helix.zookeeper.zkclient.deprecated.IZkStateListener listener) {
+    throwUnsupportedOperationException();
+  }
+
+  @Override
+  public void unsubscribeStateChanges(
+      org.apache.helix.zookeeper.zkclient.deprecated.IZkStateListener listener) {
+    throwUnsupportedOperationException();
   }
 
   @Override
   public void unsubscribeAll() {
-
+    _zkRealmToZkClientMap.values().forEach(ZkClient::unsubscribeAll);
   }
 
   @Override
   public void createPersistent(String path) {
-
+    createPersistent(path, false);
   }
 
   @Override
   public void createPersistent(String path, boolean createParents) {
-
+    createPersistent(path, createParents, ZooDefs.Ids.OPEN_ACL_UNSAFE);
   }
 
   @Override
   public void createPersistent(String path, boolean createParents, List<ACL> acl) {
-
+    getZkClient(path).createPersistent(path, createParents, acl);
   }
 
   @Override
   public void createPersistent(String path, Object data) {
-
+    create(path, data, CreateMode.PERSISTENT);
   }
 
   @Override
   public void createPersistent(String path, Object data, List<ACL> acl) {
-
+    create(path, data, acl, CreateMode.PERSISTENT);
   }
 
   @Override
   public String createPersistentSequential(String path, Object data) {
-    return null;
+    return create(path, data, CreateMode.PERSISTENT_SEQUENTIAL);
   }
 
   @Override
   public String createPersistentSequential(String path, Object data, List<ACL> acl) {
-    return null;
+    return create(path, data, acl, CreateMode.PERSISTENT_SEQUENTIAL);
   }
 
   @Override
   public void createEphemeral(String path) {
-
+    create(path, null, CreateMode.EPHEMERAL);
   }
 
   @Override
   public void createEphemeral(String path, String sessionId) {
-
+    createEphemeral(path, null, sessionId);
   }
 
   @Override
   public void createEphemeral(String path, List<ACL> acl) {
-
+    create(path, null, acl, CreateMode.EPHEMERAL);
   }
 
   @Override
   public void createEphemeral(String path, List<ACL> acl, String sessionId) {
-
+    create(path, null, acl, CreateMode.EPHEMERAL, sessionId);
   }
 
   @Override
   public String create(String path, Object data, CreateMode mode) {
-    return null;
+    return create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, mode);
   }
 
   @Override
-  public String create(String path, Object datat, List<ACL> acl, CreateMode mode) {
-    return null;
+  public String create(String path, Object data, List<ACL> acl, CreateMode mode) {
+    return create(path, data, acl, mode, null);
   }
 
   @Override
   public void createEphemeral(String path, Object data) {
-
+    create(path, data, CreateMode.EPHEMERAL);
   }
 
   @Override
   public void createEphemeral(String path, Object data, String sessionId) {
-
+    create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL, sessionId);
   }
 
   @Override
   public void createEphemeral(String path, Object data, List<ACL> acl) {
-
+    create(path, data, acl, CreateMode.EPHEMERAL);
   }
 
   @Override
   public void createEphemeral(String path, Object data, List<ACL> acl, String sessionId) {
-
+    create(path, data, acl, CreateMode.EPHEMERAL, sessionId);
   }
 
   @Override
   public String createEphemeralSequential(String path, Object data) {
-    return null;
+    return create(path, data, CreateMode.EPHEMERAL_SEQUENTIAL);
   }
 
   @Override
   public String createEphemeralSequential(String path, Object data, List<ACL> acl) {
-    return null;
+    return create(path, data, acl, CreateMode.EPHEMERAL_SEQUENTIAL);
   }
 
   @Override
   public String createEphemeralSequential(String path, Object data, String sessionId) {
-    return null;
+    return create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL_SEQUENTIAL,
+        sessionId);
   }
 
   @Override
   public String createEphemeralSequential(String path, Object data, List<ACL> acl,
       String sessionId) {
-    return null;
+    return create(path, data, acl, CreateMode.EPHEMERAL_SEQUENTIAL, sessionId);
   }
 
   @Override
   public List<String> getChildren(String path) {
-    return null;
+    return getZkClient(path).getChildren(path);
   }
 
   @Override
   public int countChildren(String path) {
-    return 0;
+    return getZkClient(path).countChildren(path);
   }
 
   @Override
   public boolean exists(String path) {
-    return false;
+    return getZkClient(path).exists(path);
   }
 
   @Override
   public Stat getStat(String path) {
-    return null;
+    return getZkClient(path).getStat(path);
   }
 
   @Override
   public boolean waitUntilExists(String path, TimeUnit timeUnit, long time) {
-    return false;
+    return getZkClient(path).waitUntilExists(path, timeUnit, time);
   }
 
   @Override
   public void deleteRecursively(String path) {
-
+    getZkClient(path).deleteRecursively(path);
   }
 
   @Override
   public boolean delete(String path) {
-    return false;
+    return getZkClient(path).delete(path);
   }
 
   @Override
+  @SuppressWarnings("unchecked")
   public <T> T readData(String path) {
-    return null;
+    return (T) readData(path, false);
   }
 
   @Override
   public <T> T readData(String path, boolean returnNullIfPathNotExists) {
-    return null;
+    return getZkClient(path).readData(path, returnNullIfPathNotExists);
   }
 
   @Override
   public <T> T readData(String path, Stat stat) {
-    return null;
+    return getZkClient(path).readData(path, stat);
   }
 
   @Override
   public <T> T readData(String path, Stat stat, boolean watch) {
-    return null;
+    return getZkClient(path).readData(path, stat, watch);
   }
 
   @Override
   public <T> T readDataAndStat(String path, Stat stat, boolean returnNullIfPathNotExists) {
-    return null;
+    return getZkClient(path).readData(path, stat, returnNullIfPathNotExists);
   }
 
   @Override
   public void writeData(String path, Object object) {
-
+    writeData(path, object, -1);
   }
 
   @Override
   public <T> void updateDataSerialized(String path, DataUpdater<T> updater) {
-
+    getZkClient(path).updateDataSerialized(path, updater);
   }
 
   @Override
-  public void writeData(String path, Object datat, int expectedVersion) {
-
+  public void writeData(String path, Object data, int expectedVersion) {
+    writeDataReturnStat(path, data, expectedVersion);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -65,6 +66,11 @@ private void postConstruct() {
     buildMetadataStoreDirectory(_namespace, helixRestNamespace.getMetadataStoreAddress());
   }
 
+  @PreDestroy
```
narendly: Let's try to have changes like this in a separate PR. Or you could create an issue and we could batch process them later.

pkuwm: Forgot to remove it. Not supposed to include it.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,486 @@
...
+
+  // ZK realm -> ZkClient
+  private final Map<String, ZkClient> _zkRealmToZkClientMap;
+
+  private volatile boolean _isClosed;
+  private PathBasedZkSerializer _pathBasedZkSerializer;
+
+  // TODO: support capacity of ZkClient number in one FederatedZkClient and do garbage collection.
+  public FederatedZkClient(RealmAwareZkClient.RealmAwareZkClientConfig clientConfig,
+      MetadataStoreRoutingData metadataStoreRoutingData) {
```
pkuwm: We will remove this metadata store routing data in param once http routing data reader is ready.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,486 @@
...
+    return new ZkClient(new ZkConnection(zkAddress), (int) _clientConfig.getConnectInitTimeout(),
+        _clientConfig.getOperationRetryTimeout(), _pathBasedZkSerializer,
+        _clientConfig.getMonitorType(), _clientConfig.getMonitorKey(),
+        _clientConfig.getMonitorInstanceName(), _clientConfig.isMonitorRootPathOnly());
+  }
+
+  private void throwUnsupportedOperationException() {
+    throw new UnsupportedOperationException(
+        "Session-aware operation is not supported by " + FEDERATED_ZK_CLIENT
+            + ". Instead, please use " + DEDICATED_ZK_CLIENT_FACTORY + " for this operation");
```
narendly: Nit: Should we tell people to use DedicatdZkClient instead of DedicatedZkClientFactory?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
```diff
@@ -0,0 +1,252 @@
...
+import org.apache.zookeeper.Op;
+import org.apache.zookeeper.Watcher;
+import org.apache.zookeeper.ZooDefs;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+
+public class TestFederatedZkClient extends ZkTestBase {
```
narendly: Great job on writing tests cases to cover all unsupported operations! This test looks great.

One more test case if I may suggest adding for FederatedZkClient is doing two writes to two different paths (path A and path B) where the two paths map to different ZK realms.

First, instrument the MetadataStoreRoutingData so that paths A and B have sharding keys that map to two different ZK realms. Then use the same FederatedZkClient to write to paths A and B.

Then create two single-realm ZkClients connected to the two realms respectively, and check that the data/ZNode you wrote to path A is in one ZK realm and not in the other, and vice versa.

Do you think that would be a good test case that tests FederatedZkClient's multi-realm functionality? :)

pkuwm: This is correct. We would definitely need this test to very it is actual a multi realm client. I’ve considered it and it is in progress. Because in the previous version there is only one zk server running, I could not  run multi realms. I was planning to start a zk server just for this test but was waiting to see if we want to start one more zk in zk tests base. I will push a test. Thanks for the review and suggestion!

narendly: @pkuwm 
Our ZkTestBase already starts multiple ZKs and you could reference them in _zkServerMap :)

You could look in our pom.xml file and search for "multiZk" keyword. I believe I have it set at 3. And you could use a protected variable `_numZk` to check how many ZKs have been started. That should make writing a test case for this much easier (start by doing a check - `if (_numZk > 1)` )

pkuwm: @narendly I've noticed that multi-zk execution is included in pom. But actually I not clear why we are running two times of tests. And actually in Intellij the test runs the default-tests but not multi-zk (maybe I need to config my Intellij).
Anyway, I've added a test for multi realm.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
pkuwm: `DedicatedZkClientFactory` is what we provide and recommend users to create a RealmAwareZkClient which is a `DedicatedZkClient` instance. `DedicatedZkClient ` is protected and we don't want users to directly use it, right?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
narendly: I don't think IntelliJ supports this - you could run it in command line with maven directly. That way, it will execute twice.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    if (rawRoutingData.size() < 2) {
+      System.out.println("There is only one ZK realm. Starting one more ZK to test multi-realm.");
+      _extraZkRealm = ZK_PREFIX + (ZK_START_PORT + 1);
+      _extraZkServer = startZkServer(_extraZkRealm);
+      // RealmTwo's sharding key: /test_sharding_key_2
+      List<String> shardingKeyList = Collections.singletonList(TEST_SHARDING_KEY_PREFIX + "2");
+      rawRoutingData.put(_extraZkRealm, shardingKeyList);
+    }
```
narendly: This is not necessary when the tests are being run by maven's surefire-plugin (what runs your mvn test's). Writing this block of code would be just for IntelliJ, so I wouldn't bother creating logic like this in your future tests.

You can have 2 or more ZKs in your tests, but we have two runs for a reason - we want to test it in a single ZK environment and also test it in a multi-ZK environment.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: I see what you're saying, but the way it's phrased here may confuse users.

Say this:

"Instead, please use DedicatedZkClientFactory to create an instance of dedicated RealmAwareZkClient for ephemeral operations".

This is because we don't use the factory for ephemeral operations.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
pkuwm: `if (_numZk > 1)` to check multi-zk env in each multi-realm test doesn't look natural to me. Why can we not just keep one execution which just initially starts multi ZK? Does it hurt the default tests which don't need multi-zk env?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
pkuwm: Same reply above.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
narendly: We could make an exception to tests written for ZkClient APIs, but in general, we want to run all tests in two environments: single-realm and multi-realm. 

If we only keep one execution with multiple ZKs, then we lose test coverage for the single-realm case.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,487 @@
...
   @Override
   public void unsubscribeDataChanges(String path, IZkDataListener listener) {
-
+    getZkClient(path).unsubscribeDataChanges(path, listener);
   }
 
   @Override
   public void subscribeStateChanges(IZkStateListener listener) {
-
+    throwUnsupportedOperationException();
```
kaisun2000: Add a message here in the exception. Something like " subscribeStateChanges not supported in FederatedZkClient"

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,487 @@
...
+          _zkRealmToZkClientMap.put(zkRealm, createZkClient(zkRealm));
+        }
+      }
+    }
+
+    return _zkRealmToZkClientMap.get(zkRealm);
+  }
+
+  private String getZkRealm(String path) {
+    String zkRealm;
```
kaisun2000: initialize this variable. 
maybe to null?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,487 @@
...
+  }
+
+  private ZkClient getZkClient(String path) {
+    // If FederatedZkClient is closed, should not return ZkClient.
+    if (isClosed()) {
+      throw new IllegalStateException(FEDERATED_ZK_CLIENT + " is closed!");
+    }
+
+    String zkRealm = getZkRealm(path);
+    if (!_zkRealmToZkClientMap.containsKey(zkRealm)) {
```
kaisun2000: concurrentMap, we should use  PutIfAbsent(), no need to reinvent the wheel, which may not be right.

See here. https://stackoverflow.com/questions/10743622/concurrenthashmap-avoid-extra-object-creation-with-putifabsent

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,487 @@
...
+      }
+    }
+
+    return _zkRealmToZkClientMap.get(zkRealm);
+  }
+
+  private String getZkRealm(String path) {
+    String zkRealm;
+    try {
+      zkRealm = _metadataStoreRoutingData.getMetadataStoreRealm(path);
```
kaisun2000: Is this _metadataStoreRoutingData query thread safe? 
@narendly, @dasahcc. 

This need to be validated.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,487 @@
...
+    }
+    if (clientConfig == null) {
+      throw new IllegalArgumentException("Client config cannot be null!");
+    }
+
+    _isClosed = false;
+    _clientConfig = clientConfig;
+    _pathBasedZkSerializer = clientConfig.getZkSerializer();
+    _metadataStoreRoutingData = metadataStoreRoutingData;
+    _zkRealmToZkClientMap = new ConcurrentHashMap<>();
```
kaisun2000: Nit, this can be done in the declaration place and make the _zkRealmToZkClientMap final.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,487 @@
...
+
+    if (zkRealm == null || zkRealm.isEmpty()) {
+      throw new NoSuchElementException("Cannot find ZK realm for the path: " + path);
+    }
+
+    return zkRealm;
+  }
+
+  private ZkClient createZkClient(String zkAddress) {
+    LOG.debug("Creating ZkClient for realm: {}.", zkAddress);
```
kaisun2000: Using this way to collect stats, we have individual internal dedicated zkclient stats. We can't have aggregated stats.

The question is that do we want to have collective status for one federatedZkClient?

kaisun2000:
One design concern:

The data/child listerner would be called on individual underlying DedicatedZkClient's event queue **concurrently**

Previously client does not really expect this behavior. If their code is not thread-safe, which they are probably not, they would have hard to track concurrency issue. 

I think we should schedule the callback in one thread. Or at least we should explicitly document this issue.  @dasahcc , @narendly , @jiajunwang , what is your take?

kaisun2000:
Another design concern:

I think the original design is that FederatedZKClient does not support session aware ephemeral node creation. 

Is there any reason we don't do it this way?

@narendly , @dasahcc , @jiajunwang

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: ^ You could still make the reference final and initialize in the constructor :)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: It's null by default

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: Does this need to be thread-safe? This data is statically populated at the very beginning and no one is updating this data, right (update not allowed)?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: Are you referring to ZkClient metrics? 

We've discussed this and I think it's worth considering, but that should be a TODO. Can we create an issue for it and tackle it at a later time?

kaisun2000: TODO seems to be good choice now. But exactly what needs to be done let us think about it.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
kaisun2000: _metadataStoreRoutingData.getMetadataStoreRealm for sure should be thread safe. All the public API in this one should be as they will be called by multiple thread. 

We probably should raise this point in the diff implementing this code. I just forget to look careful there. 

That said, it is not hard to make it thread safe as you said, think they are immutable once published. Wrapping the data to be published in a immutable collection seems to be the easiest way to me.

narendly: Could you look at how _metadataStoreRoutingData is implemented? It sounds like you think it's returning a map - it's actually an interface that does not provide any APIs that can modify the underlying data structure.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
pkuwm: @kaisun2000 Thanks for the suggestion. I am glad we got to this point.
- Did you mean `putIfAbsent(zkRealm, createZkClient(zkRealm))`? I did thought about it. However, it actually does not satisfy my expectation. `putIfAbsent(zkRealm, createZkClient(zkRealm))` runs `createZkClient(zkRealm)` every time, no matter if zkRealm presents or not. As you already know, creating a zkclient costs a lot of overhead: creating native zk instance, zk event thread, wait until connected, etc.. We definitely don't want to create a zkclient every time even one is already available, right?
- Then is there another API that could solve the pain we have above so no new zkclient is created if there is already one? Yes, that is `computeIfAbsent()`. We can make one time init and get() together as clean as `map.computeIfAbsent(zkRealm, zkclient -> createZkClient())`. This looks pretty good, right? But my only concern is that the get performance of `computeIfAbsent()` is not as good as simply a `get(zkRealm)`. Because internally in ConcurrentHashMap's implementation, `computeIfAbsent()` always synchronized the node/table, while `get()` directly reads the `transient volatile table` **without** synchonized/lock, which has a better performance over `computeIfAbsent()`.
- 'get()' + `synchronized`. Back to our use case, read is much much more than create/write because creation is only one time for each ZkClient. And I think the synchronized code block here is correct. This way makes sure creation/put is thread-safe, one time initialization does not hurt too much performance because read much heavier than write in our case. 

So I believe `get()` + `synchronized` satisfy us better. What do you think?

kaisun2000: Good point. ComputeifAbsent() is better. This guarantees no leaking of DedicatedZkClient. Note, client created must be close(). PutIfAbsent has a small chance of leaking though.

Get() + synchronized has its pitfall too. This is double-locking pattern. If test a simple value null or not, you need to make the value volatile. But for this one, I am not sure. 

So my suggestion is to use ComputeIfAbsent, and if possible, add a test case to make sure we don't leak zkClient (garbage collected, but did not call close())

pkuwm: As we discussed, we will keep current implementation, because:
- double-check lock here for the concurrent map is thread safe, as `computeIfAbsent()`
- `get()` is faster than `computeIfAbsent()` for getting an existing key, as we analyzed the code and did performance experiments to prove it
- synchronized `_zkRealmToZkClientMap` also protects the map from adding new zkclient, when `close()` is called.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,487 @@
...
 import org.apache.zookeeper.Op;
 import org.apache.zookeeper.OpResult;
+import org.apache.zookeeper.ZooDefs;
 import org.apache.zookeeper.data.ACL;
 import org.apache.zookeeper.data.Stat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
```
kaisun2000: Document the threading model change here. Namely that listeners (data/children) for ZKpath in different realm potentially can be invoked concurrently.  This means user should take care of currency issue in the listener implementation.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,487 @@
...
+    return 0L;
   }
 
   @Override
   public void close() {
+    if (isClosed()) {
+      return;
+    }
 
+    LOG.info("Closing {}.", FEDERATED_ZK_CLIENT);
```
narendly: Nit: usually we put the log at the very end. So move to the end of this method and say FederatedZkClient closed successfully?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
kaisun2000: This good implementation.  Please add a comment here why this is safe and performant ( https://stackoverflow.com/questions/7003239/double-checked-locking-with-concurrentmap)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,487 @@
...
   public long getSessionId() {
-    return 0;
+    // Session-aware is unsupported.
+    throwUnsupportedOperationException();
+    return 0L;
   }
 
   @Override
   public void close() {
+    if (isClosed()) {
```
kaisun2000: This is not thread safe way to closing. As discussed. We need to fix it.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
pkuwm: I thought about that when I was adding the log. The reason why I add logging here is because it gives us timestamp when its starts to close the clients and could potentially help us with troubleshooting (if there was any race condition related to `close()`). And actually each raw zkclient has has its own `closed` log after it is closed. So I prefer to put it at the very beginning.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
pkuwm: Added.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
pkuwm: Resolved. 
- We check closed state again before creating a new zkclient in the synchronized block. This helps with the case: the thread B starts to execute the createZkClient() synchronized block right after thread A executes close(). With current code now, check closed state again to avoid creating a new zkclient.
- We use a zkclient reference `ZkClient zkClient = _zkRealmToZkClientMap.get(zkRealm);` to avoid race condition: 
```
if (containsKey()) { // true
 // clear() could happen in this time gap.
 return get()  // returns null
}
```

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,506 @@
...
+    }
+
+    // Create mode is not session-aware, so the node does not have to be created
+    // by the expectedSessionId.
+    return getZkClient(path).create(path, dataObject, acl, mode);
+  }
+
+  private ZkClient getZkClient(String path) {
+    // If FederatedZkClient is closed, should not return ZkClient.
+    checkClosedState();
```
narendly: Why is this necessary? The raw ZkClient handles this already.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,506 @@
...
+        // after FederatedZkClient is already closed.
+        checkClosedState();
+
+        if (!_zkRealmToZkClientMap.containsKey(zkRealm)) {
+          zkClient = createZkClient(zkRealm);
+          _zkRealmToZkClientMap.put(zkRealm, zkClient);
+        }
+      }
+    }
+
```
narendly: Nit. remove empty line

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,506 @@
...
+        // Check closed state again to avoid race condition and creating a new ZkClient
+        // after FederatedZkClient is already closed.
+        checkClosedState();
+
+        if (!_zkRealmToZkClientMap.containsKey(zkRealm)) {
+          zkClient = createZkClient(zkRealm);
+          _zkRealmToZkClientMap.put(zkRealm, zkClient);
+        }
+      }
+    }
```
narendly: This is good. Makes sense. This will help us avoid zkClient leaks. Good work!

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,506 @@
...
+  }
+
+  private ZkClient getZkClient(String path) {
+    // If FederatedZkClient is closed, should not return ZkClient.
+    checkClosedState();
+
+    String zkRealm = getZkRealm(path);
+
+    // Use this zkClient reference to protect the returning zkClient from being null because of
+    // race condition.
```
narendly: This is fine, but it will never return null though right? There are only 2 outcomes that are possible - 1) we return a newly created ZkClient or 2) we throw an exception and not return anything.

narendly: The three options I listed are all fair game. I also agree that adding the else clause should do it :)

pkuwm: Option 1 you listed does not work 100% . It could still return null. That’s why I would like this way to have a zkclient reference.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,508 @@
...
+/**
+ * Implements and supports all ZK operations defined in interface {@link RealmAwareZkClient},
+ * except for session-aware operations such as creating ephemeral nodes, for which
+ * an {@link UnsupportedOperationException} will be thrown.
+ * <p>
+ * It acts as a single ZK client but will automatically route read/write/change subscription
+ * requests to the corresponding ZkClient with the help of metadata store directory service.
+ * It could connect to multiple ZK addresses and maintain a {@link ZkClient} for each ZK address.
+ * <p>
+ * Note: while the ordering of handling data/child changes listeners in one single ZK realm is
```
kaisun2000: This has nothing to do with "order" or sequential consistency in Zookeeper term. Let us don't confuse user with this consistency guarantee. 

From the user's perspective, they need to be aware their callback listerner can be invoked concurrently and should prepare to handle this concurrency accordingly, which is not the case before. Something like:

Federated ZKClient is designed to perform CRUD operation across different zookeeper realms. It would handle data and node change notification. It does not support ephemeral node operation. 

From concurrency point of view, the callback notification can be invoked concurrently for ZK path in different zookeeper realm. This is different from other ZkClient which will invoke the callback in one thread. Thus, it is the user's responsibility to take care this concurrency.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
kaisun2000: up to you. Most other codes does it my suggested way, though.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
kaisun2000: Up to you. If I were you, I would initialize it to null, easy for readers and easy to reason

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,508 @@
...
+
+  private ZkClient getZkClient(String path) {
+    // If FederatedZkClient is closed, should not return ZkClient.
+    checkClosedState();
+
+    String zkRealm = getZkRealm(path);
+
+    // Use this zkClient reference to protect the returning zkClient from being null because of
+    // race condition.
+    ZkClient zkClient = _zkRealmToZkClientMap.get(zkRealm);
```
kaisun2000: Please add a comment here that why this kind of double locking is working. The stackoverflow thread you found. Also, explain why here this is faster. This help people later maintaining the code understand why it is done this way.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,508 @@
...
+
+    if (zkClient == null) {
+      // Synchronized to avoid creating duplicate ZkClient for the same ZkRealm.
```
kaisun2000: Change the comment to state that this synchronize() is pair with close() synchronized. This is to avoid creating another internal dedicated ZkClient by thread A when thread B invoked close() which subsequently close all internal dedicated ZkClient.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,508 @@
...
+      synchronized (_zkRealmToZkClientMap) {
```
kaisun2000: Seems to me this is ok. 

We can reason it this way. The code in this synchronized() is pair with code in close() synchronized block. Thus, there are only two possibilities.
Case 1/ this block run first, the block in closed() run next. In this case, the newly created dedicated ZkClient in this block would be closed in the closed() block
Case 2/ this block run after the block in close(). In this case, we would not create dedicated ZkClient due to the flag of closed is set.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -20,344 +20,508 @@
...
+    }
+
+    // Create mode is not session-aware, so the node does not have to be created
+    // by the expectedSessionId.
+    return getZkClient(path).create(path, dataObject, acl, mode);
+  }
+
+  private ZkClient getZkClient(String path) {
+    // If FederatedZkClient is closed, should not return ZkClient.
+    checkClosedState();
```
kaisun2000: See the reasoning bellow, this is not necessary. Isn't it?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
```diff
@@ -0,0 +1,310 @@
...
+
+    // Client is closed, so operation should not be executed.
+    try {
+      _realmAwareZkClient.createPersistent(TEST_REALM_ONE_VALID_PATH);
+      Assert
+          .fail("createPersistent() should not be executed because RealmAwareZkClient is closed.");
+    } catch (IllegalStateException ex) {
+      Assert.assertEquals(ex.getMessage(), "FederatedZkClient is closed!");
+    }
+  }
```
kaisun2000: Can we add a test case that one thread close the FederatedZKClient and then another thread use it, which would generate exception?

kaisun2000:
Overall, it looks ok. 

The important things are:
1/ Add a test case for close()
2/ I think the first checkCloseState() in getZkClient is not necessary, isn't it?
3/ Add comment about synchronize block and why? I see we don't do it for other places and in production we see problems due to later change by another developer who did not fully grasp what the original author's concurrency intention. Thus, let us document it and give people who make change this later a hint.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: Also explicitly mentione that "the users should use thread-safe data structures if they wish to handle change callbacks".

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: I'm fine either way - 

1) we can fail early here
2) returning a closed ZkClient is also okay because the raw ZkClient will fail if it's closed anyway.

I am slightly in favor of option 2) because that would look more natural to users. (as in they continue getting the message from the raw ZkClient, as if they were using the raw ZkClient)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: I don't agree that we should add the link to the stack overflow post, but feel free to explain here why you know this would work.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
pkuwm: Like I said, it would be redundant here because there are only 2 possibilities:
- it is assigned a value: `zkRealm = _metadataStoreRoutingData.getMetadataStoreRealm(path);`
- or an exception is thrown. So an uninitialized zkRealm would NOT happen or be used. 
```
throw new NoSuchElementException("Cannot find ZK realm for the path: " + path);
```

It must be be initialized in below case:
```
String zkRealm;
try {
       zkRealm = _metadataStoreRoutingData.getMetadataStoreRealm(path);
This conversation was marked as resolved by pkuwm  Show conversation
} catch (NoSuchElementException ex) {
       // throw new NoSuchElementException("Cannot find ZK realm for the path: " + path);
       LOG.error();
}
// It may not have been initialized if exception is thrown.
System.out.println(zkRealm);
```

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
pkuwm: Since the first check is already tested here, I would say you wanted to test the second check within the synchronized, right? I would appreciate it if you could help and show me how to implement it.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: I didn't realize that we're clearing the map. I'm fine with doing the check and failing early here then so as not to return null. Returning null is bad.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
narendly: Testing multi-threaded code for race condition might not be so easy to do it here. I vote to create a separate issue for testing this sort of stuff if it's necessary. @kaisun2000 could provide more insight on that issue if he's willing.

For the purposes of adding FederatedZkClient, I do not think we need to create that test here.

kaisun2000: yeah, testing multi-thread version is probably not easy. More into thinking test no leak of internal zkClient. Specifically here, can we assert that the internal map from realm to zkclient is empty? And if possible the opened internal zkclient do get a close call. This is to ensure normal flow we don't leak Zookeeper session.

kaisun2000:
As discussed, two concerns:
1/ I think the "checkClosedState()" in the getZkClient is not needed for correctness. In fact, I am not sure how to reason it is correct. Put it this way, I don't know if it is correct or not. 
2/ Need a test case to ensure now Zookeeper session is leaking after close() is called.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -19,345 +19,535 @@
...
+
+      while (iterator.hasNext()) {
+        Map.Entry<String, ZkClient> entry = iterator.next();
+        String zkRealm = entry.getKey();
+        ZkClient zkClient = entry.getValue();
+
+        // Catch any exception from ZkClient's close() to avoid that there is leakage of
+        // remaining unclosed ZkClient.
+        try {
+          zkClient.close();
```
kaisun2000: Good!

pkuwm:
This PR is ready to be merged, approved by @kaisun2000 @narendly 
```
As part of ZkClient API enhancement, we wish to add FederatedZkClient, which is a wrapper of the raw ZkClient, that provides realm-aware access to ZooKeeper.
FederatedZkClient will internally maintain multiple ZooKeeper sessions connecting to different ZooKeeper realms on an as-needed basis and route requests to the appropriate ZooKeeper based on the ZK path sharding key. Ephemeral creation is not supported.
```

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -19,345 +19,535 @@
...
  */
 
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
+import com.sun.istack.internal.NotNull;
```
narendly: Can we not include this? I understand what you're trying to do here, but I don't think you're using this particular annotation the right way.

See https://stackoverflow.com/questions/42576760/where-is-notnull-in-java

Issue #798, opened 2020-02-22
Support getAllRealmKeyMappings Java API and REST endpoint


narendly:
This will help with reducing REST calls down to one single call to read all raw routing data in ZK.

PR #799, opened 2020-02-23
[helix-rest] Add endpoint to get namespace routing data

pkuwm:
### Issues

Implements #798 

### Description

Add Java API to support get mapping of all sharding keys by realm: realm -> list of sharding keys.
This will help with reducing REST calls down to one single call to read all raw routing data in ZK.

### Tests

 - testGetShardingKeysGroupByRealm

```
[INFO] Tests run: 128, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 28.216 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 128, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:01 min
[INFO] Finished at: 2020-02-22T20:29:35-08:00
```

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
@@ -58,7 +58,7 @@ public interface MetadataStoreDirectory extends AutoCloseable {
    * @param namespace namespace in metadata store directory.
    * @return Routing data map: realm -> List of sharding keys
    */
-  Map<String, List<String>> getRoutingData(String namespace);
+  Map<String, List<String>> getNamespaceRoutingData(String namespace);
 
   /**
    * Returns all path-based sharding keys in the given namespace and the realm.
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
@@ -20,6 +20,7 @@ package org.apache.helix.rest.metadatastore;
  */
 
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 
@@ -50,6 +51,14 @@ public interface MetadataStoreDirectory extends AutoCloseable {
    * @return
    */
   Collection<String> getAllShardingKeys(String namespace);
+
+  /**
+   * Returns all path-based sharding keys by realm in the given namespace.
+   *
+   * @param namespace namespace in metadata store directory.
+   * @return Map: realm -> List of sharding keys
+   */
+  Map<String, List<String>> getShardingKeysByRealm(String namespace);
 
   /**
    * Returns all path-based sharding keys in the given namespace and the realm.
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
@@ -53,12 +53,12 @@ public interface MetadataStoreDirectory extends AutoCloseable {
   Collection<String> getAllShardingKeys(String namespace);
 
   /**
-   * Returns all path-based sharding keys by realm in the given namespace.
+   * Returns routing data in the given namespace.
    *
    * @param namespace namespace in metadata store directory.
-   * @return Map: realm -> List of sharding keys
+   * @return Routing data map: realm -> List of sharding keys
    */
-  Map<String, List<String>> getShardingKeysByRealm(String namespace);
+  Map<String, List<String>> getRoutingData(String namespace);
 
   /**
    * Returns all path-based sharding keys in the given namespace and the realm.
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -112,7 +112,7 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
   }
 
   @Override
-  public Map<String, List<String>> getRoutingData(String namespace) {
+  public Map<String, List<String>> getNamespaceRoutingData(String namespace) {
     Map<String, List<String>> routingData = _realmToShardingKeysMap.get(namespace);
     if (routingData == null) {
       throw new NoSuchElementException("Namespace " + namespace + " does not exist!");
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -112,13 +112,13 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
   }
 
   @Override
-  public Map<String, List<String>> getShardingKeysByRealm(String namespace) {
-    Map<String, List<String>> shardingKeysByRealmMap = _realmToShardingKeysMap.get(namespace);
-    if (shardingKeysByRealmMap == null) {
+  public Map<String, List<String>> getRoutingData(String namespace) {
+    Map<String, List<String>> routingData = _realmToShardingKeysMap.get(namespace);
+    if (routingData == null) {
       throw new NoSuchElementException("Namespace " + namespace + " does not exist!");
     }
 
-    return shardingKeysByRealmMap;
+    return routingData;
   }
 
   @Override
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -112,6 +112,16 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
   }
 
   @Override
+  public Map<String, List<String>> getShardingKeysByRealm(String namespace) {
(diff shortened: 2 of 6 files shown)


helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -111,6 +111,16 @@ public ZkMetadataStoreDirectory(Map<String, String> routingZkAddressMap)
...
   }
 
+  @Override
+  public Map<String, List<String>> getShardingKeysByRealm(String namespace) {
+    Map<String, List<String>> shardingKeysByRealmMap = _realmToShardingKeysMap.get(namespace);
+    if (shardingKeysByRealmMap == null) {
+      throw new NoSuchElementException("Namespace " + namespace + " does not exist!");
+    }
+
+    return shardingKeysByRealmMap;
```
narendly:
```
if (!_realmToShardingKeysMap.contains(namespace)) {
  throw Exception(~~~);
}
return  _realmToShardingKeys.get(namespace);
```

seems like a more concise way of expressing this.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -167,11 +169,15 @@ public Response deleteMetadataStoreRealm(@PathParam("realm") String realm) {
...
+  public Response getShardingKeys(@QueryParam("prefix") String prefix,
+      @DefaultValue("false") @QueryParam("groupByRealm") boolean groupByRealm) {
     try {
-      if (prefix == null) {
+      if (prefix == null && !groupByRealm) {
         // For endpoint: "/sharding-keys" to get all sharding keys in a namespace.
         return getAllShardingKeys();
+      } else if (prefix == null) {
+        // For request: "/sharding-keys?groupByRealm=true" to get sharding keys by realm.
+        return getAllShardingKeysByRealm();
```
narendly:
```
if (prefix == null) {
  if (groupByRealm) {
    return getAllShardingKeysByRealm();
  }
  return getAllShardingKeys();
}
```

would be a more concise and readable way of expressing this logic. what do you think?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/datamodel/MetadataStoreShardingKeysByRealm.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+
+@JsonPropertyOrder({"realm", "shardingKeys"})
+public class MetadataStoreShardingKeysByRealm {
+  private String realm;
+  private List<String> shardingKeys;
```
narendly: Nit: do you think we should use Collection where appropriate to make things consistent? Let's first check if Jackson supports serialization of Collections.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
```diff
@@ -51,6 +52,14 @@
    */
   Collection<String> getAllShardingKeys(String namespace);
 
+  /**
+   * Returns all path-based sharding keys by realm in the given namespace.
+   *
+   * @param namespace namespace in metadata store directory.
+   * @return Map: realm -> List of sharding keys
+   */
+  Map<String, List<String>> getShardingKeysByRealm(String namespace);
```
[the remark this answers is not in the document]
pkuwm: I initially did considered it but when implementing the method, I found the `_realmToShardingKeysMap` returns a `Map<String, List<String>>`. If I make it List to Collection, we have to do extra work to iterate the map, create a new map and convert List to Collection. I just don't want to do this for performance consideration. I always try to make Rest API respond as fast as possible. As in this case, List is still fine over a Collection in the map. This is my thought.
Well, if you think this kind of consistency is more important and preferred over the performance, I am OK.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
pkuwm: I realized that we already consider multi-thread issue (reading and refreshing) for this map because it is a concurrent map. Your style `contains()` and `get()` would potentially have race condition. So only one `get()` makes advantage of a concurrent map to avoid race condition.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/datamodel/MetadataStoreShardingKeysByRealm.java
pkuwm: It should. Changed.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: Maybe the ways we think of if-else condition branch are different. I've changed it to another way which may look easier to understand.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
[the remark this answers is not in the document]
narendly: Great - thanks for providing your reasoning. I think this is minor, so if you feel that making it a List will simplify things, we could leave it as is.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -167,14 +183,22 @@ public Response deleteMetadataStoreRealm(@PathParam("realm") String realm) {
...
-  public Response getShardingKeys(@QueryParam("prefix") String prefix) {
+  public Response getShardingKeys(@QueryParam("prefix") String prefix,
+      @DefaultValue("false") @QueryParam("groupByRealm") boolean groupByRealm) {
     try {
-      if (prefix == null) {
-        // For endpoint: "/sharding-keys" to get all sharding keys in a namespace.
-        return getAllShardingKeys();
+      if (prefix != null) {
+        // For endpoint: "/sharding-keys?prefix={prefix}"
+        // Ignore groupByRealm because response already has realm info for each sharding key.
```
narendly: Have you considered creating a separate endpoint for retrieving all realm-list of keys mappings?

As I previously commented, I believe that making sure an endpoint does as little as possible and keeping it simple is an important part of a good API design. But now this endpoint has multiple queryParams, and I think this makes it a little more confusing. 

This is because this endpoint now does more than one thing depending on what queryParams are given, and in this case, the `groupByRealm` queryParam seems to be serving to specify an option/mode rather than a filter.

Could we please create a separate endpoint called `getAllRoutingData(String namespace)`? This way, we keep the original `getShardingKeys` simple, and there's no confusion about what each of these endpoints is doing. The endpoint name could be something like `/all-routing-data` or whatever you see fit.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
pkuwm: I actually don’t see your suggested version is simpler: the key is hashed two times (maybe JVM could cache hashcode) which causes more cpu time. It is more natural to just compute one time of key hash. Right, in our case here namespace could not change, but it is just a special case here. What if namespace change could be supported in the future? I would prefer to do it in a general way: one time hash and cache the result and generally avoid race condition. So decouple it from namespace change.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
pkuwm: This is a good name that I was trying to have. I tried to have a better name but did not realize that this is actually a routing data endpoint. I have to admit that I don't have a keen sense of `routing data`. I will refactor this.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
narendly: Sure - in this case either way works. Marking it as resolved 👍

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/MetadataStoreDirectory.java
```diff
@@ -51,6 +52,14 @@
    */
   Collection<String> getAllShardingKeys(String namespace);
 
+  /**
+   * Returns routing data in the given namespace.
+   *
+   * @param namespace namespace in metadata store directory.
+   * @return Routing data map: realm -> List of sharding keys
+   */
+  Map<String, List<String>> getRoutingData(String namespace);
```
narendly: Nit: should we rename this method to `getAllRoutingDataInNamespace(String namespace)`? That way there's no ambiguity and is more consistent with other methods' names in this class.

pkuwm: Changed to `getNamespaceRoutingData(namespace)` to avoid having `of`, `in`.

pkuwm:

```
RealmAwareZkClient construction needs a REST endpoint to get routing data in a namespace. This endpoint will help with reducing REST calls down to one single call to read all raw routing data in ZK.

This commit adds Java API getNamespaceRoutingData() and an endpoint "GET /routing-data".
```

PR #811, opened 2020-02-26
Fix zookeeper endpoint error when path has leading slash

### LINKED ISSUE #810: Zookeeper REST endpoint still prepends slash when path already has leading slash
If path has leading slash in Zookeeper REST endpoint: `/zookeeper//tmp/folder`, response is incorrect: response status code is 400.

This endpoint should be able to handle the case correctly.

pkuwm:
### Issues

Fixes #810

### Description

If path has leading slash in Zookeeper REST endpoint: `/zookeeper//tmp/folder`, response is incorrect: response status code is 400.

This endpoint should be able to handle the case correctly.

### Tests

 - testPathHasLeadingSlash

### Commits

### Documentation

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
@@ -62,14 +62,20 @@ public class ZooKeeperAccessor extends AbstractResource {
     if (cmd == null) {
       return badRequest("Invalid ZooKeeper command: " + commandStr);
     }
+    if (path == null) {
+      return badRequest("Path is missing in the request!");
+    }
 
     // Lazily initialize ZkBaseDataAccessor
     ServerContext _serverContext =
         (ServerContext) _application.getProperties().get(ContextPropertyKeys.SERVER_CONTEXT.name());
     _zkBaseDataAccessor = _serverContext.getByteArrayZkBaseDataAccessor();
 
-    // Need to prepend a "/" since JAX-RS regex removes it
-    path = "/" + path;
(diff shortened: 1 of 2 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -62,14 +62,20 @@ public Response get(@PathParam("path") String path, @QueryParam("command") Strin
...
+    }
 
     // Lazily initialize ZkBaseDataAccessor
     ServerContext _serverContext =
         (ServerContext) _application.getProperties().get(ContextPropertyKeys.SERVER_CONTEXT.name());
     _zkBaseDataAccessor = _serverContext.getByteArrayZkBaseDataAccessor();
 
-    // Need to prepend a "/" since JAX-RS regex removes it
-    path = "/" + path;
+    if (path.isEmpty() || path.charAt(0) != '/') {
```
dasahcc: This path is getting from URL. Why we have different behavior? Sometimes it has slash sometimes not?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -62,14 +62,20 @@ public Response get(@PathParam("path") String path, @QueryParam("command") Strin
     if (cmd == null) {
       return badRequest("Invalid ZooKeeper command: " + commandStr);
     }
+    if (path == null) {
```
narendly: I don't think this PR is necessary because it is the caller's responsibility to correctly format the URL. With the correct URL, it has already been verified and tested that these endpoints work as expected.

PR #819, opened 2020-02-27
Make RealmAwareZkClient implementations use HttpRoutingDataReader for routing data

### LINKED ISSUE #818: Use HttpRoutingDataReader in all implementations of RealmAwareZkClient
We want all implementations of RealmAwareZkClient to do a one-time query to Metadata Store Directory Service for routing data and cache it in memory. In order to accomplish that, we have introduced HttpRoutingDataReader, which is a Singleton class that makes a REST call to read routing data and caches it in memory. 

This issue tracks the work of updating the initialization logic in all instances of RealmAwareZkClient.

narendly:
### Issues

Fixes #818 

### Description

We want all implementations of RealmAwareZkClient to do a one-time query to Metadata Store Directory Service for routing data and cache it in memory. In order to accomplish that, we have introduced HttpRoutingDataReader, which is a Singleton class that makes a REST call to read routing data and caches it in memory. This diff updates the initialization logic in RealmAwareZkClients accordingly.
Changelist:
1. Update all RealmAwareZkClient initialization logic
2. Fix tests

### Tests

TestDedicatedZkClient
TestSharedZkClient
TestFederatedZkClient

- metadata-store-directory-common

```
[INFO] Tests run: 30, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.322 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 30, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  4.531 s
[INFO] Finished at: 2020-03-03T18:02:00-08:00
```
- zookeeper-api
```
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.57 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  7.604 s
[INFO] Finished at: 2020-03-03T18:02:21-08:00
```
- helix-rest
```
[INFO] Tests run: 136, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 25.661 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 136, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 


```
### Commits

### Code Quality

- [x] My diff has been formatted using helix-style.xml (helix-style-intellij.xml if IntelliJ IDE is used)

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/factory/RealmAwareZkClientFactory.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/factory/RealmAwareZkClientFactory.java
@@ -19,7 +19,9 @@ package org.apache.helix.zookeeper.api.factory;
  * under the License.
  */
 
-import org.apache.helix.msdcommon.datamodel.MetadataStoreRoutingData;
+import java.io.IOException;
+
+import org.apache.helix.msdcommon.exception.InvalidRoutingDataException;
 import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 
 
@@ -31,26 +33,20 @@ public interface RealmAwareZkClientFactory {
    * Build a RealmAwareZkClient using specified connection config and client config.
    * @param connectionConfig
    * @param clientConfig
-   * @param metadataStoreRoutingData
    * @return HelixZkClient
    */
-  // TODO: remove MetadataStoreRoutingData
   RealmAwareZkClient buildZkClient(RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig,
-      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig,
-      MetadataStoreRoutingData metadataStoreRoutingData);
+      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig)
+      throws IOException, InvalidRoutingDataException;
 
   /**
    * Builds a RealmAwareZkClient using specified connection config and default client config.
    * @param connectionConfig
-   * @param metadataStoreRoutingData
    * @return RealmAwareZkClient
    */
-
-  // TODO: remove MetadataStoreRoutingData
   default RealmAwareZkClient buildZkClient(
-      RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig,
-      MetadataStoreRoutingData metadataStoreRoutingData) {
-    return buildZkClient(connectionConfig, new RealmAwareZkClient.RealmAwareZkClientConfig(),
-        metadataStoreRoutingData);
+      RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig)
+      throws IOException, InvalidRoutingDataException {
+    return buildZkClient(connectionConfig, new RealmAwareZkClient.RealmAwareZkClientConfig());
   }
 }
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/factory/RealmAwareZkClientFactory.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/factory/RealmAwareZkClientFactory.java
@@ -33,7 +33,9 @@ public interface RealmAwareZkClientFactory {
    * Build a RealmAwareZkClient using specified connection config and client config.
    * @param connectionConfig
    * @param clientConfig
-   * @return HelixZkClient
+   * @return RealmAwareZkClient
+   * @throws IOException if Metadata Store Directory Service is unresponsive over HTTP
+   * @throws InvalidRoutingDataException if the routing data received is invalid or empty
    */
   RealmAwareZkClient buildZkClient(RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig,
       RealmAwareZkClient.RealmAwareZkClientConfig clientConfig)
@@ -43,6 +45,8 @@ public interface RealmAwareZkClientFactory {
    * Builds a RealmAwareZkClient using specified connection config and default client config.
    * @param connectionConfig
    * @return RealmAwareZkClient
+   * @throws IOException if Metadata Store Directory Service is unresponsive over HTTP
+   * @throws InvalidRoutingDataException if the routing data received is invalid or empty
    */
   default RealmAwareZkClient buildZkClient(
       RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig)
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
@@ -75,7 +75,7 @@ public class DedicatedZkClient implements RealmAwareZkClient {
       throw new IllegalArgumentException("RealmAwareZkConnectionConfig cannot be null!");
     }
     if (clientConfig == null) {
-      throw new IllegalArgumentException("RealmAwareZkConnectionConfig cannot be null!");
+      throw new IllegalArgumentException("RealmAwareZkClientConfig cannot be null!");
     }
 
     // Get the routing data from a static Singleton HttpRoutingDataReader
@@ -87,6 +87,10 @@ public class DedicatedZkClient implements RealmAwareZkClient {
     }
 
     _zkRealmShardingKey = connectionConfig.getZkRealmShardingKey();
+    if (_zkRealmShardingKey == null || _zkRealmShardingKey.isEmpty()) {
+      throw new IllegalArgumentException(
+          "RealmAwareZkConnectionConfig's ZK realm sharding key cannot be null or empty for DedicatedZkClient!");
+    }
 
     // Get the ZkRealm address based on the ZK path sharding key
     String zkRealmAddress = _metadataStoreRoutingData.getMetadataStoreRealm(_zkRealmShardingKey);
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
@@ -72,12 +72,21 @@ public class DedicatedZkClient implements RealmAwareZkClient {
   public DedicatedZkClient(RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig,
       RealmAwareZkClient.RealmAwareZkClientConfig clientConfig)
       throws IOException, InvalidRoutingDataException {
-    // Get the routing data from a static Singleton HttpRoutingDataReader
-    _metadataStoreRoutingData = HttpRoutingDataReader.getMetadataStoreRoutingData();
-
     if (connectionConfig == null) {
       throw new IllegalArgumentException("RealmAwareZkConnectionConfig cannot be null!");
     }
+    if (clientConfig == null) {
+      throw new IllegalArgumentException("RealmAwareZkConnectionConfig cannot be null!");
+    }
+
+    // Get the routing data from a static Singleton HttpRoutingDataReader
+    String msdsEndpoint = connectionConfig.getMsdsEndpoint();
+    if (msdsEndpoint == null || msdsEndpoint.isEmpty()) {
+      _metadataStoreRoutingData = HttpRoutingDataReader.getMetadataStoreRoutingData();
+    } else {
+      _metadataStoreRoutingData = HttpRoutingDataReader.getMetadataStoreRoutingData(msdsEndpoint);
+    }
+
     _zkRealmShardingKey = connectionConfig.getZkRealmShardingKey();
 
     // Get the ZkRealm address based on the ZK path sharding key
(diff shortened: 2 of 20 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/factory/RealmAwareZkClientFactory.java
```diff
@@ -31,25 +34,21 @@
    * Build a RealmAwareZkClient using specified connection config and client config.
    * @param connectionConfig
    * @param clientConfig
-   * @param metadataStoreRoutingData
    * @return HelixZkClient
```
kaisun2000: Document when/why IOExeption would be thrown and why InvalidRoutingDataException would be thrown

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/SharedZkClientFactory.java
```diff
@@ -45,21 +45,12 @@ protected SharedZkClientFactory() {
   @Override
   public RealmAwareZkClient buildZkClient(
       RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig,
-      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig,
-      MetadataStoreRoutingData metadataStoreRoutingData) {
+      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig) {
```
kaisun2000: This would be a conflict. Now my change is in, we need a rebase for this one.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
```diff
@@ -102,7 +102,8 @@ public static MetadataStoreRoutingData getMetadataStoreRoutingData()
    * @throws IOException
    */
   private static String getAllRoutingData() throws IOException {
-    // Note that MSDS_ENDPOINT should provide high-availability - it risks becoming a single point of failure if it's backed by a single IP address/host
+    // Note that MSDS_ENDPOINT should provide high-availability - it risks becoming a single point
+    // of failure if it's backed by a single IP address/host
```
kaisun2000: A little of topic. Do we have a issue to track this one? If we need to enhance this part, this is the number one issue in my opinion.

kaisun2000:
Overall looks good to me, just some nits.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/factory/RealmAwareZkClientFactory.java
narendly: That is correct. I will update the JavaDoc :)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/factory/SharedZkClientFactory.java
narendly: No worries.

I won't merge this in until other ZkClients have been merged in. I might include other ZkClients in the scope of this PR.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/HttpRoutingDataReader.java
narendly: We do not - could you please create an issue for it? Not an immediate blocker.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -80,19 +83,17 @@
   private PathBasedZkSerializer _pathBasedZkSerializer;
 
   // TODO: support capacity of ZkClient number in one FederatedZkClient and do garbage collection.
-  public FederatedZkClient(RealmAwareZkClient.RealmAwareZkClientConfig clientConfig,
-      MetadataStoreRoutingData metadataStoreRoutingData) {
-    if (metadataStoreRoutingData == null) {
-      throw new IllegalArgumentException("MetadataStoreRoutingData cannot be null!");
-    }
+  public FederatedZkClient(RealmAwareZkClient.RealmAwareZkClientConfig clientConfig)
+      throws IOException, InvalidRoutingDataException {
```
pkuwm: I don't think it is a good idea to declare exception in a constructor which I would try to avoid. For this case, I would catch them and throw an `IllegalStateException`.

narendly: Throwing exceptions in a constructor is not bad practice. In fact, it is the only way for a constructor to indicate that there is a problem; e.g. that the parameters/states are invalid.

This is a behavior agreed upon with @kaisun2000, which reminds me to add the Javadoc. IOException in the case of HTTP read failure, InvalidRoutingDataException in the case of an invalid routing data input. Users should be able to tell what's going on, and we let them distinguish failure cases by which type of exception we throw.

pkuwm: I understand it is OK to throw exceptions in a constructor. In this case, if we declare exceptions in the signature, would it make it inconvenient to initialize/build a FederatedZkClient with try...catch? Personally, I would not like that. If the routing data does not exist or IO exception, it means routing data is not available for this FederatedZkClient construction. A runtime exception is good enough: `IllegalStateException("Routing data is not available: " + e.getMessage())`. I wouldn't bother try...catch to construct a FederatedZkClient.

narendly: - FederatedZkClient is still a low-level construct
- This means that the users of this component will mostly be Helix Java APIs, and probably a few users who wish to use directly.
- In that case, I think we still need to float these exceptions up and handle each case at the Helix Java API level. That way, we make it 1) user-friendly and 2) we still achieve the low-level granularity with FederatedZkClient that users of this class would want instead of hiding failure types behind an IllegalStateException.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -56,32 +59,42 @@
...
-      MetadataStoreRoutingData metadataStoreRoutingData) {
-
+      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig)
+      throws IOException, InvalidRoutingDataException {
     if (connectionConfig == null) {
       throw new IllegalArgumentException("RealmAwareZkConnectionConfig cannot be null!");
     }
-    _zkRealmShardingKey = connectionConfig.getZkRealmShardingKey();
+    if (clientConfig == null) {
+      throw new IllegalArgumentException("RealmAwareZkConnectionConfig cannot be null!");
```
pkuwm: Typo in exception message. Should be RealmAwareZkClientConfig

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -457,17 +470,17 @@ public PathBasedZkSerializer getZkSerializer() {
...
   private void checkIfPathContainsShardingKey(String path) {
-    // TODO: replace with the singleton MetadataStoreRoutingData
     try {
-      String zkRealmForPath = _metadataStoreRoutingData.getMetadataStoreRealm(path);
-      if (!_zkRealmAddress.equals(zkRealmForPath)) {
-        throw new IllegalArgumentException("Given path: " + path + "'s ZK realm: " + zkRealmForPath
-            + " does not match the ZK realm: " + _zkRealmAddress + " and sharding key: "
-            + _zkRealmShardingKey + " for this DedicatedZkClient!");
+      String targetShardingKey = _metadataStoreRoutingData.getShardingKeyInPath(path);
+      if (!_zkRealmShardingKey.equals(targetShardingKey)) {
```
pkuwm: Nit, Shall we move this condition check out of try...catch block? I understand it not may be unrelated to your change. But since you already touch it and change the name :)

One more thing: This could throw NPE. `_zkRealmShardingKey` could be null if it is not set in connectionConfig when the user expects it to be a multi-realm mode by default.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
pkuwm: Like I said, personally, hiding failure type in this case is more user-friendly than floating these exceptions up. And I also understand and respect your opinion.

 I am not sure how Helix Java APIs could handle these 2 exceptions differently when constructing this FederatedZkClient. Regarding a few other users, we could get opinions from them. I will be OK if we can get options from other users/committers and they think floating exceptions are better :)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -80,19 +83,27 @@
...
-      throw new IllegalArgumentException("MetadataStoreRoutingData cannot be null!");
+  public FederatedZkClient(RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig,
+      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig)
+      throws IOException, InvalidRoutingDataException {
+    if (connectionConfig == null) {
+      throw new IllegalArgumentException("RealmAwareZkConnectionConfig cannot be null!");
     }
     if (clientConfig == null) {
-      throw new IllegalArgumentException("Client config cannot be null!");
+      throw new IllegalArgumentException("RealmAwareZkConnectionConfig cannot be null!");
```
pkuwm: Client config in message.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/constant/TestConstants.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Constants to be used for testing.
+ */
+public class TestConstants {
+  // ZK hostname prefix and port to be used throughout the zookeeper-api module
+  public static final String ZK_PREFIX = "localhost:";
+  public static final int ZK_START_PORT = 2127;
+
+  // Based on the ZK hostname constants, construct a set of fake routing data mappings
+  public static Map<String, Collection<String>> FAKE_ROUTING_DATA = ImmutableMap
```
pkuwm: static final?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
narendly: Good catch. Fixed

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
narendly: I am glad that you brought this up. In fact, DedicatedZkClient and SharedZkClient are single-realm APIs, so we need to check in the constructor that Zk Sharding key is not null or empty.

I will add this check. Then we don't need to worry about NPEs.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: Here's one example where using checked exceptions is beneficial.

1) You could imagine a scenario where in the future we might have different sources from which we could obtain routing data. There is MSDS-1, MSDS-2, and MSDS-3 with different endpoints. In this case, we could add logic for trying options 1, 2, and 3 in order in case we run into IOException. Now, this logic won't apply to InvalidRoutingDataException because that implies that a data source is available but the actual routing data is invalid. This way, we could handle different exceptions differently.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
narendly: Fixed. thank you.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/constant/TestConstants.java
narendly: Added final!

narendly:

We want all implementations of RealmAwareZkClient to do a one-time query to Metadata Store Directory Service for routing data and cache it in memory. In order to accomplish that, we have introduced HttpRoutingDataReader, which is a Singleton class that makes a REST call to read routing data and caches it in memory. This diff updates the initialization logic in RealmAwareZkClients accordingly.
Changelist:

Update all RealmAwareZkClient initialization logic
Fix tests

Issue #826, opened 2020-02-27
Add intermediate storage for customized state


zhangmeng916:
To support the storing and computation over customized state, Helix needs to implement some intermediate storage for better performance and easy operation. Specially, we would need a cache to store the existing customized state and refresh it when necessary, and we would need another data structure to hold the data in the format that is needed by computation.

PR #827, opened 2020-02-27
Add intermediate storage for customized state

zhangmeng916:
### Issues

(#826  )

### Description
To support the storing and computation over customized state, Helix needs to implement some intermediate storage for better performance and easy operation. This PR adds a cache for store customized state and refresh it when necessary. It also adds a new customized state output data structure to store the data that will be used for computation. 

### Tests
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/common/caches/CurrentStateCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/CurrentStateCache.java
@@ -19,33 +19,21 @@ package org.apache.helix.common.caches;
  * under the License.
  */
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import com.google.common.collect.Maps;
-import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.PropertyKey;
 import org.apache.helix.common.controllers.ControlContextProvider;
-import org.apache.helix.controller.LogUtil;
 import org.apache.helix.model.CurrentState;
-import org.apache.helix.model.LiveInstance;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * Cache to hold all CurrentStates of a cluster.
  */
-public class CurrentStateCache extends AbstractDataCache<CurrentState> {
+public class CurrentStateCache extends ParticipantStateCache<CurrentState> {
   private static final Logger LOG = LoggerFactory.getLogger(CurrentStateCache.class.getName());
-
-  private Map<String, Map<String, Map<String, CurrentState>>> _currentStateMap;
-  private Map<PropertyKey, CurrentState> _currentStateCache = Maps.newHashMap();
-  // If the cache is already refreshed with current state data.
+  // If the snapshot is already refreshed with current state data.
   private boolean _initialized = false;
   private CurrentStateSnapshot _snapshot;
 
@@ -55,149 +43,17 @@ public class CurrentStateCache extends AbstractDataCache<CurrentState> {
 
   public CurrentStateCache(ControlContextProvider contextProvider) {
     super(contextProvider);
-    _currentStateMap = Collections.emptyMap();
-    _snapshot = new CurrentStateSnapshot(_currentStateCache);
+    _snapshot = new CurrentStateSnapshot(_participantStateCache);
   }
 
-  /**
-   * This refreshes the CurrentStates data by re-fetching the data from zookeeper in an efficient
-   * way
-   *
-   * @param accessor
-   * @param liveInstanceMap map of all liveInstances in cluster
-   *
-   * @return
-   */
-  public boolean refresh(HelixDataAccessor accessor,
-      Map<String, LiveInstance> liveInstanceMap) {
-    long startTime = System.currentTimeMillis();
-
-    refreshCurrentStatesCache(accessor, liveInstanceMap);
-    Map<String, Map<String, Map<String, CurrentState>>> allCurStateMap = new HashMap<>();
-    for (PropertyKey key : _currentStateCache.keySet()) {
-      CurrentState currentState = _currentStateCache.get(key);
-      String[] params = key.getParams();
-      if (currentState != null && params.length >= 4) {
-        String instanceName = params[1];
-        String sessionId = params[2];
-        String stateName = params[3];
-        Map<String, Map<String, CurrentState>> instanceCurStateMap =
-            allCurStateMap.get(instanceName);
-        if (instanceCurStateMap == null) {
-          instanceCurStateMap = Maps.newHashMap();
-          allCurStateMap.put(instanceName, instanceCurStateMap);
-        }
-        Map<String, CurrentState> sessionCurStateMap = instanceCurStateMap.get(sessionId);
-        if (sessionCurStateMap == null) {
-          sessionCurStateMap = Maps.newHashMap();
-          instanceCurStateMap.put(sessionId, sessionCurStateMap);
-        }
-        sessionCurStateMap.put(stateName, currentState);
-      }
-    }
-
-    for (String instance : allCurStateMap.keySet()) {
-      allCurStateMap.put(instance, Collections.unmodifiableMap(allCurStateMap.get(instance)));
-    }
-    _currentStateMap = Collections.unmodifiableMap(allCurStateMap);
-
-    long endTime = System.currentTimeMillis();
-    LogUtil.logInfo(LOG, genEventInfo(),
-        "END: CurrentStateCache.refresh() for cluster " + _controlContextProvider.getClusterName()
-            + ", started at : " + startTime + ", took " + (endTime - startTime) + " ms");
-    if (LOG.isDebugEnabled()) {
-      LogUtil.logDebug(LOG, genEventInfo(),
-          String.format("Current State refreshed : %s", _currentStateMap.toString()));
-    }
-    return true;
-  }
-
-  // reload current states that has been changed from zk to local cache.
-  private void refreshCurrentStatesCache(HelixDataAccessor accessor,
-      Map<String, LiveInstance> liveInstanceMap) {
-
-    long start = System.currentTimeMillis();
-    PropertyKey.Builder keyBuilder = accessor.keyBuilder();
-
-    Set<PropertyKey> currentStateKeys = new HashSet<>();
-    for (String instanceName : liveInstanceMap.keySet()) {
-      LiveInstance liveInstance = liveInstanceMap.get(instanceName);
-      String sessionId = liveInstance.getEphemeralOwner();
-      List<String> currentStateNames =
-          accessor.getChildNames(keyBuilder.currentStates(instanceName, sessionId));
-      for (String currentStateName : currentStateNames) {
-        currentStateKeys.add(keyBuilder.currentState(instanceName, sessionId, currentStateName));
-      }
-    }
-    // All new entries from zk not cached locally yet should be read from ZK.
-    Set<PropertyKey> reloadKeys = new HashSet<>(currentStateKeys);
-    reloadKeys.removeAll(_currentStateCache.keySet());
-
-    Set<PropertyKey> cachedKeys = new HashSet<>(_currentStateCache.keySet());
-    cachedKeys.retainAll(currentStateKeys);
-
-    Set<PropertyKey> reloadedKeys = new HashSet<>();
-    Map<PropertyKey, CurrentState> newStateCache = Collections.unmodifiableMap(
-        refreshProperties(accessor, reloadKeys, new ArrayList<>(cachedKeys),
-            _currentStateCache, reloadedKeys));
-
-    // if the cache was not initialized, the previous state should not be included in the snapshot
+  protected void refreshSnapshot(Map<PropertyKey, CurrentState> newStateCache,
+      Map<PropertyKey, CurrentState> participantStateCache, Set<PropertyKey> reloadedKeys) {
     if (_initialized) {
-      _snapshot = new CurrentStateSnapshot(newStateCache, _currentStateCache, reloadedKeys);
+      _snapshot = new CurrentStateSnapshot(newStateCache, participantStateCache, reloadedKeys);
     } else {
       _snapshot = new CurrentStateSnapshot(newStateCache);
       _initialized = true;
     }
-
-    _currentStateCache = newStateCache;
-
-    if (LOG.isDebugEnabled()) {
-      LogUtil.logDebug(LOG, genEventInfo(),
-          "# of CurrentStates reload: " + reloadKeys.size() + ", skipped:" + (
-              currentStateKeys.size() - reloadKeys.size()) + ". took " + (System.currentTimeMillis()
-              - start) + " ms to reload new current states for cluster: " + _controlContextProvider
-              .getClusterName());
-    }
-  }
-
-  /**
-   * Return CurrentStates map for all instances.
-   *
-   * @return
-   */
-  public Map<String, Map<String, Map<String, CurrentState>>> getCurrentStatesMap() {
-    return Collections.unmodifiableMap(_currentStateMap);
-  }
-
-  /**
-   * Return all CurrentState on the given instance.
-   *
-   * @param instance
-   *
-   * @return
-   */
-  public Map<String, Map<String, CurrentState>> getCurrentStates(String instance) {
-    if (!_currentStateMap.containsKey(instance)) {
-      return Collections.emptyMap();
-    }
-    return Collections.unmodifiableMap(_currentStateMap.get(instance));
-  }
-
-  /**
-   * Provides the current state of the node for a given session id, the sessionid can be got from
-   * LiveInstance
-   *
-   * @param instance
-   * @param clientSessionId
-   *
-   * @return
-   */
-  public Map<String, CurrentState> getCurrentState(String instance, String clientSessionId) {
-    if (!_currentStateMap.containsKey(instance) || !_currentStateMap.get(instance)
-        .containsKey(clientSessionId)) {
-      return Collections.emptyMap();
-    }
-    return Collections.unmodifiableMap(_currentStateMap.get(instance).get(clientSessionId));
   }
 
   @Override
--- a/helix-core/src/main/java/org/apache/helix/common/caches/CurrentStateCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/CurrentStateCache.java
@@ -59,9 +59,11 @@ public class CurrentStateCache extends ParticipantStateCache<CurrentState> {
     for (String instanceName : liveInstanceMap.keySet()) {
       LiveInstance liveInstance = liveInstanceMap.get(instanceName);
       String sessionId = liveInstance.getEphemeralOwner();
-      List<String> currentStateNames = accessor.getChildNames(keyBuilder.currentStates(instanceName, sessionId));
+      List<String> currentStateNames =
+          accessor.getChildNames(keyBuilder.currentStates(instanceName, sessionId));
       for (String currentStateName : currentStateNames) {
-        participantStateKeys.add(keyBuilder.currentState(instanceName, sessionId, currentStateName));
+        participantStateKeys
+            .add(keyBuilder.currentState(instanceName, sessionId, currentStateName));
       }
     }
     return participantStateKeys;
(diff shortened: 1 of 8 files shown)


helix-core/src/main/java/org/apache/helix/common/caches/CustomizedStateCache.java
```diff
@@ -0,0 +1,186 @@
...
+import org.apache.helix.HelixDataAccessor;
+import org.apache.helix.PropertyKey;
+import org.apache.helix.common.controllers.ControlContextProvider;
+import org.apache.helix.controller.LogUtil;
+import org.apache.helix.model.CustomizedState;
+import org.apache.helix.model.LiveInstance;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CustomizedStateCache extends AbstractDataCache<CustomizedState> {
```
jiajunwang: As we discussed offline, we should either leverage propertyCache class or try to combine the common part of this cache and the current state cache.

helix-core/src/main/java/org/apache/helix/common/caches/AbstractDataCache.java
```diff
@@ -43,8 +47,13 @@
 
   protected ControlContextProvider _controlContextProvider;
 
+  protected Map<String, Map<String, Map<String, T>>> _participantStateMap;
+
+  protected Map<PropertyKey, T> _participantStateCache = Maps.newHashMap();
+
   public AbstractDataCache(ControlContextProvider controlContextProvider) {
```
jiajunwang: Thanks for the change. Looks great.
One suggestion is that this change will reduce the genericity of AbstractDataCache. Can we add a middle class called "AbstractParticipantStateCache" as a child class of the AbstractDataCache?

helix-core/src/main/java/org/apache/helix/common/caches/AbstractDataCache.java
```diff
@@ -104,6 +113,157 @@ public AbstractDataCache(ControlContextProvider controlContextProvider) {
...
+
+  /**
+   * This refreshes the participant state cache data by re-fetching the data from zookeeper in an efficient way
+   *
+   * @param accessor
+   * @param liveInstanceMap map of all liveInstances in cluster
+   *
+   * @return
+   */
+  public boolean refreshParticipantStatesCache(HelixDataAccessor accessor,
```
jiajunwang: You can keep the original name refresh() so it fits the style.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import org.apache.helix.model.CustomizedState;
+import org.apache.helix.model.Partition;
+
+public class CustomizedStateOutput {
+  // stateType -> (resourceName -> (Partition -> (instanceName -> customizedState)))
+  private final Map<String, Map<String, Map<Partition, Map<String, String>>>> _customizedStateMap;
```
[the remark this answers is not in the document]
zhangmeng916: CurrentStateOutput has more maps that CustomizedStateOutput. And those maps are with different types of keys. I feel it's actually easier to know what the map has and what exactly each level of keys are by looking at the map. If we have another class of hierarchical map, we will still need to pass the types for the key, and how many levels there are, and if we only manipulate map in that class, it is harder to users to understand the map logic.

helix-core/src/main/java/org/apache/helix/common/caches/CustomizedStateCache.java
```diff
@@ -0,0 +1,37 @@
...
+ * under the License.
+ */
+
+import org.apache.helix.common.controllers.ControlContextProvider;
+import org.apache.helix.model.CustomizedState;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CustomizedStateCache extends AbstractDataCache<CustomizedState> {
+  private static final Logger LOG = LoggerFactory.getLogger(CurrentStateCache.class.getName());
```
jiajunwang: nit, I think you can directly use CurrentStateCache.class

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * Given stateType, returns resource customized state map (resource -> parition -> instance ->
+   * customizedState)
+   * @param stateType
+   * @return
+   */
+  public Map<String, Map<Partition, Map<String, String>>> getResourceCustomizedStateMap(
+      String stateType) {
+    if (_customizedStateMap.containsKey(stateType)) {
+      return _customizedStateMap.get(stateType);
```
jiajunwang: duplicate and return is safer.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,135 @@
...
+   * @param stateType
+   * @param resourceName
+   * @return
+   */
+  public Map<Partition, Map<String, String>> getCustomizedStateMap(String stateType,
+      String resourceName) {
+    if (_customizedStateMap.containsKey(stateType)) {
+      Map<String, Map<Partition, Map<String, String>>> map = _customizedStateMap.get(stateType);
+      if (map.containsKey(resourceName)) {
+        return map.get(resourceName);
```
jiajunwang: Same here. Please copy the map and then return.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,135 @@
...
+    return Collections.emptyMap();
+  }
+
+  /**
+   * given (stateType, resource), returns (partition -> instance-> customizedState) map
+   * @param stateType
+   * @param resourceName
+   * @return
+   */
+  public Map<Partition, Map<String, String>> getCustomizedStateMap(String stateType,
```
jiajunwang: This method should call getResourceCustomizedStateMap() to avoid duplicate code.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /**
+   * given (stateType, resource, partition), returns (instance-> customizedState) map
+   * @param stateType
+   * @param resourceName
+   * @param partition
+   * @return
+   */
+  public Map<String, String> getCustomizedStateMap(String stateType, String resourceName,
```
jiajunwang: This method should call the other getCustomizedStateMap() to avoid duplicate code.
BTW, can we have a better name to differentiate?

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,135 @@
...
+        if (partitionMap != null) {
+          return partitionMap.get(partition);
+        }
+      }
+    }
+    return Collections.emptyMap();
+  }
+
+  public Set<String> getAllStateTypes() {
+    return _customizedStateMap.keySet();
```
jiajunwang: Always copy and then return.

helix-core/src/main/java/org/apache/helix/common/caches/CustomizedStateCache.java
zhangmeng916: Let's keep it separated, so that in the computation logic, there's no confusion.

helix-core/src/main/java/org/apache/helix/common/caches/CustomizedStateCache.java
```diff
@@ -0,0 +1,186 @@
...
+
+  // reload customized states that has been changed from zk to local cache.
+  private void refreshCustomizedStatesCache(HelixDataAccessor accessor,
+      Map<String, LiveInstance> liveInstanceMap, Set<String> aggregationEnabledTypes) {
+    long start = System.currentTimeMillis();
+    PropertyKey.Builder keyBuilder = accessor.keyBuilder();
+    Set<PropertyKey> customizedStateKeys = new HashSet<>();
+    List<String> resourceNames = new ArrayList<>();
+    for (String instanceName : liveInstanceMap.keySet()) {
+      for (String customizedStateName : aggregationEnabledTypes) {
```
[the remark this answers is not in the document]
zhangmeng916: It's moved to participantStateCache logic and updated.

helix-core/src/main/java/org/apache/helix/common/caches/ParticipantStateCache.java
```diff
@@ -0,0 +1,203 @@
...
+  public boolean refresh(HelixDataAccessor accessor,
+      Map<String, LiveInstance> liveInstanceMap, Boolean snapshotEnabled, Set<String> restrictedKeys) {
+    long startTime = System.currentTimeMillis();
+
+    refreshParticipantStatesCacheFromZk(accessor, liveInstanceMap, snapshotEnabled, restrictedKeys);
+    Map<String, Map<String, Map<String, T>>> allParticipantStateMap = new HashMap<>();
+    for (PropertyKey key : _participantStateCache.keySet()) {
+      T participantState = _participantStateCache.get(key);
+      String[] params = key.getParams();
+      if (participantState != null && params.length >= 4) {
```
jiajunwang: log error (or warning) if any mismatch keys.

helix-core/src/main/java/org/apache/helix/common/caches/ParticipantStateCache.java
```diff
@@ -0,0 +1,203 @@
...
+        if (stateMapLevel1 == null) {
+          stateMapLevel1 = Maps.newHashMap();
+          allParticipantStateMap.put(keyLevel1, stateMapLevel1);
+        }
+        Map<String, T> stateMapLevel2 = stateMapLevel1.get(keyLevel2);
+        if (stateMapLevel2 == null) {
+          stateMapLevel2 = Maps.newHashMap();
+          stateMapLevel1.put(keyLevel2, stateMapLevel2);
+        }
+        stateMapLevel2.put(keyLevel3, participantState);
```
jiajunwang: nit, can we simplify this block a little bit. For example, putIfAbsent, etc.

helix-core/src/main/java/org/apache/helix/common/caches/ParticipantStateCache.java
```diff
@@ -0,0 +1,203 @@
...
+      }
+    }
+
+    for (String instance : allParticipantStateMap.keySet()) {
```
jiajunwang: Since you are creating a new map every time already, is it still necessary to make it unmodifiable?

helix-core/src/main/java/org/apache/helix/common/caches/ParticipantStateCache.java
```diff
@@ -0,0 +1,203 @@
...
+      Set<String> restrictedKeys) {
+
+    long start = System.currentTimeMillis();
+    PropertyKey.Builder keyBuilder = accessor.keyBuilder();
+
+    Set<PropertyKey> participantStateKeys = new HashSet<>();
+    for (String instanceName : liveInstanceMap.keySet()) {
+      LiveInstance liveInstance = liveInstanceMap.get(instanceName);
+
+      if (restrictedKeys.isEmpty()) {
```
jiajunwang: So for the first condition, the liveInstance is kind of the restrictedKeys, right? Can we put this logic to a protected method and leave it to the child class? This will help to cleanup the abstract class.

jiajunwang: Basically you have an abstract method getParticipantStateKeys and called here. The children will implement this method separately with the different logic.

helix-core/src/main/java/org/apache/helix/common/caches/ParticipantStateCache.java
```diff
@@ -0,0 +1,203 @@
...
+    return Collections.unmodifiableMap(_participantStateMap.get(keyLevel1));
+  }
+
+  /**
+   * Provides the participant state map for a certain level1 key and level2 key
+   * @param keyLevel1
+   * @param keyLevel2
+   * @return
+   */
+  public Map<String, T> getParticipantState(String keyLevel1,
```
jiajunwang: Optional, both getParticipantStates() methods could be protected and let the children class wrap them and implement with more meaningful public methods.

zhangmeng916: I changed the key names here. I feel it's ok to leave it here as it only returns some mapping.

helix-core/src/main/java/org/apache/helix/common/caches/ParticipantStateCache.java
zhangmeng916: I tried a bit, but it cannot really be simplified by using putIfAbsent, as the next level map still need to use the previous map.

helix-core/src/main/java/org/apache/helix/common/caches/CustomizedStateCache.java
```diff
@@ -0,0 +1,37 @@
...
+public class CustomizedStateCache extends ParticipantStateCache<CustomizedState> {
+  private static final Logger LOG = LoggerFactory.getLogger(CurrentStateCache.class.getName());
+
+  public CustomizedStateCache(String clusterName) {
+    this(createDefaultControlContextProvider(clusterName));
+  }
+
+  public CustomizedStateCache(ControlContextProvider contextProvider) {
+    super(contextProvider);
+  }
```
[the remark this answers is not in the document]
zhangmeng916: Yeah, actually they're both used.

helix-core/src/main/java/org/apache/helix/common/caches/ParticipantStateCache.java
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * This refreshes the participant state cache data by re-fetching the data from zookeeper in an
+   * efficient way
+   * @param accessor
+   * @param liveInstanceMap map of all liveInstances in cluster
+   * @return
+   */
+  public boolean refresh(HelixDataAccessor accessor, Map<String, LiveInstance> liveInstanceMap,
+      Boolean snapshotEnabled, Set<String> restrictedKeys) {
```
jiajunwang: As we talked offline, let's try to hide restrictedKeys as an internal field and don't require the caller to pass on every refresh call.

helix-core/src/main/java/org/apache/helix/common/caches/CurrentStateCache.java
```diff
@@ -55,149 +46,33 @@ public CurrentStateCache(String clusterName) {
 
   public CurrentStateCache(ControlContextProvider contextProvider) {
     super(contextProvider);
-    _currentStateMap = Collections.emptyMap();
-    _snapshot = new CurrentStateSnapshot(_currentStateCache);
+    _snapshot = new CurrentStateSnapshot(_participantStateCache);
```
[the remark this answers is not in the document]
zhangmeng916: I changed the map in ParticipantCache.

helix-core/src/main/java/org/apache/helix/common/caches/ParticipantStateCache.java
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * This refreshes the participant state cache data by re-fetching the data from zookeeper in an
+   * efficient way
+   * @param accessor
+   * @param liveInstanceMap map of all liveInstances in cluster
+   * @return
+   */
+  public boolean refresh(HelixDataAccessor accessor, Map<String, LiveInstance> liveInstanceMap,
+      Boolean snapshotEnabled) {
```
jiajunwang: Should this boolean be fixed once the class is coded? Overall, I don't think we need it. Just call refreshSnapshot() anyway. If there is no such implementation, the child class won't implement it anyway.

helix-core/src/main/java/org/apache/helix/common/caches/ParticipantStateCache.java
```diff
@@ -0,0 +1,177 @@
...
+              - reloadKeys.size()) + ". took " + (System.currentTimeMillis() - start)
+              + " ms to reload new participant states for cluster: " + _controlContextProvider.getClusterName()
+              + "and state: " + this.getClass().getName());
+    }
+  }
+
+  protected abstract Set<PropertyKey> PopulateParticipantKeys(HelixDataAccessor accessor,
+      Map<String, LiveInstance> liveInstanceMap);
+
+  protected abstract void refreshSnapshot(Map<PropertyKey, T> newStateCache, Map<PropertyKey, T> participantStateCache,
```
jiajunwang: Since the customized state cache won't support snapshot for now, just comment on this method as OPTIONAL to be implemented. If the child class does not support snapshot, just do nothing. And we don't need the boolean to differentiate.

We might want to support it later though.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,126 @@
...
+    _customizedStateMap.get(stateType).get(resourceName).get(partition).put(instanceName, state);
+  }
+
+  /**
+   * Given stateType, returns resource customized state map (resource -> parition -> instance ->
+   * customizedState)
+   * @param stateType
+   * @return
+   */
+  public Map<String, Map<Partition, Map<String, String>>> getResourceCustomizedStateMap(
```
jiajunwang: Intuitively, this one should be called getCustomizedStateMap()

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,126 @@
...
+    return Collections.emptyMap();
+  }
+
+  /**
+   * given (stateType, resource), returns (partition -> instance-> customizedState) map
+   * @param stateType
+   * @param resourceName
+   * @return
+   */
+  public Map<Partition, Map<String, String>> getPartitionCustomizedStateMap(String stateType,
```
jiajunwang: Intuitively, this one should be called getResourceCustomizedStateMap()

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  /**
+   * given (stateType, resource, partition), returns (instance-> customizedState) map
+   * @param stateType
+   * @param resourceName
+   * @param partition
+   * @return
+   */
+  public Map<String, String> getInstanceCustomizedStateMap(String stateType, String resourceName,
```
jiajunwang: Intuitively, this one should be called getPartitionCustomizedStateMap()

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * given (stateType, resource, partition, instance), returns customized state
+   * @param stateType
+   * @param resourceName
+   * @param partition
+   * @param instanceName
+   * @return
+   */
+  public String getCustomizedState(String stateType, String resourceName, Partition partition,
```
jiajunwang: This one can be called getReplicaCustomizedState()

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Map;
+import java.util.Set;
+import org.apache.helix.model.CustomizedState;
+import org.apache.helix.model.Partition;
+
+public class CustomizedStateOutput {
+  // stateType -> (resourceName -> (Partition -> (instanceName -> customizedState)))
+  private final Map<String, Map<String, Map<Partition, Map<String, String>>>> _customizedStateMap;
+
+  private final Map<String, CustomizedState> _customizedStateMetaMap;
```
jiajunwang: How this is used? I would suggest adding this only when we have the usage in the PR.

zhangmeng916:
Final commit message:
1. Implement a participant state cache for generalizing functions in both current state cache and customized state cache. 
2. Implement an intermediate data structure to store the result of customized state computation and prepare the data to be the format that can be used by customized view computation later.

PR #829, opened 2020-02-27
Fix bugs in ZkRoutingDataReader and ZkRoutingDataWriter

### LINKED ISSUE #828: Fix bugs in ZkRoutingDataReader and ZkRoutingDataWriter
Fix the following bugs:
1. When the routing data is not initialized for a namespace, creating a `ZkRoutingDataReader` for that namespace causes a NoNodeException due to a `getChildren` call. Fix: initialize the base node for routing data across all zookeeper-related actors (`ZkMetadataStoreDirectory`, `ZkRoutingDataReader`, `ZkRoutingDataWriter`)
2. In the `getRoutingData()` method of `ZkRoutingDataReader`, a null check is necessary before getting sharding keys from a realm node, because the realm node may be an invalid ZNRecord. Fix: add null check. 
3. The hostname provided from system properties does not have the protocol. `ZkRoutingDataWriter` was using the hostname directly for request forwarding. Fix: add protocol to the hostname before request forwarding. 
4. `ZkRoutingDataReader`'s `handleDataDeleted` is missing an unsubscribe operation which may lead to watcher leak; `handleChildChange` is missing the resubscription operation. Fix: add both.

NealSun96:
### Issues

Fixes #828, #832 

### Description

This PR fixes the following bugs:
1. When the routing data is not initialized for a namespace, creating a `ZkRoutingDataReader` for that namespace causes a NoNodeException due to a `getChildren` call. Fix: initialize the base node for routing data across all zookeeper-related actors (`ZkMetadataStoreDirectory`, `ZkRoutingDataReader`, `ZkRoutingDataWriter`)
2. In the `getRoutingData()` method of `ZkRoutingDataReader`, a null check is necessary before getting sharding keys from a realm node, because the realm node may be an invalid ZNRecord. Fix: add null check. 
3. The hostname provided from system properties does not have the protocol. `ZkRoutingDataWriter` was using the hostname directly for request forwarding. Fix: add protocol to the hostname before request forwarding. 
4. `ZkRoutingDataReader`'s `handleDataDeleted` is missing an unsubscribe operation which may lead to watcher leak; `handleChildChange` is missing the resubscription operation. Fix: add both.

This PR also introduced an improvement to `MetadataStoreDirectoryAccessor` endpoints. Previously, the endpoints in `MetadataStoreDirectoryAccessor` are missing proper status codes for some cases: for example for `IllegalArgumentException`, it's most fitting to return 400 Bad Request to users, but currently the exception will not be handled and cause a 500 to be returned. 
A change that can fix this issue while improving the status code behavior is to bind exception types with status codes. `NoSuchElementException` should be paired with 404, `IllegalArgumentException` should be paired with 400, while `IllegalStateException` (currently used when underlying ZkClients are closed) should be paired with 500. This has two major benefits:
1. Code maintenance is made easier. The exception types clearly and meaningfully translate to status codes. Developers no longer need to traverse layer by layer to figure out what happened to each exception; all of them are handled at the accessor layer.
2. The status codes are verbose and helpful, because they now contain the underlying exception messages that explain the exact reason behind the exceptions. 

### Tests

[INFO] Tests run: 136, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 22.201 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 136, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  52.954 s
[INFO] Finished at: 2020-02-27T12:06:15-08:00

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/common/HttpConstants.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/common/HttpConstants.java
@@ -27,5 +27,5 @@ public class HttpConstants {
     DELETE
   }
 
-  public static final String HTTP_PROTOCOL = "http://";
+  public static final String HTTP_PROTOCOL_PREFIX = "http://";
 }
--- a/helix-rest/src/main/java/org/apache/helix/rest/common/HttpConstants.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/common/HttpConstants.java
@@ -26,4 +26,6 @@ public class HttpConstants {
     PUT,
     DELETE
   }
+
+  public static final String HTTP_PROTOCOL = "http://";
 }
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
@@ -96,11 +96,10 @@ public class ZkRoutingDataReader implements MetadataStoreRoutingDataReader, IZkD
           "Routing data directory ZNode " + MetadataStoreRoutingConstants.ROUTING_DATA_PATH
               + " does not exist. Routing ZooKeeper address: " + _zkAddress);
     }
-    if (allRealmAddresses != null)
-    {
+    if (allRealmAddresses != null) {
       for (String realmAddress : allRealmAddresses) {
-        ZNRecord record =
-            _zkClient.readData(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realmAddress);
+        ZNRecord record = _zkClient
+            .readData(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realmAddress);
         if (record != null) {
           List<String> shardingKeys =
               record.getListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY);
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
@@ -35,6 +35,7 @@ import org.apache.helix.zookeeper.zkclient.IZkChildListener;
 import org.apache.helix.zookeeper.zkclient.IZkDataListener;
 import org.apache.helix.zookeeper.zkclient.IZkStateListener;
 import org.apache.helix.zookeeper.zkclient.exception.ZkNoNodeException;
+import org.apache.helix.zookeeper.zkclient.exception.ZkNodeExistsException;
 import org.apache.zookeeper.Watcher;
 
 
@@ -57,6 +58,15 @@ public class ZkRoutingDataReader implements MetadataStoreRoutingDataReader, IZkD
     _zkClient = DedicatedZkClientFactory.getInstance()
         .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
             new HelixZkClient.ZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
+
+    // Ensure that ROUTING_DATA_PATH exists in ZK. If not, create
+    // create() semantic will fail if it already exists
+    try {
+      _zkClient.createPersistent(MetadataStoreRoutingConstants.ROUTING_DATA_PATH, true);
+    } catch (ZkNodeExistsException e) {
+      // This is okay
+    }
+
     _routingDataListener = routingDataListener;
     if (_routingDataListener != null) {
       // Subscribe child changes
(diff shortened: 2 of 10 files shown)


helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
```diff
@@ -134,6 +147,7 @@ public synchronized void handleChildChange(String s, List<String> list) {
 
     // Subscribe data changes again because some children might have been deleted or added
     _zkClient.unsubscribeAll();
+    _zkClient.subscribeChildChanges(MetadataStoreRoutingConstants.ROUTING_DATA_PATH, this);
```
NealSun96: After making the fix, `handleChildChange` and `handleDataDeleted` are completely equal now. I added a comment on line 127 based on what you told me to justify the duplication - sometimes the callbacks may go missing. Should we consider calling `handleDataDeleted` from `handleChildChange`, for example, to reduce duplication? @narendly

NealSun96: Resolved offline.

helix-rest/src/main/java/org/apache/helix/rest/common/HttpConstants.java
```diff
@@ -26,4 +26,6 @@
     PUT,
     DELETE
   }
+
+  public static final String HTTP_PROTOCOL = "http://";
```
narendly: Nit: How about something like HTTP_PREFIX?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
```diff
@@ -90,10 +99,12 @@ public ZkRoutingDataReader(String namespace, String zkAddress,
     for (String realmAddress : allRealmAddresses) {
```
narendly: Null check on allRealmAddress just in case?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
```diff
@@ -128,17 +142,7 @@ public synchronized void handleDataDeleted(String s) {
...
-    }
-
-    // Subscribe data changes again because some children might have been deleted or added
-    _zkClient.unsubscribeAll();
-    for (String child : _zkClient.getChildren(MetadataStoreRoutingConstants.ROUTING_DATA_PATH)) {
-      _zkClient.subscribeDataChanges(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + child,
-          this);
-    }
-    _routingDataListener.refreshRoutingData(_namespace);
+    handleDataDeleted(s);
```
narendly: Please refactor the logic out into a private method and call it `handleResubscription()` and call it in both of these methods.

helix-rest/src/main/java/org/apache/helix/rest/common/HttpConstants.java
NealSun96: Good point, I'll opt for `HTTP_PROTOCAL_PREFIX`. It is a protocol and it is also a prefix.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
NealSun96: I'll add it, but is that case actually possible?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
NealSun96: Ack.

NealSun96:
Final commit message:
## Improve MetadataStoreDirectoryAccessor endpoints and fix bugs in ZkRoutingDataReader/Writer ##
Improve the current status code design of MetadataStoreDirectoryAccessor endpoints by more clearly translate underlying exceptions to status codes. Fix 4 bugs in Reader/Writer.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
```diff
@@ -87,13 +96,18 @@ public ZkRoutingDataReader(String namespace, String zkAddress,
...
+      for (String realmAddress : allRealmAddresses) {
+        ZNRecord record =
+            _zkClient.readData(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + realmAddress);
+        if (record != null) {
+          List<String> shardingKeys =
+              record.getListField(MetadataStoreRoutingConstants.ZNRECORD_LIST_FIELD_KEY);
+          routingData
+              .put(realmAddress, shardingKeys != null ? shardingKeys : Collections.emptyList());
+        }
+      }
```
narendly: format

PR #831, opened 2020-02-27
Generate cancellation message for currentState=null desiredState=DROPPED

### LINKED ISSUE #830: Generate state transition cancellation message when currentState=null desiredState=DROPPED
In current version of MessageGenerationPhase, we skip message generation when currentState=null and desiredState=DROPPED to avoid infinite OFFLINE-> DROPPED messages. However, this also cause Helix fail to send cancellation message. This issue tracks the fix for this problem.

mgao0:
### Issues

Fixes #830 

### Description

To fix the issue that Helix fail to send cancel message when currentState=null and desiredState=DROPPED, this PR adds message generation logic to the condition before the "continue". To reduce the duplication of code, separate the related code section out and put into individual methods. 
Condition: current state = null, desired state = dropped, there is pending message. Current state is checked, and since it is null, it is set to the initial state OFFLINE for further handling. 
Before fix: all message generation logics are skipped for this instance, this handling is to fix a previous problem where under this condition, infinite "OFFLINE to DROPPED" state transition messages. And we are having the problem that because we skip the message generation logics, the cancellation message for the pending message won't be generated.
After fix: this fix separated the generation of cancellation message for pending message from other message generation logics to be a single method, and call this method when the above edge case is met.

### Tests

TestSendCancellationMessageFromOFFLINEToDROPPED

Results :

Tests run: 1084, Failures: 0, Errors: 0, Skipped: 0

[INFO] BUILD SUCCESS
[INFO] Total time:  01:09 h
[INFO] Finished at: 2020-03-02T11:16:06-08:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/controller/stages/TestSendCancellationMessageFromOFFLINEToDROPPED.java
+++ b/helix-core/src/test/java/org/apache/helix/controller/stages/TestSendCancellationMessageFromOFFLINEToDROPPED.java
@@ -1,5 +1,6 @@
 package org.apache.helix.controller.stages;
 
+import java.lang.reflect.Array;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


helix-core/src/test/java/org/apache/helix/controller/stages/TestSendCancellationMessageFromOFFLINEToDROPPED.java
```diff
@@ -0,0 +1,97 @@
...
+import org.testng.annotations.Test;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+
+/**
+ * This test checks the cancellation message generation when currentState=null and desiredState=DROPPED
+ */
+public class TestSendCancellationMessageFromOFFLINEToDROPPED extends MessageGenerationPhase {
```
jiajunwang: Can we just add this test method to some existing test classes? And the name is too long.

mgao0: I'll change it to a shorter name. It has to be a separate class because it needs to extend MessageGenerationPhase so I can run this class alone. Since it is a very specific case, it's better we can run this stage alone so we can control the input to this message generation phase.

jiajunwang: Sure, sounds good.

mgao0:
> I haven't finish reading details yet. Overall, when we try to move code to a separate method, try to simplify the logic and parameters as much as possible. What I observed here are very long parameter lists. Can we try to simplify it?
> 
> Moreover, could you provide more details about the old behavior and the issue. Then compare it with the new one. Thanks.

Good point. Please see the updated PR description, hope it explains the issue. Please let me know how I can improve it.
Per your concerns about long parameter list, I have tried to simplify it and I think this is the best I can do right now. All the parameters are needed due to the complex logic, and the purpose of moving into a separate method is to avoid duplicated code because they will be called under two different conditions.

mgao0:

dasahcc:
Please fix your author and email for commit message.

PR #838, opened 2020-02-29
Make ConfigAccessor and ZkUtil realm-aware

### LINKED ISSUE #839: Make ConfigAccessor and ZkUtil realm-aware
To make Helix Java APIs realm-aware, we first make ConfigAccessor and ZkUtil realm-aware by instrumenting these APIs with a Builder and RealmAwareZkClients.

narendly:
### Issues

Resolves #839 

### Description

To make Helix Java APIs realm-aware, we first make ConfigAccessor and ZkUtil realm-aware by instrumenting these APIs with a Builder and RealmAwareZkClients.

The Builder pattern is chosen because it is a scalable option when there are a lot of configurable parameters. It makes it easy to validate the given parameters as well.

### Tests

ConfigAccessor is already covered by TestConfigAccessor.

[INFO]
[INFO] Results:
[INFO]
[ERROR] Failures:
[INFO]
[ERROR] Tests run: 1083, Failures: 1, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:08 h
[INFO] Finished at: 2020-02-29T20:46:35-08:00

TestCompression passes when run individually

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
@@ -38,11 +38,14 @@ import org.apache.helix.model.RESTConfig;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.builder.HelixConfigScopeBuilder;
 import org.apache.helix.util.StringTemplate;
-import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.api.client.HelixZkClient;
+import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.apache.helix.zookeeper.impl.client.FederatedZkClient;
 import org.apache.helix.zookeeper.impl.factory.SharedZkClientFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 
 /**
  * Provides access to the persistent configuration of the cluster, the instances that live on it,
@@ -52,6 +55,7 @@ public class ConfigAccessor {
   private static Logger LOG = LoggerFactory.getLogger(ConfigAccessor.class);
 
   private static final StringTemplate template = new StringTemplate();
+
   static {
     // @formatter:off
     template.addEntry(ConfigScopeProperty.CLUSTER, 1, "/{clusterName}/CONFIGS/CLUSTER");
@@ -70,10 +74,45 @@ public class ConfigAccessor {
     // @formatter:on
   }
 
-  private final HelixZkClient _zkClient;
+  private final RealmAwareZkClient _zkClient;
   // true if ConfigAccessor was instantiated with a HelixZkClient, false otherwise
   // This is used for close() to determine how ConfigAccessor should close the underlying ZkClient
   private final boolean _usesExternalZkClient;
+
+  /**
+   * Constructor that creates a realm-aware ConfigAccessor using ConficAccessorBuilder.
+   * @param builder
+   */
+  public ConfigAccessor(ConfigAccessorBuilder builder) {
+    // Default mode is MULTI_REALM
+    switch ((builder._realmMode != null) ? builder._realmMode
+        : ConfigAccessorBuilder.RealmMode.MULTI_REALM) {
+      case MULTI_REALM:
+        // TODO: make sure FederatedZkClient has been created correctly
+        _zkClient = new FederatedZkClient();
+      case SINGLE_REALM:
+        // Create a HelixZkClient. Use a SharedZkClient because ConfigAccessor does not need to do ephemeral operations
+        if (builder._realmAwareZkClientConfig == null) {
+          // Use a default HelixZkClient.ZkClientConfig()
+          _zkClient = SharedZkClientFactory.getInstance()
+              .buildZkClient(new HelixZkClient.ZkConnectionConfig(builder._zkAddress),
+                  new HelixZkClient.ZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
+        } else {
+          // Use the given RealmAwareZkClientConfig
+          _zkClient = SharedZkClientFactory.getInstance()
+              .buildZkClient(new HelixZkClient.ZkConnectionConfig(builder._zkAddress),
+                  new HelixZkClient.ZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
+        }
+
+
+
+      default:
+        throw new HelixException("Invalid RealmMode given: " + builder._realmMode);
+    }
+
+    _zkClient = null;
+    _usesExternalZkClient = false;
+  }
 
   /**
    * Initialize an accessor with a Zookeeper client
@@ -94,9 +133,9 @@ public class ConfigAccessor {
    * @param zkAddress
    */
   public ConfigAccessor(String zkAddress) {
-    _zkClient = SharedZkClientFactory.getInstance().buildZkClient(
-        new HelixZkClient.ZkConnectionConfig(zkAddress),
-        new HelixZkClient.ZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
+    _zkClient = SharedZkClientFactory.getInstance()
+        .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
+            new HelixZkClient.ZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
     _usesExternalZkClient = false;
   }
 
@@ -157,7 +196,6 @@ public class ConfigAccessor {
       }
     }
     return map;
-
   }
 
   /**
@@ -265,8 +303,8 @@ public class ConfigAccessor {
       String instanceName = scopeStr.substring(scopeStr.lastIndexOf('/') + 1);
       if (!ZKUtil.isInstanceSetup(_zkClient, scope.getClusterName(), instanceName,
           InstanceType.PARTICIPANT)) {
-        throw new HelixException("instance: " + instanceName + " is NOT setup in cluster: "
-            + clusterName);
+        throw new HelixException(
+            "instance: " + instanceName + " is NOT setup in cluster: " + clusterName);
       }
     }
 
@@ -499,7 +537,6 @@ public class ConfigAccessor {
 
         if (splits[1].startsWith("SIMPLEKEYS")) {
           retKeys = new ArrayList<String>(record.getSimpleFields().keySet());
-
         } else if (splits[1].startsWith("MAPKEYS")) {
           retKeys = new ArrayList<String>(record.getMapFields().keySet());
         } else if (splits[1].startsWith("MAPMAPKEYS")) {
@@ -516,7 +553,6 @@ public class ConfigAccessor {
     } catch (Exception e) {
       return Collections.emptyList();
     }
-
   }
 
   /**
@@ -638,8 +674,8 @@ public class ConfigAccessor {
     updateClusterConfig(clusterName, clusterConfig, false);
   }
 
-
-  private void updateClusterConfig(String clusterName, ClusterConfig clusterConfig, boolean overwrite) {
+  private void updateClusterConfig(String clusterName, ClusterConfig clusterConfig,
+      boolean overwrite) {
     if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
       throw new HelixException("fail to update config. cluster: " + clusterName + " is NOT setup.");
     }
@@ -775,7 +811,6 @@ public class ConfigAccessor {
   public void setInstanceConfig(String clusterName, String instanceName,
       InstanceConfig instanceConfig) {
     updateInstanceConfig(clusterName, instanceName, instanceConfig, true);
-
   }
 
   /**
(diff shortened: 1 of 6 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
```diff
@@ -373,6 +379,91 @@ public String getZkRealmShardingKey() {
...
+        return new RealmAwareZkConnectionConfig(this);
+      }
+
+      /**
+       * Validate the internal fields of the builder before creating an instance.
+       */
+      private void validate() {
+        boolean isRealmModeSet = _realmMode != null;
+        boolean isShardingKeySet = _zkRealmShardingKey != null && !_zkRealmShardingKey.isEmpty();
+        switch (isRealmModeSet ? _realmMode : RealmMode.MULTI_REALM) {
```
pkuwm: It seems by default you set realmMode as multi realm. Why not set it multi realm by default in constructor ?
And if sharding key is set only, are we going to just allow it to have single realm by default?
 It seems easier to use and to check the code logic?

By default:
1. No sharding key or realm mode set, by default it is multi realm
2. Only sharding key is set, single realm
3. Sharding key and multi realm set, illegal

narendly: That's what this piece of logic is doing.

Moving validation logic to a validate() function is a lot cleaner and promotes better organization. The constructor only constructs.

I am not understanding your question/comment clearly - could you clarify?

pkuwm: This validation logic needs some time to actually understand. At a first glance, I was not able to understand the logic.

- if only sharding key is set and realm mode is not set, I would expect to create a single mode client, but this validation does not pass. Or is realm mode forced to set?

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -830,4 +857,75 @@ public void close() {
...
+      if (isRealmModeSet) {
+        if (_realmMode == RealmAwareZkClient.RealmMode.SINGLE_REALM && !isZkAddressSet) {
+          throw new HelixException(
+              "ConfigAccessor: RealmMode cannot be single-realm without a valid ZkAddress set!");
+        }
+        // If realm mode is set to multi-realm, we simply ignore the given ZK address
+      } else {
+        // If zkAddress is set, that implies single-realm mode. Otherwise, multi-realm mode.
+        _realmMode = isZkAddressSet ? RealmAwareZkClient.RealmMode.SINGLE_REALM
+            : RealmAwareZkClient.RealmMode.MULTI_REALM;
```
pkuwm: It seems this could be simplified a bit to make it easier to read?
```
if (_realmMode == RealmAwareZkClient.RealmMode.SINGLE_REALM && !isZkAddressSet) {
 throw new HelixException(
     "ConfigAccessor: RealmMode cannot be single-realm without a valid ZkAddress set!");
}

if (_realmMode == null) {
 _realmMode = isZkAddressSet ? RealmAwareZkClient.RealmMode.SINGLE_REALM
             : RealmAwareZkClient.RealmMode.MULTI_REALM;
}
```

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
```diff
@@ -373,6 +379,91 @@ public String getZkRealmShardingKey() {
...
+            }
+            break;
+          case SINGLE_REALM:
+            if (!isShardingKeySet) {
+              throw new IllegalArgumentException(
+                  "ZK sharding key must be set on single-realm mode!");
+            }
+            break;
+          default:
+            throw new ZkClientException("RealmAwareZkConnectionConfig: Unknown mode!");
```
pkuwm: Do we want `ZkClientException`?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
```diff
@@ -75,7 +80,7 @@ public static boolean isClusterSetup(String clusterName, HelixZkClient zkClient)
       logger.info("Fail to check cluster setup : zookeeper client is null!");
       return false;
     }
-    ArrayList<String> requiredPaths = new ArrayList<String>();
+    ArrayList<String> requiredPaths = new ArrayList<>();
```
pkuwm: Nit, shall we declare type `List<>` ?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
```diff
@@ -92,15 +97,17 @@ public static boolean isClusterSetup(String clusterName, HelixZkClient zkClient)
     requiredPaths.add(PropertyPathBuilder.controllerHistory(clusterName));
     boolean isValid = true;
 
-    BaseDataAccessor<Object> baseAccessor = new ZkBaseDataAccessor<Object>(zkClient);
-    boolean[] ret = baseAccessor.exists(requiredPaths, 0);
+    boolean[] ret = new boolean[requiredPaths.size()];
+    for (int i = 0; i < requiredPaths.size(); i++) {
+      ret[i] = zkClient.exists(requiredPaths.get(i));
```
pkuwm: It seems this changes the original behavior. `baseAccessor.exits` is actually an async call using `asyncExists()` in the bottom implementation. I think it is better to keep the original behavior.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
```diff
@@ -92,15 +97,17 @@ public static boolean isClusterSetup(String clusterName, HelixZkClient zkClient)
...
+    }
     StringBuilder errorMsg = new StringBuilder();
 
     for (int i = 0; i < ret.length; i++) {
       if (!ret[i]) {
         isValid = false;
-        errorMsg
-            .append(("Invalid cluster setup, missing znode path: " + requiredPaths.get(i)) + "\n");
+        errorMsg.append("Invalid cluster setup, missing znode path: ").append(requiredPaths.get(i))
```
pkuwm: Nit: Maybe we can modify a bit for the error message to combine the error paths together like below to avoid the extra duplicate strings `Invalid cluster setup, missing znode path` for each path.
```
Invalid cluster setup, missing znode paths: [path1, path2]
```

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
```diff
@@ -66,6 +67,10 @@ public static boolean isClusterSetup(String clusterName, String zkAddress) {
   }
 
   public static boolean isClusterSetup(String clusterName, HelixZkClient zkClient) {
```
pkuwm: If we have `public static boolean isClusterSetup(String clusterName, RealmAwareZkClient zkClient)`, it seems keeping this old API `public static boolean isClusterSetup(String clusterName, HelixZkClient zkClient)` is redundant, because `HelixZkClient` is a subclass of `RealmAwareZkClient`?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
```diff
@@ -132,7 +139,12 @@ public static boolean isInstanceSetup(String zkAddress, String clusterName, Stri
     return result;
   }
 
-  public static boolean isInstanceSetup(HelixZkClient zkclient, String clusterName,
+  public static boolean isInstanceSetup(HelixZkClient zkClient, String clusterName,
```
pkuwm: I think we can move this old API as HelixZkClient can automatically fit into RealmAwareZkClient.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
narendly: It's fine - the idea is not changing the existing implementation. But this is a quick fix. I'll change to List.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
narendly: This is a util function. Functionality wise, there is no difference.

Doing an asyncExists() would require more change (you have to look at the callbacks).

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
narendly: Feel free to do this in a future PR. This PR is not about improving or modifying the logic of ZkUtil. It simply is making things realm-aware.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
narendly: I agree. Let me remove old signatures.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
pkuwm: I did that is why I know this is an async. If you do this sync, it changes the original behavior. If one call of exists fails and throw exception this call fails. But the original behavior won’t fail all the calls. Some other calls are still possible to succeed. Like you said this is just a realm aware change, I don’t think it is a good idea to change the behavior.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
narendly: Simplified :)

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
narendly: Yes. Will remove.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
narendly: Okay. Wrapping it with a try-catch.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
narendly: I simplified the validation logic. Let me know if you think it's clearer.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
```diff
@@ -325,16 +328,18 @@ public int hashCode() {
...
-    private int _sessionTimeout = DEFAULT_SESSION_TIMEOUT;
-
-    public RealmAwareZkConnectionConfig(String zkRealmShardingKey) {
-      _zkRealmShardingKey = zkRealmShardingKey;
+    private String _zkRealmShardingKey;
+    private String _msdsEndpoint;
+    private int _sessionTimeout;
+
+    private RealmAwareZkConnectionConfig(Builder builder) {
+      _zkRealmShardingKey = builder._zkRealmShardingKey;
```
pkuwm: This “._” is a bit annoying and we may not have setters in builder...
This is one of the reasons I don’t like this leading underscore naming convention. May be in this case I would just ignore this convention and instead use a convention without leading underscore.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
```diff
@@ -373,6 +378,81 @@ public String getZkRealmShardingKey() {
...
+
+      public RealmAwareZkConnectionConfig build() {
+        validate();
+        return new RealmAwareZkConnectionConfig(this);
+      }
+
+      /**
+       * Validate the internal fields of the builder before creating an instance.
+       */
+      private void validate() {
```
pkuwm: It seems if realm mode is not set by caller, we don’t set it here by default. Shall we set a default value based on sharding key being set or not?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
pkuwm: I am still not very convinced we change the behavior. I understand the functionality may be the same. But if one call is blocked because of connection and it has to retry, other calls could be blocked. I am not sure how critical it is for here. Maybe Junkai can give an opinion. Since we are also integrating baseDataAccessor, can we just keep this original behavior by calling base data accessor?

narendly: It is not that critical. This could have easily used a set of sync exists call too.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
narendly: This is just a config, not an actual API. So we just do checks. For APIs, we might set the internal variables accordingly to make the construction logic clean and simple.

Issue #843, opened 2020-03-02
Fix writing race conditions in ZkMetadataStoreDirectory


NealSun96:
Among the writing operations of `ZkMetadataStoreDirectory`, we have validation logic set up to ensure that, for example, a sharding key insertion is valid. However, it is possible to bypass the validation logic: one simple reproducible method is to do `addShardingKey("realm1", "/a/b/c")` immediately followed by `addShardingKey("realm2", "/a/b/c")`. Since the validation logic relies on the in-memory copy of routing data, which is updated by data callback, if the second call fires before data callback, the validation logic will treat the second call as valid. This can cause the routing data to have one sharding key pointing to two realms, which is invalid. 
Following this train of thought, we discovered that it is possible for two threads to do the two mentioned-above calls at the same time, which is even more difficult to guard. 
To fix this issue, all writing operations will be wrapped in `synchronized (this)`, which ensures that only 1 writing operation of any kind will fire at a time. After any writing operation, we also manually refresh the in-memory routing data to reflect the changes.

Issue #842, opened 2020-03-02
Implement setRoutingData for MetadataStoreDirectoryService


NealSun96:
`setRoutingData` was an endpoint that was defined but not finished. It allows users to overwrite whatever routing data there is for a namespace and replace it with a new set of data. This endpoint was designed with convenience in mind because it allows users to specify the routing data in one call instead of making multiple calls to create zkrealms and add sharding keys. 
Originally, it was designed that this endpoint will validate the routing data provided by users. However, we have decided against that (related: #841 ). This is because `getRoutingData` is able to return invalid or incomplete routing data, such as one that includes a zkRealm with no sharding keys. `setRoutingData` should be able to use the response from `getRoutingData` without an error. Users will be notified in the manual that `setRoutingData` does not validate the routing data they provided, and they will be responsible for the validity.

PR #844, opened 2020-03-02
Implement setRoutingData for MetadataStoreDirectoryService

### LINKED ISSUE #841: Fix TrieRoutingData initialization and usage in MetadataStoreDirectory
Previously in `ZkMetadataStoreDirectory`, instances of `TrieRoutingData` are always created after fetching the raw routing data using `ZkRoutingDataReader`. However, this has caused uncaught exceptions and other unspecified behaviors: `TrieRoutingData` will throw exceptions when given an empty map, however it is valid to get an empty map as the "raw routing data". Similarly as an example, we have allowed users to create zkRealms without creating associated sharding keys, which will cause exceptions when `ZkMetadataStoreDirectory` attempts to construct `TrieRoutingData` during data callback. Across the API specifications, we allowed a lot of operations that will fail `TrieRoutingData` constructions. 
In light of this issue, we reconsidered the definition of "routing data validity" and we concluded that the raw routing data is allowed to be "invalid" (or incomplete), such as having a zkRealm without sharding keys, but `TrieRoutingData` is not allowed to be invalid - if it exists it's valid. As a result, we no longer always create `TrieRoutingData` after fetching raw routing data from ZooKeeper; they are decoupled. We allow namespaces to not have `TrieRoutingData` in `ZkMetadataStoreDirectory`. When calls are made that depend on `TrieRoutingData` but `TrieRoutingData` doesn't exist, we fail the call and notify the users that the current routing data is invalid.

NealSun96:
### Issues

Fixes #841, #842, #843 

### Description

This PR implements `setRoutingData`, which was an endpoint that was defined but not finished. It allows users to overwrite whatever routing data there is for a namespace and replace it with a new set of data. This endpoint was designed with convenience in mind because it allows users to specify the routing data in one call instead of making multiple calls to create zkrealms and add sharding keys. This endpoint was originally designed to validate the input but now it doesn't, please refer #842 for the reasoning. 

Related to this PR, we also discovered that there are problems related to the construction of `TrieRoutingData` in `MetadataStoreDirectory`. Please see #841 . 

Lastly, we are also fixing the race conditions of writing operations in `MetadataStoreDirectory` which are discovered during the development. Please see #843. 

### Tests

[INFO] Tests run: 140, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 24.602 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 140, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  53.115 s
[INFO] Finished at: 2020-03-02T14:38:49-08:00

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -161,11 +161,11 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
 
   @Override
   public boolean setNamespaceRoutingData(String namespace, Map<String, List<String>> routingData) {
+    if (!_routingDataWriterMap.containsKey(namespace)) {
+      throw new IllegalArgumentException(
+          "Failed to set routing data: Namespace " + namespace + " is not found!");
+    }
     synchronized (this) {
-      if (!_routingDataWriterMap.containsKey(namespace)) {
-        throw new IllegalArgumentException(
-            "Failed to set routing data: Namespace " + namespace + " is not found!");
-      }
       boolean result = _routingDataWriterMap.get(namespace).setRoutingData(routingData);
       refreshRoutingData(namespace);
       return result;
@@ -216,13 +216,13 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
 
   @Override
   public boolean addMetadataStoreRealm(String namespace, String realm) {
+    if (!_routingDataWriterMap.containsKey(namespace)) {
+      // throwing NoSuchElementException instead of IllegalArgumentException to differentiate the
+      // status code in the Accessor level
+      throw new NoSuchElementException(
+          "Failed to add metadata store realm: Namespace " + namespace + " is not found!");
+    }
     synchronized (this) {
-      if (!_routingDataWriterMap.containsKey(namespace)) {
-        // throwing NoSuchElementException instead of IllegalArgumentException to differentiate the
-        // status code in the Accessor level
-        throw new NoSuchElementException(
-            "Failed to add metadata store realm: Namespace " + namespace + " is not found!");
-      }
       boolean result = _routingDataWriterMap.get(namespace).addMetadataStoreRealm(realm);
       refreshRoutingData(namespace);
       return result;
@@ -231,13 +231,13 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
 
   @Override
   public boolean deleteMetadataStoreRealm(String namespace, String realm) {
+    if (!_routingDataWriterMap.containsKey(namespace)) {
+      // throwing NoSuchElementException instead of IllegalArgumentException to differentiate the
+      // status code in the Accessor level
+      throw new NoSuchElementException(
+          "Failed to delete metadata store realm: Namespace " + namespace + " is not found!");
+    }
     synchronized (this) {
-      if (!_routingDataWriterMap.containsKey(namespace)) {
-        // throwing NoSuchElementException instead of IllegalArgumentException to differentiate the
-        // status code in the Accessor level
-        throw new NoSuchElementException(
-            "Failed to delete metadata store realm: Namespace " + namespace + " is not found!");
-      }
       boolean result = _routingDataWriterMap.get(namespace).deleteMetadataStoreRealm(realm);
       refreshRoutingData(namespace);
       return result;
@@ -246,13 +246,13 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
 
   @Override
   public boolean addShardingKey(String namespace, String realm, String shardingKey) {
+    if (!_routingDataWriterMap.containsKey(namespace)) {
+      // throwing NoSuchElementException instead of IllegalArgumentException to differentiate the
+      // status code in the Accessor level
+      throw new NoSuchElementException(
+          "Failed to add sharding key: Namespace " + namespace + " is not found!");
+    }
     synchronized (this) {
-      if (!_routingDataWriterMap.containsKey(namespace)) {
-        // throwing NoSuchElementException instead of IllegalArgumentException to differentiate the
-        // status code in the Accessor level
-        throw new NoSuchElementException(
-            "Failed to add sharding key: Namespace " + namespace + " is not found!");
-      }
       if (_routingDataMap.containsKey(namespace) && _routingDataMap.get(namespace)
           .containsKeyRealmPair(shardingKey, realm)) {
         return true;
@@ -271,13 +271,13 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
 
   @Override
   public boolean deleteShardingKey(String namespace, String realm, String shardingKey) {
+    if (!_routingDataWriterMap.containsKey(namespace)) {
+      // throwing NoSuchElementException instead of IllegalArgumentException to differentiate the
+      // status code in the Accessor level
+      throw new NoSuchElementException(
+          "Failed to delete sharding key: Namespace " + namespace + " is not found!");
+    }
     synchronized (this) {
-      if (!_routingDataWriterMap.containsKey(namespace)) {
-        // throwing NoSuchElementException instead of IllegalArgumentException to differentiate the
-        // status code in the Accessor level
-        throw new NoSuchElementException(
-            "Failed to delete sharding key: Namespace " + namespace + " is not found!");
-      }
       boolean result = _routingDataWriterMap.get(namespace).deleteShardingKey(realm, shardingKey);
       refreshRoutingData(namespace);
       return result;
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -119,7 +119,7 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
           try {
             _routingDataMap.put(namespace, new TrieRoutingData(rawRoutingData));
           } catch (InvalidRoutingDataException e) {
-            // Do not create TrieRoutingData if the routing data is invalid
+            LOG.warn("TrieRoutingData is not created for namespace {}", namespace, e);
           }
         }
       }
@@ -194,7 +194,7 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
     // If namespace is included but not routing data, it means the routing data is invalid
     if (!_routingDataMap.containsKey(namespace)) {
       throw new IllegalStateException("Failed to get all mapping under path: Namespace " + namespace
-          + " contains invalid routing data!");
+          + " contains either empty or invalid routing data!");
     }
     return _routingDataMap.get(namespace).getAllMappingUnderPath(path);
   }
@@ -209,7 +209,7 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
     // If namespace is included but not routing data, it means the routing data is invalid
     if (!_routingDataMap.containsKey(namespace)) {
       throw new IllegalStateException("Failed to get metadata store realm: Namespace " + namespace
-          + " contains invalid routing data!");
+          + " contains either empty or invalid routing data!");
     }
     return _routingDataMap.get(namespace).getMetadataStoreRealm(shardingKey);
   }
@@ -324,7 +324,7 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
     try {
       _routingDataMap.put(namespace, new TrieRoutingData(rawRoutingData));
     } catch (InvalidRoutingDataException e) {
-      // Do not create TrieRoutingData if the routing data is invalid
+      LOG.warn("TrieRoutingData is not created for namespace {}", namespace, e);
     }
   }
 
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -113,10 +113,14 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
           _routingDataWriterMap.put(namespace, new ZkRoutingDataWriter(namespace, zkAddress));
 
           // Populate realmToShardingKeys with ZkRoutingDataReader
-          _realmToShardingKeysMap
-              .put(namespace, _routingDataReaderMap.get(namespace).getRoutingData());
-          _routingDataMap
-              .put(namespace, new TrieRoutingData(_realmToShardingKeysMap.get(namespace)));
+          Map<String, List<String>> rawRoutingData =
+              _routingDataReaderMap.get(namespace).getRoutingData();
+          _realmToShardingKeysMap.put(namespace, rawRoutingData);
+          try {
+            _routingDataMap.put(namespace, new TrieRoutingData(rawRoutingData));
+          } catch (InvalidRoutingDataException e) {
+            // Do not create TrieRoutingData if the routing data is invalid
+          }
         }
       }
     }
@@ -156,6 +160,19 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
   }
 
   @Override
+  public boolean setNamespaceRoutingData(String namespace, Map<String, List<String>> routingData) {
+    synchronized (this) {
+      if (!_routingDataWriterMap.containsKey(namespace)) {
+        throw new IllegalArgumentException(
+            "Failed to set routing data: Namespace " + namespace + " is not found!");
+      }
+      boolean result = _routingDataWriterMap.get(namespace).setRoutingData(routingData);
+      refreshRoutingData(namespace);
+      return result;
+    }
+  }
+
+  @Override
   public Collection<String> getAllShardingKeysInRealm(String namespace, String realm) {
     if (!_realmToShardingKeysMap.containsKey(namespace)) {
       throw new NoSuchElementException("Namespace " + namespace + " does not exist!");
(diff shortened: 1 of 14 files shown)


helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -113,10 +113,14 @@ private void init(String namespace, String zkAddress) throws InvalidRoutingDataE
...
-              .put(namespace, _routingDataReaderMap.get(namespace).getRoutingData());
-          _routingDataMap
-              .put(namespace, new TrieRoutingData(_realmToShardingKeysMap.get(namespace)));
+          Map<String, List<String>> rawRoutingData =
+              _routingDataReaderMap.get(namespace).getRoutingData();
+          _realmToShardingKeysMap.put(namespace, rawRoutingData);
+          try {
+            _routingDataMap.put(namespace, new TrieRoutingData(rawRoutingData));
+          } catch (InvalidRoutingDataException e) {
+            // Do not create TrieRoutingData if the routing data is invalid
```
pkuwm: Do we want to throw any exception or log this?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -155,6 +159,19 @@ private void init(String namespace, String zkAddress) throws InvalidRoutingDataE
     return routingData;
   }
 
+  @Override
+  public boolean setNamespaceRoutingData(String namespace, Map<String, List<String>> routingData) {
+    synchronized (this) {
+      if (!_routingDataWriterMap.containsKey(namespace)) {
```
pkuwm: This is like double lock checking. You may also want to check it before synchronized which also improves performance.
Considering namespace is not changed, you don’t have to check it in within synchronized block.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -169,72 +186,102 @@ private void init(String namespace, String zkAddress) throws InvalidRoutingDataE
...
 
   @Override
   public boolean deleteShardingKey(String namespace, String realm, String shardingKey) {
-    if (!_routingDataWriterMap.containsKey(namespace)) {
-      // throwing NoSuchElementException instead of IllegalArgumentException to differentiate the
-      // status code in the Accessor level
-      throw new NoSuchElementException(
-          "Failed to delete sharding key: Namespace " + namespace + " is not found!");
+    synchronized (this) {
+      if (!_routingDataWriterMap.containsKey(namespace)) {
```
pkuwm: move it before synchronized

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -169,72 +186,102 @@ private void init(String namespace, String zkAddress) throws InvalidRoutingDataE
...
 
   @Override
   public boolean addMetadataStoreRealm(String namespace, String realm) {
-    if (!_routingDataWriterMap.containsKey(namespace)) {
-      // throwing NoSuchElementException instead of IllegalArgumentException to differentiate the
-      // status code in the Accessor level
-      throw new NoSuchElementException(
-          "Failed to add metadata store realm: Namespace " + namespace + " is not found!");
+    synchronized (this) {
+      if (!_routingDataWriterMap.containsKey(namespace)) {
```
pkuwm: Before synchronized

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: We are not throwing any exception because we allow this to happen. For logs, it's a good point - in my opinion, logging this may pollute the main log because in most cases this situation isn't even qualified as an error (when a namespace is initialized, this exception clause will happen because `rawRoutingData` is empty, which is a normal situation); however, when this is an actual error situation, it's better to log it. Since this logic is in data change callback, it's very easy to spam logs that are non-essential in this case, hence why I didn't log anything here. 
Since I have conflicting views myself, I'm open to your suggestions.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: Sure.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
narendly: I think a WARN log would be okay to add here. Note that we are not expecting very frequent writes :)

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -169,19 +186,31 @@ private void init(String namespace, String zkAddress) throws InvalidRoutingDataE
...
-    if (!_routingDataMap.containsKey(namespace)) {
+    // Check _routingZkAddressMap first to see if namespace is included
+    if (!_routingZkAddressMap.containsKey(namespace)) {
       throw new NoSuchElementException(
           "Failed to get all mapping under path: Namespace " + namespace + " is not found!");
     }
+    // If namespace is included but not routing data, it means the routing data is invalid
+    if (!_routingDataMap.containsKey(namespace)) {
+      throw new IllegalStateException("Failed to get all mapping under path: Namespace " + namespace
+          + " contains invalid routing data!");
```
narendly: Could we rephrase this to say, the routing data is either _empty_ or invalid?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -169,19 +186,31 @@ private void init(String namespace, String zkAddress) throws InvalidRoutingDataE
...
-    if (!_routingDataMap.containsKey(namespace)) {
+    // Check _routingZkAddressMap first to see if namespace is included
+    if (!_routingZkAddressMap.containsKey(namespace)) {
       throw new NoSuchElementException(
           "Failed to get metadata store realm: Namespace " + namespace + " is not found!");
     }
+    // If namespace is included but not routing data, it means the routing data is invalid
+    if (!_routingDataMap.containsKey(namespace)) {
+      throw new IllegalStateException("Failed to get metadata store realm: Namespace " + namespace
+          + " contains invalid routing data!");
```
narendly: Same as above :)

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -262,17 +309,22 @@ public void refreshRoutingData(String namespace) {
...
-      _routingDataMap.put(namespace, routingData);
     } catch (InvalidRoutingDataException e) {
       LOG.error("Failed to refresh cached routing data for namespace {}", namespace, e);
+      return;
+    }
+
+    try {
+      _routingDataMap.put(namespace, new TrieRoutingData(rawRoutingData));
+    } catch (InvalidRoutingDataException e) {
+      // Do not create TrieRoutingData if the routing data is invalid
```
narendly: Probably a good idea to add a warn log here?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -209,8 +214,20 @@ public synchronized boolean setRoutingData(Map<String, List<String>> routingData
...
 
-    // TODO: Forward the request to leader
-    return true;
+    String leaderHostName = _leaderElection.getCurrentLeaderInfo().getId();
+    String url = leaderHostName + constructUrlSuffix(
+        MetadataStoreRoutingConstants.MSDS_GET_ALL_ROUTING_DATA_ENDPOINT);
+    HttpPut httpPut = new HttpPut(url);
+    String routingDataJsonString;
+    try {
+      routingDataJsonString = new ObjectMapper().writeValueAsString(routingData);
```
narendly: Could we create a constant variable ObjectMapper instead of creating one every time this endpoint is called?

private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();

and you could use this throughout the class :)

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -332,8 +349,9 @@ private String constructUrlSuffix(String... urlParams) {
     return String.join("", allUrlParameters);
   }
 
-  private boolean forwardRequestToLeader(String urlSuffix, HttpConstants.RestVerbs request_method,
-      int expectedResponseCode) throws IllegalArgumentException {
+  private boolean buildAndSendRequestToLeader(String urlSuffix,
+      HttpConstants.RestVerbs request_method, int expectedResponseCode)
```
narendly: shouldWeUseCamelCaseFor request_method?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
```diff
@@ -220,6 +228,24 @@ public Response getRoutingData() {
     return JSONRepresentation(responseMap);
   }
 
+  @PUT
+  @Path("/routing-data")
+  @Consumes(MediaType.APPLICATION_JSON)
+  public Response setRoutingData(String jsonContent) {
+    try {
+      Map<String, List<String>> routingData = new ObjectMapper()
```
narendly: Same thing as above. We can create ObjectMapper once and use it over and over.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataReader.java
```diff
@@ -130,17 +116,11 @@ public void testGetRoutingDataMSRDChildEmptyValue() {
...
-        return false;
-      }
-
-      return true;
-    }, TestHelper.WAIT_DURATION), "Routing data path should be deleted after the tests.");
+  private void clearRoutingDataPath() {
+    for (String zkRealm : _baseAccessor
+        .getChildNames(MetadataStoreRoutingConstants.ROUTING_DATA_PATH, AccessOption.PERSISTENT)) {
+      _baseAccessor.remove(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + zkRealm,
+          AccessOption.PERSISTENT);
```
narendly: Nit: Outside of helix-core module, we should try to use ZkClient instead of BaseDataAccessor.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
```diff
@@ -132,61 +132,86 @@ public void testSetRoutingData() {
   public void testAddMetadataStoreRealmNonLeader() {
     MockWriter mockWriter = new MockWriter(DUMMY_NAMESPACE, ZK_ADDR);
     mockWriter.addMetadataStoreRealm(DUMMY_REALM);
-    Assert.assertEquals("PUT", mockWriter.calledRequest.getMethod());
+    Assert.assertEquals(mockWriter.calledRequest.getMethod(), "PUT");
```
narendly: Could we use a pre-defined constant instead of "PUT"? (Cue: HttpVerbs..) Same applies below.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
```diff
@@ -132,61 +132,86 @@ public void testSetRoutingData() {
...
+            .substring(1);
+    Assert.assertEquals(mockWriter.calledRequest.getURI().toString(), expectedUrl);
+    mockWriter.close();
+  }
+
+  private void clearRoutingDataPath() {
+    for (String zkRealm : _baseAccessor
+        .getChildNames(MetadataStoreRoutingConstants.ROUTING_DATA_PATH, AccessOption.PERSISTENT)) {
+      _baseAccessor.remove(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + zkRealm,
+          AccessOption.PERSISTENT);
```
narendly: It would be better to stick with ZkClient if possible outside of helix-core module.

helix-rest/src/test/java/org/apache/helix/rest/server/MetadataStoreDirectoryAccessorTestBase.java
```diff
@@ -100,23 +100,19 @@ public void beforeClass() throws Exception {
...
-      for (String zk : _zkList) {
-        if (ZK_SERVER_MAP.get(zk).getZkClient()
-            .exists(MetadataStoreRoutingConstants.ROUTING_DATA_PATH)) {
-          return false;
+  protected void deleteRoutingDataPath() {
+    for (String zk : _zkList) {
+      ZkClient zkClient = ZK_SERVER_MAP.get(zk).getZkClient();
+      if (zkClient.exists(MetadataStoreRoutingConstants.ROUTING_DATA_PATH)) {
+        for (String zkRealm : zkClient.getChildren(MetadataStoreRoutingConstants.ROUTING_DATA_PATH)) {
+          zkClient.delete(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + zkRealm);
```
narendly: We shouldn't remove the TestHelper block. Could we add that back? In general, please try to add a verify() block after cleanup.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMetadataStoreDirectoryAccessor.java
```diff
@@ -452,6 +454,43 @@ public void testDeleteShardingKey() throws InvalidRoutingDataException {
...
+  public void testSetRoutingData() throws InvalidRoutingDataException, IOException {
+    Map<String, List<String>> routingData = new HashMap<>();
+    routingData.put(TEST_REALM_1, TEST_SHARDING_KEYS_2);
+    routingData.put(TEST_REALM_2, TEST_SHARDING_KEYS_1);
+    String routingDataString = new ObjectMapper().writeValueAsString(routingData);
+
+    Map<String, String> badFormatRoutingData = new HashMap<>();
+    badFormatRoutingData.put(TEST_REALM_1, TEST_REALM_2);
+    badFormatRoutingData.put(TEST_REALM_2, TEST_REALM_1);
+    String badFormatRoutingDataString = new ObjectMapper().writeValueAsString(badFormatRoutingData);
```
narendly: Please create a constant OBJECT_MAPPER and use it throughout.

metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/datamodel/TrieRoutingData.java
```diff
@@ -167,8 +167,8 @@ private TrieNode getLongestPrefixNodeAlongPath(String path) {
   }
 
   /*
-   * Checks for the edge case when the only sharding key in provided routing data is the delimiter
-   * or an empty string. When this is the case, the trie is valid and contains only one node, which
+   * Checks for the edge case when the only sharding key in provided routing data is the delimiter.
+   * When this is the case, the trie is valid and contains only one node, which
```
narendly: Question: we still want to throw an exception when the data is empty, right?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: Using WARN is a good middle ground. I'll add logging for this section and also the next. Thank you for the suggestion!

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: Agreed.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Sure.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Ah sorry, python habits coming through. Fixing it.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: Although I agree with the same change in `ZkRoutingDataWriter`, in this case it's actually unnecessary because `MetadataStoreDirectoryAccessor` has a per-request cycle. This object mapper will be reconstructed every time anyway. Given that, the object mapper is better a local variable because it's not used anywhere else.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataReader.java
NealSun96: I believe we talked about this before implementing and decided to use `BaseDataAccessor`. This is good to know. 
Regardless, how do you feel about using `HelixZkClient`? It's used in other tests as well (like the one for leader election), but it's a deprecated class. Is it really okay to use it among the tests? I'd expect the general consensus to be avoid using it.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataWriter.java
NealSun96: Sure.

helix-rest/src/test/java/org/apache/helix/rest/server/MetadataStoreDirectoryAccessorTestBase.java
NealSun96: Very good point. I'll try to add the same block to all similar clean ups.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataReader.java
NealSun96: Resolved offline. Using `HelixZkClient` and will migrate later.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/metadatastore/MetadataStoreDirectoryAccessor.java
NealSun96: On a second thought, the ObjectMapper will be static and is reusable. Making the changes, then.

helix-rest/src/test/java/org/apache/helix/rest/server/TestMetadataStoreDirectoryAccessor.java
NealSun96: Reusing the inherited `OBJECT_MAPPER`. Applies to many other files as well.

metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/datamodel/TrieRoutingData.java
NealSun96: The comment change here was regarding to the sharding key being either "/" or "", not routing data being empty. I believe your question is about when the entire routing data is empty: the answer is yes, we throw an exception. That behavior, as we agreed on, will not be changed.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -209,8 +215,20 @@ public synchronized boolean setRoutingData(Map<String, List<String>> routingData
...
+        MetadataStoreRoutingConstants.MSDS_GET_ALL_ROUTING_DATA_ENDPOINT);
+    HttpPut httpPut = new HttpPut(url);
+    String routingDataJsonString;
+    try {
+      routingDataJsonString = OBJECT_MAPPER.writeValueAsString(routingData);
+    } catch (JsonGenerationException | JsonMappingException e) {
+      throw new IllegalArgumentException(e.getMessage());
+    } catch (IOException e) {
+      return false;
+    }
```
narendly: Do we need to log the exception here? Otherwise you'd be failing silently.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -332,20 +350,21 @@ private String constructUrlSuffix(String... urlParams) {
...
+    switch (requestMethod) {
       case PUT:
         request = new HttpPut(url);
         break;
       case DELETE:
         request = new HttpDelete(url);
         break;
       default:
-        LOG.error("Unsupported request_method: " + request_method.name());
+        LOG.error("Unsupported requestMethod: " + requestMethod.name());
```
narendly: I think it would be better to throw an exception here because this is not supposed to happen.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/TestZkMetadataStoreDirectory.java
```diff
@@ -277,4 +315,27 @@ public void testChildChangeCallback() throws Exception {
       return false;
     }, TestHelper.WAIT_DURATION));
   }
+
+  private void clearRoutingData() throws Exception {
+    Assert.assertTrue(TestHelper.verify(() -> {
+      for (String zk : _zkList) {
```
narendly: Could we add the retry logic on false?

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataReader.java
```diff
@@ -130,17 +123,14 @@ public void testGetRoutingDataMSRDChildEmptyValue() {
...
     Assert.assertTrue(TestHelper.verify(() -> {
-      ZK_SERVER_MAP.get(ZK_ADDR).getZkClient()
-          .deleteRecursively(MetadataStoreRoutingConstants.ROUTING_DATA_PATH);
-
-      if (ZK_SERVER_MAP.get(ZK_ADDR).getZkClient()
-          .exists(MetadataStoreRoutingConstants.ROUTING_DATA_PATH)) {
-        return false;
+      for (String zkRealm : _zkClient
+          .getChildren(MetadataStoreRoutingConstants.ROUTING_DATA_PATH)) {
+        _zkClient.delete(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + zkRealm);
```
narendly: Need a retry while block

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/TestZkMetadataStoreDirectory.java
NealSun96: `verify()` already retries, right?

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataReader.java
NealSun96: Ditto above.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: Sure. This block should technically never have a problem though. I'll add a log.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
NealSun96: This is a strange one because it can only go to `default` because of developer oversights. 
If some code using the wrong http verb does make it to production, and clients are calling the endpoints with the incorrect code, we definitely want to report 500 to clients (return false) instead of 400 (`IllegalArgumentException`). 
I'll change it to exception because `return false` isn't informative enough to developers, but I wish there's a way to limit the values of a parameter.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/TestZkMetadataStoreDirectory.java
NealSun96: For your convenience:
```
  public static boolean verify(Verifier verifier, long timeout) throws Exception {
    do {
      boolean result = verifier.verify();
      if (result || (System.currentTimeMillis() - start) > timeout) {
        return result;
      }
      Thread.sleep(50);
    } while (true);
  }
```

narendly: :+1:

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/accessor/TestZkRoutingDataReader.java
narendly: :+1:

NealSun96:
Final commit message:
## Implement setRoutingData for MetadataStoreDirectoryService ##
Implement setRoutingData endpoint. Modify TrieRoutingData construction in MetadataStoreDirectory. Fix race conditions among writing operations in MetadataStoreDirectory.

PR #846, opened 2020-03-03
Make ZKHelixAdmin and ZKHelixManager Realm-aware

pkuwm:
### Issues

Implements #847 

### Description

To make Helix Java APIs realm-aware, we need to make ZKHelixAdmin and ZKHelixManager realm-aware.

### Tests

Current TestZkHelixAdmin, TestHandleNewSession  cover the tests
```
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestRebalanceRunningTask.testFixedTargetTaskAndDisabledRebalanceAndNodeAdded:265 expected:<true> but was:<false>
[INFO]
[ERROR] Tests run: 1083, Failures: 2, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:13 h
[INFO] Finished at: 2020-03-11T23:49:26-07:00

[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 19.945 s - in org.apache.helix.integration.task.TestRebalanceRunningTask
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS

[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 13.597 s - in org.apache.helix.integration.task.TestJobQueueCleanUp
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
```

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -94,7 +94,6 @@ import org.slf4j.LoggerFactory;
 public class ZKHelixAdmin implements HelixAdmin {
   private static final Logger LOG = LoggerFactory.getLogger(ZKHelixAdmin.class);
 
-
   public static final String CONNECTION_TIMEOUT = "helixAdmin.timeOutInSec";
   private static final String MAINTENANCE_ZNODE_ID = "maintenance";
   private static final int DEFAULT_SUPERCLUSTER_REPLICA = 3;
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -49,6 +49,7 @@ import org.apache.helix.InstanceType;
 import org.apache.helix.PropertyKey;
 import org.apache.helix.PropertyPathBuilder;
 import org.apache.helix.PropertyType;
+import org.apache.helix.SystemPropertyKeys;
 import org.apache.helix.controller.rebalancer.DelayedAutoRebalancer;
 import org.apache.helix.controller.rebalancer.strategy.CrushEdRebalanceStrategy;
 import org.apache.helix.controller.rebalancer.strategy.RebalanceStrategy;
@@ -107,8 +108,9 @@ public class ZKHelixAdmin implements HelixAdmin {
   private static Logger logger = LoggerFactory.getLogger(ZKHelixAdmin.class);
 
   /**
-   * @deprecated it is recommended to use the other constructors instead to avoid having to manually
-   * create and maintain a RealmAwareZkClient outside of ZkBaseDataAccessor.
+   * @deprecated it is recommended to use the builder constructor {@link Builder}
+   * instead to avoid having to manually create and maintain a RealmAwareZkClient
+   * outside of ZKHelixAdmin.
    *
    * @param zkClient A created RealmAwareZkClient
    */
@@ -119,6 +121,18 @@ public class ZKHelixAdmin implements HelixAdmin {
     _usesExternalZkClient = true;
   }
 
+  /**
+   * There are 2 realm-aware modes to connect to ZK:
+   * 1. if system property {@link SystemPropertyKeys#MULTI_ZK_ENABLED} is set to <code>"true"</code>,
+   * it will connect on multi-realm mode;
+   * 2. otherwise, it will connect on single-realm mode to the <code>zkAddress</code> provided.
+   *
+   * @param zkAddress ZK address
+   * @exception HelixException if not able to connect on multi-realm mode
+   *
+   * @deprecated it is recommended to use the builder constructor {@link Builder}
+   */
+  @Deprecated
   public ZKHelixAdmin(String zkAddress) {
     int timeOutInSec = Integer.parseInt(System.getProperty(CONNECTION_TIMEOUT, "30"));
     RealmAwareZkClient.RealmAwareZkClientConfig clientConfig =
@@ -127,13 +141,15 @@ public class ZKHelixAdmin implements HelixAdmin {
             .setZkSerializer(new ZNRecordSerializer());
 
     RealmAwareZkClient zkClient;
-    try {
-      zkClient = new FederatedZkClient(
-          new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder().build(), clientConfig);
-    } catch (IllegalStateException | IOException | InvalidRoutingDataException e) {
-      LOG.info("Not able to connect on multi-realm mode. "
-          + "Connecting on single-realm mode to ZK: {}", zkAddress);
 
+    if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED)) {
+      try {
+        zkClient = new FederatedZkClient(
+            new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder().build(), clientConfig);
+      } catch (IllegalStateException | IOException | InvalidRoutingDataException e) {
+        throw new HelixException("Not able to connect on multi-realm mode.", e);
+      }
+    } else {
       zkClient = SharedZkClientFactory.getInstance()
           .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
               clientConfig.createHelixZkClientConfig());
@@ -1883,7 +1899,7 @@ public class ZKHelixAdmin implements HelixAdmin {
       return this;
     }
 
-    public ZKHelixAdmin build() throws Exception {
+    public ZKHelixAdmin build() {
       validate();
       return new ZKHelixAdmin(this);
     }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -1896,7 +1896,11 @@ public class ZKHelixAdmin implements HelixAdmin {
       boolean isZkAddressSet = zkAddress != null && !zkAddress.isEmpty();
       if (realmMode == RealmAwareZkClient.RealmMode.SINGLE_REALM && !isZkAddressSet) {
         throw new HelixException(
-            "ConfigAccessor: RealmMode cannot be single-realm without a valid ZkAddress set!");
+            "RealmMode cannot be single-realm without a valid ZkAddress set!");
+      }
+      if (realmMode == RealmAwareZkClient.RealmMode.MULTI_REALM && isZkAddressSet) {
+        throw new HelixException(
+            "ZkAddress cannot be set on multi-realm mode!");
       }
       if (realmMode == null) {
         realmMode = isZkAddressSet ? RealmAwareZkClient.RealmMode.SINGLE_REALM
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -145,23 +145,34 @@ public class ZKHelixAdmin implements HelixAdmin {
     _usesExternalZkClient = false;
   }
 
-  private ZKHelixAdmin(Builder builder) throws IOException, InvalidRoutingDataException {
+  private ZKHelixAdmin(Builder builder) {
+    RealmAwareZkClient zkClient;
     switch (builder.realmMode) {
       case MULTI_REALM:
-        _zkClient = new FederatedZkClient(builder.realmAwareZkConnectionConfig,
-            builder.realmAwareZkClientConfig);
-        break;
+        try {
+          zkClient = new FederatedZkClient(builder.realmAwareZkConnectionConfig,
+              builder.realmAwareZkClientConfig);
+          break;
+        } catch (IOException | InvalidRoutingDataException | IllegalStateException e) {
+          if (builder.zkAddress == null || builder.zkAddress.isEmpty()) {
+            throw new IllegalStateException("Not able to connect on multi-realm mode.", e);
+          }
+          LOG.info("Not able to connect on multi-realm mode. "
+              + "Connecting on single-realm mode to ZK: {}", builder.zkAddress);
+          builder.setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM);
+        }
       case SINGLE_REALM:
         // Create a HelixZkClient: Use a SharedZkClient because ZKHelixAdmin does not need to do
         // ephemeral operations
-        _zkClient = SharedZkClientFactory.getInstance()
-            .buildZkClient(builder.realmAwareZkConnectionConfig.createZkConnectionConfig(),
+        zkClient = SharedZkClientFactory.getInstance()
+            .buildZkClient(new HelixZkClient.ZkConnectionConfig(builder.zkAddress),
                 builder.realmAwareZkClientConfig.createHelixZkClientConfig());
         break;
       default:
         throw new HelixException("Invalid RealmMode given: " + builder.realmMode);
     }
 
+    _zkClient = zkClient;
     _configAccessor = new ConfigAccessor(_zkClient);
     _usesExternalZkClient = false;
   }
@@ -1898,11 +1909,9 @@ public class ZKHelixAdmin implements HelixAdmin {
       }
 
       // Resolve RealmAwareZkClientConfig
-      boolean isZkClientConfigSet = realmAwareZkClientConfig != null;
-      // Resolve which clientConfig to use
-      realmAwareZkClientConfig =
-          isZkClientConfigSet ? realmAwareZkClientConfig.createHelixZkClientConfig()
-              : new HelixZkClient.ZkClientConfig().setZkSerializer(new ZNRecordSerializer());
+      if (realmAwareZkClientConfig == null) {
+        realmAwareZkClientConfig = new RealmAwareZkClient.RealmAwareZkClientConfig();
+      }
 
       // Resolve RealmAwareZkConnectionConfig
       if (realmAwareZkConnectionConfig == null) {
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -152,15 +152,10 @@ public class ZKHelixAdmin implements HelixAdmin {
         try {
           zkClient = new FederatedZkClient(builder.realmAwareZkConnectionConfig,
               builder.realmAwareZkClientConfig);
-          break;
         } catch (IOException | InvalidRoutingDataException | IllegalStateException e) {
-          if (builder.zkAddress == null || builder.zkAddress.isEmpty()) {
-            throw new IllegalStateException("Not able to connect on multi-realm mode.", e);
-          }
-          LOG.info("Not able to connect on multi-realm mode. "
-              + "Connecting on single-realm mode to ZK: {}", builder.zkAddress);
-          builder.setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM);
+          throw new HelixException("Not able to connect on multi-realm mode.", e);
         }
+        break;
       case SINGLE_REALM:
         // Create a HelixZkClient: Use a SharedZkClient because ZKHelixAdmin does not need to do
         // ephemeral operations
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -1868,6 +1868,7 @@ public class ZKHelixAdmin implements HelixAdmin {
     return true;
   }
 
+  // TODO: refactor builder to reduce duplicate code with other Helix Java APIs
   public static class Builder {
     private String zkAddress;
     private RealmAwareZkClient.RealmMode realmMode;
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -47,7 +47,6 @@ import org.apache.helix.HelixDefinedState;
 import org.apache.helix.HelixException;
 import org.apache.helix.InstanceType;
 import org.apache.helix.PropertyKey;
-import org.apache.helix.PropertyKey.Builder;
 import org.apache.helix.PropertyPathBuilder;
 import org.apache.helix.PropertyType;
 import org.apache.helix.controller.rebalancer.DelayedAutoRebalancer;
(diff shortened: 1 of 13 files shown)


helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -118,6 +118,11 @@ public ConfigAccessor(HelixZkClient zkClient) {
     _usesExternalZkClient = true;
   }
 
+  public ConfigAccessor(RealmAwareZkClient zkClient) {
+    _zkClient = zkClient;
+    _usesExternalZkClient = true;
+  }
```
narendly: The new ConfigAccessor should be created with a Builder. I'm going to assume that this is just a placeholder? :)

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -89,33 +91,55 @@
...
   private final boolean _usesExternalZkClient;
 
   private static Logger logger = LoggerFactory.getLogger(ZKHelixAdmin.class);
 
-  @Deprecated
-  public ZKHelixAdmin(HelixZkClient zkClient) {
+  public ZKHelixAdmin() {
+    _zkClient = new FederatedZkClient();
+    _configAccessor = new ConfigAccessor(_zkClient);
+    _usesExternalZkClient = false;
```
narendly: We should use a Builder pattern for HelixAdmin here. This is because we need to allow users to provide connection and client configs. 

For example, users should be able to set operationRetry time. If we just provide a constructor that doesn't take in any parameters, it won't be very flexible.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -89,33 +91,55 @@
...
+  }
+
+  public ZKHelixAdmin(RealmAwareZkClient zkClient) {
```
narendly: We could just leave this as HelixZkClient - do you think that'd work?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -89,33 +91,55 @@
...
 
   public ZKHelixAdmin(String zkAddress) {
     int timeOutInSec = Integer.parseInt(System.getProperty(CONNECTION_TIMEOUT, "30"));
-    HelixZkClient.ZkClientConfig clientConfig = new HelixZkClient.ZkClientConfig();
-    clientConfig.setZkSerializer(new ZNRecordSerializer())
-        .setConnectInitTimeout(timeOutInSec * 1000);
-    _zkClient = SharedZkClientFactory.getInstance()
-        .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress), clientConfig);
-    _zkClient.waitUntilConnected(timeOutInSec, TimeUnit.SECONDS);
+    RealmAwareZkClient.RealmAwareZkClientConfig clientConfig =
```
narendly: I just realized I forgot to add this logic for ConfigAccessor. Let me open another PR :)

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
narendly: On a second thought, I forgot to add this constructor. I'll go ahead and work on that as well.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1285,4 +1264,87 @@ public void handleSessionEstablishmentError(Throwable error) throws Exception {
...
+              + " Fallback to HelixZkClient.", connectionConfig.getZkRealmShardingKey(),
+          e.getMessage());
+
+      // Fallback to HelixZkClient
+      HelixZkClient.ZkConnectionConfig helixZkConnectionConfig =
+          new HelixZkClient.ZkConnectionConfig(_zkAddress);
+      helixZkConnectionConfig.setSessionTimeout(connectionConfig.getSessionTimeout());
+      zkClient = DedicatedZkClientFactory.getInstance()
+          .buildZkClient(helixZkConnectionConfig, clientConfig.createHelixZkClientConfig());
+    }
```
narendly: Do you think we could refactor these two methods to reduce duplicate logic? You could take in the factory in the parameter and switch back and forth between dedicated and shared that way.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -79,7 +83,31 @@
...
   private PathBasedZkSerializer _pathBasedZkSerializer;
 
+  public FederatedZkClient() {
+    this(new RealmAwareZkClient.RealmAwareZkClientConfig()
+        .setZkSerializer(new ZNRecordSerializer()));
+  }
+
   // TODO: support capacity of ZkClient number in one FederatedZkClient and do garbage collection.
+  public FederatedZkClient(RealmAwareZkClient.RealmAwareZkClientConfig clientConfig) {
+    try {
```
narendly: FederatedZkClient should take in connectionConfig and clientConfig. I'll just assume this is a placeholder.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1285,4 +1266,85 @@ public void handleSessionEstablishmentError(Throwable error) throws Exception {
...
+    if (_instanceType == InstanceType.ADMINISTRATOR) {
+      newClient = buildSharedZkClient(connectionConfig, clientConfig);
+    } else {
+      newClient = buildDedicatedZkClient(connectionConfig, clientConfig);
+    }
+
+    return newClient;
+  }
+
+  private RealmAwareZkClient buildSharedZkClient(
```
NealSun96: Is it possible to combine `buildSharedZkClient` and `buildDedicatedZkClient` to reduce duplication? Both instances of the factories extend `HelixZkClientFactory`.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1285,4 +1266,85 @@ public void handleSessionEstablishmentError(Throwable error) throws Exception {
...
+      helixZkConnectionConfig.setSessionTimeout(connectionConfig.getSessionTimeout());
+      zkClient = DedicatedZkClientFactory.getInstance()
+          .buildZkClient(helixZkConnectionConfig, clientConfig.createHelixZkClientConfig());
+    }
+
+    return zkClient;
+  }
+
+  private String buildShardingKey() {
+    return _clusterName.charAt(0) == '/' ? _clusterName : SLASH + _clusterName;
```
NealSun96: Nit: not much point declaring `SLASH` if it's character counter-part is a local variable.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: I tried. We have 2 options:
1. As you said, we can pass in the factory as a parameter. But we have to make `HelixZkClientFactory` public. Shall we do this?
`public abstract class HelixZkClientFactory implements RealmAwareZkClientFactory`

2. It is the implemented as I do. Keep`HelixZkClientFactory` as it is. I prefer this considering that we don't want HelixZkClientFactory public.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: I tried. We have 2 options:
1. As you said, we can pass in the factory as a parameter. But we have to make `HelixZkClientFactory` public. Shall we do this?
`public abstract class HelixZkClientFactory implements RealmAwareZkClientFactory`

2. It is the implemented as I do. Keep`HelixZkClientFactory` as it is. I prefer this considering that we don't want HelixZkClientFactory public.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -89,33 +91,77 @@
...
+              clientConfig.createHelixZkClientConfig());
+      zkClient.waitUntilConnected(timeOutInSec, TimeUnit.SECONDS);
+    }
+
+    _zkClient = zkClient;
+    _configAccessor = new ConfigAccessor(_zkClient);
+    _usesExternalZkClient = false;
+  }
+
+  private ZKHelixAdmin(Builder builder) throws IOException, InvalidRoutingDataException {
```
narendly: Why is the constructor throwing all sorts of exceptions? As we discussed, this is where we catch checked exceptions.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -89,33 +91,77 @@
...
+    switch (builder.realmMode) {
+      case MULTI_REALM:
+        _zkClient = new FederatedZkClient(builder.realmAwareZkConnectionConfig,
+            builder.realmAwareZkClientConfig);
```
narendly: This should be a try-catch.

narendly: Example: https://github.com/apache/helix/pull/863/files#diff-259ce65919d1337b08cb736839d755feR178-R192

What are your thoughts?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1799,4 +1845,71 @@ private boolean validateWeightForResourceConfig(ClusterConfig clusterConfig,
...
+        realmMode = isZkAddressSet ? RealmAwareZkClient.RealmMode.SINGLE_REALM
+            : RealmAwareZkClient.RealmMode.MULTI_REALM;
+      }
+
+      // Resolve RealmAwareZkClientConfig
+      boolean isZkClientConfigSet = realmAwareZkClientConfig != null;
+      // Resolve which clientConfig to use
+      realmAwareZkClientConfig =
+          isZkClientConfigSet ? realmAwareZkClientConfig.createHelixZkClientConfig()
+              : new HelixZkClient.ZkClientConfig().setZkSerializer(new ZNRecordSerializer());
```
narendly: This could actually be simplified to 

```
      // Resolve RealmAwareZkClientConfig
      if (_realmAwareZkClientConfig == null) {
        _realmAwareZkClientConfig = new RealmAwareZkClient.RealmAwareZkClientConfig();
      }
```

pkuwm: Done. We may want to make `ConfigAccessor` consistent with this?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
pkuwm: Above reply.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1799,4 +1856,69 @@ private boolean validateWeightForResourceConfig(ClusterConfig clusterConfig,
...
+      return this;
+    }
+
+    public ZKHelixAdmin build() throws Exception {
+      validate();
+      return new ZKHelixAdmin(this);
+    }
+
+    /**
+     * Validate the given parameters before creating an instance of ConfigAccessor.
```
narendly: ConfigAccessor -> ZkHelixAdmin?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1285,4 +1266,62 @@ public void handleSessionEstablishmentError(Throwable error) throws Exception {
...
+
+    // Fall back to HelixZkClient
+    HelixZkClient.ZkClientConfig helixZkClientConfig = clientConfig.createHelixZkClientConfig();
+    HelixZkClient.ZkConnectionConfig helixZkConnectionConfig =
+        new HelixZkClient.ZkConnectionConfig(_zkAddress)
+            .setSessionTimeout(connectionConfig.getSessionTimeout());
+
+    return zkClientFactory.buildZkClient(helixZkConnectionConfig, helixZkClientConfig);
+  }
+
```
narendly: Nit: remove extra line

pkuwm:
> Looks good to me.
> 
> Also I think it would be a good idea to test the Builder for HelixAdmin to make sure the Builder creates a working HelixAdmin (or ConfigAccessor, etc.). I am trying to think of ways to do this in other PRs too. Let us think about it.
> 
> Totally fine to do that in a separate PR.

As I mentioned, I have the same plan to test and would like to do it for all of the required Java APIs in another PR.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -89,33 +91,88 @@
...
+          zkClient = new FederatedZkClient(builder.realmAwareZkConnectionConfig,
+              builder.realmAwareZkClientConfig);
+          break;
+        } catch (IOException | InvalidRoutingDataException | IllegalStateException e) {
+          if (builder.zkAddress == null || builder.zkAddress.isEmpty()) {
+            throw new IllegalStateException("Not able to connect on multi-realm mode.", e);
+          }
+          LOG.info("Not able to connect on multi-realm mode. "
+              + "Connecting on single-realm mode to ZK: {}", builder.zkAddress);
+          builder.setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM);
```
narendly: @pkuwm 
Please throw a HelixException here and propagate the exception. 
Also note that per our offline discussion, there will no longer be a fallback to single realm mode. If multi-realm fails, we fail the creation.

pkuwm: Done.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1799,4 +1851,69 @@ private boolean validateWeightForResourceConfig(ClusterConfig clusterConfig,
...
+
+    /*
+     * Validates the given parameters before creating an instance of ZKHelixAdmin.
+     */
+    private void validate() {
+      // Resolve RealmMode based on other parameters
+      boolean isZkAddressSet = zkAddress != null && !zkAddress.isEmpty();
+      if (realmMode == RealmAwareZkClient.RealmMode.SINGLE_REALM && !isZkAddressSet) {
+        throw new HelixException(
+            "ConfigAccessor: RealmMode cannot be single-realm without a valid ZkAddress set!");
```
narendly: This is not configaccessor...

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1799,4 +1851,69 @@ private boolean validateWeightForResourceConfig(ClusterConfig clusterConfig,
...
+      return new ZKHelixAdmin(this);
+    }
+
+    /*
+     * Validates the given parameters before creating an instance of ZKHelixAdmin.
+     */
+    private void validate() {
+      // Resolve RealmMode based on other parameters
+      boolean isZkAddressSet = zkAddress != null && !zkAddress.isEmpty();
+      if (realmMode == RealmAwareZkClient.RealmMode.SINGLE_REALM && !isZkAddressSet) {
```
narendly: Also we need to throw exception when zkaddr is set on multi-realm mode.

pkuwm: Fixed.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1285,4 +1266,63 @@ public void handleSessionEstablishmentError(Throwable error) throws Exception {
...
+    }
+
+    // If multi-zk mode is not enabled, create HelixZkClient with the provided zk address.
+    HelixZkClient.ZkClientConfig helixZkClientConfig = clientConfig.createHelixZkClientConfig();
+    HelixZkClient.ZkConnectionConfig helixZkConnectionConfig =
+        new HelixZkClient.ZkConnectionConfig(_zkAddress)
+            .setSessionTimeout(connectionConfig.getSessionTimeout());
+
+    return zkClientFactory.buildZkClient(helixZkConnectionConfig, helixZkClientConfig);
+  }
```
narendly: These method names are very confusing. You're creating HelixZkClient in buildRealmAwareZkClient() and creating RealmAwareZkClient in buildHelixZkClient()? Could we try to improve the naming for these methods so that it's easier to read? I think this could be refactored for readability. Thoughts?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -102,14 +103,13 @@ public AccessResult() {
...
-  private final HelixZkClient _zkClient;
+  private final RealmAwareZkClient _zkClient;
   // true if ZkBaseDataAccessor was instantiated with a HelixZkClient, false otherwise
   // This is used for close() to determine how ZkBaseDataAccessor should close the underlying
   // ZkClient
   private final boolean _usesExternalZkClient;
 
-  @Deprecated
-  public ZkBaseDataAccessor(HelixZkClient zkClient) {
+  public ZkBaseDataAccessor(RealmAwareZkClient zkClient) {
```
narendly: Shouldn't this be deprecated?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1799,4 +1867,73 @@ private boolean validateWeightForResourceConfig(ClusterConfig clusterConfig,
             clusterConfig));
     return true;
   }
+
+  public static class Builder {
```
dasahcc: Add TODO here. We need to refactor all the builders before merging to the master.

pkuwm: Added.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: I also had the same thought. Based on the original name and the current code structure, I could not find good names. Updated with my new idea. If you have better ideas, please suggest.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -90,33 +92,98 @@
 
 
 public class ZKHelixAdmin implements HelixAdmin {
+  private static final Logger LOG = LoggerFactory.getLogger(ZKHelixAdmin.class);
+
+
```
narendly: Nit: one empty line would suffice?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -117,7 +121,7 @@
   private final String _version;
   private int _reportLatency;
 
-  protected HelixZkClient _zkclient = null;
+  protected RealmAwareZkClient _zkclient = null;
```
narendly: Nit: I know we don't have to change this, but initializing to null is redundant.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1285,4 +1266,60 @@ public void handleSessionEstablishmentError(Throwable error) throws Exception {
...
+    if (_instanceType == InstanceType.ADMINISTRATOR) {
+      return buildSelectiveZkClient(SharedZkClientFactory.getInstance(), connectionConfig,
+          clientConfig);
+    }
+
+    return buildSelectiveZkClient(DedicatedZkClientFactory.getInstance(), connectionConfig,
+        clientConfig);
+  }
+
+  private RealmAwareZkClient buildSelectiveZkClient(HelixZkClientFactory zkClientFactory,
```
narendly: My suggestion is

JavaDoc:

Resolves what type of ZkClient this HelixManager should use based on whether MULTI_ZK_ENABLED System config is set or not. Two types of ZkClients are available: 1) If MULTI_ZK_ENABLED is set to true, we create a dedicated RealmAwareZkClient that provides full ZkClient functionalities and connects to the correct ZK by querying MetadataStoreDirectoryService. 2) Otherwise, we create a dedicated HelixZkClient which plainly connects to the ZK address given.

- Rename the method to `resolveZkClient()`

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
pkuwm: How could 2 empty lines be added...
Addressed.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1285,4 +1266,60 @@ public void handleSessionEstablishmentError(Throwable error) throws Exception {
   public Long getSessionStartTime() {
     return _sessionStartTime;
   }
+
+  private RealmAwareZkClient createSingleRealmZkClient() {
```
narendly: - Rename the method to `prepareAndGetZkClient()`

- Javadoc:

Prepares connection config and client config based on the internal parameters given to HelixManager in order to create a ZkClient instance to use. Note that a shared ZkClient instance will be created if connecting as an ADMINISTRATOR to minimize the cost of creating ZkConnections.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: Added. Thx.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: - `prepareAndGet` sounds like a method does 2 things. I prefer to name a method which does one thing. `get` is better than `prepareAndGet` as there is no need to care about "prepare" since the final purpose is "get". There is already one method `createClient()` so I prefer to keep the name.
- It is a private method and the logic is pretty straightforward. That's why I didn't consider adding comments. Anyway, since you suggested here, added.

pkuwm:

```
To make Helix Java APIs realm-aware, we need to make both ZKHelixAdmin and ZKHelixManager realm-aware. This commit adds a Builder to set client config and connection config for building realm-aware ZkClients underneath.
```

Issue #848, opened 2020-03-03
Add REST endpoint for updating RestConfig


mgao0:
We currently don't have REST endpoint to update RestConfig for clusters and we do updates manually which can be risky. Need to add REST endpoint PUT POST GET APIs to support RestConfig management.

PR #849, opened 2020-03-03
Add REST APIs for get, set, update RestConfig

mgao0:
### Issues

#848 

### Description

Added PUT to create, POST to upate, and GET to get the RestConfig for each cluster under CONFIGS.
Tested by unit tests.

### Tests

testSetRestConfig, testUpdateRestConfig

testCreateClusterRESTConfig, testUpdateClusterRESTConfig, testGetClusterRESTConfig

Tests run: 102, Failures: 0, Errors: 0, Skipped: 0

[INFO] BUILD SUCCESS
[INFO] Total time:  34.405 s
[INFO] Finished at: 2020-03-03T15:48:27-08:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
@@ -655,8 +655,7 @@ public class ConfigAccessor {
     // Create "/{clusterId}/CONFIGS/REST" if it does not exist
     String parentPath = HelixUtil.getZkParentPath(zkPath);
     if (!_zkClient.exists(parentPath)) {
-      return;
-    }
+      throw new HelixException("fail to delete REST config. cluster: " + clusterName + " is NOT setup.");    }
 
     ZKUtil.dropChildren(_zkClient, parentPath, new ZNRecord(clusterName));
   }
--- a/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
@@ -633,7 +633,7 @@ public class ConfigAccessor {
 
     // Create "/{clusterId}/CONFIGS/REST" if it does not exist
     String parentPath = HelixUtil.getZkParentPath(zkPath);
-    if (!_zkClient.exists(HelixUtil.getZkParentPath(zkPath))) {
+    if (!_zkClient.exists(parentPath)) {
       ZKUtil.createOrMerge(_zkClient, parentPath, new ZNRecord(parentPath), true, true);
     }
 
@@ -642,6 +642,23 @@ public class ConfigAccessor {
     } else {
       ZKUtil.createOrUpdate(_zkClient, zkPath, restConfig.getRecord(), true, true);
     }
+  }
+
+  public void deleteRESTConfig(String clusterName) {
+    if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
+      throw new HelixException("fail to delete REST config. cluster: " + clusterName + " is NOT setup.");
+    }
+
+    HelixConfigScope scope = new HelixConfigScopeBuilder(ConfigScopeProperty.REST).forCluster(clusterName).build();
+    String zkPath = scope.getZkPath();
+
+    // Create "/{clusterId}/CONFIGS/REST" if it does not exist
+    String parentPath = HelixUtil.getZkParentPath(zkPath);
+    if (!_zkClient.exists(parentPath)) {
+      return;
+    }
+
+    ZKUtil.dropChildren(_zkClient, parentPath, new ZNRecord(clusterName));
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
@@ -37,6 +37,7 @@ import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.RESTConfig;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.builder.HelixConfigScopeBuilder;
+import org.apache.helix.util.HelixUtil;
 import org.apache.helix.util.StringTemplate;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.api.client.HelixZkClient;
@@ -602,6 +603,45 @@ public class ConfigAccessor {
     }
 
     return new RESTConfig(record);
+  }
+
+  /**
+   * Set RestConfig of a given cluster
+   * @param clusterName the cluster id
+   * @param restConfig the RestConfig to be set to the cluster
+   */
+  public void setRESTConfig(String clusterName, RESTConfig restConfig) {
+    updateRESTConfig(clusterName, restConfig, true);
+  }
+
+  /**
+   * Update RestConfig of a given cluster
+   * @param clusterName the cluster id
+   * @param restConfig the new RestConfig to be set to the cluster
+   */
+  public void updateRESTConfig(String clusterName, RESTConfig restConfig) {
+    updateRESTConfig(clusterName, restConfig, false);
+  }
+
+  private void updateRESTConfig(String clusterName, RESTConfig restConfig, boolean overwrite) {
+    if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
+      throw new HelixException("fail to update REST config. cluster: " + clusterName + " is NOT setup.");
+    }
+
+    HelixConfigScope scope = new HelixConfigScopeBuilder(ConfigScopeProperty.REST).forCluster(clusterName).build();
+    String zkPath = scope.getZkPath();
+
+    // Create "/{clusterId}/CONFIGS/REST" if it does not exist
+    String parentPath = HelixUtil.getZkParentPath(zkPath);
+    if (!_zkClient.exists(HelixUtil.getZkParentPath(zkPath))) {
+      ZKUtil.createOrMerge(_zkClient, parentPath, new ZNRecord(parentPath), true, true);
+    }
+
+    if (overwrite) {
+      ZKUtil.createOrReplace(_zkClient, zkPath, restConfig.getRecord(), true);
+    } else {
+      ZKUtil.createOrUpdate(_zkClient, zkPath, restConfig.getRecord(), true, true);
+    }
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
@@ -625,7 +625,7 @@ public class ConfigAccessor {
 
   private void updateRESTConfig(String clusterName, RESTConfig restConfig, boolean overwrite) {
     if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
-      throw new HelixException("fail to update REST config. cluster: " + clusterName + " is NOT setup.");
+      throw new HelixException("Fail to update REST config. cluster: " + clusterName + " is NOT setup.");
     }
 
     HelixConfigScope scope = new HelixConfigScopeBuilder(ConfigScopeProperty.REST).forCluster(clusterName).build();
@@ -646,16 +646,16 @@ public class ConfigAccessor {
 
   public void deleteRESTConfig(String clusterName) {
     if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
-      throw new HelixException("fail to delete REST config. cluster: " + clusterName + " is NOT setup.");
+      throw new HelixException("Fail to delete REST config. cluster: " + clusterName + " is NOT setup.");
     }
 
     HelixConfigScope scope = new HelixConfigScopeBuilder(ConfigScopeProperty.REST).forCluster(clusterName).build();
     String zkPath = scope.getZkPath();
 
-    // Create "/{clusterId}/CONFIGS/REST" if it does not exist
+    // Check if "/{clusterId}/CONFIGS/REST" exists
     String parentPath = HelixUtil.getZkParentPath(zkPath);
     if (!_zkClient.exists(parentPath)) {
-      throw new HelixException("fail to delete REST config. cluster: " + clusterName + " is NOT setup.");    }
+      throw new HelixException("Fail to delete REST config. cluster: " + clusterName + " does not have a rest config.");    }
 
     ZKUtil.dropChildren(_zkClient, parentPath, new ZNRecord(clusterName));
   }
--- a/helix-core/src/test/java/org/apache/helix/TestConfigAccessor.java
+++ b/helix-core/src/test/java/org/apache/helix/TestConfigAccessor.java
@@ -33,6 +33,7 @@ import org.apache.helix.model.builder.HelixConfigScopeBuilder;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
+
 public class TestConfigAccessor extends ZkUnitTestBase {
   @Test
   public void testBasic() throws Exception {
@@ -116,9 +117,8 @@ public class TestConfigAccessor extends ZkUnitTestBase {
         "should be [HELIX_ENABLED, HELIX_ENABLED_TIMESTAMP, HELIX_HOST, HELIX_PORT, participantConfigKey]");
     Assert.assertEquals(keys.get(4), "participantConfigKey");
 
-    keys =
-        configAccessor.getKeys(ConfigScopeProperty.PARTITION, clusterName, "testResource",
-            "testPartition");
+    keys = configAccessor
+        .getKeys(ConfigScopeProperty.PARTITION, clusterName, "testResource", "testPartition");
     Assert.assertEquals(keys.size(), 1, "should be [partitionConfigKey]");
     Assert.assertEquals(keys.get(0), "partitionConfigKey");
 
@@ -166,7 +166,6 @@ public class TestConfigAccessor extends ZkUnitTestBase {
     configAccessor.close();
     configAccessorZkAddr.close();
     System.out.println("END " + clusterName + " at " + new Date(System.currentTimeMillis()));
-
   }
 
   // HELIX-25: set participant Config should check existence of instance
@@ -186,8 +185,8 @@ public class TestConfigAccessor extends ZkUnitTestBase {
 
     try {
       configAccessor.set(participantScope, "participantConfigKey", "participantConfigValue");
-      Assert
-          .fail("Except fail to set participant-config because participant: localhost_12918 is not added to cluster yet");
+      Assert.fail(
+          "Except fail to set participant-config because participant: localhost_12918 is not added to cluster yet");
     } catch (HelixException e) {
       // OK
     }
@@ -196,8 +195,8 @@ public class TestConfigAccessor extends ZkUnitTestBase {
     try {
       configAccessor.set(participantScope, "participantConfigKey", "participantConfigValue");
     } catch (Exception e) {
-      Assert
-          .fail("Except succeed to set participant-config because participant: localhost_12918 has been added to cluster");
+      Assert.fail(
+          "Except succeed to set participant-config because participant: localhost_12918 has been added to cluster");
     }
 
     String participantConfigValue = configAccessor.get(participantScope, "participantConfigKey");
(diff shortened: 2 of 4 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -511,6 +512,89 @@ public Response removeClusterStateModelDefinition(@PathParam("clusterId") String
...
+    } catch (Exception ex) {
+      _logger.error(
+          "Failed to update rest config, cluster " + clusterId + " new config: " + content
+              + ", Exception: " + ex);
+      return serverError(ex);
+    }
+    return OK();
+  }
+
+  @GET
```
zhangmeng916: We may also need a "DELETE" API. You can double check.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -511,6 +512,89 @@ public Response removeClusterStateModelDefinition(@PathParam("clusterId") String
...
+
+    RESTConfig config = new RESTConfig(record);
+    ConfigAccessor configAccessor = getConfigAccessor();
+    try {
+      configAccessor.setRESTConfig(clusterId, config);
+    } catch (HelixException ex) {
+      return notFound(ex.getMessage());
+    } catch (Exception ex) {
+      _logger.error(
+          "Failed to create rest config, cluster " + clusterId + " new config: " + content
```
zhangmeng916: try "code reformat" to see whether it'll look better.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -511,6 +512,89 @@ public Response removeClusterStateModelDefinition(@PathParam("clusterId") String
...
 
+  @PUT
+  @Path("{clusterId}/restconfig")
+  public Response createClusterRESTConfig(@PathParam("clusterId") String clusterId,
+      String content) {
+    ZNRecord record;
+    try {
+      record = toZNRecord(content);
+    } catch (IOException e) {
+      _logger.error("Failed to deserialize user's input " + content + ", Exception: " + e);
```
pkuwm: I suggest using `{}` to make advantage of sl4j logging.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -605,6 +606,45 @@ public RESTConfig getRESTConfig(String clusterName) {
...
+  private void updateRESTConfig(String clusterName, RESTConfig restConfig, boolean overwrite) {
+    if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
+      throw new HelixException("fail to update REST config. cluster: " + clusterName + " is NOT setup.");
+    }
+
+    HelixConfigScope scope = new HelixConfigScopeBuilder(ConfigScopeProperty.REST).forCluster(clusterName).build();
+    String zkPath = scope.getZkPath();
+
+    // Create "/{clusterId}/CONFIGS/REST" if it does not exist
+    String parentPath = HelixUtil.getZkParentPath(zkPath);
```
jiajunwang: I see "HelixUtil.getZkParentPath(zkPath)" twice. I guess you are trying to use parentPath.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -605,6 +606,45 @@ public RESTConfig getRESTConfig(String clusterName) {
...
+    if (!_zkClient.exists(HelixUtil.getZkParentPath(zkPath))) {
+      ZKUtil.createOrMerge(_zkClient, parentPath, new ZNRecord(parentPath), true, true);
+    }
+
+    if (overwrite) {
```
jiajunwang: Actually, if it is "overwrite", there is no need to check existence. Just call createOrReplace().
If it is not "overwrite", there is no need to check existence either, just call createOrUpdate().

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -511,6 +512,89 @@ public Response removeClusterStateModelDefinition(@PathParam("clusterId") String
...
+    if (!record.getId().equals(clusterId)) {
+      return badRequest("ID does not match the cluster name in input!");
+    }
+
+    RESTConfig config = new RESTConfig(record);
+    ConfigAccessor configAccessor = getConfigAccessor();
+    try {
+      configAccessor.setRESTConfig(clusterId, config);
+    } catch (HelixException ex) {
+      return notFound(ex.getMessage());
```
jiajunwang: Could it because of write failure? I'm not sure if the accessor could throw any HelixException as well?
I know it probably lists details in the message, but the response type would be notfound in any case, which is not good.

mgao0: I checked, all the exceptions are caught in ZKUtil.createOrMerge/Replace/Update, and only log the errors, including the write failure or other types of errors. Only one case would throw HelixException which is ZKUtil.isClusterSetup(clusterName, _zkClient) returns false, so I used the notFound response type.  Please let me know if you still think it is an issue and we should fix it some other way. Thanks.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
mgao0: You are right. I'm trying to use the parentPath. I'm trying to resolve the problem that when we want to update the rest config in the path "/{clusterId}/CONFIGS/REST{clusterId}", the parent path "/{clusterId}/CONFIGS/REST" doesn't exist yet, so I would create the parentPath first.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
mgao0: Please see the comment above, I checked the existence of parent path, not current path.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: I searched for sl4j logging, but couldn't find materials for what you referred to as using{}, could you please provide an example?

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
jiajunwang: What I meant is that line 637 could be:
_zkClient.exists(parentPath)

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: I think it's because I used the old style sheet, will update to the new style sheet. Thanks.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
jiajunwang: I see.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
jiajunwang: It's better to have a more precise exception for notFound error. Or, if we use HelixException, we should use a more generic error accordingly.
Since this is a common issue, I'm fine if you add a TODO here. We can address this later.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: I added the "DELETE" API, please help me review again. Thank you.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
zhangmeng916: I think he meant something like this:
 logger.info("Initialized {} pipeline", Pipeline.Type.TASK.name());

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -511,6 +512,107 @@ public Response removeClusterStateModelDefinition(@PathParam("clusterId") String
...
+      _logger.error("Failed to create rest config, cluster " + clusterId + " new config: " + content
+          + ", Exception: " + ex);
+      return serverError(ex);
+    }
+    return OK();
+  }
+
+  @POST
+  @Path("{clusterId}/restconfig")
+  public Response updateClusterRESTConfig(@PathParam("clusterId") String clusterId,
```
dasahcc: Can you follow the style of ClusterConfig? We have entry update/delete operation with URL param.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: I see. Thank you for explanation @zhangmeng916. @pkuwm Here I'm following the format that is used by other methods in this class. But the way you suggested does provide a clearer and easier way to form the message. Later we can have a separate PR to change all the log messages in Helix to adopt the sl4j {} function.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: I have updated the PR. Can you review again? Thanks.

dasahcc: I think you have missunderstanding about delete the Config and delete entry of the Config.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -605,6 +606,62 @@ public RESTConfig getRESTConfig(String clusterName) {
...
+  /**
+   * Update RestConfig of a given cluster
+   * @param clusterName the cluster id
+   * @param restConfig the new RestConfig to be set to the cluster
+   */
+  public void updateRESTConfig(String clusterName, RESTConfig restConfig) {
+    updateRESTConfig(clusterName, restConfig, false);
+  }
+
+  private void updateRESTConfig(String clusterName, RESTConfig restConfig, boolean overwrite) {
```
dasahcc: Is this necessary? We should have update function with scope arugment,right?

mgao0: I've adopted the pattern for ClusterConfig here. The overwrite is to be called to distinguish set and update operations. The RESTConfig instance passed in gives us the information to be used to set/update the rest config, and the scope can be derived from information in RESTConfig in the implementation.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -511,6 +512,125 @@ public Response removeClusterStateModelDefinition(@PathParam("clusterId") String
...
+    try {
+      record = toZNRecord(content);
+    } catch (IOException e) {
+      _logger.error("Failed to deserialize user's input " + content + ", Exception: " + e);
+      return badRequest("Input is not a valid ZNRecord!");
+    }
+
+    if (!record.getId().equals(clusterId)) {
+      return badRequest("ID does not match the cluster name in input!");
+    }
```
narendly: Could we double-check if the ID for the restConfig is supposed to match up with the cluster name?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -511,6 +512,125 @@ public Response removeClusterStateModelDefinition(@PathParam("clusterId") String
     return OK();
   }
 
+  @PUT
+  @Path("{clusterId}/restconfig")
+  public Response createClusterRESTConfig(@PathParam("clusterId") String clusterId,
```
narendly: Nit: is there a reason we don't just say createRestConfig? Cluster seems a little redundant?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -511,6 +512,125 @@ public Response removeClusterStateModelDefinition(@PathParam("clusterId") String
...
+      return serverError(ex);
+    }
+    return OK();
+  }
+
+  @GET
+  @Path("{clusterId}/restconfig")
+  public Response getClusterRESTConfig(@PathParam("clusterId") String clusterId) {
+    ConfigAccessor accessor = getConfigAccessor();
+    RESTConfig config = null;
```
narendly: Initializing to null is redundant.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: This check is not imposed on rest config id, this is to check the if the id of ZNRecord user uses to update the rest config matches the cluster name. I'm not sure what our users are doing right now, but this could prevent accidentally update the wrong cluster.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: I've changed the method names.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: I initialize the config to null mainly to check later if the accessor got the config successfully. If there is an Exception, we catch these exceptions, so by comparing if the config is null we know if there is error in fetching the information.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: Since it is a not necessary check. I removed it so it is easier for users to use.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
dasahcc: There are lots of common code for updateRESTConfig, updateClusterConfig and so on. I am fine with this PR. But can you have another PR to abstract the logic to a private method to reduce the redundant code?

mgao0: Ok. I'll create another PR to abstract the logic.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
narendly: Please read https://docs.oracle.com/javase/specs/jls/se7/jls7.pdf

Object references not initialized default to null. So initializing to null is redundant.

narendly:
Please fix/improve the code where change is possible. The fact that existing code is doing it in a certain suboptimal way doesn't justify us making the same mistake. I am fine, however, with creating a separate issue for refactoring work that will involve changes in existing logic though.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -511,6 +512,121 @@ public Response removeClusterStateModelDefinition(@PathParam("clusterId") String
...
+  }
+
+  @DELETE
+  @Path("{clusterId}/restconfig")
+  public Response deleteRESTConfig(@PathParam("clusterId") String clusterId) {
+    ConfigAccessor accessor = getConfigAccessor();
+    try {
+      accessor.deleteRESTConfig(clusterId);
+    } catch (HelixException ex) {
+      _logger.info("Failed to delete rest config for cluster " + clusterId
```
pkuwm: Failed to delete but finally returning OK response? Would 404 be more appropriate?

helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
```diff
@@ -576,6 +577,70 @@ public void testEnableWagedRebalanceForAllResources() {
...
+    restConfigRest.set(RESTConfig.SimpleFields.CUSTOMIZED_HEALTH_URL, null);
+    entity = Entity.entity(OBJECT_MAPPER.writeValueAsString(restConfigRest.getRecord()),
+        MediaType.APPLICATION_JSON_TYPE);
+    post("clusters/" + cluster + "/restconfig", ImmutableMap.of("command", Command.delete.name()),
+        entity, Response.Status.OK.getStatusCode());
+    restConfigZK = _configAccessor.getRESTConfig(cluster);
+    Assert.assertEquals(restConfigZK, new RESTConfig(cluster),
+        "rest config from response: " + new RESTConfig(cluster) + " vs rest config actually: "
+            + restConfigZK);
+    System.out.println("End test :" + TestHelper.getTestMethodName());
```
pkuwm: Is the test already ended?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: I tried it, the code doesn't compile after removing "=null". I agree that initialization to null is functional redundant, but it benefits the style check and readability of the code, so I would keep it here.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
mgao0: I was considering returns OK would be fine since what you want to delete is already not there. But on second thought, return NOT_FOUND probably is more intuitive for the users. Changing it to NOT_FOUND.

helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
mgao0: Added one more test after the first draft. Moving it down.

mgao0:
> Please fix/improve the code where change is possible. The fact that existing code is doing it in a certain suboptimal way doesn't justify us making the same mistake. I am fine, however, with creating a separate issue for refactoring work that will involve changes in existing logic though.

Rolling out functional service for our customers at the earliest to unblock them is our top priority. We proved the current way works fine, and the new way although is clearer in code, it bears the risk of breaking things for our customers. Therefore, I considered it sub-priority for us. Also, refactoring for one thing all together is easier for us to track and debug if anything goes wrong in the future.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -604,6 +605,61 @@ public RESTConfig getRESTConfig(String clusterName) {
     return new RESTConfig(record);
   }
 
+  /**
```
narendly: If there is a change to config accessor, should we add tests in TestConfigAccessor as well?

mgao0: There are tests for rest config covering set and update in TestConfigAccessor. I just added a few more cases for delete.

helix-core/src/test/java/org/apache/helix/TestConfigAccessor.java
```diff
@@ -204,4 +207,56 @@ public void testSetNonexistentParticipantConfig() throws Exception {
...
+    HelixConfigScope scope = new HelixConfigScopeBuilder(ConfigScopeProperty.REST).forCluster(clusterName).build();
+    Assert.assertNull(configAccessor.getRESTConfig(clusterName));
+
+    RESTConfig restConfig = new RESTConfig(clusterName);
+    restConfig.set(RESTConfig.SimpleFields.CUSTOMIZED_HEALTH_URL, "TEST_URL");
+    configAccessor.setRESTConfig(clusterName, restConfig);
+    Assert.assertEquals(restConfig, configAccessor.getRESTConfig(clusterName));
+  }
+
+  @Test (expectedExceptions = HelixException.class)
```
narendly: Could you mark it clearly where the exception should be expected? Because it's not often easy to tell, I prefer the style of 

```
try {
//your logic
Assert.fail("Exception expected!");
catch (exception) {
// Exception expected..
}
```

mgao0:
This PR is ready to be merged, approved by @narendly .
Final commit message: 
Title: Add REST APIs for get, set, update RestConfig
Body: Added endpoints PUT to create, POST to update, and GET to get, DELETE to delete the RestConfig for clusters in ClusterAccessor. Added corresponding methods in ConfigAccessor to be used by ClusterAccessor.

Issue #850, opened 2020-03-04
Add customized view computation to Helix generic controller


zhangmeng916:
Helix will support customized state input from its users and the aggregation of customized state to customized view. This would require Helix generic controller to have an extra stage to computer customized view from customized state.

PR #851, opened 2020-03-04
Modify Helix generic controller to include new stage for customized view aggregation

zhangmeng916:
### Issues

(#850 )

### Description
Helix will support customized state input from its users and the aggregation of customized state to customized view. This PR implement the computation logic for customized view aggregation and also modify Helix generic controller to include an extra stage to compute customized view from customized state.

### Tests
TestCustomizedStateComputationStage
TestCustomizedViewStage
TestComputeAndCleanupCustomizedView

[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/api/listeners/CustomizedStateRootChangeListener.java
+++ b/helix-core/src/main/java/org/apache/helix/api/listeners/CustomizedStateRootChangeListener.java
@@ -19,8 +19,9 @@ package org.apache.helix.api.listeners;
  * under the License.
  */
 
-import org.apache.helix.NotificationContext;
+import java.util.List;
 
+import org.apache.helix.NotificationContext;
 
 /**
  * Interface to implement to respond to changes in the root path of customized state
@@ -30,7 +31,9 @@ public interface CustomizedStateRootChangeListener {
   /**
    * Invoked when root path customized state changes
    * @param instanceName name of the instance whose state changed
+   * @param customizedStateTypes the state types under the root
    * @param changeContext the change event and state
    */
-  void onCustomizedStateRootChange(String instanceName, NotificationContext changeContext);
+  void onCustomizedStateRootChange(String instanceName,
+      List<String> customizedStateTypes, NotificationContext changeContext);
 }
--- a/helix-core/src/main/java/org/apache/helix/api/listeners/CustomizedStateRootChangeListener.java
+++ b/helix-core/src/main/java/org/apache/helix/api/listeners/CustomizedStateRootChangeListener.java
@@ -19,21 +19,18 @@ package org.apache.helix.api.listeners;
  * under the License.
  */
 
-import java.util.List;
 import org.apache.helix.NotificationContext;
-import org.apache.helix.model.CustomizedState;
+
 
 /**
- * Interface to implement to respond to changes in the customized state
+ * Interface to implement to respond to changes in the root path of customized state
  */
-public interface CustomizedStateChangeListener {
+public interface CustomizedStateRootChangeListener {
 
   /**
-   * Invoked when customized state changes
+   * Invoked when root path customized state changes
    * @param instanceName name of the instance whose state changed
-   * @param customizedStatesInfo a list of the customized states
    * @param changeContext the change event and state
    */
-  void onCustomizedStateChange(String instanceName, List<CustomizedState> customizedStatesInfo,
-      NotificationContext changeContext);
+  void onCustomizedStateRootChange(String instanceName, NotificationContext changeContext);
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -822,12 +822,12 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
     logger.info("START: GenericClusterController.onCustomizedStateRootChange()");
     HelixManager manager = changeContext.getManager();
     Builder keyBuilder = new Builder(manager.getClusterName());
-   if (customizedStateTypes.isEmpty()) {
-     customizedStateTypes =
-         manager.getHelixDataAccessor().getChildNames(
-             manager.getHelixDataAccessor().keyBuilder().customizedStatesRoot(instanceName));
-   }
+    if (customizedStateTypes.isEmpty()) {
+      customizedStateTypes = manager.getHelixDataAccessor()
+          .getChildNames(keyBuilder.customizedStatesRoot(instanceName));
+    }
 
+    // TODO: remove the synchronization here once we move this update into dataCache.
     synchronized (_lastSeenCustomizedStateTypes) {
       Set<String> lastSeenCustomizedStateTypes = _lastSeenCustomizedStateTypes.get();
       for (String customizedState : customizedStateTypes) {
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -137,6 +138,8 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
 
   final AtomicReference<Map<String, LiveInstance>> _lastSeenInstances;
   final AtomicReference<Map<String, LiveInstance>> _lastSeenSessions;
+
+  final AtomicReference<Set<String>> _lastSeenCustomizedStateTypes;
 
   // By default not reporting status until controller status is changed to activate
   // TODO This flag should be inside ClusterStatusMonitor. When false, no MBean registering.
@@ -545,6 +548,7 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
     _taskRegistry = taskRegistry;
     _lastSeenInstances = new AtomicReference<>();
     _lastSeenSessions = new AtomicReference<>();
+    _lastSeenCustomizedStateTypes = new AtomicReference<>();
     _clusterName = clusterName;
     _lastPipelineEndTimestamp = TopStateHandoffReportStage.TIMESTAMP_NOT_RECORDED;
     _clusterStatusMonitor = new ClusterStatusMonitor(_clusterName);
@@ -813,23 +817,36 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
 
   @Override
   @PreFetch(enabled = false)
-  public void onCustomizedStateRootChange(String instanceName, NotificationContext changeContext) {
+  public void onCustomizedStateRootChange(String instanceName, List<String> customizedStateTypes,
+      NotificationContext changeContext) {
     logger.info("START: GenericClusterController.onCustomizedStateRootChange()");
-    notifyCaches(changeContext, ChangeType.CUSTOMIZED_STATE_ROOT);
     HelixManager manager = changeContext.getManager();
-    List<String> customizedStateTypes =
-        manager.getHelixDataAccessor().getChildNames(
-        manager.getHelixDataAccessor().keyBuilder().customizedStatesRoot(instanceName));
+    Builder keyBuilder = new Builder(manager.getClusterName());
 
-    for (String customizedState : customizedStateTypes) {
-      try {
-        manager.addCustomizedStateChangeListener(this, instanceName, customizedState);
-        logger.info(
-            manager.getInstanceName() + " added customized state listener for " + instanceName
-                + ", listener: " + this);
-      } catch (Exception e) {
-        logger.error("Fail to add customized state listener for instance: " + instanceName, e);
+    synchronized (_lastSeenCustomizedStateTypes) {
+      Set<String> lastSeenCustomizedStateTypes = _lastSeenCustomizedStateTypes.get();
+      for (String customizedState : customizedStateTypes) {
+        try {
+          if (lastSeenCustomizedStateTypes == null || !lastSeenCustomizedStateTypes
+              .contains(customizedState)) {
+            manager.addCustomizedStateChangeListener(this, instanceName, customizedState);
+            logger.info(
+                manager.getInstanceName() + " added customized state listener for " + instanceName
+                    + ", listener: " + this);
+          }
+        } catch (Exception e) {
+          logger.error("Fail to add customized state listener for instance: " + instanceName, e);
+        }
       }
+
+      for (String previousCustomizedState : lastSeenCustomizedStateTypes) {
+        if (!customizedStateTypes.contains(previousCustomizedState)) {
+          manager.removeListener(keyBuilder.customizedStates(instanceName, previousCustomizedState),
+              this);
+        }
+      }
+
+      _lastSeenCustomizedStateTypes.set(new HashSet<>(customizedStateTypes));
     }
   }
 
@@ -1333,19 +1350,5 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
 
     eventThread.setDaemon(true);
     eventThread.start();
-  }
-
-  private List<String> getEnabledCustomizedStates(HelixManager manager) {
-    CustomizedStateConfig customizedStateConfig =
-        manager.getConfigAccessor().getCustomizedStateConfig(_clusterName);
-    if (customizedStateConfig != null) {
-      return customizedStateConfig.getAggregationEnabledTypes();
-    }
-    return Collections.emptyList();
-  }
-
-  private void addCustomizedStateListeners(HelixManager manager, String customizedState,
-      String instance) {
-
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -468,10 +468,13 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
       registry
           .register(ClusterEventType.OnDemandRebalance, dataRefresh, autoExitMaintenancePipeline,
               dataPreprocess, externalViewPipeline, rebalancePipeline);
+      // TODO: We now include rebalance pipeline in customized state change for correctness.
+      // However, it is not efficient, and we should improve this by splitting the pipeline or
+      // controller roles to multiple hosts.
       registry.register(ClusterEventType.CustomizedStateChange, dataRefresh, dataPreprocess,
-          externalViewPipeline, customizedViewPipeline, rebalancePipeline);
-      registry.register(ClusterEventType.CustomizeStateAggregationConfigChange, dataRefresh,
-          dataPreprocess, externalViewPipeline, customizedViewPipeline, rebalancePipeline);
+          customizedViewPipeline, rebalancePipeline);
+      registry.register(ClusterEventType.CustomizeStateConfigChange, dataRefresh, dataPreprocess,
+          customizedViewPipeline, rebalancePipeline);
       return registry;
     }
   }
@@ -966,7 +969,7 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
         "START: GenericClusterController.onCustomizedStateConfigChange() for cluster "
             + _clusterName);
     notifyCaches(context, ChangeType.CUSTOMIZED_STATE_CONFIG);
-    pushToEventQueues(ClusterEventType.CustomizeStateAggregationConfigChange, context,
+    pushToEventQueues(ClusterEventType.CustomizeStateConfigChange, context,
         Collections.<String, Object> emptyMap());
     logger.info(
         "END: GenericClusterController.onCustomizedStateConfigChange() for cluster "
@@ -1323,7 +1326,7 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
     if (customizedStateConfig != null) {
       return customizedStateConfig.getAggregationEnabledTypes();
     }
-    return new ArrayList<>();
+    return Collections.emptyList();
   }
 
   private void addCustomizedStateListeners(HelixManager manager, String customizedState,
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -35,6 +35,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 
 import com.google.common.collect.Sets;
 import org.I0Itec.zkclient.exception.ZkInterruptedException;
@@ -47,6 +48,8 @@ import org.apache.helix.api.exceptions.HelixMetaDataAccessException;
 import org.apache.helix.api.listeners.ClusterConfigChangeListener;
 import org.apache.helix.api.listeners.ControllerChangeListener;
 import org.apache.helix.api.listeners.CurrentStateChangeListener;
+import org.apache.helix.api.listeners.CustomizedStateChangeListener;
+import org.apache.helix.api.listeners.CustomizedStateConfigChangeListener;
 import org.apache.helix.api.listeners.IdealStateChangeListener;
 import org.apache.helix.api.listeners.InstanceConfigChangeListener;
 import org.apache.helix.api.listeners.LiveInstanceChangeListener;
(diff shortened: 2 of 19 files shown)


helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -96,13 +99,15 @@
 
   // Special caches
   private CurrentStateCache _currentStateCache;
+  private CustomizedStateCache _customizedStateCache;
```
jiajunwang: We aggregate _currentStateCache in addition to externalView because it is used in the rebalance pipeline. For customize state though, the customized view is the only thing we need. Can we just combine these 2 refresh workflow and keep one customized View map in the cache?

helix-core/src/main/java/org/apache/helix/common/caches/CustomizedStateCache.java
```diff
@@ -49,11 +49,15 @@ public CustomizedStateCache(ControlContextProvider contextProvider) {
...
-            .getListFields().get(
-            CustomizedStateAggregationConfig.CustomizedStateAggregationProperty.AGGREGATION_ENABLED_TYPES
-                .name()));
+
+    Set<String> restrictedKeys = new HashSet<>();
+    if (accessor.getProperty(accessor.keyBuilder().customizedStateAggregationConfig()) != null) {
+      restrictedKeys = new HashSet<>(
+          accessor.getProperty(accessor.keyBuilder().customizedStateAggregationConfig()).getRecord()
+              .getListFields().get(
+              CustomizedStateAggregationConfig.CustomizedStateAggregationProperty.AGGREGATION_ENABLED_TYPES
```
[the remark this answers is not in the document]
zhangmeng916: yes.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -96,13 +99,15 @@
...
   private InstanceMessagesCache _instanceMessagesCache;
 
   // Other miscellaneous caches
   private Map<String, Long> _instanceOfflineTimeMap;
   private Map<String, Map<String, String>> _idealStateRuleMap;
   private Map<String, Map<String, Set<String>>> _disabledInstanceForPartitionMap = new HashMap<>();
   private Set<String> _disabledInstanceSet = new HashSet<>();
+  private Set<String> _aggregationEnabledTypes = new HashSet<>();
```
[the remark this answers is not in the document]
zhangmeng916: I've removed it, it used to be used by customized state cache, but now we've changed that cache and access the customized state config directly in the cache. So we do not need this variable.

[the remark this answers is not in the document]
zhangmeng916: I added the aggregationEnableType back, as it will break some test. The main reason is that if we put the accessing to ZK inside customized state cache, it will break some assumption in the mock zk Helix accessor about refresh logic, and the calculation for "config" type would be wrong.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -286,6 +292,27 @@ private void refreshResourceConfig(final HelixDataAccessor accessor,
     }
   }
 
+  private void refreshCustomizedStateAggregationConfig(final HelixDataAccessor accessor,
```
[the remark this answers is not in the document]
zhangmeng916: As the customized state config is a cluster level config, and following the current pattern, I feel it is more natural to put it together with cluster config, resource config, etc.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1072,6 +1152,25 @@ protected void checkLiveInstancesObservation(List<LiveInstance> liveInstances,
         }
       }
 
+      for (String instance : curInstances.keySet()) {
```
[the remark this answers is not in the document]
zhangmeng916: The reason we need to add listeners in both functions is that only after customized state config is updated, the paths of individual states can be decided and listeners can be added to them. For example, after we add listener in live instance change, if we do not add listener in customized state config, we will miss the paths of the new customized state, and all the changes on this path will be missed until next live instance change. This is not preferred.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1072,6 +1152,25 @@ protected void checkLiveInstancesObservation(List<LiveInstance> liveInstances,
...
+        if (lastInstances == null || !lastInstances.containsKey(instance)) {
+          try {
```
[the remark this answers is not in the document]
zhangmeng916: Please see the updated code.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
zhangmeng916: So _customizedStateCache is used by CustomizedStateComputation stage, which happened before customized view aggregation stage. Therefore, we do need this cache to avoid access to ZK during computation.

helix-core/src/main/java/org/apache/helix/common/caches/CustomizedStateCache.java
```diff
@@ -46,14 +46,9 @@ public CustomizedStateCache(ControlContextProvider contextProvider) {
 
   @Override
   protected Set<PropertyKey> PopulateParticipantKeys(HelixDataAccessor accessor,
-      Map<String, LiveInstance> liveInstanceMap) {
+      Map<String, LiveInstance> liveInstanceMap, Set<String> restrictedKeys) {
```
jiajunwang: I think we discussed to put the restrictedKeys to the CustomizedStateCache constructor as a final field. So you don't need to add this to these methods? The downside of the current design is that the current state cache will have to keep the unnecessary input.

helix-core/src/main/java/org/apache/helix/common/caches/ParticipantStateCache.java
```diff
@@ -57,10 +57,11 @@ public ParticipantStateCache(ControlContextProvider controlContextProvider) {
    * @param liveInstanceMap map of all liveInstances in cluster
    * @return
    */
-  public boolean refresh(HelixDataAccessor accessor, Map<String, LiveInstance> liveInstanceMap) {
+  public boolean refresh(HelixDataAccessor accessor, Map<String, LiveInstance> liveInstanceMap,
+      Set<String> restrictedKeys) {
```
jiajunwang: Following the previous comment, I think this restrictedKeys can be a field only for the customizedStateCache class. As long as we input it in the constructor, we will be able to read it in the customizedStateCache class method. For this parent class, it does no need to know it.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -122,7 +124,7 @@
   // TODO: make this be per _manager or per _listener instaed of per callbackHandler -- Lei
   private CallbackProcessor _batchCallbackProcessor;
   private boolean _watchChild = true; // Whether we should subscribe to the child znode's data
-                                      // change.
+  // change.
```
pkuwm: Maybe no need to format the whole file? Or you can move this comment before the code:
// comment
code;

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregateStage.java
```diff
@@ -0,0 +1,171 @@
...
+public class CustomizedViewAggregateStage extends AbstractAsyncBaseStage {
+  private static Logger LOG = LoggerFactory.getLogger(CustomizedViewAggregateStage.class);
+
+  @Override
+  public AsyncWorkerType getAsyncWorkerType() {
+    return AsyncWorkerType.CustomizedStateViewComputeWorker;
+  }
+
+  @Override
+  public void execute(final ClusterEvent event) throws Exception {
```
pkuwm: The implementation in this method is too long. Can we try to split it into several private methods like:
```
removeStaleTypes()
updateCustomizedView()
```

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -118,7 +118,12 @@ public synchronized void refresh(HelixDataAccessor accessor) {
     if (propertyRefreshed.contains(HelixConstants.ChangeType.IDEAL_STATE)
         || propertyRefreshed.contains(HelixConstants.ChangeType.LIVE_INSTANCE)
         || propertyRefreshed.contains(HelixConstants.ChangeType.INSTANCE_CONFIG)
+<<<<<<< HEAD
```
pkuwm: Fix conflicts :)

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1233,4 +1316,24 @@ private void initPipeline(Thread eventThread, BaseControllerDataProvider cache)
...
     eventThread.start();
   }
+
+  private List<String> getEnabledCustomizedStates(HelixManager manager) {
+    CustomizedStateConfig customizedStateConfig =
+        manager.getConfigAccessor().getCustomizedStateConfig(_clusterName);
+    if (customizedStateConfig != null) {
+      return customizedStateConfig.getAggregationEnabledTypes();
+    }
+    return new ArrayList<>();
```
pkuwm: `Collections.singletonList()` may be more lightweight as a returning empty list?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
zhangmeng916: Have moved it to resource provider.

helix-core/src/test/java/org/apache/helix/controller/stages/TestCustomizedViewStage.java
```diff
@@ -0,0 +1,105 @@
...
+    Assert.assertEquals(cache.getCustomizedViewCacheMap().values(),
+        accessor.getChildValues(accessor.keyBuilder().customizedViews()));
+
+    // Assure there is no customized view got updated
+    List<CustomizedView> oldCustomizedViews =
+        accessor.getChildValues(accessor.keyBuilder().customizedViews());
+    runStage(event, customizedViewComputeStage);
+    List<CustomizedView> newCustomizedViews =
+        accessor.getChildValues(accessor.keyBuilder().customizedViews());
+    Assert.assertEquals(oldCustomizedViews, newCustomizedViews);
```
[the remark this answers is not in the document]
zhangmeng916: Discussed in another PR, and will keep it as a list for now as we're verifying based on index.

helix-core/src/main/java/org/apache/helix/manager/zk/ControllerManagerHelper.java
```diff
@@ -102,6 +104,7 @@ public void removeListenersFromController(GenericHelixController controller) {
     _manager.removeListener(keyBuilder.clusterConfig(), controller);
     _manager.removeListener(keyBuilder.resourceConfigs(), controller);
     _manager.removeListener(keyBuilder.instanceConfigs(), controller);
+    _manager.removeListener(keyBuilder.customizedStateConfig(), controller);
```
jiajunwang: Please follow the reverse order in which the listeners are added.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -743,6 +743,8 @@ private void createZKPaths(String clusterName) {
     _zkClient.createPersistent(path);
     path = PropertyPathBuilder.resourceConfig(clusterName);
     _zkClient.createPersistent(path);
+    path = PropertyPathBuilder.customizedStateConfig(clusterName);
```
jiajunwang: This is OK for this PR, but ideally, we should separate this type of API change into other PRs.

helix-core/src/test/java/org/apache/helix/integration/TestZkCallbackHandlerLeak.java
```diff
@@ -123,7 +123,7 @@ public boolean verify() throws Exception {
     // printHandlers(participantManagerToExpire);
     int controllerHandlerNb = controller.getHandlers().size();
     int particHandlerNb = participantManagerToExpire.getHandlers().size();
-    Assert.assertEquals(controllerHandlerNb, 11,
+    Assert.assertEquals(controllerHandlerNb, 12,
         "HelixController should have 10 (5+2n) callback handlers for 2 (n) participant");
```
jiajunwang: update the comment? Or the math won't be right.

helix-core/src/test/java/org/apache/helix/integration/TestZkCallbackHandlerLeak.java
```diff
@@ -247,7 +247,7 @@ public boolean verify() throws Exception {
 
     int controllerHandlerNb = controller.getHandlers().size();
     int particHandlerNb = participantManager.getHandlers().size();
-    Assert.assertEquals(controllerHandlerNb, 7 + 2 * n,
+    Assert.assertEquals(controllerHandlerNb, 8 + 2 * n,
         "HelixController should have 10 (6+2n) callback handlers for 2 participant, but was "
```
jiajunwang: Same here.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
jiajunwang: Is this change covered by any unit tests?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1072,6 +1146,15 @@ protected void checkLiveInstancesObservation(List<LiveInstance> liveInstances,
         }
       }
 
+      List<String> customizedStates = getEnabledCustomizedStates(manager);
+      for (String customizedState: customizedStates) {
+        for (String instance : curInstances.keySet()) {
+          if (lastInstances == null || !lastInstances.containsKey(instance)) {
+            addCustomizedStateListeners(manager, customizedState, instance);
```
jiajunwang: When you have add, we should have a corresponding remove to prevent listener leakage.

jiajunwang:
One more thing, the added test cases are not enough. Please add tests for the following features:
1. We need to add a case in the test listener leakage for the new listener introduced.
2. Admin change to create the new paths.
3. In the integration test, please test adjusting the view config and see if the view changed accordingly. Moreover, we should use the customized view provider to test.
4. Try to cover some error case, for example, instances are not reporting the configured type. Or 2 instances are reporting different sets of types.
Please try to cover all the use cases that we mentioned in the design doc.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
zhangmeng916: added in TestZkHelixAdmin

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -775,6 +811,38 @@ public void onStateChange(String instanceName, List<CurrentState> statesInfo,
     logger.info("END: GenericClusterController.onStateChange()");
   }
 
+  @Override
```
dasahcc: Prefetch ?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -569,12 +570,19 @@ public void addCurrentStateChangeListener(org.apache.helix.CurrentStateChangeLis
...
+        ChangeType.CUSTOMIZED_STATE_ROOT, new EventType[]{EventType.NodeChildrenChanged});
+  }
+
   @Override
   public void addCustomizedStateChangeListener(CustomizedStateChangeListener listener,
-      String instanceName, String customizedStateName) throws Exception {
-    addListener(listener, new Builder(_clusterName).customizedStates(instanceName, customizedStateName),
-        ChangeType.CUSTOMIZED_STATE, new EventType[] { EventType.NodeChildrenChanged
-        });
+      String instanceName, String customizedStateType) throws Exception {
```
dasahcc: We subscribe on root of customized state for children change. Do we still need for each indivdual?

zhangmeng916: Yes, because under customized state, there're different types of customized states, and their children are different resources, and then customized state data. Only listen to one level of children is not enough.

zhangmeng916:
> One more thing, the added test cases are not enough. Please add tests for the following features:
> 
> 1. We need to add a case in the test listener leakage for the new listener introduced.
This one is tested in TestZkCallbackHandlerLeak.

> 2. Admin change to create the new paths.
This one is tested in TestZkHelixAdmin.

> 3. In the integration test, please test adjusting the view config and see if the view changed accordingly. Moreover, we should use the customized view provider to test.
Tested in TestComputeAndCleanupCustomizedView.
The provider has unit test checked in, and the end to end integration test will cover from provider to router, which will be checked in by @mgao0.

> 4. Try to cover some error case, for example, instances are not reporting the configured type. Or 2 instances are reporting different sets of types.
Added test cases in TestComputeAndCleanupCustomizedView

>    Please try to cover all the use cases that we mentioned in the design doc.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -424,6 +446,20 @@ public void invoke(NotificationContext changeContext) throws Exception {
         List<CurrentState> currentStates = preFetch(_propertyKey);
         currentStateChangeListener.onStateChange(instanceName, currentStates, changeContext);
 
+      } else if (_changeType == CUSTOMIZED_STATE_ROOT) {
+        CustomizedStateRootChangeListener customizedStateRootChangeListener =
+            (CustomizedStateRootChangeListener) _listener;
+        String instanceName = PropertyPathConfig.getInstanceNameFromPath(_path);
+        customizedStateRootChangeListener.onCustomizedStateRootChange(instanceName,
```
jiajunwang: If prefetch, I guess we should get all the children for types and pass to the onCustomizedStateRootChange() method?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
jiajunwang: That would be one additional layer than the current callbackhandler infrastructure can support.
I would like to have a try to improve it after this PR is in.

helix-core/src/main/java/org/apache/helix/api/listeners/CustomizedStateRootChangeListener.java
```diff
@@ -0,0 +1,36 @@
...
+ * Interface to implement to respond to changes in the root path of customized state
+ */
+public interface CustomizedStateRootChangeListener {
+
+  /**
+   * Invoked when root path customized state changes
+   * @param instanceName name of the instance whose state changed
+   * @param changeContext the change event and state
+   */
+  void onCustomizedStateRootChange(String instanceName, NotificationContext changeContext);
```
jiajunwang: Shall this method be passed a list of types in case anyone defined it with PreFetch == true?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -259,45 +266,52 @@ private void parseListenerProperties() {
...
-      listenerClass = ControllerChangeListener.class;
+        break;
+      case LIVE_INSTANCE:
+        listenerClass = LiveInstanceChangeListener.class;
+        break;
+      case CURRENT_STATE:
+        listenerClass = CurrentStateChangeListener.class;
+        break;
+      case CUSTOMIZED_STATE_ROOT:
+        listenerClass = CustomizedStateRootChangeListener.class;
```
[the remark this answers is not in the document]
jiajunwang: Yeah, I double checked later. It is not doable given the current infrastructure only take the first method it found.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -450,8 +451,10 @@ public void invoke(NotificationContext changeContext) throws Exception {
         CustomizedStateRootChangeListener customizedStateRootChangeListener =
             (CustomizedStateRootChangeListener) _listener;
         String instanceName = PropertyPathConfig.getInstanceNameFromPath(_path);
-        customizedStateRootChangeListener.onCustomizedStateRootChange(instanceName,
-            changeContext);
+        List<String> customizedStateTypes =
```
jiajunwang: If prefetch enabled, then we read. Otherwise, we should pass an empty list.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -813,23 +817,36 @@ public void onStateChange(String instanceName, List<CurrentState> statesInfo,
...
-    for (String customizedState : customizedStateTypes) {
-      try {
-        manager.addCustomizedStateChangeListener(this, instanceName, customizedState);
-        logger.info(
-            manager.getInstanceName() + " added customized state listener for " + instanceName
-                + ", listener: " + this);
-      } catch (Exception e) {
-        logger.error("Fail to add customized state listener for instance: " + instanceName, e);
+    synchronized (_lastSeenCustomizedStateTypes) {
+      Set<String> lastSeenCustomizedStateTypes = _lastSeenCustomizedStateTypes.get();
```
jiajunwang: If we already lock on _lastSeenCustomizedStateTypes, do we still need to use AtomicReference?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -813,23 +817,36 @@ public void onStateChange(String instanceName, List<CurrentState> statesInfo,
...
+      for (String customizedState : customizedStateTypes) {
```
jiajunwang: Note that if prefetch is false, you should not expect customizedStateTypes to be filled with data.
The code works here because you are not following the prefetch configuration in the callback handler logic.

zhangmeng916: I made the change. However, I feel we actually do not need the prefetch logic for this event. It's redundant now. Please let me know.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
zhangmeng916: I saw a todo here:
 // TODO: remove the synchronization here once we move this update into dataCache.
    synchronized (_lastSeenInstances) {
I'm not sure whether there's any pending task for this that could also apply here, so I kept it.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -822,6 +822,11 @@ public void onCustomizedStateRootChange(String instanceName, List<String> custom
     logger.info("START: GenericClusterController.onCustomizedStateRootChange()");
     HelixManager manager = changeContext.getManager();
     Builder keyBuilder = new Builder(manager.getClusterName());
+   if (customizedStateTypes.isEmpty()) {
+     customizedStateTypes =
+         manager.getHelixDataAccessor().getChildNames(
+             manager.getHelixDataAccessor().keyBuilder().customizedStatesRoot(instanceName));
```
jiajunwang: nit, you have the keyBuilder above.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: Then please also carry over the TODO part.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
zhangmeng916: This is because the previous code was not correctly formatted in the switch section.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregateStage.java
zhangmeng916: If you take a look at it now, it's simpler than before. Right now it only has remove stale state types, and update customized view logic.

zhangmeng916:
Final commit message:
Add new stages in Helix generic controller for customized view aggregation.
- Add extra stages and pipelines in controller for customized state computation and customized view aggregation.
- Add refresh logic in resource data provider for customized view related data refresh.
- Add customized state event handling in CallbackHandler.
- Add integration test for customized view aggregation.
- Modify existing tests to verify new logic.

Issue #854, opened 2020-03-04
Make ZkBaseDataAccessor Realm-aware


pkuwm:
To make Helix Java APIs realm-aware, we need to make ZkBaseDataAccessor to use FederatedZkClient for realm-aware multi-realm operations.

PR #855, opened 2020-03-04
Make ZkBaseDataAccessor realm-aware

pkuwm:
Draft: to add a test for multi-realm mode and depends on #819 

### Issues

Implements #854 

### Description

To make Helix Java APIs realm-aware, we need to make ZkBaseDataAccessor realm-aware to use FederatedZkClient for multi-realm mode.

### Tests

TestZkBaseDataAccessor covers the cases.

```
[INFO] Results:
[INFO]
[ERROR] Failures:
[WARNING] Flakes:
[WARNING] org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceNonRack.testLackEnoughInstances(org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceNonRack)
[INFO]   Run 1: PASS
[ERROR]   Run 2: TestCrushAutoRebalanceNonRack.testLackEnoughInstances:250 » ZkClient Failed to...
[INFO]
[INFO]
[ERROR] Tests run: 1082, Failures: 1, Errors: 0, Skipped: 1, Flakes: 1
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:09 h
[INFO] Finished at: 2020-03-11T22:06:21-07:00

[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 18.273 s - in org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceNonRack
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 9.489 s - in org.apache.helix.integration.TestPauseSignal
[INFO]
[INFO] Results:
[INFO]

```

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -137,24 +137,15 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
   }
 
   private ZkBaseDataAccessor(Builder builder) {
-    RealmAwareZkClient zkClient;
     switch (builder.realmMode) {
       case MULTI_REALM:
         try {
-          zkClient = new FederatedZkClient(builder.realmAwareZkConnectionConfig,
+          _zkClient = new FederatedZkClient(builder.realmAwareZkConnectionConfig,
               builder.realmAwareZkClientConfig);
-          // Break here to exit.
-          break;
         } catch (IOException | InvalidRoutingDataException | IllegalStateException e) {
-          if (builder.zkAddress == null || builder.zkAddress.isEmpty()) {
-            throw new IllegalStateException("Not able to connect on multi-realm mode.", e);
-          }
-          LOG.info("Not able to connect on multi-realm mode. "
-              + "Connecting on single-realm mode to ZK: {}", builder.zkAddress);
-          builder.setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM);
+          throw new HelixException("Not able to connect on multi-realm mode.", e);
         }
-        // No break here. If connecting on multi-realm fails and ZK address is valid, connecting
-        // on single-realm is allowed.
+        break;
 
       case SINGLE_REALM:
         // Create a HelixZkClient: Use a SharedZkClient because ZkBaseDataAccessor does not need to
@@ -163,24 +154,24 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
           case DEDICATED:
             // If DEDICATED, then we use a dedicated HelixZkClient because we must support ephemeral
             // operations
-            zkClient = DedicatedZkClientFactory.getInstance()
+            _zkClient = DedicatedZkClientFactory.getInstance()
                 .buildZkClient(new HelixZkClient.ZkConnectionConfig(builder.zkAddress),
                     builder.realmAwareZkClientConfig.createHelixZkClientConfig());
             break;
           case SHARED:
           default:
-            zkClient = SharedZkClientFactory.getInstance()
+            _zkClient = SharedZkClientFactory.getInstance()
                 .buildZkClient(new HelixZkClient.ZkConnectionConfig(builder.zkAddress),
                     builder.realmAwareZkClientConfig.createHelixZkClientConfig());
-            zkClient.waitUntilConnected(HelixZkClient.DEFAULT_CONNECTION_TIMEOUT,
+            _zkClient.waitUntilConnected(HelixZkClient.DEFAULT_CONNECTION_TIMEOUT,
                 TimeUnit.MILLISECONDS);
+            break;
         }
         break;
       default:
         throw new HelixException("Invalid RealmMode given: " + builder.realmMode);
     }
 
-    _zkClient = zkClient;
     _usesExternalZkClient = false;
   }
 
@@ -1383,6 +1374,10 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
     }
   }
 
+  /*
+   * This is used for constructors that do not take a Builder in as a parameter
+   * because of the fallback behavior.
+   */
   private RealmAwareZkClient buildRealmAwareZkClient(
       RealmAwareZkClient.RealmAwareZkClientConfig clientConfig, String zkAddress,
       ZkClientType zkClientType) {
@@ -1391,7 +1386,8 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
           new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder().build(), clientConfig);
     } catch (IllegalStateException | IOException | InvalidRoutingDataException e) {
       // Fall back to connect on single-realm mode if failed to connect on multi-realm mode and
-      // ZK address is not empty.
+      // ZK address is not empty. This is for backward-compatibility for constructors that take
+      // in zkAddress or zkClient.
       LOG.info("Not able to connect on multi-realm mode, caused by: {}. "
           + "Connecting on single-realm mode to ZK: {}.", e.getMessage(), zkAddress);
     }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -51,7 +51,6 @@ import org.apache.helix.zookeeper.zkclient.exception.ZkBadVersionException;
 import org.apache.helix.zookeeper.zkclient.exception.ZkException;
 import org.apache.helix.zookeeper.zkclient.exception.ZkNoNodeException;
 import org.apache.helix.zookeeper.zkclient.exception.ZkNodeExistsException;
-import org.apache.helix.zookeeper.zkclient.serialize.BasicZkSerializer;
 import org.apache.helix.zookeeper.zkclient.serialize.ZkSerializer;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.KeeperException.Code;
@@ -259,7 +258,11 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
   @Deprecated
   public ZkBaseDataAccessor(String zkAddress, ZkSerializer zkSerializer,
       ZkClientType zkClientType) {
-    this(zkAddress, new BasicZkSerializer(zkSerializer), zkClientType);
+    RealmAwareZkClient.RealmAwareZkClientConfig clientConfig =
+        new RealmAwareZkClient.RealmAwareZkClientConfig().setZkSerializer(zkSerializer);
+
+    _zkClient = buildRealmAwareZkClient(clientConfig, zkAddress, zkClientType);
+    _usesExternalZkClient = false;
   }
 
   /**
@@ -275,8 +278,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    * @deprecated it is recommended to use the builder constructor {@link Builder}
    */
   @Deprecated
-  public ZkBaseDataAccessor(String zkAddress,
-      org.apache.helix.zookeeper.zkclient.serialize.PathBasedZkSerializer pathBasedZkSerializer,
+  public ZkBaseDataAccessor(String zkAddress, PathBasedZkSerializer pathBasedZkSerializer,
       ZkClientType zkClientType) {
     RealmAwareZkClient.RealmAwareZkClientConfig clientConfig =
         new RealmAwareZkClient.RealmAwareZkClientConfig().setZkSerializer(pathBasedZkSerializer);
@@ -1357,6 +1359,13 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
       return this;
     }
 
+    /**
+     * Returns a <code>ZkBaseDataAccessor</code> instance.
+     * <p>
+     * Note: in multi-realm mode, if and only if ZK client type is set to <code>FEDERATED</code>,
+     * <code>ZkBaseDataAccessor</code> can access to multi-realm. Otherwise, it can only access to
+     * single-ream.
+     */
     public ZkBaseDataAccessor<?> build() {
       validate();
       return new ZkBaseDataAccessor<>(this);
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -1338,30 +1338,30 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
   private RealmAwareZkClient buildRealmAwareZkClient(
       RealmAwareZkClient.RealmAwareZkClientConfig clientConfig, String zkAddress,
       ZkClientType zkClientType) {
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -102,14 +107,31 @@ public AccessResult() {
...
   private final boolean _usesExternalZkClient;
 
+  /**
+   * This default constructor attempts to connect to ZK on multi-realm mode.
+   *
+   * @exception IllegalStateException if connecting to ZK on multi-realm mode fails
+   */
+  public ZkBaseDataAccessor() {
+    this(EMPTY_ZK_ADDRESS);
+  }
```
narendly: ZkBaseDataAccessor needs to be able to take in Connection and Client configs. Instead of a constructor that takes in 0 parameters, could we use the builder pattern to achieve this? See the example for ConfigAccessor.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -1248,4 +1251,41 @@ public void close() {
...
+
+  private RealmAwareZkClient buildRealmAwareZkClient(
+      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig, String zkAddress,
+      ZkClientType zkClientType) {
+    RealmAwareZkClient zkClient;
+
+    try {
+      zkClient = new FederatedZkClient(
+          new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder().build(), clientConfig);
+    } catch (IllegalStateException | IllegalArgumentException | IOException | InvalidRoutingDataException e) {
```
narendly: Under what circumstances will `IllegalArgumentException` be thrown?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -1248,4 +1240,123 @@ public void close() {
...
+      // Resolve RealmAwareZkConnectionConfig
+      if (realmAwareZkConnectionConfig == null) {
+        // If not set, create a default one
+        realmAwareZkConnectionConfig =
+            new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder().build();
+      }
+    }
+  }
+
+  private ZkBaseDataAccessor(Builder builder) throws IOException, InvalidRoutingDataException {
```
narendly: Could we move up this constructor so it's with other public constructors?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -1248,4 +1240,123 @@ public void close() {
       _zkClient.close();
     }
   }
+
+  public static class Builder {
+    private String zkAddress;
+    private RealmAwareZkClient.RealmMode realmMode;
+    private RealmAwareZkClient.RealmAwareZkConnectionConfig realmAwareZkConnectionConfig;
+    private RealmAwareZkClient.RealmAwareZkClientConfig realmAwareZkClientConfig;
+
```
narendly: We'd need ZkBaseDataAccessor-specific parameters here as well such as ZkClientType (See https://github.com/apache/helix/pull/863 for reference)

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -1248,4 +1240,123 @@ public void close() {
...
+    public ZkBaseDataAccessor<?> build() throws Exception {
+      validate();
+      return new ZkBaseDataAccessor<>(this);
+    }
+
+    /**
+     * Validate the given parameters before creating an instance of ConfigAccessor.
+     */
+    private void validate() {
```
narendly: Ideally, validation logic should also resolve things like ZkClientType (if it's not set), so that the constructor should only worry about "constructing", not figuring out what's set or not. Let's try to separate logic as much as we can - this will help future development much easier in my opinion.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -102,21 +108,75 @@ public AccessResult() {
...
 
+  /**
+   * @deprecated it is recommended to use the other constructors instead to avoid having to manually
+   * create and maintain a RealmAwareZkClient outside of ZkBaseDataAccessor.
+   *
+   * @param zkClient A created RealmAwareZkClient
+   */
   @Deprecated
-  public ZkBaseDataAccessor(HelixZkClient zkClient) {
+  public ZkBaseDataAccessor(RealmAwareZkClient zkClient) {
```
narendly: Do we have a PR for HelixDataAccessor as well?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -102,21 +108,75 @@ public AccessResult() {
...
+          if (builder.zkAddress == null || builder.zkAddress.isEmpty()) {
+            throw new IllegalStateException("Not able to connect on multi-realm mode.", e);
+          }
+          LOG.info("Not able to connect on multi-realm mode. "
+              + "Connecting on single-realm mode to ZK: {}", builder.zkAddress);
+          builder.setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM);
+        }
+        // No break here. If connecting on multi-realm fails and ZK address is valid, connecting
+        // on single-realm is allowed.
+
```
narendly: As discussed offline, please make sure you're throwing a HelixException and propagating the Exception. 

Note that we shouldn't fall back to single-realm mode. We should fail here.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -1248,4 +1289,122 @@ public void close() {
...
+        zkClient = DedicatedZkClientFactory.getInstance()
+            .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
+                clientConfig.createHelixZkClientConfig());
+        break;
+      case SHARED:
+      default:
+        zkClient = SharedZkClientFactory.getInstance()
+            .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
+                clientConfig.createHelixZkClientConfig());
+        break;
```
narendly: Is this break necessary?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -1248,4 +1289,122 @@ public void close() {
...
+      // Resolve RealmAwareZkConnectionConfig
+      if (realmAwareZkConnectionConfig == null) {
+        // If not set, create a default one
+        realmAwareZkConnectionConfig =
+            new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder().build();
+      }
+    }
+  }
+
+  private RealmAwareZkClient buildRealmAwareZkClient(
```
narendly: Let's clearly note that this private method is only to be used for constructors that do **not** take a Builder in as a parameter because of the fallback behavior.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -1248,4 +1289,122 @@ public void close() {
...
+      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig, String zkAddress,
+      ZkClientType zkClientType) {
+    try {
+      return new FederatedZkClient(
+          new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder().build(), clientConfig);
+    } catch (IllegalStateException | IOException | InvalidRoutingDataException e) {
+      // Fall back to connect on single-realm mode if failed to connect on multi-realm mode and
+      // ZK address is not empty.
+      LOG.info("Not able to connect on multi-realm mode, caused by: {}. "
+          + "Connecting on single-realm mode to ZK: {}.", e.getMessage(), zkAddress);
```
narendly: Great use of logging, but let's also note here that this is for backward-compatibility for constructors that take in zkaddress or zkclient.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
pkuwm: As we discussed offline, we decided not to add a builder constructor for HelixDataAccessor. 

The reason is: Underneath it is using a BaseDataAccessor, which could an instance of ZkBaseDataAccessor/ZkCacheBaseDataAccessor/HelixPropertyStore, etc.. If we offer a builder constructor, we have to determine which instance of BaseDataAccessor. If we just offer a builder constructor like this Builder().setBaseDataAccessor(), I don’t think we need it.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
pkuwm: Done.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -1248,4 +1280,127 @@ public void close() {
...
+      return this;
+    }
+
+    public ZkBaseDataAccessor<?> build() throws Exception {
+      validate();
+      return new ZkBaseDataAccessor<>(this);
+    }
+
+    /**
+     * Validate the given parameters before creating an instance of ConfigAccessor.
```
pkuwm: Change comment.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -199,20 +238,13 @@ public ZkBaseDataAccessor(String zkAddress, ZkSerializer zkSerializer,
    * @param pathBasedZkSerializer
    * @param zkClientType
    */
-  public ZkBaseDataAccessor(String zkAddress, PathBasedZkSerializer pathBasedZkSerializer,
+  public ZkBaseDataAccessor(String zkAddress,
```
dasahcc: I think we should make the behavior description in the Java doc.

pkuwm: As discussed offline, we don't want the fallback behavior and also make these old constructors deprecated.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -102,28 +118,85 @@ public AccessResult() {
...
+  private ZkBaseDataAccessor(Builder builder) {
+    switch (builder.realmMode) {
+      case MULTI_REALM:
+        try {
+          if (builder.zkClientType == ZkBaseDataAccessor.ZkClientType.DEDICATED) {
+            // Use a realm-aware dedicated zk client
+            _zkClient = DedicatedZkClientFactory.getInstance()
+                .buildZkClient(builder.realmAwareZkConnectionConfig,
+                    builder.realmAwareZkClientConfig);
+          } else if (builder.zkClientType == ZkBaseDataAccessor.ZkClientType.SHARED) {
```
narendly: Nit: This is already inside ZkBaseDataAccessor, so no need to qualify using ZkBaseDataAccessor. Did you copy the code from the other PRs? IDE automatically puts these things.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -198,21 +271,17 @@ public ZkBaseDataAccessor(String zkAddress, ZkSerializer zkSerializer,
    * @param zkAddress
    * @param pathBasedZkSerializer
    * @param zkClientType
+   *
+   * @deprecated it is recommended to use the builder constructor {@link Builder}
    */
-  public ZkBaseDataAccessor(String zkAddress, PathBasedZkSerializer pathBasedZkSerializer,
+  @Deprecated
+  public ZkBaseDataAccessor(String zkAddress,
+      org.apache.helix.zookeeper.zkclient.serialize.PathBasedZkSerializer pathBasedZkSerializer,
```
narendly: why do we need the long package name for this serializer?

pkuwm: It is to reduce duplicate code for 2 constructors that have serializer parameters. `BasicZkSerializer` implements the long path `PathBasedZkSerializer` interface in zookeeper module. So we need this long path zookeeper `PathBasedZkSerializer`. 
```
public ZkBaseDataAccessor(String zkAddress, ZkSerializer zkSerializer, ZkClientType zkClientType) {
  this(zkAddress, new BasicZkSerializer(zkSerializer), zkClientType);
}
```
Cons: `new BasicZkSerializer(zkSerializer)` just follows the behavior in ZkClientConfig. If it is changed in ZkClientConfig to use another instance other than BasicZkSerializer, the one here would not follow that change. 
Well, if we don't want to reduce the duplicate code, we can just keep the original code.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -102,28 +118,85 @@ public AccessResult() {
...
+            // Use a realm-aware dedicated zk client
+            _zkClient = DedicatedZkClientFactory.getInstance()
+                .buildZkClient(builder.realmAwareZkConnectionConfig,
+                    builder.realmAwareZkClientConfig);
+          } else if (builder.zkClientType == ZkClientType.SHARED) {
+            // Use a realm-aware shared zk client
+            _zkClient = SharedZkClientFactory.getInstance()
+                .buildZkClient(builder.realmAwareZkConnectionConfig,
+                    builder.realmAwareZkClientConfig);
+          } else {
```
dasahcc: So we do agree to support shared zkclient while in multi realm mode?

narendly: This is a realm aware shared Zk client. If the user wants to use it this way, then they can

pkuwm: @dasahcc This is a realm-aware shared ZkClient. Users could construct the client with the sharding key on multi-realm mode.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
narendly: I'd prefer to leave it as is and tackle this later. This sort of refactoring doesn't have to be in the scope of this PR. Having 2 constructors is fine for now.

pkuwm:

```
This commits makes ZkBaseDataAccessor realm-aware by building according realm-aware ZkClients in the constructor. A Builder is provided to set realm-aware client config and connection config.
```

PR #858, opened 2020-03-04
Add error handling to getInstanceById

### LINKED ISSUE #857: Adding logs and exceptions to getInstanceById
Based on some user reports, the `getInstanceById` endpoint sometimes fail with a 500 for certain instances. Upon inspecting the logs, we realize that the endpoint is failing silently. 
To help with this situation and reinforce the ability to debug this endpoint, we decided to add logs and throw exceptions across all vulnerable points in relevant code.

NealSun96:
### Issues

Fixes #857

### Description

This PR add logs and exception throws across all vulnerable points in relevant code to the `getInstanceById` endpoint. We want to prevent the endpoint from silently failing.

### Tests
- [] The following is the result of the "mvn test" command on the appropriate module:

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/util/InstanceValidationUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/util/InstanceValidationUtil.java
@@ -257,6 +257,9 @@ public class InstanceValidationUtil {
   public static boolean isInstanceStable(HelixDataAccessor dataAccessor, String instanceName) {
     PropertyKey.Builder keyBuilder = dataAccessor.keyBuilder();
     ClusterConfig clusterConfig = dataAccessor.getProperty(keyBuilder.clusterConfig());
+    if (clusterConfig == null) {
+      throw new HelixException("Missing cluster config!");
+    }
     if (!clusterConfig.isPersistIntermediateAssignment()) {
       throw new HelixException("isInstanceStable needs persist assignment on!");
     }
@@ -266,6 +269,9 @@ public class InstanceValidationUtil {
       IdealState idealState = dataAccessor.getProperty(keyBuilder.idealStates(idealStateName));
       if (idealState == null || !idealState.isEnabled() || !idealState.isValid()
           || TaskConstants.STATE_MODEL_NAME.equals(idealState.getStateModelDefRef())) {
+        _logger.warn(
+            "idealState is either null, not enabled, not valid, or has Task as stateModelDefRef. IdealState: {}",
+            instanceName);
         continue;
       }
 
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/util/InstanceValidationUtil.java
```diff
@@ -266,6 +269,9 @@ public static boolean isInstanceStable(HelixDataAccessor dataAccessor, String in
       IdealState idealState = dataAccessor.getProperty(keyBuilder.idealStates(idealStateName));
       if (idealState == null || !idealState.isEnabled() || !idealState.isValid()
           || TaskConstants.STATE_MODEL_NAME.equals(idealState.getStateModelDefRef())) {
+        _logger.warn(
+            "idealState is either null, not enabled, not valid, or has Task as stateModelDefRef. IdealState: {}",
+            instanceName);
```
narendly: This won't make things fail, will it?

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
```diff
@@ -101,9 +103,15 @@ public InstanceInfo getInstanceInfo(String clusterId, String instanceName,
...
           partitions.addAll(currentState.getPartitionStateMap().keySet());
+        } else {
+          LOG.warn(
+              "Current state is either null or partitionStateMap is missing. InstanceName: {}, SessionId: {}, ResourceName: {}",
+              instanceName, sessionId, resourceName);
         }
       }
       instanceInfoBuilder.partitions(partitions);
+    } else {
+      LOG.warn("Missing instance config for {}", instanceName);
```
narendly: LiveInstance?

helix-core/src/main/java/org/apache/helix/util/InstanceValidationUtil.java
NealSun96: While this block doesn't fail the code, it's triggered in situations where idealState is null or is invalid. Just like the logs I did in `InstanceServiceImpl.java`, I think it's a good idea to log these unusual situations for debug purposes.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
NealSun96: Oops, let me fix that.

helix-core/src/main/java/org/apache/helix/util/InstanceValidationUtil.java
narendly: This block mostly exists to filter out non-existing IdealStates and Task Framework idealStates. The reason I pointed this out is because I don't think it will add much in debugging and it will just end up making the log file large. Let's not log this.

NealSun96: I mostly want to ensure the case of `idealState == null` is caught. That situation shouldn't happen. 
Regardless, I don't have a strong opinion about this.

narendly: You're not catching anything here, right? Also if IdealState is null, you skip over that resource, right? Would that ever cause a problem?

NealSun96:
Final commit message:
## Add logs and throw exceptions in getInstanceById ##

PR #859, opened 2020-03-04
Update customize state using updater

mgao0:
### Description

Currently the update customized state method is made synchronized for concurrency control. This PR modifies the implementation of update to leave the responsibility of concurrency control to ZooKeeper by using updater to update the customize state. With delete method already implemented with updater, we can prevent unexpected change of the customize state data.

### Tests

===============================================
Default Suite
Total tests run: 6, Failures: 0, Skips: 0
===============================================

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/customizedstate/CustomizedStateProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/customizedstate/CustomizedStateProvider.java
@@ -113,7 +113,7 @@ public class CustomizedStateProvider {
   /**
    * Delete the customized state for a specified resource and a specified partition
    */
-  public void deletePerPartitionCustomizedState(String customizedStateName, String resourceName,
+  public synchronized void deletePerPartitionCustomizedState(String customizedStateName, String resourceName,
       String partitionName) {
     PropertyKey.Builder keyBuilder = _helixDataAccessor.keyBuilder();
     PropertyKey propertyKey =
--- a/helix-core/src/main/java/org/apache/helix/customizedstate/CustomizedStateProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/customizedstate/CustomizedStateProvider.java
@@ -50,7 +50,7 @@ public class CustomizedStateProvider {
    * Update a specific customized state based on the resource name and partition name. The
    * customized state is input as a single string
    */
-  public synchronized void updateCustomizedState(String customizedStateName, String resourceName,
+  public void updateCustomizedState(String customizedStateName, String resourceName,
       String partitionName, String customizedState) {
     Map<String, String> customizedStateMap = new HashMap<>();
     customizedStateMap.put(CustomizedState.CustomizedStateProperty.CURRENT_STATE.name(), customizedState);
@@ -61,29 +61,16 @@ public class CustomizedStateProvider {
    * Update a specific customized state based on the resource name and partition name. The
    * customized state is input as a map
    */
-  public synchronized void updateCustomizedState(String customizedStateName, String resourceName,
+  public void updateCustomizedState(String customizedStateName, String resourceName,
       String partitionName, Map<String, String> customizedStateMap) {
     PropertyKey.Builder keyBuilder = _helixDataAccessor.keyBuilder();
     PropertyKey propertyKey =
         keyBuilder.customizedState(_instanceName, customizedStateName, resourceName);
     ZNRecord record = new ZNRecord(resourceName);
-    Map<String, Map<String, String>> mapFields = new HashMap<>();
-    CustomizedState existingState = getCustomizedState(customizedStateName, resourceName);
-    if (existingState != null
-        && existingState.getRecord().getMapFields().containsKey(partitionName)) {
-      Map<String, String> existingMap = new HashMap<>();
-      for (String key : customizedStateMap.keySet()) {
-        existingMap.put(key, customizedStateMap.get(key));
-      }
-
-      mapFields.put(partitionName, existingMap);
-    } else {
-      mapFields.put(partitionName, customizedStateMap);
-    }
-    record.setMapFields(mapFields);
+    record.setMapField(partitionName, customizedStateMap);
     if (!_helixDataAccessor.updateProperty(propertyKey, new CustomizedState(record))) {
-      throw new HelixException(
-          String.format("Failed to persist customized state %s to zk for instance %s, resource %s",
+      throw new HelixException(String
+          .format("Failed to persist customized state %s to zk for instance %s, resource %s",
               customizedStateName, _instanceName, record.getId()));
     }
   }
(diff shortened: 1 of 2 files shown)


mgao0:
This PR is ready to be merged, approved by @zhangmeng916 .

helix-core/src/main/java/org/apache/helix/customizedstate/CustomizedStateProvider.java
```diff
@@ -113,7 +113,7 @@ public CustomizedState getCustomizedState(String customizedStateName, String res
   /**
    * Delete the customized state for a specified resource and a specified partition
    */
-  public void deletePerPartitionCustomizedState(String customizedStateName, String resourceName,
+  public synchronized void deletePerPartitionCustomizedState(String customizedStateName, String resourceName,
```
jiajunwang: Why we need the synchronized keyword at the first place? ZK should be the one who takes care of concurrent control, right?

zhangmeng916: There can be multiple threads that try to delete partitions in a same resource. Do we want to leave the currency to the Zookeeper or maybe we can prevent it from the beginning. There are some similar pattern in the code base. We can discuss to decide whether we want to follow.

jiajunwang: ZK is the source of truth. Even we control the concurrency in one single node, we can not prevent multiple nodes concurrent access. So leverage ZK capability is the right way.
In detail, updater prevents unexpected overwrite. So I don't think the additional check is necessary. Unless we have a concrete counter-example.

jiajunwang:
> @jiajunwang Please see line 76, we rely on the record we get from zk for the update, and set the record to the local map to update the znode. So I think if there is any change during get and update, the data would be messed up.

I see your point. But if we are using the "updater", and use it in the correct way, I believe we don't need this additional sync. Actually, let's address this problem in this PR. Make the logic around 76 use updater.

mgao0:
This PR is ready to be merged, approved by @zhangmeng916

helix-core/src/test/java/org/apache/helix/integration/paticipant/TestCustomizedStateUpdate.java
```diff
@@ -19,54 +19,84 @@
...
+  @AfterClass
+  public void afterClass() throws Exception {
+    super.afterClass();
+    _manager.disconnect();
+  }
 
-    HelixDataAccessor dataAccessor = manager.getHelixDataAccessor();
+  @BeforeMethod
+  public void beforeMethod() {
+    HelixDataAccessor dataAccessor = _manager.getHelixDataAccessor();
```
narendly: This could go to beforeClass, right?

helix-core/src/test/java/org/apache/helix/integration/paticipant/TestCustomizedStateUpdate.java
```diff
@@ -113,40 +143,176 @@ public void testUpdateCustomizedState() throws Exception {
...
+        .deletePerPartitionCustomizedState(CUSTOMIZE_STATE_NAME, RESOURCE_NAME, PARTITION_NAME2);
+    Assert.assertNotNull(_mockProvider
+        .getPerPartitionCustomizedState(CUSTOMIZE_STATE_NAME, RESOURCE_NAME, PARTITION_NAME1));
+    Assert.assertNull(_mockProvider
+        .getPerPartitionCustomizedState(CUSTOMIZE_STATE_NAME, RESOURCE_NAME, PARTITION_NAME2));
+  }
+
+  @Test
+  public void testSimultaneousUpdateCustomizedState() {
+    int n = 10;
```
narendly: Not a good idea to use 'n's like this - please either make it a int constant or make it more descriptive.

helix-core/src/main/java/org/apache/helix/customizedstate/CustomizedStateProvider.java
```diff
@@ -61,29 +61,16 @@ public synchronized void updateCustomizedState(String customizedStateName, Strin
...
       String partitionName, Map<String, String> customizedStateMap) {
     PropertyKey.Builder keyBuilder = _helixDataAccessor.keyBuilder();
     PropertyKey propertyKey =
         keyBuilder.customizedState(_instanceName, customizedStateName, resourceName);
     ZNRecord record = new ZNRecord(resourceName);
-    Map<String, Map<String, String>> mapFields = new HashMap<>();
-    CustomizedState existingState = getCustomizedState(customizedStateName, resourceName);
-    if (existingState != null
-        && existingState.getRecord().getMapFields().containsKey(partitionName)) {
-      Map<String, String> existingMap = new HashMap<>();
```
jiajunwang: It is not clear what the original code tries to do here. It seems trying to merge the map. But not really doing so. Can you confirm the original logic is right or not?

The concern is that the default update method is doing deep merge. That is if you have an existing state map with some stale states, even after merging, those stale states will be kept in the map. But I assume we should clean those stale states for updating?

zhangmeng916: So this is user to update a customized state for a certain partition. I don't think we have the knowledge to clean up stale states, meaning if the user only update "CURRENT_STATE" of customized state, we update it, and keep the other fields unchanged. If they would like to do a full update, they will use the other API and provide the full map, and the original map will be overwritten.

jiajunwang: I don't think you get my point, or maybe I did not get the point. Anyway, let's go with an example,
1. The original map {"CSTypeA":"foo", "CSTypeB":"bar"}
2. User try to update with map {"CSTypeC":"foo", "CSTypeD":"bar"}
3. I assume the expect result for the custmize state is {"CSTypeC":"foo", "CSTypeD":"bar"}.
4. But really by the default merge method we get: {"CSTypeA":"foo", "CSTypeB":"bar", "CSTypeC":"foo", "CSTypeD":"bar"}
CSTypeA and CSTypeB is the stale state that I meant.

zhangmeng916: Yeah, we do want the full map like instead of replacing the previous map. Because for customized state, the key are fixed, e.g. "previous state", "current state", etc. We do not allow user to delete entries, and if they do not have value, the entry will have null value. 

For this API, it's updating the per partition customized state, not adding new type of customized states.
So it's a definite map of
{"start time": xxx
"end time:xxx
"previous state":xxx
"current state":xxx
}
and customers update one entry or the full map.

jiajunwang: I see. That makes sense.

mgao0:

jiajunwang:
> This PR is ready to be merged, approved by @jiajunwang.
> Final commit message: Use updater to update customized state

@mgao0 Could you please at least include the purpose of this change in the message? Thanks. More information would be better.

mgao0:
Final commit message: 
Title: Use updater to update customized state for concurrency control
Body: Currently the update customized state method is made synchronized for concurrency control. This commit modifies the implementation of update to leave the responsibility of concurrency control to ZooKeeper by using updater to update the customize state. With delete method already implemented with updater, we can prevent unexpected change of the customize state data.

Issue #860, opened 2020-03-05
Make ClusterSetup realm-aware


narendly:
This issue tracks the work of making ClusterSetup, a Helix Java API, realm-aware.

PR #861, opened 2020-03-05
Make ClusterSetup realm-aware

narendly:
### Issues

#860 

### Description

We make ClusterSetup, a Helix Java API, realm-aware so that this could be used in a multi-ZK environment. 
Changelist:
1. Add a Builder to enable users to set internal ZkClient parameters
2. Add the realm-aware behavior in existing constructors

### Tests

TestClusterSetup covers this.

**helix-core:**

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestHierarchicalDataStore.testHierarchicalDataStore:50->add:81 » ZkClient Inpu...
[INFO] 
[ERROR] Tests run: 1083, Failures: 3, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:07 h
[INFO] Finished at: 2020-03-11T02:14:14-07:00
```

All tests pass when run individually.

> mvn test -Dtest=TestHierarchicalDataStore,TestEnableCompression,TestControllerLeadershipChange,TestConfigAccessor,TestClusterSetup

> [INFO] Tests run: 19, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 76.508 s - in TestSuite
> [INFO] 
> [INFO] Results:
> [INFO] 
> [INFO] Tests run: 19, Failures: 0, Errors: 0, Skipped: 0
> [INFO] 
> [INFO] BUILD SUCCESS
> [INFO] Total time:  02:37 min
> [INFO] Finished at: 2020-03-11T19:34:16-07:00

### Commits

### Documentation (Optional)

### Code Quality


helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -135,37 +138,74 @@
...
 
-  public ClusterSetup(HelixZkClient zkClient, HelixAdmin zkHelixAdmin) {
+  public ClusterSetup(RealmAwareZkClient zkClient, HelixAdmin zkHelixAdmin) {
     _zkServerAddress = zkClient.getServers();
     _zkClient = zkClient;
     _admin = zkHelixAdmin;
     _usesExternalZkClient = true;
   }
 
+  private ClusterSetup(Builder builder) throws IOException, InvalidRoutingDataException {
```
pkuwm: Like you mentioned, do we need the checked exception signature? If yes, what is the consideration to have it thrown checked exceptions?

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -135,37 +138,74 @@
...
+    switch (builder._realmMode) {
+      case MULTI_REALM:
```
pkuwm: If it fails, shall we fall it back to connect to single ZK?

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -135,37 +138,74 @@
...
+    try {
+      zkClient = new FederatedZkClient(
+          new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder().build(),
+          new RealmAwareZkClient.RealmAwareZkClientConfig());
+    } catch (InvalidRoutingDataException | IOException | IllegalStateException e) {
+      // Note: IllegalStateException is for HttpRoutingDataReader if MSDS endpoint cannot be found
+      // Fall back to single-realm mode using SharedZkClient (HelixZkClient)
+      // This is to preserve backward-compatibility
+      if (zkServerAddress == null || zkServerAddress.isEmpty()) {
+        throw new IllegalArgumentException("ZK server address is null or empty!");
```
pkuwm: If zk address is not set for multi-realm mode, a user would not expect an `IllegalArgumentException`. Shall we throw an `IllegalStateException` as the user doesn't set this **argument**? And also include the original exception message/stack?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -78,6 +78,7 @@
 import org.apache.helix.util.HelixUtil;
 import org.apache.helix.util.RebalanceUtil;
 import org.apache.helix.zookeeper.api.client.HelixZkClient;
```
jiajunwang: nit, is this import needed anymore?

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -135,37 +138,74 @@
   public static final String removeConstraint = "removeConstraint";
 
   private static final Logger _logger = LoggerFactory.getLogger(ClusterSetup.class);
-  private final String _zkServerAddress;
-  private final HelixZkClient _zkClient;
-  // true if ZkBaseDataAccessor was instantiated with a HelixZkClient, false otherwise
+  private String _zkServerAddress;
```
jiajunwang: It seems this field can still be final. Why do we need to remove the final keyword?

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
```diff
@@ -1570,4 +1610,66 @@ public static void main(String[] args) throws Exception {
     int ret = processCommandLineArgs(args);
     System.exit(ret);
   }
+
+  public static class Builder {
```
jiajunwang: I see similar builders in multiple PRs for the different accessor or tools. Can we have a parent ZkClient builder which contains basic info and validation, and these builders can be the children of that parent.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
narendly: Good cross validation. This shouldn't be needed. Let me update the PR :+1:

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
narendly: @jiajunwang This was an option considered, but let me note here that having a parent builder doesn't actually remove duplicate code because all setters and constructors will have to either be re-written or overriden. This is because setters return Builder objects that aren't compatible with children builders. Take a look at the ZkClusterVerifier PR - in that PR, it makes more sense to have a parent Builder, but we don't see the reduction in duplicate code.

In general, having an inheritance hierarchy for static Builders is not necessarily a good idea because 1) it doesn't reduce code duplication, 2) Builder's setter and validation logic need to / may evolve differently, and inheritance might make that difficult in the future.

jiajunwang: For 2) the setter and validation logic should be common for the ZK parameters. And ZK related input and validation will be evolved the same. That's why a single class would help us to reduce maintenance costs.

For 1) There are multiple options to avoid the problem. Not sure if we have these options evaluated. If any document, that would be great so I can take a look : )
1. Pass the ZK Client builder to the verifier's builders. So there is no inheritance, but the ZK client building logic is in the client builder only.
2. Use a template class as mentioned here, https://stackoverflow.com/questions/4031857/way-to-make-java-parent-class-method-return-object-of-child-class
3. Since ZK parameters are in most cases requried, include them in the constructor and you don't need to worry about the return value. In the verifier builder, the constructor calls the super(...) constructor to finish the ZK client build.
There could be more ways.

narendly: There aren't any documents. However, I think this input is valid. 

Let's use this to track this issue. I think it would be wise to first focus on making Java APIs realm-aware, and take a step back and look at all of our Helix Java API classes and determine how we should refactor the Builders.

https://github.com/apache/helix/issues/873

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
narendly: If the user wants to create this class on multi-realm mode, FederatedZkClient will be used. FederatedZkClient does not and should not expose ZkAddress because it's an internal state info.

In fact, this `_zkServerAddress` was not being used anywhere and doesn't need to exist. I've removed this field from this class.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
narendly: If we want to preserve the existing behavior, I say we just remove this check (because the original implementation doesn't have this check). 

I am not sure what you're asking exactly... could you explain?

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
narendly: Discussed with @pkuwm offline.

We decided to keep these checked exceptions. As a matter of fact, all builder-based constructors should throw checked exceptions because things could go wrong on multi-realm mode.

helix-core/src/main/java/org/apache/helix/tools/ClusterSetup.java
narendly: Discussed offline. We shouldn't fall back on multi-realm mode.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
narendly: Needed. HelixZkClient is still used.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -123,23 +127,23 @@ public ConfigAccessor(RealmAwareZkClient zkClient) {
...
-          zkAddress, e);
-      zkClient = SharedZkClientFactory.getInstance()
-          .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
-              new HelixZkClient.ZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
-    }
-    _zkClient = zkClient;
     _usesExternalZkClient = false;
+
+    // If the multi ZK config is enabled, use FederatedZkClient on multi-realm mode
+    if (Boolean.parseBoolean(System.getProperty(SystemPropertyKeys.MULTI_ZK_ENABLED))) {
```
pkuwm: Nit: not sure whether or not we would allow users to set upper case “TRUE” and honor it as true. I just use getBoolean() to restrict it to true if and only if string is lowercase “true”.

narendly: They should both be considered as true. Either way works.

narendly:



We make ClusterSetup, a Helix Java API, realm-aware so that this could be used in a multi-ZK environment.
Changelist:

Add a Builder to enable users to set internal ZkClient parameters
Add the realm-aware behavior in existing constructors

Issue #875, opened 2020-03-06
Add domain construction in Helix participant manager


zhangmeng916:
Helix has a topology defined in cluster config, e.g.,  "TOPOLOGY": "/zone/host", and it will be used to construct the "domain" field in participant config, e.g. ""DOMAIN": "zone=0,host=host1". In cloud environment, Helix will provide a default topology value for the cluster, e.g., for Azure, the format is /faultDomain/hostName. Participant manager will need to construct the domain field based on this topology.

PR #876, opened 2020-03-06
Add construction of domain in Helix participant logic

zhangmeng916:
### Issues

- [X ] My PR addresses the following Helix issues and references them in the PR description:

(#875 )

### Description

Helix has a topology defined in cluster config, e.g., "TOPOLOGY": "/zone/host", and it will be used to construct the "domain" field in participant config, e.g. ""DOMAIN": "zone=0,host=host1". In Azure, Helix will provide a default topology value for the cluster, and the format is /faultDomain/hostName. In this pR, we add the construction function in Participant manager to construct the domain field based on this topology.

### Tests

### Commits

- [X ] My commits all reference appropriate Apache Helix GitHub issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":

### Code Quality

- [X ] My diff has been formatted using helix-style.xml 

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
+++ b/helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
@@ -144,9 +144,6 @@ public class AzureCloudInstanceInformationProcessor
         String vmssName = computeNode.path(INSTANCE_SET_NAME).getValueAsText();
         String azureTopology = AzureConstants.AZURE_TOPOLOGY;
         String[] parts = azureTopology.trim().split("/");
-        if (parts.length != 2) {
-          throw new HelixException("Invalid Azure topology definition: " + azureTopology);
-        }
         //The hostname will be filled in by each participant
         String domain = parts[0] + "=" + platformFaultDomain + "," + parts[1] + "=";
 
--- a/helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
+++ b/helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
@@ -142,8 +142,16 @@ public class AzureCloudInstanceInformationProcessor
         String vmName = computeNode.path(INSTANCE_NAME).getTextValue();
         String platformFaultDomain = computeNode.path(DOMAIN).getTextValue();
         String vmssName = computeNode.path(INSTANCE_SET_NAME).getValueAsText();
+        String azureTopology = AzureConstants.AZURE_TOPOLOGY;
+        String[] parts = azureTopology.trim().split("/");
+        if (parts.length != 2) {
+          throw new HelixException("Invalid Azure topology definition: " + azureTopology);
+        }
+        //The hostname will be filled in by each participant
+        String domain = parts[0] + "=" + platformFaultDomain + "," + parts[1] + "=";
+
         AzureCloudInstanceInformation.Builder builder = new AzureCloudInstanceInformation.Builder();
-        builder.setInstanceName(vmName).setFaultDomain(platformFaultDomain)
+        builder.setInstanceName(vmName).setFaultDomain(domain)
             .setInstanceSetName(vmssName);
         azureCloudInstanceInformation = builder.build();
       }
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -200,6 +201,22 @@ private void joinCluster() {
     }
   }
 
+  private String ConstructDomainField(String faultDomain) {
+    Boolean topologyEnabled = _configAccessor.getClusterConfig(_clusterName).isTopologyAwareEnabled();
```
dasahcc: Do we need this? If we know the environment is Azure, then we can directly get the topology from AzureConstants instead of reading from ClusterConfig.

zhangmeng916: In this logic, we do not parse the cloud provider directly. That logic is in the processor. I was thinking that if we read from clusterConfig, it'll be generic, and independent of cloud provider. thoughts?

zhangmeng916: Discussed offline. We'll move the logic to cloud instance information processor and not to depend on cluster config.

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
```diff
@@ -142,8 +142,16 @@ public AzureCloudInstanceInformation parseCloudInstanceInformation(List<String>
         String vmName = computeNode.path(INSTANCE_NAME).getTextValue();
         String platformFaultDomain = computeNode.path(DOMAIN).getTextValue();
         String vmssName = computeNode.path(INSTANCE_SET_NAME).getValueAsText();
+        String azureTopology = AzureConstants.AZURE_TOPOLOGY;
+        String[] parts = azureTopology.trim().split("/");
+        if (parts.length != 2) {
+          throw new HelixException("Invalid Azure topology definition: " + azureTopology);
+        }
```
dasahcc: This is not necessary, right? Unless we offer flexibility for user to customize it.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -178,7 +178,7 @@ private void joinCluster() {
           LOG.info(_instanceName + " is auto-registering cluster: " + _clusterName);
           CloudInstanceInformation cloudInstanceInformation = getCloudInstanceInformation();
           String domain = cloudInstanceInformation
-              .get(CloudInstanceInformation.CloudInstanceField.FAULT_DOMAIN.name());
+              .get(CloudInstanceInformation.CloudInstanceField.FAULT_DOMAIN.name()) + _instanceName;
```
dasahcc: This is domain string right? Then it should looked like "FD=xxx, InstanceName=xxxx". But the code is adding them together?

zhangmeng916: The prefix is in the processor: 
//The hostname will be filled in by each participant
 String domain = parts[0] + "=" + platformFaultDomain + "," + parts[1] + "=";

helix-core/src/main/java/org/apache/helix/cloud/azure/AzureCloudInstanceInformationProcessor.java
zhangmeng916: I was thinking whether it's possible the file was accidentally changed.

dasahcc: But this is not from the file. This is from Java class AzureConstants.

zhangmeng916: oh yes, I was thinking about my property file. Removed.

zhangmeng916:
This PR is ready to merge, approved by @dasahcc 
Final commit message:
Add the construction of domain field in Azure cloud instance information processor. It will read the topology from AzureConstants, and construct the domain based on the format of the topology.

PR #883, opened 2020-03-10
Support enableCompression in workflow and job configs

### LINKED ISSUE #884: Allow users to set enableCompression for workflow and job configs
See title.

narendly:
### Issues

Fixes #884 

### Description

Previously, there was no way for users to set enableCompression in workflows and jobs configs. This allows them to set that field in the Builder.

### Tests

```
Failed tests: 
  TestEnableCompression.testEnableCompressionResource:108 expected:<true> but was:<false>
  TestHelixAdminCli.testDeactivateCluster:604 » Helix There are still LEADER in ...

Tests run: 1087, Failures: 2, Errors: 0, Skipped: 1

[INFO] BUILD FAILURE
[INFO] Total time:  01:08 h
[INFO] Finished at: 2020-03-11T23:48:14-07:00
```

```
Results :

Tests run: 11, Failures: 0, Errors: 0, Skipped: 0

[INFO] BUILD SUCCESS
[INFO] Total time:  47.491 s
[INFO] Finished at: 2020-03-12T09:48:20-07:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/JobConfig.java
+++ b/helix-core/src/main/java/org/apache/helix/task/JobConfig.java
@@ -35,6 +35,8 @@ import org.apache.helix.HelixProperty;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.task.beans.JobBean;
 import org.apache.helix.task.beans.TaskBean;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+
 
 /**
  * Provides a typed interface to job configurations.
@@ -455,6 +457,7 @@ public class JobConfig extends ResourceConfig {
     private boolean _ignoreDependentJobFailure = DEFAULT_IGNORE_DEPENDENT_JOB_FAILURE;
     private int _numberOfTasks = DEFAULT_NUMBER_OF_TASKS;
     private boolean _rebalanceRunningTask = DEFAULT_REBALANCE_RUNNING_TASK;
+    private boolean _enableCompression = TaskConstants.DEFAULT_TASK_ENABLE_COMPRESSION;
 
     public JobConfig build() {
       if (_targetResource == null && _taskConfigMap.isEmpty()) {
@@ -536,11 +539,11 @@ public class JobConfig extends ResourceConfig {
       }
       if (cfg.containsKey(JobConfigProperty.DisableExternalView.name())) {
         b.setDisableExternalView(
-            Boolean.valueOf(cfg.get(JobConfigProperty.DisableExternalView.name())));
+            Boolean.parseBoolean(cfg.get(JobConfigProperty.DisableExternalView.name())));
       }
       if (cfg.containsKey(JobConfigProperty.IgnoreDependentJobFailure.name())) {
         b.setIgnoreDependentJobFailure(
-            Boolean.valueOf(cfg.get(JobConfigProperty.IgnoreDependentJobFailure.name())));
+            Boolean.parseBoolean(cfg.get(JobConfigProperty.IgnoreDependentJobFailure.name())));
       }
       if (cfg.containsKey(JobConfigProperty.JobType.name())) {
         b.setJobType(cfg.get(JobConfigProperty.JobType.name()));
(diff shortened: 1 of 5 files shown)


narendly:
> Do we want to just add this field or support a builder can carry over all the fields from JobConfig?

Just this field only because we want to curb the way user creates workflows and jobs now. Users should ideally use PropertyStore to store user-generated custom fields.

pkuwm:
> > Do we want to just add this field or support a builder can carry over all the fields from JobConfig?
> 
> Just this field only because we want to curb the way user creates workflows and jobs now. Users should ideally use PropertyStore to store user-generated custom fields.

Can we add a log to tell users their properties are ignored if any? I think that would be helpful for the end users.

helix-core/src/main/java/org/apache/helix/task/JobConfig.java
```diff
@@ -553,7 +556,11 @@ public static Builder fromMap(Map<String, String> cfg) {
       }
       if (cfg.containsKey(JobConfigProperty.RebalanceRunningTask.name())) {
         b.setRebalanceRunningTask(
-            Boolean.valueOf(cfg.get(JobConfigProperty.RebalanceRunningTask.name())));
+            Boolean.parseBoolean(cfg.get(JobConfigProperty.RebalanceRunningTask.name())));
+      }
+      if (cfg.containsKey(ZNRecord.ENABLE_COMPRESSION_BOOLEAN_FIELD)) {
```
pkuwm: It is redundant to check `containsKey`: if the value is not set, parseBoolean still returns false.
Unless the default value enableCompression is true, but I don't think this should be the case.

helix-core/src/main/java/org/apache/helix/task/WorkflowConfig.java
```diff
@@ -552,6 +559,11 @@ public Builder setConfigMap(Map<String, String> cfg) {
         }
       }
 
+      if (cfg.containsKey(ZNRecord.ENABLE_COMPRESSION_BOOLEAN_FIELD)) {
```
pkuwm: No need to check containsKey, unless the default value enableCompression is true.

narendly:
> > > Do we want to just add this field or support a builder can carry over all the fields from JobConfig?
> > 
> > 
> > Just this field only because we want to curb the way user creates workflows and jobs now. Users should ideally use PropertyStore to store user-generated custom fields.
> 
> Can we add a log to tell users their properties are ignored if any? I think that would be helpful for the end users.

I don't think adding logs is necessary. Logs are helpful for leaving historical information that can't be found anywhere else. In this case, however, users can easily check whether their fields ever made it to the znode by inspecting the znode.

Also, we want to limit the kind of fields we allow users to include in workflow and job config znodes by having them use Builders.

The biggest worry of all is again for heavy users of Task Framework, this might pollute the log.

helix-core/src/main/java/org/apache/helix/task/JobConfig.java
```diff
@@ -553,8 +556,11 @@ public static Builder fromMap(Map<String, String> cfg) {
       }
       if (cfg.containsKey(JobConfigProperty.RebalanceRunningTask.name())) {
         b.setRebalanceRunningTask(
-            Boolean.valueOf(cfg.get(JobConfigProperty.RebalanceRunningTask.name())));
+            Boolean.parseBoolean(cfg.get(JobConfigProperty.RebalanceRunningTask.name())));
       }
+      // If not set, parseBoolean returns false for null
+      b.setEnableCompression(
```
dasahcc: Then we should do a check here, right? If get null, we will get exception for parse boolean.

narendly: @dasahcc 

This is the implementation for parseBoolean() if the comment wasn't clear:

```
    public static boolean parseBoolean(String s) {
        return ((s != null) && s.equalsIgnoreCase("true"));
    }
```
Do you see why the null check is not necessary?

narendly:
```
    public static boolean parseBoolean(String s) {
        return ((s != null) && s.equalsIgnoreCase("true"));
    }
```

dasahcc: OK. The null should be final. But it make this entry explicitly in ZNRecord, right? Usually, our check is if the entry is not in ZNRecord, we use default value. But we dont carry the default value and explicitly set in ZNRecord to reduce the size of data set.

In this case, it is better we follow our convention: do the check when we carry it over.

narendly: Okay I see what you mean. I will revert to the original pattern.

PR #888, opened 2020-03-12
Modify data providers and add customized view aggregation stage

zhangmeng916:
### Issues

(#850 )

### Description
To support customized view aggregation, we need to change the current data providers to add the refresh for customized state and customized view. We also need to have two extra stages that perform computation from customized state to customized view. 

### Tests

[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -38,6 +38,7 @@ import org.apache.helix.HelixProperty;
 import org.apache.helix.PropertyKey;
 import org.apache.helix.common.caches.AbstractDataCache;
 import org.apache.helix.common.caches.CurrentStateCache;
+import org.apache.helix.common.caches.CustomizedStateCache;
 import org.apache.helix.common.caches.InstanceMessagesCache;
 import org.apache.helix.common.caches.PropertyCache;
 import org.apache.helix.common.controllers.ControlContextProvider;
@@ -45,6 +46,8 @@ import org.apache.helix.controller.LogUtil;
 import org.apache.helix.model.ClusterConfig;
 import org.apache.helix.model.ClusterConstraints;
 import org.apache.helix.model.CurrentState;
+import org.apache.helix.model.CustomizedState;
+import org.apache.helix.model.CustomizedStateConfig;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.LiveInstance;
@@ -96,6 +99,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
 
   // Special caches
   private CurrentStateCache _currentStateCache;
+  private CustomizedStateCache _customizedStateCache;
   private InstanceMessagesCache _instanceMessagesCache;
 
   // Other miscellaneous caches
@@ -103,6 +107,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
   private Map<String, Map<String, String>> _idealStateRuleMap;
   private Map<String, Map<String, Set<String>>> _disabledInstanceForPartitionMap = new HashMap<>();
   private Set<String> _disabledInstanceSet = new HashSet<>();
+  private Set<String> _aggregationEnabledTypes = new HashSet<>();
 
   public BaseControllerDataProvider() {
     this(AbstractDataCache.UNKNOWN_CLUSTER, AbstractDataCache.UNKNOWN_PIPELINE);
@@ -217,6 +222,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
       }
     }, false);
     _currentStateCache = new CurrentStateCache(this);
+    _customizedStateCache = new CustomizedStateCache(this, _aggregationEnabledTypes);
     _instanceMessagesCache = new InstanceMessagesCache(_clusterName);
   }
 
@@ -286,6 +292,27 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     }
   }
 
+  private void refreshCustomizedStateConfig(final HelixDataAccessor accessor,
+      Set<HelixConstants.ChangeType> refreshedType) {
+    if (_propertyDataChangedMap.get(HelixConstants.ChangeType.CUSTOMIZED_STATE_CONFIG)
+        .getAndSet(false)) {
+      CustomizedStateConfig customizedStateConfig =
+          accessor.getProperty(accessor.keyBuilder().customizedStateConfig());
+      if (customizedStateConfig != null) {
+        _aggregationEnabledTypes = new HashSet<>(customizedStateConfig.getRecord().getListFields()
+            .get(CustomizedStateConfig.CustomizedStateProperty.AGGREGATION_ENABLED_TYPES.name()));
+      }
+      LogUtil.logInfo(logger, getClusterEventId(), String
+          .format("Reloaded CustomizedStateConfig for cluster %s, %s pipeline.",
+              _clusterName, getPipelineName()));
+      refreshedType.add(HelixConstants.ChangeType.CUSTOMIZED_STATE_CONFIG);
+    } else {
+      LogUtil.logInfo(logger, getClusterEventId(), String
+          .format("No customized state config change for %s cluster, %s pipeline",
+              _clusterName, getPipelineName()));
+    }
+  }
+
   private void updateMaintenanceInfo(final HelixDataAccessor accessor) {
     _maintenanceSignal = accessor.getProperty(accessor.keyBuilder().maintenance());
     _isMaintenanceModeEnabled = _maintenanceSignal != null;
@@ -323,6 +350,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     refreshLiveInstances(accessor, refreshedTypes);
     refreshInstanceConfigs(accessor, refreshedTypes);
     refreshResourceConfig(accessor, refreshedTypes);
+    refreshCustomizedStateConfig(accessor, refreshedTypes);
     _stateModelDefinitionCache.refresh(accessor);
     _clusterConstraintsCache.refresh(accessor);
     updateMaintenanceInfo(accessor);
@@ -333,6 +361,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     // Refresh derived data
     _instanceMessagesCache.refresh(accessor, _liveInstanceCache.getPropertyMap());
     _currentStateCache.refresh(accessor, _liveInstanceCache.getPropertyMap());
+    _customizedStateCache.refresh(accessor, _liveInstanceCache.getPropertyMap());
 
     // current state must be refreshed before refreshing relay messages
     // because we need to use current state to validate all relay messages.
@@ -363,6 +392,8 @@ public class BaseControllerDataProvider implements ControlContextProvider {
       LogUtil.logDebug(logger, getClusterEventId(),
           "InstanceConfigs: " + getInstanceConfigMap().keySet());
       LogUtil.logDebug(logger, getClusterEventId(), "ClusterConfigs: " + getClusterConfig());
+      LogUtil.logDebug(logger, getClusterEventId(),
+          "CustomizedStateConfig: " + getCustomizedStateConfig());
     }
   }
 
@@ -372,6 +403,14 @@ public class BaseControllerDataProvider implements ControlContextProvider {
 
   public void setClusterConfig(ClusterConfig clusterConfig) {
     _clusterConfig = clusterConfig;
+  }
+
+  public Set<String> getCustomizedStateConfig() {
+    return _aggregationEnabledTypes;
+  }
+
+  public void setCustomizedStateConfig(Set<String> aggregationEnabledTypes) {
+    _aggregationEnabledTypes = aggregationEnabledTypes;
   }
 
   @Override
@@ -527,6 +566,17 @@ public class BaseControllerDataProvider implements ControlContextProvider {
    */
   public Map<String, CurrentState> getCurrentState(String instanceName, String clientSessionId) {
     return _currentStateCache.getParticipantState(instanceName, clientSessionId);
+  }
+
+  /**
+   * Provides the customized state of the node for a given state type (resource -> customizedState)
+   * @param instanceName
+   * @param customizedStateType
+   * @return
+   */
+  public Map<String, CustomizedState> getCustomizedState(String instanceName,
+      String customizedStateType) {
+    return _customizedStateCache.getParticipantState(instanceName, customizedStateType);
   }
 
   /**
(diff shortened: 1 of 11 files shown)


helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -156,6 +161,23 @@ private void refreshTargetExternalViews(final HelixDataAccessor accessor) {
     }
   }
 
+  public void refreshCustomizedViewMap(final HelixDataAccessor accessor) {
+    // As we are not listening on customized view change, customized view will be
+    // refreshed once during the cache's first refresh() call, or when full refresh is required
+    List<String> stateTypes = accessor.getChildNames(accessor.keyBuilder().customizedViews());
+    if (_propertyDataChangedMap.get(HelixConstants.ChangeType.CUSTOMIZED_VIEW).getAndSet(false)) {
+      for (String stateType: stateTypes) {
+        _customizedViewCacheMap.get(stateType).refresh(accessor);
```
jiajunwang: The map is initialized as en empty map. Do we need to check for NPE here?

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateComputationStage.java
```diff
@@ -0,0 +1,107 @@
...
+
+
+public class CustomizedStateComputationStage extends AbstractBaseStage {
+  private static Logger LOG = LoggerFactory.getLogger(CustomizedStateComputationStage.class);
+
+  @Override
+  public void process(ClusterEvent event) throws Exception {
+    HelixManager helixManager = event.getAttribute(AttributeName.helixmanager.name());
+    Set<String> aggregationEnabledTypes = new HashSet<>();
+    if (helixManager.getHelixDataAccessor().getProperty(
```
jiajunwang: Haven't we refreshed this in the BaseControllerDataProvider, can we just use that result here?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -217,6 +222,7 @@ public String getObjName(StateModelDefinition obj) {
       }
     }, false);
     _currentStateCache = new CurrentStateCache(this);
+    _customizedStateCache = new CustomizedStateCache(this, _aggregationEnabledTypes);
```
jiajunwang: If you passed "this", I guess you don't need to pass "_aggregationEnabledTypes" in addition. You can always call BaseControllerDataProvider.getCustomizedStateConfig() in the CustomizedStateCache. And that method is safer and cleaner.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -96,13 +99,15 @@
 
   // Special caches
   private CurrentStateCache _currentStateCache;
+  private CustomizedStateCache _customizedStateCache;
```
jiajunwang: I think if we add it here, it will be done by the Task pipeline as well. Is it necessary?

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
```diff
@@ -0,0 +1,171 @@
...
+    cache.removeCustomizedViewTypes(customizedViewTypesToRemove);
+
+    // update customized view
+    for (String stateType : customizedStateOutput.getAllStateTypes()) {
+      Map<String, CustomizedView> curCustomizedViews = new HashMap<>();
+      CustomizedViewCache customizedViewCache = customizedViewCacheMap.get(stateType);
+      if (customizedViewCache != null) {
+        curCustomizedViews = customizedViewCache.getCustomizedViewMap();
+      } else {
+        customizedViewCache = new CustomizedViewCache(event.getClusterName(), stateType);
```
jiajunwang: This logic confused me. Are we setting curCustomizedViews or customizedViewCache? Or both?
Could you please split these 2 parts?

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
```diff
@@ -0,0 +1,171 @@
...
+      }
+
+      for (Resource resource : resourceMap.values()) {
+        try {
+          computeCustomizedStateView(resource, stateType, customizedStateOutput, curCustomizedViews,
+              newCustomizedViews);
+          // Keep MBeans for existing resources and unregister MBeans for dropped resources
+          if (clusterStatusMonitor != null) {
+            clusterStatusMonitor.retainResourceMonitor(monitoringResources);
```
jiajunwang: I think this has been done in the external view stage. There is no need to do it twice.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
```diff
@@ -0,0 +1,171 @@
...
+          List<PropertyKey> keys = new ArrayList<>();
+          for (Iterator<CustomizedView> it = newCustomizedViews.iterator(); it.hasNext(); ) {
+            CustomizedView view = it.next();
+            String resourceName = view.getResourceName();
+            keys.add(keyBuilder.customizedView(stateType, resourceName));
+          }
+          // add/update customized-views
+          if (newCustomizedViews.size() > 0) {
+            dataAccessor.setChildren(keys, newCustomizedViews);
+            customizedViewCache.refresh(dataAccessor);
```
jiajunwang: Shall we just pass the in-memory content to the cache instead of requiring it to read from ZK again?

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
zhangmeng916: The logic here is that if customizedViewCache is not null, we will get the mapping (resource -> customizedView) from the cache, and use that in the later computation stage. However, if the customizedViewCache is null, meaning the stateType does not exist in the customizedViewCache list, we will initialize a new customizedViewCache for it.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
zhangmeng916: I tried to use in-memory content, and that does require a couple of set functions in CustomizedViewCache. I feel it's not clean. Please let me know if you think it's worth.

zhangmeng916: I've confirmed that routing table would not read from the customized view cache output from this stage, so we are safe to remove the refresh of cache

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -123,8 +136,11 @@ public synchronized void refresh(HelixDataAccessor accessor) {
     }
 
     // Refresh resource controller specific property caches
+    refreshCustomizedStateConfig(accessor);
+    _customizedStateCache.refresh(accessor, _liveInstanceCache.getPropertyMap());
     refreshExternalViews(accessor);
```
jiajunwang: Maybe we shall consider moving these 3 calculations to the update stages in the async thread. But let's leave it here for now. We will revisit anyway.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateComputationStage.java
```diff
@@ -0,0 +1,92 @@
...
+
+public class CustomizedStateComputationStage extends AbstractBaseStage {
+  private static Logger LOG = LoggerFactory.getLogger(CustomizedStateComputationStage.class);
+
+  @Override
+  public void process(ClusterEvent event) throws Exception {
+    _eventId = event.getEventId();
+    ResourceControllerDataProvider cache =
+        event.getAttribute(AttributeName.ControllerDataProvider.name());
+    final Map<String, Resource> resourceMap = event.getAttribute(AttributeName.RESOURCES.name());
```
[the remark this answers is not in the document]
zhangmeng916: I followed CurrentStateComputationStage convention here. What is exactly the difference for them?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
zhangmeng916: ok, will discuss in the controller PR.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateComputationStage.java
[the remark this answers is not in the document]
jiajunwang: The RESOURCES_TO_REBALANCE is a set of valid resources that we filtered. I think it would be a better input.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -139,6 +155,27 @@ protected void dumpDebugInfo() {
     }
   }
 
+  private void refreshCustomizedStateConfig(final HelixDataAccessor accessor) {
+    if (_propertyDataChangedMap.get(HelixConstants.ChangeType.CUSTOMIZED_STATE_CONFIG)
+        .getAndSet(false)) {
+      CustomizedStateConfig customizedStateConfig =
+          accessor.getProperty(accessor.keyBuilder().customizedStateConfig());
+      if (customizedStateConfig != null) {
+        _aggregationEnabledTypes =
```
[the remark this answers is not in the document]
jiajunwang: In that case, should be good. I thought it was just for protecting the input.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -89,7 +89,7 @@
   // Property caches
   private final PropertyCache<ResourceConfig> _resourceConfigCache;
   private final PropertyCache<InstanceConfig> _instanceConfigCache;
-  private final PropertyCache<LiveInstance> _liveInstanceCache;
+  protected final PropertyCache<LiveInstance> _liveInstanceCache;
```
jiajunwang: nit, safer to add a protected readonly method for getting the list instead of returning the list. Otherwise, the children may modify this list.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -139,6 +155,28 @@ protected void dumpDebugInfo() {
...
+            new HashSet<>(customizedStateConfig.getAggregationEnabledTypes());
+      } else {
+        _aggregationEnabledTypes.clear();
+      }
+      _customizedStateCache.setAggregationEnabledTypes(_aggregationEnabledTypes);
```
jiajunwang: Can we explicitly call this outside the refreshCustomizedStateConfig and just before _customizedStateCache.refresh()? Since _customizedStateCache.setAggregationEnabledTypes and refresh must be ordered, putting them together is easier for us to understand and maintain.

helix-core/src/test/java/org/apache/helix/controller/stages/TestCustomizedViewStage.java
```diff
@@ -0,0 +1,106 @@
...
+    Assert.assertEquals(cache.getCustomizedViewCacheMap().values(),
+        accessor.getChildValues(accessor.keyBuilder().customizedViews()));
+
+    // Assure there is no customized view got updated when running the stage again
+    List<CustomizedView> oldCustomizedViews =
+        accessor.getChildValues(accessor.keyBuilder().customizedViews());
+    runStage(event, customizedViewComputeStage);
+    List<CustomizedView> newCustomizedViews =
+        accessor.getChildValues(accessor.keyBuilder().customizedViews());
+    Assert.assertEquals(oldCustomizedViews, newCustomizedViews);
```
pkuwm: More accurate to convert the 2 lists to sets for assertion? If the ordering is different, this assert would fail.

zhangmeng916:
Final commit message:
Add two stages for customized state view aggregation. 

1. One stage is the computation stage for customized state. It takes the Zookeeper data of customized states and converts them to the formatted output used by the other stage. 
2. The other stage is customized view aggregation stage. It will take the output from the customized state computation stage, and output the customized view to Zookeeper.
3. The two stages together compute the customized view from the customized states. 
4. Unit tests are added to verify the correctness of the two stages.

helix-core/src/test/java/org/apache/helix/controller/stages/TestCustomizedViewStage.java
zhangmeng916: I thought the children were retrieved in a certain order, but by looking at the code, seems actually it's random based on the order whichever child is gotten back first. Is it right?

pkuwm: Ordering is not guaranteed. In ZK's Znode implementation, it is using a set to store children. And it is converted to a list when returning to the client. So the ordering could be random.
If you are just checking if they have the same element without ordering, I would suggest using Sets so ordering is ignored. Only if you also want to check the ordering, a list is more appropriate.

helix-core/src/test/java/org/apache/helix/controller/stages/TestCustomizedViewStage.java
```diff
@@ -0,0 +1,106 @@
...
+
+    CustomizedViewAggregationStage customizedViewComputeStage =
+        new CustomizedViewAggregationStage();
+    Pipeline dataRefresh = new Pipeline();
+    dataRefresh.addStage(new ReadClusterDataStage());
+    runPipeline(event, dataRefresh);
+    runStage(event, new ResourceComputationStage());
+    runStage(event, new CustomizedStateComputationStage());
+    runStage(event, customizedViewComputeStage);
+    Assert.assertEquals(cache.getCustomizedViewCacheMap().values(),
```
pkuwm: And also here. Collection vs List. Would be better to make them 2 sets.

helix-core/src/test/java/org/apache/helix/controller/stages/TestCustomizedViewStage.java
zhangmeng916: Actually I found in TestExternalVeiwStage the following code.
Assert.assertEquals(oldExternalViews, newExternalViews);
    for (int i = 0; i < oldExternalViews.size(); i++) {
      Assert.assertEquals(oldExternalViews.get(i).getStat().getVersion(),
          newExternalViews.get(i).getStat().getVersion());
    }

I believe this test doesn't fail. I think my test is using the same logic. Since I would also like to examine each item based on index, like the above code, let's keep it as list for now.  If we encounter any test failure, let's revisit it.

helix-core/src/test/java/org/apache/helix/controller/stages/TestCustomizedViewStage.java
zhangmeng916: Please see the comments above.

PR #892, opened 2020-03-13
Add integration tests for Helix Java APIs

narendly:
### Issues

- [x] My PR addresses the following Helix issues and references them in the PR description:

Fixes #891

### Description

- [x] Here are some details about my PR, including screenshots of any UI changes:

This commit adds a comprehensive integration test for Helix Java APIs. All Helix Java APIs are tested using regular resource rebalancing and task framework.

There are a few minor changes to Helix Java API code. Those will be a separate PR for easier reviewing. Note that this PR contains all of those changes.

Depends on the following PRs:
https://github.com/apache/helix/pull/893
https://github.com/apache/helix/pull/894
https://github.com/apache/helix/pull/895
https://github.com/apache/helix/pull/896

### Tests

- [x] The following tests are written for this issue:

TestMultiZkEnvironment

- [x] The following is the result of the "mvn test" command on the appropriate module:

```
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestEnableCompression.testEnableCompressionResource:117 expected:<true> but was:<false>
[INFO]
[ERROR] Tests run: 1097, Failures: 1, Errors: 0, Skipped: 0
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:10 h
[INFO] Finished at: 2020-03-16T21:31:06-07:00
[INFO] ------------------------------------------------------------------------
```

mvn test -Dtest=TestControllerLeadershipChange,TestEnableCompression

```
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 31.587 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:12 min
[INFO] Finished at: 2020-03-13T14:02:11-07:00
[INFO] ------------------------------------------------------------------------
```

### Commits

- [x] My commits all reference appropriate Apache Helix GitHub issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
[...]

PR #895, opened 2020-03-13
Update listClusters() in ZkHelixAdmin

### LINKED ISSUE #891: Test newly instrumented Helix Java APIs
We need a comprehensive integration test that tests the realm-aware functionalities of new Helix Java APIs and fix any bugs.

narendly:
### Issues

Resolves #891 

### Description

This PR updates getClusters() in ZkHelixAdmin in realm-aware mode. It is difficult to reason about getting all clusters - and this method, on single-realm mode, is also broken anyways because in the case users create nested clusters, this method will end up returning nothing. This update is backward-compatible for single-realm users.

Ideally, users should use Metadata Store Directory Service to get the list of all clusters.

### Tests

See https://github.com/apache/helix/pull/892

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -954,6 +954,12 @@ public class ZKHelixAdmin implements HelixAdmin {
 
   @Override
   public List<String> getClusters() {
+    if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED)
+        || _zkClient instanceof FederatedZkClient) {
+      throw new UnsupportedOperationException(
+          "getClusters() is not supported in multi-realm mode! Use Metadata Store Directory Service instead!");
+    }
+
     List<String> zkToplevelPathes = _zkClient.getChildren("/");
     List<String> result = new ArrayList<String>();
     for (String pathName : zkToplevelPathes) {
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1925,7 +1931,10 @@ private void validate() {
 
       // Resolve RealmAwareZkClientConfig
       if (realmAwareZkClientConfig == null) {
-        realmAwareZkClientConfig = new RealmAwareZkClient.RealmAwareZkClientConfig();
+        // ZkHelixAdmin should have ZNRecordSerializer set by default
+        realmAwareZkClientConfig = new RealmAwareZkClient.RealmAwareZkClientConfig()
+            .setZkSerializer(
+                new org.apache.helix.zookeeper.datamodel.serializer.ZNRecordSerializer());
```
pkuwm: Nit, I think currently we can still use the deprecated ZNRecordSerializer. Once the deprecated one is removed, we will also refactor and change it here. If we put the long path here, maybe we would not find it when removing the deprecated one.

narendly: But then we'd be using a deprecated component. I don't think using the package prefix is wrong per se..

narendly:

This PR updates getClusters() in ZkHelixAdmin in realm-aware mode. It is difficult to reason about getting all clusters - and this method, on single-realm mode, is also broken anyways because in the case users create nested clusters, this method will end up returning nothing. This update is backward-compatible for single-realm users.

Ideally, users should use Metadata Store Directory Service to get the list of all clusters.

PR #908, opened 2020-03-24
Make Helix REST realm-aware

### LINKED ISSUE #907: Make Helix REST realm-aware
Helix REST needs to start using a realm-aware ZkClient on multi-zk mode. Also it needs to become a listener on routing data because we don't want to restart the HelixRestServer every time we update the routing data.

narendly:
### Issues

Fixes #907 

### Description

Helix REST needs to start using a realm-aware ZkClient on multi-zk mode. Also it needs to become a listener on routing data because we don't want to restart the HelixRestServer every time we update the routing data.

Changelist:
1. Make ServerContext listen on routing data paths if run on multi-zk mode
2. Make HelixRestServer use RealmAwareZkClient (FederatedZkClient) on multi-zk mode

### Tests

TestRoutingDataUpdate

on helix-rest module:

> [INFO] Results:
> [INFO] 
> [ERROR] Failures: 
> [ERROR]   TestResourceAccessor.testResourceHealth:289 expected:<HEALTHY> but was:<UNHEALTHY>
> [INFO] 
> [ERROR] Tests run: 144, Failures: 1, Errors: 0, Skipped: 7
> [INFO] 
> [INFO] BUILD FAILURE
> [INFO] Total time:  37.847 s
> [INFO] Finished at: 2020-03-23T22:02:47-07:00

TestResourceAccessor is an unrelated failure that will be fixed in future iterations.

### Commits

### Code Quality


helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
```diff
@@ -58,25 +58,9 @@ public ZkRoutingDataReader(String namespace, String zkAddress,
...
         .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
             new HelixZkClient.ZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
-
-    // Ensure that ROUTING_DATA_PATH exists in ZK. If not, create
-    // create() semantic will fail if it already exists
-    try {
-      _zkClient.createPersistent(MetadataStoreRoutingConstants.ROUTING_DATA_PATH, true);
-    } catch (ZkNodeExistsException e) {
-      // This is okay
-    }
```
NealSun96: Why is this section removed? In a scenario where `ZkRoutingDataReader` is instantiated by itself, `subscribeRoutingDataChanges` might encounter `NoNodeException` if we don't ensure that ROUTING_DATA_PATH exists in ZK.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
```diff
@@ -82,20 +82,14 @@ public ZkRoutingDataWriter(String namespace, String zkAddress) {
         .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
             new HelixZkClient.ZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
 
-    // Ensure that ROUTING_DATA_PATH exists in ZK. If not, create
-    // create() semantic will fail if it already exists
-    try {
-      _zkClient.createPersistent(MetadataStoreRoutingConstants.ROUTING_DATA_PATH, true);
-    } catch (ZkNodeExistsException e) {
-      // This is okay
-    }
```
NealSun96: Likewise, without this section, some ZNode creation might fail if they didn't create the parent (ROUTING_DATA_PATH) and the parent doesn't exist.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -165,5 +238,93 @@ public void close() {
...
+
+  @Override
+  public void handleDataDeleted(String dataPath) {
+    if (_zkClientForListener == null || _zkClientForListener.isClosed()) {
+      return;
+    }
+    // Resubscribe
+    _zkClientForListener.unsubscribeAll();
+    ZkRoutingDataReader.subscribeRoutingDataChanges(_zkClientForListener, this, this);
+    resetZkResources();
```
NealSun96: Is it possible to reduce duplication among the handle functions?

narendly: Updated. It's hard to reduce duplication further because we need to sneak in unsubscribe/resubscribe for some callback functions.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataWriter.java
narendly: I'll re-add.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
narendly: Adding it back.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -339,4 +348,22 @@ public synchronized void close() {
...
+   */
+  private void createRoutingDataPath(HelixZkClient zkClient, String zkAddress) {
+    try {
+      zkClient.createPersistent(MetadataStoreRoutingConstants.ROUTING_DATA_PATH, true);
+    } catch (ZkNodeExistsException e) {
+      // The node already exists and it's okay
+    }
+    // Make sure ROUTING_DATA_PATH is mapped to the routing ZK so that FederatedZkClient used
+    // in Helix REST can subscribe to the routing data path
+    ZNRecord znRecord = new ZNRecord(MetadataStoreRoutingConstants.ROUTING_DATA_PATH);
```
dasahcc: The path contains"/" symbol. Would that be OK?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
```diff
@@ -138,40 +122,51 @@ public synchronized void handleChildChange(String s, List<String> list) {
...
     _routingDataListener.refreshRoutingData(_namespace);
   }
 
+  /**
+   * Subscribes to the routing data paths using the provided ZkClient.
+   * @param zkClient
+   * @param childListener
+   * @param dataListener
+   */
+  public static void subscribeRoutingDataChanges(RealmAwareZkClient zkClient,
```
dasahcc: Would you like to make it as public util function? If yes, then move it to some util class, otherwise, I think we have internal member of _zkClient, does not need to make the zkclient is passing in.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
```diff
@@ -138,40 +122,51 @@ public synchronized void handleChildChange(String s, List<String> list) {
...
     }
 
     // Renew subscription
     _zkClient.unsubscribeAll();
-    _zkClient.subscribeChildChanges(MetadataStoreRoutingConstants.ROUTING_DATA_PATH, this);
-    for (String child : _zkClient.getChildren(MetadataStoreRoutingConstants.ROUTING_DATA_PATH)) {
-      _zkClient.subscribeDataChanges(MetadataStoreRoutingConstants.ROUTING_DATA_PATH + "/" + child,
-          this);
-    }
+    ZkRoutingDataReader.subscribeRoutingDataChanges(_zkClient, this, this);
```
dasahcc: Let's make a util class for this purpose.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -20,42 +20,67 @@
...
-public class ServerContext {
+public class ServerContext implements IZkDataListener, IZkChildListener, IZkStateListener {
+  private static final Logger LOG = LoggerFactory.getLogger(ServerContext.class);
+
   private final String _zkAddr;
-  private HelixZkClient _zkClient;
-  private ZKHelixAdmin _zkHelixAdmin;
-  private ClusterSetup _clusterSetup;
-  private ConfigAccessor _configAccessor;
+  private volatile RealmAwareZkClient _zkClient;
```
dasahcc: The volatile keyword is trying to make variable sharing between threads. Do we have many threads for using these variables, which are not able to get the instance?

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -20,42 +20,67 @@
...
+  private volatile ZKHelixAdmin _zkHelixAdmin;
+  private volatile ClusterSetup _clusterSetup;
+  private volatile ConfigAccessor _configAccessor;
   // A lazily-initialized base data accessor that reads/writes byte array to ZK
   // TODO: Only read (deserialize) is supported at this time. This baseDataAccessor should support write (serialize) as needs arise
   private volatile ZkBaseDataAccessor<byte[]> _byteArrayZkBaseDataAccessor;
   // 1 Cluster name will correspond to 1 helix data accessor
   private final Map<String, HelixDataAccessor> _helixDataAccessorPool;
   // 1 Cluster name will correspond to 1 task driver
   private final Map<String, TaskDriver> _taskDriverPool;
```
dasahcc: Why we need a pool? Since you already make the task driver could accept the RealmAzreZkClient, we can pass in the FederatedZkClient as input, right? And TaskDriver does not have any session aware operations. So we can use FederatedZkClient for multiple cluster accessing purpose.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
narendly: That's okay. This is just an ID - it's for display purposes only.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
narendly: As suggested, I moved it to RealmAwareZkClient.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/accessor/ZkRoutingDataReader.java
narendly: Moved to RealmAwareZkClient interface.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
narendly: Yes. Jersey spins up a thread to serve each incoming request.

Also volatile is used to as a double-checked locking idiom. So it is needed here for lazy initialization.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
narendly: Your line of thinking is correct, but if we made that change here, Helix REST will no longer be backward-compatible. We leave it here so that for users that use Helix REST in a single ZK mode will be able to use it the way they were able to use before (1 TaskDriver per cluster)

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
dasahcc: That was confusing user. In our convention, the path to a certain will be "/ID of ZNode1/ID of ZNode2".  Usually we dont have folder name is one thing and id of ZNRecord is another thing. Suggest to make it consisnt.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
dasahcc: Can you elaborate why it is not backward compatible? I think FederatedZKClient can support single ZK access, right? We can make it accept single ZK while combining with multi ZK access. Otherwise, we already started multiple Servlet for different ZK works fine and I did not get the idea of making REST realm-aware.

narendly: If multiZk system config is not set and MSDS routing data is not available (which is the case for a single-zk realm user), you cannot create a FederatedZkClient. Does that make sense? So we need to preserve the behavior if the user is not using the multi-zk feature.

dasahcc: I agree with that. But the routing logic should not be at service layer. Otherwise, each service using TaskDriver or ConfigAccessor will have their own routing logic. 

TaskDriver / Config Accessor is instantiated from RealmAwareZKClient. Then can we make some smart things underneath for sharding key routing?

narendly: I understand your concern and where you're coming from.  But it's still difficult. TaskDriver and ConfigAccessor are different APIs by design.
- TaskDriver/HelixManager/HelixDataAccessor -> only serve 1 single Helix cluster
- ConfigAccessor/ClusterSetup/etc.. -> can serve many Helix clusters
So we need to keep a pool regardless.
To help you understand:

```
  public TaskDriver(HelixManager manager) {
  }
  @Deprecated
  public TaskDriver(RealmAwareZkClient client, String clusterName) {
  }
  @Deprecated
  public TaskDriver(RealmAwareZkClient client, ZkBaseDataAccessor<ZNRecord> baseAccessor,
      String clusterName) {}
  @Deprecated
  public TaskDriver(HelixAdmin admin, HelixDataAccessor accessor, ConfigAccessor cfgAccessor,
      HelixPropertyStore<ZNRecord> propertyStore, String clusterName) {
    this(admin, accessor, propertyStore, clusterName);
  }
```
You see that every constructor of TaskDriver takes in a cluster name.

Although TaskDriver does not support any ephemeral operations, by design it's meant to be a single cluster API.

I think what you wanted to see was to pass in some FederatedZkClient into TaskDriver and use one TaskDriver for all clusters, but 1) that does not work, and 2) if we wanted to make that work, that's going to require a lot of code change in TaskDriver.

narendly:

narendly:
Helix REST needs to start using a realm-aware ZkClient on multi-zk mode. Also it needs to become a listener on routing data because we don't want to restart the HelixRestServer every time we update the routing data.

Changelist:

Make ServerContext listen on routing data paths if run on multi-zk mode
Make HelixRestServer use RealmAwareZkClient (FederatedZkClient) on multi-zk mode

PR #915, opened 2020-03-26
Make multiZkEnabled configurable in HelixRestNamespace

### LINKED ISSUE #914: Make multiZk in helix-rest configurable in HelixRestNamespace
It was observed that we need more fine-grained control over this multiZkEnabled config because there could exists namespaces with differing modes. Because multiple namespaces may be co-deployed, we cannot simply make it a system config.

narendly:
### Issues

Fixes #914 

### Description

It was observed that we need more fine-grained control over this multiZkEnabled config because there could exists namespaces with differing modes. Because multiple namespaces may be co-deployed, we cannot simply make it a system config.

### Tests

Covered by existing tests.

[ERROR] Failures: 
[ERROR]   TestResourceAccessor.testResourceHealth:289 expected:<HEALTHY> but was:<UNHEALTHY>
[INFO] 
[ERROR] Tests run: 144, Failures: 1, Errors: 0, Skipped: 7
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  38.519 s
[INFO] Finished at: 2020-03-26T15:37:48-07:00
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M3:test (default-test) on project helix-rest: There are test failures.

Known failure. This will be dealt with in a different PR.

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/common/HelixRestNamespace.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/common/HelixRestNamespace.java
@@ -35,6 +35,7 @@ public class HelixRestNamespace {
     METADATA_STORE_TYPE,
     METADATA_STORE_ADDRESS,
     IS_DEFAULT,
+    MULTI_ZK_ENABLED,
     MSDS_ENDPOINT
   }
 
@@ -67,6 +68,11 @@ public class HelixRestNamespace {
   private boolean _isDefault;
 
   /**
+   * Flag indicating whether this namespace should have multi-zk feature enabled.
+   */
+  private boolean _isMultiZkEnabled;
+
+  /**
    * Endpoint for accessing MSDS for this namespace.
    */
   private String _msdsEndpoint;
@@ -77,15 +83,17 @@ public class HelixRestNamespace {
 
   public HelixRestNamespace(String name, HelixMetadataStoreType metadataStoreType,
       String metadataStoreAddress, boolean isDefault) throws IllegalArgumentException {
-    this(name, metadataStoreType, metadataStoreAddress, isDefault, null);
+    this(name, metadataStoreType, metadataStoreAddress, isDefault, false, null);
   }
 
   public HelixRestNamespace(String name, HelixMetadataStoreType metadataStoreType,
-      String metadataStoreAddress, boolean isDefault, String msdsEndpoint) {
+      String metadataStoreAddress, boolean isDefault, boolean isMultiZkEnabled,
+      String msdsEndpoint) {
     _name = name;
     _metadataStoreAddress = metadataStoreAddress;
     _metadataStoreType = metadataStoreType;
     _isDefault = isDefault;
+    _isMultiZkEnabled = isMultiZkEnabled;
     _msdsEndpoint = msdsEndpoint;
     validate();
   }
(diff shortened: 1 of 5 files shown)


helix-rest/src/main/java/org/apache/helix/rest/common/HelixRestNamespace.java
```diff
@@ -35,6 +35,7 @@
     METADATA_STORE_TYPE,
     METADATA_STORE_ADDRESS,
     IS_DEFAULT,
+    MULTI_ZK_ENABLED,
```
dasahcc: Why we dont just rely on the system property but a new field?

narendly: This is what the description says - please refer to the description carefully and let me know if it answers your question.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -111,20 +114,9 @@ public RealmAwareZkClient getRealmAwareZkClient() {
       synchronized (this) {
         if (_zkClient == null) {
           // If the multi ZK config is enabled, use FederatedZkClient on multi-realm mode
-          if (Boolean.parseBoolean(System.getProperty(SystemPropertyKeys.MULTI_ZK_ENABLED))) {
-            LOG.info("ServerContext: initializing FederatedZkClient with routing ZK at {}!",
-                _zkAddr);
+          if (_isMultiZkEnabled || Boolean
+              .parseBoolean(System.getProperty(SystemPropertyKeys.MULTI_ZK_ENABLED))) {
```
NealSun96: Aren't we moving away from system properties? 
If we allow both it could be confusing, unless we specifically document for it.

narendly: > Aren't we moving away from system properties?
Why?

> If we allow both it could be confusing, unless we specifically document for it.
This is for convenience. Users that have many namespaces can simply set the system property. Otherwise for more fine-grained control, they can set it at the namespace granularity.

NealSun96: Then the system config should have a default value = "false". When user wants to enable fine-grained control, they shouldn't be manually setting `MULTI_ZK_ENABLED = 'false'` and then setting the per-namespace values; that will only be confusing.

narendly: It is false if it's not set by default.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -136,7 +128,20 @@ public RealmAwareZkClient getRealmAwareZkClient() {
               // Refresh data subscription
               _zkClientForListener.unsubscribeAll();
               _zkClientForListener.subscribeRoutingDataChanges(this, this);
+              LOG.info("ServerContext: subscribed to routing data in routing ZK at {}!", _zkAddr);
```
dasahcc: Why we change the place? Could _zkClientForListener be null? It should be constructed when the ServerContext built.

narendly: I just changed the order so that control flow is not broken if FederatedZkClient creation throws an exception.

You can imagine a scenario in which, in the previous order, if FederatedZkClient creation failed and threw an exception, this whole resubscription logic would be skipped altogether, which is not what we want.

narendly:
This PR is ready to be merged, approved by @NealSun96 

It was observed that we need more fine-grained control over this multiZkEnabled config because there could exists namespaces with differing modes. Because multiple namespaces may be co-deployed, we cannot simply make it a system config.

PR #924, opened 2020-03-31
Fix ZkHelixPropertyStore loses Zookeeper notification issue

### LINKED ISSUE #921: ZkHelixPropertyStore miss ZK notification
ZkHelixPropertyStore misses ZK notification after Zk session expires.

The root cause of this issue is that Share ZkClient fireAllEvent() would not be triggered after session expires.

kaisun2000:
### Issues

Fixes #921 

### Description

    ZkHelixPropertyStore loses ZK notification after session expires.
    The issue was caused by a bug in Share ZkClient code path. More
    specifically, Share ZkClient would not call fireAllEvent when ZK
    session expires. Thus, ZkHelixPropertyStore would not install
    watches for corresponding ZkPath. Thus, lose Zookeeper
    nofiticaition when changes happens.

### Tests

testSessionExpirationWithSharedZkClient

ksun-mn1:helix-core ksun$ mvn test

Failed tests: 
  TestZkConnectionLost.testLostZkConnection » ThreadTimeout Method org.testng.in...
  TestJobFailureDependence.testWorkflowFailureJobThreshold » ThreadTimeout Metho...
  TestHelixAdminCli.testInstanceOperations:469 » ZkClient Failed to delete /Test...

Tests run: 1093, Failures: 3, Errors: 0, Skipped: 4

[INFO] BUILD FAILURE
[INFO] Total time:  01:37 h
[INFO] Finished at: 2020-04-06T18:55:19-07:00
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.18.1:test (default-test) on project helix-core: There are test failures.
[ERROR] 
[ERROR] Please refer to /Users/ksun/dev_branch_helix/helix/helix-core/target/surefire-reports for the individual test results.
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] 

Individual test case run will all succeed.
### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java
@@ -72,6 +72,11 @@ public class ZkCacheBaseDataAccessor<T> implements HelixPropertyStore<T> {
 
   private HelixZkClient _zkClient = null;
 
+  // test only
+  public HelixZkClient getZkClient() {
+    return _zkClient;
+  }
+
   public ZkCacheBaseDataAccessor(ZkBaseDataAccessor<T> baseAccessor, List<String> wtCachePaths) {
     this(baseAccessor, null, wtCachePaths, null);
   }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java
@@ -72,11 +72,6 @@ public class ZkCacheBaseDataAccessor<T> implements HelixPropertyStore<T> {
 
   private HelixZkClient _zkClient = null;
 
-  // test only
-  public HelixZkClient getZkClient() {
-    return _zkClient;
-  }
-
   public ZkCacheBaseDataAccessor(ZkBaseDataAccessor<T> baseAccessor, List<String> wtCachePaths) {
     this(baseAccessor, null, wtCachePaths, null);
   }
--- a/helix-core/src/main/java/org/apache/helix/store/zk/ZkHelixPropertyStore.java
+++ b/helix-core/src/main/java/org/apache/helix/store/zk/ZkHelixPropertyStore.java
@@ -23,6 +23,7 @@ import java.util.List;
 
 import org.apache.helix.manager.zk.ZkBaseDataAccessor;
 import org.apache.helix.manager.zk.ZkCacheBaseDataAccessor;
+import org.apache.helix.zookeeper.api.client.HelixZkClient;
 import org.apache.helix.zookeeper.zkclient.serialize.ZkSerializer;
 
 
@@ -46,5 +47,10 @@ public class ZkHelixPropertyStore<T> extends ZkCacheBaseDataAccessor<T> {
   public ZkHelixPropertyStore(String zkAddress, ZkSerializer serializer, String chrootPath,
       ZkBaseDataAccessor.ZkClientType zkClientType) {
     super(zkAddress, serializer, chrootPath, null, null, MONITOR_TYPE, chrootPath, zkClientType);
+  }
+
+  // test only
+  protected HelixZkClient getTestZkClient() {
+    return getZkClient();
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/store/zk/ZkHelixPropertyStore.java
+++ b/helix-core/src/main/java/org/apache/helix/store/zk/ZkHelixPropertyStore.java
@@ -48,9 +48,4 @@ public class ZkHelixPropertyStore<T> extends ZkCacheBaseDataAccessor<T> {
       ZkBaseDataAccessor.ZkClientType zkClientType) {
     super(zkAddress, serializer, chrootPath, null, null, MONITOR_TYPE, chrootPath, zkClientType);
   }
-
-  // test only
-  protected HelixZkClient getTestZkClient() {
-    return getZkClient();
-  }
 }
--- a/helix-core/src/main/java/org/apache/helix/store/zk/ZkHelixPropertyStore.java
+++ b/helix-core/src/main/java/org/apache/helix/store/zk/ZkHelixPropertyStore.java
@@ -23,7 +23,6 @@ import java.util.List;
 
 import org.apache.helix.manager.zk.ZkBaseDataAccessor;
 import org.apache.helix.manager.zk.ZkCacheBaseDataAccessor;
-import org.apache.helix.zookeeper.api.client.HelixZkClient;
 import org.apache.helix.zookeeper.zkclient.serialize.ZkSerializer;
 
 
--- a/helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
+++ b/helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
@@ -146,8 +146,9 @@ public class ZkTestHelper {
     LOG.info("After expiry. sessionId: " + Long.toHexString(curZookeeper.getSessionId()));
   }
 
-  public static void expireSharedZkClientSession(HelixZkClient client) throws Exception {
+  public static void expireSession(HelixZkClient client) throws Exception {
     final CountDownLatch waitExpireSession = new CountDownLatch(1);
+    final CountDownLatch waitNewSession = new CountDownLatch(1);
     final ZkClient zkClient = (ZkClient) client;
 
     IZkStateListener listener = new IZkStateListener() {
@@ -156,6 +157,9 @@ public class ZkTestHelper {
         LOG.info("IZkStateListener#handleStateChanged, state: " + state);
         if (state == KeeperState.Expired) {
           waitExpireSession.countDown();
+        }
+        if (state == KeeperState.SyncConnected) {
+          waitNewSession.countDown();
         }
       }
 
@@ -195,71 +199,6 @@ public class ZkTestHelper {
 
     // make sure session expiry really happens
     waitExpireSession.await();
-    zkClient.unsubscribeStateChanges(listener);
-
-    connection = (ZkConnection) zkClient.getConnection();
-    curZookeeper = connection.getZookeeper();
-
-    // wait util connected
-    while (curZookeeper.getState() != States.CONNECTED) {
-      Thread.sleep(10);
-    }
-    String newSessionId = Long.toHexString(curZookeeper.getSessionId());
-    LOG.info("After session expiry. sessionId: " + newSessionId + ", zk: " + curZookeeper);
-    Assert.assertFalse(newSessionId.equals(oldSessionId),
-        "Fail to expire current session, zk: " + curZookeeper);
-  }
-
-  public static void expireSession(HelixZkClient client) throws Exception {
-    final CountDownLatch waitNewSession = new CountDownLatch(1);
-    final ZkClient zkClient = (ZkClient) client;
-
-    IZkStateListener listener = new IZkStateListener() {
-      @Override
-      public void handleStateChanged(KeeperState state) throws Exception {
-        LOG.info("IZkStateListener#handleStateChanged, state: " + state);
-      }
-
-      @Override
-      public void handleNewSession(final String sessionId) throws Exception {
-        // make sure zkclient is connected again
-        zkClient.waitUntilConnected(HelixZkClient.DEFAULT_CONNECTION_TIMEOUT, TimeUnit.SECONDS);
-
-        LOG.info("handleNewSession. sessionId: {}.", sessionId);
-        waitNewSession.countDown();
-      }
-
-      @Override
-      public void handleSessionEstablishmentError(Throwable var1) throws Exception {
-      }
-    };
-
-    zkClient.subscribeStateChanges(listener);
-
-    ZkConnection connection = ((ZkConnection) zkClient.getConnection());
-    ZooKeeper curZookeeper = connection.getZookeeper();
-    String oldSessionId = Long.toHexString(curZookeeper.getSessionId());
-    LOG.info("Before session expiry. sessionId: " + oldSessionId + ", zk: " + curZookeeper);
-
-    Watcher watcher = new Watcher() {
-      @Override
-      public void process(WatchedEvent event) {
-        LOG.info("Watcher#process, event: " + event);
-      }
-    };
-
-    final ZooKeeper dupZookeeper =
-        new ZooKeeper(connection.getServers(), curZookeeper.getSessionTimeout(), watcher,
-            curZookeeper.getSessionId(), curZookeeper.getSessionPasswd());
-    // wait until connected, then close
-    while (dupZookeeper.getState() != States.CONNECTED) {
-      Thread.sleep(10);
-    }
-    Assert.assertEquals(dupZookeeper.getState(), States.CONNECTED,
-        "Fail to connect to zk using current session info");
-    dupZookeeper.close();
-
-    // make sure session expiry really happens
     waitNewSession.await();
     zkClient.unsubscribeStateChanges(listener);
 
--- a/helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
+++ b/helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
@@ -210,7 +210,6 @@ public class ZkTestHelper {
         "Fail to expire current session, zk: " + curZookeeper);
   }
 
-
   public static void expireSession(HelixZkClient client) throws Exception {
     final CountDownLatch waitNewSession = new CountDownLatch(1);
     final ZkClient zkClient = (ZkClient) client;
--- a/helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
+++ b/helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
@@ -146,6 +146,71 @@ public class ZkTestHelper {
     LOG.info("After expiry. sessionId: " + Long.toHexString(curZookeeper.getSessionId()));
   }
 
+  public static void expireSharedZkClientSession(HelixZkClient client) throws Exception {
+    final CountDownLatch waitExpireSession = new CountDownLatch(1);
+    final ZkClient zkClient = (ZkClient) client;
+
+    IZkStateListener listener = new IZkStateListener() {
+      @Override
+      public void handleStateChanged(KeeperState state) throws Exception {
+        LOG.info("IZkStateListener#handleStateChanged, state: " + state);
+        if (state == KeeperState.Expired) {
+          waitExpireSession.countDown();
+        }
+      }
+
+      @Override
+      public void handleNewSession(final String sessionId) throws Exception {
+      }
+
+      @Override
+      public void handleSessionEstablishmentError(Throwable var1) throws Exception {
+      }
+    };
+
+    zkClient.subscribeStateChanges(listener);
+
+    ZkConnection connection = ((ZkConnection) zkClient.getConnection());
+    ZooKeeper curZookeeper = connection.getZookeeper();
+    String oldSessionId = Long.toHexString(curZookeeper.getSessionId());
+    LOG.info("Before session expiry. sessionId: " + oldSessionId + ", zk: " + curZookeeper);
+
+    Watcher watcher = new Watcher() {
+      @Override
+      public void process(WatchedEvent event) {
+        LOG.info("Watcher#process, event: " + event);
+      }
+    };
+
+    final ZooKeeper dupZookeeper =
+        new ZooKeeper(connection.getServers(), curZookeeper.getSessionTimeout(), watcher,
+            curZookeeper.getSessionId(), curZookeeper.getSessionPasswd());
+    // wait until connected, then close
+    while (dupZookeeper.getState() != States.CONNECTED) {
+      Thread.sleep(10);
+    }
+    Assert.assertEquals(dupZookeeper.getState(), States.CONNECTED,
+        "Fail to connect to zk using current session info");
+    dupZookeeper.close();
+
+    // make sure session expiry really happens
+    waitExpireSession.await();
+    zkClient.unsubscribeStateChanges(listener);
+
+    connection = (ZkConnection) zkClient.getConnection();
+    curZookeeper = connection.getZookeeper();
+
+    // wait util connected
+    while (curZookeeper.getState() != States.CONNECTED) {
+      Thread.sleep(10);
+    }
+    String newSessionId = Long.toHexString(curZookeeper.getSessionId());
+    LOG.info("After session expiry. sessionId: " + newSessionId + ", zk: " + curZookeeper);
+    Assert.assertFalse(newSessionId.equals(oldSessionId),
+        "Fail to expire current session, zk: " + curZookeeper);
+  }
+
+
   public static void expireSession(HelixZkClient client) throws Exception {
     final CountDownLatch waitNewSession = new CountDownLatch(1);
     final ZkClient zkClient = (ZkClient) client;
--- a/helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
+++ b/helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
@@ -147,7 +147,6 @@ public class ZkTestHelper {
   }
 
   public static void expireSession(HelixZkClient client) throws Exception {
-    final CountDownLatch waitExpireSession = new CountDownLatch(1);
     final CountDownLatch waitNewSession = new CountDownLatch(1);
     final ZkClient zkClient = (ZkClient) client;
 
@@ -155,16 +154,15 @@ public class ZkTestHelper {
       @Override
       public void handleStateChanged(KeeperState state) throws Exception {
         LOG.info("IZkStateListener#handleStateChanged, state: " + state);
-        if (state == KeeperState.Expired) {
-          waitExpireSession.countDown();
-        }
-        if (state == KeeperState.SyncConnected) {
-          waitNewSession.countDown();
-        }
       }
 
       @Override
       public void handleNewSession(final String sessionId) throws Exception {
+        // make sure zkclient is connected again
+        zkClient.waitUntilConnected(HelixZkClient.DEFAULT_CONNECTION_TIMEOUT, TimeUnit.SECONDS);
+
+        LOG.info("handleNewSession. sessionId: {}.", sessionId);
+        waitNewSession.countDown();
       }
 
       @Override
@@ -198,7 +196,6 @@ public class ZkTestHelper {
     dupZookeeper.close();
 
     // make sure session expiry really happens
-    waitExpireSession.await();
     waitNewSession.await();
     zkClient.unsubscribeStateChanges(listener);
 
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
@@ -40,7 +40,7 @@ import org.testng.annotations.Test;
 public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
   @Test
   public void testSessionExpirationWithSharedZkClient() throws Exception {
-    int CURSTATECNT = 10;
+    int curstateCnt = 10;
     String className = TestHelper.getTestClassName();
     String methodName = TestHelper.getTestMethodName();
     String clusterName = className + "_" + methodName;
@@ -69,7 +69,7 @@ public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
     // create 10 current states using external base accessor
     List<String> paths = new ArrayList<>();
     List<ZNRecord> records = new ArrayList<>();
-    for (int i = 0; i < CURSTATECNT; i++) {
+    for (int i = 0; i < curstateCnt; i++) {
       String path = PropertyPathBuilder
           .instanceCurrentState(clusterName, "localhost_8901", "session_0", "TestDB" + i);
       ZNRecord record = new ZNRecord("TestDB" + i);
@@ -79,7 +79,7 @@ public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
     }
 
     boolean[] success = extBaseAccessor.createChildren(paths, records, AccessOption.PERSISTENT);
-    for (int i = 0; i < CURSTATECNT; i++) {
+    for (int i = 0; i < curstateCnt; i++) {
       Assert.assertTrue(success[i], "Should succeed in create: " + paths.get(i));
     }
 
@@ -95,7 +95,7 @@ public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
     ZkTestHelper.asyncExpireSession(dupZkclient);
     ret = TestHelper.verify(() -> {
       long curSessionId = dupZkclient.getSessionId();
-      return curSessionId != sessionId && curSessionId != 0 ? true : false;
+      return curSessionId != sessionId && curSessionId != 0;
     }, 10000);
     Assert.assertTrue(ret, "kill session timed out!");
 
@@ -105,19 +105,19 @@ public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
     ZkTestHelper.asyncExpireSession(dupZkclient);
     ret = TestHelper.verify(() -> {
       long curSessionId = dupZkclient.getSessionId();
-      return curSessionId != sessionId1 && curSessionId != 0 ? true : false;
+      return curSessionId != sessionId1 && curSessionId != 0;
     }, 10000);
     Assert.assertTrue(ret, "kill session second time timed out!");
 
     // remove the currentstates
     paths.clear();
-    for (int i = 0; i < CURSTATECNT; i++) {
+    for (int i = 0; i < curstateCnt; i++) {
       String path = PropertyPathBuilder
           .instanceCurrentState(clusterName, "localhost_8901", "session_0", "TestDB" + i);
       paths.add(path);
     }
     success = extBaseAccessor.remove(paths, 0);
-    for (int i = 0; i < CURSTATECNT; i++) {
+    for (int i = 0; i < curstateCnt; i++) {
       Assert.assertTrue(success[i], "Should succeed in remove:" + paths.get(i));
     }
 
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
@@ -91,30 +91,29 @@ public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
         .buildZkClient(new HelixZkClient.ZkConnectionConfig(ZK_ADDR));
 
     // kill the session to make sure shared zkClient re-installs watcher
-    long sessionId = dupZkclient.getSessionId();
+    final long sessionId = dupZkclient.getSessionId();
     ZkTestHelper.asyncExpireSession(dupZkclient);
-    while (true) {
+    ret = TestHelper.verify(()-> {
       long curSessionId = dupZkclient.getSessionId();
-      if (curSessionId == sessionId || curSessionId == 0) {
-        Thread.sleep(500);
-      } else {
-        sessionId = curSessionId;
-        break;
+      if (curSessionId != sessionId && curSessionId != 0) {
+        return true;
       }
-    }
+      return false;
+    }, 10000);
+    Assert.assertTrue(ret, "kill session timed out!");
 
     // kill the session one more time to cover code path ZkClient resetting flag that
     // indicates first time synconnect happened.
+    final long sessionId1 = dupZkclient.getSessionId();
     ZkTestHelper.asyncExpireSession(dupZkclient);
-    while (true) {
+    ret = TestHelper.verify(()->{
       long curSessionId = dupZkclient.getSessionId();
-      if (curSessionId == sessionId || curSessionId == 0) {
-        Thread.sleep(500);
-      } else {
-        sessionId = curSessionId;
-        break;
+      if (curSessionId != sessionId1 && curSessionId != 0) {
+        return true;
       }
-    }
+      return false;
+    }, 10000);
+    Assert.assertTrue(ret, "kill session second time timed out!");
 
     // remove the currentstates
     paths.clear();
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
@@ -91,11 +91,30 @@ public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
         .buildZkClient(new HelixZkClient.ZkConnectionConfig(ZK_ADDR));
 
     // kill the session to make sure shared zkClient re-installs watcher
-    ZkTestHelper.expireSession(dupZkclient);
+    long sessionId = dupZkclient.getSessionId();
+    ZkTestHelper.asyncExpireSession(dupZkclient);
+    while (true) {
+      long curSessionId = dupZkclient.getSessionId();
+      if (curSessionId == sessionId || curSessionId == 0) {
+        Thread.sleep(500);
+      } else {
+        sessionId = curSessionId;
+        break;
+      }
+    }
 
     // kill the session one more time to cover code path ZkClient resetting flag that
     // indicates first time synconnect happened.
-    ZkTestHelper.expireSession(dupZkclient);
+    ZkTestHelper.asyncExpireSession(dupZkclient);
+    while (true) {
+      long curSessionId = dupZkclient.getSessionId();
+      if (curSessionId == sessionId || curSessionId == 0) {
+        Thread.sleep(500);
+      } else {
+        sessionId = curSessionId;
+        break;
+      }
+    }
 
     // remove the currentstates
     paths.clear();
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
@@ -70,8 +70,8 @@ public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
     List<String> paths = new ArrayList<>();
     List<ZNRecord> records = new ArrayList<>();
     for (int i = 0; i < CURSTATECNT; i++) {
-      String path = PropertyPathBuilder.instanceCurrentState(clusterName, "localhost_8901",
-          "session_0", "TestDB" + i);
+      String path = PropertyPathBuilder
+          .instanceCurrentState(clusterName, "localhost_8901", "session_0", "TestDB" + i);
       ZNRecord record = new ZNRecord("TestDB" + i);
 
       paths.add(path);
@@ -83,8 +83,8 @@ public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
       Assert.assertTrue(success[i], "Should succeed in create: " + paths.get(i));
     }
 
-    TestHelper.verifyWithTimeout("verifyZkCache", 5000,
-        zkCacheInitPaths, accessor._zkCache._cache, _gZkClient, true  );
+    TestHelper.verifyWithTimeout("verifyZkCache", 5000, zkCacheInitPaths, accessor._zkCache._cache,
+        _gZkClient, true);
 
     // dup shared ZkClient
     HelixZkClient dupZkclient = SharedZkClientFactory.getInstance()
@@ -93,12 +93,9 @@ public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
     // kill the session to make sure shared zkClient re-installs watcher
     final long sessionId = dupZkclient.getSessionId();
     ZkTestHelper.asyncExpireSession(dupZkclient);
-    ret = TestHelper.verify(()-> {
+    ret = TestHelper.verify(() -> {
       long curSessionId = dupZkclient.getSessionId();
-      if (curSessionId != sessionId && curSessionId != 0) {
-        return true;
-      }
-      return false;
+      return curSessionId != sessionId && curSessionId != 0 ? true : false;
     }, 10000);
     Assert.assertTrue(ret, "kill session timed out!");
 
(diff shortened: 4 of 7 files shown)


helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
```diff
@@ -88,6 +92,57 @@ public void afterClass() {
...
+    ZkTestHelper.expireSession(testClient);
+
+    listener.reset();
+    int expectDeleteNodes = 1 + firstLevelNr + firstLevelNr * secondLevelNr;
+    store.remove("/", 0);
+    // wait until all callbacks have been received
+    for (int i = 0; i < 10; i++) {
+      if (listener._deleteKeys.size() == expectDeleteNodes)
+        break;
+      Thread.sleep(500);
```
pkuwm: `sleep()` doesn't look solid and makes the test unstable. I suggest using `TestHelper.varify()` to replace sleep().

helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java
```diff
@@ -72,6 +72,11 @@
 
   private HelixZkClient _zkClient = null;
 
+  // test only
+  public HelixZkClient getZkClient() {
```
pkuwm: This should be introduced. If getZkClient() is really necessary, you can make `_zkClient` as protected.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1073,10 +1073,24 @@ protected void processStateChanged(WatchedEvent event) {
 
     fireStateChangedEvent(event.getState());
 
+    // This following is the case for sharedZkClient
     if (!isManagingZkConnection()) {
+      /*
```
pkuwm: Comment block format:
```
/*
 * comments
 */
```

helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
```diff
@@ -88,6 +92,57 @@ public void afterClass() {
...
+    setNodes(store, 'a', true);
+
+    // kill the session to make sure shared zkClient re-installs watcher
+    // Note, current shared ZkClient does not issue new session. Thus, we need
+    // kill _zkConnectionManager's session. Otherwise, ZkTestHelper.expireSesson
+    // would not work.
+    SharedZkClient sharedClient = (SharedZkClient) store.getTestZkClient();
+    HelixZkClient  testClient =  sharedClient.getConnectionManager();
+    ZkTestHelper.expireSession(testClient);
+
```
pkuwm: Remove extra empty line.

helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
```diff
@@ -88,6 +92,57 @@ public void afterClass() {
...
+    // kill the session one more time to cover Shared ZkClient resetting flag
+    // indicating first time synconnect happened.
+    ZkTestHelper.expireSession(testClient);
+
+    listener.reset();
+    int expectDeleteNodes = 1 + firstLevelNr + firstLevelNr * secondLevelNr;
+    store.remove("/", 0);
+    // wait until all callbacks have been received
+    for (int i = 0; i < 10; i++) {
+      if (listener._deleteKeys.size() == expectDeleteNodes)
```
pkuwm: Style:
```
if () {
}
```

helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
```diff
@@ -88,6 +92,57 @@ public void afterClass() {
...
+    // test dataCreate callbacks
+    listener.reset();
+    setNodes(store, 'a', true);
+
+    // kill the session to make sure shared zkClient re-installs watcher
+    // Note, current shared ZkClient does not issue new session. Thus, we need
+    // kill _zkConnectionManager's session. Otherwise, ZkTestHelper.expireSesson
+    // would not work.
+    SharedZkClient sharedClient = (SharedZkClient) store.getTestZkClient();
+    HelixZkClient  testClient =  sharedClient.getConnectionManager();
```
pkuwm: Extra spaces.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java
narendly: This should *not* be introduced.

helix-core/src/main/java/org/apache/helix/store/zk/ZkHelixPropertyStore.java
```diff
@@ -47,4 +48,9 @@ public ZkHelixPropertyStore(String zkAddress, ZkSerializer serializer, String ch
       ZkBaseDataAccessor.ZkClientType zkClientType) {
     super(zkAddress, serializer, chrootPath, null, null, MONITOR_TYPE, chrootPath, zkClientType);
   }
+
+  // test only
+  protected HelixZkClient getTestZkClient() {
+    return getZkClient();
+  }
```
narendly: Exposing the internal HelixZkClient is not desirable

helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
narendly: +1. This is not a show of good craftsmanship..

helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
```diff
@@ -88,6 +92,57 @@ public void afterClass() {
...
+    store.remove("/", 0);
+    // wait until all callbacks have been received
+    for (int i = 0; i < 10; i++) {
+      if (listener._deleteKeys.size() == expectDeleteNodes)
+        break;
+      Thread.sleep(500);
+    }
+
+    System.out.println("createKey#:" + listener._createKeys.size() + ", changeKey#:"
+        + listener._changeKeys.size() + ", deleteKey#:" + listener._deleteKeys.size());
```
narendly: Possible to avoid print statements?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/SharedZkClient.java
```diff
@@ -112,4 +112,9 @@ public String create(final String path, Object datat, final List<ACL> acl,
   protected boolean isManagingZkConnection() {
     return false;
   }
+
+  // test only
+  public HelixZkClient getConnectionManager() {
+    return _connectionManager;
+  }
```
narendly: Can you avoid adding public methods like this that shouldn't otherwise be used in production?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1073,10 +1073,24 @@ protected void processStateChanged(WatchedEvent event) {
...
+       Invoking fireAllEvents is important to make sure HelixPropertyStore working correctly.
+      */
+      if ((event.getState() == KeeperState.SyncConnected) && (!_isNewSessionEventFired) && (!"0"
+          .equals(getHexSessionId()))) {
+        fireAllEvents();
+        _isNewSessionEventFired = true;
+      } else if (event.getState() == KeeperState.Expired) {
+        _isNewSessionEventFired = false;
+      }
       return;
```
narendly: In general, it's not good practice to add subclass-specific logic in its parent class. Can you think of another way to do this?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1073,10 +1073,24 @@ protected void processStateChanged(WatchedEvent event) {
...
     }
 
+    // The following is the case for non sharedZkClient
```
narendly: parent class shouldn't know about what its sub classes are doing?

helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
```diff
@@ -88,6 +92,57 @@ public void afterClass() {
...
   }
 
+  @Test
+  public void testSessionExpirationWithSharedZkClient() throws Exception {
+    /*
+    This test is to make sure SharedZkClient would work with
+    HelixPropertyStore when session expiration happens. More specifically
+    HelixPropertyStore would register all the callbacks after session
+    expiration.
+     */
```
narendly: This is not how you would enter JavaDoc in a public method. Could you please review the javadoc best practice guidelines?

helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
```diff
@@ -88,6 +92,57 @@ public void afterClass() {
...
+
+    // kill the session to make sure shared zkClient re-installs watcher
+    // Note, current shared ZkClient does not issue new session. Thus, we need
+    // kill _zkConnectionManager's session. Otherwise, ZkTestHelper.expireSesson
+    // would not work.
+    SharedZkClient sharedClient = (SharedZkClient) store.getTestZkClient();
+    HelixZkClient  testClient =  sharedClient.getConnectionManager();
+    ZkTestHelper.expireSession(testClient);
+
+
```
narendly: empty line?

helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
kaisun2000: This is copied from the other test. I can change this a little bit.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1073,10 +1073,24 @@ protected void processStateChanged(WatchedEvent event) {
...
+       This keeps the same behavior as original SharedZkClient.
+       Invoking fireAllEvents is important to make sure HelixPropertyStore working correctly.
+      */
+      if ((event.getState() == KeeperState.SyncConnected) && (!_isNewSessionEventFired) && (!"0"
+          .equals(getHexSessionId()))) {
+        fireAllEvents();
+        _isNewSessionEventFired = true;
+      } else if (event.getState() == KeeperState.Expired) {
+        _isNewSessionEventFired = false;
+      }
```
pkuwm: If we remove ` _isNewSessionEventFired = false;` in `reconnect()` and do below code, it seems more concise and easier to maintain. `_isNewSessionEventFired` is only changed in this code block. What do you think?

```java
    if (event.getState() == KeeperState.SyncConnected && !_isNewSessionEventFired && !"0".equals(getHexSessionId())) {
      if (isManagingZkConnection()) {
        /*
         * Before the new zookeeper instance is connected to the zookeeper service and its session
         * is established, its session id is 0.
         * New session event is not fired until the new zookeeper session receives the first
         * SyncConnected state(the zookeeper session is established).
         * Now the session id is available and non-zero, and we can fire new session events.
         */
        fireNewSessionEvents();
      }
      /*
       * Set it true to avoid firing events again for the same session next time
       * when SyncConnected events are received.
       */
      _isNewSessionEventFired = true;

      /*
       * With this first SyncConnected state, we just get connected to zookeeper service after
       * reconnecting when the session expired. Because previous session expired, we also have to
       * notify all listeners that something might have changed.
       */
      fireAllEvents();
    } else if (event.getState() == KeeperState.Expired) {
      if (isManagingZkConnection()) {
        reconnectOnExpiring();
      }
    }
```

helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java
kaisun2000: Protect is not going to work, as the test case live in another package. 

Java has not way to access otherwise private members in the class. In this case, we have to expire the session inside this class. This is a common pattern I did not see addressed in our code base. Unless you guys can identify another pattern, I guess we have to leave it this way.

helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
kaisun2000: Java has not way to access otherwise private members in the class. In this case, we have to expire the session inside this class. This is a common problem I did not see addressed in our code base. Unless you guys can identify another pattern, I guess we have to leave it this way.

helix-core/src/main/java/org/apache/helix/store/zk/ZkHelixPropertyStore.java
kaisun2000: Java has not way to access otherwise private members in the class. In this case, we have to expire the session inside this class. This is a common problem I did not see addressed in our code base. Unless you guys can identify another pattern, I guess we have to leave it this way.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/SharedZkClient.java
kaisun2000: Java has not way to access otherwise private members in the class. In this case, we have to expire the session inside this class. This is a common problem I did not see addressed in our code base. Unless you guys can identify another pattern, I guess we have to leave it this way.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: Strictly speaking, this is not from sub-class. This is one object of a class that can act on different roles. But I get what you are thinking.

That said, traditionally it is like this. 

I can write in an another concise way wrapping all this logic into the if block of 1093. 

Thought about it. The drawback of that way would be losing readability, and thus the intention of the logic. This would pave the way of next hidden bug when people do refactoring later, just as how this bug is introduced.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java
lei-xia: +1, we should not add a public method to a public class just for testing, specially this is a public API.  If you test class is sitting in the same module with this class, protected will work.  Another way is to make it protected, and wrap ZkCacheBaseDataAccessor to a new test class.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: See my comment in the other similar comment. We can discuss it over there.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/SharedZkClient.java
lei-xia: Agreed here, we should never add a public method to a public class just for testing.  If you really need to expose ZkClient, either change it to protected, and inherit this class in your test, such that:

class SharedZkClient {
   protected _zkClient;
}

class testSharedZkClient extends SharedZkClient {

 Protected getZkClient() {
      return _zkClient;
 }

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: Good point. 
Yes, this is another way to do it, thought about it.
Pro: concise
Con: hard to understand the intention. This may pave the way to another hidden bug when people do refactoring next time, as how we get to this issue.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/SharedZkClient.java
lei-xia: Another way is, if your test class is put in the same module with sharedZkClient, such as org.apache.helix.zookeeper.impl.client,  you should be able to access protected instance variables or methods from the test code.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
lei-xia: I think we have the format style file? You can apply the style file in intellij.

helix-core/src/main/java/org/apache/helix/store/zk/ZkHelixPropertyStore.java
kaisun2000: We have another thread of this discussion in this pull. Close this one

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: I applied the style file. And it did not pick this up.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/SharedZkClient.java
kaisun2000: Here is the case:

test case is in package org.apache.helix.store.zk, here is the how it is constructed.

`    ZkSerializer serializer = new ZNRecordSerializer();
    ZkHelixPropertyStore<ZNRecord> store =
        new ZkHelixPropertyStore<>(ZK_ADDR, serializer, subRoot, subscribedPaths);`

We need the internal shared ZkClient in the store; and further the internal _ZkconnectionManager.

Note, ZkHelixPropertyStore uses ZkCacheBasedDataAccessor which lives in package org.apache.helix.manager.zk;

Note, ShareZkClient and ZkConnectionManager lives in package org.apache.helix.zookeeper.impl.client;

So the solution (in this case, but not in general) can be:

I can put the test in the org.apache.helix.zookeeper.impl.client, zookeeper-api module

I can also initialize the HelixPropertystore like this:

1/ construct SharedZkClient 

2/ construct ZkBaseDataAccessor using this sharedZkCLient.

3/ construct ZkHelixPropertyStore (using ZkBaseDataAccessor.

by making SharedZkClient's _zkConnectionManager protected, we can use it here.  

The problem is that org.apache.helix.zookeeper.impl.client is in Zookeeper-api, we need a test scope dependency to Helix-core and Helix-common module. Does this test dependency work here?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: To me, it is easier to understand:
1. only managed connection will do: fireNewSessionEvents() and reconnectOnExpiring.
2. first sync connected: _isNewSessionEventFired = true; expired: _isNewSessionEventFired = false.

kaisun2000:
@jiajunwang 

> I see the main issue is about the additional public method for fetching the client. I don't like that either.
> An alternative way is that, since you are for sure given a shared zkclient in that property store, you can call ShareZkClientFactory with the same parameters and it will return you with another shared ZkClient instance. But, since they will use the same connection manager, when you expires the newly created client, the one that in the HelixPropertyStore will also be expired for once.
> Can you have a try? In this way, no new method is required.

How do we get the zkconnection manager from the new sharedZkClient? It is not public. 

If we make zkConnectionManager protected, we need to access it from the module in zookeeper-api, which means move (rewrite) this test case there. But this case use the code from HelixPropertyStore, which is helix-core. This is circular dependency.

jiajunwang:
> @jiajunwang
> 
> > I see the main issue is about the additional public method for fetching the client. I don't like that either.
> > An alternative way is that, since you are for sure given a shared zkclient in that property store, you can call ShareZkClientFactory with the same parameters and it will return you with another shared ZkClient instance. But, since they will use the same connection manager, when you expires the newly created client, the one that in the HelixPropertyStore will also be expired for once.
> > Can you have a try? In this way, no new method is required.
> 
> How do we get the zkconnection manager from the new sharedZkClient? It is not public.
> 
> If we make zkConnectionManager protected, we need to access it from the module in zookeeper-api, which means move (rewrite) this test case there. But this case use the code from HelixPropertyStore, which is helix-core. This is circular dependency.

Why you need to get zkconnection manager? The object you operate on is the send shared zkclient that you get from the factory.

kaisun2000:
As discussed offline, ZkTestHelper.expireSession() only works with zkConnectionmanager, but not shared zkclient. 

But we can add another expireSharedZkClientSession() though for ZkTestHelper.

helix-core/src/test/java/org/apache/helix/store/zk/TestZkHelixPropertyStore.java
kaisun2000: New approach use TestHelper.verifyWithTimeout().

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: @jiajunwang, what is your take here?

helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
```diff
@@ -146,6 +146,70 @@ public void process(WatchedEvent event) {
...
+  public static void expireSharedZkClientSession(HelixZkClient client) throws Exception {
+    final CountDownLatch waitExpireSession = new CountDownLatch(1);
+    final ZkClient zkClient = (ZkClient) client;
+
+    IZkStateListener listener = new IZkStateListener() {
+      @Override
+      public void handleStateChanged(KeeperState state) throws Exception {
+        LOG.info("IZkStateListener#handleStateChanged, state: " + state);
+        if (state == KeeperState.Expired) {
+          waitExpireSession.countDown();
```
jiajunwang, line 158: I guess this wait only check if there is an expiring happened. The original wait is to ensure the session has been re-established. So if you want to achieve the same goal, maybe we need to compare the older session and newer session then end the wait.
Given that saying, I guess you can just modify the existing method to include this check. Note that if this not happen, we can directly fail the test by throwing an exception.

kaisun2000: You are right, here this latch guarantees expiring happens. But later, the code still ensures new session established.

`     waitExpireSession.await();
    zkClient.unsubscribeStateChanges(listener);

    connection = (ZkConnection) zkClient.getConnection();
    curZookeeper = connection.getZookeeper();

    // wait util connected
    while (curZookeeper.getState() != States.CONNECTED) {
      Thread.sleep(10);
    } `

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1073,10 +1073,24 @@ protected void processStateChanged(WatchedEvent event) {
...
       return;
     }
 
+    // The following is the case for ZkClient managing the connection
```
lei-xia, line 1093: This comment is a bit confusing.  You mean "The following only applies to dedicated ZkClient in which it manages ZkConnection directly", right?

helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
jiajunwang: This work. Since it should work for all the client, can we just modify the original expiring method? 2 similar methods for the same purpose is not good.

helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
```diff
@@ -146,6 +146,70 @@ public void process(WatchedEvent event) {
...
+
+    // make sure session expiry really happens
+    waitExpireSession.await();
+    zkClient.unsubscribeStateChanges(listener);
+
+    connection = (ZkConnection) zkClient.getConnection();
+    curZookeeper = connection.getZookeeper();
+
+    // wait util connected
+    while (curZookeeper.getState() != States.CONNECTED) {
```
jiajunwang, line 204: Can we subscribe to the state change instead of busy loop?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: I like less code : )

And actually I can understand the concise version. So I guess it's better.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: I guess it can be simplified to:
// else, if managing the connection

Anyway, if we do what Huizhi suggested, it should be clearer.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: The I will use this concise version. I will add a verbose note about the intention too.

helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
kaisun2000: There are many ways to do it. Let us be specific to avoid delaying this important fix. 
So the code can be something like this:

  ```
public void handleStateChanged(KeeperState state) throws Exception {
        LOG.info("IZkStateListener#handleStateChanged, state: " + state);
        if (state == KeeperState.Expired) {
          waitExpireSession.countDown();
        } 
        if (state == KeeperState.Synconnected) {
          waitNewSession.countDown();
        }

```

Later something like this to wait for latch?

```
boolean expiredCompleted = waitExpireSession.await(5L, TimeUnit.SECONDS);
if (!expiredCompleted)  {
   throw Exception(" Expiration of Session failed.");
}
boolean newSessionCompleted = waitNewSession.await(5L, TimeUnit.SECONDS);
if (!newSessionCompleted) {
  throw Exception(" Expiration of Session failed.");
}
zkClient.unsubscribeStateChanges(listener);
```

Note, the wait for expiration and synconnected are both necessary. The reason is that disconnect can give you sysconnected too. 

Note, the timeout in await() is to prevent hanging here forever to block test from proceeding.

File: helix-core/src/test/java/org/apache/helix/ZkTestHelper.java (hunk 15 above)
Comment (kaisun2000) [reply helix-core/src/test/java/org/apache/helix/ZkTestHelper.java:158]: See my comment later about how to enhance with handleStateChanged. If we agree with that approach. I will use this logic in TestHelper.expireSession() and remove this one.

File: zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java (hunk 16 above)
Comment (kaisun2000) [reply zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java:1093]: @lei-xia, right. Your description is more readable. But @jiajunwang and @hulu like the more concise way. Unless you like this separation of two branch way, I will adopt the more concise way. And add another comment about the goal fo what we will achieve. (Intention)

So later if people refactor this code. They know the intention at lease.

File: helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java [hunk 18]
```diff
@@ -37,6 +38,83 @@
 import org.testng.annotations.Test;
 
 public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
+  @Test
+  public void testSessionExpirationWithSharedZkClient() throws Exception {
+    int CURSTATECNT = 10;
```
pkuwm, line 43: If it is just a local var, no need to make it capitalized.

helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
pkuwm: +1 to get rid of busy sleeping.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
```diff
@@ -37,6 +38,83 @@
...
+      paths.add(path);
+      records.add(record);
+    }
+
+    boolean[] success = extBaseAccessor.createChildren(paths, records, AccessOption.PERSISTENT);
+    for (int i = 0; i < CURSTATECNT; i++) {
+      Assert.assertTrue(success[i], "Should succeed in create: " + paths.get(i));
+    }
+
+    TestHelper.verifyWithTimeout("verifyZkCache", 5000,
```
pkuwm, line 86: Should assign the result to ret?

helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
```diff
@@ -146,6 +146,70 @@ public void process(WatchedEvent event) {
     LOG.info("After expiry. sessionId: " + Long.toHexString(curZookeeper.getSessionId()));
   }
 
+  public static void expireSharedZkClientSession(HelixZkClient client) throws Exception {
```
pkuwm, line 149: Naming *Shared*ZkClient doesn’t match the parameter type HelixZkClient.

FYI, it seems the only difference between expireSession and this is, this does not have handleNewSession implementation logic. Would you consider reducing the duplicate code?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
```diff
@@ -37,6 +38,83 @@
...
+      String path = PropertyPathBuilder.instanceCurrentState(clusterName, "localhost_8901",
+          "session_0", "TestDB" + i);
+      paths.add(path);
+    }
+    success = extBaseAccessor.remove(paths, 0);
+    for (int i = 0; i < CURSTATECNT; i++) {
+      Assert.assertTrue(success[i], "Should succeed in remove:" + paths.get(i));
+    }
+
+    TestHelper.verifyWithTimeout("verifyZkCache", 5000,
```
pkuwm, line 114: Assert this?

helix-core/src/test/java/org/apache/helix/ZkTestHelper.java
kaisun2000: It is different for a reason. Shared ZkClient does not have handleNewSession. Will combine see the above discussion with JJ.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
kaisun2000: good point.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
kaisun2000: See above.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: By they way, setting _isNewSessionEventFired = false should be above reconnect(). Otherwise, there is a small chance of race condition. In reconnect() or current code, we still have this race condition.
```
    } else if (event.getState() == KeeperState.Expired) {
      if (isManagingZkConnection()) {
        reconnectOnExpiring();
      }
    }
```

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1101,7 +1104,10 @@ protected void processStateChanged(WatchedEvent event) {
         fireAllEvents();
       }
     } else if (event.getState() == KeeperState.Expired) {
-      reconnectOnExpiring();
+      _isNewSessionEventFired = false;
```
jiajunwang, line 1107: What if reconnect is not done successfully in the following reconnectOnExpiring() call?
In that case, if we change this flag to false, will it cause additional fireNewSessionEvents()? For example if we have a stale new session event in the queue.

jiajunwang: Discussed wit Kai offline, this should be fine as long as processStateChanged is synchronized called in one thread.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1101,7 +1104,10 @@ protected void processStateChanged(WatchedEvent event) {
...
+      if (isManagingZkConnection()) {
```
jiajunwang, line 1108: As we moved the isManagingZkConnection closer to these methods, it might be cleaner if we just move it inside reconnectOnExpiring() and fireNewSessionEvents(). So we have better protection and the code looks cleaner.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: BTW, please update the comment of _isNewSessionEventFired, it is out of date.

pkuwm: I was also thinking about the same concern.  It should be fine as it is protected by the eventLock. And also if reconnecting fails because of IOException, the old/expired zk is kept, but the expired zk could not do anything.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
pkuwm: Do you think this should be changed, since I've seen you agreement but no change?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
```diff
@@ -37,6 +38,100 @@
...
+    // kill the session to make sure shared zkClient re-installs watcher
+    long sessionId = dupZkclient.getSessionId();
+    ZkTestHelper.asyncExpireSession(dupZkclient);
+    while (true) {
+      long curSessionId = dupZkclient.getSessionId();
+      if (curSessionId == sessionId || curSessionId == 0) {
+        Thread.sleep(500);
+      } else {
+        sessionId = curSessionId;
+        break;
```
pkuwm, line 102: If you have to use sleep to wait for the sync connected, I do see duplicate code and it could replaced by TestHelper.varify(), which saves you the duplicate code and also offers a timeout option in case it is forever blocked here.

kaisun2000: fixed.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: updated.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: As @pkuwm pointed out, processStateChanged is private method always called under lock protection, the same as fireAllEvent, and reconnectionOnExpiring.

jiajunwang: Yeah, but it is a good practice to check closer to where the real business logic exists.

One example could be, if in the future we want to call reconnectOnExpiring() somewhere else, we don't need to add one more check there.

High-level, these 2 methods shall only be called for non-shared clients. But we don't have a way to block them for shared client implementation. So we just put the check inside, then they won't cause any more problems even if we want to do anything to change the zkclient.

kaisun2000: Can you be more specific as what you want to do with the change? Sketch how you think about the fireNewSessionEvent, fireAllEvent, reconnectOnExpiring look like? 

To me, there are many ways to put the if and else check. Either way works for me.

kaisun2000: maybe something like this:

```
private void reconnectOnExpiring() {
  if (!isManagingZkConnection()) {
    return;
  }
  ... // the rest is the same
}
```

```
private void fireNewSessionEvents() {
  if (!isManagingZkConnection()) {
    return;
  }
 ...// the rest is the same.

  }
```

jiajunwang: Yeah, this is what I meant.

kaisun2000: done.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
```diff
@@ -37,6 +38,99 @@
...
+
+    // kill the session to make sure shared zkClient re-installs watcher
+    final long sessionId = dupZkclient.getSessionId();
+    ZkTestHelper.asyncExpireSession(dupZkclient);
+    ret = TestHelper.verify(()-> {
+      long curSessionId = dupZkclient.getSessionId();
+      if (curSessionId != sessionId && curSessionId != 0) {
+        return true;
+      }
+      return false;
```
pkuwm, line 101: It could be simplified to `return curSessionId != sessionId && curSessionId != 0`?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
```diff
@@ -37,6 +38,99 @@
...
+      records.add(record);
+    }
+
+    boolean[] success = extBaseAccessor.createChildren(paths, records, AccessOption.PERSISTENT);
+    for (int i = 0; i < CURSTATECNT; i++) {
+      Assert.assertTrue(success[i], "Should succeed in create: " + paths.get(i));
+    }
+
+    TestHelper.verifyWithTimeout("verifyZkCache", 5000,
+        zkCacheInitPaths, accessor._zkCache._cache, _gZkClient, true  );
```
pkuwm, line 87: Unnecessary extra space before `)`

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
```diff
@@ -37,6 +38,99 @@
...
+
+    // dup shared ZkClient
+    HelixZkClient dupZkclient = SharedZkClientFactory.getInstance()
+        .buildZkClient(new HelixZkClient.ZkConnectionConfig(ZK_ADDR));
+
+    // kill the session to make sure shared zkClient re-installs watcher
+    final long sessionId = dupZkclient.getSessionId();
+    ZkTestHelper.asyncExpireSession(dupZkclient);
+    ret = TestHelper.verify(()-> {
```
pkuwm, line 96: I've seen multiple places not formatted. Can you format your changes here?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
pkuwm: Style?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1150,6 +1156,10 @@ private void reconnect() {
   }
 
   private void fireNewSessionEvents() {
+    // only managing zkclient fire handleNewSession event
+    if (isManagingZkConnection()) {
```
pkuwm, line 1160: Reversed way? Managing connection should fire handle new session events.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1101,11 +1103,16 @@ protected void processStateChanged(WatchedEvent event) {
...
       }
     } else if (event.getState() == KeeperState.Expired) {
+      _isNewSessionEventFired = false;
       reconnectOnExpiring();
     }
   }
 
   private void reconnectOnExpiring() {
+    // only managing zkclient fire handleNewSession event
+    if (isManagingZkConnection()) {
```
pkuwm, line 1113: Reversed.
And please correct the comment.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1080,22 +1080,21 @@ protected void processStateChanged(WatchedEvent event) {
...
         _isNewSessionEventFired = true;
 
+        /*
+         * Before the new zookeeper instance is connected to the zookeeper service and its session
+         * is established, its session id is 0.
+         * New session event is not fired until the new zookeeper session receives the first
+         * SyncConnected state(the zookeeper session is established).
+         * Now the session id is available and non-zero, and we can fire new session events.
+         */
+        fireNewSessionEvents();
```
jiajunwang, line 1096: Thanks for put the check inside the function. But why we need to move it to after _isNewSessionEventFired = true;?
Is it concerning? Because when _isNewSessionEventFired is set to true, the event is not sent yet.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
kaisun2000: done.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
kaisun2000: formatted.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: Any comments on "why we need to move it to after _isNewSessionEventFired = true;? Is it concerning?" ?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
kaisun2000: changed. 

there are many ways to write same bool logic. there is no preferred way by all the people. (say what if JJ prefer another way, how do we resolve it?) let us refrain from pick this kind stuff up. It incurs cost without adding value.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: I changed it back to the original way.

jiajunwang:
@kaisun2000, please follow the merge steps to finish this PR merge process if you think it is good enough.

@pkuwm, please take another look to ensure this looks good to you.

kaisun2000:

Final message:
    fix ZkHelixPropertyStore loses Zookeeper notification issue
    
    ZkHelixPropertyStore loses ZK notification after session expires.
    THe issue was caused by a bug in Share ZkClient code path. More
    specifically, Share ZkClient would not call fireAllEvent when ZK
    session expires. Thus, ZkHelixPropertyStore would not install
    watches for corresponding ZkPath. Thus, lose Zookeeper
    nofiticaition when changes happens.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
```diff
@@ -37,6 +38,93 @@
...
+    boolean ret =
+        TestHelper.verifyZkCache(zkCacheInitPaths, accessor._zkCache._cache, _gZkClient, true);
+    Assert.assertTrue(ret, "zkCache doesn't match data on Zk");
+
+    // create 10 current states using external base accessor
+    List<String> paths = new ArrayList<>();
+    List<ZNRecord> records = new ArrayList<>();
+    for (int i = 0; i < CURSTATECNT; i++) {
+      String path = PropertyPathBuilder
+          .instanceCurrentState(clusterName, "localhost_8901", "session_0", "TestDB" + i);
```
pkuwm, line 74: @kaisun2000 My understanding is, the current states created by this invalid session "session_0" would also be invalid. Can you help me understand if these invalid current states would work for this testing purpose? Thanks.

kaisun2000: CurrentStates here only serves the purpose of some directory in zookeeper. This test has nothing to do with controller rebalancing logic.  Note, TestHelper.verifyZkCache only valid the data in zookeeper and propertycache for this currentstates path are the same. That is the gist of this test.

Put it another way, using currentstate property builder is just a convenience. All the other test in this suite is like this.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
```diff
@@ -37,6 +38,93 @@
 import org.testng.annotations.Test;
 
 public class TestZkCacheAsyncOpSingleThread extends ZkUnitTestBase {
+  @Test
+  public void testSessionExpirationWithSharedZkClient() throws Exception {
+    int CURSTATECNT = 10;
```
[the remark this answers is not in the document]
kaisun2000: addressed.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
```diff
@@ -37,6 +38,93 @@
...
+    // dup shared ZkClient
+    HelixZkClient dupZkclient = SharedZkClientFactory.getInstance()
+        .buildZkClient(new HelixZkClient.ZkConnectionConfig(ZK_ADDR));
+
+    // kill the session to make sure shared zkClient re-installs watcher
+    final long sessionId = dupZkclient.getSessionId();
+    ZkTestHelper.asyncExpireSession(dupZkclient);
+    ret = TestHelper.verify(() -> {
+      long curSessionId = dupZkclient.getSessionId();
+      return curSessionId != sessionId && curSessionId != 0 ? true : false;
```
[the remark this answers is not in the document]
kaisun2000: addressed.

pkuwm:
@kaisun2000 Just a reminder: please run the tests in an appropriate module before merging the PR. Thanks.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkCacheAsyncOpSingleThread.java
```diff
@@ -37,6 +38,93 @@
...
+    }, 10000);
+    Assert.assertTrue(ret, "kill session second time timed out!");
+
+    // remove the currentstates
+    paths.clear();
+    for (int i = 0; i < curstateCnt; i++) {
+      String path = PropertyPathBuilder
+          .instanceCurrentState(clusterName, "localhost_8901", "session_0", "TestDB" + i);
+      paths.add(path);
+    }
```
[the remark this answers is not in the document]
kaisun2000: This is necessary.  This removal of currentstate should trigger propertystore to reflect this removal in its cache as well. 

Without this fix, this is not going to be true because the watcher to zookeeper is lost.

kaisun2000:
I did run another round of mvn test; same results. 4 flaking test. Running independently they succeed. The same as before this fix and added test,

Issue #932, opened 2020-04-07
Fix flaky tests in TestResourceAccessor


NealSun96:
Two of the tests in `TestResourceAccessor`, namely `testPartitionHealth()` and `testResourceHealth()` have been flaky. The tests will fail if there is an delay to the health check API call, because the newly added external view will be deleted during this delay; the tests were only passing because the API calls were fast enough; at the same time, there is an oddity that adding a delay before external view causes it to remain in ZooKeeper, not deleted. While the exact cause is unclear, one possibility is due to mismatching current states and external views, causing the controller to remove the external view.
Future updates will be posted in this issue.

PR #933, opened 2020-04-07
Fix MetadataStoreDirectory routing data cache refresh bug

### LINKED ISSUE #927: MetadataStoreDirectory routing data cache refresh bug
The current routing data cache refresh behavior in `ZkMetadataStoreDirectory` is incorrect. Using the following steps will cause an erroneous state:
1. Create a valid set of routing data;
2. Remove the routing data using `setRoutingData()`;
3. All "routing data fetching" endpoints will still retain the routing data from step 1; all "using routing data to route" endpoints will still use the routing data from step 1. 

This is due to the internal routing data cache not being cleared before updating in `ZkMetadataStoreDirectory`. In the case when exceptions are raised during updating (which is excepted when routing data is empty/invalid and is correctly handled in `ZkMetadataStoreDirectory`), the old data is not erased.

NealSun96:
### Issues

Fixes #927 

### Description

This PR fixes the MetadataStoreDirectory routing data cache refresh bug. In `ZkMetadataStoreDirectory`, it is now ensured that routing data cache is cleared before updating; when an exception is raised, the old data is still erased correctly. 

Also, this PR fixes an edge case during `TrieRoutingData` creation. When there are zkRealms defined without a single sharding key, `TrieRoutingData` now correctly identifies the raw routing data as invalid and raise an exception. 

### Tests

```
[ERROR] Tests run: 145, Failures: 1, Errors: 0, Skipped: 7, Time elapsed: 28.895 s <<< FAILURE! - in TestSuite
[ERROR] testResourceHealth(org.apache.helix.rest.server.TestResourceAccessor)  Time elapsed: 0.295 s  <<< FAILURE!
java.lang.AssertionError: expected:<HEALTHY> but was:<UNHEALTHY>
	at org.apache.helix.rest.server.TestResourceAccessor.testResourceHealth(TestResourceAccessor.java:289)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestResourceAccessor.testResourceHealth:289 expected:<HEALTHY> but was:<UNHEALTHY>
[INFO] 
[ERROR] Tests run: 145, Failures: 1, Errors: 0, Skipped: 7
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  33.443 s
[INFO] Finished at: 2020-04-03T13:47:22-07:00
```
The failed test is known flaky and is tracked by https://github.com/apache/helix/issues/932. 

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -322,14 +322,19 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
     }
 
     Map<String, List<String>> rawRoutingData;
+    // Remove the raw data first in case of failure on creation
+    _realmToShardingKeysMap.remove(namespace);
     try {
       rawRoutingData = _routingDataReaderMap.get(namespace).getRoutingData();
       _realmToShardingKeysMap.put(namespace, rawRoutingData);
     } catch (InvalidRoutingDataException e) {
+      _routingDataMap.remove(namespace);
       LOG.error("Failed to refresh cached routing data for namespace {}", namespace, e);
       return;
     }
 
+    // Remove routing data first in case of failure on creation
+    _routingDataMap.remove(namespace);
     try {
       _routingDataMap.put(namespace, new TrieRoutingData(rawRoutingData));
     } catch (InvalidRoutingDataException e) {
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -321,20 +321,20 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
       return;
     }
 
-    Map<String, List<String>> rawRoutingData;
     // Remove the raw data first in case of failure on creation
     _realmToShardingKeysMap.remove(namespace);
+    // Remove routing data first in case of failure on creation
+    _routingDataMap.remove(namespace);
+
+    Map<String, List<String>> rawRoutingData;
     try {
       rawRoutingData = _routingDataReaderMap.get(namespace).getRoutingData();
       _realmToShardingKeysMap.put(namespace, rawRoutingData);
     } catch (InvalidRoutingDataException e) {
-      _routingDataMap.remove(namespace);
       LOG.error("Failed to refresh cached routing data for namespace {}", namespace, e);
       return;
     }
 
-    // Remove routing data first in case of failure on creation
-    _routingDataMap.remove(namespace);
     try {
       _routingDataMap.put(namespace, new TrieRoutingData(rawRoutingData));
     } catch (InvalidRoutingDataException e) {
--- a/metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/datamodel/TrieRoutingData.java
+++ b/metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/datamodel/TrieRoutingData.java
@@ -48,6 +48,10 @@ public class TrieRoutingData implements MetadataStoreRoutingData {
       throw new InvalidRoutingDataException("routingData cannot be null or empty");
     }
 
+    if (!containsShardingKey(routingData)) {
+      throw new InvalidRoutingDataException("routingData needs at least 1 sharding key");
+    }
+
     if (isRootShardingKey(routingData)) {
       Map.Entry<String, List<String>> entry = routingData.entrySet().iterator().next();
       _rootNode = new TrieNode(Collections.emptyMap(), "/", true, entry.getKey());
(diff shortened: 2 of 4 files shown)


helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -322,14 +322,19 @@ public void refreshRoutingData(String namespace) {
     }
 
     Map<String, List<String>> rawRoutingData;
+    // Remove the raw data first in case of failure on creation
+    _realmToShardingKeysMap.remove(namespace);
```
NealSun96, line 326: **NOTE**: This check is added for the case when "METADATA_STORE_ROUTING_DATA" is deleted. 
There is one edge case that's *not* handled: when "METADATA_STORE_ROUTING_DATA" contains no node, and then "METADATA_STORE_ROUTING_DATA" gets deleted. In that case, before deletion, the routing data should be empty; after deletion, the routing data should return 404 due to missing namespace. However, since there is no listener on "METADATA_STORE_ROUTING_DATA" itself, its deletion event cannot be handled. Please share you thoughts on whether this edge case should be handled, or if you think this falls into misuse.

narendly: I don't think you need to worry about the case you're describing. MSD makes the assumption that the root path already exists.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -322,14 +322,19 @@ public void refreshRoutingData(String namespace) {
...
       rawRoutingData = _routingDataReaderMap.get(namespace).getRoutingData();
       _realmToShardingKeysMap.put(namespace, rawRoutingData);
     } catch (InvalidRoutingDataException e) {
+      _routingDataMap.remove(namespace);
       LOG.error("Failed to refresh cached routing data for namespace {}", namespace, e);
       return;
     }
 
+    // Remove routing data first in case of failure on creation
+    _routingDataMap.remove(namespace);
```
narendly, line 337: Why don't you just put `_routingDataMap.remove(namespace);` up there next to `_realmToShardingKeysMap.remove(namespace);`?

Why is there a need for this duplicate code? routingDataMap entry should be cleared anyway prior to update, right?

metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/datamodel/TrieRoutingData.java
```diff
@@ -48,6 +48,10 @@ public TrieRoutingData(Map<String, List<String>> routingData)
       throw new InvalidRoutingDataException("routingData cannot be null or empty");
     }
 
+    if (!containsShardingKey(routingData)) {
+      throw new InvalidRoutingDataException("routingData needs at least 1 sharding key");
+    }
```
narendly, line 53: Explain why this should be invalid? I thought we agreed that it's acceptable to have a realm without any keys.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: Makes sense, let me change that.

metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/datamodel/TrieRoutingData.java
NealSun96: We agreed that it's okay for some realms to not have keys (which is handled correctly by skipping them), but not when **all** realms have no keys. The design of `TrieRoutingData` is "a set of data with routing capability", therefore empty routing data shouldn't be created; a group of realms with 0 key doesn't have routing capability either. 
In that situation, the current `TrieRoutingData` will create a structure in an erroneous state. The promise of `TrieRoutingData` is to contain at least 1 sharding key; with 0 sharding keys, some functionality will break. For example, `isShardingKeyInsertionValid("/")` called on a `TrieRoutingData` with 0 sharding key will return false, even though it should be true.

narendly: So it's okay to have a bunch of realms with no keys as long as we have some realm with a key?

NealSun96: Yes. The focus should be on "is there a sharding key". Without any sharding key the routing data is definitely invalid; with at least one it is valid.

narendly: Okay. Is this clear in our documentation, both in code and user manual? We need to document this behavior so it's crystal clear to anyone that looks at this code.

narendly: Also what would happen if a user initially creates N realms with keys, and ends up deleting all keys for the N realms? What happens when the user tries to delete the last remaining key?

NealSun96: In my user manual I have briefly talked about invalid routing data. I will expand some sections for that. 
In terms of your second question, the refresh logic will trigger and `TrieRoutingData` will be removed from MSD when the last key is deleted, similar to how when the user deletes all routing data, `TrieRoutingData` is removed. Note that the raw routing data is not removed because the realms are still there. Also a reminder, when we don't have `TrieRoutingData`, the accessor behaves correctly.

NealSun96:
Final commit message:
## Fix MetadataStoreDirectory routing data cache refresh bug ##
This PR ensures that routing data cache is cleared before updating in ZkMetadataStoreDirectory. Also, this PR fixes an edge case during TrieRoutingData creation when no zk realms has any sharding key.

PR #934, opened 2020-04-08
Add cache update/delete in customized view aggregation stage

zhangmeng916:
### Issues

(#942  )

### Description

This PR adds the customized view cache update/delete operation in customized view aggregation stage. Previously, we depend on routing table provider to update customized view cache, as the controller is NOT listening to customized view change, and we found there're some cases we see inconsistency between cache and zk due to the lag between routing table provider and controller. To handle cache in the customized view aggregation stage could avoid such problems.

Detailed race condition:
Assume the customized state config changes, and it triggers a customized view update, before the router catches the change and processes the customized view change and updates cache, there's another customized state config change happen, in the customized view aggregation stage, we compare the new change with that in the cache, which is stale, and we will make the wrong decision.

### Tests

### Commits

### Code Quality

- [X ] My diff has been formatted using helix-style.xml 

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
@@ -39,6 +39,7 @@ import org.apache.helix.controller.pipeline.Pipeline;
 import org.apache.helix.controller.stages.MissingTopStateRecord;
 import org.apache.helix.model.CustomizedState;
 import org.apache.helix.model.CustomizedStateConfig;
+import org.apache.helix.model.CustomizedView;
 import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.ResourceAssignment;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -257,6 +258,24 @@ public void updateExternalViews(List<ExternalView> externalViews) {
...
   }
 
+  /**
+   * Update the cached customized view map
+   * @param customizedViews
+   */
+  public void updateCustomizedViews(String customizedStateType,
+      List<CustomizedView> customizedViews) {
+    for (CustomizedView cv : customizedViews) {
+      if (_customizedViewCacheMap.containsKey(customizedStateType)) {
```
jiajunwang, line 268: You can move this condition out of the loop by checking and adding key if not exists.
Then in the loop, you only need to set property.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -281,12 +300,24 @@ public void removeExternalViews(List<String> resourceNames) {
...
   }
 
+  /**
+   * Remove dead customized views for a certain state type from customized view cache
+   * @param stateType a specific customized state type
+   * @param resourceNames the names of resources whose customized view is stale
+   */
+  public void removeCustomizedViews(String stateType, List<String> resourceNames) {
+    for (String resourceName : resourceNames) {
+      _customizedViewCacheMap.get(stateType).getCustomizedViewCache()
```
jiajunwang, line 316: Safer to check if the stateType exists or not.

zhangmeng916:
final commit message:
"Add customized view cache update in customized view aggregation stage. This is to ensure that the cache does not have stale data"

Issue #940, opened 2020-04-09
Full auto rebalancer with CrushEd algorithm changed partition assignment unexpected.


jiajunwang:
It was reported and verified that the current full auto rebalancer with CrushEd strategy may change partition assignment if the rebalance is triggered rapidly (2 or 3 times per second). This caused unexpected partition shuffling.

Till now, based on our investigation, the problem might be related to the unstable map key iterator in Java 8 or after. Since this usually happens with a short while, it might be overlooked since we bump up JDK version.

We plan to fix it while ensuring the algorithm's backward compatibility.

PR #941, opened 2020-04-09
Fix unexpceted partition movements in the CrushEd strategy.

jiajunwang:
### Issues

#940 

### Description

This is a workaround fix to ensure backward compatibility. An additional cache map is used to keep the stable partition list so as to remove the randomness in the algorithm input.
Note that the right fix would be cleaner that we sort the list inside the strategy class. However, that will also change all existing cluster assignments in production.

### Tests

TestResourceControllerDataProvider.testStablePartitionListCache
* Note that integration test is not quite doable unless we have a way to let JDK keySet() method return the same set with different iteration orders. Please comment if any good idea.

[ERROR] Failures:
[ERROR]   TestRebalancePipeline.testMsgTriggeredRebalance:208 expected:<true> but was:<false>
[ERROR] org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceTopoplogyAwareDisabled.testLackEnoughLiveInstances(org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceTopoplogyAwareDisabled)
[ERROR]   Run 1: TestCrushAutoRebalanceTopoplogyAwareDisabled.testLackEnoughLiveInstances:79->TestCrushAutoRebalanceNonRack.testLackEnoughLiveInstances:228->TestCrushAutoRebalanceNonRack.validateIsolation:301 expected:<2> but was:<1>
[INFO]   Run 2: PASS
[INFO]
[INFO]
[ERROR] Tests run: 1119, Failures: 2, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time: 01:14 h
[INFO] Finished at: 2020-04-10T16:17:10-07:00

Rerun failed test

[INFO] Tests run: 13, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 37.316 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 13, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 43.478 s
[INFO] Finished at: 2020-04-10T16:32:30-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
@@ -314,14 +314,14 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
   }
 
   /**
-   * Return the cached stable partition list of the specified resource. If no such cached item,
-   * return empty list.
    * This is for a backward compatible workaround to fix https://github.com/apache/helix/issues/940.
    *
    * @param resourceName
+   * @return the cached stable partition list of the specified resource. If no such cached item,
+   * return null.
    */
   public List<String> getStablePartitionList(String resourceName) {
-    return _stablePartitionListCache.getOrDefault(resourceName, Collections.EMPTY_LIST);
+    return _stablePartitionListCache.get(resourceName);
   }
 
   /**
@@ -335,8 +335,8 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
     for (String resourceName : idealStateMap.keySet()) {
       Set<String> newPartitionSet = idealStateMap.get(resourceName).getPartitionSet();
       List<String> cachedPartitionList = getStablePartitionList(resourceName);
-      if (cachedPartitionList.size() != newPartitionSet.size() || !newPartitionSet
-          .containsAll(cachedPartitionList)) {
+      if (cachedPartitionList == null || cachedPartitionList.size() != newPartitionSet.size()
+          || !newPartitionSet.containsAll(cachedPartitionList)) {
         _stablePartitionListCache.put(resourceName, new ArrayList<>(newPartitionSet));
       }
     }
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
@@ -19,17 +19,17 @@ package org.apache.helix.controller.dataproviders;
  * under the License.
  */
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
 import java.util.concurrent.ConcurrentHashMap;
+
 import org.apache.helix.HelixConstants;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.PropertyKey;
-import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.common.caches.AbstractDataCache;
 import org.apache.helix.common.caches.PropertyCache;
 import org.apache.helix.controller.LogUtil;
@@ -37,6 +37,7 @@ import org.apache.helix.controller.pipeline.Pipeline;
 import org.apache.helix.controller.stages.MissingTopStateRecord;
 import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.ResourceAssignment;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -67,6 +68,15 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
 
   // Maintain a set of all ChangeTypes for change detection
   private Set<HelixConstants.ChangeType> _refreshedChangeTypes;
+
+  // CrushEd strategy needs to have a stable partition list input. So this cached list persist the
+  // previous seen partition lists. If the members in a list are not modified, the old list will be
+  // used in the algorithm to ensure it calculates the same result.
+  // Refer to https://github.com/apache/helix/issues/940.
+  // TODO: Sorting the partition list in CrushEd strategy instead of using the cache to reduce
+  // TODO: dependency. Note that this will change the cluster partition assignment and potentially
+  // TODO: cause shuffling. So it is non backward compatible.
+  private final Map<String, List<String>> _stablePartitionListCache = new HashMap<>();
 
   public ResourceControllerDataProvider() {
     this(AbstractDataCache.UNKNOWN_CLUSTER);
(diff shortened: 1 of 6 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/AutoRebalancer.java
```diff
@@ -64,7 +64,12 @@ public IdealState computeNewIdealState(String resourceName,
 
     LOG.info("Computing IdealState for " + resourceName);
 
-    List<String> partitions = new ArrayList<>(currentIdealState.getPartitionSet());
+    // This is part of the backward compatible workaround to fix
+    // https://github.com/apache/helix/issues/940.
+    // TODO: remove the workaround once we are able to apply the simple fix without majorly
```
lei-xia, line 69: When do you think we can fix it permanently?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -294,4 +304,32 @@ public void clearMonitoringRecords() {
...
+
+  /**
+   * If the list items are the same, return the stable partition list cache. Otherwise, refresh
+   * the cache with new list and return the new list.
+   * This is for a backward compatible workaround to fix https://github.com/apache/helix/issues/940.
+   *
+   * @param resourceName
+   * @param newPartitionSet
+   */
+  public List<String> getOrSetStablePartitionList(String resourceName,
```
lei-xia, line 316: This method is confusing. Another option is to call this method when the cluster data cache refresh IdealStates, this can keep all logic within data cache. 
i.e,
split to two methods: 1) refreshPartitionList(), this is private.  and 2) getPartitionList(), this is public and used by the rebalancer (and probably in other stages too where a partition list is needed, so there will be consistent over all stages).

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AutoRebalancer.java
jiajunwang: There are 2 plans in my mind:
1. Encourage customers to use WAGED. Needs to wait until WAGED is stable.
2. Adding a CRUSHedV2 with the sorting logic, and leave V1 as it is now. Then deprecate V1. For this one, I would prefer to do it after we find all issues in the CRUSHed otherwise we might have V3 later.

After compare between these 2 options, I personally prefer the first option. That's why I actually removed the CRUSHedV2 after I added it to my PR. What's your opinion?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
jiajunwang: Changed with some additional modification to ensure safety.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -294,4 +312,33 @@ public void clearMonitoringRecords() {
...
+   *
+   * @param idealStateMap
+   */
+  final void refreshStablePartitionList(Map<String, IdealState> idealStateMap) {
+    _stablePartitionListCache.keySet().retainAll(idealStateMap.keySet());
+    for (String resourceName : idealStateMap.keySet()) {
+      Set<String> newPartitionSet = idealStateMap.get(resourceName).getPartitionSet();
+      List<String> cachedPartitionList = getStablePartitionList(resourceName);
+      if (cachedPartitionList.size() != newPartitionSet.size() || !newPartitionSet
+          .containsAll(cachedPartitionList)) {
```
lei-xia, line 339: Keep in mind "newPartitionSet.containsAll(cachedPartitionList))" could cause a performance (delay) issue, the operation is O(N^2), given if we have large number of partitions.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
```diff
@@ -417,4 +417,25 @@ public int compare(String ins1, String ins2) {
...
+  // https://github.com/apache/helix/issues/940.
+  // TODO: remove the workaround once we are able to apply the simple fix without majorly
+  // TODO: impacting user's clusters.
+  protected List<String> getStablePartitionList(ResourceControllerDataProvider clusterData,
+      IdealState currentIdealState) {
+    List<String> partitions =
+        clusterData.getStablePartitionList(currentIdealState.getResourceName());
+    Set<String> currentPartitionSet = currentIdealState.getPartitionSet();
+    if (partitions.size() != currentPartitionSet.size() || !currentPartitionSet
+        .containsAll(partitions)) {
```
lei-xia, line 431: Let us think about if we  have any other way to do this.  We actually found containsAll() has been a big performance issue before, specially if you compare a set with a list (which take O(N) to iterator).

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
```diff
@@ -417,4 +417,25 @@ public int compare(String ins1, String ins2) {
       return p1.compareTo(p2);
     }
   }
+
+  // This is for a backward compatible workaround to fix
+  // https://github.com/apache/helix/issues/940.
+  // TODO: remove the workaround once we are able to apply the simple fix without majorly
+  // TODO: impacting user's clusters.
+  protected List<String> getStablePartitionList(ResourceControllerDataProvider clusterData,
```
lei-xia, line 425: Why do  we compare here again? You are saying the IdealState used here could be different from  cluster data cache?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
jiajunwang: Why it is O(N^2)? The containsAll of the set is looping on the input list and then call contains() to check. So it should be O(N). Note that we are copying the input set to a list anyway. So this won't change the latency dramatically. The O() calculation has the same result.

But I agree that this might still cause an additional delay. Since this is related to the later 2 comments, let's keep in the same threads. I will resolve this one.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
jiajunwang: I thought about this, but the other ways are not safe or creating too many additional dependencies. For example, if IS has not been changed, then we don't update the partition lists.
But there is a big assumption there. Doing that might cause some bugs in the future.

I don't have a better idea now. Please suggest.

jiajunwang: Run a quick profile about containsAll() of Set.

```
  public static void main(String[] args) {
    Set<String> set;
    List<String> input = new ArrayList<>();
    Random ran = new Random();
    for (int i = 0; i < 10000000; i++)  {
      input.add("test" + ran.nextInt() + i);
    }
    set = new HashSet<>(input);
    long mid = System.currentTimeMillis();
    set.containsAll(input);
    long end = System.currentTimeMillis();
    System.out.println("Phase I: " + (mid - start) + " Phase II: " + (end - mid));
  }
```
The result is Phase I: 5560 Phase II: 1242

So containsAll is not a major concern here. I think.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
jiajunwang: If the method is put in a different class, it would be safer to protect from the OO design perspective. In detail, after we put refreshStablePartitionList as private into ResourceControllerDataProvider, we don't have a mechanism to guarantee refreshStablePartitionList() is called and the refresh is done completely.

It is optional though, especially it is a workaround. If it is a big concern to you, I can remove it.

jiajunwang:
This PR is ready to be merged, approved by @lei-xia

PR #946, opened 2020-04-10
Move routing table provider initialization

mgao0:
### Issues

In current test we initialize routing table provider after updating customized state, and there are several re-initialization of routing table provider at different places during the test to set different data sources. Although routing table provider will fetch correct results in this way, it does not represent a common way how our customers should use routing table provider. This PR fixes this problem.

### Description

This PR moves the initialization of routing table provider to before class so it is initialized before any updates. In addition, in this PR re-initializations of routing table provider during the test are removed. Also, added several checks into the validation method to cover some edge cases. 

### Tests

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/integration/TestCustomizedViewAggregation.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/TestCustomizedViewAggregation.java
@@ -319,6 +319,22 @@ public class TestCustomizedViewAggregation extends ZkUnitTestBase {
   }
 
   /**
+   * Set the data sources (customized state types) for routing table provider
+   * @param customizedStateTypes list of customized state types that routing table provider will include in the snapshot shown to users
+   */
+  private void setRoutingTableProviderDataSources(List<CustomizedStateType> customizedStateTypes) {
+    List<String> customizedViewSources = new ArrayList<>();
+    _routingTableProviderDataSources.clear();
+    for (CustomizedStateType type : customizedStateTypes) {
+      customizedViewSources.add(type.name());
+      _routingTableProviderDataSources.add(type.name());
+    }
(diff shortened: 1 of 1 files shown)


helix-core/src/test/java/org/apache/helix/integration/TestCustomizedViewAggregation.java
```diff
@@ -120,6 +118,23 @@ public void beforeClass() throws Exception {
...
+    customizedStateConfigBuilder.setAggregationEnabledTypes(customizedStateTypes);
+    HelixDataAccessor accessor = _manager.getHelixDataAccessor();
+    accessor.setProperty(accessor.keyBuilder().customizedStateConfig(),
+        customizedStateConfigBuilder.build());
+    _aggregationEnabledTypes.addAll(customizedStateTypes);
+
+    Map<PropertyType, List<String>> dataSource = new HashMap<>();
+    dataSource.put(PropertyType.CUSTOMIZEDVIEW, customizedStateTypes);
+    _routingTableProvider = new RoutingTableProvider(_spectator, dataSource);
+    _routingTableProviderDataSources.addAll(customizedStateTypes);
```
zhangmeng916, line 137: can we try somewhere in the test process to initialize a new routing table provider with different types? A little concerned that we have only one fixed data resources input here. Let's put the initialization before the update of config/states, and see whether that's stable.

mgao0: Good point. I added back two tests on this.

mgao0:
This PR is ready to be merged, approved by @zhangmeng916

Final commit message:
This PR moves the initialization of routing table provider to before class so it is initialized before any updates. Also, added several checks into the validation method to cover some edge cases.

Issue #939, opened 2020-04-09
Fix flaky tests


narendly:
[ERROR]   TestRebalancePipeline.testMsgTriggeredRebalance:195 expected:<1> but was:<0>
[ERROR]   TestEnableCompression.testEnableCompressionResource:117 expected:<true> but was:<false>

jiajunwang:
Let me take a look at the testMsgTriggeredRebalance failure.

jiajunwang:
Adding 2 unstable tests that happen in the other test runs.

TestControllerLeadershipChange.testMissingTopStateDurationMonitoring:262 expected:<true> but was:<false>
org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceNonRack.testLackEnoughInstances(org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceNonRack)
  Run 1: PASS
  Run 2: TestCrushAutoRebalanceNonRack.testLackEnoughInstances:250 » ZkClient Failed to...

pkuwm:
```
Results :

Failed tests:
  TestBasicSpectator.TestSpectator:58 expected:<true> but was:<false>
org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceTopoplogyAwareDisabled.testLackEnoughInstances(org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceTopoplogyAwareDisabled)
  Run 1: PASS
  Run 2: TestCrushAutoRebalanceTopoplogyAwareDisabled.testLackEnoughInstances:86->TestCrushAutoRebalanceNonRack.testLackEnoughInstances:250 » ZkClient

  TestWorkflowTermination.testWorkflowJobFail:251->verifyWorkflowCleanup:257 expected:<true> but was:<false>
  TestWorkflowTermination.testWorkflowRunningTimeout:131->verifyWorkflowCleanup:257 expected:<true> but was:<false>

Tests run: 1111, Failures: 4, Errors: 0, Skipped: 1
```

pkuwm:
```
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR] org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceNonRack.test(org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceNonRack)
[INFO]   Run 1: PASS
[...]

PR #955, opened 2020-04-15
Fix MSD routing data refreshing

pkuwm:
### Issues

Fixes part of #939 

### Description

1. Routing data update methods in ZkMetadataStoreDirectory writes routing data to ZK and refresh routing cache. However, it should not refresh routing data if writing to ZK fails. Refreshing routing should only be proceeded if writing to ZK is successful.

2. The test is flaky because of removing namespace when refreshing routing data. It is caused by race condition between the read request after updating and data change callback to refresh routing (which is second refresh). Namespace should only be removed when exception is thrown.

### Tests

Fix TestZkMetadataStoreDirectory

Before: TestZkMetadataStoreDirectory failed within 5 attempts.
After: 50 attempts of running TestZkMetadataStoreDirectory passed

In helix-rest:

```
[INFO] Tests run: 155, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 40.216 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 155, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  47.437 s
[INFO] Finished at: 2020-04-22T17:39:58-07:00
```

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
@@ -175,9 +175,11 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
           "Failed to set routing data: Namespace " + namespace + " is not found!");
     }
     synchronized (this) {
-      boolean result = _routingDataWriterMap.get(namespace).setRoutingData(routingData);
+      if (!_routingDataWriterMap.get(namespace).setRoutingData(routingData)) {
+        return false;
+      }
       refreshRoutingData(namespace);
-      return result;
+      return true;
     }
   }
 
@@ -232,9 +234,11 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
           "Failed to add metadata store realm: Namespace " + namespace + " is not found!");
     }
     synchronized (this) {
-      boolean result = _routingDataWriterMap.get(namespace).addMetadataStoreRealm(realm);
+      if (!_routingDataWriterMap.get(namespace).addMetadataStoreRealm(realm)) {
+        return false;
+      }
       refreshRoutingData(namespace);
-      return result;
+      return true;
     }
   }
 
@@ -247,9 +251,11 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
           "Failed to delete metadata store realm: Namespace " + namespace + " is not found!");
     }
     synchronized (this) {
-      boolean result = _routingDataWriterMap.get(namespace).deleteMetadataStoreRealm(realm);
+      if (!_routingDataWriterMap.get(namespace).deleteMetadataStoreRealm(realm)) {
+        return false;
+      }
       refreshRoutingData(namespace);
-      return result;
+      return true;
     }
   }
 
@@ -272,9 +278,11 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
             "Failed to add sharding key: Adding sharding key " + shardingKey
                 + " makes routing data invalid!");
       }
-      boolean result = _routingDataWriterMap.get(namespace).addShardingKey(realm, shardingKey);
+      if (!_routingDataWriterMap.get(namespace).addShardingKey(realm, shardingKey)) {
+        return false;
+      }
       refreshRoutingData(namespace);
-      return result;
+      return true;
     }
   }
 
@@ -287,9 +295,11 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
           "Failed to delete sharding key: Namespace " + namespace + " is not found!");
     }
     synchronized (this) {
-      boolean result = _routingDataWriterMap.get(namespace).deleteShardingKey(realm, shardingKey);
+      if (!_routingDataWriterMap.get(namespace).deleteShardingKey(realm, shardingKey)) {
+        return false;
+      }
       refreshRoutingData(namespace);
-      return result;
+      return true;
     }
   }
 
@@ -304,16 +314,6 @@ public class ZkMetadataStoreDirectory implements MetadataStoreDirectory, Routing
    */
   @Override
   public void refreshRoutingData(String namespace) {
-    // Safe to ignore the callback if any of the maps are null.
-    // If routingDataMap is null, then it will be populated by the constructor anyway
-    // If routingDataMap is not null, then it's safe for the callback function to update it
-    if (_routingZkAddressMap == null || _realmToShardingKeysMap == null
-        || _routingDataReaderMap == null || _routingDataWriterMap == null) {
-      LOG.warn(
-          "refreshRoutingData callback called before ZKMetadataStoreDirectory was fully initialized. Skipping refresh!");
-      return;
-    }
-
     // Check if namespace exists; otherwise, return as a NOP and log it
     if (!_routingZkAddressMap.containsKey(namespace)) {
       LOG.error(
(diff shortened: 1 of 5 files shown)


narendly:
@NealSun96 Could you please review this change?

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -304,42 +314,35 @@ public boolean deleteShardingKey(String namespace, String realm, String sharding
    */
   @Override
   public void refreshRoutingData(String namespace) {
-    // Safe to ignore the callback if any of the maps are null.
-    // If routingDataMap is null, then it will be populated by the constructor anyway
-    // If routingDataMap is not null, then it's safe for the callback function to update it
-    if (_routingZkAddressMap == null || _realmToShardingKeysMap == null
-        || _routingDataReaderMap == null || _routingDataWriterMap == null) {
-      LOG.warn(
-          "refreshRoutingData callback called before ZKMetadataStoreDirectory was fully initialized. Skipping refresh!");
```
NealSun96, line 313: This if statement shouldn't be deleted. The reasoning is clearly shown in the logging: the callback could be triggered before `ZKMetadataStoreDirectory` is fully initialized. Removing this if statement leaves the rest of the logic unguarded.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -304,42 +314,35 @@ public boolean deleteShardingKey(String namespace, String realm, String sharding
...
       return;
     }
 
+    TrieRoutingData trieRoutingData;
     try {
-      _routingDataMap.put(namespace, new TrieRoutingData(rawRoutingData));
+      trieRoutingData = new TrieRoutingData(rawRoutingData);
     } catch (InvalidRoutingDataException e) {
       LOG.warn("TrieRoutingData is not created for namespace {}", namespace, e);
+      _realmToShardingKeysMap.remove(namespace);
```
NealSun96, line 339: The raw data shouldn't be removed here. The raw data is allowed to contain invalid routing data.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -304,42 +314,35 @@ public boolean deleteShardingKey(String namespace, String realm, String sharding
...
-    // Remove routing data first in case of failure on creation
-    _routingDataMap.remove(namespace);
-
     Map<String, List<String>> rawRoutingData;
     try {
       rawRoutingData = _routingDataReaderMap.get(namespace).getRoutingData();
-      _realmToShardingKeysMap.put(namespace, rawRoutingData);
     } catch (InvalidRoutingDataException e) {
       LOG.error("Failed to refresh cached routing data for namespace {}", namespace, e);
+      _realmToShardingKeysMap.remove(namespace);
```
NealSun96, line 329: As you pointed out yesterday, it's better to assign empty data here.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
```diff
@@ -304,42 +314,35 @@ public boolean deleteShardingKey(String namespace, String realm, String sharding
...
+      trieRoutingData = new TrieRoutingData(rawRoutingData);
     } catch (InvalidRoutingDataException e) {
       LOG.warn("TrieRoutingData is not created for namespace {}", namespace, e);
+      _realmToShardingKeysMap.remove(namespace);
+      _routingDataMap.remove(namespace);
+      return;
     }
+
+    _realmToShardingKeysMap.put(namespace, rawRoutingData);
+    _routingDataMap.put(namespace, trieRoutingData);
```
NealSun96, line 345: Overall, I think it's more clean to do `try {overwriteOldData} catch {eraseOldData}` for both blocks of the logic.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/TestZkMetadataStoreDirectory.java
```diff
@@ -344,6 +345,10 @@ public void testDataDeletionCallback() throws Exception {
               + " contains either empty or invalid routing data!")) {
             return false;
           }
+        } catch (IllegalArgumentException iae) {
+          if (!iae.getMessage().equals("Provided path is not a valid Zookeeper path: anyKey")) {
+            return false;
```
NealSun96, line 350: The meaning of the block on line 343 is "ensure that an `IllegalStateException` is raised and the message is 'Failed to get...'". Is this block ensuring a `IllegalArgumentException` is raised? That's not what this test case is testing.

pkuwm: Refreshing routing has a delay. And this is why `TestHelper.verify()` is used. Before routing is deleted in cached, namespace is still in cache, right? So the request still goes to TrieRouting, and IllegalArgumentException would be thrown because `anyKey` is invalid. In this case, we should catch the exception and retry, right?

The better I think is this. I will update PR.
```
      for (String namespace : _routingZkAddrMap.keySet()) {
        try {
          _metadataStoreDirectory.getMetadataStoreRealm(namespace, "anyKey");
          Assert.fail("Should not successfully get routing data");
        } catch (IllegalStateException e) {
          // If other IllegalStateException, it is unexpected and this test should fail.
          if (!e.getMessage().equals("Failed to get metadata store realm: Namespace " + namespace
              + " contains either empty or invalid routing data!")) {
            throw e;
          }
        } catch (IllegalArgumentException iae) {
          // If routing data is not yet refreshed, return false and retry.
          if (iae.getMessage().equals("Provided path is not a valid Zookeeper path: anyKey")) {
            return false;
          }
          // If other IllegalArgumentException, it is not expected and this test should fail.
          throw iae;
        }
      }
      return true;
    }, TestHelper.WAIT_DURATION));
```

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: When the `ZkMetadataStoreDirectory` is initialized, it initializes `ZkRoutingDataReader` which subscribe to ZNode changes and calls `refreshRoutingData()` when an event happens. It's possible that an event comes in before `ZkMetadataStoreDirectory` finishes initialization, which we don't want to handle. 
This is my understanding to it. This was introduced to me first by @narendly , and he was also the first to write this statement so he could provide further comments if necessary.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: My understanding is that you prefer not deleting namespaces, and I agree with your point, so it's safe to change that behavior. We can modify the test cases to reflect that.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
NealSun96: The validation process = the exception. If something is caught it means the data is invalid. So validation will always go first before update.

helix-rest/src/test/java/org/apache/helix/rest/metadatastore/TestZkMetadataStoreDirectory.java
NealSun96: I see your point now, thanks for the explanation. Was this test case flaky because of that? This is a very good catch! 👍  
I think it makes more sense to replace "anyKey" with an actual existing key in the routing data; the handling of `IllegalArgumentException` is only caused by "anyKey", and is not what this test case should focus on. My point is that using "anyKey" is a mistake. We shouldn't change the test case to accommodate that mistake; we can just correct it by using something like `TEST_SHARDING_KEYS`.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
pkuwm: @NealSun96 Thanks for the explanation.
After I check the code carefully, I don't see what you said is the case. `ZkRoutingDataReader` is initialized after the maps are created. So I think this check is safe to remove.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
pkuwm: Discussed offline. I've changed to assign an empty map to `_realmToShardingKeysMap` namespace. And the tests pass.

helix-rest/src/main/java/org/apache/helix/rest/metadatastore/ZkMetadataStoreDirectory.java
pkuwm: I prefer to catch the exception specifically for this operation `trieRoutingData = new TrieRoutingData(rawRoutingData);` to make it clean.

pkuwm:
This PR is ready to be merged, approved by @NealSun96 

Commit:
```
testSetNamespaceRoutingData is flaky because namespace is removed when refreshing routing data. It is caused by race condition between the read request after updating and data change callback to refresh routing. And routing data cache should not be refreshed if writing routing to ZK fails.
```

PR #968, opened 2020-04-23
Improve logging for isClusterSetup

### LINKED ISSUE #967: If isClusterSetup fails, we don't get the errorMsg
If isClusterSetup fails, we don't get the errorMsg

narendly:
### Issues

Fixes #967 

### Description

This improves logging.
### Tests

N/A

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
@@ -112,13 +112,13 @@ public final class ZKUtil {
     for (int i = 0; i < ret.length; i++) {
       if (!ret[i]) {
         isValid = false;
-        errorMsg.append("Invalid cluster setup, missing znode path: ").append(requiredPaths.get(i))
-            .append("\n");
+        errorMsg.append("Invalid cluster setup for cluster: ").append(clusterName)
+            .append(", missing znode path: ").append(requiredPaths.get(i)).append("\n");
       }
     }
 
     if (!isValid) {
-      logger.debug(errorMsg.toString());
+      logger.warn(errorMsg.toString());
     }
 
     return isValid;


helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
```diff
@@ -112,13 +112,13 @@ public static boolean isClusterSetup(String clusterName, RealmAwareZkClient zkCl
...
-        errorMsg.append("Invalid cluster setup, missing znode path: ").append(requiredPaths.get(i))
-            .append("\n");
+        errorMsg.append("Invalid cluster setup for cluster: ").append(clusterName)
+            .append(", missing znode path: ").append(requiredPaths.get(i)).append("\n");
       }
     }
 
     if (!isValid) {
-      logger.debug(errorMsg.toString());
+      logger.warn(errorMsg.toString());
```
dasahcc, line 121: Will this generate a lot of logs? That is the reason we move this to debug log instead of warning or even info.

narendly: 1. Generally the pipeline shouldn't call this every time.
2. If this condition is met, that means the cluster is not in the right state anyways, so it's good that we make this a warn log to be able to diagnose the problem appropriately. Clusters that aren't ill-formed won't get this log.

dasahcc: It is better to ask original author to see whether it is a concern or not. @lei-xia

dasahcc:
I checked all the callers. Anyway they will make an exception along with that. So it should be fine to enable warn log back. 

Let's whether there is any other concerns from @lei-xia . If not, I am gonna approve it.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKUtil.java
dasahcc: @narendly One thing I need to correct. It does not have to be cluster in ill-form. If there is some network or ZK server issue, partial of the path has been read. You will get this error frequently. But it does not means the cluster is not well setup.

For your change, I am fine. But I think we should think about to reduce the places of using isClusterSetup.

narendly: @dasahcc 
We still need to record this in order to diagnose whether it was a connectivity issue that caused this. Otherwise, how do you prove that it was a connectivity issue?

narendly: Additionally, what you said doesn't really strengthen your case about logging because connectivity issues are supposed to be transient. Then how does that make logging worse? If anything, this would serve as an important marker that lets you know that there were connectivity issues.

dasahcc: There was an issue for the HelixAdmin get resource. The cluster just added 1 min ago. And we verify the cluster does create successfully. Using another HelixAdmin to create a resource make it failed. It is network issue.

It will not be related to the log warning or debug, but we need to think about the whether we should apply this isClusterSetup check anywhere. And as I said, I am OK with the log itself.

narendly: What you're talking about is whether it's appropriate to call isClusterSetup. That's not the focus of this PR - please create a separate issue to track that work if you think is necessary.

narendly:

Issue #969, opened 2020-04-23
Batch Aync ZK calls may return partial results


jiajunwang:
The batch read or write ZK operations are built based on the Async ZK calls. The calls might fail due to the transient ZK connectivity issue. In this case, the batch read will return a partial result or throw exception. Either way is less idea.

For resolve this issue, we want to do the following 2 things,
1. Retry when the operation fails because of the ZK connectivity issue.
2. Fail the batch operation in the critical pipeline logic to ensure the Helix components finish the calculation based on a complete view.

PR #970, opened 2020-04-23
Add async call retry to resolve the transient ZK connection issue.

jiajunwang:
### Issues

#969 

### Description

If any exceptions happen during the async call, the current design will fail the operation and may eventually return a partial result.
This change makes the ZkClient retry operation if the error is because of a temporary ZK connection issue (CONNECTIONLOSS, SESSIONEXPIRED, SESSIONMOVED).
So the async call has a larger chance to finish the operation. Note that if the exception is due to business logic, the async call will still fail and the right return code will be sent to the callback handler.

### Tests

TestZkClientAsyncRetry

Keep running 75 times to ensure it's stable.
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 6.558 s - in org.apache.helix.zookeeper.impl.client.TestZkClientAsyncRetry
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 10.896 s
[INFO] Finished at: 2020-04-25T15:56:37-07:00
//======================================================================
Attempt 75 TestZkClientAsyncRetry
//======================================================================

**zookeeper-api**

[INFO] Tests run: 24, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 7.118 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 24, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 13.684 s
[INFO] Finished at: 2020-04-25T16:20:32-07:00

**helix-core**

[INFO] Tests run: 1144, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,621.077 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 01:17 h
[INFO] Finished at: 2020-04-27T12:34:24-07:00

**helix-rest**

[INFO] Tests run: 159, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 193.83 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  03:19 min
[INFO] Finished at: 2020-04-25T16:51:55-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -27,7 +27,10 @@ import javax.management.JMException;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.exception.ZkClientException;
 import org.apache.helix.zookeeper.zkclient.annotation.PreFetch;
+import org.apache.helix.zookeeper.zkclient.callback.ZkAsyncCallMonitorContext;
 import org.apache.helix.zookeeper.zkclient.callback.ZkAsyncCallbacks;
+import org.apache.helix.zookeeper.zkclient.callback.ZkAsyncRetryCallContext;
+import org.apache.helix.zookeeper.zkclient.callback.ZkAsyncRetryThread;
 import org.apache.helix.zookeeper.zkclient.exception.ZkBadVersionException;
 import org.apache.helix.zookeeper.zkclient.exception.ZkException;
 import org.apache.helix.zookeeper.zkclient.exception.ZkInterruptedException;
@@ -90,6 +93,7 @@ public class ZkClient implements Watcher {
 
   private boolean _shutdownTriggered;
   private ZkEventThread _eventThread;
+  private final ZkAsyncRetryThread _asyncCallRetryThread;
   // TODO PVo remove this later
   private Thread _zookeeperEventThread;
   private volatile boolean _closed;
@@ -182,6 +186,9 @@ public class ZkClient implements Watcher {
     _pathBasedZkSerializer = zkSerializer;
     _operationRetryTimeoutInMillis = operationRetryTimeout;
     _isNewSessionEventFired = false;
+
+    _asyncCallRetryThread = new ZkAsyncRetryThread(zkConnection.getServers());
+    _asyncCallRetryThread.start();
 
     connect(connectionTimeout, this);
 
@@ -1736,15 +1743,23 @@ public class ZkClient implements Watcher {
       data = (datat == null ? null : serialize(datat, path));
     } catch (ZkMarshallingError e) {
       cb.processResult(KeeperException.Code.MARSHALLINGERROR.intValue(), path,
-          new ZkAsyncCallbacks.ZkAsyncCallContext(_monitor, startT, 0, false), null);
+          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
       return;
     }
+    doAsyncCreate(path, data, mode, startT, cb);
+  }
+
+  private void doAsyncCreate(final String path, final byte[] data, final CreateMode mode,
+      final long startT, final ZkAsyncCallbacks.CreateCallbackHandler cb) {
     retryUntilConnected(() -> {
       ((ZkConnection) getConnection()).getZookeeper()
-          .create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE,
-              // Arrays.asList(DEFAULT_ACL),
-              mode, cb, new ZkAsyncCallbacks.ZkAsyncCallContext(_monitor, startT,
-                  data == null ? 0 : data.length, false));
+          .create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, mode, cb,
+              new ZkAsyncRetryCallContext(_asyncCallRetryThread, _monitor, startT, 0, false) {
+                @Override
+                protected void doRetry() {
+                  doAsyncCreate(path, data, mode, System.currentTimeMillis(), cb);
+                }
+              });
       return null;
     });
   }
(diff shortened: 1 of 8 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1758,50 +1776,66 @@ public void asyncSetData(final String path, Object datat, final int version,
...
+      final ZkAsyncCallbacks.SetDataCallbackHandler cb) {
     retryUntilConnected(() -> {
       ((ZkConnection) getConnection()).getZookeeper().setData(path, data, version, cb,
-          new ZkAsyncCallbacks.ZkAsyncCallContext(_monitor, startT,
-              data == null ? 0 : data.length, false));
+          new ZkAsyncRetryCallContext(_asyncCallRetryThread, cb, _monitor, startT,
+              data == null ? 0 : data.length, false) {
+            @Override
+            protected void doRetry() {
+              doAsyncSetData(path, data, version, System.currentTimeMillis(), cb);
```
dasahcc, line 1793: Is this recursively self calling OK?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -122,44 +121,66 @@ public void handle() {
...
+              return;
+            } else {
+              LOG.warn(
+                  "Cannot request to retry the operation. The retry request thread may have been stopped.");
+            }
+          } catch (Throwable t) {
+            LOG.error("Failed to request to retry the operation.", t);
+          }
+        } else {
+          LOG.warn(
```
dasahcc, line 158: Would this be to many introduced for log?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: This is the design. If no connectivity issue, it won't be triggered. The assumption here is that the connectivity issue is transient and won't happen continuously.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
jiajunwang: I don't think so.
The first two logs only happen when retry is not possible because of some unknown error. We need to know in this case.
The third one happens if Helix devs changed the Helix code in a strange way. It never happens with the existing PR code.
Overall, if everything works fine, we won't see any error messages.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
dasahcc: I am not quite familiar with the way you use. I just wonder would that cause infinite call then stack overflow?

jiajunwang: If I say it won't, I guess you won't believe it so easily. Please take a look at the code carefully. In general, it is not a recursive call, it is a callback triggered in a different thread after this method is done. Even we keep retrying, only one call exists in the stack.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
alirezazamani: about "Helix devs changed the Helix code in a strange way": 
It would be great if you can identify some of the places that code change can cause the issue and put some comments in that parts? so people will be careful about their changes that they are making later.

jiajunwang: @alirezazamani I understand your concern. But there are many many ways to make it happen. I won't be able to comment about every case. Moreover, strictly speaking, this final check is not like an NPE check, this is not a failure case. This is just meaning that the caller does not want a retry. And this is not something we have to avoid, if logic requires, we shall still do it. It is just not happening, and a little bit strang for the current logic. But we shall not discourage people to use it in a different way in the future. At that moment, if this log becomes too verbose, we can downgrade it or remove it.

jiajunwang:
This PR is ready to be merged, approved by @alirezazamani @dasahcc 
Also, re-run the helix-core and helix-rest tests, all passed.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -122,44 +121,66 @@ public void handle() {
...
   /**
-   * Default callback for zookeeper async api
+   * Default callback for zookeeper async api.
    */
-  public static abstract class DefaultCallback {
-    AtomicBoolean _lock = new AtomicBoolean(false);
-    int _rc = -1;
+  public static abstract class DefaultCallback implements CancellableZkAsyncCallback {
+    AtomicBoolean _isOperationDone = new AtomicBoolean(false);
+    int _rc = UNKNOWN_RET_CODE;
```
kaisun2000, line 128: why change this value from -1 to 255?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -122,44 +121,66 @@ public void handle() {
...
+              // The retry operation will be done asynchronously. Once it is done, the same callback
+              // handler object shall be triggered to ensure the result is notified to the right
+              // caller(s).
+              return;
+            } else {
+              LOG.warn(
+                  "Cannot request to retry the operation. The retry request thread may have been stopped.");
+            }
+          } catch (Throwable t) {
+            LOG.error("Failed to request to retry the operation.", t);
```
kaisun2000, line 155: Need retry, retri-able context, but retry operation failed? What to do here? Mark done, return some retriable RC value like CONNECTIONLOSS is not what the customer expect to handle right?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -183,6 +190,9 @@ protected ZkClient(IZkConnection zkConnection, int connectionTimeout, long opera
     _operationRetryTimeoutInMillis = operationRetryTimeout;
     _isNewSessionEventFired = false;
 
+    _asyncCallRetryThread = new ZkAsyncRetryThread(zkConnection.getServers());
```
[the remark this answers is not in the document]
jiajunwang: Good point, let me do it in a separate PR.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -172,22 +193,52 @@ public int getRc() {
...
+     * @return true if the error is transient and the operation may succeed when being retried.
+     */
+    private boolean needRetry(int rc) {
+      try {
+        switch (Code.get(rc)) {
+        /** Connection to the server has been lost */
+        case CONNECTIONLOSS:
+          /** The session has been expired by the server */
+        case SESSIONEXPIRED:
+          /** Session moved to another server, so operation is ignored */
```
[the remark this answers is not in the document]
jiajunwang: I think we are not using the async call to create ephemeral nodes. But this might be a concern. If that becomes the case, let's make the same change for async call.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
jiajunwang: -1 is a valid error code.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
jiajunwang: It is. Because retry is not possible. So we have to return with a result instead of letting the caller pending forever.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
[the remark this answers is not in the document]
jiajunwang: Sorry I was confused by myself. Name already given in this PR.
"ZkClient-AsyncCallback-Retry-" + getId() + "-" + name.

PR #973, opened 2020-04-27
TaskStateModelFactory configurable thread pool size

### LINKED ISSUE #972: Allow TaskStateModelFactory to use configurable thread pool size
We are improving the task framework by allowing task thread pools to have configurable sizes. As a part of the participant side change, `TaskStateModelFactory` is changed to allocate thread pool sizes based on values provided in `InstanceConfig` or `ClusterConfig`. 

Both `InstanceConfig` and `ClusterConfig` are now expected to contain a field keyed as "TARGET_TASK_THREAD_POOL_SIZE", where the value for `InstanceConfig` only determines the thread pool size for the specified instance, and the value for `ClusterConfig` determines the thread pool sizes for all instances of that cluster. If `InstanceConfig` isn't able to provide the pool size, then `ClusterConfig` will be used to determine the pool size; if neither of them are able to provide the pool size, `TaskStateModelFactory` will use a default value.

NealSun96:
### Issues

Fixes #972

### Description

We are improving the task framework by allowing task thread pools to have configurable sizes. As a part of the participant side change, `TaskStateModelFactory` is changed to allocate thread pool sizes based on values provided in `InstanceConfig` or `ClusterConfig`.

Both `InstanceConfig` and `ClusterConfig` are now expected to contain a field keyed as "TARGET_TASK_THREAD_POOL_SIZE", where the value for `InstanceConfig` only determines the thread pool size for the specified instance, and the value for `ClusterConfig` determines the thread pool sizes for all instances of that cluster. If `InstanceConfig` isn't able to provide the pool size, then `ClusterConfig` will be used to determine the pool size; if neither of them are able to provide the pool size, `TaskStateModelFactory` will use a default value.

### Tests

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestZkConnectionLost.testLostZkConnection:140 » Helix Workflow "testLostZkConn...
[ERROR]   TestStateTransitionTimeout.testStateTransitionTimeOut:166 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1144, Failures: 4, Errors: 0, Skipped: 1
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:15 h
[INFO] Finished at: 2020-05-12T12:22:14-07:00
```
Rerun failed classes
```
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 71.032 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:16 min
[INFO] Finished at: 2020-05-12T14:13:00-07:00
```

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -137,6 +140,7 @@
   private final static int MAX_REBALANCE_PREFERENCE = 10;
   private final static int MIN_REBALANCE_PREFERENCE = 0;
   public final static boolean DEFAULT_GLOBAL_REBALANCE_ASYNC_MODE_ENABLED = true;
+  public static final int TARGET_TASK_THREAD_POOL_SIZE_NOT_SET = -1;
```
narendly, line 143: Why does this need to be a public field? 

Also, it seems that this constant and some of the other enums are duplicated in different places. Is it possible to put this in one place? (Hint. TaskConstants.java)

NealSun96: Since `InstanceConfig` and `ClusterConfig` are both using the same enum and constant, how about only defining them in `InstanceConfig` and make them public, then `ClusterConfig` can reuse the values?

All the field names we have so far are enums in one of the several `ConfigProperty`s, and I think it's better to maintain that convention; a separate file like TaskConstants.java having a `InstanceConfig` field name might come as a surprise to others.

helix-core/src/main/java/org/apache/helix/model/InstanceConfig.java
```diff
@@ -506,6 +508,16 @@ public void setMaxConcurrentTask(int maxConcurrentTask) {
     _record.setIntField(InstanceConfigProperty.MAX_CONCURRENT_TASK.name(), maxConcurrentTask);
   }
 
+  /**
+   * Get the target size of task thread pool
+   * @return the target size of task thread pool
+   */
+  public int getTargetTaskThreadPoolSize() {
```
narendly, line 515: Could we add setters as well with some sort of validation logic?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -44,11 +49,11 @@
   private final ScheduledExecutorService _taskExecutor;
   private final ScheduledExecutorService _timerTaskExecutor;
   private ThreadPoolExecutorMonitor _monitor;
-  public final static int TASK_THREADPOOL_SIZE = 40;
+  public final static int DEFAULT_TASK_THREAD_POOL_SIZE = 40;
```
narendly, line 52: Could we potentially move this constant to TaskConstants? Do you think that would be a good idea?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -102,4 +107,40 @@ public boolean isShutdown() {
   public boolean isTerminated() {
     return _taskExecutor.isTerminated();
   }
+
+  /*
+   * Get target thread pool size from InstanceConfig first; if that fails, get it from
+   * ClusterConfig; if that fails, fall back to the default value.
+   */
+  private static int _getTaskThreadPoolSize(HelixManager manager) {
```
narendly, line 115: Why are we including an underscore for a method name?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -102,4 +107,40 @@ public boolean isShutdown() {
...
+    try {
+      InstanceConfig instanceConfig =
+          configAccessor.getInstanceConfig(manager.getClusterName(), manager.getInstanceName());
+      int targetTaskThreadPoolSize = instanceConfig.getTargetTaskThreadPoolSize();
+      if (_verifyTargetThreadPoolSize(targetTaskThreadPoolSize)) {
+        return targetTaskThreadPoolSize;
+      }
+    } catch (HelixException e) {
+      // Pass if InstanceConfig doesn't exist.
+    }
```
narendly, line 127: Try-catch is not necessary here. There are other ways to check if InstanceConfig exists or not.
E.g.) Perform one read, if it's null, then it will return null. If exists, then you'll have read the config.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -102,4 +107,40 @@ public boolean isShutdown() {
...
+    // Fallback to cluster config since instance config doesn't provide the value
+    try {
+      ClusterConfig clusterConfig = configAccessor.getClusterConfig(manager.getClusterName());
+      int targetTaskThreadPoolSize = clusterConfig.getTargetTaskThreadPoolSize();
+      if (_verifyTargetThreadPoolSize(targetTaskThreadPoolSize)) {
+        return targetTaskThreadPoolSize;
+      }
+    } catch (HelixException e) {
+      // Pass if ClusterConfig doesn't exist.
+    }
```
narendly, line 138: Again, this might be an incorrect use of a try-catch clause

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -102,4 +107,40 @@ public boolean isShutdown() {
...
+
+    return DEFAULT_TASK_THREAD_POOL_SIZE;
+  }
+
+  private static boolean _verifyTargetThreadPoolSize(int targetTaskThreadPoolSize) {
```
narendly, line 143: You could do this in the setter?

helix-core/src/main/java/org/apache/helix/model/InstanceConfig.java
NealSun96: I agree.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Since it's only used in this class, would it be better to have it here? And make it private?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Thank you for correcting!

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: This is to check against the default value: when the field is not specified, a -1 is returned from `getTargetTaskThreadPoolSize()`. This check makes sure the default value is filtered out. 

I don't see any other good way to do `getIntField()` while handling the "not defined" case differently, since `getIntField()` always asks for a default value.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: IMO, the cluster level default value is different from the node level pool size.
Better to use a different name for the cluster level default number.
DEFAULT_TARGET_TASK_THREAD_POOL_SIZE
DEFAULT_TARGET_TASK_THREAD_POOL_SIZE_NOT_SET

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -102,4 +107,36 @@ public boolean isShutdown() {
...
+  /*
+   * Get target thread pool size from InstanceConfig first; if that fails, get it from
+   * ClusterConfig; if that fails, fall back to the default value.
+   */
+  private static int getTaskThreadPoolSize(HelixManager manager) {
+    ConfigAccessor configAccessor = manager.getConfigAccessor();
+    // Check instance config first for thread pool size
+    InstanceConfig instanceConfig =
+        configAccessor.getInstanceConfig(manager.getClusterName(), manager.getInstanceName());
+    if (instanceConfig != null) {
```
jiajunwang, line 120: nit, shall we throw exception if the instance config or cluster config is not found?
This mostly because the read is not done successfully. In which case, we don't know the real configured numbers. So better stop initialization instead of guess it shall be 40.

NealSun96: For `InstanceConfig` it's actually valid to get null. When an instance is set up for the first time it's not going to have `InstanceConfig` when this code is executed, so an exception is not very appropriate because that situation is actually correct. 

I do think the `ClusterConfig` should have been there, but I still think an exception may be too strong.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: I realized that this is actually incorrect. I'll move it to TaskConstants.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
narendly: +1 to what @jiajunwang  said.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
narendly: I am okay with returning the default size because creating a thread pool with a default size won't be the end of the world and won't affect the user too much. And the user can always try to set it again or restart if they want to change the thread pool size. 

But we should make it clear to the user. I don't see any log or notification here, so that could be a problem.

NealSun96: So the context has changed, and what I said earlier is no longer valid. Please take a look if you want to further this discussion:

1. Thread pool creation is now after HelixManager connection, which means `InstanceConfig` and `ClusterConfig` are expected to exist. It should be treated as an error if they don't exist. 
2. I verified with test cases and believe that an earlier comment made by @narendly is incorrect: if `InstanceConfig` or `ClusterConfig` doesn't exist, `getInstanceConfig()`/`getClusterConfig()` will throw an HelixException because `isInstanceSetup()`/`isClusterSetup()` will fail. 

Here are the implications of these two points:

With the current design as shown here, if `InstanceConfig`/`ClusterConfig` doesn't exist, the attempt to create a thread pool will result an HelixException. I believe it's fine to **not** catch the exception and let it propagate up, because if `InstanceConfig`/`ClusterConfig` doesn't exist, it's an erroneous state. This aligns with what @jiajunwang is saying, and is the current design as we see it. 

Alternatively, we can catch the HelixException, log it, and use the default value (this is @narendly 's idea). 

I'm leaning towards the current design (let HelixException propagate upwards if configs don't exist). Anyone has a different opinion?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: @narendly Reminder on this conversation. 

I want to know if there's a good way to use `getIntField()` while handling the "not defined" case differently. If not, I will proceed with the current design.

NealSun96:
> Just to confirm, will the logic to report the current pool size be in a different PR?

Yes it will be a different PR.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Additionally, we are also updating `LiveInstance`, which I'm giving the same treatment. Read: if `InstanceConfig`, `ClusterConfig`, `LiveInstance` don't exist, the `TaskStateModelFactory` will **not** handle the exception and will let the process fail, because they should exist. I believe if they don't exist, somewhere else will also fail on participant side.

I'm open to your suggestions, what do you think? @jiajunwang @narendly

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -41,23 +47,15 @@
...
-
-          @Override
-          public Thread newThread(Runnable r) {
-            return new Thread(r, "TaskStateModelFactory-task_thread-" + threadId.getAndIncrement());
-          }
-        }));
+    this(manager, taskFactoryRegistry, null);
   }
 
+  // This constructor is only for internal usage. Do not use!
```
NealSun96, line 58: We talked about this in an early meeting, how we want to not let users to use this constructor, therefore I added this comment. 
However, I realized that with this constructor, we actually cannot set `CURRENT_TASK_THREAD_POOL_SIZE`, because we can't get the pool size. I'm now more inclined to disable this constructor altogether. @jiajunwang @narendly

helix-core/src/test/java/org/apache/helix/task/TestTaskStateModelFactory.java
```diff
@@ -0,0 +1,95 @@
+package org.apache.helix.task;
+
```
pkuwm, line 2: Apache license?

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
```diff
@@ -48,6 +49,7 @@
     TASK_EXEC_THREAD
   }
 
+  public static final int CURRENT_TASK_THREAD_POOL_SIZE_NOT_SET = -1;
```
pkuwm, line 52: I see 2 not set constants. Did you consider using one common constant and putting it in the constants class?

jiajunwang: in addition to what Huizhi said, shall we make it 40 for backward compatibility?

helix-core/src/test/java/org/apache/helix/task/TestTaskStateModelFactory.java
NealSun96: Not sure what happened there... Added it. Thanks for catching it!

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
NealSun96: These values all have different meanings (see one discussion above). Also, since they are all used only in one place (LiveInstance, ClusterConfig, InstanceConfig), they should stay in the scope.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -81,6 +79,12 @@ public Thread newThread(Runnable r) {
 
   @Override
   public TaskStateModel createNewStateModel(String resourceName, String partitionKey) {
+    if (_taskExecutor == null) {
```
jiajunwang, line 82: To be safe, add a sync control on this block. There is no guarantee we won't call createNewStateModel in parallel in the future.
And better to put all the logics inside the initializeTaskExecutor().

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -81,6 +79,12 @@ public Thread newThread(Runnable r) {
...
+      int taskThreadPoolSize = getTaskThreadPoolSize();
+      updateLiveInstanceWithThreadPoolSize(taskThreadPoolSize);
```
jiajunwang, line 84: Curious about this design.
The createNewStateModel() is called when we have partition assigned, right? But I think the controller needs to know the pool size before assign the tasks. How it works?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -41,23 +47,15 @@
...
-    this(manager, taskFactoryRegistry,
-        Executors.newScheduledThreadPool(TASK_THREADPOOL_SIZE, new ThreadFactory() {
-          private AtomicInteger threadId = new AtomicInteger(0);
-
-          @Override
-          public Thread newThread(Runnable r) {
-            return new Thread(r, "TaskStateModelFactory-task_thread-" + threadId.getAndIncrement());
-          }
-        }));
+    this(manager, taskFactoryRegistry, null);
```
jiajunwang, line 55: As mentioned below, I don't think we like lazy initialize the executor. Any reason for this change?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -102,4 +106,73 @@ public boolean isShutdown() {
...
+      if (verifyTargetThreadPoolSize(targetTaskThreadPoolSize)) {
+        return targetTaskThreadPoolSize;
+      }
+    }
+
+    return TaskConstants.DEFAULT_TASK_THREAD_POOL_SIZE;
+  }
+
+  /*
+   * Checks against the default values of -1 when pool sizes are not defined; we don't want -1's
```
jiajunwang, line 139: You are also excluding 0. Let's make the comment clear and fitting the code.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -102,4 +106,73 @@ public boolean isShutdown() {
...
+   */
+  private static boolean verifyTargetThreadPoolSize(int targetTaskThreadPoolSize) {
+    return targetTaskThreadPoolSize > 0;
+  }
+
+  /*
+   * Update LiveInstance with the current used thread pool size
+   */
+  private void updateLiveInstanceWithThreadPoolSize(int taskThreadPoolSize) {
```
jiajunwang, line 148: Instead of updating after the live node created, we shall ensure it is included when the node is created the first time. Or we may have some serious issue that the controller violates the capacity limitation.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Agreed on the sync part. 

Why do we need all logic inside `initializeTaskExecutor()`?

jiajunwang: Do you have a scenario that you need to call these methods separately? Put each line of code in separate methods are not efficient. Please try to balance the code.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Answering "why lazy initialization" question here as well:

When `TaskStateModelFactory` is initialized, it is usually during factory registration. However, factory registration can occur before HelixManager connection, or after. When it occurs before HelixManager connection, `TaskStateModelFactory` actually cannot access the configs, therefore it cannot create the thread pool before manager connection. 

We have 2 ways to make it work: 1. we enforce `TaskStateModelFactory` initialization to be after HelixManager connection, 2. we lazily initialize the thread pool. I chose 2 because it is backward compatible - clients don't need to change their existing code in which they registered `TaskStateModelFactory` before HelixManager connection (which exists). 

In terms of `createNewStateModel()`, it is called during message handling when the state model of the message doesn't exist for the resource name and partition key. From the code I read, I don't think "the controller needs to know the pool size before assign the tasks", because as we can see from the original code of this class, the task thread pool is never exposed anywhere other than `createNewStateModel()`. In fact, it's not used anywhere other than `createNewStateModel()` either, therefore making it safe to initialize in this function. 

Feel free to correct me. @jiajunwang

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Combined with the previous comment.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: This is a good point. Let me think about this.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
jiajunwang: 1. why we have to relies on the HelixManager to query for the config? You can use data accessor whenever it is requried. But if we don't even have the cluster/zk information, then it is not possible. This is one of the cons (or limitations) that we rely on ZK to pass the configuration.
2. The controller needs to know the real capacity otherwise how it ensures the assigned tasks fit the instance's capacity? I think this is the purpose of this design, no?
3. Alternatively, you can report the accepted capacity without really applying it to any executor. I mean, when liveinstance is created, put the capacity there. But at that moment, there is no real executor exists. Then when we create the executor, we apply for exactly the same number there.
To be clear, this is not a preferred design. In this way, the CURRENT_THREAD_POOL_SIZE becomes meaningless. But that is the only thing I can think of now.

NealSun96: Reading your later comment, I think I know what you mean by "controller needs to know the pool size". Let me think about this.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: By separating the code into multiple functions based on its purpose, I made it easier to understand. If I combine them, I either name the function `initializeTaskExecutorAndTaskMonitorAndUpdateLiveInstance()` or I force developers to read the code with no clue from the function titles. I'm sure it was clear for you to take a glance and understand what this block is doing because the function names explain them. 

This is a nit, so I don't have that strong if an opinion, but I believe it is correct to break down code into logical groups. To answer your question, "do you have a scenario that you need to call these methods separately?", the answer is yes. I could be utilizing `initializeTaskMonitor()` separately in the other constructor; I'm not doing it because I think that constructor should be gone (see conversation above).

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
NealSun96: @jiajunwang This is an interesting point, but wouldn't it be dangerous to use a default value like 40? If the actual thread pool is 20 and the write to LiveInstance fails, then the controller will be looking at an exaggerated number.

jiajunwang: What I meant here is actully since you have a default value anyway. There should not be a case we want the CURRENT_TASK_THREAD_POOL_SIZE_NOT_SET defined. We just always return 40 if no additional configuration is given.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
jiajunwang: It depends, to me the whole init logic is one logical group.
1. each method is small jumping between these methods to read the code is not comfortable.
2. Do you have a scenario that you need to call these methods separately? Or how you support the statement that they are different logical groups?
3. The method name does not need to contain all the logics inside. I don't quite get this point.

helix-core/src/main/java/org/apache/helix/model/InstanceConfig.java
```diff
@@ -56,11 +56,13 @@
...
     DELAY_REBALANCE_ENABLED,
     MAX_CONCURRENT_TASK,
-    INSTANCE_CAPACITY_MAP
+    INSTANCE_CAPACITY_MAP,
+    TARGET_TASK_THREAD_POOL_SIZE
   }
 
   public static final int WEIGHT_NOT_SET = -1;
   public static final int MAX_CONCURRENT_TASK_NOT_SET = -1;
+  public static final int TARGET_TASK_THREAD_POOL_SIZE_NOT_SET = -1;
```
narendly, line 65: private?

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
narendly: This field is not necessary.

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
```diff
@@ -190,6 +192,24 @@ public void setWebserviceUrl(String url) {
     _record.setSimpleField(LiveInstanceProperty.ZKPROPERTYTRANSFERURL.toString(), url);
   }
 
+  /**
+   * Get the current task thread pool size of the instance
+   * @return the current task thread pool size
+   */
+  public int getCurrentTaskThreadPoolSize() {
+    return _record.getIntField(LiveInstanceProperty.CURRENT_TASK_THREAD_POOL_SIZE.name(),
+        CURRENT_TASK_THREAD_POOL_SIZE_NOT_SET);
```
narendly, line 201: Please try to understand the proposed design well. This suggests to me that you should perhaps spend more time trying to understand what this field represents.

For example, what does it mean for an instance to have a task thread pool of size -1? (it is meaningless, we shouldn't create a task pool with size -1).

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -41,23 +47,15 @@
 
   private final HelixManager _manager;
   private final Map<String, TaskFactory> _taskFactoryRegistry;
-  private final ScheduledExecutorService _taskExecutor;
+  private ScheduledExecutorService _taskExecutor;
```
narendly, line 50: Update:

Neal and I had an offline discussion. To simplify the design, we will create a temporary ZK connection to read the task pool size configs and close right away. This would greatly simplify the code and this won't be that much of a burden since the connection is closed right away.

This is to populate the current task size pool at the time of LiveInstance (ephemeral node) initialization.

@NealSun96 Please remember to put a TODO here to potentially revisit the logic - ideally, we shouldn't have to create multiple ZK client connections when there already is a manager there.

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
NealSun96: If there is ever a case where CURRENT_TASK_THREAD_POOL_SIZE is not set, I don't think the controller should proceed. But @narendly raised a good point offline - it could be an old participant that doesn't have this field. I'm using the default 40 instead as there's no better options in that case. Thank you all for the input.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: With the recent changes, `initializeTaskMonitor()` is called by both constructors so it makes sense for it to be factored out. I'm not entirely convinced that it's correct to put all logic in one function - I always thought it's a good practice to break functions down when necessary into logical chunks to avoid big messy functions. 
Since this section of code is outdated, we can discuss in the new conversation if necessary.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Resolved offline, thank you @jiajunwang!

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
NealSun96: Discussed offline: this is meant to be a bit signaling an error, because field should always be set. But it's true that old participants don't have this field. I'm using the default 40 instead as there's no better options in that case. Thank you.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Fyi @jiajunwang

helix-core/src/test/java/org/apache/helix/task/TestTaskStateModelFactory.java
```diff
@@ -0,0 +1,124 @@
...
+    msds.stopServer();
+  }
+
+  @Test(dependsOnMethods = "testConfigAccessorCreationMultiZk")
+  public void testConfigAccessorCreationSingleZk() {
+    MockParticipantManager anyParticipantManager = _participants[0];
+
+    // Save previously-set system configs
+    String prevMultiZkEnabled = System.getProperty(SystemPropertyKeys.MULTI_ZK_ENABLED);
+    // Turn on multiZk mode in System config
```
narendly, line 107: Did you mean to say turn "off"? :)

helix-core/src/test/java/org/apache/helix/task/TestTaskStateModelFactory.java
```diff
@@ -0,0 +1,124 @@
...
+        .setInstanceConfig(anyParticipantManager.getClusterName(),
+            anyParticipantManager.getInstanceName(), instanceConfig);
+
+    // Start a msds server
+    // TODO: TestMultiZkHelixJavaApis already defined MSDS_SERVER_ENDPOINT, which goes into
+    // HttpRoutingDataReader and is recorded as final. As a result this test case has to use the
+    // same endpoint. There's no workaround at this moment.
+    final String msdsHostName = "localhost";
+    final int msdsPort = 11117;
+    final String msdsNamespace = "multiZkTest";
```
narendly, line 61: This is okay. I'm not sure if we want to make it a TODO, but "multiZkTest" is a generic enough a name, so I'm not too bothered by this at the moment. The reason this is tricky is that in production, in order to set up a different MSDS, you would have to change the JVM config and restart the JVM, thereby updating the `final` field.

I think the right TODO here is that now that we have a few tests that need to use the MSDS, so it would be a good idea to move this MSDS creation to the abstract test base so that we make it transparent to all tests, not a single test that tests multi-zk functionalities. What do you think? That way, we could make these values like msdsHostName, port, namespace all accessible to child classes, so you don't have to redefine them (just like how we do with ZK_ADDR and such).

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -709,6 +714,33 @@ public void setInstanceCapacityKeys(List<String> capacityKeys) {
...
+  /**
+   * Set the global target size of task thread pools for this cluster.
+   * @param globalTargetTaskThreadPoolSize - the new global target task thread pool size
+   * @throws IllegalArgumentException - when the provided new thread pool size is not greater than 0
+   */
+  public void setGlobalTargetTaskThreadPoolSize(int globalTargetTaskThreadPoolSize)
+      throws IllegalArgumentException {
+    if (globalTargetTaskThreadPoolSize <= 0) {
+      throw new IllegalArgumentException("globalTargetTaskThreadPoolSize must be greater than 0!");
+    }
```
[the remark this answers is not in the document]
NealSun96: I was thinking that negative values and 0 both produce unusable thread pools. But your second point is convincing - this is a legitimate use.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -102,4 +96,53 @@ public boolean isShutdown() {
...
+    }
+  }
+
+  /*
+   * Create a config accessor to get the thread pool size
+   */
+  protected ConfigAccessor createConfigAccessor() {
+    if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED)) {
+      String clusterName = _manager.getClusterName();
+      String shardingKey = clusterName.charAt(0) == '/' ? clusterName : "/" + clusterName;
```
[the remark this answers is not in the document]
NealSun96: In `ZkHelixManager`, the sharding building is done like this:
```
  private String buildShardingKey() {
    return _clusterName.charAt(0) == '/' ? _clusterName : "/" + _clusterName;
  }
```
I figured the state model factory should do the same. It doesn't hurt to check anyways.

helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkHelixJavaApis.java
```diff
@@ -149,6 +150,7 @@ public void beforeClass() throws Exception {
     System.setProperty(MetadataStoreRoutingConstants.MSDS_SERVER_ENDPOINT_KEY,
         "http://" + msdsHostName + ":" + msdsPort + "/admin/v2/namespaces/" + msdsNamespace);
 
+    HttpRoutingDataReader.reset();
```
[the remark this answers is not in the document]
NealSun96: `reset()` is to avoid racing over routing data; this test and `TestTaskStateModelFactory` both specify routing data using the same MSDS endpoint. Without `reset()`, the test that starts later will be using the cached routing data specified by the test that starts earlier and fail.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -109,7 +109,11 @@
     // https://github.com/apache/helix/wiki/Weight-aware-Globally-Evenly-distributed-Rebalancer#rebalance-coordinator
     //
     // Default to be true.
-    GLOBAL_REBALANCE_ASYNC_MODE
+    GLOBAL_REBALANCE_ASYNC_MODE,
+
+    // The target size of task thread pools for each participant. This is the global value
+    // that's used when participants don't specify their individual pool sizes.
+    GLOBAL_TARGET_TASK_THREAD_POOL_SIZE
```
jiajunwang, line 119: Can you share the reason why we change to GLOBAL_XXXX instead of DEFAULT_XXXX? I think default value is closer to how it is used. If participant specifies their values individually, it is not globally applied, right?

helix-core/src/test/java/org/apache/helix/task/TestTaskStateModelFactory.java
NealSun96: Yes, it was a copy-paste mistake.

helix-core/src/test/java/org/apache/helix/task/TestTaskStateModelFactory.java
NealSun96: That's exactly what I planned to do at first, but then I realized that `TestMultiZkHelixJavaApis` doesn't extend any class, which makes refactoring a bit more difficult. I will update to TODO accordingly.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -44,39 +48,29 @@
...
+        _manager.getInstanceName());
+    configAccessor.close();
+    _taskExecutor = createTaskExecutor(threadPoolSize);
+    _timerTaskExecutor = createTimerTaskExecutor();
+    initializeTaskMonitor();
   }
 
+  // FIXME: DO NOT USE! This size of provided thread pool will not be reflected to controller
+  // properly, the controller may over schedule tasks to this participant.
   public TaskStateModelFactory(HelixManager manager, Map<String, TaskFactory> taskFactoryRegistry,
```
jiajunwang, line 73: Deprecate?

And if anyone is using this method, what is the current thread pool size in the live instance node?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -44,39 +48,29 @@
   private final ScheduledExecutorService _taskExecutor;
   private final ScheduledExecutorService _timerTaskExecutor;
   private ThreadPoolExecutorMonitor _monitor;
-  public final static int TASK_THREADPOOL_SIZE = 40;
 
   public TaskStateModelFactory(HelixManager manager, Map<String, TaskFactory> taskFactoryRegistry) {
```
jiajunwang, line 52: You have some duplicate code in your 2 constructors. Why not keep the original style?
And, I still think merging the init logic in a single method or fewer methods make the code cleaner.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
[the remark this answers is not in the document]
jiajunwang: Why not make buildShardingKey() a util method? Will we have more and more this kind of code elsewhere?

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -1045,4 +1048,54 @@ private static void setNextJobPurgeTime(String workflow, long currentTime, long
...
+  public static int getTargetThreadPoolSize(ConfigAccessor configAccessor, String clusterName,
+      String instanceName) {
+    // Check instance config first for thread pool size
+    try {
+      InstanceConfig instanceConfig = configAccessor.getInstanceConfig(clusterName, instanceName);
+      if (instanceConfig != null) {
+        int targetTaskThreadPoolSize = instanceConfig.getTargetTaskThreadPoolSize();
+        if (targetTaskThreadPoolSize > 0) {
+          return targetTaskThreadPoolSize;
+        }
```
jiajunwang, line 1069: nit, It would be helpful to note what happens if it is "else" condition.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -1045,4 +1048,54 @@ private static void setNextJobPurgeTime(String workflow, long currentTime, long
...
+        LOG.warn("Got null as ClusterConfig for cluster {}. Returning default value. ",
+            clusterName);
+      }
+    } catch (HelixException e) {
+      LOG.warn(
+          "Encountered an exception while fetching ClusterConfig in cluster {}. Returning default value. Exception: ",
+          clusterName, e);
+    }
+
+    return TaskConstants.DEFAULT_TASK_THREAD_POOL_SIZE;
```
jiajunwang, line 1099: If it is HelixException, is it safe to just return the default value?

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -1045,4 +1048,54 @@ private static void setNextJobPurgeTime(String workflow, long currentTime, long
...
+
+  /**
+   * Get target thread pool size from InstanceConfig first; if that fails, get it from
+   * ClusterConfig; if that fails, fall back to the default value.
+   * @param configAccessor - accessor used for the configs
+   * @param clusterName - the cluster name for InstanceConfig and ClusterConfig
+   * @param instanceName - the instance name for InstanceConfig
+   * @return target thread pool size
+   */
+  public static int getTargetThreadPoolSize(ConfigAccessor configAccessor, String clusterName,
```
jiajunwang, line 1060: You have 2 very similar code blocks, especially the warning messages. Is it possible to make it concise? I understand the warning messages might look the same in that case, but it won't hurt IMO. Since we just need to know that the configuration is missing.
Duplicate code is a bigger problem.

helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkHelixJavaApis.java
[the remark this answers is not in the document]
jiajunwang: Shall we put this reset into the root test class? Or to the basic ZK API test class? So that we don't need to bother changing it if more tests modify the routing data.

helix-core/src/test/java/org/apache/helix/model/TestLiveInstance.java
```diff
@@ -0,0 +1,51 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import org.apache.helix.task.TaskConstants;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class TestLiveInstance {
```
jiajunwang, line 26: Can you checkout TestZKLiveInstanceData and see if you can merge them?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
narendly: @jiajunwang 
I shared this thought with Neal offline, but let me also share it here.

The reason I think it would be more appropriate to just add a FIXME or a warning here is because there's a slim chance that we might need to use this constructor. This might be useful later when we want to make the thread pool size change take place dynamically, without requiring a JVM restart.

I didn't think it'd be a good idea to deprecate and undeprecate public APIs in the future, so I've asked Neal to put a FIXME here. If you think otherwise, let us know.

helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkHelixJavaApis.java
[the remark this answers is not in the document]
narendly: If that's the case, we could do this in the` beforeClass()`.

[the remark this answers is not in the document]
NealSun96: @jiajunwang What you said makes the most sense but we can't do that now because `TestMultiZkHelixJavaApis.java` doesn't have a base test class; it should be done as a part of a refactoring ticket. I can move it to `beforeClass()`. @narendly

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: I can no longer use the original style because `_taskExecutor` is created differently - the first constructor's logic is no longer a subset of the second constructor's. 

As a result of that, I factored out the timer task thread pool creation etc into methods to avoid duplicate code.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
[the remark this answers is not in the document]
NealSun96: I was tempted to do it but I don't want to expand this PR with that change - there could always be debates about it, and I want this PR approved fast to unblock other items. If we don't have high urgency for this, can we have this as a later item?

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
NealSun96: Ok, adding a comment.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
NealSun96: When HelixException is thrown on this line, it means the cluster is not set up. I don't have a strong preference on "returning default" vs "rethrow the exception". What do you think?

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
NealSun96: This is the first thing I tried with this PR, and my conclusion was there is no way to not have duplicate lines here. The main reason is due to the "try catch" + "optional return" structure, making it very difficult to factor out. Also, since the getters have different signatures, that's an additional difficulty to factor it out. If you have a good way to make it not duplicate, I'd love to try it.

helix-core/src/test/java/org/apache/helix/model/TestLiveInstance.java
NealSun96: Combined.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: @jiajunwang To answer your question there will be no current thread pool size defined, so it will come out as the default value when accessed.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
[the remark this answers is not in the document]
NealSun96: Went ahead and did it.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: IMO, there could be 2 potential misunderstandings if we use GLOBAL.
1. This size is the total thread count of all the participant thread pool size.
2. This size will be applied to every single participant.

To avoid those confusing, using DEFAULT is easier.

Regarding the internal DEFAULT value which is 40,
1. it is internal only, so even we name the above one as DEFAULT, it won't confuse the users.
2. I think we only have NOT_SET which is -1. There is no DEFAULT TARGET thread pool size, right? The 40 is the default thread pool size which is corresponding to the current thread poll size.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
jiajunwang: That won't be the case. If we have a dynamic pool size later, the change will be done inside this class (configuration, or some new defined Helix executors) instead of being exposed as a public constructor, I guess.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
jiajunwang: I'm not convinced. It seems that if you put the get threadpool size and createTaskExecutor(threadPoolSize) into one method, then you can put this method as a parameter so you can do exactly what was here before.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
jiajunwang: I would prefer exception because it fails the logic earlier.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
NealSun96: As a counter argument, "global target" sounds like the target value that's applied globally, which is correct here. The confusion over "default" may also apply to developers too. 

Personally, I think different people have different interpretation to the field names, but what I can do is to provide well-written javadocs that explain clearly what these fields do on CRUD APIs, then the naming doesn't matter as much. What do you think?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: If it'll be us using it, how about we make it private? Users shouldn't gain access to it anyways, right?

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
NealSun96: Resolved offline. Now the logic rethrows exceptions.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
narendly: Ultimately, you can make the call. I still think global makes more sense in this context and more intuitive for devs to understand, but it seems that Jiajun has used "default" for the WAGED rebalancer related configs that serve a similar purpose.

I don't think it'd be hugely off either way, so you could either get a third opinion or make the call yourself :) But I agree that we should make this clear in both JavaDoc and a user manual of some sort.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
narendly: That could be another option. The point here is to prevent users from using it in order to retain control. But I'm not sure it's possible to make it private - isn't this required by the interface?

NealSun96: Deprecating it as resolved offline.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
NealSun96: Based on a vote within the dev team, we have decided on using "GLOBAL_XXX".

NealSun96:
Final commit message:
## Allow Configurable Thread Pool Size in TaskStateModelFactory ##
This PR changes TaskStateModelFactory logic such that it now obtains the thread pool size from InstanceConfig/ClusterConfig that may be specified by the users. This PR also added setters and getters for InstanceConfig/ClusterConfig for thread pool size fields. This PR also added thread pool size reporting during LiveInstance creation, with setters and getters for LiveInstance.

NealSun96:
Final commit message:
## Allow Configurable Thread Pool Size in TaskStateModelFactory ##
This PR changes TaskStateModelFactory logic such that it now obtains the thread pool size from InstanceConfig/ClusterConfig that may be specified by the users. This PR also added setters and getters for InstanceConfig/ClusterConfig for thread pool size fields. This PR also added thread pool size reporting during LiveInstance creation, with setters and getters for LiveInstance.

PR #974, opened 2020-04-28
Enforce result check for data accessors batch get calls to prevent partial batch read.

jiajunwang:
### Issues

#969 

### Description

Enforce result check for data accessors' batch read calls to prevent partial read.
This will help to ensure the main Helix logic does not calculate based on incomplete input.

### Tests

No new features or logic introduced. This PR only changes the usage of the accessors in the Helix logic to be more strict.

**zookeeper-api**

[INFO] Tests run: 20, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.226 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 20, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 6.486 s
[INFO] Finished at: 2020-04-27T23:06:28-07:00

**helix-core**

[INFO] Tests run: 1144, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,657.683 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 01:17 h
[INFO] Finished at: 2020-04-28T15:20:07-07:00

**helix-rest**

[INFO] Tests run: 159, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 44.191 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 52.589 s
[INFO] Finished at: 2020-04-27T23:05:45-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
@@ -139,6 +139,7 @@ public interface BaseDataAccessor<T> {
    * @param options Set the type of ZNode see the valid values in {@link AccessOption}
    * @return List of record data stored at each ZNode
    */
+  @Deprecated
   List<T> get(List<String> paths, List<Stat> stats, int options);
 
   /**
@@ -166,6 +167,7 @@ public interface BaseDataAccessor<T> {
    * @param options Set the type of ZNode see the valid values in {@link AccessOption}
    * @return A list of children of the parent ZNode
    */
+  @Deprecated
   List<T> getChildren(String parentPath, List<Stat> stats, int options);
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
@@ -30,7 +30,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.apache.helix.AccessOption;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.PropertyType;
-import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.common.controllers.ControlContextProvider;
 import org.apache.helix.controller.LogUtil;
 import org.apache.helix.model.ResourceAssignment;
@@ -42,6 +41,7 @@ import org.apache.helix.task.RuntimeJobDag;
 import org.apache.helix.task.TaskConstants;
 import org.apache.helix.task.WorkflowConfig;
 import org.apache.helix.task.WorkflowContext;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -194,9 +194,9 @@ public class TaskDataCache extends AbstractDataCache {
       }
     }
 
-    List<ZNRecord> contexts = accessor.getBaseDataAccessor().get(contextPaths, null, 0);
+    List<ZNRecord> contexts = accessor.getBaseDataAccessor().get(contextPaths, null, 0, true);
     List<ZNRecord> prevAssignments =
-        accessor.getBaseDataAccessor().get(prevAssignmentPaths, null, 0);
+        accessor.getBaseDataAccessor().get(prevAssignmentPaths, null, 0, true);
 
     for (int i = 0; i < contexts.size(); i++) {
       ZNRecord context = contexts.get(i);
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -41,6 +41,7 @@ import org.apache.helix.NotificationContext.Type;
 import org.apache.helix.PropertyKey;
 import org.apache.helix.PropertyPathConfig;
 import org.apache.helix.SystemPropertyKeys;
+import org.apache.helix.api.exceptions.HelixMetaDataAccessException;
 import org.apache.helix.api.listeners.BatchMode;
 import org.apache.helix.api.listeners.ClusterConfigChangeListener;
 import org.apache.helix.api.listeners.ConfigChangeListener;
@@ -525,7 +526,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
 
   private <T extends HelixProperty> List<T> preFetch(PropertyKey key) {
     if (_preFetchEnabled) {
-      return _accessor.getChildValues(key);
+      return _accessor.getChildValues(key, true);
     } else {
       return Collections.emptyList();
     }
@@ -600,7 +601,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
             case TARGET_EXTERNAL_VIEW: {
               // check if bucketized
               BaseDataAccessor<ZNRecord> baseAccessor = new ZkBaseDataAccessor<>(_zkClient);
-              List<ZNRecord> records = baseAccessor.getChildren(path, null, 0);
+              List<ZNRecord> records = baseAccessor.getChildren(path, null, 0, 1, 0);
               for (ZNRecord record : records) {
                 HelixProperty property = new HelixProperty(record);
                 String childPath = path + "/" + record.getId();
@@ -637,7 +638,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
               break;
             }
           }
-        } catch (ZkNoNodeException e) {
+        } catch (ZkNoNodeException | HelixMetaDataAccessException e) {
           logger.warn(
               "fail to subscribe child/data change. path: " + path + ", listener: " + _listener, e);
         }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -601,7 +601,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
             case TARGET_EXTERNAL_VIEW: {
               // check if bucketized
               BaseDataAccessor<ZNRecord> baseAccessor = new ZkBaseDataAccessor<>(_zkClient);
-              List<ZNRecord> records = baseAccessor.getChildren(path, null, 0, 1, 0);
+              List<ZNRecord> records = baseAccessor.getChildren(path, null, 0, 0, 0);
               for (ZNRecord record : records) {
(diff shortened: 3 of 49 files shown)


helix-core/src/main/java/org/apache/helix/messaging/DefaultMessagingService.java
```diff
@@ -186,7 +186,8 @@ private HelixDataAccessor getRecipientDataAccessor(final Criteria recipientCrite
       Map<String, String> sessionIdMap = new HashMap<String, String>();
       if (recipientCriteria.isSessionSpecific()) {
         Builder keyBuilder = targetDataAccessor.keyBuilder();
-        List<LiveInstance> liveInstances = targetDataAccessor.getChildValues(keyBuilder.liveInstances());
+        List<LiveInstance> liveInstances =
+            targetDataAccessor.getChildValues(keyBuilder.liveInstances(), true);
```
dasahcc, line 190: Are we retrying for these? Or we need some way to monitor the failure ones.

jiajunwang: Retry is in https://github.com/apache/helix/pull/970.
I thought about this. If the liveInstance list is not complete, I guess we will only send partial messages, I guess it would be bad since there might be some relationship between the messages if they are done within the same pipeline.

But I don't have a strong preference. After PR 970 is in, there is very little chance we see this error. Please let me know if you think this shall be false (and why). I can change it and comment with the reason.

dasahcc: I agree it will have little chance. Was worrying about if we throw an exception, it could totally miss the message delivery.

helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
```diff
@@ -166,6 +167,7 @@
    * @param options Set the type of ZNode see the valid values in {@link AccessOption}
    * @return A list of children of the parent ZNode
    */
+  @Deprecated
   List<T> getChildren(String parentPath, List<Stat> stats, int options);
 
   /**
```
lei-xia, line 173: What is the unit of retryInterval, second or millisecond?  Should be use long?  Usually have three consecutive parameters with the same type is not easy to use.

jiajunwang: Lei, let's address the interface design question in separate PR if necessary? This design is not changed here.

In the current change, I am retrying once in the callers. But zero retry also works, I think, since we have added the async internal retry on connectivity issues. Do you think we shall just make it zero retry?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -701,7 +701,7 @@ private void syncSessionToController(HelixManager manager) {
       }
     }
 
-    List<Message> newMessages = accessor.getProperty(keys);
+    List<Message> newMessages = accessor.getProperty(keys, true);
```
dasahcc, line 704: This is not a good idea. If we throws exception here, then there is no event to trigger message read. Then it could only relies on @mgao0 's periodical message refresh. But if user does not set this up?

Do not throw exception at least can read some message back and process them. Also when ST done with these partial messages, delete message operation can trigger another round of read. I know relies on this logic is not robust. But at least, it is not worse scenario than throw exception here.

jiajunwang: Good point. I will change it back and add this to the comment.
Please help to check the other changes. My goal is to make all calls strictly checked unless we have a good reason.

helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
jiajunwang: Since we already checked in the zkclient async call retry PR, let me change to zero retry for the newly changed callers in this PR.

helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
```diff
@@ -194,9 +194,9 @@ private void refreshContextsAndPreviousAssignments(HelixDataAccessor accessor) {
       }
     }
 
-    List<ZNRecord> contexts = accessor.getBaseDataAccessor().get(contextPaths, null, 0);
+    List<ZNRecord> contexts = accessor.getBaseDataAccessor().get(contextPaths, null, 0, true);
```
kaisun2000, line 197: Does it mean if passing true here, it would throw exception? Then how do we handle this exception?

kaisun2000: By the way, if passing false here, it would retry this batch async reads infinitely?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -525,7 +526,7 @@ public void invoke(NotificationContext changeContext) throws Exception {
 
   private <T extends HelixProperty> List<T> preFetch(PropertyKey key) {
     if (_preFetchEnabled) {
-      return _accessor.getChildValues(key);
+      return _accessor.getChildValues(key, true);
```
kaisun2000, line 529: Same question. Before this change, partial read would not terminate rest flow, so onLiveInstance(), onMessage() etc would be called. 
Now if exception thrown, this is not the case. This would 
1/ result onLiveInstance() etc not called, but the exception got caught in the ZkEvent thread.
2/ For Molly's change to add perodical checking messages PULL request, this would break the time thread which calls onMessage(). Basically if the exception thrown from onMessage() when timer thread calls this one, this would stop later timer thread I believe. (We can test. I recall Java Concurrency Controller book mentioned this issue there.) This would defeat the purpose of Molly's change.

kaisun2000:
There are too many places that previously read partial data but now thrown exception. We need to reason for each case that when exception thrown, what is the difference of behavior and what should be right behavior.

helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
jiajunwang: It will fail the current pipeline, which is what we want. Otherwise, the rebalance will be done based on incomplete information.
If false, it will return partial result.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang: "partial read would not terminate rest flow", true. But it will cause the controller to make the wrong decision. Which is even worse. To prevent this is the target of this change.

I don't think it will fail the ZkEvent thread. All the callers as I can find has been try-catched. Do you have an example of exception?

jiajunwang:
> There are too many places that previously read partial data but now thrown exception. We need to reason for each case that when exception thrown, what is the difference of behavior and what should be right behavior.

I believe, on the contrary, we need a reason for each case that does not throw an exception. The partial reads cause many problems. We need to be very strict.
Moreover, the batch retry logic has been checked in to ensure connectivity issues won't be a concern anymore.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
kaisun2000: Note, if exception thrown. For this CallbackHandler, before this change, partial read result would be feed to OnMessage() or OnWhatever; now, onMessage() would not be called, instead, zkEvent would get an exception, catch the exception nad do logging. This is what we want.

Then, in Molly's case, the timer thread can see exception too, right? The thing is that I suspect that timer's internal thread may not schedule more OnMessage() if the previous one has exception thrown.  This needs to be tested and verified. Otherwise, the timer thread may never work after exception thrown if my suspicion is right.

kaisun2000:
I agree, partial result hide problem. 

But theoretically, now you can have an exception previously you don't. Now the problem is that this exception may not be caught properly and that would cause other problems like pipeline stopping, or thread get terminated etc. For these, we need to evaluate case by case. 

This is my point -- all the changed places, the potential exception is handled properly.  Maybe you already considered them. Just a reminder.

helix-core/src/main/java/org/apache/helix/messaging/DefaultMessagingService.java
dasahcc: I still not suggest to have this for DefaultMessaging service. Because this changes behavior and it is used by customer. It may break their logic.

Except this one, others are in our control and should be fine for me.

jiajunwang: Reasonable, I will change this to false.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang: @kaisun2000, that change has not been checked in yet. But it is a good point.
@mgao0, Please be aware of what Kai mentioned.

jiajunwang:
Thanks for the review @dasahcc and @kaisun2000 
I have converted back some of the changes that enforce strictly check. Could you please take a look at the remaining ones?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
kaisun2000: Let me add this point to @mgao0's pull request.

kaisun2000:
@jiajunwang, I guess you see my point.
The other thinking is to let this api throw exception explicitly and run all unit test to see if they are ok. But that can be very time consuming.

I cannot review review each use case.  I am willing to trust you made all the right choice. I will give you approval.

kaisun2000:
go

jiajunwang:
> @jiajunwang, I guess you see my point.
> The other thinking is to let this api throw exception explicitly and run all unit test to see if they are ok. But that can be very time consuming.
> 
> I cannot review review each use case. I am willing to trust you made all the right choice. I will give you approval.

I have run all the unit test against the current change. But the main concern is that our unit test cannot cover all the scenarios. Hopefully, our stress test process will help to catch if anything left. But please let me know if any other practical tests that is in your mind. I'd like to run whatever test possible to ensure this is a safe change.

jiajunwang:

PR #845, opened 2020-03-02
Async write operation should not throw Exception for serializing error

jiajunwang:
### Issues

- [x] My PR addresses the following Helix issues and references them in the PR description:

#805 

### Description

- [x] Here are some details about my PR, including screenshots of any UI changes:

This change will make the async write operations return error through the async callback instead of throwing exceptions. This change will fix the batch write/create failure due to one single node serializing failure.
In addition, according to serializer interface definition, change ZK related serializers to throw ZkMarshallingError instead of ZkClientException.

### Tests

- [x] The following tests are written for this issue:

testAsyncWriteOperations

- [x] The following is the result of the "mvn test" command on the appropriate module:

Failed tests: 
  TestJobQueueCleanUp.testJobQueueAutoCleanUp » ThreadTimeout Method org.testng....
  TestWorkflowTermination.testWorkflowJobFail:251->verifyWorkflowCleanup:257 expected:<true> but was:<false>

Tests run: 1086, Failures: 2, Errors: 0, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------

retrun:

Tests run: 8, Failures: 0, Errors: 0, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 57.613 s
[INFO] Finished at: 2020-03-02T16:43:29-08:00
[INFO] ------------------------------------------------------------------------

### Commits

- [x] My commits all reference appropriate Apache Helix GitHub issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Documentation (Optional)

- [ ] In case of new functionality, my PR adds documentation in the following wiki page:

(Link the GitHub wiki you added)

### Code Quality

- [x] My diff has been formatted using helix-style.xml 
(helix-style-intellij.xml if IntelliJ IDE is used)

PR #976, opened 2020-04-28
Fix incorrect exception type and confusing assertion messages

### LINKED ISSUE #975: Assertion messages and exception type in TestZNRecordSizeLimit are incorrect
```
    try {
      _gZkClient.writeData(path2, largeRecord);
    } catch (HelixException e) {
      Assert.fail("Should not fail because data size is larger than 1M since compression applied");
    }
```

- The message should be "Should succeed because data size is smaller than 1M since compression applied"
- `HelixException` should be `ZkMarshallingError` because the exception type is changed in ZNRecord serializer.

pkuwm:
### Issues

Fixes #975 

### Description

The message should be "Should succeed because data size is **smaller** than 1M since compression applied"
`HelixException` should be `ZkMarshallingError` because the exception type is changed in ZNRecord serializer.

### Tests

TestZNRecordSizeLimit

Doesn't change any other code. Just ran this test.

Default Suite
Total tests run: 4, Failures: 0, Skips: 0

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestZNRecordSizeLimit.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestZNRecordSizeLimit.java
@@ -45,7 +45,7 @@ public class TestZNRecordSizeLimit extends ZkUnitTestBase {
   private static Logger LOG = LoggerFactory.getLogger(TestZNRecordSizeLimit.class);
 
   private static final String ASSERTION_MESSAGE =
-      "Should succeed because compressed data is smaller than 1M";
+      "Should succeed because compressed data is smaller than 1M. Caused by: ";
 
   @Test
   public void testZNRecordSizeLimitUseZNRecordSerializer() {
@@ -90,7 +90,7 @@ public class TestZNRecordSizeLimit extends ZkUnitTestBase {
     try {
       _gZkClient.writeData(path2, largeRecord);
     } catch (ZkMarshallingError e) {
-      Assert.fail(ASSERTION_MESSAGE);
+      Assert.fail(ASSERTION_MESSAGE + e);
     }
     record = _gZkClient.readData(path2);
     Assert.assertNotNull(record);
@@ -100,7 +100,7 @@ public class TestZNRecordSizeLimit extends ZkUnitTestBase {
     try {
       _gZkClient.writeData(path1, largeRecord);
     } catch (ZkMarshallingError e) {
-      Assert.fail(ASSERTION_MESSAGE);
+      Assert.fail(ASSERTION_MESSAGE + e);
     }
     ZNRecord recordNew = _gZkClient.readData(path1);
     try {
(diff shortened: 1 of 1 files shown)


helix-core/src/test/java/org/apache/helix/manager/zk/TestZNRecordSizeLimit.java
```diff
@@ -86,8 +89,8 @@ public void testZNRecordSizeLimitUseZNRecordSerializer() {
     _gZkClient.createPersistent(path2, true);
     try {
       _gZkClient.writeData(path2, largeRecord);
-    } catch (HelixException e) {
-      Assert.fail("Should not fail because data size is larger than 1M since compression applied");
+    } catch (ZkMarshallingError e) {
```
jiajunwang, line 92: What is this test log testing?

If the expectation is no error, then we don't care what is the exception. If the operation fails, just throw it. If we use Assert.fail(), we get a message, but we are actually not sure the detail of the exception, because we never include it in the assert fail message.
If the expectation is we have an error, then the assert in the catch block will fail the test.

Can we make it straight?

pkuwm: We are testing a `ZkMarshallingError` should not be thrown by the serializer. This message specifies the reason why this bad exception is thrown.
FYI, exception stack trace is added to the message.

jiajunwang: I don't have a strong preference here. But I think it is redundant try-catch. Otherwise, why we are not adding this try-catch then throw exception blocks to all the test code? Because all test code, as long as it is not testing for the negative cases, will have the same assumption that we shall not have a certain Exception thrown.

pkuwm: I understand your point. And I agree we shouldn’t do redundant code. But in this specific case here, since compression is such a critical part, I would prefer to make it bold. If there is anything change in the logic and triggers the test failed, the message should tell the reason. And one more reason I use assert.fail is to avoid that if someone changes the test and catch the exception Without failing the test , the exception thrown from here may not fail the test. This is what I would like to prevent.

dasahcc:
Something wrong with your change? Only on test change, where is the code change? @pkuwm

pkuwm:
> Something wrong with your change? Only on test change, where is the code change? @pkuwm

@dasahcc No main code change in this PR. This is a minor fix for the test. The exception change was already done in another PR: [here](https://github.com/apache/helix/pull/845)

pkuwm:

```
In TestZNRecordSizeLimit, the assertion message should indicate the data is smaller than 1 MB threshold. HelixException should be ZkMarshallingError because the exception type is changed in ZNRecord serializer.
```

Issue #985, opened 2020-05-01
Auto rebalancer state assignment logic triggers extra top state transitions.


jiajunwang:
1. The current states are: [N1:M, N2:S, N3,S]
2. The desired states are: [N4:M, N2:S, N1:S]
3. Due to the sorting logic based on current states, we will have a transient preference list ordered like: [N2, N1, N4]. In which case, the controller will assign master to N2 before N4 has a slave state replica.
4. When N4 finishes the Offline to Slave transition, the same sorting logic will sort the preference list to be: [N4, N2, N1]. Then we have another mastership handoff.
We don't want step 3. But only the state transition in step 4.

PR #986, opened 2020-05-01
Adjust the auto rebalancer state assignment logic to reduce top state transition.

jiajunwang:
### Issues

#985

### Description

The old state assignment logic assign the states to selected nodes according to the priority of the current replica state that is on the instance. Moreover, the sorting algorithm is designed to prioritize both current topstate and current secondary states equally. The result is that we will have premature mastership handoff to a current seconardy state host before the real desired master host is ready. For example,
1. The current states are: [N1:M, N2:S, N3,S]
2. The desired states are: [N4:M, N2:S, N1:S]
3. Due to the sorting logic based on current states, we will have a transient preference list ordered like: [N2, N1, N4]. In which case, the controller will assign master to N2 before N4 has a slave state replica.
4. When N4 finishes the Offline to Slave transition, the same sorting logic will sort the preference list to be: [N4, N2, N1]. Then we have another mastership handoff.
To be clear, we don't want step 3. But only the state transition in step 4.

In this PR, we refactor the sorting logic so that it will only move the master whenever the candidate has a "ready" state replica, in which case, only one mastership handoff happens.

### Tests

Add the test scenario to TestAbstractRebalancer.

[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestWorkflowTimeout.testWorkflowTimeoutWhenWorkflowCompleted:116 expected:<true> but was:<false>
[INFO] 
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time: 01:21 h
[INFO] Finished at: 2020-04-30T23:30:07-07:00

Rerun:
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 30.931 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 37.212 s
[INFO] Finished at: 2020-04-30T23:32:13-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
@@ -305,45 +305,60 @@ public abstract class AbstractRebalancer<T extends BaseControllerDataProvider> i
     Set<String> liveAndEnabled = new HashSet<>(liveInstances);
     liveAndEnabled.removeAll(disabledInstancesForPartition);
 
-    for (String state : statesPriorityList) {
-      // Use the the specially ordered preferenceList for choosing instance for top state.
-      if (state.equals(statesPriorityList.get(0))) {
-        List<String> preferenceListForTopState = new ArrayList<>(preferenceList);
-        Collections.sort(preferenceListForTopState,
-            new TopStatePreferenceListComparator(currentStateMap, stateModelDef));
-        preferenceList = preferenceListForTopState;
-      }
+    // Sort the instances based on replicas' state priority in the current state
+    List<String> sortedPreferenceList = new ArrayList<>(preferenceList);
+    sortedPreferenceList.sort(new StatePriorityComparator(currentStateMap, stateModelDef));
 
+    // Assign the state to the instances that appear in the preference list.
+    for (String state : statesPriorityList) {
       int stateCount =
           getStateCount(state, stateModelDef, liveAndEnabled.size(), preferenceList.size());
       for (String instance : preferenceList) {
         if (stateCount <= 0) {
-          break;
+          break; // continue assigning for the next state
         }
-        if (!assigned.contains(instance) && liveAndEnabled.contains(instance)) {
-          if (HelixDefinedState.ERROR.toString().equals(currentStateMap.get(instance))) {
-            bestPossibleStateMap.put(instance, HelixDefinedState.ERROR.toString());
-          } else {
-            bestPossibleStateMap.put(instance, state);
-            stateCount--;
+        if (assigned.contains(instance) || !liveAndEnabled.contains(instance)) {
+          continue; // continue checking for the next available instance
+        }
+        String proposedInstance = instance;
+        // Additional check and alternate the assignment for reducing top state handoff.
+        if (state.equals(stateModelDef.getTopState()) && !stateModelDef.getSecondTopStates()
+            .contains(currentStateMap.getOrDefault(instance, stateModelDef.getInitialState()))) {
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
```diff
@@ -305,45 +305,60 @@ public static int getStateCount(String state, StateModelDefinition stateModelDef
...
         }
-        if (!assigned.contains(instance) && liveAndEnabled.contains(instance)) {
-          if (HelixDefinedState.ERROR.toString().equals(currentStateMap.get(instance))) {
-            bestPossibleStateMap.put(instance, HelixDefinedState.ERROR.toString());
-          } else {
-            bestPossibleStateMap.put(instance, state);
-            stateCount--;
+        String proposedInstance = instance;
+        // Additional check and alternate the assignment for reducing top state handoff.
+        if (state.equals(stateModelDef.getTopState()) && !stateModelDef.getSecondTopStates()
```
dasahcc, line 325: I think you already sorted the preference list based on the state. Why not just replace the looping of preference list with the sorted list but add this logic?

This logic makes two order of preference list intersected each other that logic is not very clear.

jiajunwang: I thought about it. But even I move this logic to the sorting logic, the sorting itself will be comparing one list with 2 logics. This interesting trick is just moved to another place. I don't have a way to eliminate it for now.
Then why I want to put it here instead of the sorting logic? Because the sorting logic is away from this business logic block. That would be even harder for me to reason even with comments.Put it here, I

jiajunwang: Please consider this as a skiplist type design/implementation.

dasahcc: I see you point. Purely relies on sorting is fine because the previous author use it in wrong way. We suppose to sort it before looping. 

Anyway, I am OK to keep the current behavior. But can you add a TODO here. To remind that we can refactor the logic based on sorting with comprehensive test in the future?

jiajunwang: Not exactly. Actually, I don't agree this is a TODO work to put everything in a sorting method. That is just an alternative design, and I don't think it is cleaner than the current one.
Considering the proposed universal sorting method, it is actually merging 2 orders (preference list order and the current state order). However, there is no clear way to define the priority score.
Note the logic here is that if the node is not on secondary state or above but it is attempted to be a master, we should find alternatives according to the current state order.
This logic cannot be done with only the 2 orders. We have to know the proposed states assignment. For example, how many top states and where to assign based on the preference state.
Give this, to merging the sort in one method (instead of the current design), we will need to:
1. propose a state assignment based on the preference list
2. input this proposed state mapping as one additional input to the sorting method, and then sort accordingly.
3. calculate another state assignment based on the re-sorted list.

It seems more confusing and slower than the current design.

dasahcc: Hmm. The sorting logic could be complicated. Alright, let's revisit the logic if it requires refactoring later.

jiajunwang:

PR #923, opened 2020-03-31
Fix the scheduling decision for multiple currentStates

alirezazamani:
### Issues
- [x] My PR addresses the following Helix issues and references them in the PR title:
Fixes #461
Fixes #922

### Description
- [x] Here are some details about my PR, including screenshots of any UI changes:
This commit specifically targets the scenario where there are multiple currentstates and previous assignment existed for one partition. In this case, since paMap has only one field for the mapping, the final result of scheduling is not deterministic. In this case, tasks can be stuck in INIT -> RUNNING and RUNNING -> DROPPED on the wrong instance. This PR fixes this scenario.

### Tests
- [x] The following tests are written for this issue:
TestTargetedTaskStateChange
TestTaskSchedulingTwoCurrentStates

- [x] The following is the result of the "mvn test" command on the appropriate module:
Test Result 1: mvn test
Failed tests:
  TestWorkflowTermination.testWorkflowPausedTimeout:170->verifyWorkflowCleanup:257 expected:<true> but was:<false>

Tests run: 1094, Failures: 1, Errors: 0, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:13 h
[INFO] Finished at: 2020-03-30T17:11:26-07:00
[INFO] ------------------------------------------------------------------------

The failed test has passed when I ran it individually.

### Commits

- [x] My commits all reference appropriate Apache Helix GitHub issues in their subject lines, and I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Code Quality

- [x] My diff has been formatted using helix-style.xml

Issue #922, opened 2020-03-31
Continuous wrong assignment of targeted tasks


alirezazamani:
There are scenarios where there are multiple currentstates and previous assignment existed in ZK for one partition. This will confused the controller for the task scheduling stage. In this case, since paMap has only place for each task, the final result of scheduling is not deterministic. In this case, tasks can be stuck in INIT -> RUNNING and RUNNING -> DROPPED on the wrong instance.

Issue #461, opened 2019-09-10
Investigate FixedTargetTaskAssignmentCalculator


alirezazamani:
Right now the targeted assignment for targeted jobs relies on jobContext which might be different from Currentstate.

alirezazamani:
Here is the outcome of the investigation on FixedTargetedTasks:
1- These tasks will be assigned to the instances based on the on the outcome of the class:FixedTargetTaskAssignmentCalculator -> function:computeAssignmentAndChargeResource.

2- In the function:updatePreviousAssignedTasksStatus of the class:AbstractTaskDispatcher, whenever currentState is INIT (currentState==INIT or (currentState==null and Context==INIT) or (currentState==null and Context==null)), this partition will be added to assignedPartitions and will be excluded from being assignment in the method mentioned in Step1. In conclusion, whenever this task is targeted task, it makes sense to not take any decision about this task and let the FixedTargetTaskAssignmentCalculator to decide on the new assignment of the targeted task.

3- However, for the cases where current state is actually INIT, we might not have any assignment, because in function:computeAssignmentAndChargeResource, line 260: in condition "prevInstance.equals(instance)", we break and there won't be any assignment. As a result, this break should be removed which allows this function to take new decision for the tasks.

4- Finally, with these changes, One of the tests will fails. This test is called TestTaskNumAttempts.
The reason is potentially related to pending messages and race condition that can happen when there is pending message for the tasks.

5- More investigation is needed.

PR #994, opened 2020-05-04
Remove the scheduling decision based on PreviousAssignment

### LINKED ISSUE #984: TODO: Scheduling the tasks without considering PreviousAssignment
It is better to avoid having multiple sources of truth while making decisions on the tasks that have been already scheduled. Hence, it is better to rely in CurrentState completely and get rid of PreviousAssignment.

alirezazamani:
### Issues
Fixes #984 

### Description
In this commit, several parts of the code have been removed in order to avoid
making scheduling decision based on PreviousAssignment. Instead we are relying on
the CurrentState. 
In order to facilitate this, the tasks that reaches the terminal state 
(Error, Time_out, Aborted and completed) will be dropped and rescheduled in later
pipeline if needed (If they need to be rescheduled). This helps us to reduce the
number of existing current states for each task. 

### Tests
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestWorkflowTermination.testWorkflowJobFail:251->verifyWorkflowCleanup:257 expected:<true> but was:<false>
[ERROR]   TestWorkflowTimeout.testWorkflowTimeoutWhenWorkflowCompleted:116 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1144, Failures: 2, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:21 h
[INFO] Finished at: 2020-05-01T15:25:20-07:00

The failed tests has passed when I ran it individually. 

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
+++ b/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
@@ -66,7 +66,7 @@ public abstract class AbstractTaskDispatcher {
   // Job Update related methods
 
   public void updatePreviousAssignedTasksStatus(
-      Map<String, SortedSet<Integer>> prevInstanceToTaskAssignments, Set<String> excludedInstances,
+      Map<String, SortedSet<Integer>> currentInstanceToTaskAssignments, Set<String> excludedInstances,
       String jobResource, CurrentStateOutput currStateOutput, JobContext jobCtx, JobConfig jobCfg,
       ResourceAssignment prevTaskToInstanceStateAssignment, TaskState jobState,
       Map<String, Set<Integer>> assignedPartitions, Set<Integer> partitionsToDropFromIs,
@@ -78,11 +78,11 @@ public abstract class AbstractTaskDispatcher {
     AssignableInstanceManager assignableInstanceManager = cache.getAssignableInstanceManager();
 
     // Iterate through all instances
-    for (String instance : prevInstanceToTaskAssignments.keySet()) {
+    for (String instance : currentInstanceToTaskAssignments.keySet()) {
       assignedPartitions.put(instance, new HashSet<>());
 
       // Set all dropping transitions first. These are tasks coming from Participant disconnects
-      // that have some active current state (INIT or RUNNING) and the requestedState of DROPPED.
+      // and have the requestedState of DROPPED.
       // These need to be prioritized over any other state transitions because of the race condition
       // with the same pId (task) running on other instances. This is because in paMap, we can only
       // define one transition per pId
@@ -99,7 +99,7 @@ public abstract class AbstractTaskDispatcher {
       }
 
       // If not an excluded instance, we must instantiate its entry in assignedPartitions
-      Set<Integer> pSet = prevInstanceToTaskAssignments.get(instance);
+      Set<Integer> pSet = currentInstanceToTaskAssignments.get(instance);
 
       // We need to remove all task pId's to be dropped because we already made an assignment in
       // paMap above for them to be dropped. The following does this.
@@ -107,8 +107,7 @@ public abstract class AbstractTaskDispatcher {
         pSet.removeAll(tasksToDrop.get(instance));
       }
 
-      // Used to keep track of partitions that are in one of the final states: COMPLETED, TIMED_OUT,
-      // TASK_ERROR, ERROR.
+      // Used to keep track of partitions that are in one the INIT or DROPPED states.
       Set<Integer> donePartitions = new TreeSet<>();
       for (int pId : pSet) {
         final String pName = pName(jobResource, pId);
@@ -120,17 +119,6 @@ public abstract class AbstractTaskDispatcher {
               "Instance {} does not match the assigned participant for pId {} in the job context. Skipping task scheduling.",
               instance, pId);
           continue;
-        }
-        // This avoids a race condition in the case that although currentState is in the following
-        // error condition, the pending message (INIT->RUNNNING) might still be present.
-        // This is undesirable because this prevents JobContext from getting the proper update of
-        // fields including task state and task's NUM_ATTEMPTS
-        if (currState == TaskPartitionState.ERROR || currState == TaskPartitionState.TASK_ERROR
-            || currState == TaskPartitionState.TIMED_OUT
-            || currState == TaskPartitionState.TASK_ABORTED) {
-          // Do not increment the task attempt count here - it will be incremented at scheduling
-          // time
-          markPartitionError(jobCtx, pId, currState);
         }
 
         // Check for pending state transitions on this (partition, instance). If there is a pending
@@ -242,16 +230,16 @@ public abstract class AbstractTaskDispatcher {
         }
           break;
         case COMPLETED: {
-          // The task has completed on this partition. Mark as such in the context object.
-          donePartitions.add(pId);
+          // The task has completed on this partition. Drop it from the instance and add it to assignedPartitions in
+          // order to avoid scheduling it again in this pipeline.
+          assignedPartitions.get(instance).add(pId);
+          paMap.put(pId, new PartitionAssignment(instance, TaskPartitionState.DROPPED.name()));
           if (LOG.isDebugEnabled()) {
             LOG.debug(String.format(
                 "Task partition %s has completed with state %s. Marking as such in rebalancer context.",
                 pName, currState));
           }
           partitionsToDropFromIs.add(pId);
-          markPartitionCompleted(jobCtx, pId);
-
           // This task is COMPLETED, so release this task
           assignableInstanceManager.release(instance, taskConfig, quotaType);
         }
@@ -263,7 +251,11 @@ public abstract class AbstractTaskDispatcher {
         case TASK_ABORTED:
 
         case ERROR: {
-          donePartitions.add(pId); // The task may be rescheduled on a different instance.
+          // First make this task which is in terminal state to be dropped.
+          // Later on, in next pipeline in handleAdditionalAssignments, the task will be retried if possible.
+          // (meaning it is not ABORTED and max number of attempts has not been reached yet)
+          assignedPartitions.get(instance).add(pId);
+          paMap.put(pId, new PartitionAssignment(instance, TaskPartitionState.DROPPED.name()));
           if (LOG.isDebugEnabled()) {
             LOG.debug(String.format(
                 "Task partition %s has error state %s with msg %s. Marking as such in rebalancer context.",
@@ -389,13 +381,57 @@ public abstract class AbstractTaskDispatcher {
       return stateFromContext == null ? TaskPartitionState.INIT : stateFromContext;
     }
     TaskPartitionState currentState = TaskPartitionState.valueOf(currentStateString);
-    jobCtx.setAssignedParticipant(pId, instance);
-    jobCtx.setPartitionState(pId, currentState);
-    String taskMsg = currentStateOutput.getInfo(jobResource, new Partition(pName), instance);
-    if (taskMsg != null) {
-      jobCtx.setPartitionInfo(pId, taskMsg);
-    }
+    // Update job context based on current state
+    updatePartitionInformationInJobContext(currentStateOutput, jobResource, currentState, jobCtx,
+        pId, pName, instance);
     return currentState;
+  }
+
+  /**
+   * Based on the CurrentState of this task and Context information, the task information in the job
+   * context gets updated.
+   * @param currentStateOutput
+   * @param jobResource
+   * @param currentState
+   * @param jobCtx
+   * @param pId
+   * @param pName
+   * @param instance
+   */
+  private void updatePartitionInformationInJobContext(CurrentStateOutput currentStateOutput,
+      String jobResource, TaskPartitionState currentState, JobContext jobCtx, Integer pId,
+      String pName, String instance) {
+    // The assigned participant needs to be updated regardless of the current state and context
+    // information because it will prevent controller to stuck in race condition while there is two
+    // current states. In the updatePreviousAssignedTasksStatus, we check
+    // instance.equals(jobCtx.getAssignedParticipant(pId)) and bypass the assignment if instance is
+    // not equal to job context's AssignedParticipant for this pId.
+    jobCtx.setAssignedParticipant(pId, instance);
+    // If job context needs to be updated with new state, update it accordingly
+    // This check is necessary because we are relying on current state and we do not want to update
+    // context as long as current state existed. We just want to update context information
+    // (specially finish time) once.
+    if (!currentState.equals(jobCtx.getPartitionState(pId))) {
+      jobCtx.setPartitionState(pId, currentState);
+      String taskMsg = currentStateOutput.getInfo(jobResource, new Partition(pName), instance);
+      if (taskMsg != null) {
+        jobCtx.setPartitionInfo(pId, taskMsg);
+      }
+      if (currentState == TaskPartitionState.COMPLETED) {
+        markPartitionCompleted(jobCtx, pId);
+      }
+      // This avoids a race condition in the case that although currentState is in the following
+      // error condition, the pending message (INIT->RUNNNING) might still be present.
+      // This is undesirable because this prevents JobContext from getting the proper update of
+      // fields including task state and task's NUM_ATTEMPTS
+      if (currentState == TaskPartitionState.ERROR || currentState == TaskPartitionState.TASK_ERROR
+          || currentState == TaskPartitionState.TIMED_OUT
+          || currentState == TaskPartitionState.TASK_ABORTED) {
+        // Do not increment the task attempt count here - it will be incremented at scheduling
+        // time
+        markPartitionError(jobCtx, pId, currentState);
+      }
+    }
   }
 
   /**
@@ -511,7 +547,7 @@ public abstract class AbstractTaskDispatcher {
   // Compute real assignment from theoretical calculation with applied throttling
   // This is the actual assigning part
   protected void handleAdditionalTaskAssignment(
-      Map<String, SortedSet<Integer>> prevInstanceToTaskAssignments, Set<String> excludedInstances,
+      Map<String, SortedSet<Integer>> currentInstanceToTaskAssignments, Set<String> excludedInstances,
       String jobResource, CurrentStateOutput currStateOutput, JobContext jobCtx,
       final JobConfig jobCfg, final WorkflowConfig workflowConfig, WorkflowContext workflowCtx,
       final WorkflowControllerDataProvider cache,
@@ -580,7 +616,7 @@ public abstract class AbstractTaskDispatcher {
       // TODO: isRebalanceRunningTask() was originally put in place to allow users to move
       // ("rebalance") long-running tasks, but there hasn't been a clear use case for this
       // Previously, there was a bug in the condition above (it was || where it should have been &&)
-      dropRebalancedRunningTasks(tgtPartitionAssignments, prevInstanceToTaskAssignments, paMap,
+      dropRebalancedRunningTasks(tgtPartitionAssignments, currentInstanceToTaskAssignments, paMap,
           jobCtx);
     }
 
@@ -588,11 +624,11 @@ public abstract class AbstractTaskDispatcher {
     if (!TaskUtil.isGenericTaskJob(jobCfg) && existsLiveInstanceOrCurrentStateChange) {
       // Drop current jobs only if they are assigned to a different instance, regardless of
       // the jobCfg.isRebalanceRunningTask() setting
-      dropRebalancedRunningTasks(tgtPartitionAssignments, prevInstanceToTaskAssignments, paMap,
+      dropRebalancedRunningTasks(tgtPartitionAssignments, currentInstanceToTaskAssignments, paMap,
           jobCtx);
     }
     // Go through ALL instances and assign/throttle tasks accordingly
-    for (Map.Entry<String, SortedSet<Integer>> entry : prevInstanceToTaskAssignments.entrySet()) {
+    for (Map.Entry<String, SortedSet<Integer>> entry : currentInstanceToTaskAssignments.entrySet()) {
       String instance = entry.getKey();
       if (!tgtPartitionAssignments.containsKey(instance)) {
         // There is no assignment made for this instance, so it is safe to skip
--- a/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
+++ b/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
@@ -107,7 +107,7 @@ public abstract class AbstractTaskDispatcher {
         pSet.removeAll(tasksToDrop.get(instance));
       }
 
-      // Used to keep track of partitions that are in one the INIT or DROPPED states.
+      // Used to keep track of partitions that are in either INIT or DROPPED states
       Set<Integer> donePartitions = new TreeSet<>();
       for (int pId : pSet) {
         final String pName = pName(jobResource, pId);
@@ -401,9 +401,10 @@ public abstract class AbstractTaskDispatcher {
   private void updatePartitionInformationInJobContext(CurrentStateOutput currentStateOutput,
       String jobResource, TaskPartitionState currentState, JobContext jobCtx, Integer pId,
       String pName, String instance) {
-    // The assigned participant needs to be updated regardless of the current state and context
-    // information because it will prevent controller to stuck in race condition while there is two
-    // current states. In the updatePreviousAssignedTasksStatus, we check
+    // The assignedParticipant field needs to be updated regardless of the current state and context
+    // information because it will prevent controller to assign the task to the wrong participant
+    // for targeted tasks when two CurrentStates exist for one task.
+    // In the updatePreviousAssignedTasksStatus, we check
     // instance.equals(jobCtx.getAssignedParticipant(pId)) and bypass the assignment if instance is
     // not equal to job context's AssignedParticipant for this pId.
     jobCtx.setAssignedParticipant(pId, instance);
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -99,16 +99,15 @@ public void updatePreviousAssignedTasksStatus(
...
 
       // We need to remove all task pId's to be dropped because we already made an assignment in
       // paMap above for them to be dropped. The following does this.
       if (tasksToDrop.containsKey(instance)) {
         pSet.removeAll(tasksToDrop.get(instance));
       }
 
-      // Used to keep track of partitions that are in one of the final states: COMPLETED, TIMED_OUT,
-      // TASK_ERROR, ERROR.
+      // Used to keep track of partitions that are in one the INIT or DROPPED states.
```
narendly, line 110: Did you mean "that are in either INIT or DROPPED states"?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -242,16 +230,16 @@ public void updatePreviousAssignedTasksStatus(
         }
           break;
         case COMPLETED: {
-          // The task has completed on this partition. Mark as such in the context object.
-          donePartitions.add(pId);
+          // The task has completed on this partition. Drop it from the instance and add it to assignedPartitions in
+          // order to avoid scheduling it again in this pipeline.
+          assignedPartitions.get(instance).add(pId);
+          paMap.put(pId, new PartitionAssignment(instance, TaskPartitionState.DROPPED.name()));
```
narendly, line 236: I don't know if this would be the correct thing to do.

Sure, once a task is complete, it could be dropped from the node, but you also want to keep the state of the task in COMPLETED in JobContext.

I'm afraid doing this might cause the otherwise-completed task's state to change to DROPPED. Then we would have no way of knowing if this task just got dropped or completed and dropped.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -263,7 +251,11 @@ public void updatePreviousAssignedTasksStatus(
         case TASK_ABORTED:
 
         case ERROR: {
-          donePartitions.add(pId); // The task may be rescheduled on a different instance.
+          // First make this task which is in terminal state to be dropped.
+          // Later on, in next pipeline in handleAdditionalAssignments, the task will be retried if possible.
+          // (meaning it is not ABORTED and max number of attempts has not been reached yet)
+          assignedPartitions.get(instance).add(pId);
+          paMap.put(pId, new PartitionAssignment(instance, TaskPartitionState.DROPPED.name()));
```
narendly, line 258: Again, I am not sure if this would cause the JobContext to have the wrong state for the task partitions. Did you check? What state does the task go into between being dropped and getting rescheduled? (It should be ERROR or TASK_ABORTED, or whatever it was before being dropped)

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -389,13 +381,57 @@ private TaskPartitionState updateJobContextAndGetTaskCurrentState(
...
+   * @param currentState
+   * @param jobCtx
+   * @param pId
+   * @param pName
+   * @param instance
+   */
+  private void updatePartitionInformationInJobContext(CurrentStateOutput currentStateOutput,
+      String jobResource, TaskPartitionState currentState, JobContext jobCtx, Integer pId,
+      String pName, String instance) {
+    // The assigned participant needs to be updated regardless of the current state and context
```
narendly, line 404: the assignedParticipant "field" in JobContext needs to be updated...

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -389,13 +381,57 @@ private TaskPartitionState updateJobContextAndGetTaskCurrentState(
...
+    // information because it will prevent controller to stuck in race condition while there is two
```
narendly, line 405: ... because this will prevent the controller from being stuck in a race condition when there are two current states found for a single task.

-> Okay, but what kind of race condition are we talking about?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -389,13 +381,57 @@ private TaskPartitionState updateJobContextAndGetTaskCurrentState(
...
+    // current states. In the updatePreviousAssignedTasksStatus, we check
+    // instance.equals(jobCtx.getAssignedParticipant(pId)) and bypass the assignment if instance is
+    // not equal to job context's AssignedParticipant for this pId.
```
narendly, line 408: Does this help us avoid the race condition?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -389,13 +381,57 @@ private TaskPartitionState updateJobContextAndGetTaskCurrentState(
...
     jobCtx.setAssignedParticipant(pId, instance);
-    jobCtx.setPartitionState(pId, currentState);
-    String taskMsg = currentStateOutput.getInfo(jobResource, new Partition(pName), instance);
-    if (taskMsg != null) {
-      jobCtx.setPartitionInfo(pId, taskMsg);
+    // If job context needs to be updated with new state, update it accordingly
+    // This check is necessary because we are relying on current state and we do not want to update
+    // context as long as current state existed. We just want to update context information
```
narendly, line 412: -> update context while its current state exists...

But why do we not want to update the context when there's a current state for it? Shouldn't we update the context based on the latest current state?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -389,13 +381,57 @@ private TaskPartitionState updateJobContextAndGetTaskCurrentState(
...
+    // (specially finish time) once.
+    if (!currentState.equals(jobCtx.getPartitionState(pId))) {
```
narendly, line 414: This condition checks whether jobContext's state is out of date or not, right?

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -380,72 +379,38 @@ private static boolean isJobComplete(JobContext ctx, Set<Integer> allPartitions,
...
-
-    // Add all pIds existing in CurrentStateOutput as well because task currentStates copied over
-    // from previous sessions won't show up in prevInstanceToTaskAssignments
-    // We need to add these back here in order for these task partitions to be dropped (after a
-    // copy-over, the Controller will send a message to drop the state currentState)
-    // partitions: (partition -> instance -> currentState)
+    // Generate currentInstanceToTaskAssignment with CurrentStateOutput as source of truth
+    // Add all pIds existing in CurrentStateOutput
+    // We need to add these pIds here and make decision about the existing tasks with current state
+    // in updatePreviousAssignedTasksStatus method
```
narendly, line 399: "add these pIds to `result`" and update their states in JobContext in `updatePreviousAssignedTasksStatus` method...

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -380,72 +379,38 @@ private static boolean isJobComplete(JobContext ctx, Set<Integer> allPartitions,
...
-          // ** for tasks that are just in INIT state, we do not add them here because old
-          // Participants, upon connection reset, set tasks' currentStates to INIT. We cannot
-          // consider those tasks active **
-          if (currState == TaskPartitionState.INIT && requestedState != null
-              && requestedState.equals(TaskPartitionState.RUNNING.name())
-              || currState == TaskPartitionState.RUNNING) {
-            result.get(instance).add(pId);
-          }
-
+          // We must add all pIds back here
```
narendly, line 410: This comment isn't giving any information. Why must we?

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -380,72 +379,38 @@ private static boolean isJobComplete(JobContext ctx, Set<Integer> allPartitions,
...
+          result.get(instance).add(pId);
           // Check if this task needs to be dropped. If so, we need to add to tasksToDrop no matter
-          // what its current state is so that it will be dropped
+          // what its current state is so that it will be dropped.
```
narendly, line 413: Nit: This dot isn't necessary...
If anything, repeat again what "a task needs to be dropped" means. This is trying to drop tasks on a reconnected instance with a new sessionId that have all of their requestedState == DROPPED.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -462,10 +427,10 @@ private static boolean isJobComplete(JobContext ctx, Set<Integer> allPartitions,
    * If partition is missing from prevInstanceToTaskAssignments (e.g. previous assignment is
    * deleted) it is added from context. Otherwise, the context won't be updated.
    * @param jobCtx Job Context
-   * @param prevInstanceToTaskAssignments instance -> partitionIds from previous assignment
+   * @param currentInstanceToTaskAssignments instance -> partitionIds from current assignment
```
narendly, line 430: "from CurrentStateOutput"

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -242,16 +230,16 @@ public void updatePreviousAssignedTasksStatus(
         }
           break;
         case COMPLETED: {
-          // The task has completed on this partition. Mark as such in the context object.
-          donePartitions.add(pId);
+          // The task has completed on this partition. Drop it from the instance and add it to assignedPartitions in
+          // order to avoid scheduling it again in this pipeline.
```
dasahcc, line 234: Why you remove the done partition from COMPLETED state? Then this partition will not be removed from pSet, right?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: Yes. Done.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: The reason is the behavior has been changed in this PR. We have an assignment for it and the assignment is dropped. So, it goes to assigned partitions. Once it is dropped from participant, then we do not even consider it for the new assignment.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: You are right. I also had this concern. But note that we do not change the JobContext to DROPPED. Let’s say controller sends COMPLETED to DROPPED message to the participant. Now, the current state will be COMPLETED until participant processes the message. Once it is processed, the current state will be null, and that task will not be considered again. So, to answer your question, it is totally depending on the behavior of the participant. If participant set the current state  to DROPPED, which is not the case here, your concern was correct. Since participant removes the current state of the task, we do not need to be worry about that case.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: The answer to this question is also similar to COMPLETED case :). Please see it above.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: Fixed. Thanks.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: The race condition that here is scenario that I am talking about is the scenario where we have two current states of a task in different participants. Since we have only one slot for each task in paMap, we want to make sure that the task is DROPPED from the wrong one and is assigned to the correct instance. I am basically talking about this #922 and #461. Let me rephrase this comment to make it clear.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: Exactly. We add this check in PR #923 to help resolve this scenario. Maybe using race condition is not the good word here. I rephrased the comments.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: Exactly. I added a comment about it.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
alirezazamani: Fixed. Thanks.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
alirezazamani: I agree. Removed the comment. Basically, we have explained everything in the comments already.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: Here we check the if current state has been changed and it is not equal to context information. This check is necessary because we do not need to update the context in every pipeline. Also, the more important part of this check is for the fields such as finish time. Let’s say we have a delay retry time for the task. This functionality depends on the finish time in the context. If we update finish time as long as we have current state (note that we set it to current time), we will never schedule the task with delay retry time. Because finish time in the context and delay retry time will be messed up. That is why we only update the context at the first time that the current state has been changed.

alirezazamani:
> I have a general suggestion. Let's do the renaming in another PR. Then we can have clear understand what is the logical change.

There are not that much name changes. It is basically prevAssignment to currentState. The rest are similar. If you still think the name change is necessary, I can change the names back to original. Please let me know.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -238,21 +238,20 @@ private ResourceAssignment computeResourceMapping(String jobResource,
...
 
     if (LOG.isDebugEnabled()) {
       LOG.debug("All partitions: " + allPartitions + " taskAssignment: "
-          + prevInstanceToTaskAssignments + " excludedInstances: " + excludedInstances);
+          + currentInstanceToTaskAssignments + " excludedInstances: " + excludedInstances);
     }
 
     // Release resource for tasks in terminal state
-    updatePreviousAssignedTasksStatus(prevInstanceToTaskAssignments, excludedInstances, jobResource,
+    updatePreviousAssignedTasksStatus(currentInstanceToTaskAssignments, excludedInstances, jobResource,
```
NealSun96, line 254: Should this function be renamed?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -121,17 +120,6 @@ public void updatePreviousAssignedTasksStatus(
               instance, pId);
           continue;
         }
```
NealSun96, line 122: Would it be a problem that this skipping check is now done after markPartitionCompleted/markPartitionError?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -389,13 +381,59 @@ private TaskPartitionState updateJobContextAndGetTaskCurrentState(
...
+    // This check is necessary because we are relying on current state and we do not want to update
+    // context as long as current state existed. We just want to update context information
+    // (specially finish time) once.
+    // This condition checks whether jobContext's state is out of date or not.
+    if (!currentState.equals(jobCtx.getPartitionState(pId))) {
+      jobCtx.setPartitionState(pId, currentState);
+      String taskMsg = currentStateOutput.getInfo(jobResource, new Partition(pName), instance);
+      if (taskMsg != null) {
+        jobCtx.setPartitionInfo(pId, taskMsg);
+      }
```
NealSun96, line 421: Is there a possibility that a message exists without state changes?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
dasahcc: The concern for doing this is that will this reflect the change to JobContext? So if final state will all be DROPPED. It is hard for us to determine whether the tasks is completed or it requires another retry? Because if task has problem in the middle of running, it will be DROPPED as well.

dasahcc: Also, I dont think you will get DROPPED state as current state, right? Because the current state will be deleted from the current state if the to state is DROPPED after it completes state transition.

alirezazamani: Exactly. Since we do not get DROPPED state in current state, we will not mark the the task as DROPPED in job context. If the task is completed and we send COMPLETE to DROPPED message, then the participant will DROP the currentstate and this function will not see the task on the instance again because we are based on CurrentState now. Hence, we will not mark the task DROPPED in jobContext. Basically since in this function context follows the currentState, then DROPPED context will not be possible. Right?

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
alirezazamani: I think it should be fine to keep it as it is. Because we are updating the state of the tasks that has been assigned in previous pipelines based on the currentstate of the tasks.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: This shouldn't be an issue because this check is specifically added before to bypass the scenario where we have two current states which non of the markPartitionCompleted/markPartitionError will not be hit because it is either Running or Dropped. Good question though.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: The participant read the message, handle state change and then delete the message. So when messages get deleted from ZK after ST is completed. So we should be safe here.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
narendly: @alirezazamani -

Okay, I think your point about not updating when there's no delta is valid, but I'm not sure if what you said about delayed scheduling is entirely true. Delayed scheduling of tasks is a feature that was previously working, right? If what you said about it was the case, then how come our delayed scheduling feature was working all this time?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -389,13 +381,59 @@ private TaskPartitionState updateJobContextAndGetTaskCurrentState(
...
-    jobCtx.setPartitionState(pId, currentState);
-    String taskMsg = currentStateOutput.getInfo(jobResource, new Partition(pName), instance);
-    if (taskMsg != null) {
-      jobCtx.setPartitionInfo(pId, taskMsg);
+    // If job context needs to be updated with new state, update it accordingly
+    // This check is necessary because we are relying on current state and we do not want to update
+    // context as long as current state existed. We just want to update context information
+    // (specially finish time) once.
+    // This condition checks whether jobContext's state is out of date or not.
+    if (!currentState.equals(jobCtx.getPartitionState(pId))) {
```
narendly, line 416: @alirezazamani Another concern here:

Is it possible for a task to be retried so fast that it ends up being in the same state? For example,

task_error -> (controller reschedules it) -> (controller sends messages, error -> init, init -> running, running -> complete) -> participant processes message so quickly but it goes into task_error again, and by the time controller gets to this task, the states are the same as the previous run. Basically, controller sees task_error ->  task_error.

Is that going to be an issue? I guess partition id is set before this check so that will be updated accordingly, but what about finish time or the time at which the task was marked as error?

alirezazamani: @narendly Thank you for mentioning these corner cases. 

I went through the logic and I am sure the code handles this corner case as well. Let me explain why (please note that if task goes to terminal states, with this PR we are dropping 
 the task first and schedule them in next pipeline):

Here are the steps:
1- Requested State is set to TASK_ERROR by the participant. 
2- Controller sends RUNNING -> TASK_ERROR to participant.
3- Participant process it and makes current state as TASK_ERROR.
4- Controller see this (i.e. TASK_ERROR) current state and send TASK_ERROR -> DROPPED message.
5- Participant gets the message and drops/removes current states.
6- Controller does not see current state (because it is removed) and sets the context of the task to be INIT.  Schedule the task again on the participant by sending INIT-> RUNNING message. (this happens in handleAdditionalTaskAssignment method).
*Now controller by itself does not send RUNNING -> COMPLETE because COMPLETE state should be requested by the participant. The participant will either request TASK_ERROR or COMPLETE state in current state.
7- If the task goes to error state we see this delta again (because context is set to be INIT before scheduling the task) and mark it ERROR with new finish time. 

There are two points here that helps controller to handle this cornet case:
1- controller will not send running -> complete unless participant requested state is COMPLETE. So if the task goes to TASK_ERROR in participant, the requested state will be TASK_ERROR and controller always respects the requested state and send RUNNING -> TASK_ERROR. 
2- Dropping the task in terminal state makes the controller to reschedule the task with jobContext as INIT. This behavior helps the controller to see these deltas in future pipelines and update the information correctly in jobContext.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: @narendly 

The reason that delayed scheduling was working is because we were relying on previousAssignment. Let's say we are relying on previousAssignment and we decided to send the task to TASK_ERROR. This task will be recorded in previousAssignment and next pipeline we iterate through the task existed in previous assignment and record finish time and context information. Then this task will not be existed in future pipelines previousAssignment and we do not update finish time in the future pipelines. (Actually some of the test guided me toward this change).

However, if we want to be independent of previousAssignment and rely on currentState (which is also means we are relying on participant's reactions), we need to consider that currentState might not change for several pipelines. In this case, it is necessary to monitor these deltas and not update the context information (specially time sensitive information) multiple times.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
narendly: > 1- controller will not send running -> complete unless participant requested state is COMPLETE

Are you sure this is true? Can you verify that init -> running / running -> complete are not sent out at the same time? The target state during message generation phase should be complete if I recall correctly.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
narendly: Thanks for the explanation. The reason we need to know if CurrentState changed is because we no longer have prevAssignment.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: I double checked again and it seems the only way that controller sends COMPLETED message to participant should be initiated by participant by setting the requested state to be completed.

narendly: Sounds good. Overall I think this analysis is sound. Let's make sure we do a careful round of  E2E testing as well as load testing.

alirezazamani:
mvn test result before merge:
Helix-Core
```
[INFO] Tests run: 1144, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,939.921 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:22 h
[INFO] Finished at: 2020-05-26T10:42:39-07:00
```
Helix-rest:
```
[INFO] Tests run: 159, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 46.025 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  51.874 s
[INFO] Finished at: 2020-05-26T16:31:12-07:00
```

alirezazamani:

Final commit message:

Make the task scheduling decision independent of the PreviousAssignment

In this commit, the previous logic which was based on PreviousAssignment, has been changed. Instead, the task scheduling is done based on the CurrentState.

Issue #996, opened 2020-05-05
Allow periodic message full refresh for participants to prevent missing messages


mgao0:
With our currently logic, if participant somehow fails to read messages from ZooKeepr, it could retry the read. However, if there is a long period that ZooKeeper sever is not available, such as a long GC, retry reading doesn't help. With no other trigger for the participant to try to read Zk again, we could potentially miss some messages that are leftover in the Zk server. Therefore, we add a configurable periodic message full refresh to prevent this situation.

PR #1000, opened 2020-05-11
Add message periodic refresh

mgao0:
### Issues

#996

### Description

This PR adds a scheduled periodic task to read leftover messages from zk.
There are several situations that we would need this periodic refresh:
1. A read from ZK fails and there is no more event to trigger another read
2. Watcher is not registered properly so it fails to trigger the callback
3. The message is not processed correctly and is discarded but no more incoming events to trigger another read

What would not be fixed by this PR is out of order problem, which is another issue may cause missing messages. For example, we may miss an event because its callback comes in after a finalize type event, so it will be ignored. Since this periodic refresh is done through generating an event, this out of order issue may affect refresh events too.

The refresh is done in callback handler and will be triggered if the time has passed for as long as a user defined interval since last processed event. The refresh is achieved by queuing an empty callback type event into the event processor.

### Tests

TestPerioidicRefresh

Running

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -761,8 +761,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
       }
     }
 
-    resetTriggerTask();
-
+    initTriggerTask();
     updateNotificationTime(System.nanoTime());
     try {
       NotificationContext changeContext = new NotificationContext(_manager);
@@ -884,23 +883,12 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
             _batchCallbackProcessor.resetEventQueue();
           }
         }
-
-        NotificationContext changeContext = new NotificationContext(_manager);
-        changeContext.setType(NotificationContext.Type.FINALIZE);
-        changeContext.setChangeType(_changeType);
-        invoke(changeContext);
-
-        if (_periodicTriggerExecutor != null) {
-          if (isShutdown) {
-            _periodicTriggerExecutor.shutdownNow();
-            if (!_scheduledTriggerFuture.isCancelled()) {
-              _scheduledTriggerFuture.cancel(true);
-            }
-          } else {
-            resetTriggerTask();
-          }
-        }
+        shutDownTriggerTask(isShutdown);
       }
+      NotificationContext changeContext = new NotificationContext(_manager);
+      changeContext.setType(NotificationContext.Type.FINALIZE);
+      changeContext.setChangeType(_changeType);
+      invoke(changeContext);
     } catch (Exception e) {
       String msg = "Exception while resetting the listener:" + _listener;
       ZKExceptionHandler.getInstance().handle(msg, e);
@@ -931,22 +919,33 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
   }
 
   /**
-   * Used to initialize or reset a periodic refresh task
+   * Used to shut down a periodic triggered task
+   */
+  private void shutDownTriggerTask(boolean isShutdown) {
+    if (_scheduledTriggerFuture != null) {
+      _scheduledTriggerFuture.cancel(true);
+    }
+    if (_periodicTriggerExecutor != null && isShutdown) {
+      _periodicTriggerExecutor.shutdownNow();
+      _periodicTriggerExecutor = null;
+    }
+  }
+
+  /**
+   * Used to initialize a periodic triggered task
    * Schedule tasks in a task executor with fixed intervals
    */
-  private void resetTriggerTask() {
+  private void initTriggerTask() {
     if (_periodicTriggerInterval <= 0) {
       return;
     }
+    shutDownTriggerTask(false);
     _lastInvokeTime = System.currentTimeMillis();
     if (_periodicTriggerExecutor == null) {
       _periodicTriggerExecutor = new ScheduledThreadPoolExecutor(1);
       // When cancelling the task future, it removes the task from the queue
       // so we won't have a memory leakage when we cancel scheduled task
       _periodicTriggerExecutor.setRemoveOnCancelPolicy(true);
-    }
-    if (_scheduledTriggerFuture != null) {
-      _scheduledTriggerFuture.cancel(true);
     }
     _scheduledTriggerFuture = _periodicTriggerExecutor
         .scheduleWithFixedDelay(new TriggerTask(), _periodicTriggerInterval,
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -28,6 +28,10 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.helix.BaseDataAccessor;
@@ -111,9 +115,12 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
    * define the next possible notification types
    */
   private static Map<Type, List<Type>> nextNotificationType = new HashMap<>();
+
   static {
-    nextNotificationType.put(Type.INIT, Arrays.asList(Type.CALLBACK, Type.FINALIZE));
-    nextNotificationType.put(Type.CALLBACK, Arrays.asList(Type.CALLBACK, Type.FINALIZE));
+    nextNotificationType
+        .put(Type.INIT, Arrays.asList(Type.CALLBACK, Type.FINALIZE, Type.PERIODIC_REFRESH));
+    nextNotificationType
+        .put(Type.CALLBACK, Arrays.asList(Type.CALLBACK, Type.FINALIZE, Type.PERIODIC_REFRESH));
     nextNotificationType.put(Type.FINALIZE, Arrays.asList(Type.INIT));
   }
 
@@ -132,6 +139,11 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
   private boolean _batchModeEnabled = false;
   private boolean _preFetchEnabled = true;
   private HelixCallbackMonitor _monitor;
+  private final long _periodicRefreshInterval;
+  private ScheduledExecutorService _periodicRefreshExecutor;
+  private volatile long _lastEventTime;
+  private ScheduledFuture<?> _scheduledRefreshFuture;
+  private static final long VALUE_NOT_SET = -1;
 
   // TODO: make this be per _manager or per _listener instaed of per callbackHandler -- Lei
   private CallbackProcessor _batchCallbackProcessor;
@@ -203,6 +215,29 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
     }
   }
 
+  class RefreshTask implements Runnable {
+    @Override
+    public void run() {
+      try {
+        long currentTime = System.currentTimeMillis();
+        if (_lastEventTime + _periodicRefreshInterval <= currentTime) {
+          NotificationContext changeContext = new NotificationContext(_manager);
+          changeContext.setType(NotificationContext.Type.PERIODIC_REFRESH);
+          changeContext.setChangeType(_changeType);
+          enqueueTask(changeContext);
+        } else {
+          long remainingTime = _lastEventTime + _periodicRefreshInterval - currentTime;
+          _scheduledRefreshFuture.cancel(false);
+          _scheduledRefreshFuture = _periodicRefreshExecutor
+              .scheduleWithFixedDelay(this, remainingTime, _periodicRefreshInterval,
+                  TimeUnit.MILLISECONDS);
+        }
+      } catch (Exception e) {
+        logger.warn("Exception caught in periodic refresh task.", e);
+      }
+    }
+  }
+
   /**
    * maintain the expected notification types
    * this is fix for HELIX-195: race condition between FINALIZE callbacks and Zk callbacks
@@ -211,12 +246,25 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
 
   public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
       Object listener, EventType[] eventTypes, ChangeType changeType) {
-    this(manager, client, propertyKey, listener, eventTypes, changeType, null);
+    this(manager, client, propertyKey, listener, eventTypes, changeType, null, -1);
+  }
+
+  public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
+      Object listener, EventType[] eventTypes, ChangeType changeType,
+      long periodicRefreshInterval) {
+    this(manager, client, propertyKey, listener, eventTypes, changeType, null,
+        periodicRefreshInterval);
   }
 
   public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
       Object listener, EventType[] eventTypes, ChangeType changeType,
       HelixCallbackMonitor monitor) {
+    this(manager, client, propertyKey, listener, eventTypes, changeType, monitor, -1);
+  }
+
+  public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
+      Object listener, EventType[] eventTypes, ChangeType changeType, HelixCallbackMonitor monitor,
+      long periodicRefreshInterval) {
     if (listener == null) {
       throw new HelixException("listener could not be null");
     }
@@ -246,6 +294,17 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
     parseListenerProperties();
 
     init();
+
+    /**
+     * Async batch read was designed that if an exception is thrown, the operation fails and returns a partial result.
+     * The issue is separately addressed by checking read result in batch read calls [https://github.com/apache/helix/pull/974]
+     * and retry batch read for ZK connectivity issues [https://github.com/apache/helix/pull/970].
+     * We also add this one to periodical read stale messages from ZkServer in the case we don't see any event for a certain period.
+     */
+    _periodicRefreshInterval = periodicRefreshInterval;
+    if (_periodicRefreshInterval != VALUE_NOT_SET) {
+      initRefreshTask();
+    }
   }
 
   private void parseListenerProperties() {
@@ -394,16 +453,17 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
         logger.warn("Callback handler received event in wrong order. Listener: {}, path: {}, "
             + "expected types: {}, but was {}", _listener, _path, _expectTypes, type);
         return;
-
       }
       _expectTypes = nextNotificationType.get(type);
 
-      if (type == Type.INIT || type == Type.FINALIZE) {
+      if (type == Type.INIT || type == Type.FINALIZE || type == Type.PERIODIC_REFRESH) {
         subscribeForChanges(changeContext.getType(), _path, _watchChild);
       } else {
         // put SubscribeForChange run in async thread to reduce the latency of zk callback handling.
         subscribeForChangesAsyn(changeContext.getType(), _path, _watchChild);
       }
+
+      _lastEventTime = System.currentTimeMillis();
 
       if (_changeType == IDEAL_STATE) {
         IdealStateChangeListener idealStateChangeListener = (IdealStateChangeListener) _listener;
@@ -814,8 +874,15 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
           if (isShutdown) {
             _batchCallbackProcessor.shutdown();
             _batchCallbackProcessor = null;
+            if (_periodicRefreshExecutor != null) {
+              _periodicRefreshExecutor.shutdownNow();
+            }
           } else {
             _batchCallbackProcessor.resetEventQueue();
+            if (_periodicRefreshExecutor != null) {
+              _periodicRefreshExecutor.shutdownNow();
+              initRefreshTask();
+            }
           }
         }
       }
@@ -850,5 +917,13 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
         + _preFetchEnabled + ", _batchModeEnabled=" + _batchModeEnabled + ", _path='" + _path + '\''
         + ", _listener=" + _listener + ", _changeType=" + _changeType + ", _manager=" + _manager
         + ", _zkClient=" + _zkClient + '}';
+  }
+
+  private void initRefreshTask() {
+    _lastEventTime = System.currentTimeMillis();
+    _periodicRefreshExecutor = Executors.newSingleThreadScheduledExecutor();
+    _scheduledRefreshFuture = _periodicRefreshExecutor
+        .scheduleWithFixedDelay(new RefreshTask(), _periodicRefreshInterval,
+            _periodicRefreshInterval, TimeUnit.MILLISECONDS);
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -114,13 +114,12 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
    * define the next possible notification types
    */
   private static Map<Type, List<Type>> nextNotificationType = new HashMap<>();
-
   static {
     nextNotificationType
-        .put(Type.INIT, Arrays.asList(Type.CALLBACK, Type.FINALIZE, Type.PERIODIC_REFRESH));
+        .put(Type.INIT, Arrays.asList(Type.CALLBACK, Type.FINALIZE));
     nextNotificationType
-        .put(Type.CALLBACK, Arrays.asList(Type.CALLBACK, Type.FINALIZE, Type.PERIODIC_REFRESH));
-    nextNotificationType.put(Type.FINALIZE, Arrays.asList(Type.INIT, Type.PERIODIC_REFRESH));
+        .put(Type.CALLBACK, Arrays.asList(Type.CALLBACK, Type.FINALIZE));
+    nextNotificationType.put(Type.FINALIZE, Arrays.asList(Type.INIT));
   }
 
   // processor to handle async zk event resubscription.
@@ -138,7 +137,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
   private boolean _batchModeEnabled = false;
   private boolean _preFetchEnabled = true;
   private HelixCallbackMonitor _monitor;
-  private final long _periodicRefreshInterval;
+  private final long _periodicRefreshTriggerInterval;
   private ScheduledThreadPoolExecutor _periodicRefreshExecutor;
   private ScheduledFuture<?> _scheduledRefreshFuture;
   private volatile long _lastEventTime;
@@ -217,22 +216,21 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
     @Override
     public void run() {
       try {
-        long currentTime = System.currentTimeMillis();
-        long remainingTime = _lastEventTime + _periodicRefreshInterval - currentTime;
-        if (remainingTime <= 0) {
-          NotificationContext changeContext = new NotificationContext(_manager);
-          changeContext.setType(NotificationContext.Type.PERIODIC_REFRESH);
-          changeContext.setChangeType(_changeType);
-          enqueueTask(changeContext);
-        } else {
-          _scheduledRefreshFuture.cancel(false);
-          // remaining time must > 0, otherwise exception will be thrown and the schedule fails
-          _periodicRefreshExecutor
-              .scheduleWithFixedDelay(this, remainingTime, _periodicRefreshInterval,
-                  TimeUnit.MILLISECONDS);
+        while (true) {
+          long currentTime = System.currentTimeMillis();
+          long remainingTime = _lastEventTime + _periodicRefreshTriggerInterval - currentTime;
+          if (remainingTime <= 0) {
+            NotificationContext changeContext = new NotificationContext(_manager);
+            changeContext.setType(Type.CALLBACK);
+            changeContext.setChangeType(_changeType);
+            enqueueTask(changeContext);
+            break;
+          } else {
+            Thread.sleep(remainingTime);
+          }
         }
       } catch (Exception e) {
-        logger.warn("Exception caught in periodic refresh task.", e);
+        logger.warn("Exception during execute refresh task. Error: ", e);
       }
     }
   }
@@ -245,25 +243,25 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
 
   public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
       Object listener, EventType[] eventTypes, ChangeType changeType) {
-    this(manager, client, propertyKey, listener, eventTypes, changeType, null, SystemPropertyKeys.PROPERTY_NOT_SET);
+    this(manager, client, propertyKey, listener, eventTypes, changeType, null, -1);
   }
 
   public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
       Object listener, EventType[] eventTypes, ChangeType changeType,
-      long periodicRefreshInterval) {
+      long periodicRefreshTriggerInterval) {
     this(manager, client, propertyKey, listener, eventTypes, changeType, null,
-        periodicRefreshInterval);
+        periodicRefreshTriggerInterval);
   }
 
   public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
       Object listener, EventType[] eventTypes, ChangeType changeType,
       HelixCallbackMonitor monitor) {
-    this(manager, client, propertyKey, listener, eventTypes, changeType, monitor, SystemPropertyKeys.PROPERTY_NOT_SET);
+    this(manager, client, propertyKey, listener, eventTypes, changeType, monitor, -1);
   }
 
   public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
       Object listener, EventType[] eventTypes, ChangeType changeType, HelixCallbackMonitor monitor,
-      long periodicRefreshInterval) {
+      long periodicRefreshTriggerInterval) {
     if (listener == null) {
       throw new HelixException("listener could not be null");
     }
@@ -300,8 +298,8 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
      * and retry batch read for ZK connectivity issues [https://github.com/apache/helix/pull/970].
      * We also add this one to periodical read stale messages from ZkServer in the case we don't see any event for a certain period.
      */
-    _periodicRefreshInterval = periodicRefreshInterval;
-    if (_periodicRefreshInterval != SystemPropertyKeys.PROPERTY_NOT_SET) {
+    _periodicRefreshTriggerInterval = periodicRefreshTriggerInterval;
+    if (_periodicRefreshTriggerInterval > 0) {
       initRefreshTask();
     }
   }
@@ -452,12 +450,11 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
         logger.warn("Callback handler received event in wrong order. Listener: {}, path: {}, "
             + "expected types: {}, but was {}", _listener, _path, _expectTypes, type);
         return;
-      }
-      if (type != Type.PERIODIC_REFRESH) {
-        _expectTypes = nextNotificationType.get(type);
-      }
 
-      if (type == Type.INIT || type == Type.FINALIZE || type == Type.PERIODIC_REFRESH) {
+      }
+      _expectTypes = nextNotificationType.get(type);
+
+      if (type == Type.INIT || type == Type.FINALIZE) {
         subscribeForChanges(changeContext.getType(), _path, _watchChild);
       } else {
         // put SubscribeForChange run in async thread to reduce the latency of zk callback handling.
@@ -780,15 +777,15 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
     try {
       updateNotificationTime(System.nanoTime());
       if (dataPath != null && dataPath.startsWith(_path)) {
-        NotificationContext changeContext = new NotificationContext(_manager);
-        changeContext.setType(NotificationContext.Type.CALLBACK);
-        changeContext.setPathChanged(dataPath);
-        changeContext.setChangeType(_changeType);
-        enqueueTask(changeContext);
-      }
-    } catch (Exception e) {
-      String msg =
-          "exception in handling data-change. path: " + dataPath + ", listener: " + _listener;
+      NotificationContext changeContext = new NotificationContext(_manager);
+      changeContext.setType(NotificationContext.Type.CALLBACK);
+      changeContext.setPathChanged(dataPath);
+      changeContext.setChangeType(_changeType);
+      enqueueTask(changeContext);
+    }
+  } catch (Exception e) {
+    String msg =
+        "exception in handling data-change. path: " + dataPath + ", listener: " + _listener;
       ZKExceptionHandler.getInstance().handle(msg, e);
     }
   }
@@ -926,7 +923,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
     // so we won't have a memory leakage when we cancel scheduled task
     _periodicRefreshExecutor.setRemoveOnCancelPolicy(true);
     _scheduledRefreshFuture = _periodicRefreshExecutor
-        .scheduleWithFixedDelay(new RefreshTask(), _periodicRefreshInterval,
-            _periodicRefreshInterval, TimeUnit.MILLISECONDS);
+        .scheduleWithFixedDelay(new RefreshTask(), _periodicRefreshTriggerInterval,
+            _periodicRefreshTriggerInterval, TimeUnit.MILLISECONDS);
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -120,7 +120,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
         .put(Type.INIT, Arrays.asList(Type.CALLBACK, Type.FINALIZE, Type.PERIODIC_REFRESH));
     nextNotificationType
         .put(Type.CALLBACK, Arrays.asList(Type.CALLBACK, Type.FINALIZE, Type.PERIODIC_REFRESH));
-    nextNotificationType.put(Type.FINALIZE, Arrays.asList(Type.INIT));
+    nextNotificationType.put(Type.FINALIZE, Arrays.asList(Type.INIT, Type.PERIODIC_REFRESH));
   }
 
   // processor to handle async zk event resubscription.
@@ -142,7 +142,6 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
   private ScheduledThreadPoolExecutor _periodicRefreshExecutor;
   private ScheduledFuture<?> _scheduledRefreshFuture;
   private volatile long _lastEventTime;
-  private static final long VALUE_NOT_SET = -1;
 
   // TODO: make this be per _manager or per _listener instaed of per callbackHandler -- Lei
   private CallbackProcessor _batchCallbackProcessor;
@@ -246,7 +245,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
 
   public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
       Object listener, EventType[] eventTypes, ChangeType changeType) {
-    this(manager, client, propertyKey, listener, eventTypes, changeType, null, VALUE_NOT_SET);
+    this(manager, client, propertyKey, listener, eventTypes, changeType, null, SystemPropertyKeys.PROPERTY_NOT_SET);
   }
 
   public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/messaging/DefaultMessagingService.java
```diff
@@ -70,7 +70,7 @@ public DefaultMessagingService(HelixManager manager) {
 
     _taskExecutor = new HelixTaskExecutor(
         new ParticipantStatusMonitor(isParticipant, manager.getInstanceName()),
-        new MessageQueueMonitor(manager.getClusterName(), manager.getInstanceName()));
+        new MessageQueueMonitor(manager.getClusterName(), manager.getInstanceName()), manager);
```
mgao0, line 73: jiajunwang 14 hours ago Contributor
Did you miss some files here? MessageQueueMonitor does not have such a constructor and it is not changed in this PR.

 @jiajunwang
jiajunwang 14 hours ago Contributor
In general, passing the manager to the monitor object would be an overkill. If it is really necessary, then we shall change the constructor to accept the manager only, since it includes all the information.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -152,6 +155,11 @@ public HelixTaskExecutor(ParticipantStatusMonitor participantStatusMonitor) {
 
   public HelixTaskExecutor(ParticipantStatusMonitor participantStatusMonitor,
       MessageQueueMonitor messageQueueMonitor) {
+    this(participantStatusMonitor, messageQueueMonitor, null);
```
mgao0, line 158: @jiajunwang 14 hours ago Contributor
Deprecate it?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -714,7 +724,7 @@ private void syncSessionToController(HelixManager manager) {
 
   @Override
   @PreFetch(enabled = false)
-  public void onMessage(String instanceName, List<Message> messages,
+  public synchronized void onMessage(String instanceName, List<Message> messages,
```
mgao0, line 727: @jiajunwang 14 hours ago Contributor
Why? The periodic refresh does not need to be done synchronized, I think.
If your timer action just adds the refresh event to the queue, there is no need to modify this.

mgao0:
@jiajunwang 14 hours ago Contributor
Why do you need to call it here synchronously? I think it's simpler if we just put this event into the event queue.
Given this saying, this logic will be fitting better if it is in the HelixManager level. What I'm trying to propose if the HelixManger is initialized as a Participant, then we start this timer thread there. It just injects message event with a certain interval. Of course, if a real message event comes earlier, then we reset the timer.

Put this logic in the executor does not make sense to me.

mgao0:
@jiajunwang  Sorry if it's confusing. This PR is for the periodic refresh. I have copied and pasted all your comments here and I will address them here.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
kaisun2000: Previously only CallbacHandler (for message) would execute this onMessage() and it would be on ZkClient's event thread. 
Now you have thread to execute this method. Make it would synchronized would be safe.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -1135,6 +1145,34 @@ private void removeMessageFromZK(HelixDataAccessor accessor, Message message,
     }
   }
 
+  private void startMessageRefreshThread() {
```
kaisun2000, line 1148: How about let us add a comment about the reason why we add this thread?
Something like:
Async batch read has bug that prevent it from reading all the messages successfully. The issue is separately addressed by pull request [xxxx]. We also add this one as periodical "pull" messages from ZkServer method to preemptively read and process messages from controller.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -1135,6 +1145,34 @@ private void removeMessageFromZK(HelixDataAccessor accessor, Message message,
...
+    int periodicMessageRefreshInterval =
+        HelixUtil.getSystemPropertyAsInt(SystemPropertyKeys.PERIODIC_MESSAGE_REFRESH_INTERVAL, -1);
+    if (periodicMessageRefreshInterval > 0) {
+      _messageRefreshTimer = new Timer();
+      try {
+        {
+          if (_manager == null) {
```
pkuwm, line 1155: Do you think it may be cleaner if we check this null in constructor instead and not start this thread if null, which avoids this throw HelixException, catch and log?

helix-core/src/test/java/org/apache/helix/messaging/handling/TestPerioidicMessageRefresh.java
```diff
@@ -0,0 +1,141 @@
+package org.apache.helix.messaging.handling;
+
```
pkuwm, line 2: Don't forget Apache license :)

helix-core/src/test/java/org/apache/helix/messaging/handling/TestPerioidicMessageRefresh.java
```diff
@@ -0,0 +1,141 @@
...
+public class TestPerioidicMessageRefresh extends ZkStandAloneCMTestBase {
+  private HelixManager manager;
+  private HelixDataAccessor accessor;
+  private PropertyKey.Builder keyBuilder;
+  private HelixTaskExecutor executor;
+  private TestMessageHandlerFactory factory;
+
+  public class TestMessageHandlerFactory implements MultiTypeMessageHandlerFactory {
+    int _handlersCreated = 0;
+    ConcurrentHashMap<String, String> _processedMsgIds = new ConcurrentHashMap<>();
```
pkuwm, line 32: Use interface: `Map<String, String> _processedMsgIds`?

helix-core/src/test/java/org/apache/helix/messaging/handling/TestPerioidicMessageRefresh.java
```diff
@@ -0,0 +1,141 @@
...
+      public TestMessageHandler(Message message, NotificationContext context) {
+        super(message, context);
+        // TODO Auto-generated constructor stub
+      }
+
+      @Override
+      public HelixTaskResult handleMessage() throws InterruptedException {
+        HelixTaskResult result = new HelixTaskResult();
+        _processedMsgIds.put(_message.getMsgId(), _message.getMsgId());
+        Thread.sleep(100);
```
pkuwm, line 44: What's the purpose of sleep?

helix-core/src/test/java/org/apache/helix/messaging/handling/TestPerioidicMessageRefresh.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.helix.model.Message;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+public class TestPerioidicMessageRefresh extends ZkStandAloneCMTestBase {
+  private HelixManager manager;
+  private HelixDataAccessor accessor;
+  private PropertyKey.Builder keyBuilder;
+  private HelixTaskExecutor executor;
+  private TestMessageHandlerFactory factory;
```
pkuwm, line 28: Underscore as prefix to be consistent?

helix-core/src/test/java/org/apache/helix/messaging/handling/TestPerioidicMessageRefresh.java
```diff
@@ -0,0 +1,141 @@
...
+      public boolean verify() throws Exception {
+        return factory._processedMsgIds.size() == 0;
+      }
+    }, 2000);
+
+    // Start periodic message refresh
+    Properties properties = new Properties();
+    properties
+        .setProperty(SystemPropertyKeys.PERIODIC_MESSAGE_REFRESH_INTERVAL, String.valueOf(20));
+    System.setProperties(properties);
```
pkuwm, line 126: I suggest using  `setProperty()` as just one single property is set.
It is a good habit to clear the system property after this test so it eliminates possibility of affecting other tests after this one.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -1135,6 +1145,34 @@ private void removeMessageFromZK(HelixDataAccessor accessor, Message message,
     }
   }
 
+  private void startMessageRefreshThread() {
+    int periodicMessageRefreshInterval =
+        HelixUtil.getSystemPropertyAsInt(SystemPropertyKeys.PERIODIC_MESSAGE_REFRESH_INTERVAL, -1);
```
kaisun2000, line 1150: Another point is that:
 OnMessage called by the timer thread can throw exception. The thing is that I suspect that timer's internal thread may not schedule more OnMessage() if the previous one has exception thrown. This needs to be tested and verified. Otherwise, the timer thread may never work after exception thrown if my suspicion is right.

Java Concurrency Control book talked in detail about this issue. I suggest let at least let test throwing exception from OnMessage and see if this is a problem.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
narendly: If we make sure periodic refresh timer is reset whenever there's been recent events, this sort of synchronization won't be necessary by design. See my more detailed comment below.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -781,8 +841,15 @@ void reset(boolean isShutdown) {
           if (isShutdown) {
             _batchCallbackProcessor.shutdown();
             _batchCallbackProcessor = null;
+            if (_periodicRefreshExecutor != null) {
+              _periodicRefreshExecutor.shutdownNow();
+            }
           } else {
             _batchCallbackProcessor.resetEventQueue();
+            if (_periodicRefreshExecutor != null) {
+              _periodicRefreshExecutor.shutdownNow();
```
NealSun96, line 850: Just making sure: this line automatically cancels the scheduled runnable in `_scheduledRefreshFuture` right? If the refresh period is right after this line and before the next line, it's okay?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -210,12 +245,25 @@ protected void handleEvent(NotificationContext event) {
 
   public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
       Object listener, EventType[] eventTypes, ChangeType changeType) {
-    this(manager, client, propertyKey, listener, eventTypes, changeType, null);
+    this(manager, client, propertyKey, listener, eventTypes, changeType, null, -1);
```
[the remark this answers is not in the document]
mgao0: Nice catch. Fixed

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -202,6 +214,29 @@ protected void handleEvent(NotificationContext event) {
...
+          NotificationContext changeContext = new NotificationContext(_manager);
+          changeContext.setType(NotificationContext.Type.PERIODIC_REFRESH);
+          changeContext.setChangeType(_changeType);
+          enqueueTask(changeContext);
+        } else {
+          long remainingTime = _lastEventTime + _periodicRefreshInterval - currentTime;
+          _scheduledRefreshFuture.cancel(false);
+          _scheduledRefreshFuture = _periodicRefreshExecutor
+              .scheduleWithFixedDelay(this, remainingTime, _periodicRefreshInterval,
+                  TimeUnit.MILLISECONDS);
```
NealSun96, line 232: Instead of doing this, can you just "sleep" for `remainingTime`, then do the check again? I think that automatically adjusts the schedule.

mgao0: Good idea! I changed it to sleep. One concern I had is that the thread could be interrupted while sleeping, but since we only have one thread here, I guess it's fine.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
mgao0: Yes, the shutdownNow method will automatically cancel the scheduled tasks. Now I changed to another way for reset (not shutting down), I only cancel the scheduled future now, instead of shut down the executor and initialize another one, I think it's more efficient.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
mgao0: Added this comment in CallbackHandler. Thanks for reminding!

helix-core/src/main/java/org/apache/helix/messaging/DefaultMessagingService.java
mgao0: Resolving this conversation since we are not doing it this way now.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
mgao0: Per our conversation, moved this change from HelixTaskExecutor to CallbackHandler. Instead of call onMessage directly, for refresh, we enqueue a specific event to the queue to trigger onMessage.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -202,6 +212,26 @@ protected void handleEvent(NotificationContext event) {
...
+      try {
+        long currentTime = System.currentTimeMillis();
+        if (_lastEventTime + _periodicRefreshInterval <= currentTime) {
+          NotificationContext changeContext = new NotificationContext(_manager);
+          changeContext.setType(NotificationContext.Type.PERIODIC_REFRESH);
+          changeContext.setChangeType(_changeType);
+          enqueueTask(changeContext);
+        } else {
+          long remainingTime = _lastEventTime + _periodicRefreshInterval - currentTime;
+          Thread.sleep(remainingTime);
```
NealSun96, line 227: You need to do the refresh check again here, otherwise a whole new period/cycle will pass without refresh. :)

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -1135,6 +1145,34 @@ private void removeMessageFromZK(HelixDataAccessor accessor, Message message,
...
+          @Override
+          public void run() {
+            // Trigger read messages from zk
+            NotificationContext changeContext = new NotificationContext(_manager);
+            changeContext.setChangeType(HelixConstants.ChangeType.MESSAGE);
+            onMessage(_manager.getInstanceName(), null, changeContext);
+          }
+        };
+        _messageRefreshTimer.schedule(messageRefreshTask, periodicMessageRefreshInterval,
+            periodicMessageRefreshInterval);
```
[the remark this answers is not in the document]
mgao0: Good suggestion. In current design, I put the check in invoke method, and if the onMessage (in the case of messages) is not called for a period of time, it will do the refresh.

helix-core/src/test/java/org/apache/helix/messaging/handling/TestPerioidicMessageRefresh.java
```diff
@@ -0,0 +1,141 @@
...
+    startMessageRefreshThread.setAccessible(true);
+    startMessageRefreshThread.invoke(executor);
+
+    // Now we should be able to get the messages and process them
+    TestHelper.verify(new TestHelper.Verifier() {
+      @Override
+      public boolean verify() throws Exception {
+        return factory._processedMsgIds.size() == messageIds.size();
+      }
+    }, 2000);
```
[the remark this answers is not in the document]
mgao0: Didn't know we have this. Thanks.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
mgao0: Didn't think of that! Thanks for catching this. I think I would prefer my previous method then, the framework handles it better. And I searched, to cancel a future and to schedule another one, is merely to remove a task from workQueue and create another task and put it into workQueue, so it won't be too costly, especially if we are doing it for relatively long interval (20, 30 minutes).

NealSun96: Ok. I personally liked this approach because it's cleaner, but I'll let you be the judge of that.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -110,9 +112,12 @@
    * define the next possible notification types
    */
   private static Map<Type, List<Type>> nextNotificationType = new HashMap<>();
+
   static {
-    nextNotificationType.put(Type.INIT, Arrays.asList(Type.CALLBACK, Type.FINALIZE));
-    nextNotificationType.put(Type.CALLBACK, Arrays.asList(Type.CALLBACK, Type.FINALIZE));
+    nextNotificationType
```
dasahcc, line 117: Is this agreed? I am not sure. I was preferring to have a separate type to by pass even if there is a callback wrong order issue.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -392,17 +450,18 @@ public void invoke(NotificationContext changeContext) throws Exception {
         logger.warn("Callback handler received event in wrong order. Listener: " + _listener
```
dasahcc, line 450: My assumption is that you have a new type to by pass when there is a callback type wrong order. But here you may still have the problem that current type is FINALIZE and coming with CALLBACK. And your type PERODIC_REFRESH will also be recognized as a wrong order type.

I would suggest rethink the value of adding the new type. What is the purpose to have a new type.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -292,6 +295,8 @@ public ZKHelixManager(String clusterName, String instanceName, InstanceType inst
         .getSystemPropertyAsInt(SystemPropertyKeys.PARTICIPANT_HEALTH_REPORT_LATENCY,
             ParticipantHealthReportTask.DEFAULT_REPORT_LATENCY);
 
+    _messageRefreshInterval =  HelixUtil.getSystemPropertyAsLong(SystemPropertyKeys.MESSAGE_REFRESH_INTERVAL, VALUE_NOT_SET);
```
dasahcc, line 298: Do we need a variable for that? I dont think we will reuse the value in later stage.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
dasahcc: There would be a good example of Delayed Scheduler. Try to understand that mode.

kaisun2000:
I just saw this diff. In fact, this is a totally different from what it was before. I have quite some reservation to add periodical refresh in CallbackHandler. 

CallbackHandler is the interface with zkclient managing the lifecycle of a specific Helix property path. Specifically they handle notification (further calling back to application logic) and registering further notification logic. 

The part actually already had some notable issues (lose notification, or installing "leaked path etc) that should be addressed. We should not complicate this part with further here.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
mgao0: I agree. I have updated the code, please take a look.

mgao0:
> I just saw this diff. In fact, this is a totally different from what it was before. I have quite some reservation to add periodical refresh in CallbackHandler.
> 
> CallbackHandler is the interface with zkclient managing the lifecycle of a specific Helix property path. Specifically they handle notification (further calling back to application logic) and registering further notification logic.
> 
> The part actually already had some notable issues (lose notification, or installing "leaked path etc) that should be addressed. We should not complicate this part with further here.

Doing periodic refresh in callback handler can actually resolve the issue that you mentioned such as lost notification or missing watcher. This periodic refresh logic is very independent, it's a separate thread, and it put an event in the event queue only if there is no event processed for a while. That being said, it the callback handler is functioning well and events are processed normally, it won't add much burden to the performance since the periodic refresh event won't be generated.
Is there any specific concern you have on the periodic refresh?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang: Let's just use CALLBACK for the periodic triggering.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -131,6 +136,10 @@
   private boolean _batchModeEnabled = false;
   private boolean _preFetchEnabled = true;
   private HelixCallbackMonitor _monitor;
+  private final long _periodicRefreshInterval;
```
jiajunwang, line 139: For generosity, shall we call it periodicCallbackInterval, or periodicTriggerInterval?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang: This logic seems to be incorrect... But I guess you will change it anyway.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -202,6 +211,30 @@ protected void handleEvent(NotificationContext event) {
     }
   }
 
+  class RefreshTask implements Runnable {
+    @Override
+    public void run() {
+      try {
+        long currentTime = System.currentTimeMillis();
+        long remainingTime = _lastEventTime + _periodicRefreshInterval - currentTime;
+        if (remainingTime <= 0) {
```
jiajunwang, line 220: What if there is already one periodic event in the queue? Shall we keep adding new ones?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -818,4 +884,15 @@ public String getContent() {
...
   }
+
+  private void initRefreshTask() {
+    _lastEventTime = System.currentTimeMillis();
+    _periodicRefreshExecutor = new ScheduledThreadPoolExecutor(1);
+    // When cancelling the task future, it removes the task from the queue
+    // so we won't have a memory leakage when we cancel scheduled task
+    _periodicRefreshExecutor.setRemoveOnCancelPolicy(true);
+    _scheduledRefreshFuture = _periodicRefreshExecutor
+        .scheduleWithFixedDelay(new RefreshTask(), _periodicRefreshInterval,
```
jiajunwang, line 895: I think we talked about at least 2 of the following points:
1. If the callback is triggered, the timer should be reset and the wait until the interval passes.
2. If the queue is not empty, then we don't need to trigger the refresh.

jiajunwang: They need to be done to avoid unnecessary performance impact.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
mgao0: Resolving this comment since we do not directly call onMessage in the current design.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
mgao0: @jiajunwang The previous PR contains the first condition you mentioned. I added the second condition in the updated PR. And both checks are still being done in the callback handler instead of ZK event queue.
As per our discussion on Friday, I tried to do this in the ZKEvent queue, and found it hard to do, when an event coming in at the ZKEvent level, we don't know the change type of the event, thus not able to reset the timer accordingly. Currently at the ZKEvent level, no Helix logics are involved so I think it's better we keep it that way.
For your concern on parallel invoke, we synchronize on Helix manager in the invoke method, so it should be fine. 
Let's discuss more if you still have concern about this.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
mgao0: Updated this part. Please take a look.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -210,12 +249,25 @@ protected void handleEvent(NotificationContext event) {
 
   public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
       Object listener, EventType[] eventTypes, ChangeType changeType) {
-    this(manager, client, propertyKey, listener, eventTypes, changeType, null);
+    this(manager, client, propertyKey, listener, eventTypes, changeType, null, -1);
+  }
+
+  public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
```
jiajunwang, line 255: You don't need this one I guess.
Just add the following one,
```
public CallbackHandler(HelixManager manager, RealmAwareZkClient client, PropertyKey propertyKey,
      Object listener, EventType[] eventTypes, ChangeType changeType, HelixCallbackMonitor monitor,
      long periodicRefreshTriggerInterval)
```

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -693,15 +758,15 @@ public void handleDataChange(String dataPath, Object data) {
...
-        NotificationContext changeContext = new NotificationContext(_manager);
-        changeContext.setType(NotificationContext.Type.CALLBACK);
-        changeContext.setPathChanged(dataPath);
-        changeContext.setChangeType(_changeType);
-        enqueueTask(changeContext);
-      }
-    } catch (Exception e) {
-      String msg =
-          "exception in handling data-change. path: " + dataPath + ", listener: " + _listener;
+      NotificationContext changeContext = new NotificationContext(_manager);
```
jiajunwang, line 761: format

jiajunwang: I go back and check your PR description. You did click the "Code Quality" bullet. Please ensure that when you click it, you really did so.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -200,6 +208,37 @@ protected void handleEvent(NotificationContext event) {
...
+          long currentTime = System.currentTimeMillis();
+          long remainingTime = _lastEventTime + _periodicRefreshTriggerInterval - currentTime;
+          if (remainingTime <= 0) {
+            // If there is no event in the queue, meaning this long idle time could be due to lack of events
+            // Otherwise, if there is event in the queue, this long idle time is due to slow process of events
+            if (_batchCallbackProcessor == null || _batchCallbackProcessor.queueIsEmpty()) {
+              NotificationContext changeContext = new NotificationContext(_manager);
+              changeContext.setType(Type.CALLBACK);
+              changeContext.setChangeType(_changeType);
+              enqueueTask(changeContext);
```
jiajunwang, line 231: Still the same question, if batch mode is false, this will be running in parallel with the ZKEvent thread, right? How to avoid conflict?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -200,6 +208,37 @@ protected void handleEvent(NotificationContext event) {
...
+            }
+            break;
+          } else {
+            Thread.sleep(remainingTime);
```
jiajunwang, line 235: I know that you are not holding any lock here, but I would prefer using wait(remainingTime)

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -171,6 +171,10 @@
   private CallbackHandler _leaderElectionHandler = null;
   protected final List<HelixTimerTask> _controllerTimerTasks = new ArrayList<>();
 
+  /** Set this variable to negative can disable the message periodic refresh feature */
+  private long _messageRefreshTriggerInterval;
```
jiajunwang, line 175: final?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -405,6 +413,11 @@ void checkConnected(long timeout) {
 
   void addListener(Object listener, PropertyKey propertyKey, ChangeType changeType,
       EventType[] eventType) {
+    addListener(listener, propertyKey, changeType, eventType, -1);
+  }
+
+  void addListener(Object listener, PropertyKey propertyKey, ChangeType changeType,
+      EventType[] eventType, long periodicRefreshTriggerInterval) {
```
jiajunwang, line 420: Refresh and Trigger are duplicates somehow.
I think we just call it periodicTriggerInterval for generic.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -548,14 +561,14 @@ public void addConfigChangeListener(org.apache.helix.ScopedConfigChangeListener
...
-        new EventType[] { EventType.NodeChildrenChanged });
+        new EventType[] { EventType.NodeChildrenChanged }, _messageRefreshTriggerInterval);
   }
 
   @Deprecated
   @Override
   public void addMessageListener(org.apache.helix.MessageListener listener, String instanceName) {
     addListener(listener, new Builder(_clusterName).messages(instanceName), ChangeType.MESSAGE,
-        new EventType[] { EventType.NodeChildrenChanged });
+        new EventType[] { EventType.NodeChildrenChanged }, _messageRefreshTriggerInterval);
```
jiajunwang, line 571: optional, but I guess we don't need to add support for the deprecated methods.

And one question, why not just casting the deprecated listener class to the newer one? So it becomes maintenance-free.

helix-core/src/test/java/org/apache/helix/TestPeriodicRefresh.java
```diff
@@ -0,0 +1,232 @@
...
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import static org.apache.helix.HelixConstants.ChangeType.MESSAGE;
+
+
+/**
+ * This class tests that if there are no incoming events, the onMessage method in message listener will be called by message periodic refresh
+ */
+public class TestPeriodicRefresh extends ZkUnitTestBase {
```
jiajunwang, line 45: @kaisun2000 since you raise this concern, please review this test case and verify if this is enough or not : )

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -131,6 +135,10 @@
   private boolean _batchModeEnabled = false;
   private boolean _preFetchEnabled = true;
   private HelixCallbackMonitor _monitor;
+  private final long _periodicRefreshTriggerInterval;
```
jiajunwang, line 138: I think I mentioned in some earlier comments that Refresh is not general enough for the CallbackHandler. Let's just call it _periodicTriggerInterval

mgao0: Didn't get what you meant before. Changed it.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
mgao0: Good idea. This is better way to do for the deprecated listener. Updated it.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang: The problem of parallelism here is not 2 threads try to run at the same time. As you mentioned it is protected by synchronization.

The concern is out of order. I mentioned this scenario during the discussion. But I guess the argument that it won't make it worse since we are basically doing a similar batch mode processing here.
Meaning, whatever potential issues that we saw in the batch mode may appear in this refresh.

Let's put this concern aside. Since if any problem, we will need to fix it for the batch mode anyway.

jiajunwang:
Thanks for the update. Could you please add more information in the description about what the issue this refresh trying to address? I mean we discussed several scenarios that cannot be addressed by the other fixes, then we decided to proceed to finish this PR. But let's make it clear what is the missing part.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -245,6 +294,17 @@ public CallbackHandler(HelixManager manager, RealmAwareZkClient client, Property
...
+
+    /**
+     * Async batch read was designed that if an exception is thrown, the operation fails and returns a partial result.
+     * The issue is separately addressed by checking read result in batch read calls [https://github.com/apache/helix/pull/974]
+     * and retry batch read for ZK connectivity issues [https://github.com/apache/helix/pull/970].
+     * We also add this one to periodical read stale messages from ZkServer in the case we don't see any event for a certain period.
+     */
+    _periodicTriggerInterval = periodicTriggerInterval;
+    if (_periodicTriggerInterval > 0) {
+      initRefreshTask();
```
jiajunwang, line 306: nit, pass the periodicTriggerInterval to initRefreshTask() and do the branch inside. That makes this code cleaner.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -200,6 +206,43 @@ protected void handleEvent(NotificationContext event) {
...
+              changeContext.setType(Type.CALLBACK);
+              changeContext.setChangeType(_changeType);
+              enqueueTask(changeContext);
+            }
+            break;
+          } else {
+            wait(remainingTime);
+          }
+        }
+      } catch (Exception e) {
```
jiajunwang, line 242: For the interruption exception, we may want to exit the execution.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -403,6 +463,10 @@ public void invoke(NotificationContext changeContext) throws Exception {
         subscribeForChangesAsyn(changeContext.getType(), _path, _watchChild);
       }
 
+      if (_periodicTriggerInterval > 0) {
```
jiajunwang, line 466: This is not the direct indicator, how about check if _periodicRefreshExecutor != null?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -781,8 +845,15 @@ void reset(boolean isShutdown) {
...
             _batchCallbackProcessor.shutdown();
             _batchCallbackProcessor = null;
+            if (_periodicRefreshExecutor != null) {
+              _periodicRefreshExecutor.shutdownNow();
+            }
           } else {
             _batchCallbackProcessor.resetEventQueue();
+            if (_scheduledRefreshFuture != null) {
+              _periodicRefreshExecutor.shutdownNow();
+              initRefreshTask();
```
jiajunwang, line 855: The next event would be FINALIZE, so even the periodic task is triggered, it won't really invoke the handling logic successfully. I think we shall not to init here.

Shall we do it in the init() call?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -131,6 +133,10 @@
   private boolean _batchModeEnabled = false;
   private boolean _preFetchEnabled = true;
   private HelixCallbackMonitor _monitor;
+  private final long _periodicTriggerInterval;
+  private ScheduledThreadPoolExecutor _periodicRefreshExecutor;
```
jiajunwang, line 137: _periodicTriggerExecutor

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -131,6 +133,10 @@
...
+  private ScheduledFuture<?> _scheduledRefreshFuture;
```
jiajunwang, line 138: _scheduledTriggerFuture

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -200,6 +206,43 @@ protected void handleEvent(NotificationContext event) {
...
+    }
+  }
+
+  /**
+   * This class represents each periodic refresh task
+   * If the deadline passed, enqueue an event to do a refresh
+   * If the deadline has not passed (some events happened in between so don't need to do a refresh),
+   * wait until reaching the interval and check again
+   */
+  class RefreshTask implements Runnable {
```
jiajunwang, line 221: TriggerTask?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang, line 306: initTriggerTask()?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -171,6 +171,10 @@
   private CallbackHandler _leaderElectionHandler = null;
   protected final List<HelixTimerTask> _controllerTimerTasks = new ArrayList<>();
 
+  /** Set this variable to negative can disable the message periodic refresh feature */
+  private final long _messageRefreshTriggerInterval;
+  private final long DEFAULT_PERIODIC_MESSAGE_REFRESH_INTERVAL = 2 * 60 * 1000; //2 minutes
```
jiajunwang, line 176: static

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -200,6 +206,45 @@ protected void handleEvent(NotificationContext event) {
...
+    @Override
+    public void run() {
+      try {
+        while (true) {
+          long currentTime = System.currentTimeMillis();
+          long remainingTime = _lastInvokeTime + _periodicTriggerInterval - currentTime;
+          if (remainingTime <= 0) {
+            // If there is no event in the queue, meaning this long idle time could be due to lack of events
+            // Otherwise, if there is event in the queue, this long idle time is due to slow process of events
+            if (_batchCallbackProcessor == null || _batchCallbackProcessor.queueIsEmpty()) {
```
jiajunwang, line 231: If _batchCallbackProcessor is not empty, but the callback handling takes a while. then this loop will do a busy loop, right?
More graceful design is to wait until the queue becomes non-empty then loop to check the remaining time again.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -403,6 +462,10 @@ public void invoke(NotificationContext changeContext) throws Exception {
         subscribeForChangesAsyn(changeContext.getType(), _path, _watchChild);
       }
 
+      if (_periodicTriggerExecutor != null) {
```
jiajunwang, line 465: Can we put this in the initTriggerTask() method?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -818,4 +883,38 @@ public String getContent() {
...
+      _periodicTriggerExecutor.shutdownNow();
+      _periodicTriggerExecutor = null;
+    }
+  }
+
+  /**
+   * Used to initialize a periodic triggered task
+   * Schedule tasks in a task executor with fixed intervals
+   */
+  private void initTriggerTask() {
```
jiajunwang, line 904: nit, take _periodicTriggerInterval as the parameter, and don't refer to the global private field. This is a good style in which you can reduce the impact of a private field in the class. So less dependency if you want to make any further change.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -818,4 +883,38 @@ public String getContent() {
...
+    shutDownTriggerTask(false);
+    _lastInvokeTime = System.currentTimeMillis();
+    if (_periodicTriggerExecutor == null) {
+      _periodicTriggerExecutor = new ScheduledThreadPoolExecutor(1);
+      // When cancelling the task future, it removes the task from the queue
+      // so we won't have a memory leakage when we cancel scheduled task
+      _periodicTriggerExecutor.setRemoveOnCancelPolicy(true);
+    }
+    _scheduledTriggerFuture = _periodicTriggerExecutor
+        .scheduleWithFixedDelay(new TriggerTask(), _periodicTriggerInterval,
```
jiajunwang, line 917: I overlooked this part before. Why do you need to use scheduleWithFixedDelay() here? The TriggerTask object will keep running in the thread, right?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -200,6 +206,45 @@ protected void handleEvent(NotificationContext event) {
...
+   * If the deadline has not passed (some events happened in between so don't need to do a refresh),
+   * wait until reaching the interval and check again
+   */
+  class TriggerTask implements Runnable {
+    @Override
+    public void run() {
+      try {
+        while (true) {
+          long currentTime = System.currentTimeMillis();
+          long remainingTime = _lastInvokeTime + _periodicTriggerInterval - currentTime;
```
jiajunwang, line 227: To reduce the usage of _periodicTriggerInterval in the whole callback class, could you please pass the interval as a constructor parameter? That would be a cleaner design.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -405,6 +413,11 @@ void checkConnected(long timeout) {
 
   void addListener(Object listener, PropertyKey propertyKey, ChangeType changeType,
       EventType[] eventType) {
+    addListener(listener, propertyKey, changeType, eventType, -1);
+  }
+
+  void addListener(Object listener, PropertyKey propertyKey, ChangeType changeType,
```
jiajunwang, line 419: private?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
mgao0: Great suggestion! Fixed. Thanks!

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
mgao0: Discussed offline. A single long running thread is sufficient. Updated.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
mgao0: Updated.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang: I don't think it will deadlock. But this also works.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang: Let's put it to the very beginning of this method?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -132,6 +136,18 @@
   private boolean _batchModeEnabled = false;
   private boolean _preFetchEnabled = true;
   private HelixCallbackMonitor _monitor;
+  private final long _periodicTriggerInterval;
+  private final static ExecutorService _periodicTriggerExecutor =
+      Executors.newFixedThreadPool(40, new ThreadFactory() {
+        public Thread newThread(Runnable runnable) {
+          Thread thread = Executors.defaultThreadFactory().newThread(runnable);
+          thread.setDaemon(true);
+          thread.setName("TriggerTaskThread");
```
jiajunwang, line 145: PeriodicTriggerTask. thread is not necessary, IMHO.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -352,6 +434,10 @@ public String getPath() {
   }
 
   public void enqueueTask(NotificationContext changeContext) throws Exception {
+    if (_periodicTriggerExecutor != null) {
```
jiajunwang, line 437: This check is no longer valid. We should use the interval or just use a boolean (_periodicTriggerEnabled) so it is clearer

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -132,6 +136,18 @@
   private boolean _batchModeEnabled = false;
   private boolean _preFetchEnabled = true;
   private HelixCallbackMonitor _monitor;
+  private final long _periodicTriggerInterval;
+  private final static ExecutorService _periodicTriggerExecutor =
```
jiajunwang, line 140: the static final field follows different name conventions.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -201,6 +217,58 @@ protected void handleEvent(NotificationContext event) {
...
+    private final long _triggerInterval;
+
+    public TriggerTask(long triggerInterval) {
+      this._triggerInterval = triggerInterval;
+    }
+
+    @Override
+    public void run() {
+      logger.info("Starting trigger task.");
+      while (!Thread.currentThread().isInterrupted()) {
```
jiajunwang, line 242: As we discussed this may cause the runnable to occupy the thread forever, let's try the schedule executor back.
Unfortunately, the simpler design fails the tests. I hope the older design will work.

helix-core/src/test/java/org/apache/helix/TestPeriodicRefresh.java
narendly: https://github.com/apache/helix/pull/1000/files#diff-a7737834d48d58f89a95636b50b2858589a27e7043625bffaaa68ece67e3c361R22-R38

jiajunwang:
Close due to inactive.

Issue #1001, opened 2020-05-11
Avoid race condition between removing a message and write "READ" state message back to ZK


mgao0:
Usual process for a participant handles a message is that receive new message -> mark message "READ" state and write back to ZK -> complete ST and remove the message. We could potentially see race conditions between the 2nd and 3rd step if we enable state transition cancellation, which could result a message kept being written back to zk although it is already processed.

PR #1002, opened 2020-05-11
Use updaters to update read messages to ZK

mgao0:
### Issues

#1001

### Description

This PR uses an updater to replace set, so it checks whether there is a message in the path beforehand, and only write the "READ" state message to ZK if there is a message, to avoid write back a message which is already removed.

### Tests

TestHelixTaskExecutor -> testNoWriteReadStateForRemovedMessage

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestWorkflowTermination.testWorkflowRunningTimeout:131->verifyWorkflowCleanup:257 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1147, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:15 h
[INFO] Finished at: 2020-05-12T18:59:45-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -37,6 +37,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.helix.AccessOption;
 import org.apache.helix.ConfigAccessor;
 import org.apache.helix.Criteria;
 import org.apache.helix.HelixConstants;
@@ -69,6 +70,8 @@ import org.apache.helix.participant.statemachine.StateModel;
 import org.apache.helix.participant.statemachine.StateModelFactory;
 import org.apache.helix.util.HelixUtil;
 import org.apache.helix.util.StatusUpdateUtil;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.apache.helix.zookeeper.zkclient.DataUpdater;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -524,12 +527,26 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
   private void updateMessageState(List<Message> readMsgs, HelixDataAccessor accessor,
       String instanceName) {
     Builder keyBuilder = accessor.keyBuilder();
-    List<PropertyKey> readMsgKeys = new ArrayList<>();
+    List<String> readMsgPaths = new ArrayList<>();
+    List<DataUpdater<ZNRecord>> updaters = new ArrayList<>();
     for (Message msg : readMsgs) {
-      readMsgKeys.add(msg.getKey(keyBuilder, instanceName));
+      readMsgPaths.add(msg.getKey(keyBuilder, instanceName).getPath());
       _knownMessageIds.add(msg.getId());
+      /**
+       * We use the updater to avoid race condition between writing message to zk as READ state and removing message after ST is done
+       * If there is no message at this path, meaning the message is removed so we do not write the message
+       */
+      updaters.add(new DataUpdater<ZNRecord>() {
+        @Override
+        public ZNRecord update(ZNRecord currentData) {
+          if (currentData == null) {
+            return null;
+          }
+          return msg.getRecord();
+        }
+      });
     }
-    accessor.setChildren(readMsgKeys, readMsgs);
+    accessor.updateChildren(readMsgPaths, updaters, AccessOption.PERSISTENT);
   }
 
   private void shutdownAndAwaitTermination(ExecutorService pool) {
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -540,6 +540,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
         @Override
         public ZNRecord update(ZNRecord currentData) {
           if (currentData == null) {
+            LOG.warn("Message {} targets at {} has already been removed before it is set as READ on instance {}", msg.getId(), msg.getTgtName(), instanceName);
             return null;
           }
           return msg.getRecord();
--- a/helix-core/src/test/java/org/apache/helix/MockAccessor.java
+++ b/helix-core/src/test/java/org/apache/helix/MockAccessor.java
@@ -245,8 +245,7 @@ public class MockAccessor implements HelixDataAccessor {
   @Override
   public <T extends HelixProperty> boolean[] updateChildren(List<String> paths,
       List<DataUpdater<ZNRecord>> updaters, int options) {
-    // TODO Auto-generated method stub
-    throw new HelixException("Method not implemented!");
+    return _baseDataAccessor.updateChildren(paths, updaters, options);
   }
 
   @Override
--- a/helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
+++ b/helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
@@ -19,6 +19,8 @@ package org.apache.helix.messaging.handling;
  * under the License.
  */
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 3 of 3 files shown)


helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -524,12 +537,26 @@ public void finishTask(MessageTask task) {
...
       _knownMessageIds.add(msg.getId());
+      /**
+       * We use the updater to avoid race condition between writing message to zk as READ state and removing message after ST is done
+       * If there is no message at this path, meaning the message is removed so we do not write the message
+       */
+      updaters.add(new DataUpdater<ZNRecord>() {
+        @Override
+        public ZNRecord update(ZNRecord currentData) {
+          if (currentData == null) {
+            return null;
```
dasahcc, line 553: We should throw exception here.

mgao0: I checked the code, "return null" won't cause an exception to be thrown. The zkclient won't set null to zk, and accessor will return a boolean[] for updateChildren, where it returns false for the node whose updater returns null. If you think throw exception is a better way to handle it, I can change it too.

helix-core/src/main/java/org/apache/helix/messaging/DefaultMessagingService.java
```diff
@@ -70,7 +70,7 @@ public DefaultMessagingService(HelixManager manager) {
 
     _taskExecutor = new HelixTaskExecutor(
         new ParticipantStatusMonitor(isParticipant, manager.getInstanceName()),
-        new MessageQueueMonitor(manager.getClusterName(), manager.getInstanceName()));
+        new MessageQueueMonitor(manager.getClusterName(), manager.getInstanceName()), manager);
```
jiajunwang, line 73: Did you miss some files here? MessageQueueMonitor does not have such a constructor and it is not changed in this PR.

jiajunwang: In general, passing the manager to the monitor object would be an overkill. If it is really necessary, then we shall change the constructor to accept the manager only, since it includes all the information.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -152,6 +158,11 @@ public HelixTaskExecutor(ParticipantStatusMonitor participantStatusMonitor) {
 
   public HelixTaskExecutor(ParticipantStatusMonitor participantStatusMonitor,
```
jiajunwang, line 159: Deprecate it?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: This is the assumption, if the current data does not exist, return null. I think it is fine.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -524,12 +537,26 @@ public void finishTask(MessageTask task) {
...
+       * We use the updater to avoid race condition between writing message to zk as READ state and removing message after ST is done
+       * If there is no message at this path, meaning the message is removed so we do not write the message
+       */
+      updaters.add(new DataUpdater<ZNRecord>() {
+        @Override
+        public ZNRecord update(ZNRecord currentData) {
+          if (currentData == null) {
+            return null;
+          }
+          return msg.getRecord();
```
jiajunwang, line 555: If you always return msg.getRecord() why not just set? You are not updating the currentData anyway.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -714,7 +741,7 @@ private void syncSessionToController(HelixManager manager) {
 
   @Override
   @PreFetch(enabled = false)
-  public void onMessage(String instanceName, List<Message> messages,
+  public synchronized void onMessage(String instanceName, List<Message> messages,
```
jiajunwang, line 744: Why? The periodic refresh does not need to be done synchronized, I think.
If your timer action just adds the refresh event to the queue, there is no need to modify this.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -1135,6 +1162,34 @@ private void removeMessageFromZK(HelixDataAccessor accessor, Message message,
...
+            throw new HelixException("Periodic message refresh skipped. No helix manager is set.");
+          }
+        }
+        TimerTask messageRefreshTask = new TimerTask() {
+          @Override
+          public void run() {
+            // Trigger read messages from zk
+            NotificationContext changeContext = new NotificationContext(_manager);
+            changeContext.setChangeType(HelixConstants.ChangeType.MESSAGE);
+            onMessage(_manager.getInstanceName(), null, changeContext);
```
jiajunwang, line 1182: Why do you need to call it here synchronously? I think it's simpler if we just put this event into the event queue.
Given this saying, this logic will be fitting better if it is in the HelixManager level. What I'm trying to propose if the HelixManger is initialized as a Participant, then we start this timer thread there. It just injects message event with a certain interval. Of course, if a real message event comes earlier, then we reset the timer.

Put this logic in the executor does not make sense to me.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
mgao0: I use the updater here to check if the current data exists, if not then don't set the new data. And if there is current data, then just set the new data, that's why I'm not updating currentData here.

jiajunwang: I see, could you please comment the reason here?

jiajunwang: Nevermind, I see you already noted it down above.

mgao0:
> I think the design is a little bit awkward. Please check out my comments and see if that would be an easier design.

I moved all your comments on periodic message refresh to https://github.com/apache/helix/pull/1000. I decided to separate the PR since it's trying to resolve two different issues. Sorry for the confusion, please refer to the link above for the replies on your comments.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -524,12 +527,27 @@ public void finishTask(MessageTask task) {
...
       _knownMessageIds.add(msg.getId());
+      /**
+       * We use the updater to avoid race condition between writing message to zk as READ state and removing message after ST is done
+       * If there is no message at this path, meaning the message is removed so we do not write the message
+       */
+      updaters.add(new DataUpdater<ZNRecord>() {
+        @Override
+        public ZNRecord update(ZNRecord currentData) {
+          if (currentData == null) {
+            LOG.warn("Message {} targets at {} has already been removed before it is set as READ on instance {}", msg.getId(), msg.getTgtName(), instanceName);
```
mgao0, line 543: @jiajunwang I have added the logging. Can you please check if the message is good?

mgao0:
This PR is ready to be merged, approved by @dasahcc @jiajunwang .
Final commit message:
Use updaters to update messages with READ state to ZK

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -524,12 +527,27 @@ public void finishTask(MessageTask task) {
...
+    List<DataUpdater<ZNRecord>> updaters = new ArrayList<>();
     for (Message msg : readMsgs) {
-      readMsgKeys.add(msg.getKey(keyBuilder, instanceName));
+      readMsgPaths.add(msg.getKey(keyBuilder, instanceName).getPath());
       _knownMessageIds.add(msg.getId());
+      /**
+       * We use the updater to avoid race condition between writing message to zk as READ state and removing message after ST is done
+       * If there is no message at this path, meaning the message is removed so we do not write the message
+       */
+      updaters.add(new DataUpdater<ZNRecord>() {
```
lei-xia, line 539: One optimization here maybe to create just one updater instead of one per each message. You can define the update logic in the updater, i.e, read the message and mark the read field as "read".

PR #1006, opened 2020-05-13
Fix the RuntimeJobDag issue for queues

### LINKED ISSUE #1005: RunTimeJobDag might have multiple instances of a job
It has been observed that if ParallelJobs has a value more than one, If controller switch happens or several running jobs finish in the same pipeline, the _readyJobList in the RunTimeJobDag will have multiple instance of the same job. This will cause problems for the queue that has ParallelJobs set to a value grater than one. It means that after controller switch, the jobs will run one by one. We need to add a check to make sure _readyJobList contains unique jobs.

alirezazamani:
### Issues
Fixes #1005 

### Description
In this PR, a fix has been implemented to avoid _readyJobList in RuntimeJobDag to contain multiple entries of a job. In this RuntimeJobDah.finishJob we add the next job without considering of it has already been added or not. In this PR, added a test which proves this can happen and I proposed a fix for this issue.

### Tests
TestEnqueueJobs.testQueueParallelJobs

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestRecurringJobQueue.deleteRecreateRecurrentQueue » ThreadTimeout Method org....
[ERROR]   TestWorkflowTermination.testWorkflowJobFail:223 » Helix Workflow "testWorkflow...
[INFO] 
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:24 h
[INFO] Finished at: 2020-05-13T13:55:21-07:00
```

The failed tests have passed when I ran them individually.
```
mvn test -Dtest="TestRecurringJobQueue,TestWorkflowTermination"
[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 137.835 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  02:23 min
[INFO] Finished at: 2020-05-13T14:21:58-07:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
+++ b/helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
@@ -146,13 +146,13 @@ public class RuntimeJobDag extends JobDag {
     }
     // Add finished job's successors to ready-list
 
-    // If it is a jobQueue, there should be a check to make sure that the a job has not been added
-    // to the _readyJobList multiple times. This check is necessary because once the controller
-    // switch happens, the _readyJobList and _inflightJobList will be created from scratch. In this
-    // case, since there might be many jobs that have been finished before, we do not want to have a
-    // job several times to the _readyJobList. If _readyJobList has multiple instances of the same
-    // job, it can compromise the functionality of the parallel jobs.
     if (_isJobQueue) {
+      // If it is a jobQueue, there should be a check to make sure that the a job has not been added
+      // to the _readyJobList multiple times. This check is necessary because once the controller
+      // switch happens, the _readyJobList and _inflightJobList will be created from scratch. In this
+      // case, since there might be many jobs that have been finished before, we do not want to have a
+      // job several times to the _readyJobList. If _readyJobList has multiple instances of the same
+      // job, it can compromise the functionality of the parallel jobs.
       while (_lastJob != null && _parentsToChildren.containsKey(_lastJob)) {
         String nextJob = _parentsToChildren.get(_lastJob).iterator().next();
         if (!_readyJobList.contains(nextJob)) {
--- a/helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
+++ b/helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
@@ -146,8 +146,13 @@ public class RuntimeJobDag extends JobDag {
     }
     // Add finished job's successors to ready-list
     if (_isJobQueue) {
-      if (_lastJob != null && _parentsToChildren.containsKey(_lastJob)) {
-        _readyJobList.offer(_parentsToChildren.get(_lastJob).iterator().next());
+      while (_lastJob != null && _parentsToChildren.containsKey(_lastJob)) {
+        String nextJob = _parentsToChildren.get(_lastJob).iterator().next();
+        if (!_readyJobList.contains(nextJob)) {
+          _readyJobList.offer(nextJob);
+          break;
+        }
+        _lastJob = nextJob;
       }
     } else if (_successorMap.containsKey(job)) {
         for (String successor : _successorMap.get(job)) {
--- a/helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
+++ b/helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
@@ -145,6 +145,13 @@ public class RuntimeJobDag extends JobDag {
           String.format("Job: %s has either finished already, never been scheduled, or been removed from DAG", job));
     }
     // Add finished job's successors to ready-list
+
+    // If it is a jobQueue, there should be a check to make sure that the a job has not been added
+    // to the _readyJobList multiple times. This check is necessary because once the controller
+    // switch happens, the _readyJobList and _inflightJobList will be created from scratch. In this
+    // case, since there might be many jobs that have been finished before, we do not want to have a
+    // job several times to the _readyJobList. If _readyJobList has multiple instances of the same
+    // job, it can compromise the functionality of the parallel jobs.
     if (_isJobQueue) {
       while (_lastJob != null && _parentsToChildren.containsKey(_lastJob)) {
         String nextJob = _parentsToChildren.get(_lastJob).iterator().next();
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
@@ -105,10 +105,13 @@ public class TestEnqueueJobs extends TaskTestBase {
 
   @Test
   public void testQueueParallelJobs() throws InterruptedException {
+    final int parallelJobs = 3;
+    final int numberOfJobsAddedBeforeControllerSwitch = 4;
+    final int totalNumberOfJobs = 7;
     String queueName = TestHelper.getTestMethodName();
     JobQueue.Builder builder = TaskTestUtil.buildJobQueue(queueName);
     WorkflowConfig.Builder workflowCfgBuilder = new WorkflowConfig.Builder()
-        .setWorkflowId(queueName).setParallelJobs(3).setAllowOverlapJobAssignment(true);
+        .setWorkflowId(queueName).setParallelJobs(parallelJobs).setAllowOverlapJobAssignment(true);
     _driver.start(builder.setWorkflowConfig(workflowCfgBuilder.build()).build());
     JobConfig.Builder jobBuilder =
         new JobConfig.Builder().setTargetResource(WorkflowGenerator.DEFAULT_TGT_DB)
@@ -116,12 +119,12 @@ public class TestEnqueueJobs extends TaskTestBase {
             .setJobCommandConfigMap(Collections.singletonMap(MockTask.JOB_DELAY, "10000"));
 
     // Add 4 jobs to the queue
-    for (int i = 0; i <= 3; i++) {
+    for (int i = 0; i < numberOfJobsAddedBeforeControllerSwitch; i++) {
       _driver.enqueueJob(queueName, "JOB" + i, jobBuilder);
     }
 
     // Wait until all of the enqueued jobs (Job0 to Job3) are finished
-    for (int i = 0; i <= 3; i++) {
+    for (int i = 0; i < numberOfJobsAddedBeforeControllerSwitch; i++) {
       _driver.pollForJobState(queueName, TaskUtil.getNamespacedJobName(queueName, "JOB" + i),
           TaskState.COMPLETED);
     }
@@ -130,7 +133,7 @@ public class TestEnqueueJobs extends TaskTestBase {
     _controller.syncStop();
 
     // Add 3 more jobs to the queue which should run in parallel after the Controller is started
-    for (int i = 4; i <= 6; i++) {
+    for (int i = numberOfJobsAddedBeforeControllerSwitch; i < totalNumberOfJobs; i++) {
       _driver.enqueueJob(queueName, "JOB" + i, jobBuilder);
     }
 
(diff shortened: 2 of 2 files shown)


helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
```diff
@@ -146,8 +146,13 @@ public boolean finishJob(String job) {
     }
     // Add finished job's successors to ready-list
     if (_isJobQueue) {
-      if (_lastJob != null && _parentsToChildren.containsKey(_lastJob)) {
-        _readyJobList.offer(_parentsToChildren.get(_lastJob).iterator().next());
+      while (_lastJob != null && _parentsToChildren.containsKey(_lastJob)) {
```
dasahcc, line 149: Better to have some comments here. I think this is an optimization instead of fix. Even if you add the scheduled one as next job. It will be skipped in schedule job part in WorkflowDispatcher. 

Would that cause problem for get another next job?

alirezazamani: I will add the comments. Actually this is a fix. If you look at the test I added, it is failing without this fix. If you run the newly added test, you will notice that the jobs are not running in parallel because of this issue.

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
```diff
@@ -146,8 +146,13 @@ public boolean finishJob(String job) {
...
+        String nextJob = _parentsToChildren.get(_lastJob).iterator().next();
```
narendly, line 150: Could nextJob be null?

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -96,4 +102,60 @@ public void testJobSubmitGenericWorkflows() throws InterruptedException {
...
+    WorkflowConfig.Builder workflowCfgBuilder = new WorkflowConfig.Builder()
+        .setWorkflowId(queueName).setParallelJobs(3).setAllowOverlapJobAssignment(true);
+    _driver.start(builder.setWorkflowConfig(workflowCfgBuilder.build()).build());
+    JobConfig.Builder jobBuilder =
+        new JobConfig.Builder().setTargetResource(WorkflowGenerator.DEFAULT_TGT_DB)
+            .setCommand(MockTask.TASK_COMMAND).setMaxAttemptsPerTask(2)
+            .setJobCommandConfigMap(Collections.singletonMap(MockTask.JOB_DELAY, "10000"));
+
+    // Add 4 jobs to the queue
+    for (int i = 0; i <= 3; i++) {
```
narendly, line 119: Let's make these magic numbers descriptive variables?

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
narendly: +1. I believe we're making this improvement because we've seen an issue in production. Could you add a block comment here what the issue was and why this would fix the issue to give more context?

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -96,4 +102,60 @@ public void testJobSubmitGenericWorkflows() throws InterruptedException {
...
+    Set<Long> startTimes = new HashSet<>();
+    Set<Long> endTime = new HashSet<>();
+
+    for (int i = 4; i <= 6; i++) {
+      JobContext jobContext =
+          _driver.getJobContext(TaskUtil.getNamespacedJobName(queueName, "JOB" + i));
+      startTimes.add(jobContext.getStartTime());
+      endTime.add(jobContext.getFinishTime());
+    }
+    Assert.assertTrue(Collections.min(endTime) > Collections.max(startTimes));
```
pkuwm, line 159: Could you help understand: If you are only checking one: maxEndtime > maxStartime, why are HashSets necessary?

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
```diff
@@ -146,8 +146,13 @@ public boolean finishJob(String job) {
...
+        if (!_readyJobList.contains(nextJob)) {
```
pkuwm, line 151: What's the time complexity of this operation?

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
alirezazamani: Sure. Added comments about it. I will change the PR for master branch.

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
alirezazamani: Since _parentsToChildren.containsKey(_lastJob) is only added if there is actually a children for that job, as long as we are in the while loop, it means that nextJob is not null.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
alirezazamani: I understand you concern about time complexity. However, if you look at the whole structure of the code, finish job is only called once and only if the job is finished. Also parallel jobs is not usually set to a large value, I think we are safe here. Also note that we need that for correctness of TF.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
alirezazamani: Changed it to one long and get minimum/maximum in every loop.

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
narendly: could you add that explanation in the code?

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
```diff
@@ -145,9 +145,21 @@ public boolean finishJob(String job) {
           String.format("Job: %s has either finished already, never been scheduled, or been removed from DAG", job));
     }
     // Add finished job's successors to ready-list
+
+    // If it is a jobQueue, there should be a check to make sure that the a job has not been added
+    // to the _readyJobList multiple times. This check is necessary because once the controller
+    // switch happens, the _readyJobList and _inflightJobList will be created from scratch. In this
+    // case, since there might be many jobs that have been finished before, we do not want to have a
+    // job several times to the _readyJobList. If _readyJobList has multiple instances of the same
+    // job, it can compromise the functionality of the parallel jobs.
```
narendly, line 154: Have a follow up question. What do you mean the job gets added several times to the readyJobList?

If there's a controller switch, we'll be creating readyJobList/inflightJobList from scratch, and add everything from the beginning (or the first available job from the step-by-step topological sort). Then as long as the controller doesn't switch, it only gets added once, right?

In other words, for each controller should add a job exactly once. Is that correct?

Also even though a job might get added to readyJobList multiple times (exactly once in different controllers), our scheduling logic just skips it I believe (if jobState is complete/progress we just skip?). Could you help me understand how this could help? It seems that although this wouldn't hurt to do...

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -1,12 +1,18 @@
 package org.apache.helix.integration.task;
 
+import java.util.Collections;
+import java.util.HashSet;
```
pkuwm, line 4: Don't forget to remove unused imports.

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
pkuwm: This is within a while loop. How many times this check will be run?

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
alirezazamani: It is very hard to explain. I will do my best to explain it here. (If it is still not clear, I suggest you to just run the test I added to this PR and it will be very clear to you.)
Explanation: 
Here is the _readyJobList contents after the controller switch in my test (without this fix):

> [testQueueParallelJobs_JOB0, testQueueParallelJobs_JOB1, testQueueParallelJobs_JOB2]
> [testQueueParallelJobs_JOB1, testQueueParallelJobs_JOB2, testQueueParallelJobs_JOB1]
> [testQueueParallelJobs_JOB2, testQueueParallelJobs_JOB1, testQueueParallelJobs_JOB2]
> [testQueueParallelJobs_JOB1, testQueueParallelJobs_JOB2, testQueueParallelJobs_JOB3]
> [testQueueParallelJobs_JOB2, testQueueParallelJobs_JOB3, testQueueParallelJobs_JOB2]
> [testQueueParallelJobs_JOB3, testQueueParallelJobs_JOB2, testQueueParallelJobs_JOB3]
> [testQueueParallelJobs_JOB2, testQueueParallelJobs_JOB3, testQueueParallelJobs_JOB4]
> [testQueueParallelJobs_JOB3, testQueueParallelJobs_JOB4, testQueueParallelJobs_JOB3]
> [testQueueParallelJobs_JOB4, testQueueParallelJobs_JOB3, testQueueParallelJobs_JOB4]
> [testQueueParallelJobs_JOB3, testQueueParallelJobs_JOB4]
> [testQueueParallelJobs_JOB4, testQueueParallelJobs_JOB4]
> [testQueueParallelJobs_JOB4]
> [testQueueParallelJobs_JOB5]
> [testQueueParallelJobs_JOB6]
> 

We end up having jut JOB4 in the DAG and the reason is because once each job is finished, we add new job to the _readyList (_readyList is updated in finishJob method). If there is only one Job in the _readyList, we will wait until 4 is finished and add JOB5 and so on. In this case, we will not run the whole queue in parallel. 

Also interesting to note here that even before controller swicth here is _readyList which is also wrong and we see two JOBs with the same name in the queue. However, this is not as critical as the the one I explained above:

> [testQueueParallelJobs_JOB0, testQueueParallelJobs_JOB1]
> [testQueueParallelJobs_JOB1]
> [testQueueParallelJobs_JOB0, testQueueParallelJobs_JOB1, testQueueParallelJobs_JOB2]
> [testQueueParallelJobs_JOB1, testQueueParallelJobs_JOB2]
> [testQueueParallelJobs_JOB2]
> [testQueueParallelJobs_JOB3, testQueueParallelJobs_JOB3]
> [testQueueParallelJobs_JOB3]

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
alirezazamani: Sure. I will add that now.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
alirezazamani: Usually in normal cases where we do not have controller switch we don't run this loop multiple times because _lastJob is keeping us at the end for the list. Also if we have controller switch, it will help us in the long run because we do not need to loop through the jobs that has been scheduled before multiple times.

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
alirezazamani: Please note that if we have a lot of jobs in the queue and restart the controller, if we have for example 1000 jobs already in the queue and completed, how many times we have to loop through each job for the first pipeline. It will make the controller very slow (for the fist pipeline) because each job will go to _readyList multiple times.

alirezazamani: With this fix, this is the _readyList after controller switch:
[testQueueParallelJobs_JOB0, testQueueParallelJobs_JOB1, testQueueParallelJobs_JOB2]
[testQueueParallelJobs_JOB1, testQueueParallelJobs_JOB2, testQueueParallelJobs_JOB3]
[testQueueParallelJobs_JOB2, testQueueParallelJobs_JOB3, testQueueParallelJobs_JOB4]
[testQueueParallelJobs_JOB3, testQueueParallelJobs_JOB4, testQueueParallelJobs_JOB5]
[testQueueParallelJobs_JOB4, testQueueParallelJobs_JOB5, testQueueParallelJobs_JOB6]
[testQueueParallelJobs_JOB5, testQueueParallelJobs_JOB6]
[testQueueParallelJobs_JOB6]

alirezazamani: Without the fix:
Please note the we add job to the _readyJob list once we call finishJob (once a job is completed).
Let's say controller switch happened and we started from scratch.
We have [Job0, Job1 , Job2] in the _readyList.
To schedule the jobs, we get next job from readyList (Job0), see that this job is completed, we add the children of Job0 which is Job1. Hence the _readyList will be [Job1, Job2, Job1]. That is why we have same job added to the _readyList.

In conclusion, if a job is finished and controller adds a wrong job, the _readyJobList will fall behind. For example in my test, we add Job4 multiple times. Since Job4 goes to in-progress state and it takes some time to finish, we do not add new job (i.e. job5) until Job4 is finished and hence we do not run jobs in parallel if this happens.

narendly: @alirezazamani Thank for explaining with concrete examples. So the root cause seems to be that we do not do if (!_readyJobList.contains(nextJob)) when we add the jobs to the ready list.

Good find and good fix! :)

alirezazamani:
This PR is ready to be merged, approved by @dasahcc and @narendly 

Fix adding same job multiple times to RuntimeJobDag 

In this commit, a fix has been implemented to avoid
 _readyJobList in RuntimeJobDag to contain multiple
entries of the same job.

Issue #1003, opened 2020-05-13
 Need more accurate error message for resetPartition API


xyuanlu:
resetPartition will return an "instance not alive" exception if the given instance 
1. was part of given cluster but is not alive now, or
2. is not part of this cluster ever.

It would be preferable to have different error messages for these two different cases for better debuggability.

PR #1007, opened 2020-05-13
Add more accurate error message for resetPartition 

xyuanlu:
### Issues

Address #1003

### Description

When user reset one or more particular partition(s) using resetPartition API, Helix will throw an exception if the given instance can not be found under aliveInstance path. This could be caused by either of of the following reasons:
1. The instance was part of given cluster but is not alive now, or
2. The instance is not part of this cluster ever.
This PR differentiate these two causes for better debuggability.

This PR also adds a unit test for resetPartition since there is no existing test for it previously.

### Tests

testResetPartition

> [ERROR] Failures: 
[ERROR]   TestMessagingService.TestMessageSimpleSendReceiveAsync:240
[ERROR]   TestStateTransitionTimeout.testStateTransitionTimeOut:166 expected:<true> but was:<false>
[ERROR]   TestRebalanceRunningTask.testFixedTargetTaskAndDisabledRebalanceAndNodeAdded:297 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1146, Failures: 4, Errors: 0, Skipped: 1
rerun:
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 14.992 s - in org.apache.helix.integration.controller.TestControllerLeadershipChange
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 31.445 s - in org.apache.helix.integration.messaging.TestMessagingService
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4.014 s - in org.apache.helix.integration.paticipant.TestStateTransitionTimeout
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 21.551 s - in org.apache.helix.integration.task.TestRebalanceRunningTask

### Commits

### Documentation (Optional)

- [X] In case of new functionality, my PR adds documentation in the following wiki page:X

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -563,11 +563,11 @@ public class ZKHelixAdmin implements HelixAdmin {
   private enum ResetPartitionFailureReason {
     INSTANCE_NOTALIVE("%s is not alive in cluster %s"),
     INSTANCE_NONEXISTENT("%s does not exist in cluster %s"),
-    RESOURCE_NONEXISTENT("resource %s is not added"),
-    PARTITION_NONEXISTENT("not all %s exist"),
-    PARTITION_NOT_ERROR("%s is NOT found"),
-    STATE_MODEL_NONEXISTENT("%s is NOT found"),
-    PENDING_MSG("a pending message exists: %s");
+    RESOURCE_NONEXISTENT("resource %s is not added to cluster %s"),
+    PARTITION_NONEXISTENT("not all %s exist in cluster %s"),
+    PARTITION_NOT_ERROR("%s is NOT found in cluster %s"),
+    STATE_MODEL_NONEXISTENT("%s is NOT found in cluster %s"),
+    PENDING_MSG("a pending message %s exists for resource %s");
 
     private String message;
 
@@ -575,8 +575,10 @@ public class ZKHelixAdmin implements HelixAdmin {
       this.message = message;
     }
 
-    public String getMessage() {
-      return "Can't reset state for %s.%s on %s, because " + message;
+    public String getMessage(String resourceName, List<String> partitionNames, String instanceName,
+        String arg1, String arg2) {
+      return String.format("Can't reset state for %s.%s on %s, because " + message, resourceName,
+          partitionNames, instanceName, arg1, arg2);
     }
   }
 
@@ -597,16 +599,17 @@ public class ZKHelixAdmin implements HelixAdmin {
       String instanceConfigPath = PropertyPathBuilder.instanceConfig(clusterName, instanceName);
       throw new HelixException(String.format(
           _zkClient.exists(instanceConfigPath) ? ResetPartitionFailureReason.INSTANCE_NOTALIVE
-              .getMessage() : ResetPartitionFailureReason.INSTANCE_NONEXISTENT.getMessage(),
-          resourceName, partitionNames, instanceName, instanceName, clusterName));
+              .getMessage(resourceName, partitionNames, instanceName, instanceName, clusterName)
+              : ResetPartitionFailureReason.INSTANCE_NONEXISTENT
+                  .getMessage(resourceName, partitionNames, instanceName, instanceName,
+                      clusterName)));
     }
 
     // check resource group exists
     IdealState idealState = accessor.getProperty(keyBuilder.idealStates(resourceName));
     if (idealState == null) {
-      throw new HelixException(String
-          .format(ResetPartitionFailureReason.RESOURCE_NONEXISTENT.getMessage(), resourceName,
-              partitionNames, instanceName, resourceName));
+      throw new HelixException(String.format(ResetPartitionFailureReason.RESOURCE_NONEXISTENT
+          .getMessage(resourceName, partitionNames, instanceName, resourceName, clusterName)));
     }
 
     // check partition exists in resource group
@@ -615,9 +618,9 @@ public class ZKHelixAdmin implements HelixAdmin {
         (idealState.getRebalanceMode() == RebalanceMode.CUSTOMIZED) ? idealState.getRecord()
             .getMapFields().keySet() : idealState.getRecord().getListFields().keySet();
     if (!partitions.containsAll(resetPartitionNames)) {
-      throw new HelixException(String
-          .format(ResetPartitionFailureReason.PARTITION_NONEXISTENT.getMessage(), resourceName,
-              partitionNames, instanceName, partitionNames));
+      throw new HelixException(String.format(ResetPartitionFailureReason.PARTITION_NONEXISTENT
+          .getMessage(resourceName, partitionNames, instanceName, partitionNames.toString(),
+              clusterName)));
     }
 
     // check partition is in ERROR state
@@ -626,9 +629,9 @@ public class ZKHelixAdmin implements HelixAdmin {
         accessor.getProperty(keyBuilder.currentState(instanceName, sessionId, resourceName));
     for (String partitionName : resetPartitionNames) {
       if (!curState.getState(partitionName).equals(HelixDefinedState.ERROR.toString())) {
-        throw new HelixException(String
-            .format(ResetPartitionFailureReason.PARTITION_NOT_ERROR.getMessage(), resourceName,
-                partitionNames, instanceName, partitionNames));
+        throw new HelixException(String.format(ResetPartitionFailureReason.PARTITION_NOT_ERROR
+            .getMessage(resourceName, partitionNames, instanceName, partitionNames.toString(),
+                clusterName)));
       }
     }
 
@@ -636,9 +639,8 @@ public class ZKHelixAdmin implements HelixAdmin {
     String stateModelDef = idealState.getStateModelDefRef();
     StateModelDefinition stateModel = accessor.getProperty(keyBuilder.stateModelDef(stateModelDef));
     if (stateModel == null) {
-      throw new HelixException(String
-          .format(ResetPartitionFailureReason.STATE_MODEL_NONEXISTENT.getMessage(), resourceName,
-              partitionNames, instanceName, stateModelDef));
+      throw new HelixException(String.format(ResetPartitionFailureReason.STATE_MODEL_NONEXISTENT
+          .getMessage(resourceName, partitionNames, instanceName, stateModelDef, clusterName)));
     }
 
     // check there is no pending messages for the partitions exist
@@ -650,9 +652,9 @@ public class ZKHelixAdmin implements HelixAdmin {
         continue;
       }
 
-      throw new HelixException(String
-          .format(ResetPartitionFailureReason.PENDING_MSG.getMessage(), resourceName,
-              partitionNames, instanceName, message.toString()));
+      throw new HelixException(String.format(ResetPartitionFailureReason.PENDING_MSG
+          .getMessage(resourceName, partitionNames, instanceName, message.toString(),
+              message.getResourceName())));
     }
 
     String adminName = null;
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -597,11 +597,9 @@ public class ZKHelixAdmin implements HelixAdmin {
       // check if the instance exists in the cluster
       String instanceConfigPath = PropertyPathBuilder.instanceConfig(clusterName, instanceName);
       throw new HelixException(String.format(
-          _zkClient.exists(instanceConfigPath) ? ResetPartitionFailureReason.INSTANCE_NOT_ALIVE
-              .getMessage(resourceName, partitionNames, instanceName, instanceName, clusterName)
-              : ResetPartitionFailureReason.INSTANCE_NON_EXISTENT
-                  .getMessage(resourceName, partitionNames, instanceName, instanceName,
-                      clusterName)));
+          (_zkClient.exists(instanceConfigPath) ? ResetPartitionFailureReason.INSTANCE_NOT_ALIVE
+              : ResetPartitionFailureReason.INSTANCE_NON_EXISTENT)
+              .getMessage(resourceName, partitionNames, instanceName, instanceName, clusterName)));
     }
 
     // check resource group exists
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -573,39 +573,41 @@ public class ZKHelixAdmin implements HelixAdmin {
     // check the instance is alive
     LiveInstance liveInstance = accessor.getProperty(keyBuilder.liveInstance(instanceName));
     if (liveInstance == null) {
-      // check if the instance has ever exist in the cluster
+      // check if the instance has never exist in the cluster
       String instancePath = PropertyPathBuilder.instance(clusterName, instanceName);
+      String errMessage;
       if (!_zkClient.exists(instancePath)) {
-        throw new HelixException(
-            "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-            + ", because " + instanceName + " never existed in cluster " + clusterName);
+        errMessage = String.format(
+            "Can't reset state for %s.%s on %s, because %s has never existed" + " in cluster %s",
+            resourceName, partitionNames, instanceName, instanceName, clusterName);
       } else {
-        throw new HelixException(
-            "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-            + ", because " + instanceName + " is not alive in cluster " + clusterName + " anymore");
+        errMessage = String.format("Can't reset state for %s.%s on %s, because %s does not live in"
+                + " cluster %s anymore", resourceName, partitionNames, instanceName, instanceName,
+            clusterName);
       }
+      throw new HelixException(errMessage);
     }
 
     // check resource group exists
     IdealState idealState = accessor.getProperty(keyBuilder.idealStates(resourceName));
     if (idealState == null) {
-      throw new HelixException(
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because " + resourceName + " is not added");
+      throw new HelixException(String
+          .format("Can't reset state for %s.%s on %s, because %s is not added", resourceName,
+              partitionNames, instanceName, resourceName));
     }
 
     // check partition exists in resource group
     Set<String> resetPartitionNames = new HashSet<String>(partitionNames);
     Set<String> partitions;
     if (idealState.getRebalanceMode() == RebalanceMode.CUSTOMIZED) {
-       partitions = new HashSet<String>(idealState.getRecord().getMapFields().keySet());
+      partitions = idealState.getRecord().getMapFields().keySet();
     } else {
-      partitions = new HashSet<String>(idealState.getRecord().getListFields().keySet());
+      partitions = idealState.getRecord().getListFields().keySet();
     }
     if (!partitions.containsAll(resetPartitionNames)) {
-      throw new HelixException(
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because not all " + partitionNames + " exist");
+      throw new HelixException(String
+          .format("Can't reset state for %s.%s on %s, because not all %s exist", resourceName,
+              partitionNames, instanceName, partitionNames));
     }
 
     // check partition is in ERROR state
@@ -614,9 +616,9 @@ public class ZKHelixAdmin implements HelixAdmin {
         accessor.getProperty(keyBuilder.currentState(instanceName, sessionId, resourceName));
     for (String partitionName : resetPartitionNames) {
       if (!curState.getState(partitionName).equals(HelixDefinedState.ERROR.toString())) {
-        throw new HelixException(
-            "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-                + ", because not all " + partitionNames + " are in ERROR state");
+        throw new HelixException(String
+            .format("Can't reset state for %s.%s on %s, because not all %s are in ERROR state",
+                resourceName, partitionNames, partitionNames));
       }
     }
 
@@ -624,9 +626,9 @@ public class ZKHelixAdmin implements HelixAdmin {
     String stateModelDef = idealState.getStateModelDefRef();
     StateModelDefinition stateModel = accessor.getProperty(keyBuilder.stateModelDef(stateModelDef));
     if (stateModel == null) {
-      throw new HelixException(
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because " + stateModelDef + " is NOT found");
+      throw new HelixException(String
+          .format("Can't reset state for %s.%s on %s, because %s is NOT found", resourceName,
+              partitionNames, stateModelDef));
     }
 
     // check there is no pending messages for the partitions exist
@@ -638,9 +640,9 @@ public class ZKHelixAdmin implements HelixAdmin {
         continue;
       }
 
-      throw new HelixException(
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because a pending message exists: " + message);
+      throw new HelixException(String
+          .format("Can't reset state for %s.%s on %s, because a pending message exists: %s",
+              resourceName, partitionNames, message));
     }
 
     String adminName = null;
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -566,8 +566,7 @@ public class ZKHelixAdmin implements HelixAdmin {
     RESOURCE_NON_EXISTENT("resource %s is not added to cluster %s"),
     PARTITION_NON_EXISTENT("not all %s exist in cluster %s"),
     PARTITION_NOT_ERROR("%s is NOT found in cluster %s"),
-    STATE_MODEL_NON_EXISTENT("%s is NOT found in cluster %s"),
-    PENDING_MSG("a pending message %s exists for resource %s");
+    STATE_MODEL_NON_EXISTENT("%s is NOT found in cluster %s");
 
     private String message;
 
@@ -652,9 +651,10 @@ public class ZKHelixAdmin implements HelixAdmin {
         continue;
       }
 
-      throw new HelixException(String.format(ResetPartitionFailureReason.PENDING_MSG
-          .getMessage(resourceName, partitionNames, instanceName, message.toString(),
-              message.getResourceName())));
+      throw new HelixException(String.format(
+          "Can't reset state for %s.%s on %s, because a pending message %s exists for resource %s",
+          resourceName, partitionNames, instanceName, message.toString(),
+          message.getResourceName()));
     }
 
     String adminName = null;
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -575,11 +575,8 @@ public class ZKHelixAdmin implements HelixAdmin {
       this.message = message;
     }
 
-    public String getMessage(String instanceName, String resourceName,
-        List<String> partitionNames) {
-      return String
-          .format("Can't reset state for %s.%s on %s, because ", resourceName, partitionNames,
-              instanceName) + message;
+    public String getMessage() {
+      return "Can't reset state for %s.%s on %s, because " + message;
     }
   }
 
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -588,20 +596,16 @@ public void resetPartition(String clusterName, String instanceName, String resou
...
     Set<String> resetPartitionNames = new HashSet<String>(partitionNames);
+    Set<String> partitions;
     if (idealState.getRebalanceMode() == RebalanceMode.CUSTOMIZED) {
-      Set<String> partitions = new HashSet<String>(idealState.getRecord().getMapFields().keySet());
-      if (!partitions.containsAll(resetPartitionNames)) {
-        throw new HelixException(
-            "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-                + ", because not all " + partitionNames + " exist");
-      }
+       partitions = new HashSet<String>(idealState.getRecord().getMapFields().keySet());
```
pkuwm, line 601: Suggest diamond `new HashSet<>()`
And it seems the partitions are only used for next step `containsAll()`. Would it be fine just use the reference to keySet() `partitions = Address unclean helix controller leadership switch`, instead of making a copy?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,93 @@ public void testLegacyEnableDisablePartition() {
...
+      // This exception is expected because the instance is not alive.
+      Assert.assertEquals(expected.getMessage(),
+          "Can't reset state for " + testResource  + "/[1, 2] on " + instanceName
+              + ", because " + instanceName + " is not alive in cluster " + clusterName + " anymore");
+    }
+
+    HelixManager manager = initializeHelixManager(clusterName, instanceConfig.getInstanceName());
+    try {
+      manager.connect();
+    } catch (Exception e) {
```
pkuwm, line 606: No need to catch this exception. If an exception is thrown from connect(), the test will throw this exception and fail.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,9 +573,17 @@ public void resetPartition(String clusterName, String instanceName, String resou
     // check the instance is alive
     LiveInstance liveInstance = accessor.getProperty(keyBuilder.liveInstance(instanceName));
     if (liveInstance == null) {
-      throw new HelixException(
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because " + instanceName + " is not alive");
+      // check if the instance has ever exist in the cluster
```
narendly, line 576: "has ever existed"

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,9 +573,17 @@ public void resetPartition(String clusterName, String instanceName, String resou
...
+      String instancePath = PropertyPathBuilder.instance(clusterName, instanceName);
+      if (!_zkClient.exists(instancePath)) {
+        throw new HelixException(
+            "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
+            + ", because " + instanceName + " never existed in cluster " + clusterName);
+      } else {
+        throw new HelixException(
+            "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
+            + ", because " + instanceName + " is not alive in cluster " + clusterName + " anymore");
```
narendly, line 585: "not live"

narendly, line 585: Can you just create the error message in the if-else block and throw only once to avoid code duplication?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,93 @@ public void testLegacyEnableDisablePartition() {
...
+    System.out.println("START " + clusterName + " at " + new Date(System.currentTimeMillis()));
+    HelixAdmin admin = new ZKHelixAdmin(_gZkClient);
+    admin.addCluster(clusterName, true);
+    admin.addInstance(clusterName, new InstanceConfig(instanceName));
+    admin.enableInstance(clusterName, instanceName, true);
+
+    InstanceConfig instanceConfig = admin.getInstanceConfig(clusterName, instanceName);
+
+    // Test the sanity check for resetPartition.
+    // resetPartition is expected to throw an exception when provided with a non-exist instance.
```
narendly, line 579: "nonexistent"

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,93 @@ public void testLegacyEnableDisablePartition() {
...
+          Arrays.asList("1", "2"));
+      Assert.fail("Should throw HelixException");
+    } catch (HelixException expected) {
+      // This exception is expected because the instance name is made up.
+      Assert.assertEquals(expected.getMessage(),
+          "Can't reset state for " + testResource  + "/[1, 2] on WrongTestInstance," +
+              " because WrongTestInstance never existed in cluster " + clusterName);
+    }
+
+    // resetPartition is expected to throw an exception when provided with a non-alive instance.
```
narendly, line 591: non-live

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,93 @@ public void testLegacyEnableDisablePartition() {
...
+    try {
+      admin.resetPartition(clusterName, instanceName, testResource,
+          Arrays.asList("1", "2"));
+      Assert.fail("Should throw HelixException");
+    } catch (HelixException expected) {
+      // This exception is expected because the instance is not alive.
+      Assert.assertEquals(expected.getMessage(),
+          "Can't reset state for " + testResource  + "/[1, 2] on " + instanceName
+              + ", because " + instanceName + " is not alive in cluster " + clusterName + " anymore");
```
narendly, line 600: not live

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
xyuanlu: TFTR. Will fix.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
xyuanlu: TFTR. Will fix.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,9 +573,17 @@ public void resetPartition(String clusterName, String instanceName, String resou
...
     LiveInstance liveInstance = accessor.getProperty(keyBuilder.liveInstance(instanceName));
     if (liveInstance == null) {
-      throw new HelixException(
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because " + instanceName + " is not alive");
+      // check if the instance has ever exist in the cluster
+      String instancePath = PropertyPathBuilder.instance(clusterName, instanceName);
+      if (!_zkClient.exists(instancePath)) {
+        throw new HelixException(
+            "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
```
jiajunwang, line 580: Let's use "." instead of "/" to connect the resource name and partition name?
And use String.format() to make the code cleaner.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
xyuanlu: I think I need to catch it so that compile won't fail.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,35 +573,41 @@ public void resetPartition(String clusterName, String instanceName, String resou
...
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because " + resourceName + " is not added");
+      throw new HelixException(String
+          .format("Can't reset state for %s.%s on %s, because %s is not added", resourceName,
+              partitionNames, instanceName, resourceName));
     }
 
     // check partition exists in resource group
     Set<String> resetPartitionNames = new HashSet<String>(partitionNames);
+    Set<String> partitions;
```
junlong-gao, line 601: Looks like either branch of the below if/else needs this, might as well just init `partitions` here?

Also, since the original code makes a copy, better keep the original semantics here to prevent accidental mutation.

- Junlong
: )

junlong-gao: Looks like there is a slight difference `getMapFields` vs `getListFields` requires this if-else, so perhaps a ternary operator or the current if-else is already unavoidable.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,35 +573,41 @@ public void resetPartition(String clusterName, String instanceName, String resou
...
-              + ", because " + instanceName + " is not alive");
+      // check if the instance has never exist in the cluster
+      String instancePath = PropertyPathBuilder.instance(clusterName, instanceName);
+      String errMessage;
+      if (!_zkClient.exists(instancePath)) {
+        errMessage = String.format(
+            "Can't reset state for %s.%s on %s, because %s has never existed" + " in cluster %s",
+            resourceName, partitionNames, instanceName, instanceName, clusterName);
+      } else {
+        errMessage = String.format("Can't reset state for %s.%s on %s, because %s does not live in"
```
narendly, line 584: does not -> is not

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,82 @@ public void testLegacyEnableDisablePartition() {
...
+    } catch (HelixException expected) {
+      // This exception is expected because partitions do not exist.
+      Assert.assertEquals(expected.getMessage(), String
+          .format("Can't reset state for %s.[1, 2] on %s, because not all [1, 2] exist",
+              testResource, instanceName));
+    }
+
+    // clean up
+    manager.disconnect();
+    admin.dropCluster(clusterName);
```
narendly, line 635: Please add a TestHelper.verify() to verify that the cluster has indeed been fully deleted at this point.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,82 @@ public void testLegacyEnableDisablePartition() {
...
+    System.out.println("START " + clusterName + " at " + new Date(System.currentTimeMillis()));
+    HelixAdmin admin = new ZKHelixAdmin(_gZkClient);
+    admin.addCluster(clusterName, true);
+    admin.addInstance(clusterName, new InstanceConfig(instanceName));
+    admin.enableInstance(clusterName, instanceName, true);
+
+    InstanceConfig instanceConfig = admin.getInstanceConfig(clusterName, instanceName);
+
+    // Test the sanity check for resetPartition.
+    // resetPartition is expected to throw an exception when provided with a nonexist instance.
```
narendly, line 579: nonexistent

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,82 @@ public void testLegacyEnableDisablePartition() {
...
+    try {
+      admin.resetPartition(clusterName, "WrongTestInstance", testResource, Arrays.asList("1", "2"));
```
narendly, line 581: Make "WrongTestInstance" a string constant throughout?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,82 @@ public void testLegacyEnableDisablePartition() {
...
+      // This exception is expected because the instance is not alive.
+      Assert.assertEquals(expected.getMessage(), String.format(
+          "Can't reset state for %s.[1, 2] on %s, because %s does not live in cluster %s anymore",
+          testResource, instanceName, instanceName, clusterName));
+    }
+
+    HelixManager manager = initializeHelixManager(clusterName, instanceConfig.getInstanceName());
+    manager.connect();
+
+    // resetPartition is expected to throw an exception when provided with a non-exist resource.
```
narendly, line 604: nonexistent

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,82 @@ public void testLegacyEnableDisablePartition() {
...
+    }
+
+    IdealState idealState = new IdealState(testResource);
+    idealState.setNumPartitions(3);
+    admin.addStateModelDef(clusterName, "MasterSlave", new MasterSlaveSMD());
+    idealState.setStateModelDefRef("MasterSlave");
+    idealState.setRebalanceMode(IdealState.RebalanceMode.FULL_AUTO);
+    admin.addResource(clusterName, testResource, idealState);
+
+    admin.enableResource(clusterName, testResource, true);
```
narendly, line 622: Do you need to add a verifier check here? It might take some time for the controller to rebalance.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,82 @@ public void testLegacyEnableDisablePartition() {
...
+    String className = TestHelper.getTestClassName();
+    String methodName = TestHelper.getTestMethodName();
+    String clusterName = className + "_" + methodName;
+    String instanceName = "TestInstance";
+    String testResource = "TestResource";
+    System.out.println("START " + clusterName + " at " + new Date(System.currentTimeMillis()));
+    HelixAdmin admin = new ZKHelixAdmin(_gZkClient);
+    admin.addCluster(clusterName, true);
+    admin.addInstance(clusterName, new InstanceConfig(instanceName));
+    admin.enableInstance(clusterName, instanceName, true);
```
narendly, line 574: Do we need a verifier check after this?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,35 +573,41 @@ public void resetPartition(String clusterName, String instanceName, String resou
...
     if (liveInstance == null) {
-      throw new HelixException(
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because " + instanceName + " is not alive");
+      // check if the instance has never exist in the cluster
+      String instancePath = PropertyPathBuilder.instance(clusterName, instanceName);
+      String errMessage;
+      if (!_zkClient.exists(instancePath)) {
+        errMessage = String.format(
+            "Can't reset state for %s.%s on %s, because %s has never existed" + " in cluster %s",
```
jiajunwang, line 581: Why there is still a "+", it is not necessary, right?

jiajunwang: BTW, if this is for the format, just use the auto format supported by the IDE after you import the Helix code style.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,35 +573,41 @@ public void resetPartition(String clusterName, String instanceName, String resou
...
+            resourceName, partitionNames, instanceName, instanceName, clusterName);
+      } else {
+        errMessage = String.format("Can't reset state for %s.%s on %s, because %s does not live in"
+                + " cluster %s anymore", resourceName, partitionNames, instanceName, instanceName,
```
jiajunwang, line 585: The check only shows it is not alive for now. "anymore" is not accurate, we can remove it.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,35 +573,41 @@ public void resetPartition(String clusterName, String instanceName, String resou
     // check the instance is alive
     LiveInstance liveInstance = accessor.getProperty(keyBuilder.liveInstance(instanceName));
     if (liveInstance == null) {
-      throw new HelixException(
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because " + instanceName + " is not alive");
+      // check if the instance has never exist in the cluster
+      String instancePath = PropertyPathBuilder.instance(clusterName, instanceName);
+      String errMessage;
```
jiajunwang, line 578: nit, You can init the errMessage like String.format( "Can't reset state for %s.%s on %s,", resourceName, partitionNames, instanceName), then in the different branches, append more content.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
jiajunwang: nit, you could use (idealState.getRebalanceMode() == RebalanceMode.CUSTOMIZED)? idealState.getRecord().getMapFields().keySet() : partitions = idealState.getRecord().getListFields().keySet() to simplify it a little bit.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,35 +573,41 @@ public void resetPartition(String clusterName, String instanceName, String resou
...
-      if (!partitions.containsAll(resetPartitionNames)) {
-        throw new HelixException(
-            "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-                + ", because not all " + partitionNames + " exist");
-      }
+      partitions = idealState.getRecord().getListFields().keySet();
+    }
+    if (!partitions.containsAll(resetPartitionNames)) {
+      throw new HelixException(String
+          .format("Can't reset state for %s.%s on %s, because not all %s exist", resourceName,
```
jiajunwang, line 609: It looks like there are many error messages look like this, I think we can have a private method to generate the formated messaged based on the input, the parameters could be: partitionNames, instanceName, failure reason (so the callers just have different logic of generating the different reasons).

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
xyuanlu: Please correct me if I am wrong. In my understanding, if the instance has ever lived in the cluster, we would be able to find it under /instance Path but not under the /liveInstance ?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
xyuanlu: The auto formatter collapse 2 lines into 1 line and leave the "+" here.  I missed it.  Will fix

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
xyuanlu: TFTR. May I ask are you suggesting verify if the resource is enabled? In my understanding, if not, then we won't get the expected exception for nonexistent partitions in the next try/catch block.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
xyuanlu: Same as above. If the instance is not corrected added and enabled, we wont get all following expected exceptions.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
xyuanlu: TFTR. Will fix.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
xyuanlu: Please correct me if I am wrong. For example, I guess if we get "resource not added" exception while "nonexistent partition" is expected, we would know the resource is not correctly added.

xyuanlu:
Thanks for the review! There are still some issue/comments open to further discuss. I will address them in the following commits.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -560,6 +560,51 @@ private void processMaintenanceMode(String clusterName, final boolean enabled,
...
+      String resourceName, List<String> partitionNames, String strArg,
+      ResetPartitionFailureReason reason) {
+    String errMessage = String
+        .format("Can't reset state for %s.%s on %s, because ", resourceName, partitionNames,
+            instanceName);
+
+    switch (reason) {
+      case INSTANCE_NOTALIVE:
+        errMessage +=
+            String.format("%s has never existed in cluster %s", instanceName, clusterName);
```
jiajunwang, line 583: nit, this is about style.
In Java, you can put fields into enum items. This format string can be a field of the ResetPartitionFailureReason enum item. Then you don't need the switch here.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -560,6 +560,82 @@ public void testLegacyEnableDisablePartition() {
...
+
+    InstanceConfig instanceConfig = admin.getInstanceConfig(clusterName, instanceName);
+
+    // Test the sanity check for resetPartition.
+    // resetPartition is expected to throw an exception when provided with a nonexistent instance.
+    try {
+      admin.resetPartition(clusterName, wrongTestInstance, testResource, Arrays.asList("1", "2"));
```
jiajunwang, line 582: This test case works but not very clean. Since the testResource has not been created, the resetPartition may fail due to other reasons. So the test logic here might be impacted if we change that logic. For example, we check the resource existence first in the future.

To avoid that impact, let's have the resource created and then test reset.
BTW, in that case, you will need to use the verifier to wait until resource assignment has been done and state converged.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,35 +618,36 @@ public void resetPartition(String clusterName, String instanceName, String resou
...
+      // check if the instance has never exist in the cluster
+      String instancePath = PropertyPathBuilder.instance(clusterName, instanceName);
+      if (!_zkClient.exists(instancePath)) {
+        throw new HelixException(
+            resetPartitionLogHelper(clusterName, instanceName, resourceName, partitionNames, null,
+                ResetPartitionFailureReason.INSTANCE_NOTALIVE));
+      } else {
+        throw new HelixException(
+            resetPartitionLogHelper(clusterName, instanceName, resourceName, partitionNames, null,
+                ResetPartitionFailureReason.INSTANCE_NONEXISTENT));
```
pkuwm, line 630: Nit, just about style :) To reduce the duplicate code, I would do something like this
```
String failureReason = _zkClient.exists(instancePath) ? ResetPartitionFailureReason.INSTANCE_NONEXISTENT) :  ResetPartitionFailureReason.INSTANCE_NOTALIVE);
throw new HelixException(...);
```

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -560,6 +560,29 @@ private void processMaintenanceMode(String clusterName, final boolean enabled,
...
+    STATE_MODEL_NONEXISTENT("%s is NOT found"),
+    PENDING_MSG("a pending message exists: %s");
+
+    private String message;
+
+    ResetPartitionFailureReason(String message) {
+      this.message = message;
+    }
+
+    public String getMessage(String instanceName, String resourceName,
```
jiajunwang, line 578: The method should either return a formated message or a pure format string. The current design is mixing and would be hard to use.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,35 +596,31 @@ public void resetPartition(String clusterName, String instanceName, String resou
...
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because " + instanceName + " is not alive");
+      // check if the instance has added to the cluster
+      String instanceConfigPath = PropertyPathBuilder.instanceConfig(clusterName, instanceName);
+      throw new HelixException(String.format(
+          _zkClient.exists(instanceConfigPath) ? ResetPartitionFailureReason.INSTANCE_NOTALIVE
+              .getMessage(instanceName, resourceName, partitionNames)
+              : ResetPartitionFailureReason.INSTANCE_NONEXISTENT
+                  .getMessage(instanceName, resourceName, partitionNames), instanceName,
+          clusterName));
```
jiajunwang, line 606: I would suggest passing the cluster name to getMessage, and let it return the fully formated string.

jiajunwang:
BTW, please resolve the conversation if the issue has been resolved.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
xyuanlu: Some messages require 2 args an others need one arg. Please correct me if I am wrong but I think it is hard to return a fully formated string if we do not add extra logic here.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
xyuanlu: This is a unit test for sanity check. There is no running controller.  We have end-to-end tests for resetPartition() under webapp/TestResetPartitionState and integration/TestResetPartitionState.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
narendly: Could you add a verify() to make sure the cluster has been removed successfully?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -560,6 +560,28 @@ private void processMaintenanceMode(String clusterName, final boolean enabled,
...
   }
 
+  private enum ResetPartitionFailureReason {
+    INSTANCE_NOT_ALIVE("%s is not alive in cluster %s"),
+    INSTANCE_NON_EXISTENT("%s does not exist in cluster %s"),
+    RESOURCE_NON_EXISTENT("resource %s is not added to cluster %s"),
+    PARTITION_NON_EXISTENT("not all %s exist in cluster %s"),
+    PARTITION_NOT_ERROR("%s is NOT found in cluster %s"),
+    STATE_MODEL_NON_EXISTENT("%s is NOT found in cluster %s"),
+    PENDING_MSG("a pending message %s exists for resource %s");
```
jiajunwang, line 570: Let's don't bother including PENDING_MSG here. This is not compatible with the other types. And if there is no other better way, we should not be blocked on this.

Just leave it as it originally be.

xyuanlu: Updated. THX

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -573,35 +594,32 @@ public void resetPartition(String clusterName, String instanceName, String resou
     // check the instance is alive
     LiveInstance liveInstance = accessor.getProperty(keyBuilder.liveInstance(instanceName));
     if (liveInstance == null) {
-      throw new HelixException(
-          "Can't reset state for " + resourceName + "/" + partitionNames + " on " + instanceName
-              + ", because " + instanceName + " is not alive");
+      // check if the instance exists in the cluster
+      String instanceConfigPath = PropertyPathBuilder.instanceConfig(clusterName, instanceName);
+      throw new HelixException(String.format(
+          _zkClient.exists(instanceConfigPath) ? ResetPartitionFailureReason.INSTANCE_NOT_ALIVE
```
jiajunwang, line 600: nit, but will the following code be easier:

(_zkClient.exists(instanceConfigPath) ? ResetPartitionFailureReason.INSTANCE_NOT_ALIVE : ResetPartitionFailureReason.INSTANCE_NON_EXISTENT)
              .getMessage(resourceName, partitionNames, instanceName, instanceName, clusterName);

xyuanlu:
This PR is ready to be merged, approved by @jiajunwang .

xyuanlu:

Add more accurate error message for resetPartition

This commit adds more specific messages for failure reason for reserPartion. It also adds a unit test for resetPartition.

PR #1019, opened 2020-05-19
ConnectionTimeout change to ensure HelixManager initial connection

### LINKED ISSUE #1018: ConnectionTime change to ensure ZkHelixManager can connect to Zookeeper when some quorum node fails
Zookeeper server quorum is health as long as the majority quorum nodes are health. One failing node or observer should not block HelixManager from establish a session to zookeeper. 

Different Zookeeper client has different behavior in terms of establishing the first session with ZooKeeper quorum.  The following table is a summary, assuming only one DNS name is supplied.

Zk Version | TCP connection timeout | Prob retry same server | Resolve DNS between connection within session
-----------| -------------------------|------------------------|-----------------------------------
< 3.4.13 | sessionTimeout / # IPs resolved from ZK connection string |0%; unless there is only 1 IP resolved from the ZK connection string | No
 3.4.13+ | sessionTimeout | 1/N; N is the count of resolved IP addresses from the ZK connection string | Yes

Assuming a typical 3.4.13+ client, 5 quorum node, 1 node be unresponsive, we would like less than 1% initial connection failure to unhealthy server, we need DEFAULT_CONNECTION_TIMEOUT to be 4 times of DEFAULT_SESSION_TIMEOUT.

kaisun2000:
### Issues

(fix #1018 )

### Description

- [x] Here are some details about my PR:

ConnectionTimeout change to ensure HelixManager initial connection.

Ensure HelixManager initial session establishment to
ZooKeeper server would be highly like to succeed given one quorum
node unresponsive. This is done by make default connection timeout
value 4 times of default session timeout value.

### Tests

running, will add later.

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/ZkClient.java
@@ -62,7 +62,10 @@ public class ZkClient extends org.apache.helix.zookeeper.zkclient.ZkClient imple
   private static Logger LOG = LoggerFactory.getLogger(ZkClient.class);
 
   public static final int DEFAULT_OPERATION_TIMEOUT = Integer.MAX_VALUE;
-  public static final int DEFAULT_CONNECTION_TIMEOUT = 60 * 1000;
+  // Four times of DEFAULT_SESSION_TIMEOUT would give less 1/125 chance of initial connection
+  // to Zookeeper failure assuming one observer in the 5 observer Zookeeper Quorum is down
+  // with one DNS name configured as connection string with 3.4.13+ Zookeeper client version
+  public static final int DEFAULT_CONNECTION_TIMEOUT = 120 * 1000;
   public static final int DEFAULT_SESSION_TIMEOUT = 30 * 1000;
 
   /**


zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/ZkClient.java
```diff
@@ -62,7 +62,10 @@
   private static Logger LOG = LoggerFactory.getLogger(ZkClient.class);
 
   public static final int DEFAULT_OPERATION_TIMEOUT = Integer.MAX_VALUE;
-  public static final int DEFAULT_CONNECTION_TIMEOUT = 60 * 1000;
+  // Four times of DEFAULT_SESSION_TIMEOUT would give less 1/125 chance of initial connection
+  // to Zookeeper failure assuming one observer in the 5 observer Zookeeper Quorum is down
```
jiajunwang, line 66: 1. Please provide the formula instead of one set of input and result in the comment. Not all Helix deployment connects to a ZK which has 5 zk servers.
2. If we need 4 times of the session timeout, why hardcode here?
3. Since this configuration depends on the zk server count, I would prefer to specify this in the ZK_CONNECTION_TIMEOUT system property instead of modifying the default connection values. Note that this change might impact all the Helix users that do not specify the ZK_CONNECTION_TIMEOUT, and too long timeout may also cause problems.

kaisun2000: For 3, indeed changing the property is a way. 

The fundamental problem is that if you want to fail fast or not. 

In helix case, only ZkHelixManager uses CONNECTION_TIMEOUT in connect(). If connect() does not succeed, it would throw exceptions and the application won't proceed. 

So it seems there is no downside to do the default change too.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/ZkClient.java
```diff
@@ -62,7 +62,10 @@
...
+  // with one DNS name configured as connection string with 3.4.13+ Zookeeper client version
+  public static final int DEFAULT_CONNECTION_TIMEOUT = 120 * 1000;
```
dasahcc, line 68: We already have some customers only onboard the zookeeper-api module and not for Helix. My suggestion is not change the default behavior but let it to be config change.

kaisun2000:
@jiajunwang, @dasahcc, considering that user can configure java property ZK_CONNECTION_TIMEOUT ( "zk.connection.timeout"), then, let us recommend to change this value instead of changing the DEFAULT_CONNECTION_TIMEOUT. 

Put it another way, let us abandon this pull for now.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/ZkClient.java
narendly: @kaisun2000 

I also think that we should leave the constant as is and make it configurable in the system config.

Note that the HelixManager usage is _**the one you found**_. This is a public constant, and you don't know if it's being used elsewhere. So there is an issue of backward-compatibility if you change the value of the constant.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/ZkClient.java
narendly: +1. This is a public constant, so we need to avoid changing its value.

jiajunwang:
Close the PR on behave of @kaisun2000

PR #1031, opened 2020-05-27
Add getIdealAssignmentForWagedFullAuto in HelixUtil for WAGED rebalancer

### LINKED ISSUE #1030: Support getIdealAssignmentForWagedFullAuto() in HelixUtil
We provide a util method called getIdealAssignmentForFullAuto() so that the user can easily predict how Helix will rebalance using the full-auto rebalance mode.

However, we lack an equivalent method for the WAGED rebalancer.

This issue tracks the work of adding a method, getIdealAssignmentForWagedFullAuto() that returns to the user the cluster-wide assignment result obtained from running a rebalance using WAGED.

narendly:
### Issues

Resolves #1030 

### Description

This commit adds a method, getIdealAssignmentForWagedFullAuto() in HelixUtil that returns to the user the cluster-wide assignment result obtained from running a rebalance using WAGED. The user will be able to use this method to predict how Helix will be rebalancing resources using the WAGED rebalancer.

### Tests

testRebalanceTool

[ERROR] Failures: 
[ERROR]   TestEnableCompression.testEnableCompressionResource:116 Â» ThreadTimeout Method...
[ERROR]   TestDeleteJobFromJobQueue.testForceDeleteJobFromJobQueue:75 Â» Helix Failed to ...
[INFO] 
[ERROR] Tests run: 1148, Failures: 3, Errors: 0, Skipped: 0

Individual runs of failed tests pass.

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
@@ -1,0 +1,88 @@
+package org.apache.helix.controller.rebalancer.waged;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import org.apache.helix.HelixRebalanceException;
+import org.apache.helix.controller.dataproviders.ResourceControllerDataProvider;
+import org.apache.helix.controller.rebalancer.waged.constraints.ConstraintBasedAlgorithmFactory;
+import org.apache.helix.controller.stages.CurrentStateOutput;
+import org.apache.helix.manager.zk.ZkBucketDataAccessor;
+import org.apache.helix.model.ClusterConfig;
+import org.apache.helix.model.Resource;
+import org.apache.helix.model.ResourceAssignment;
+
+
+/**
+ * This rebalancer is a version of WagedRebalancer that only reads the existing assignment metadata
+ * to compute the best possible assignment but never writes back the resulting assignment metadata
+ * from global or partial rebalance. It does so by using a modified version of
+ * AssignmentMetadataStore, ReadOnlyAssignmentMetadataStore.
+ *
+ * This class is to be used in the cluster verifiers, tests, or util methods.
+ */
+public class ReadOnlyWagedRebalancer extends WagedRebalancer {
+  public ReadOnlyWagedRebalancer(String metadataStoreAddress, String clusterName,
+      Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preferences) {
+    super(new ReadOnlyAssignmentMetadataStore(metadataStoreAddress, clusterName),
+        ConstraintBasedAlgorithmFactory.getInstance(preferences), Optional.empty());
+  }
+
+  @Override
+  protected Map<String, ResourceAssignment> computeBestPossibleAssignment(
+      ResourceControllerDataProvider clusterData, Map<String, Resource> resourceMap,
+      Set<String> activeNodes, CurrentStateOutput currentStateOutput, RebalanceAlgorithm algorithm)
+      throws HelixRebalanceException {
+    return getBestPossibleAssignment(getAssignmentMetadataStore(), currentStateOutput,
+        resourceMap.keySet());
+  }
+
+  private static class ReadOnlyAssignmentMetadataStore extends AssignmentMetadataStore {
+    ReadOnlyAssignmentMetadataStore(String metadataStoreAddress, String clusterName) {
+      super(new ZkBucketDataAccessor(metadataStoreAddress), clusterName);
+    }
+
+    @Override
+    public boolean persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
+      // If baseline hasn't changed, skip updating the metadata store
+      if (compareAssignments(_globalBaseline, globalBaseline)) {
+        return false;
+      }
+      // Update the in-memory reference only
+      _globalBaseline = globalBaseline;
+      return true;
+    }
+
+    @Override
+    public boolean persistBestPossibleAssignment(
+        Map<String, ResourceAssignment> bestPossibleAssignment) {
+      // If bestPossibleAssignment hasn't changed, skip updating the metadata store
+      if (compareAssignments(_bestPossibleAssignment, bestPossibleAssignment)) {
+        return false;
+      }
+      // Update the in-memory reference only
+      _bestPossibleAssignment = bestPossibleAssignment;
+      return true;
+    }
+  }
+}
--- a/helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
@@ -28,22 +28,42 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import com.google.common.base.Joiner;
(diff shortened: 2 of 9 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
```diff
@@ -0,0 +1,88 @@
+package org.apache.helix.controller.rebalancer.waged;
```
jiajunwang, line 1: Let's put this in the util or tool package? It should not be used as a rebalancer.

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
```diff
@@ -140,6 +160,81 @@ public static String serializeByComma(List<String> objects) {
...
+   * calculated using the read-only WAGED rebalancer.
+   * @param metadataStoreAddress
+   * @param clusterConfig
+   * @param instanceConfigs
+   * @param liveInstances
+   * @param newIdealStates
+   * @param newResourceConfigs
+   * @return
+   */
+  public static Map<String, ResourceAssignment> getIdealAssignmentForWagedFullAuto(
```
jiajunwang, line 174: I feel RebalanceUtil is a better placement, although the existing method is already in the HelixUtil. What do you think?

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
```diff
@@ -140,6 +160,81 @@ public static String serializeByComma(List<String> objects) {
...
+    HelixDataAccessor helixDataAccessor =
+        new ZKHelixDataAccessor(clusterConfig.getClusterName(), baseDataAccessor);
+
+    // Obtain a refreshed dataProvider (cache) and overwrite cluster parameters with the given parameters
+    ResourceControllerDataProvider dataProvider =
+        new ResourceControllerDataProvider(clusterConfig.getClusterName());
+    dataProvider.requireFullRefresh();
+    dataProvider.refresh(helixDataAccessor);
+    dataProvider.setClusterConfig(clusterConfig);
+    dataProvider.setInstanceConfigMap(instanceConfigs.stream()
```
jiajunwang, line 189: The WAGED rebalancer considers all resources, so if the "newResourceConfigs" only contains user-specified items and we overwrite the cached map with this input map, it may return a different result. The other list/map fields have a similar concern.
1. Can we do merge instead of overwriting? This serves for the users who want to add or modify some items.
2. How to handle requests to remove some items? I didn't have the answer yet. Will update if I have a good idea.

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
```diff
@@ -140,6 +160,81 @@ public static String serializeByComma(List<String> objects) {
...
+    try {
+      // Run the required stages to obtain the BestPossibleOutput
+      RebalanceUtil.runStage(event, new ResourceComputationStage());
+      RebalanceUtil.runStage(event, new CurrentStateComputationStage());
+      RebalanceUtil.runStage(event, new BestPossibleStateCalcStage());
+    } catch (Exception e) {
+      LOG.error("getIdealAssignmentForWagedFullAuto(): Failed to compute ResourceAssignments!", e);
+    } finally {
+      // Close all ZK connections
+      baseDataAccessor.close();
```
jiajunwang, line 217: If the process fails before runStage, for example, while refreshing the cache, this accessor connection will be leaked.

helix-core/src/main/java/org/apache/helix/util/RebalanceUtil.java
```diff
@@ -164,4 +167,25 @@ public static void scheduleOnDemandPipeline(String clusterName, long delay,
...
+   * An example usage is as follows:
+   *       runStage(event, new ResourceComputationStage());
+   *       runStage(event, new CurrentStateComputationStage());
+   *       runStage(event, new BestPossibleStateCalcStage());
+   * By running these stages, we are able to obtain BestPossibleStateOutput in the event object.
+   * @param event
+   * @param stage
+   * @throws Exception
+   */
+  public static void runStage(ClusterEvent event, Stage stage) throws Exception {
```
jiajunwang, line 184: It might be better if we have a runStage**s** method as util. It would be easier to use, and more generic.
Of course, we will have to refactor calcBestPossState() in the verifier to make the change.

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
dasahcc: If it will be exposed to customers, let's put it HelixUtil. It is better to have some centralized place for user using that.

narendly: Already reviewed that option, but if we were to place this in RebalanceUtil, we should move the existing function there too, but that would cause a backward-incompatible change. Let's keep it in here for consistency.

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
narendly: @jiajunwang 

I've considered the case you mentioned, and what we have here is I believe the right thing to do. We should overwrite completely (replace), that way, we could allow users to keep, modify, or remove certain items in the Collection. Users still have the ability to retrieve the existing ResourceConfigs in the cluster, so that's not a big concern.

Moreover, this fits nicely with the API methods already provided in the DataProvider interface. If we want to support merge or remove, that would require further change in the DataProvider interface, which is undesirable because we'd be making API changes for the sake of having a util method or for testing.

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
narendly: It shouldn't fail, but I can wrap the entire logic in a try-catch as well.

helix-core/src/main/java/org/apache/helix/util/RebalanceUtil.java
narendly: I have considered this option, and I decided against it because we lose generality. The nature of runStage sometimes requires the modification of which stages to run and depending on the test, it requires the tests to add/modify/remove Attributes in ClusterEvent. So this is the right granularity.

If we provide some composite method like runStages that runs multiple stages, the use case for it would be limited and not as configurable.

helix-core/src/test/java/org/apache/helix/integration/rebalancer/WagedRebalancer/TestWagedRebalance.java
```diff
@@ -164,7 +214,7 @@ public void testWithInstanceTag() throws Exception {
     validate(_replica);
   }
 
-  @Test(dependsOnMethods = "test")
+  @Test(dependsOnMethods = "testRebalanceTool")
```
jiajunwang, line 217: nit, but why the other tests need to depend on testRebalanceTool?

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
```diff
@@ -140,6 +160,82 @@ public static String serializeByComma(List<String> objects) {
...
   }
 
+  /**
+   * Returns the expected ideal ResourceAssignments for the given resources in the cluster
+   * calculated using the read-only WAGED rebalancer.
+   * @param metadataStoreAddress
+   * @param clusterConfig
+   * @param instanceConfigs
+   * @param liveInstances
+   * @param newIdealStates
```
jiajunwang, line 170: The "new" here might be confusing. So the requirement here is, actually, the caller should include all the resource IS/Configs into the input, right?

helix-core/src/test/java/org/apache/helix/integration/rebalancer/WagedRebalancer/TestWagedRebalance.java
narendly: Doesn't have to depend on testRebalanceTool, but I like to enforce strict ordering for the tests I write..

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
narendly: Yes. Let me rename them. I'll be removing "new".

helix-core/src/test/java/org/apache/helix/integration/rebalancer/WagedRebalancer/TestWagedRebalance.java
jiajunwang: I would prefer to configure the tests in a way of real logic dependencies. Easier for us to maintain.
Moreover, I think we won't remember why it was initially done like this. Let do it in a meaningful way the first time.

narendly: Okay. I reverted this change.

narendly:

Add getIdealAssignmentForWagedFullAuto in HelixUtil for WAGED rebalancer

This commit adds a method, getIdealAssignmentForWagedFullAuto() in HelixUtil that returns to the user the cluster-wide assignment result obtained from running a rebalance using WAGED. The user will be able to use this method to predict how Helix will be rebalancing resources using the WAGED rebalancer.

Issue #1032, opened 2020-05-27
Add check for customized state path existence before adding listener to it


zhangmeng916:
Helix has implemented the feature for customized view aggregation. During participant starts up, the path of "customized state" will be created under the instance node. However, many customers are still with old helix-lib, and won't have this path created. If we add listener without checking the existence of this path, there are some exception log thrown by zkclient for no node, and it may confuse the customer.

PR #1033, opened 2020-05-27
Add path existing check for customized state

zhangmeng916:
### Issues

- [ X] My PR addresses the following Helix issues and references them in the PR description:

(#1032 )

### Description

This PR adds check for the existence of customized state path before adding listener to it. The check is mainly for backward compatibility, as our customers may still use previous helix, and do not have customized state root path created. In that case, we should not add listener to non existing path.

### Tests

- [X ] The following is the result of the "mvn test" command on the appropriate module:

helix-core:
[ERROR] Tests run: 1147, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 4,717.53 s <<< FAILURE! - in TestSuite
[ERROR] testDisableCluster(org.apache.helix.integration.rebalancer.TestClusterInMaintenanceModeWhenReachingMaxPartition)  Time elapsed: 5.078 s  <<< FAILURE!
	at org.apache.helix.integration.rebalancer.TestClusterInMaintenanceModeWhenReachingMaxPartition.testDisableCluster(TestClusterInMaintenanceModeWhenReachingMaxPartition.java:119)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestClusterInMaintenanceModeWhenReachingMaxPartition.testDisableCluster:119 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1147, Failures: 1, Errors: 0, Skipped: 1
[INFO] 

helix-rest:
[INFO] Tests run: 160, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 180.82 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 160, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  03:04 min
[INFO] Finished at: 2020-05-27T14:11:30-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -1256,14 +1256,18 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
 
         for (String instance : curInstances.keySet()) {
           if (lastInstances == null || !lastInstances.containsKey(instance)) {
-            try {
-              manager.addCustomizedStateRootChangeListener(this, instance);
-              logger.info(manager.getInstanceName() + " added customized root change listener for"
-                  + " " + instance
-                  + ", listener: " + this);
-            } catch (Exception e) {
-              logger.error("Fail to add customized root change listener for instance: " + instance,
-                  e);
+            PropertyKey propertyKey =
+                keyBuilder.customizedStatesRoot(instance);
+            if (propertyKey.getPath() != null) {
+              try {
+                manager.addCustomizedStateRootChangeListener(this, instance);
+                logger.info(
+                    manager.getInstanceName() + " added customized root change listener for" + " " + instance
+                        + ", listener: " + this);
+              } catch (Exception e) {
+                logger.error("Fail to add customized root change listener for instance: " + instance,
+                    e);
+              }
             }
         }
       }
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -1256,18 +1256,14 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
 
         for (String instance : curInstances.keySet()) {
           if (lastInstances == null || !lastInstances.containsKey(instance)) {
-            PropertyKey propertyKey =
-                keyBuilder.customizedStatesRoot(instance);
-            if (propertyKey.getPath() != null) {
-              try {
-                manager.addCustomizedStateRootChangeListener(this, instance);
-                logger.info(
-                    manager.getInstanceName() + " added customized root change listener for" + " " + instance
-                        + ", listener: " + this);
-              } catch (Exception e) {
-                logger.error("Fail to add customized root change listener for instance: " + instance,
-                    e);
-              }
+            try {
+              manager.addCustomizedStateRootChangeListener(this, instance);
+              logger.info(manager.getInstanceName() + " added root path listener for customized "
+                  + "state change for" + " " + instance + ", listener: " + this);
+            } catch (Exception e) {
+              logger.error(
+                  "Fail to add root path listener for customized state change" + "for instance: "
+                      + instance, e);
             }
         }
       }
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -1259,10 +1259,10 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
             try {
               manager.addCustomizedStateRootChangeListener(this, instance);
               logger.info(manager.getInstanceName() + " added root path listener for customized "
-                  + "state change for" + " " + instance + ", listener: " + this);
+                  + "state change for " + instance + ", listener: " + this);
             } catch (Exception e) {
               logger.error(
-                  "Fail to add root path listener for customized state change" + "for instance: "
+                  "Fail to add root path listener for customized state change for instance: "
                       + instance, e);
             }
         }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -73,7 +73,6 @@ import org.apache.helix.model.LiveInstance;
 import org.apache.helix.model.Message;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.monitoring.mbeans.HelixCallbackMonitor;
-import org.apache.helix.zookeeper.api.client.HelixZkClient;
 import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.zkclient.IZkChildListener;
@@ -639,8 +638,13 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
             }
           }
         } catch (ZkNoNodeException | HelixMetaDataAccessException e) {
-          logger.warn(
-              "fail to subscribe child/data change. path: " + path + ", listener: " + _listener, e);
+          if (_changeType == CUSTOMIZED_STATE_ROOT) {
+            logger.warn(
+                "fail to subscribe child/data change. path: " + path + ", listener: " + _listener);
+          } else {
+            logger.warn(
+                "fail to subscribe child/data change. path: " + path + ", listener: " + _listener, e);
+          }
         }
       }
     }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -144,19 +144,19 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
         "Singleton", "CallbackHandler-AsycSubscribe") {
       @Override
       protected void handleEvent(SubscribeChangeEvent event) {
-        logger.info("Resubscribe change listener to path: " + event.path + ", for listener: "
-            + event.listener + ", watchChild: " + event.watchChild);
+        logger.info("Resubscribe change listener to path: {}, for listener: {}, watchChild: {}",
+            event.path, event.listener, event.watchChild);
         try {
           if (event.handler.isReady()) {
             event.handler.subscribeForChanges(event.callbackType, event.path, event.watchChild);
           } else {
-            logger.info("CallbackHandler is not ready, stop subscribing changes listener to path: "
-                + event.path + ", for listener: " + event.listener + ", watchChild: "
-                + event.watchChild);
+            logger.info("CallbackHandler is not ready, stop subscribing changes listener to "
+                    + "path:{} for listener: {} watchChild: {}", event.path, event.listener,
+                event.listener);
           }
         } catch (Exception e) {
-          logger.error("Failed to resubscribe change to path: " + event.path + " for listener "
-              + event.listener, e);
+          logger.error("Failed to resubscribe change to path: {} for listener: {}", event.path,
+              event.listener, e);
         }
       }
     };
@@ -259,7 +259,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
 
     if (asyncBatchModeEnabled != null) {
       _batchModeEnabled = Boolean.parseBoolean(asyncBatchModeEnabled);
-      logger.info("isAsyncBatchModeEnabled by default: " + _batchModeEnabled);
+      logger.info("isAsyncBatchModeEnabled by default: {}", _batchModeEnabled);
     }
 
     if (batchMode != null) {
@@ -336,8 +336,8 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
         _preFetchEnabled = preFetchInMethod.enabled();
       }
     } catch (NoSuchMethodException e) {
-      logger.warn("No method " + callbackMethod.getName() + " defined in listener "
-          + _listener.getClass().getCanonicalName());
+      logger.warn("No method {} defined in listener {}", callbackMethod.getName(),
+          _listener.getClass().getCanonicalName());
     }
   }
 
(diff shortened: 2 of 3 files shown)


helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1256,14 +1256,18 @@ protected void checkLiveInstancesObservation(List<LiveInstance> liveInstances,
...
-              manager.addCustomizedStateRootChangeListener(this, instance);
-              logger.info(manager.getInstanceName() + " added customized root change listener for"
-                  + " " + instance
-                  + ", listener: " + this);
-            } catch (Exception e) {
-              logger.error("Fail to add customized root change listener for instance: " + instance,
-                  e);
+            PropertyKey propertyKey =
+                keyBuilder.customizedStatesRoot(instance);
+            if (propertyKey.getPath() != null) {
```
narendly, line 1261: You're creating a PropertyKey object with the given instance name and checking whether it's null? How would it be null if you've just created the path?

Did you mean to check whether the path exists in ZooKeeper? If so, shouldn't you do an exist() check with a ZkClient-based API?

Unless I'm mistaken and we actually somehow pre-populate the KeyBuilder with a whitelist of instances for the CustomizedView feature?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1256,14 +1256,18 @@ protected void checkLiveInstancesObservation(List<LiveInstance> liveInstances,
...
+              try {
+                manager.addCustomizedStateRootChangeListener(this, instance);
+                logger.info(
+                    manager.getInstanceName() + " added customized root change listener for" + " " + instance
+                        + ", listener: " + this);
+              } catch (Exception e) {
+                logger.error("Fail to add customized root change listener for instance: " + instance,
```
narendly, line 1268: "customized root change" is confusing. Could we explicitly state what it is? For example,

"Added (or Failed to add) the root path for Customized State ..."

Also, this log doesn't seem entirely necessary. Don't need to log every time we add a listener - I'd make this a debug log so as not to pollute the log.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
zhangmeng916: oh..I was thinking to use exist checking, but put the wrong check, let me update it.
This is just for backward incompatibility check. We're totally fine without this check, but the participant would need to be on the latest version to have customized state path created. If they don't, there can be some error logs when adding listener to customized state path. It's still fine, but might cause confusion.

narendly: Yes, for backward compatibility, you need to check the root path existence in ZK. This check is not very meaningful. You're basically doing `if ("~~/instanceName" != null)`.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
zhangmeng916: I updated the log information. For log.info or log.debug, in the controller, all other listeners are added with log.info. I think this is for easy debug purpose. Please let me know if the previous convention should be changed.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -591,8 +591,11 @@ public void addCurrentStateChangeListener(org.apache.helix.CurrentStateChangeLis
   @Override
   public void addCustomizedStateRootChangeListener(CustomizedStateRootChangeListener listener,
       String instanceName) throws Exception {
-    addListener(listener, new Builder(_clusterName).customizedStatesRoot(instanceName),
-        ChangeType.CUSTOMIZED_STATE_ROOT, new EventType[]{EventType.NodeChildrenChanged});
+    PropertyKey propertyKey = _dataAccessor.keyBuilder().customizedStatesRoot(instanceName);
+    if (_zkclient.exists(propertyKey.getPath())) {
+      addListener(listener, new Builder(_clusterName).customizedStatesRoot(instanceName),
+          ChangeType.CUSTOMIZED_STATE_ROOT, new EventType[]{EventType.NodeChildrenChanged});
+    }
```
narendly, line 598: Do you think it would be useful to log (warn) if this evaluates to false?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -591,8 +591,11 @@ public void addCurrentStateChangeListener(org.apache.helix.CurrentStateChangeLis
   @Override
   public void addCustomizedStateRootChangeListener(CustomizedStateRootChangeListener listener,
       String instanceName) throws Exception {
-    addListener(listener, new Builder(_clusterName).customizedStatesRoot(instanceName),
-        ChangeType.CUSTOMIZED_STATE_ROOT, new EventType[]{EventType.NodeChildrenChanged});
+    PropertyKey propertyKey = _dataAccessor.keyBuilder().customizedStatesRoot(instanceName);
+    if (_zkclient.exists(propertyKey.getPath())) {
```
pkuwm, line 595: I am concerned about the performance. This check adds more time for adding listeners because of network latency. I wonder if there is a better way so this check could be avoided.

zhangmeng916: If we don't check path, the only way I can think of is to check participant side version.  Doesn't sounds a clean way though.

pkuwm: Maybe we could just change the the log in callback handler not to log the exception stack trace.

zhangmeng916: That would impact all the exceptions right? Then it's not good for debugging purpose.

pkuwm: It is just a warn log. I think it is OK not to print the stack trace. And if the stack trace is necessary, we could keep the stack trace for the other paths, but for this customized state path, just log the message.

narendly: exists() check is not a heavyweight ZK operation and considering we only call this when there's a LiveInstance change (or periodic rebalance), I don't think performance is a concern here.

lei-xia: Dummy question, why participant needs to listen on customized view changes? Only controller needs to listen on the change, not?

zhangmeng916: Participant does not need to listen on customized view change. The backward compatibility issue happens when we have a new version controller, while the client is on an old version participant. The new path INSTANCES/CUSTOMIZED_STATES does not exist, but controller will try to add listener to that path during live instance change. Then zkclient will have a no node exception. It's fine, just a warn, but kind of pollute client's log and make them confused.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
mgao0: Did we say that if participant's version is older than the controller version, it's fine? If so then I think we may not need to log it.

narendly:
Here's the copy of an offline discussion


> Jiajun Wang:man-woman-girl-girl:  12:27
> Meng will confirm. But I think that scenario was discussed and address in the original design.
> 12:28
> I will suggest keeping the root path listener (default feature for the new version controller) but the children's paths listeners will depend on the configuration.
> 
> 
> 
> 
> 
> 12:29
> that is the minimum code change and logic is the simpliest.
> 12:30
> Basically, we only need to avoid the default subscription to children nodes in the root path subscribe process (edited) 
> 
> Hunter Lee 12:30
> So are we comfortable with creating CallbackHandlers for nonexistent root paths?
> 
> Jiajun Wang  12:30
> Then all the rest parts have been handled by the current code.
> New
> 12:32
> @hulee That's the only downside. But my argument is that:
> It is a default feature for the new controller.
> It won't cause watcher leakage. It is always associated with a live callback handler, although never triggered,
> The new participant will always have this folder. That's the agreement.
> 12:33
> The alternative design will create many additional codes for the "perfect" backward compatibility.
> 12:33
> Which is not necessary IMO
> 
> Hunter Lee  12:35
> Ok. Although not perfect, I think that approach strikes the right balance

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
dasahcc: Shall we have some logs to indicate it is not able to subscribe? Maybe warning or something?

zhangmeng916:
I updated this rb with a solution to avoid stack trace in our customers who have not updated their helix-lib version. As discussed before, this is the simplest while robust way to solve the issue without performance penalty. For further improvement on customized state listener, we track in this issue: https://github.com/apache/helix/issues/1046 and tackle it later.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1258,12 +1258,12 @@ protected void checkLiveInstancesObservation(List<LiveInstance> liveInstances,
           if (lastInstances == null || !lastInstances.containsKey(instance)) {
             try {
               manager.addCustomizedStateRootChangeListener(this, instance);
-              logger.info(manager.getInstanceName() + " added customized root change listener for"
-                  + " " + instance
-                  + ", listener: " + this);
+              logger.info(manager.getInstanceName() + " added root path listener for customized "
```
pkuwm, line 1261: I understand you may want to keep logging as original. But since you are changing the logging code here, I suggest changing it to use parameterized logging: `logger.info("Hello {}", world);`. This parameterized logging has benefits of reduce string concatenation overhead when logging level info is not enabled, and easier to read in the code. Same as the other places in this PR.
Nit, I realized this for loop needs unindent.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -639,8 +638,13 @@ private void subscribeForChanges(NotificationContext.Type callbackType, String p
             }
           }
         } catch (ZkNoNodeException | HelixMetaDataAccessException e) {
-          logger.warn(
-              "fail to subscribe child/data change. path: " + path + ", listener: " + _listener, e);
+          if (_changeType == CUSTOMIZED_STATE_ROOT) {
+            logger.warn(
+                "fail to subscribe child/data change. path: " + path + ", listener: " + _listener);
```
pkuwm, line 643: `fail to` -> `Failed to` ? Same for the others.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: `+ "state change for" + " "`
Can we please refine the format as well.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1258,12 +1258,12 @@ protected void checkLiveInstancesObservation(List<LiveInstance> liveInstances,
...
+                  + "state change for" + " " + instance + ", listener: " + this);
             } catch (Exception e) {
-              logger.error("Fail to add customized root change listener for instance: " + instance,
-                  e);
+              logger.error(
+                  "Fail to add root path listener for customized state change" + "for instance: "
```
jiajunwang, line 1265: Why use 2 separate sections for the constant string contents?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
zhangmeng916: I updated this rb with a solution to avoid stack trace in our customers who have not updated their helix-lib version. As discussed before, this is the simplest while robust way to solve the issue without performance penalty. For further improvement on customized state listener, we track in this issue: #1046 and tackle it later.

zhangmeng916:
This PR is ready to merge. Approved by @pkuwm 
Final commit message: 
Remove stack trace when customized state root does not exist and throws no node exception.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -639,14 +639,20 @@ private void subscribeForChanges(NotificationContext.Type callbackType, String p
...
         } catch (ZkNoNodeException | HelixMetaDataAccessException e) {
-          logger.warn(
-              "fail to subscribe child/data change. path: " + path + ", listener: " + _listener, e);
+          // TODO: avoid calling getChildren for path that does not exist
+          if (_changeType == CUSTOMIZED_STATE_ROOT) {
+            logger.warn("Failed to subscribe child/data change. path: {}, listener: {}", path,
+                _listener);
+          } else {
+            logger.warn("Failed to subscribe child/data change. path: {}, listener: {}", path,
+                _listener, e);
```
narendly, line 648: Can we just merge these two conditions? There's practically no difference.

zhangmeng916: @narendly tried to remove duplicates. However, this will make the log and stacktrace to be in two entries. Won't look good.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -639,14 +639,18 @@ private void subscribeForChanges(NotificationContext.Type callbackType, String p
...
           }
         } catch (ZkNoNodeException | HelixMetaDataAccessException e) {
-          logger.warn(
-              "fail to subscribe child/data change. path: " + path + ", listener: " + _listener, e);
+          logger.warn("Failed to subscribe child/data change. path: {}, listener: {}", path,
+              _listener);
+          // TODO: avoid calling getChildren for path that does not exist
+          if (_changeType != CUSTOMIZED_STATE_ROOT) {
+            logger.warn("", e);
+          }
```
narendly, line 647:
```
if (_changeType != CUSTOMIZED_STATE_ROOT) {
  LOG.warn("Failed to subscribe .... Instance doesn't support Customized State!, path, _listerner);
} else {
  LOG.warn("Failed to.... ");
}
```
How about this? this way you don't print all the exceptions stack trace, and it doesn't look like it's duplicate code, and the message is clear.

zhangmeng916:
This PR is ready to merge. Approved by @pkuwm
Final commit message:
Remove stack trace when customized state root does not exist and throws no node exception.
Change logging in Callback handler to be parameterized logging.

PR #1040, opened 2020-05-29
Remove previousAssignment in processTaskWithPendingMessage method

### LINKED ISSUE #1039: TODO: Remove previousAssignment from processTaskWithPendingMessage method
processTaskWithPendingMessage method is still getting previousAssignment to do scheduling for the tasks with pendingMessages. We need to make decision based on currentState only if possible.

alirezazamani:
### Issues
Fixes #1039 

### Description
The processTaskWithPendingMessage method is relying on the previousAssignment.
In this PR, this method has been modified and previousAssignment has been
replaced with currentState.

### Tests
TestUpdatePreviousAssignedTaskStatusWithPendingMessage

Helix-Core:
```
[ERROR] Tests run: 1146, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5,255.063 s <<< FAILURE! - in TestSuite
[ERROR] testJobQueueAutoCleanUp(org.apache.helix.integration.task.TestJobQueueCleanUp)  Time elapsed: 300.012 s  <<< FAILURE!
org.testng.internal.thread.ThreadTimeoutException: Method org.testng.internal.TestNGMethod.testJobQueueAutoCleanUp() didn't finish within the time-out 300000

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 1146, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:27 h
[INFO] Finished at: 2020-05-29T14:14:16-07:00
```

The failed tests have passed when I ran them individually.
mvn test -Dtest="TestJobQueueCleanUp"
```
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 16.521 s - in org.apache.helix.integration.task.TestJobQueueCleanUp
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  22.301 s
[INFO] Finished at: 2020-05-29T14:38:54-07:00
```
Helix-rest:
```
[INFO] Tests run: 159, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 41.274 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  46.810 s
[INFO] Finished at: 2020-05-29T14:40:07-07:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
+++ b/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
@@ -66,9 +66,9 @@ public abstract class AbstractTaskDispatcher {
   // Job Update related methods
 
   public void updatePreviousAssignedTasksStatus(
-      Map<String, SortedSet<Integer>> currentInstanceToTaskAssignments, Set<String> excludedInstances,
-      String jobResource, CurrentStateOutput currStateOutput, JobContext jobCtx, JobConfig jobCfg,
-      ResourceAssignment prevTaskToInstanceStateAssignment, TaskState jobState,
+      Map<String, SortedSet<Integer>> currentInstanceToTaskAssignments,
+      Set<String> excludedInstances, String jobResource, CurrentStateOutput currStateOutput,
+      JobContext jobCtx, JobConfig jobCfg, TaskState jobState,
       Map<String, Set<Integer>> assignedPartitions, Set<Integer> partitionsToDropFromIs,
       Map<Integer, PartitionAssignment> paMap, TargetState jobTgtState,
       Set<Integer> skippedPartitions, WorkflowControllerDataProvider cache,
@@ -130,8 +130,8 @@ public abstract class AbstractTaskDispatcher {
           // If there is a pending message whose destination state is different from the current
           // state, just make the same assignment as the pending message. This is essentially
           // "waiting" until this state transition is complete
-          processTaskWithPendingMessage(prevTaskToInstanceStateAssignment, pId, pName, instance,
-              pendingMessage, jobState, currState, paMap, assignedPartitions);
+          processTaskWithPendingMessage(pId, pName, instance, pendingMessage, jobState, currState,
+              paMap, assignedPartitions);
           continue;
         }
 
@@ -439,7 +439,6 @@ public abstract class AbstractTaskDispatcher {
   /**
    * Create an assignment based on an already-existing pending message. This effectively lets the
    * Controller to "wait" until the pending state transition has been processed.
-   * @param prevAssignment
    * @param pId
    * @param pName
    * @param instance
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -449,43 +448,31 @@ private void updatePartitionInformationInJobContext(CurrentStateOutput currentSt
...
-        LOG.warn(String.format(
-            "Task pending to-state is %s while previous assigned state is %s. This should not"
-                + "happen.",
-            pendingMessage.getToState(), prevState));
+  private void processTaskWithPendingMessage(Integer pId, String pName, String instance,
+      Message pendingMessage, TaskState jobState, TaskPartitionState currState,
+      Map<Integer, PartitionAssignment> paMap, Map<String, Set<Integer>> assignedPartitions) {
+
+    if (jobState == TaskState.TIMING_OUT && currState == TaskPartitionState.INIT
+        && pendingMessage.getToState().equals(TaskPartitionState.RUNNING.name())) {
```
dasahcc, line 456: Could this cause problem? If there is a cancel message, so the toState is not RUNNING, it should INIT? Then are we cancelling the CANCEL message?

alirezazamani: This part is basically similar to the previous version of the code. 

From the comments and the code: If there is a pending message for this task and this pending message is INIT->RUNNING, if the job goes to TIMING_OUT state, we need to cancel this state transition. Hence we would set the task back to INIT state in this case. Then if Helix can cancel the ST (if enabled) this ST will be cancelled by controller.

Also without this change we are in the same situation: We decided to send INIT->RUNNING message for this task (hence in previous pipeline we set previousAssignment to RUNNING and send INIT-> RUNNING message), in next pipelines while the message has not been processed yet, job goes to TIMING_OUT state, hence this message is not needed we can cancel it (if enabled). In this case we set the paMap back to INIT.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -449,43 +448,31 @@ private void updatePartitionInformationInJobContext(CurrentStateOutput currentSt
...
+      // While job is timing out, if the task is pending on INIT->RUNNING, set it back to INIT,
+      // so that Helix will cancel the transition.
+      paMap.put(pId, new PartitionAssignment(instance, TaskPartitionState.INIT.name()));
```
narendly, line 459: I'm just brainstorming here:

I think a better state to assign here might be DROPPED? Previously, we would send things to INIT but that would be a long-winded way to cause this to be dropped. And I'm not sure with the recent changes, the tasks that are timing out that are in INIT state will indeed be dropped eventually.

If they don't, then there's a problem - init state takes up a quota in the quota calculation, and we might end up seeing quota leaks.

Have you tried to drop here directly?

alirezazamani: @narendly Thank you for the suggestion. I think what you are suggesting would also work. However, I think the logic behind this part of code should be untouched as much as possible. If we want to do send DROPPED here (although I think it would work), we need to release the quota and do further logic changes. The reason is once we DROP the the task from instance, we will miss currentState and we will not see that task in other methods in order to release quota. Hence, I prefer to send DROPPED for the task using two level approach (i.e. first send it to INIT and then DROP it).

Also since there is a pending INIT->RUNNING message existed, it is not very safe to release the quota here. Because the task might go to RUNNING state and before we cancel it, the quota information will be mismatched with the actual scenario. So if we first let the task to go to INIT state and then releasing the quota and sending it to DROPPED will be better option in my opinion.

To answer your question, we will actually handle this case for releasing the quota in updatePreviousAssignedTasksStatus method. Here the code that would do that:

```
        case INIT: {
          // INIT is a temporary state for tasks
          // Two possible scenarios for INIT:
          // 1. Task is getting scheduled for the first time. In this case, Task's state will go
          // from null->INIT->RUNNING, and this INIT state will be transient and very short-lived
          // 2. Task is getting scheduled for the first time, but in this case, job is timed out or
          // timing out. In this case, it will be sent back to INIT state to be removed. Here we
          // ensure that this task then goes from INIT to DROPPED so that it will be released from
          // AssignableInstance to prevent resource leak
          if (jobState == TaskState.TIMED_OUT || jobState == TaskState.TIMING_OUT
              || jobTgtState == TargetState.DELETE) {
            // Job is timed out or timing out or targetState is to be deleted, so its tasks will be
            // sent back to INIT
            // In this case, tasks' IdealState will be removed, and they will be sent to DROPPED
            partitionsToDropFromIs.add(pId);
            
            assignedPartitions.get(instance).add(pId);
            paMap.put(pId, new PartitionAssignment(instance, TaskPartitionState.DROPPED.name()));

            // Also release resources for these tasks
            **assignableInstanceManager.release(instance, taskConfig, quotaType);**
            break;
          } 
```

Your comment actually guided me to add another safety check that we can added for INIT case. I added two lines to make sure we DROP the task if job is TIMED_OUT or TIMING_OUT and the task is in INIT state. Thanks for the comments and careful review.

alirezazamani:

Final commit message:

Remove previousAssignment in processTaskWithPendingMessage method

The processTaskWithPendingMessage method is relying on the previousAssignment.
In this commit, this method has been modified and previousAssignment has been
replaced with currentState.

Issue #1042, opened 2020-05-29
Validate instance topology configuration before let it comes online


xyuanlu:
Helix does not check if an instance's topology config is defined correctly when the instance is added or joins cluster. This will cause Helix to fail the rebalance. We could add the sanity check when 
    1. Instance is added to the cluster,
    2. Instance joins cluster,
    3. User update InstanceConfig.
    4. Controller prepares the topology structure for rebalance   <-- Currently we only check here
    
Since we already have sanity check logic in controller/rebalancer/topology/Topology.java and controller/rebalancer/waged/model/AssignableNode.java, it's better to extract the logic into an util function first and use the same logic to do sanity check in multiple places. 

Reference: https://cwiki.apache.org/confluence/display/HELIX/Helix+Topology-aware+Rebalance+Strategy#HelixTopology-awareRebalanceStrategy-TopologyInformation

PR #1043, opened 2020-05-30
Code clean up Topology.java

xyuanlu:
### Issues

#1042 Validate instance topology configuration before let it comes online 

### Description

We need sanity check for topology definition in InstanceConfig in multiple places. Since we already have sanity check logic in controller/rebalancer/topology/Topology.java and controller/rebalancer/waged/model/AssignableNode.java, it's better to extract the logic into an single util function first and use the same logic to do sanity check in multiple places.

This PR is the first PR for issue #1042. It extracts the sanity check logic in multiple functions into one single function in Topology.java to align with the topology verification logic in AssignableNode.java. 

It also changes the implementation for InstanceConfig.getDomainAsMap since to resolve CPU usage hotspot showed in CPU profiler. 

### Tests

No new test is added since this PR is code clean up.

```
[ERROR] Failures: 
[ERROR]   TestWagedRebalance.testChangeIdealState:271->validate:614 expected:<true> but was:<false>                                                      
[ERROR]   TestWorkflowTermination.testWorkflowRunningTimeout:126 » Helix Workflow "testW...                                                              
[INFO]
[ERROR] Tests run: 1145, Failures: 3, Errors: 0, Skipped: 1


[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 40.747 s - in org.apache.helix.integration.rebalancer.WagedRebalancer.TestWagedRebalance
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 9.246 s - in org.apache.helix.integration.task.TestDeleteJobFromJobQueue
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 32.675 s - in org.apache.helix.integration.task.TestWorkflowTermination



```
### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
@@ -83,10 +83,11 @@ public class Topology {
         // Customized cluster topology definition is configured.
         String[] topologyStr = topologyDef.trim().split("/");
         int lastValidTypeIdx = topologyStr.length - 1;
-        for (; lastValidTypeIdx >= 0; --lastValidTypeIdx) {
+        while(lastValidTypeIdx >= 0) {
           if (topologyStr[lastValidTypeIdx].length() != 0) {
             break;
           }
+          --lastValidTypeIdx;
         }
         if (lastValidTypeIdx < 0) {
           throw new HelixException("Invalid cluster topology definition " + topologyDef);
@@ -195,8 +196,9 @@ public class Topology {
 
     for (String instance : _allInstances) {
       InstanceConfig insConfig = _instanceConfigMap.get(instance);
-      LinkedHashMap<String, String> instanceTopologyMap = new LinkedHashMap<>();
-      if (computeInstanceTopologyMap(_clusterConfig, instance, insConfig, instanceTopologyMap)) {
+      Map<String, String> instanceTopologyMap =
+          computeInstanceTopologyMap(_clusterConfig, instance, insConfig);
+      if (instanceTopologyMap != null) {
         int weight = insConfig.getWeight();
         if (weight < 0 || weight == InstanceConfig.WEIGHT_NOT_SET) {
           weight = DEFAULT_NODE_WEIGHT;
@@ -207,46 +209,58 @@ public class Topology {
     return root;
   }
 
-  private boolean computeInstanceTopologyMap(ClusterConfig clusterConfig, String instance,
-      InstanceConfig instanceConfig, LinkedHashMap<String, String> instanceTopologyMap) {
+  /**
+   * This function returns a LinkedHashMap<String, String> object representing
+   * the topology path for an instance.
+   * LinkedHashMap is used here since the order of the path needs to be preserved
+   * when creating the topology tree.
+   *
+   * @param clusterConfig    clusterConfig of the cluster.
+   * @param instance         Name of the given instance.
+   * @param instanceConfig   instanceConfig of the instance.
+   * @return an LinkedHashMap object representing the topology path for the input instance.
+   */
+  private LinkedHashMap<String, String> computeInstanceTopologyMap(ClusterConfig clusterConfig,
+      String instance, InstanceConfig instanceConfig) {
+    LinkedHashMap<String, String> instanceTopologyMap = new LinkedHashMap<>();
     if (clusterConfig.isTopologyAwareEnabled()) {
       String topologyDef = clusterConfig.getTopology();
       if (topologyDef == null) {
-        // Return a map using default cluster topology definition, i,e. /root/zone/instance
+        // Return a ordered map using default cluster topology definition, i,e. /root/zone/instance
         String zone = instanceConfig.getZoneId();
         if (zone == null) {
           // we have the hierarchy style of domain id for instance.
           if (instanceConfig.getInstanceEnabled() && (clusterConfig.getDisabledInstances() == null
               || !clusterConfig.getDisabledInstances().containsKey(instance))) {
-            // if enabled instance missing ZONE_ID information, fails the rebalance.
+            // if enabled instance missing ZONE_ID information, fail the rebalance.
             throw new HelixException(String
                 .format("ZONE_ID for instance %s is not set, fail the topology-aware placement!",
                     instance));
           } else {
             // if the disabled instance missing ZONE setting, ignore it should be fine.
             logger.warn("ZONE_ID for instance {} is not set, ignore the instance!", instance);
-            return false;
+            return null;
           }
         }
         instanceTopologyMap.put(Types.ZONE.name(), zone);
         instanceTopologyMap.put(Types.INSTANCE.name(), instance);
       } else {
         /*
-         * Return a map representing the cluster structure using cluster topology defined in
-         * TOPOLOGY in ClusterConfig.
+         * Return a ordered map representing the instance path. The topology order is defined in
+         * ClusterConfig.topology.
          */
         String domain = instanceConfig.getDomainAsString();
         if (domain == null || domain.isEmpty()) {
           if (instanceConfig.getInstanceEnabled() && (clusterConfig.getDisabledInstances() == null
               || !clusterConfig.getDisabledInstances().containsKey(instance))) {
-            // if enabled instance missing domain information, fails the rebalance.
+            // if enabled instance missing domain information, fail the rebalance.
             throw new HelixException(String
                 .format("Domain for instance %s is not set, fail the topology-aware placement!",
                     instance));
           } else {
             // if the disabled instance missing domain setting, ignore it should be fine.
             logger.warn("Domain for instance {} is not set, ignore the instance!", instance);
-            return false;
+            return null;
           }
         }
         Map<String, String> domainAsMap;
@@ -258,26 +272,34 @@ public class Topology {
               domain, instance));
         }
         String[] topologyKeys = topologyDef.trim().split("/");
+        int numOfmatchedKeys = 0;
         for (String key : topologyKeys) {
           if (!key.isEmpty()) {
             // if a key does not exist in the instance domain config, apply the default domain value.
             String value = domainAsMap.get(key);
-            if (value == null || value.length()==0) {
+            if (value == null || value.length() == 0) {
               value = DEFAULT_PATH_PREFIX + key;
+            } else {
+              numOfmatchedKeys++;
             }
             instanceTopologyMap.put(key, value);
           }
+        }
+        if (numOfmatchedKeys != domainAsMap.size()) {
+          logger.warn(
+              "Key-value pairs in InstanceConfig.Domain {} do not align with keys in ClusterConfig.Topology {}",
+              domain, topologyKeys.toString());
         }
       }
     } else {
       instanceTopologyMap.put(Types.INSTANCE.name(), instance);
     }
-    return true;
+    return instanceTopologyMap;
   }
 
   /**
    * Add an end node to the tree, create all the paths to the leaf node if not present.
-   * Input param 'pathNameMap' must have a certain order where the order of the keys is
+   * Input param 'pathNameMap' must have a certain order where the order of the keys should be
    * the same as the path order in ClusterConfig.topology.
    */
   private void addEndNode(Node root, String instanceName, Map<String, String> pathNameMap,
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
@@ -22,9 +22,10 @@ package org.apache.helix.controller.rebalancer.topology;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -55,10 +56,12 @@ public class Topology {
   private final List<String> _liveInstances;
   private final Map<String, InstanceConfig> _instanceConfigMap;
   private final ClusterConfig _clusterConfig;
-  private static final String DEFAULT_PATH_PREFIX = "Helix_default_";
+  private static final String DEFAULT_DOMAIN_PREFIX = "Helix_default_";
 
   private String _faultZoneType;
   private String _endNodeType;
+  private LinkedHashSet<String> _clusterTopologyKeys;
+  private Map<String, String> _defaultDomainPathValues = new HashMap<>();
 
   public Topology(final List<String> allNodes, final List<String> liveNodes,
       final Map<String, InstanceConfig> instanceConfigMap, ClusterConfig clusterConfig) {
@@ -76,28 +79,28 @@ public class Topology {
           _allInstances.removeAll(_instanceConfigMap.keySet())));
     }
     _clusterConfig = clusterConfig;
+    _clusterTopologyKeys = new LinkedHashSet<>();
 
     if (clusterConfig.isTopologyAwareEnabled()) {
       String topologyDef = _clusterConfig.getTopology();
       if (topologyDef != null) {
         // Customized cluster topology definition is configured.
-        String[] topologyStr = topologyDef.trim().split("/");
-        int lastValidTypeIdx = topologyStr.length - 1;
-        while(lastValidTypeIdx >= 0) {
-          if (topologyStr[lastValidTypeIdx].length() != 0) {
-            break;
+        String[] topologyKeys = topologyDef.trim().split("/");
+        int lastValidTypeIdx = 0;
+        for (int i = 0; i < topologyKeys.length; i++) {
+          if (topologyKeys[i].length() != 0) {
+            _clusterTopologyKeys.add(topologyKeys[i]);
+            lastValidTypeIdx = i;
           }
-          --lastValidTypeIdx;
         }
-        if (lastValidTypeIdx < 0) {
+        if (_clusterTopologyKeys.size() == 0) {
           throw new HelixException("Invalid cluster topology definition " + topologyDef);
         }
-        _endNodeType = topologyStr[lastValidTypeIdx];
+        _endNodeType = topologyKeys[lastValidTypeIdx];
         _faultZoneType = clusterConfig.getFaultZoneType();
         if (_faultZoneType == null) {
           _faultZoneType = _endNodeType;
-        }
-        if (Arrays.stream(topologyStr).noneMatch(type -> type.equals(_faultZoneType))) {
+        } else if (!_clusterTopologyKeys.contains(_faultZoneType)) {
           throw new HelixException(String
               .format("Invalid fault zone type %s, not present in topology definition %s.",
                   _faultZoneType, topologyDef));
@@ -194,19 +197,37 @@ public class Topology {
     root.setId(computeId("root"));
     root.setType(Types.ROOT.name());
 
-    for (String instance : _allInstances) {
-      InstanceConfig insConfig = _instanceConfigMap.get(instance);
-      Map<String, String> instanceTopologyMap =
-          computeInstanceTopologyMap(_clusterConfig, instance, insConfig);
-      if (instanceTopologyMap != null) {
+    // TODO: Currently we add disabled instance to the topology tree. Since they are not considered
+    // TODO: in relabalnce, maybe we should skip adding them to the tree for consistence.
+    for (String instanceName : _allInstances) {
+      InstanceConfig insConfig = _instanceConfigMap.get(instanceName);
+      LinkedHashMap<String, String> instanceTopologyMap;
+      try {
+        instanceTopologyMap =
+            computeInstanceTopologyMap(_clusterConfig.isTopologyAwareEnabled(), instanceName,
+                insConfig, _clusterTopologyKeys, _defaultDomainPathValues);
         int weight = insConfig.getWeight();
         if (weight < 0 || weight == InstanceConfig.WEIGHT_NOT_SET) {
           weight = DEFAULT_NODE_WEIGHT;
         }
-        addEndNode(root, instance, instanceTopologyMap, weight, _liveInstances);
+        addEndNode(root, instanceName, instanceTopologyMap, weight, _liveInstances);
+      } catch (IllegalArgumentException e) {
+        if (isInstanceEnabled(_clusterConfig, instanceName, insConfig)) {
+          throw e;
+        } else {
+          logger
+              .warn("Topology setting {} for instance {} is unset or invalid, ignore the instance!",
+                  insConfig.getDomainAsString(), instanceName);
+        }
       }
     }
     return root;
+  }
+
+  private boolean isInstanceEnabled(ClusterConfig clusterConfig, String instanceName,
+      InstanceConfig instanceConfig) {
+    return (instanceConfig.getInstanceEnabled() && (clusterConfig.getDisabledInstances() == null
+        || !clusterConfig.getDisabledInstances().containsKey(instanceName)));
   }
 
   /**
@@ -215,94 +236,71 @@ public class Topology {
    * LinkedHashMap is used here since the order of the path needs to be preserved
    * when creating the topology tree.
    *
-   * @param clusterConfig    clusterConfig of the cluster.
-   * @param instance         Name of the given instance.
-   * @param instanceConfig   instanceConfig of the instance.
    * @return an LinkedHashMap object representing the topology path for the input instance.
    */
-  private LinkedHashMap<String, String> computeInstanceTopologyMap(ClusterConfig clusterConfig,
-      String instance, InstanceConfig instanceConfig) {
+  private LinkedHashMap<String, String> computeInstanceTopologyMap(boolean isTopologyAwareEnabled,
+      String instanceName, InstanceConfig instanceConfig, LinkedHashSet<String> clusterTopologyKeys,
+      Map<String, String> defaultDomainPathValuesCache) throws IllegalArgumentException {
     LinkedHashMap<String, String> instanceTopologyMap = new LinkedHashMap<>();
-    if (clusterConfig.isTopologyAwareEnabled()) {
-      String topologyDef = clusterConfig.getTopology();
-      if (topologyDef == null) {
+    if (isTopologyAwareEnabled) {
+      if (clusterTopologyKeys == null || clusterTopologyKeys.size() == 0) {
         // Return a ordered map using default cluster topology definition, i,e. /root/zone/instance
         String zone = instanceConfig.getZoneId();
         if (zone == null) {
-          // we have the hierarchy style of domain id for instance.
-          if (instanceConfig.getInstanceEnabled() && (clusterConfig.getDisabledInstances() == null
-              || !clusterConfig.getDisabledInstances().containsKey(instance))) {
-            // if enabled instance missing ZONE_ID information, fail the rebalance.
-            throw new HelixException(String
-                .format("ZONE_ID for instance %s is not set, fail the topology-aware placement!",
-                    instance));
-          } else {
-            // if the disabled instance missing ZONE setting, ignore it should be fine.
-            logger.warn("ZONE_ID for instance {} is not set, ignore the instance!", instance);
-            return null;
-          }
+          throw new IllegalArgumentException(String
+              .format("ZONE_ID for instance %s is not set, fail the topology-aware placement!",
+                  instanceName));
         }
         instanceTopologyMap.put(Types.ZONE.name(), zone);
-        instanceTopologyMap.put(Types.INSTANCE.name(), instance);
+        instanceTopologyMap.put(Types.INSTANCE.name(), instanceName);
       } else {
         /*
          * Return a ordered map representing the instance path. The topology order is defined in
          * ClusterConfig.topology.
          */
-        String domain = instanceConfig.getDomainAsString();
-        if (domain == null || domain.isEmpty()) {
-          if (instanceConfig.getInstanceEnabled() && (clusterConfig.getDisabledInstances() == null
-              || !clusterConfig.getDisabledInstances().containsKey(instance))) {
-            // if enabled instance missing domain information, fail the rebalance.
-            throw new HelixException(String
-                .format("Domain for instance %s is not set, fail the topology-aware placement!",
-                    instance));
-          } else {
-            // if the disabled instance missing domain setting, ignore it should be fine.
-            logger.warn("Domain for instance {} is not set, ignore the instance!", instance);
-            return null;
-          }
-        }
         Map<String, String> domainAsMap;
         try {
           domainAsMap = instanceConfig.getDomainAsMap();
         } catch (IllegalArgumentException e) {
-          throw new HelixException(String.format(
-              "Domain %s for instance %s is not valid, fail the topology-aware placement!",
-              domain, instance));
+          throw new IllegalArgumentException(String
+              .format("Domain %s for instance %s is not valid, fail the topology-aware placement!",
+                  instanceConfig.getDomainAsString(), instanceName), e);
         }
-        String[] topologyKeys = topologyDef.trim().split("/");
-        int numOfmatchedKeys = 0;
-        for (String key : topologyKeys) {
-          if (!key.isEmpty()) {
-            // if a key does not exist in the instance domain config, apply the default domain value.
-            String value = domainAsMap.get(key);
-            if (value == null || value.length() == 0) {
-              value = DEFAULT_PATH_PREFIX + key;
-            } else {
-              numOfmatchedKeys++;
-            }
-            instanceTopologyMap.put(key, value);
+
+        if (domainAsMap == null || domainAsMap.isEmpty()) {
+          throw new IllegalArgumentException(String
+              .format("Domain for instance %s is not set, fail the topology-aware placement!",
+                  instanceName));
+        }
+        int numOfMatchedKeys = 0;
+        for (String key : clusterTopologyKeys) {
+          // if a key does not exist in the instance domain config, using the default domain value.
+          String value = domainAsMap.get(key);
+          if (value == null || value.length() == 0) {
+            value =
+                defaultDomainPathValuesCache.computeIfAbsent(key, k -> (DEFAULT_DOMAIN_PREFIX + key));
+          } else {
+            numOfMatchedKeys++;
           }
+          instanceTopologyMap.put(key, value);
         }
-        if (numOfmatchedKeys != domainAsMap.size()) {
+        if (numOfMatchedKeys != domainAsMap.size()) {
           logger.warn(
-              "Key-value pairs in InstanceConfig.Domain {} do not align with keys in ClusterConfig.Topology {}",
-              domain, topologyKeys.toString());
+              "Key-value pairs in InstanceConfig.Domain {} do not align with keys in ClusterConfig.Topology "
+                  + "{}, using default domain value instead", instanceConfig.getDomainAsString(),
+              clusterTopologyKeys.toString());
         }
       }
     } else {
-      instanceTopologyMap.put(Types.INSTANCE.name(), instance);
+      instanceTopologyMap.put(Types.INSTANCE.name(), instanceName);
     }
     return instanceTopologyMap;
   }
 
   /**
    * Add an end node to the tree, create all the paths to the leaf node if not present.
-   * Input param 'pathNameMap' must have a certain order where the order of the keys should be
-   * the same as the path order in ClusterConfig.topology.
    */
-  private void addEndNode(Node root, String instanceName, Map<String, String> pathNameMap,
+  private void addEndNode(Node root, String instanceName, LinkedHashMap<String, String> pathNameMap,
       int instanceWeight, List<String> liveInstances) {
     Node current = root;
     List<Node> pathNodes = new ArrayList<>();
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
@@ -55,6 +55,7 @@ public class Topology {
   private final List<String> _liveInstances;
   private final Map<String, InstanceConfig> _instanceConfigMap;
   private final ClusterConfig _clusterConfig;
+  private static final String DEFAULT_PATH_PREFIX = "Helix_default_";
 
   private String _faultZoneType;
   private String _endNodeType;
@@ -223,8 +224,7 @@ public class Topology {
                     instance));
           } else {
             // if the disabled instance missing ZONE setting, ignore it should be fine.
-            logger.warn(String
-                .format("ZONE_ID for instance %s is not set, ignore the instance!", instance));
+            logger.warn("ZONE_ID for instance {} is not set, ignore the instance!", instance);
             return false;
           }
         }
@@ -245,8 +245,7 @@ public class Topology {
                     instance));
           } else {
             // if the disabled instance missing domain setting, ignore it should be fine.
-            logger.warn(
-                String.format("Domain for instance %s is not set, ignore the instance!", instance));
+            logger.warn("Domain for instance {} is not set, ignore the instance!", instance);
             return false;
           }
         }
@@ -262,7 +261,11 @@ public class Topology {
         for (String key : topologyKeys) {
           if (!key.isEmpty()) {
             // if a key does not exist in the instance domain config, apply the default domain value.
-            instanceTopologyMap.put(key, domainAsMap.getOrDefault(key, "Helix_default_" + key));
+            String value = domainAsMap.get(key);
+            if (value == null || value.length()==0) {
+              value = DEFAULT_PATH_PREFIX + key;
+            }
+            instanceTopologyMap.put(key, value);
           }
         }
       }
@@ -274,8 +277,10 @@ public class Topology {
 
   /**
    * Add an end node to the tree, create all the paths to the leaf node if not present.
+   * Input param 'pathNameMap' must have a certain order where the order of the keys is
+   * the same as the path order in ClusterConfig.topology.
    */
-  private void addEndNode(Node root, String instanceName, LinkedHashMap<String, String> pathNameMap,
+  private void addEndNode(Node root, String instanceName, Map<String, String> pathNameMap,
       int instanceWeight, List<String> liveInstances) {
     Node current = root;
     List<Node> pathNodes = new ArrayList<>();
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
@@ -258,25 +258,18 @@ public class Topology {
          * Return a ordered map representing the instance path. The topology order is defined in
          * ClusterConfig.topology.
          */
-        Map<String, String> domainAsMap = new HashMap<>();
-        String[] pathPairs = instanceConfig.getDomainAsString().trim().split(",");
-        for (String pair : pathPairs) {
-          String[] values = pair.trim().split("=");
-          if (values.length != 2 || values[0].isEmpty() || values[1].isEmpty()) {
-            throw new IllegalArgumentException(String.format(
-                "Domain-Value pair %s for instance %s is not valid, failed the topology-aware placement!",
-                pair, instanceName));
-          }
-          domainAsMap.put(values[0], values[1]);
+        Map<String, String> domainAsMap = instanceConfig.getDomainAsMap();
+        if (domainAsMap.isEmpty()) {
+          throw new IllegalArgumentException(String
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +185,104 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-    return root;
+    return true;
   }
 
-
   /**
    * Add an end node to the tree, create all the paths to the leaf node if not present.
    */
-  private Node addEndNode(Node root, String instanceName, Map<String, String> pathNameMap,
+  private void addEndNode(Node root, String instanceName, LinkedHashMap<String, String> pathNameMap,
```
pkuwm, line 278: `LinkedHashMap` implements `Map` interface. I suggest keeping `Map`.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +185,104 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
     root.setType(Types.ROOT.name());
 
-    for (String ins : _allInstances) {
-      InstanceConfig config = _instanceConfigMap.get(ins);
-      Map<String, String> pathValueMap = new HashMap<>();
-      if (_topologyAwareEnabled) {
-        String zone = config.getZoneId();
+    for (String instance : _allInstances) {
+      InstanceConfig insConfig = _instanceConfigMap.get(instance);
+      LinkedHashMap<String, String> instanceTopologyMap = new LinkedHashMap<>();
```
pkuwm, line 197: I suggest using `Map` interface as type.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +185,104 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
+          if (instanceConfig.getInstanceEnabled() && (clusterConfig.getDisabledInstances() == null
+              || !clusterConfig.getDisabledInstances().containsKey(instance))) {
+            // if enabled instance missing domain information, fails the rebalance.
+            throw new HelixException(String
+                .format("Domain for instance %s is not set, fail the topology-aware placement!",
+                    instance));
+          } else {
+            // if the disabled instance missing domain setting, ignore it should be fine.
+            logger.warn(
+                String.format("Domain for instance %s is not set, ignore the instance!", instance));
```
pkuwm, line 249: Instead of `String.format`, I suggest parameterized logging: `logger.warn("Domain for instance {} is not set, ignore the instance!", instance);`
`String.format` is slower. And parameterized logging eliminates string concatenation if WARN level logging is not enabled.
Same for the other logging you changed in this PR.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +185,104 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-
-      String[] pathPairs = domain.trim().split(",");
-      Map<String, String> pathValueMap = new HashMap<>();
-      for (String pair : pathPairs) {
-        String[] values = pair.trim().split("=");
-        if (values.length != 2 || values[0].isEmpty() || values[1].isEmpty()) {
+        Map<String, String> domainAsMap;
+        try {
+          domainAsMap = instanceConfig.getDomainAsMap();
+        } catch (IllegalArgumentException e) {
```
pkuwm, line 256: In what case the `IllegalArgumentException` is thrown from `getDomainAsMap`? Is this same behavior with the old condition check?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +185,104 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-        if (!_types.contains(type)) {
-          logger.warn(String
-              .format("Path %s defined in domain of instance %s not recognized, ignored!", pair,
-                  ins));
-          continue;
+        String[] topologyKeys = topologyDef.trim().split("/");
+        for (String key : topologyKeys) {
+          if (!key.isEmpty()) {
+            // if a key does not exist in the instance domain config, apply the default domain value.
+            instanceTopologyMap.put(key, domainAsMap.getOrDefault(key, "Helix_default_" + key));
```
pkuwm, line 265: I suggest making "Helix_default_" a constant because the constant eliminates redundant string objects: constant is only one string object but this would create one string each time.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Please correct me if I am wrong. I think Map does not guarantee the order in which pairs were inserted into the map. When constructing the topology tree, it is important to maintain the original order for the path.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Same as above.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: TFTR. Will update.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: GetDomainAsMap will throw IllegalArgumentException if 
1. The Splitter.on() param matches the empty string.  
2. Has duplicate Key.  
For example the controller does not know how to categorize the instance if user specified two different rack.

There is one behavior change (witch I should have documented in the pr description) is that, GetDomainAsMap won't throw exception if either key or value is an empty string. I think the new behavior is reasonable since
1. We ignore keys in DOMAIN that are not defined in ClusterConfig. When we have empty key, they should be treated as non-defined key. Instead of throw exception, it is ok to just ignore them. They are no harm to the topology tree anyway.
2. We will use default values for keys that are defined in ClusterConfig but missing InstanceConfig.DOMAIN. If there is an empty value defined, we should use the default value as well.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: TFTR. Fixed.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +186,106 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-          continue;
+        String[] topologyKeys = topologyDef.trim().split("/");
+        for (String key : topologyKeys) {
+          if (!key.isEmpty()) {
+            // if a key does not exist in the instance domain config, apply the default domain value.
+            String value = domainAsMap.get(key);
+            if (value == null || value.length()==0) {
+              value = _defaultPathPrefix + key;
+            }
+            instanceTopologyMap.put(key, value);
```
zhangmeng916, line 268: I'd like to know the rationale of setting default value. Basically if the "domain" field is not set, we throw an exception, but if the "domain" is set with a random value, we think it is fine and give it a default zone value. How do we know it doesn't cause problem for rebalance.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -55,16 +55,10 @@
   private final List<String> _liveInstances;
   private final Map<String, InstanceConfig> _instanceConfigMap;
   private final ClusterConfig _clusterConfig;
-  private final boolean _topologyAwareEnabled;
+  private final String _defaultPathPrefix = "Helix_default_";
```
narendly, line 58: Please review the convention for defining constant Strings in Java.

https://stackoverflow.com/questions/9639007/defining-constant-string-in-java

With that said, is this prefix necessary?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +186,106 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
     root.setType(Types.ROOT.name());
 
-    for (String ins : _allInstances) {
-      InstanceConfig config = _instanceConfigMap.get(ins);
-      Map<String, String> pathValueMap = new HashMap<>();
-      if (_topologyAwareEnabled) {
-        String zone = config.getZoneId();
+    for (String instance : _allInstances) {
+      InstanceConfig insConfig = _instanceConfigMap.get(instance);
+      LinkedHashMap<String, String> instanceTopologyMap = new LinkedHashMap<>();
```
narendly, line 198: @xyuanlu 
Let's use a Map to declare?

And what you said about LinkedHashMap is true, but Map is a higher-level construct (interface) and LinkedHashMap is an implementation. Even if you wanted a LinkedHashMap instance, how do you guarantee that the map has the right order? If you want to be 100% sure about the order, you have two options:

1. Make it crystal clear that the user must pass in ordered map (TreeMap or LinkedHashMap)
2. do your own sorting internally.

With that said, what you said about using LinkedHashMap for declaration implies that maybe we should review the difference between declaration and implementation. Here's a helpful explanation: https://stackoverflow.com/questions/11715485/what-is-the-difference-between-declaration-and-definition-in-java

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +186,106 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
+          weight = DEFAULT_NODE_WEIGHT;
+        }
+        addEndNode(root, instance, instanceTopologyMap, weight, _liveInstances);
+      }
+    }
+    return root;
+  }
+
+  private boolean computeInstanceTopologyMap(ClusterConfig clusterConfig, String instance,
+      InstanceConfig instanceConfig, LinkedHashMap<String, String> instanceTopologyMap) {
```
narendly, line 211: Use Map instead...

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
narendly: Let's use `Map`? Making the function signature LinkedHashMap doesn't necessarily guarantee the map has the right order.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: TFTR.
Please correct me if I am wrong. As a general rule, we're supposed use its interface in declaration rather than the actual implementation like the following.
`Map<String, String> instanceTopologyMap = new LinkedHashMap<>();`
The reason behind this is that the implementation of these interfaces is usually not relevant when handling them in the following logic. However, in my understanding, in some special cases like this, if we need a map when order is important, it is better to use a more specific interface or an implementation to explicitly state that the order is important. Or maybe still use the general interface with comment?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: TFTR.
That's a good point. We probably should revisite the old sanity check logic and modify if needed.
One minor correcting. If the "domain" is set with a random value with no k-v split identifier found ('=' in this case), we will throw IllegalArgumentException.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: For private function like this, can we assume caller will strictly follow the instruction (of course with more detailed comment added) and guarantee the map has the right order? Or we could do one of the following to guarantee order:
1. pass in a separate array (or LinkedHashSet as line 63 in the old code) to keep the order of the key
2. add a lamda function to sorting internally. 

THX

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: I am planning to extract this function into a public util function.  It's better to explicitly state that the order is important or maybe still use the general map interface with comment?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
narendly: Yes, what you said about using the interface is on point.
I understand where you are coming from - you want to make it explicit that the parameter that's being passed in is ordered. But let's revisit what I said above - making the declaration LinkedHashMap does not help you in that case.

LinkedHashMap's ordering is not deterministic. It does not enforce the right ordering you need in the map. So you aren't really enforcing anything there other than the map will have some order.

If you want to enforce some sort of ordering of elements, you should sort internally. Say you need it sorted by alphabetical order, then you should convert this map into a TreeMap. 

Alternatively, stating that the topology map must be ordered in some way is also acceptable in JavaDoc, but in that case, the right thing to do is to be able to detect if the ordering has a problem, be able to throw an exception of some sort, alerting the user.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
narendly: That is acceptable, but what is this order exactly? How should users order the map when passing topology map in?

Would it be better to sort it yourself - would that be easier for the user and be safer?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
narendly: If it's a private method, then you can explicitly state that in the JavaDoc and use Map to declare. We want to avoid unnecessary sorting/computation if we can be sure that the parameters are well-formed.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: This is an out parameter. So the order will be kept when k-v pairs are added to the LinkedHashMap in this function. Forcing the caller to pass in a LinkedHashMap will  guarantee the order in which pairs were inserted into the map when user use the map later.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: For this in class variable defined here in a private function for class constructor, i guess it is safe to assume the order will be maintained as expected  since this private function has the full control of this object. Of course, more detailed comment/doc should be added.

I think it is safe to use the interface map to declare a in class object here. The public util function(computeInstanceTopologyMap) however, takes this LinkedHashMap instance, populate it and return as an out param.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -55,16 +55,10 @@
   private final List<String> _liveInstances;
   private final Map<String, InstanceConfig> _instanceConfigMap;
   private final ClusterConfig _clusterConfig;
-  private final boolean _topologyAwareEnabled;
+  private static final String DEFAULT_PATH_PREFIX = "Helix_default_";
```
dasahcc, line 58: Helix usually have all capitalized strings as our naming convention. Is this newly introduced or exists before?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -81,62 +75,42 @@ public Topology(final List<String> allNodes, final List<String> liveNodes,
...
       String topologyDef = _clusterConfig.getTopology();
       if (topologyDef != null) {
         // Customized cluster topology definition is configured.
-        String[] types = topologyDef.trim().split("/");
-        for (int i = 0; i < types.length; i++) {
-          if (types[i].length() != 0) {
-            _types.add(types[i]);
+        String[] topologyStr = topologyDef.trim().split("/");
+        int lastValidTypeIdx = topologyStr.length - 1;
+        for (; lastValidTypeIdx >= 0; --lastValidTypeIdx) {
```
dasahcc, line 86: Why not change this to be while clause since this already lost advantage of for loop for instantiation.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: TFTR.

I did not change the prefix. Please refer to line 105 in the old code.

104          for (String type : _types) {
105            _defaultDomainPathValues.put(type, "Helix_default_" + type);
106            lastType = type;
107          }

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Will update. 
THX

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: I changed the out param to a return value. It would be more clear and the definition in the caller can be changed to Map interface.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: There will be a following PR for this issue.  May I suggest using this PR solely for code clean up? We probably should schedule a discussion to revisit the old sanity check logic and modify if needed.
THX

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -81,62 +75,43 @@ public Topology(final List<String> allNodes, final List<String> liveNodes,
       throw new HelixException(String.format("Config for instances %s is not found!",
           _allInstances.removeAll(_instanceConfigMap.keySet())));
     }
-
     _clusterConfig = clusterConfig;
-    _types = new LinkedHashSet<>();
```
jiajunwang, line 86: I would prefer to keep it. A LinkedHashSet or ArrayList would be much easier to handle compared with array.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +187,129 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
         }
-        String type = values[0];
-        String value = values[1];
-
-        if (!_types.contains(type)) {
-          logger.warn(String
-              .format("Path %s defined in domain of instance %s not recognized, ignored!", pair,
-                  ins));
-          continue;
+        String[] topologyKeys = topologyDef.trim().split("/");
```
jiajunwang, line 274: Could you just input the String[] topologyStr (or a LinkedHashSet which I would recommend) to this method so you don't do the calculation again?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +187,129 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
+              value = DEFAULT_PATH_PREFIX + key;
+            } else {
+              numOfmatchedKeys++;
+            }
+            instanceTopologyMap.put(key, value);
+          }
+        }
+        if (numOfmatchedKeys != domainAsMap.size()) {
+          logger.warn(
+              "Key-value pairs in InstanceConfig.Domain {} do not align with keys in ClusterConfig.Topology {}",
```
jiajunwang, line 290: Mention here that the default domain value has been used. Otherwise, this warning looks like a major problem.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +187,129 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-          throw new HelixException(String
-              .format("Domain for instance %s is not set, failed the topology-aware placement!",
-                  ins));
-        } else {
-          // if the disabled instance missing domain setting, ignore it should be fine.
-          logger
-              .warn(String.format("Domain for instance %s is not set, ignore the instance!", ins));
-          continue;
+  private LinkedHashMap<String, String> computeInstanceTopologyMap(ClusterConfig clusterConfig,
+      String instance, InstanceConfig instanceConfig) {
```
jiajunwang, line 224: instance -> instanceName

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -81,62 +75,43 @@ public Topology(final List<String> allNodes, final List<String> liveNodes,
...
-                    _faultZoneType, topologyDef));
-          }
-          _useDefaultTopologyDef = false;
+        }
+        _endNodeType = topologyStr[lastValidTypeIdx];
+        _faultZoneType = clusterConfig.getFaultZoneType();
+        if (_faultZoneType == null) {
+          _faultZoneType = _endNodeType;
+        }
+        if (Arrays.stream(topologyStr).noneMatch(type -> type.equals(_faultZoneType))) {
```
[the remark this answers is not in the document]
xyuanlu: Please correct me if I am wrong. 
Comparing to previous approach (set. contains()), the stream.noneMatch() does introduced a o(n)worst case complexity. 
However, since there is no new LinkedHashSet created, it saved a O(n) time for constructing the LinkedHashSet and a O(n) space complexity for the object itself. 

It is better to pass the already parsed array to computeInstanceTopologyMap so we do not compute again in line 274. However, I think we could just use the original array topologyStr, no need to introduce another LinkedHashSet.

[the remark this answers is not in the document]
xyuanlu: One more optimization. We could put this check in the else branch for line 97. If we already set the _faultZoneType to be the same as _endNodeType, then _faultZoneType must belongs to topologyStr.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +187,129 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-            // if the disabled instance missing ZONE setting, ignore it should be fine.
-            logger.warn(String
-                .format("ZONE_ID for instance %s is not set, failed the topology-aware placement!",
-                    ins));
-            continue;
-          }
-
+    for (String instance : _allInstances) {
+      InstanceConfig insConfig = _instanceConfigMap.get(instance);
+      Map<String, String> instanceTopologyMap =
```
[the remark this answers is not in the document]
xyuanlu: Shouldn't we use interface for declaration (as previous comments suggested)? 
Or in this case it's better to explicitly state the ordering is important?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: update.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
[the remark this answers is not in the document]
jiajunwang: Considering the case "/domain/rack///node/", you will need to process the array. So either you have a new array or a LinkedHashSet. The latter can save us some time here.
I was thinking about these 2 points together. And my conclusion is that keep the original logic is better : )

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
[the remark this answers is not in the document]
jiajunwang: If your later usage assumes anything about the iteration order, then we should make this Map explicitly an ordered one.
And addEndNode() will require an ordered map. I suggest you change the addEndNode() signature so others won't be able to pass any non-ordered map. With that change, you have to change this Map to LinkedHashMap here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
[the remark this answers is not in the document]
xyuanlu: Had an offline discussion. Although the time consumed here in the constructor is the same, it is better to have a smaller loop with no empty string when creating the domain k-v map for all instance. Having extra empty string in the for loop may introduce performance issue considering the number of instance.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +187,129 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
+      instanceTopologyMap.put(Types.INSTANCE.name(), instance);
     }
-    return root;
+    return instanceTopologyMap;
   }
 
-
   /**
    * Add an end node to the tree, create all the paths to the leaf node if not present.
+   * Input param 'pathNameMap' must have a certain order where the order of the keys should be
```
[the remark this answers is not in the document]
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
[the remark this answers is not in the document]
xyuanlu: updated

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +187,129 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
+      String instance, InstanceConfig instanceConfig) {
+    LinkedHashMap<String, String> instanceTopologyMap = new LinkedHashMap<>();
+    if (clusterConfig.isTopologyAwareEnabled()) {
+      String topologyDef = clusterConfig.getTopology();
+      if (topologyDef == null) {
+        // Return a ordered map using default cluster topology definition, i,e. /root/zone/instance
+        String zone = instanceConfig.getZoneId();
+        if (zone == null) {
+          // we have the hierarchy style of domain id for instance.
+          if (instanceConfig.getInstanceEnabled() && (clusterConfig.getDisabledInstances() == null
```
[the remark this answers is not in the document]
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +187,129 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
+  private LinkedHashMap<String, String> computeInstanceTopologyMap(ClusterConfig clusterConfig,
+      String instance, InstanceConfig instanceConfig) {
+    LinkedHashMap<String, String> instanceTopologyMap = new LinkedHashMap<>();
+    if (clusterConfig.isTopologyAwareEnabled()) {
+      String topologyDef = clusterConfig.getTopology();
+      if (topologyDef == null) {
+        // Return a ordered map using default cluster topology definition, i,e. /root/zone/instance
+        String zone = instanceConfig.getZoneId();
+        if (zone == null) {
+          // we have the hierarchy style of domain id for instance.
```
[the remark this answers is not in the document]
xyuanlu: Updated

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +187,129 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-        String[] values = pair.trim().split("=");
-        if (values.length != 2 || values[0].isEmpty() || values[1].isEmpty()) {
+        instanceTopologyMap.put(Types.ZONE.name(), zone);
+        instanceTopologyMap.put(Types.INSTANCE.name(), instance);
+      } else {
+        /*
+         * Return a ordered map representing the instance path. The topology order is defined in
+         * ClusterConfig.topology.
+         */
+        String domain = instanceConfig.getDomainAsString();
```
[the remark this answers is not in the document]
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -81,62 +78,43 @@ public Topology(final List<String> allNodes, final List<String> liveNodes,
...
-            _faultZoneType = _endNodeType;
-          }
-          if (!_types.contains(_faultZoneType)) {
-            throw new HelixException(String
-                .format("Invalid fault zone type %s, not present in topology definition %s.",
-                    _faultZoneType, topologyDef));
-          }
-          _useDefaultTopologyDef = false;
+        }
+        _endNodeType = topologyKeys[lastValidTypeIdx];
```
jiajunwang, line 99: Does "_clusterTopologyKeys.get(_clusterTopologyKeys.size() - 1)"  works the same? But you won't need lastValidTypeIdx in that way.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +190,124 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-                .format("ZONE_ID for instance %s is not set, failed the topology-aware placement!",
-                    ins));
-            continue;
-          }
-
+    // TODO: Currently we add disabled instance to the topology tree. Since they are not considered
+    // TODO: in relabalnce, maybe we should skip adding them to the tree for consistence.
+    for (String instanceName : _allInstances) {
+      InstanceConfig insConfig = _instanceConfigMap.get(instanceName);
+      LinkedHashMap<String, String> instanceTopologyMap;
```
jiajunwang, line 204: nit, just put in the try block. It is not used outside.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -81,62 +78,43 @@ public Topology(final List<String> allNodes, final List<String> liveNodes,
...
           }
         }
-        if (_types.size() == 0) {
-          logger.error("Invalid cluster topology definition " + topologyDef);
+        if (_clusterTopologyKeys.size() == 0) {
           throw new HelixException("Invalid cluster topology definition " + topologyDef);
-        } else {
-          String lastType = null;
-          for (String type : _types) {
-            _defaultDomainPathValues.put(type, "Helix_default_" + type);
```
jiajunwang, line 105: Where do you update this map? I didn't see in your latest version.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +190,124 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
         }
-        String type = values[0];
-        String value = values[1];
 
-        if (!_types.contains(type)) {
-          logger.warn(String
-              .format("Path %s defined in domain of instance %s not recognized, ignored!", pair,
-                  ins));
-          continue;
+        if (domainAsMap == null || domainAsMap.isEmpty()) {
```
jiajunwang, line 270: It won't be null, I guess. Don't overprotect the code, it may hide issues.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +190,124 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-          // if the disabled instance missing domain setting, ignore it should be fine.
-          logger
-              .warn(String.format("Domain for instance %s is not set, ignore the instance!", ins));
-          continue;
+  private LinkedHashMap<String, String> computeInstanceTopologyMap(boolean isTopologyAwareEnabled,
+      String instanceName, InstanceConfig instanceConfig, LinkedHashSet<String> clusterTopologyKeys,
+      Map<String, String> defaultDomainPathValuesCache) throws IllegalArgumentException {
+    LinkedHashMap<String, String> instanceTopologyMap = new LinkedHashMap<>();
+    if (isTopologyAwareEnabled) {
+      if (clusterTopologyKeys == null || clusterTopologyKeys.size() == 0) {
```
jiajunwang, line 246: It won't be null. Don't overprotect the code, it may hide issues.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +190,124 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
+              .format("Domain for instance %s is not set, fail the topology-aware placement!",
+                  instanceName));
+        }
+        int numOfMatchedKeys = 0;
+        for (String key : clusterTopologyKeys) {
+          // if a key does not exist in the instance domain config, using the default domain value.
+          String value = domainAsMap.get(key);
+          if (value == null || value.length() == 0) {
+            value =
+                defaultDomainPathValuesCache.computeIfAbsent(key, k -> (DEFAULT_DOMAIN_PREFIX + key));
```
jiajunwang, line 281: I see you want a lazy initialization. It's good design.
But since you are modifying the input parameter implicitly, it might be clearer to just refer to _defaultDomainPathValues. Otherwise, there is an implicit assumption that the caller needs to pass the same map for all the calls.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +190,124 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-        pathValueMap.put(type, value);
-      }
-
-      int weight = insConfig.getWeight();
-      if (weight < 0 || weight == InstanceConfig.WEIGHT_NOT_SET) {
-        weight = DEFAULT_NODE_WEIGHT;
       }
-      root = addEndNode(root, ins, pathValueMap, weight, _liveInstances);
+    } else {
+      instanceTopologyMap.put(Types.INSTANCE.name(), instanceName);
```
jiajunwang, line 295: Add the corresponding comment as you did for the other 2 conditions above.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
jiajunwang: I talked with Xiaoyuan, since the method logic requires the input map has a fixed iterate order, I suggest her to use the explicit type. Please note that if you input a Map instantiate by HashMap, then the method won't work as expected!

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: I am not sure if there is a get() function for LinkedHashSet.

https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashSet.html#method.summary

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated

xyuanlu:
I ran a CPU profiler for my previous version (34bf638). It shows the lamda function and InstanceConfig.getDomaiAsMap are two hotspots for CPU consumption. I changed them back to the original implementation. 
After these changes, the new result of running testCreateClusterTopology is ~2%, comparing to 
 2~3% for the original code and 6% with previous version.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: I changed it back to initialing the map in the constrictor. CPU profiler shows the overhead of having a lamda function here is too heavy.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +191,122 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-              "Domain-Value pair %s for instance %s is not valid, failed the topology-aware placement!",
-              pair, ins));
+        instanceTopologyMap.put(Types.ZONE.name(), zone);
+        instanceTopologyMap.put(Types.INSTANCE.name(), instanceName);
+      } else {
+        /*
+         * Return a ordered map representing the instance path. The topology order is defined in
+         * ClusterConfig.topology.
+         */
+        Map<String, String> domainAsMap = new HashMap<>();
```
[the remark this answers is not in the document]
xyuanlu: I kept the original logic (before code clean up) here. One difference is that getDomainAsMap wont throw exception if key or value is empty.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -212,123 +191,122 @@ private static Node cloneTree(Node root, Map<Node, Integer> newNodeWeight, Set<N
...
-              .format("Path %s defined in domain of instance %s not recognized, ignored!", pair,
-                  ins));
-          continue;
+        int numOfMatchedKeys = 0;
+        for (String key : clusterTopologyKeys) {
+          // if a key does not exist in the instance domain config, using the default domain value.
+          String value = domainAsMap.get(key);
+          if (value == null || value.length() == 0) {
+            value = _defaultDomainPathValues.get(key);
+                //defaultDomainPathValuesCache.computeIfAbsent(key, k -> (DEFAULT_DOMAIN_PREFIX + key));
```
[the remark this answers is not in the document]
xyuanlu: Forgot to remove the comment. Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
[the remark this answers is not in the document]
jiajunwang: You can check once you get the map in the same way. Please try to avoid duplicate logic.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: I kept the original implementation for performance.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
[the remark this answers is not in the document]
xyuanlu: As our offline discussion. I updated the InstanceConfig.getDomainAsMap.

xyuanlu:

Code clean up -- Topology.java

helix-core/src/main/java/org/apache/helix/model/InstanceConfig.java
```diff
@@ -150,12 +150,22 @@ public String getDomainAsString() {
...
+    }
+
+    String[] pathPairs = domain.trim().split(",");
+    for (String pair : pathPairs) {
+      String[] values = pair.trim().split("=");
+      if (values.length != 2 || values[0].isEmpty() || values[1].isEmpty()) {
+        throw new IllegalArgumentException(
+            String.format("Domain-Value pair %s is not valid.", pair));
+      }
+      domainAsMap.put(values[0], values[1]);
```
jiajunwang, line 165: We shall trim at the final values level but not the above layers. What do you think?

jiajunwang:
Just noticed one more issue. Please check.

Moreover, please specify your final commit description. We will squash and merge all commits into one and use that description.

helix-core/src/main/java/org/apache/helix/model/InstanceConfig.java
xyuanlu: I think if the value or key is null, then this pair shouldn't be count as a valid input. 
Since currently the callers of this function only checks if key exists in the returned map and uses the corresponding value. So i think the user all assume all k-v pairs in  returned map are valid. 
Thus, I think we should throw exception for empty value. (Original version already handles no splitter and null key errors).

xyuanlu:
TFTR. Updated.

helix-core/src/main/java/org/apache/helix/model/InstanceConfig.java
jiajunwang: Consider this case:
"a = b, c = d", the current logic will have a map of {"a ": " b", "c " : " d"} (notice the space), and what we expect is {"a": "b", "c" : "d"}.

xyuanlu: Oh I got what you are saying. Yes thats correct. Will update.

xyuanlu:

Code clean up -- Topology.java
This change did come code refactor work for Topology.java. It extracts the sanity check logic in multiple functions into one single function so it can be used else where as well.

Also, mvn test result is updated as well.

PR #1056, opened 2020-06-03
(WIP) Task delay metrics endpoints on TaskDriver

### LINKED ISSUE #1020: Support configurable task thread pool size for Helix REST
In order to support configurable task thread pool size, we are adding endpoints to Helix REST to support relevant operations. The endpoints will be added to ClusterAccessor and PerInstanceAccessor as the operations are performed on ClusterConfig, InstanceConfig and LiveInstance.

### LINKED ISSUE #1055: Create endpoints on TaskDriver for task delay metrics
We are adding 3 endpoints to TaskDriver that return values of `SubmissionToProcessDelay`, `SubmissionToScheduleDelay`, and `ControllerInducedDelay`. The purpose of the endpoints is to offer an option for users that cannot rely directly on the controller emitted metrics to see the metric values. In order for the metrics to be accessible from TaskDriver, the metric values are written to ZooKeeper every time a new metric value is reported. The metric values that are synced to Zk is the mean value. Note: since they are synced on reporting time, if there is no reporting, there is no update to these values. These endpoint values are fresh when there are task framework resources actively running.

NealSun96:
### Issues

Fixes #1055

### Description

We are adding 3 endpoints to TaskDriver that return values of `SubmissionToProcessDelay`, `SubmissionToScheduleDelay`, and `ControllerInducedDelay`. The purpose of the endpoints is to offer an option for users that cannot rely directly on the controller emitted metrics to see the metric values. In order for the metrics to be accessible from TaskDriver, the metric values are written to ZooKeeper every time a new metric value is reported. The metric values that are synced to Zk is the mean value. Note: since they are synced on reporting time, if there is no reporting, there is no update to these values. These endpoint values are fresh when there are task framework resources actively running. 

### Tests

TestJobMonitor, testGetSubmissionToProcessDelay, testGetSubmissionToProcessDelayIllegalArgument, testGetSubmissionToScheduleDelay, testGetSubmissionToScheduleDelayIllegalArgument, testControllerInducedDelay, testControllerInducedDelayIllegalArgument

```
[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 1178, Failures: 2, Errors: 0, Skipped: 1
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:16 h
[INFO] Finished at: 2020-06-02T19:20:23-07:00

```

Rerun:
```
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 38.517 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  43.765 s
[INFO] Finished at: 2020-06-03T09:14:54-07:00
```

### Commits

### Code Quality


jiajunwang:
I have 2 questions,
1. What is the expected ZK throughput? It sounds a very scary idea. Note that ZK writes traffic is the current performance bottleneck.
2. If we are doing this, I suggest doing this as an additional metric reporting channel outside the monitor classes, or even outside the Helix lib. MBean is a generic way to report metrics. To expose that data to any other metric collection system or to ZK is not related to the MBeans. Adding this option and extra logic into the monitors just increase the unnecessary complexity of those classes.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/JobMonitor.java
```diff
@@ -41,6 +45,12 @@
   private static final Logger LOG = LoggerFactory.getLogger(JobMonitor.class);
   private static final long DEFAULT_RESET_INTERVAL_MS = 60 * 60 * 1000; // 1 hour
 
+  // The root path of ZNodes that are used to sync certain metrics to ZooKeeper. This path is the
+  // first level, and its children nodes will represent each job type. Since the metrics are
+  // aggregated by job types, the metric values will be recorded in each ZNode belonging to a job
+  // type as simple fields.
+  private static final String SYNC_ZK_ROOT_PATH = "/JOB_MONITOR_METRICS";
```
dasahcc, line 52: Why we need a ZNode to persist metrics data?

pkuwm: Is there a reason we don't use MBean for such metrics?

NealSun96: Right now the metrics are emitted by the controller. We want to expose the metrics to the client side, therefore ZNodes are used as a medium to store the metrics for access. 

MBean is still used; the ZNodes are for "syncing" or "copying" the metrics to client side.

dasahcc: It's now necessary for doing in this way:
1. These metrics can be done in controller by looking at the stats of the ZNode property.
2. Even if you would like to have it in client side. You should make it reported in MBean since all the data can be derived from config/context. So ingraph can fetch MBean to report it. This is not a good idea to have metrics reporting to introduce new ZNode create/read/write.

jiajunwang: I'm not trying to say no to this idea before I fully understand the context here. But,
1. Who are this "We"?
2. Do we have a design that illustrates the motivation and potential risks? I would like to take a look and see if there are any other ways.

NealSun96:
> I have 2 questions,
> 
> 1. What is the expected ZK throughput? It sounds a very scary idea. Note that ZK writes traffic is the current performance bottleneck.
> 2. If we are doing this, I suggest doing this as an additional metric reporting channel outside the monitor classes, or even outside the Helix lib. MBean is a generic way to report metrics. To expose that data to any other metric collection system or to ZK is not related to the MBeans. Adding this option and extra logic into the monitors just increase the unnecessary complexity of those classes.

1. Unfortunately I can't answer this question with confidence - as noted in the design doc, @narendly believes this is not a major performance concern; without much context on performance bottleneck myself, I trust @narendly's assessment. 

In terms of throughput, the syncing happens every time a data point is added to this metric. 

Lastly, to achieve the goal of "providing metrics on client side", this is the only method I can think of: using ZK. Is there any other good ways to achieve this goal?

2. There isn't any means to access the gauges outside JobMonitor, which is why I placed the logic inside JobMonitor. I don't think adding support for JobMonitor to optionally sync values to Zk is "complexity"; it's just an option. 

But if you insist, I'm happy to leave the logic to an outer class. I will need to have JobMonitor providing access to the gauges.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/JobMonitor.java
pkuwm: I would not think it is a good idea to use ZK as syncing store for metrics.. ZK is not designed for this purpose. ZK is such a core part of infra. Syncing metrics to ZK definitely increase read/write traffic to ZK, which may impact the other critical services. If we have a better design not to use ZK as a syncing store, exposing metrics to client is fine.

Could you explain any case "users that cannot rely directly on the controller emitted metrics to see the metric values."?  I would like to see if there is an alternative solution that doesn't use ZK.

jiajunwang:
> > I have 2 questions,
> > 
> > 1. What is the expected ZK throughput? It sounds a very scary idea. Note that ZK writes traffic is the current performance bottleneck.
> > 2. If we are doing this, I suggest doing this as an additional metric reporting channel outside the monitor classes, or even outside the Helix lib. MBean is a generic way to report metrics. To expose that data to any other metric collection system or to ZK is not related to the MBeans. Adding this option and extra logic into the monitors just increase the unnecessary complexity of those classes.
> 
> 1. Unfortunately I can't answer this question with confidence - as noted in the design doc, @narendly believes this is not a major performance concern; without much context on performance bottleneck myself, I trust @narendly's assessment.
>    In terms of throughput, the syncing happens every time a data point is added to this metric.
>    Lastly, to achieve the goal of "providing metrics on client side", this is the only method I can think of: using ZK. Is there any other good ways to achieve this goal?
> 2. There isn't any means to access the gauges outside JobMonitor, which is why I placed the logic inside JobMonitor. I don't think adding support for JobMonitor to optionally sync values to Zk is "complexity"; it's just an option.
>    But if you insist, I'm happy to leave the logic to an outer class. I will need to have JobMonitor providing access to the gauges.

1. Then the problem is how you can convince us : ) This current explanation is not satisfying, to be frank. Please schedule a meeting so we can talk offline.
2. Please check how we are using MBeanServer in our code base. That is the way. "complexity" means,
     - The JobMonitor will be a larger file. And there are 2 systems of reporting mechanism inside one class.
     - MBean is the generic mean for Java process to report it's status. The emitting part should be built above that. If you mixed emitting part with the MBean, it definitely makes it harder to understand.
    - If you want to add this type of support to the other monitors, you will need to do this again. However, if you do this outside as service, then you just need to register your MBean there. I guess no more than 10 lines of change given the ZK based metric emitting service is designed well.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/JobMonitor.java
narendly, line 52: Let's review whether creating this root path would be appropriate. Can you find a better place to persist the data?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/JobMonitor.java
```diff
@@ -182,6 +198,59 @@ public void updateControllerInducedDelayGauge(long delay) {
     _controllerInducedDelayGauge.updateValue(delay);
   }
 
+  /**
+   * Sync the current SubmissionToProcessDelay mean to ZooKeeper
+   * @param baseDataAccessor
+   */
+  public void syncSubmissionToProcessDelayToZk(BaseDataAccessor<ZNRecord> baseDataAccessor) {
```
narendly, line 205: As discussed offline, this implementation could be improved.
- This is too specific and not very easy to maintain or scalable - suppose you want to add more metrics that you need to persist, then you'd have to implement methods for each additional metric.
- Can you look into inversion of control? Perhaps you could create a component with the appropriate set of interface methods that allow persisting of metric numbers to a metadata store in a more generic fashion. For example, you could have something like,

> <I> MetricStorage
> <Class> ZkMetricStorage (takes a Zk connection to initialize)

Also, this MetricStorage implementation could take in a dynamic list of metric names to be persisted, etc. You could get creative with this. No need to feel like you have to cover all use cases, but designing a component in a way that is extendable and generic enough will save you and help with maintainability.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -1240,16 +1240,18 @@ protected void finishJobInRuntimeJobDag(TaskDataCache clusterDataCache, String w
...
    * @param clusterStatusMonitor
    * @param workflowConfig
    * @param jobConfig
    * @param currentTimestamp
+   * @param helixManager
    */
   protected static void reportSubmissionToProcessDelay(BaseControllerDataProvider dataProvider,
       final ClusterStatusMonitor clusterStatusMonitor, final WorkflowConfig workflowConfig,
-      final JobConfig jobConfig, final long currentTimestamp) {
+      final JobConfig jobConfig, final long currentTimestamp, final HelixManager helixManager) {
```
narendly, line 1254: It's strongly recommend we check whether the given ZkConnection is valid. What if helixManager is null or not connected?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/JobMonitor.java
```diff
@@ -182,6 +198,59 @@ public void updateControllerInducedDelayGauge(long delay) {
...
+  public void syncControllerInducedDelayToZk(BaseDataAccessor<ZNRecord> baseDataAccessor) {
+    syncMetricMeanValueToZk(baseDataAccessor,
+        JobMonitorMetricZnodeField.CONTROLLER_INDUCED_PROCESS_DELAY, _controllerInducedDelayGauge);
+  }
+
+  private void syncMetricMeanValueToZk(BaseDataAccessor<ZNRecord> baseDataAccessor,
+      JobMonitorMetricZnodeField field, HistogramDynamicMetric metric) {
+    String zkPath = buildZkPathForJobMonitorMetric(_jobType);
+
+    if (!baseDataAccessor.update(zkPath, currentData -> {
```
narendly, line 232: Is there a reason we're doing an update instead of set?

NealSun96:
After discussion, this work item will not be done.

PR #1058, opened 2020-06-04
Fix ReadOnlyWagedRebalancer so that it computes mapping from scratch

### LINKED ISSUE #1057: ReadOnlyWagedRebalancer only returns BestPossibleMapping in ZK
ReadOnlyWagedRebalancer only returns BestPossibleMapping in ZK

This is an issue because if the mapping was not there in ZK, the waged rebalancer mapping calculation tool would always return nothing. It should do the calculation from scratch.

narendly:
### Issues

Fixes #1057

### Description

Previously, ReadOnlyWagedRebalancer would only read from the previously computed best possible mapping and returns it. This commit changes it so that it computes things from scratch - it can read the previously computed best possible mapping but shouldn't just return it without doing any calculation.

### Tests

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestEnableCompression.testEnableCompressionResource:116 Â» ThreadTimeout Method...
[ERROR]   TestAutoRebalanceWithDisabledInstance.testDisableEnableInstanceAutoRebalance:69 expected:<true> but was:<false>
[ERROR]   TestDeleteJobFromJobQueue.testForceDeleteJobFromJobQueue:75 Â» Helix Failed to ...
[INFO] 
[ERROR] Tests run: 1149, Failures: 3, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
```

Other tests succeed when run individually.

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
@@ -48,15 +48,6 @@ public class ReadOnlyWagedRebalancer extends WagedRebalancer {
         ConstraintBasedAlgorithmFactory.getInstance(preferences), Optional.empty());
   }
 
-  @Override
-  protected Map<String, ResourceAssignment> computeBestPossibleAssignment(
-      ResourceControllerDataProvider clusterData, Map<String, Resource> resourceMap,
-      Set<String> activeNodes, CurrentStateOutput currentStateOutput, RebalanceAlgorithm algorithm)
-      throws HelixRebalanceException {
-    return getBestPossibleAssignment(getAssignmentMetadataStore(), currentStateOutput,
-        resourceMap.keySet());
-  }
-
   private static class ReadOnlyAssignmentMetadataStore extends AssignmentMetadataStore {
     ReadOnlyAssignmentMetadataStore(String metadataStoreAddress, String clusterName) {
       super(new ZkBucketDataAccessor(metadataStoreAddress), clusterName);


jiajunwang:
To avoid introducing unnecessary instability into the verifier, can we have 2 classes. One for the verifier, one for the simulate API. The verifier class can be a child class or an anonymous class.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
```diff
@@ -428,4 +432,20 @@ public String toString() {
     return verifierName + "(" + _clusterName + "@" + _zkClient + "@resources["
        + (_resources != null ? Arrays.toString(_resources.toArray()) : "") + "])";
   }
+
+  private class ReadOnlyWagedRebalancer extends org.apache.helix.controller.rebalancer.waged.ReadOnlyWagedRebalancer {
```
jiajunwang, line 436: Let's have a different name for this private class. I would still prefer the old name DryrunWagedRebalancer.
But feel free to have a better name for this method.

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
```diff
@@ -175,27 +175,34 @@ public static String serializeByComma(List<String> objects) {
...
       ResourceControllerDataProvider dataProvider =
-          new ResourceControllerDataProvider(clusterConfig.getClusterName());
+          new ResourceControllerDataProvider(globalSyncClusterConfig.getClusterName());
       dataProvider.requireFullRefresh();
       dataProvider.refresh(helixDataAccessor);
-      dataProvider.setClusterConfig(clusterConfig);
+      dataProvider.setClusterConfig(globalSyncClusterConfig);
       dataProvider.setInstanceConfigMap(instanceConfigs.stream()
           .collect(Collectors.toMap(InstanceConfig::getInstanceName, Function.identity())));
       dataProvider.setLiveInstances(
```
jiajunwang, line 208: As we see in the test, the current way overriding the live instance info won't work. We need to retain the session Id somehow. Otherwise, the current state won't be read correctly.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
narendly: Okay. Changed to Dryrun

PR #1066, opened 2020-06-05
Prevent parallel controller pipelines run causing two master replicas

### LINKED ISSUE #1027: Gracefully fix the double-masters situation.
There are 2 known cases that may lead the partition to have 2 masters temporarily.
- Unclean helix controller leader switch, which leads to 2 pipelines running in parallel.
- Long ZK propagation latency causes the controller leader to see out-of-date information. Then the rebalance result may lead to 2 masters.

The rebalancer will fix the additional master eventually, but the default operations are arbitrary and it may cause the oldest master to be leftover. Since many application requires the master to have the latest data, this caused serious application issue.

We are trying to address 3 issues here:
1. Unclean controller leadership switch
2. Long ZK propagation latency caused out-of-date data read
3. To gracefully recover the abnormal states (give there are more possible causes that lead to the problematic states).

pkuwm:
### Issues

Fixes #1027 

### Description

There is a case that after controller leader switches, pipelines of both old Helix controller leader and new leader are running in parallel. Different assignment decisions are sent to different participants so there are double masters for a single partition.

The PR addresses this issue by blocking non-leader controller to send messages.

### Tests

- `testNoMessageSentOnControllerLosesLeadership`
- `TestRawZkClient.testAsyncCreateByInvalidSession`

```
[INFO] Tests run: 47, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 61.443 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 47, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:05 min
[INFO] Finished at: 2020-10-04T21:39:20-07:00


[WARNING] Tests run: 1211, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,282.946 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[WARNING] Tests run: 1211, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:11 h
[INFO] Finished at: 2020-10-06T04:14:23-07:00

```
### Commits

### Documentation (Optional)

### Code Quality


helix-core/src/main/java/org/apache/helix/controller/ControllerLeaderSession.java
```diff
@@ -0,0 +1,36 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * A thin wrapper to include a ZK session. This is created for controller leader to retrieve ZK
+ * session from Helix Manager in the mean time when checking leadership.
+ */
+public class ControllerLeaderSession {
```
narendly, line 26: Why do we need to wrap? Wouldn't AtomicReference work here?

pkuwm: `AtomicReference` would work, but it over qualifies. It guarantees the variable could be read and write **atomically**. It is designed for multi threading.

Here in our case, we only need a way to get the value set out of scope as below:
```
var name;
foo(name); // sets name = "helix" inside method
print(name); // prints "helix"
```
We don't need atomicity. `AtomicReference` would be heavier than this light wrapper as internally it needs more memory work to guarantee atomicity. So I go with the thin wrapper which satisfies us better.

helix-core/src/main/java/org/apache/helix/HelixManager.java
```diff
@@ -425,6 +426,19 @@ void addExternalViewChangeListener(org.apache.helix.ExternalViewChangeListener l
...
+  /**
+   * Checks if the cluster manager is leader and sets its ZK session in
+   * {@link ControllerLeaderSession}.
+   *
+   * @param controllerLeaderSession To include ZK session of the cluster manager in return
+   *
+   * @return true if this is a controller and a leader of the cluster. Zk session of the cluster
+   * manager is set in controllerLeaderSession
+   */
+  default boolean isLeader(ControllerLeaderSession controllerLeaderSession) {
```
lei-xia, line 438: This could be confusion because there is no "leader" concept for a Helix manager. Leader only applies to a controller instance, but HelixManager is a broader concept here.

pkuwm:
Got it. Thanks for pointing that out. I will get a better name for it.
Maybe "ZkSessionWrapper". If you have a better name, please let me know.

helix-core/src/main/java/org/apache/helix/HelixManager.java
pkuwm, line 438: > Got it. Thanks for pointing that out. I will get a better name for it. Maybe "ZkSessionWrapper". If you have a better name, please let me know.
> […](#)
> On Fri, Jun 5, 2020 at 12:52 PM Lei Xia ***@***.***> wrote: ***@***.**** commented on this pull request. ------------------------------ In helix-core/src/main/java/org/apache/helix/HelixManager.java <[#1066 (comment)](https://github.com/apache/helix/pull/1066#discussion_r436132217)>: > @@ -425,6 +426,19 @@ void addExternalViewChangeListener(org.apache.helix.ExternalViewChangeListener l */ boolean isLeader(); + /** + * Checks if the cluster manager is leader and sets its ZK session in + * ***@***.*** ControllerLeaderSession}. + * + * @param controllerLeaderSession To include ZK session of the cluster manager in return + * + * @return true if this is a controller and a leader of the cluster. Zk session of the cluster + * manager is set in controllerLeaderSession + */ + default boolean isLeader(ControllerLeaderSession controllerLeaderSession) { 

> This could be confusion because there is no "leader" concept for a Helix manager. Leader only applies to a controller instance, but HelixManager is a broader concept here.

Thinking in another aspect, the java doc also confuses. That's why I call it `ControllerLeaderSession`. Though we don't have a "leader" concept for helix manager, but a helix manager could be attached to a controller which could be a leader. And this is already public API, we could not change the name of the old API now. For this new one, maybe we could think of a new name. Maybe like `isInstanceLeader()`?

Let me also fix the java doc to avoid any confusion.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageDispatchStage.java
```diff
@@ -78,7 +78,13 @@ protected void processEvent(ClusterEvent event, MessageOutput messageOutput) thr
         batchMessage(dataAccessor.keyBuilder(), messagesToSend, resourceMap, liveInstanceMap,
             manager.getProperties());
 
-    List<Message> messagesSent = sendMessages(dataAccessor, outputMessages);
+    String expectedSession = event.getAttribute(AttributeName.CONTROLLER_LEADER_SESSION.name());
+    // An early check for expected leader session. If the sessions don't match, it means the
+    // controller lost leadership, then messages should not be sent. This potentially avoid
+    // double masters for a single partition.
+    List<Message> messagesSent = manager.getSessionId().equals(expectedSession)
```
dasahcc, line 85: Shall we throw an exception to break the pipeline? If we know session is not matched. Then there is no reason to keep following pipeline based on stale data.

pkuwm: I proposed to do some checks at stages like `ReadClusterDataStage` and terminate the remaining pipeline if controller loses leadership. In our issue, I see that leader lost leadership at `ReadClusterDataStage`. So there is no need to continue the pipeline and it would help save some computation resource or write traffic to EV/persist assignment. But in the discussion we agreed on not doing that and just blocking messages. We could definitely pick it up and discuss again. @jiajunwang

helix-core/src/main/java/org/apache/helix/HelixManager.java
```diff
@@ -425,6 +426,20 @@ void addExternalViewChangeListener(org.apache.helix.ExternalViewChangeListener l
...
+   *
+   * @param instanceLeaderSession To include ZK session ID of the cluster manager in return
+   *
+   * @return true if the instance is a leader of the cluster and Zk session of the cluster
+   * manager is return in param {@link InstanceLeaderSession}
+   */
+  default boolean isInstanceLeader(InstanceLeaderSession instanceLeaderSession) {
+    throw new UnsupportedOperationException(
+        "Checking leader and returning session is not supported.");
+  }
```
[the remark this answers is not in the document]
pkuwm: I was making it the same name `isLeader(session)`. @lei-xia reminded that helix manager does not have a controller concept so no leader/follower concept. I thought the instance itself managed by the helix manager is a controller. So I name it `isInstanceLeader(session)`. We could discuss and have an agreement. @jiajunwang Could you also take a look? Thanks.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageDispatchStage.java
pkuwm: I've updated it to stop the pipeline at this point: thrown StageException.

helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
```diff
@@ -95,6 +95,24 @@
...
+   * This will use async api for better performance. If the children already exist it will return
+   * false.
+   *
+   * @param paths the paths to the children ZNodes
+   * @param records List of data to write to each of the path
+   * @param options Set the type of ZNode see the valid values in {@link AccessOption}
+   * @param expectedSession The expected ZK session to create children
+   * @return For each child: true if creation succeeded, false otherwise (e.g. if the child exists)
+   */
+  default boolean[] createChildren(List<String> paths, List<T> records, int options,
```
jiajunwang, line 110: Instead of polluting the interface, shall we have a new interface called sessionAwareZkClient or something like that? Note that an implementation can implement multiple interfaces. So we just need to let the right class implement that interface. And this one won't need to have this ugly and not supported method.

helix-core/src/main/java/org/apache/helix/HelixDataAccessor.java
```diff
@@ -188,6 +188,22 @@
...
+   * Adds multiple children to a parent. If successful, the children will be created by the expected
+   * ZK session. If current ZK session does not match expected session, the creation operation will
+   * fail.
+   *
+   * @param keys property keys
+   * @param children list of children znodes to be created
+   * @param expectedSession expected ZK session to create the children znodes
+   * @return array where true means the child was added and false means it was not
+   */
+  default <T extends HelixProperty> boolean[] createChildren(List<PropertyKey> keys, List<T> children,
```
jiajunwang, line 201: same here

helix-core/src/main/java/org/apache/helix/HelixManager.java
```diff
@@ -425,6 +426,20 @@ void addExternalViewChangeListener(org.apache.helix.ExternalViewChangeListener l
...
+  /**
+   * Checks whether the cluster manager is leader and sets its ZK session in param
+   * {@link InstanceLeaderSession} if and only if it is leader.
+   *
+   * @param instanceLeaderSession To include ZK session ID of the cluster manager in return
+   *
+   * @return true if the instance is a leader of the cluster and Zk session of the cluster
+   * manager is returned in param {@link InstanceLeaderSession}
+   */
+  default boolean isInstanceLeader(InstanceLeaderSession instanceLeaderSession) {
```
jiajunwang, line 438: Same here

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
```diff
@@ -40,5 +40,6 @@
   PipelineType,
   LastRebalanceFinishTimeStamp,
   ControllerDataProvider,
-  STATEFUL_REBALANCER
+  STATEFUL_REBALANCER,
+  CONTROLLER_LEADER_SESSION
```
jiajunwang, line 44: This can be more generic. Just call it event_session?
As you can imagine, this can be used in multiple ways.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -401,6 +402,8 @@ void checkConnected(long timeout) {
       throw new HelixException(
           "HelixManager is not connected within retry timeout for cluster " + _clusterName);
     }
+
+    _sessionId = ZKUtil.toHexSessionId(_zkclient.getSessionId());
```
jiajunwang, line 406: This may cause some issues. If the session is updated randomly, a new session may be populated before new session handling is done.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2119,8 +2126,41 @@ private String getHexSessionId() {
    * 2. create mode is EPHEMERAL or EPHEMERAL_SEQUENTIAL
    */
   private boolean isSessionAwareOperation(String expectedSessionId, CreateMode mode) {
-    return expectedSessionId != null && !expectedSessionId.isEmpty() && (
-        mode == CreateMode.EPHEMERAL || mode == CreateMode.EPHEMERAL_SEQUENTIAL);
+    return expectedSessionId != null && !expectedSessionId.isEmpty() && mode.isEphemeral();
+  }
+
+  private ZooKeeper getExpectedZookeeper(final String expectedSessionId) {
```
jiajunwang, line 2132: This is a great idea!

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: I was waiting for you to point it out and discuss :) I had the same thought when I considered adding this assignment. _sessionId is only assigned in waitUntilConnected(). Each time handleNewSession() also calls waitUntilConnected(). handleNewSession(expectedSession) handles expected session. If the sessionId changes before handleNewSession() is done, I expect handleNewSession() should be able to handle the expired session because it is handling an expected session? Maybe I miss some corner cases? Let's carefully think about this here.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -901,6 +904,11 @@ public void addPreConnectCallback(PreConnectCallback callback) {
 
   @Override
   public boolean isLeader() {
+    return isInstanceLeader(null);
+  }
+
+  @Override
+  public boolean isInstanceLeader(InstanceLeaderSession instanceLeaderSession) {
```
jiajunwang, line 911: As we discussed,
1. make it private if not used anywhere else.
2. make it getSessionIdIfLeader() which returns an Optional object if the node is leader.

helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
pkuwm: Yes. The logic is: sendMessages() -> HelixDataAccessor.createChildren() -> BaseDataAccessor.createChildren() -> ZkClient.asyncCreate().

So we need the expectedSession to be passed all way down to zkclient from sendMessages().

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: I actually also tried to use this API in `create()` that we previous added. But inside `create()` there is a `isSessionAwareOperation()` check. If we don't need this check, we could just use the this method `getExpectedZookeeper`. And I actually don't think we need the check "isSessionAwareOperation()" since we only expose `createEphemeral()` to have expectedSession. We should be good.

helix-core/src/main/java/org/apache/helix/HelixManager.java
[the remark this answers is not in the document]
pkuwm: @dasahcc An update. I've synced up with @jiajunwang about this. We decided to change it a bit to `Optional<String> getSessionIdIfLeader()` to avoid the new wrapper `InstanceLeaderSession`.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1738,29 +1705,36 @@ public Stat writeDataGetStat(final String path, Object datat, final int expected
     return writeDataReturnStat(path, datat, expectedVersion);
   }
 
-  public void asyncCreate(final String path, Object datat, final CreateMode mode,
```
jiajunwang, line 1741: Why changing the name? I think datat stands for data type. In this scenario, it means this is an object with a certain data type. "data" is very generic so it might be less meaningful.
I don't have a strong preference on the names. But if you are going to change it, please change all the datat in this java file at least : ) Or even better, the whole project.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2114,13 +2088,43 @@ private String getHexSessionId() {
...
+            "Failed to get expected zookeeper instance! There is a session id mismatch. Expected: "
+                + expectedSessionId + ". Actual: " + actualSessionId);
+      }
+
+      /*
+       * Cache the zookeeper reference and make sure later zooKeeper.create() is being run
+       * under this zookeeper connection. This is to avoid locking zooKeeper.create() which
+       * may cause potential performance issue.
+       */
+      return ((ZkConnection) getConnection()).getZookeeper();
```
jiajunwang, line 2124: return zk?

helix-core/src/main/java/org/apache/helix/HelixManager.java
```diff
@@ -419,6 +420,17 @@ void addExternalViewChangeListener(org.apache.helix.ExternalViewChangeListener l
...
   Long getSessionStartTime();
 
+  /**
+   * Checks whether the cluster manager is leader and returns the session ID associated to the
+   * connection of cluster data store, if and only if it is leader.
+   *
+   * @return {@code Optional<String>} session ID is present inside the {@code Optional} object
+   * if the cluster manager is leader. Otherwise, returns an empty {@code Optional} object.
+   */
+  default Optional<String> getSessionIdIfLead() {
```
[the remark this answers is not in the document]
jiajunwang: Another concern is getSessionId() would become less useful with this call. And would it be confusing for the caller which one to call?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -372,6 +372,12 @@ public void asyncCreate(String path, Object datat, CreateMode mode,
     _rawZkClient.asyncCreate(path, datat, mode, cb);
   }
 
+  @Override
+  public void asyncCreate(String path, Object datat, CreateMode mode,
```
jiajunwang, line 376: It's better to support it for the dedicated zkclient only. For the other clients, especially customer-facing interface, I would prefer not to open this API for now.

helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
pkuwm: I understand your point. By having a new interface, we could add javadoc to explain not use it if they don't care about session (maybe helix internal use only). They still keep using BaseDataAccessor without being confused by the extra session aware api. This is the benefit.

And another benefit is we may add more session aware apis to the new interface in the future if we need.

The downside of having a new interface is 
1. we have to maintain the extra interfaces: SessionAwareBaseDataAccessor, SessionAwareHelixDataAccessor, (maybe session aware zkclient?).
2. Need more code changes to implement the new interfaces (SessionAwareBaseDataAccessor) in this PR.

Let's discuss further.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: In that case, please change them in a batch in a separate PR. This change won't help with this PR.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: It looks good to me.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
jiajunwang: By doing so, I believe additional interface would help. Or you need to do the ugly default methods everywhere.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: Since there is not much value to change them, I would just leave them as they are: `datat`

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: Thanks for confirming. Simplified it.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: How do we reason this is correct? Note, previously, we have `acquireEventLock` when checking sessionId is equal to expected sessionId.
```
if (isSessionAwareOperation(expectedSessionId, mode)) {
          acquireEventLock();
          try {
            final String actualSessionId = Long.toHexString(zooKeeper.getSessionId());
            if (!actualSessionId.equals(expectedSessionId)) {
              throw new ZkSessionMismatchedException(
                  "Failed to create ephemeral node! There is a session id mismatch. Expected: "
                      + expectedSessionId + ". Actual: " + actualSessionId);
            }

            /*
             * Cache the zookeeper reference and make sure later zooKeeper.create() is being run
             * under this zookeeper connection. This is to avoid locking zooKeeper.create() which
             * may cause potential performance issue.
             */
            zooKeeper = ((ZkConnection) getConnection()).getZookeeper();
```

kaisun2000: So here, we are saying the native Zookeeper object `zookeeper` can't be changed even session changed. Still one Zookeeeper object one session invariant, right? Save us one lock.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1740,27 +1707,34 @@ public Stat writeDataGetStat(final String path, Object datat, final int expected
 
   public void asyncCreate(final String path, Object datat, final CreateMode mode,
       final ZkAsyncCallbacks.CreateCallbackHandler cb) {
+    asyncCreate(path, datat, mode, cb, null);
+  }
+
+  public void asyncCreate(final String path, Object data, final CreateMode mode,
```
kaisun2000, line 1713: I assume we should invoke this one in ZkBaseDataAccessor#createChildren. But I did not see it. 

Am I missing something here?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -149,8 +149,7 @@ public void handleSessionEstablishmentError(Throwable error) {
    * Tests session expiry for the helix's IZkStateListener.
    */
   @Test
-  void testSessionExpiry()
-      throws Exception {
+  void testSessionExpiry() throws Exception {
```
kaisun2000, line 152: Nit, I saw this non-essential beautification all the time. Let us not add them to divert the attention.

pkuwm: These are formatted by helix-style. I guess I formatted the whole file. Usually I don't do that. Well just a few lines I think.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: @kaisun2000 Yes, correct! We don't have to lock it actually.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: @kaisun2000 You've got an eye on it. We have a discussion here regarding the new API vs new interface: https://github.com/apache/helix/pull/1066#discussion_r437912277
You may put some opinions there :)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
pkuwm: There is trade-off: with additional interfaces, we need to maintain the additional interfaces, and we need to change the zkclient type with SessionAwareZkClient in HelixManager as well, because these BaseDataAccessors are using the zkclient in helix manager underneath. Let's discuss it further :)

kaisun2000:
The current pull request has one important assumption if I understand correctly. 

Controller assignment to clusters in super cluster change changes  when controller has a session expiration and establishing a new session. 

The thing is that is there any other event can cause controller assignment to cluster change without controller going through session expiration. For example, new cluster are added? Or any others?
@jiajunwang, @pkuwm?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
pkuwm: @jiajunwang PR updated.
We decided to keep current existing interfaces as they are. Instead, we use the srcSessionId in message as the expected session id to create message znode.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2234,13 +2205,59 @@ private String getHexSessionId() {
...
+
+    ZNRecord record;
+    try {
+      record = (ZNRecord) data;
+    } catch (ClassCastException e) {
+      LOG.debug("Failed to parse expected session id!", e);
+      return null;
+    }
+
+    // Check it is a message and get src session id as expected session id for message.
```
jiajunwang, line 2256: This logic is Helix logic specific, not a good idea to put in ZkClient.

To make it generic, one way is adding a new interface of ZnRecord (SessionAwareZnRecordUpdateRequest, etc.) which provides a method such as getExpectedSessionId() to return this information.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -695,12 +695,19 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
     event.addAttribute(AttributeName.STATEFUL_REBALANCER.name(),
         _rebalancerRef.getRebalancer(manager));
 
-    if (!manager.isLeader()) {
-      logger.error("Cluster manager: " + manager.getInstanceName() + " is not leader for " + manager
-          .getClusterName() + ". Pipeline will not be invoked");
+    Optional<String> leaderSession = manager.getSessionIdIfLead();
```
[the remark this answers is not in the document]
pkuwm: @jiajunwang 
> can we actually add the session Id to the ClusterEvent when it is constructed, and then use that session Id directly here?

I don't think it is accurate: when the event is constructed and we add this session id **S0** to event, if the manager session id changes to **S1** when handling the event, the event would not be processed. But actually I think it would be more accurate to use the session id when starting to process the pipeline.

Or even session id changes to S1 and the event is not handled, there will be future events to keep the correctness. In this case, I think it is fine to add the session id to when event is created. But I think it may affect a bit performance: originally event0 will process the pipeline, but it is discarded and next event1 will be processed later.

So I think it is better to add the session right before running the pipeline.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2234,13 +2205,59 @@ private String getHexSessionId() {
...
+  }
+
+  private String parseExpectedSessionId(Object data) {
+    if (data == null) {
+      return null;
+    }
+
+    ZNRecord record;
+    try {
+      record = (ZNRecord) data;
```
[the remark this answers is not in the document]
pkuwm: I've thought about it. `data instanceof Message` adds one more check.
```
if (!data instanceof Message) {
  return null;
}

ZNRecord record = (ZNRecord) data;
...
```

Since most of the time, this API is for message and we need to cast the type, I prefer this so no need to always do the `instanceof` check.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1827,18 +1796,20 @@ public void asyncCreate(final String path, Object datat, final CreateMode mode,
           new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
       return;
     }
-    doAsyncCreate(path, data, mode, startT, cb);
+    doAsyncCreate(path, data, mode, startT, cb, parseExpectedSessionId(datat));
```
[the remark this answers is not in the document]
pkuwm: For this fix, we only need it for creating messages. For updating messages, it should be done by participants. So we don't need that (correct me if I am wrong). If we want to do it for all the write/create operations, we have to protect all the async/sync operations, which I don't think we want to do now. 
If in the future we also need to protect write operations, we could definitely quickly add the same logic as in create. What do you think?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: Yep, I agree we should separate helix and zk logic cleanly. Let me change and see if this will make it cleaner.

pkuwm: @jiajunwang I've cleaned up the code in zkClient. I think it is cleaner. My thought is we don't have to create a new class extends ZNRecord which only introduces an expectedSessionId. It is fine to just add the field/methods to existing ZNRecord. And we move the expected session population logic to ZkHelixDataAccessor's createChildren. It seems to me a clean way. What do you think?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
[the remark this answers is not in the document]
jiajunwang: The reason we want this change is to prevent 2 pipelines running. So if the S0 session event is handled in S1, do you think it would be risky that we have more confusion and leads to a bad result?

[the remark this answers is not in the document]
pkuwm: Synced with @jiajunwang offline. We agreed to add event session when event is created. Updated the PR.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -445,7 +448,8 @@ private Message createStateTransitionCancellationMessage(HelixManager manager, R
       message.setFromState(fromState);
       message.setToState(toState);
       message.setTgtSessionId(sessionId);
-      message.setSrcSessionId(manager.getSessionId());
+      message.setSrcSessionId(managerSessionId);
+      message.setExpectedSessionId(managerSessionId);
```
lei-xia, line 452: What is the difference between SrcSessionId and ExpectedSessionId?

helix-core/src/main/java/org/apache/helix/HelixManager.java
lei-xia, line 430: getSessionIdIfLeader?  Also, why do we need this if we already have manager.getSessionId?

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
pkuwm: The difference I suppose is: say, a message is forwarded/sent from node1 to node2. So srcSessionId = sessionNode1. This message doesn't have to be written by node1's zkclient, but another zkclient `session-0`. So expectedSessionId = `session-0`. I'd like to differentiate  srcSessionId and write expectedSessionId.
Let me know I miss anything.

helix-core/src/main/java/org/apache/helix/HelixProperty.java
```diff
@@ -189,7 +200,7 @@ public final String getId() {
    * Get the backing ZNRecord
    * @return ZNRecord object associated with this property
    */
-  public final ZNRecord getRecord() {
+  public ZNRecord getRecord() {
```
lei-xia, line 203: why change this?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -312,6 +312,17 @@ private void forceRebalance(HelixManager manager, ClusterEventType eventType) {
...
+    // If session is not present, this cluster manager is not leader for the cluster.
+    if (!leaderSession.isPresent()) {
+      logger.warn("Cluster manager {} is not leader for {}. Event {} is discarded.",
+          manager.getInstanceName(), manager.getClusterName(), event);
+      return;
+    }
+
+    // Pipeline should be run and Zk writes should be completed by the event session.
+    event.addAttribute(AttributeName.EVENT_SESSION.name(), leaderSession.get());
```
lei-xia, line 325: In HandleEvent you already check if the sessionId matches the expected sessionId, why do we check again here?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -757,9 +762,22 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
     }
     event.addAttribute(AttributeName.ControllerDataProvider.name(), dataProvider);
 
-    logger.info(String.format("START: Invoking %s controller pipeline for cluster %s event: %s  %s",
-        manager.getClusterName(), dataProvider.getPipelineName(), event.getEventType(),
-        event.getEventId()));
+    String eventSessionId = event.getAttribute(AttributeName.EVENT_SESSION.name());
+    if (eventSessionId != null) {
+      String managerSessionId = manager.getSessionId();
+      if (!eventSessionId.equals(managerSessionId)) {
```
lei-xia, line 768: if(eventSessionId != null && manager.getSessionId.equals(eventSessionId)) {
  ...
}

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -757,9 +762,22 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
...
+        logger.warn(
+            "Controller pipeline is not invoked because cluster manager {} lost leadership for "
```
lei-xia, line 770: This warn message is misleading, it does not necessarily mean the current controller is not the leader of the cluster. This just means we discarded event that was generated from an old session.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1360,8 +1390,20 @@ private boolean updateControllerState(NotificationContext changeContext, PauseSi
         String uid = UUID.randomUUID().toString().substring(0, 8);
         ClusterEvent event = new ClusterEvent(_clusterName, ClusterEventType.Resume,
             String.format("%s_%s", uid, Pipeline.Type.DEFAULT.name()));
+
+        HelixManager manager = changeContext.getManager();
```
lei-xia, line 1394: Seems we have multiple places to check this event session ID matches. In pushEventToQueue, in HandleEvent(), in ForceRebalance().  Do we really need to validate this everywhere?  Can we combine them together in one place?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1360,8 +1390,20 @@ private boolean updateControllerState(NotificationContext changeContext, PauseSi
...
+        Optional<String> leaderSession = manager.getSessionIdIfLead();
```
lei-xia, line 1395: Also using of manager.getSessionIdIfLead here could be confusing. We used getSessionId in other places, what is the difference?  Where we should use getSessionIdIfLead and where we should use getSessionId?

helix-core/src/main/java/org/apache/helix/controller/stages/MessageDispatchStage.java
```diff
@@ -78,7 +78,17 @@ protected void processEvent(ClusterEvent event, MessageOutput messageOutput) thr
...
 
+    String expectedSession = event.getAttribute(AttributeName.EVENT_SESSION.name());
+    // An early check for expected leader session. If the sessions don't match, it means the
+    // controller lost leadership, then messages should not be sent and pipeline should stop.
+    // This avoids potential double masters for a single partition.
+    if (expectedSession != null && !expectedSession.equals(manager.getSessionId())) {
+      throw new StageException(
+          "Controller: " + manager.getInstanceName() + " lost leadership! Expected session: "
+              + expectedSession + ", actual: " + manager.getSessionId());
+    }
```
lei-xia, line 89: should we do this for all Stage? or just messageDispatchStage?  If for all stages, maybe we should put this check before where each stage was called.

helix-core/src/main/java/org/apache/helix/model/Message.java
```diff
@@ -953,4 +970,35 @@ public boolean isValid() {
     }
     return true;
   }
+
+  // A class represents session aware ZNRecord for message. The message should be written to zk
+  // by the expected session.
+  // TODO: remove this class once public session-aware ZNRecord is available
+  private static class SessionAwareZNRecord extends ZNRecord implements SessionAwareZkWriteData {
```
lei-xia, line 977: I still do not think adding this new subtype ZNRecord is an elegant solution here, specially you put it as a private class here only for message could make it look a bit hacky. Let us discuss what other options here.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/SessionAwareZkWriteData.java
```diff
@@ -0,0 +1,34 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * An interface representing data being written to ZK is session aware:
+ * data is supposed to be written by expected ZK session. If ZkClient's actual session
+ * doesn't match expected session, data is not written to ZK.
+ */
+public interface SessionAwareZkWriteData {
```
lei-xia, line 27: Again this name is confusing. What SessionAware exactly mean here?  And "ZkWriteData", does that mean I should only use this type when I write a data to ZK, should we return this type in a ZK read operation?   A better name could be SessionAwareZkDataType or SessionAwareZkDataRecord, something like that.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: The cluster manager's session changed, lost and then acquired leadership again. Yep, you are right. I'll make the log clearer.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: `String managerSessionId = manager.getSessionId();` is having `managerSessionId` to be used in if check and log. If we do it in the way you suggested, either we don't log this manager session, or we need `manager.getSessionId()` again in log. `getSessionId` also wait for connected. And if during these two getSessionId, the session id changes, we have different session ids.
I prefer to log manager session id so this is why I put a variable here.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: The purpose of this check here is to only add events to the queue if and only if controller is leader. We attach the session id as event session id. If controller is not leader, we discard the event and don't push it to queue.

In HandleEvent(), we check it as a filter if we should start the pipeline for the event. Because during the event in the queue, controller's session may change. If it changes, there is no need to start the pipeline for the stale event.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: As mentioned above, they are different: ForceRebalance() and pushEventToQueue() are creating new events. We want to check the leadership and add the session to the event.

For later checks such as in HandleEvent(), they are just using the event session id to check current controller session. If controller session changes, we stop the event/pipeline.

So, since we have 3 places that create new events, I don't think we could combine them together in one place(I've thought about it and tried).

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: `manager.getSessionIdIfLead` checks whether the cluster manager is leader and returns the session ID associated to the connection of cluster data store, if and only if it is leader. It combines `isLeader()` and `getSessionId()` in one method to ensure the session id returning is what we check leadership in isLeader().

Without it, we may do this:
```
if (isLeader()) {
  sessionId = manager.getSessionId();
}
```
The problem is race condition. sessionId may change after isLeader().

So we need `manager.getSessionIdIfLead` to check leadership and get sessionId at the same time.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageDispatchStage.java
pkuwm: We've discussed this question. We decided not to do the check for all stages (too many stages). Performance is not concern. Correctness is the most critical. We only do it for to ensure correctness here.

helix-core/src/main/java/org/apache/helix/model/Message.java
pkuwm: This is the most challenging part remaining in the PR that we consider. Multiple options are considered and tried. This one is by far a relatively cleaner one.

Considering:
1. not adding new session aware methods/interfaces ---> attach expected session together with data object
2. making it generic in zkclient ---> new interface `SessionAwareZkDataRecord`
3. Less hacky way than checking ZNRecord in zkclient ----> SessionAwareZNRecord extends ZNRecord implements SessionAwareZkWriteData. We don't have a public implementation `SessionAwareZNRecord`. So this is needed. I've added TODO if in future we implement public `SessionAwareZNRecord`, we should remove this private class.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/SessionAwareZkWriteData.java
pkuwm: Yep I actually also thought about `SessionAwareZkDataRecord`. But considering the data is only session-aware for write but not read, I made it `SessionAwareZkDataRecord` to indicate reading data may not use this one as session aware. We could think of a better name to make it clear.
@jiajunwang What's your choice?

helix-core/src/main/java/org/apache/helix/model/Message.java
lei-xia: If we believe this is the right solution, why not make it public?

helix-core/src/main/java/org/apache/helix/controller/stages/MessageDispatchStage.java
lei-xia: Why checking in every stage is too much? Do you mean code wise is too much, actually you just need the same code, in different place, right?

helix-core/src/main/java/org/apache/helix/HelixProperty.java
```diff
@@ -148,7 +154,12 @@ public String toString() {
     }
   }
 
-  private Stat _stat;
+  protected Stat _stat;
+
+  public HelixProperty() {
+    _record = DEFAULT_ZNRECORD;
+    _stat = DEFAULT_STAT;
+  }
```
lei-xia, line 162: Again, why do we need a default constructor here?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
lei-xia: that is fine, what I meant is we do not two levels of embeded if.. statements in this case.  

if(eventSessionId != null && managerSessionId.equals(eventSessionId))...

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
lei-xia: Then when we created the event, just add the current session to the event attributes, is that good enough? We will check whether current controller is a leader and whether session matches anyway in handleEvent?

lei-xia: What I meant, just need one line in the event create places:

    event.addAttribute(AttributeName.EVENT_SESSION.name(), manager.getSessionId());

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -971,19 +972,24 @@ public void addPreConnectCallback(PreConnectCallback callback) {
 
   @Override
   public boolean isLeader() {
+    return getSessionIdIfLead().isPresent();
+  }
+
+  @Override
+  public Optional<String> getSessionIdIfLead() {
```
lei-xia, line 979: Still, if you think this is the right name, let us rename isLeader() to isLead(), just make them consistent :).

helix-core/src/main/java/org/apache/helix/HelixProperty.java
pkuwm: If we don't have this default constructor, in Message's constructor, we need to call `super(ZNRecord)`. HelixProperty defines `ZNRecord` but not `SessionAwareZNRecord`. Then if we want Message's SessionAwareZNRecord, we may override `getRecord()` to convert `ZNRecord` to `SessionAwareZNRecord`: 
```
Override
public ZNRecord getRecord() {
  return new SessionAwareZNRecord(_record);
}
```
The problem is this creates a new record from the original one `_record`. We could not access the original one: `message.getRecord().setSimpleField()`.

So we need to create SessionAwareZNRecord in Message's constructor and override parent's ZNRecord. Without the default constructor, we won't be able to do `_record = new SessionAwareZNRecord()` in Message.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: I got your point. To achieve one level if statement, we could move managerSessionId outside. Then the scope of managerSessionId is bigger.
```
String eventSessionId = event.getAttribute(AttributeName.EVENT_SESSION.name());
String managerSessionId = manager.getSessionId();
if (eventSessionId != null && !eventSessionId.equals(managerSessionId)) {
...
}
```

But after this if statement we don't need managerSessionId. This what I try to do, following `Effective Java Item - 45 : Minimize the scope of local variables`: https://gist.github.com/kpgalligan/4cf93abd88044df71da5

helix-core/src/main/java/org/apache/helix/controller/stages/MessageDispatchStage.java
pkuwm: What I was thinking about is, we also check this stage in `ReadClusterDataStage()` because this place is the highly possible one that session expires. In our issue, the session expiry also happened in `ReadClusterDataStage`.

If we don't check it at every stage, it just costs extra resource(CPU/memory/network IO) which we could tolerate. I think checking at every stage may be excessive because most of stages are pretty fast(`ReadClusterDataStage` takes longer time) and very tiny possibility that session changes. As long as we protect sending message, correctness is ensured. We wanted to make minimum code change.

helix-core/src/main/java/org/apache/helix/model/Message.java
pkuwm: We still want to make minimum changes in this PR. Putting it private gives us more flexibility to change in future: may add more stuff to this class. If in future we find that we need `SessionAwareZNRecord` in other places, we should create such thing as public. We need to discuss more and have a more comprehensive design for public `SessionAwareZNRecord`. What do you think?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
pkuwm: Why we name it `getSessionIfLead()` is because following if should be an adj., eg. `putIfAbsent()`, `computeIfPresent()` etc.. Lead is an adj. Leader is not. So I think from usage, I prefer `getSessionIfLead()`. `isLeader()` is also fine I think.

helix-core/src/main/java/org/apache/helix/HelixManager.java
pkuwm: manager.getSessionIdIfLead checks whether the cluster manager is leader and returns the session ID associated to the connection of cluster data store, if and only if it is leader. It combines isLeader() and getSessionId() in one method to ensure the session id returning is what we check leadership in isLeader().

Without it, we may do this:

if (isLeader()) {
  sessionId = manager.getSessionId();
}
The problem is race condition. sessionId may change after isLeader().

So we need manager.getSessionIdIfLead to check leadership and get sessionId at the same time.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: > Then when we created the event, just add the current session to the event attributes, is that good enough? We will check whether current controller is a leader and whether session matches anyway in handleEvent?

It is OK to check session when handleEvent, which could reduce the checks when pushing event. Downside is when we create the events, manager session has already expired. Then we still push stale events to the event queue.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: To make code clean, I decided to put it the check in one place in handleEvent(). It is fine to push stale events into queue.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: Changed.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: Since there are 3 places creating new events, I've moved the check to handleEvent().

helix-core/src/main/java/org/apache/helix/HelixManager.java
kaisun2000: Basically, the idea is that check leader and get leader sessionID at one operation, later if sessionID changes, we use sessionAwareWriter interface to fail it.

pkuwm: @kaisun2000 You are right. This is the core idea.

helix-core/src/main/java/org/apache/helix/model/Message.java
pkuwm: @lei-xia I've considered and tried to break the PR down to 2 smaller ones: fixing the issue without new interface and adding new interface. But the first one PR would look quite hacky and I don't feel comfortable merging the commit to code base. So I decided to keep this PR and make `SessionAwareZNRecord` public. Even in the future we need to add more APIs to the interface, we could also add API implementation to `SessionAwareZNRecord`. In this way, I feel it is cleaner to check in.

helix-core/src/main/java/org/apache/helix/HelixManager.java
pkuwm: One more thing for `why do we need this if we already have manager.getSessionId?`:
`getSessionId` just directly returns the existing sessionId in manager. But `getSessionIdIfLeader()` does more thing by reading zk to check leadership. This is the main difference.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -757,9 +753,20 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
     }
     event.addAttribute(AttributeName.ControllerDataProvider.name(), dataProvider);
 
-    logger.info(String.format("START: Invoking %s controller pipeline for cluster %s event: %s  %s",
-        manager.getClusterName(), dataProvider.getPipelineName(), event.getEventType(),
-        event.getEventId()));
+    // If manager session changes, no need to run pipeline for the stale event.
+    Optional<String> eventSessionId = event.getAttribute(AttributeName.EVENT_SESSION.name());
+    String managerSessionId = manager.getSessionId();
```
kaisun2000, line 758: If I understand correctly, the sessionId here can still be stale. eventually, we need sessionaware write of zkclient used by ZkHelixManager to ensure correctness, right?

kaisun2000: I think there are  other event not trigger from ZkClient, such as periodical timer triggered, or onDemandtimer triggered. Do we need special treatment here? or Special treatment when event to controller is constructed?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1360,6 +1369,8 @@ private boolean updateControllerState(NotificationContext changeContext, PauseSi
         String uid = UUID.randomUUID().toString().substring(0, 8);
         ClusterEvent event = new ClusterEvent(_clusterName, ClusterEventType.Resume,
             String.format("%s_%s", uid, Pipeline.Type.DEFAULT.name()));
+        event.addAttribute(AttributeName.EVENT_SESSION.name(),
+            changeContext.getManager().getSessionIdIfLead());
```
kaisun2000, line 1373: This part is actually debatable? updateControllerState is used to pause or restart paused pipeline. Do it care about session? Can we lose this signal due to session check? How do we reason this one?

I am not very familiar with this logic. @jiajunwang , @lei-xia, @dasahcc

kaisun2000:
A high level question. The following approach basically encode the session info to the data of ZkClient? Last time in the design discussion I guess people raised the concern that this mean ZkClient need to understand the data format. Normally lower level do not understand higher level data. 

Though you can also think of session id in the data as meta data, just like TCP has out of line control data. 

Either way is fine to me. Just want to make sure @pkuwm and @lei-xia agree with this approach. 

```
  private String parseExpectedSessionId(Object data) {
    if (!(data instanceof SessionAwareZkWriteData)) {
      return null;
    }
    return ((SessionAwareZkWriteData) data).getExpectedSessionId();
```

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: @kaisun2000 Here if session id from manager is different from event session -- the session is changed. We check it and stop processing the event or sending messages.
The check in Zkclient is the last step. Most of the time this check here will stop the stale session sending messages.
Regarding other events, each time an event is created, we add the event_session to it. The event session is from cluster manager.

`if (!eventSessionId.isPresent() || !eventSessionId.get().equals(managerSessionId)) {` only it's leader and session doesn't change, it continues, otherwise, pipeline is not invoked.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/SessionAwareZkWriteData.java
pkuwm: Will decide this new interface in another PR.

pkuwm:
> A high level question. The following approach basically encode the session info to the data of ZkClient? Last time in the design discussion I guess people raised the concern that this mean ZkClient need to understand the data format. Normally lower level do not understand higher level data.
> 
> Though you can also think of session id in the data as meta data, just like TCP has out of line control data.
> 
> Either way is fine to me. Just want to make sure @pkuwm and @lei-xia agree with this approach.
> 
> ```
>   private String parseExpectedSessionId(Object data) {
>     if (!(data instanceof SessionAwareZkWriteData)) {
>       return null;
>     }
>     return ((SessionAwareZkWriteData) data).getExpectedSessionId();
> ```

We add a new interface SessionAwareZkWriteData to make it generic: representing data is session aware. In future, we may implement this interface to have a public one like SessionAwareZNRecord . With this new interface, ZkClient doesn’t need to change: treating SessionAwareZkWriteData as generic one. 

How about this? @lei-xia @jiajunwang @kaisun2000 
Considering the new interface would tie session to zk, it is not generic enough. Eg. if we change the storage to mysql, it doesn’t have session.
How about this, splitting the PR into 2: 1st one (this one) doesn’t have new interface, 2nd one will have the new interface. If we don’t agree on adding new interface, we just discard the 2nd PR, being tracked in https://github.com/apache/helix/issues/1219.

helix-core/src/main/java/org/apache/helix/HelixProperty.java
jiajunwang: How about do this?
  public HelixProperty(SessionAwareZNRecord record, String id) {
    _record = record;
    _stat = new Stat(_record.getVersion(), _record.getCreationTime(), _record.getModifiedTime(),
        _record.getEphemeralOwner());
  }

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -695,12 +697,6 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
     event.addAttribute(AttributeName.STATEFUL_REBALANCER.name(),
         _rebalancerRef.getRebalancer(manager));
 
-    if (!manager.isLeader()) {
```
jiajunwang, line 698: Delaying this check might cause confusing and seems not necessary to this PR. Let's don't change it in this PR even if desired.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
```diff
@@ -40,5 +40,8 @@
   PipelineType,
   LastRebalanceFinishTimeStamp,
   ControllerDataProvider,
-  STATEFUL_REBALANCER
+  STATEFUL_REBALANCER,
+
+  /** Event should be processed by the expected session. */
```
jiajunwang, line 45: This claim is not related to this field but with the internal logic. Let's just say this is the session that in which we get the event.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/datamodel/SessionAwareZNRecord.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public SessionAwareZNRecord(String id) {
+    super(id);
+  }
+
+  public SessionAwareZNRecord(ZNRecord record) {
+    super(record);
+  }
+
+  public SessionAwareZNRecord(ZNRecord record, String id) {
```
jiajunwang, line 45: Is this one needed?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/datamodel/SessionAwareZNRecord.java
```diff
@@ -0,0 +1,68 @@
...
+ * writing to zk will fail. It is supposed to be used within Helix only.
+ * <p>
+ * If this ZNRecord is not supposed to be written only by the expected zk session,
+ * {@link ZNRecord} is recommended to use.
+ */
+public class SessionAwareZNRecord extends ZNRecord {
+  @JsonIgnore
+  private String expectedSessionId;
+
+  public SessionAwareZNRecord(String id) {
```
jiajunwang, line 37: Is this one needed?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
pkuwm: Ah good catch. I intended to replace this line with the new one: check leader and get session at the same time. But somehow it goes to later right before invoking pipeline. Resolved.

helix-core/src/main/java/org/apache/helix/HelixManager.java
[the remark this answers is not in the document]
pkuwm: @jiajunwang Thanks for sharing your thought. I think the concept "fake constant session" is like an empty `Optional` object? With `Optional`, it looks more straightforward and clear than a fake constant session.
With a fake constant, we still need an `if` to check leadership for those storage systems that have leadership. I would say `null` is kind of this fake constant: if no session concept, then return null. And I think `Optional` is better than `null` here.

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
```diff
@@ -788,14 +788,18 @@ protected void setupInstances(String clusterName, int[] instances) {
     }
   }
 
-  protected void runPipeline(ClusterEvent event, Pipeline pipeline) {
+  protected void runPipeline(ClusterEvent event, Pipeline pipeline, boolean shouldThrowException)
```
jiajunwang, line 791: How about giving us a break by keeping the existing method with shouldThrowException default to be false?
So you have much less changed files.

helix-core/src/main/java/org/apache/helix/model/Message.java
```diff
@@ -937,6 +946,13 @@ private boolean isNullOrEmpty(String data) {
     return data == null || data.length() == 0 || data.trim().length() == 0;
   }
 
+  private void initStat() {
```
jiajunwang, line 949: This looks ugly.

Let's do this,

  public Message(ZNRecord record, String id) {
    _record = new SessionAwareZNRecord(record, id);
    _stat = new Stat(_record.getVersion(), _record.getCreationTime(), _record.getModifiedTime(),
        _record.getEphemeralOwner());
  }

Then the other constructors refer to this one.

helix-core/src/main/java/org/apache/helix/model/Message.java
```diff
@@ -206,10 +194,27 @@ public Message(ZNRecord record, String id) {
    * @param id unique message identifier
    */
   public Message(Message message, String id) {
```
jiajunwang, line 196: This seems not being used, just deprecate it?

helix-core/src/main/java/org/apache/helix/controller/stages/MessageDispatchStage.java
```diff
@@ -78,7 +79,17 @@ protected void processEvent(ClusterEvent event, MessageOutput messageOutput) thr
         batchMessage(dataAccessor.keyBuilder(), messagesToSend, resourceMap, liveInstanceMap,
             manager.getProperties());
 
+    // An early check for expected leader session. If the sessions don't match, it means the
+    // controller's session changes, then messages should not be sent and pipeline should stop.
+    Optional<String> expectedSession = event.getAttribute(AttributeName.EVENT_SESSION.name());
+    if (!expectedSession.isPresent() || !expectedSession.get().equals(manager.getSessionId())) {
+      throw new StageException(String.format(
```
jiajunwang, line 86: Can we skip the dispatch but continue the pipeline for the other stages?
I'm not sure, this is a question.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -703,9 +705,15 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
...
-      logger.error("Cluster manager: " + manager.getInstanceName() + " is not leader for " + manager
-          .getClusterName() + ". Pipeline will not be invoked");
+    // If manager session changes, no need to run pipeline for the stale event.
```
jiajunwang, line 708: This is right, but the original logic was, if the manager is not leader, do not run the pipeline. Where is that logic going?

helix-core/src/main/java/org/apache/helix/HelixProperty.java
jiajunwang: I know there are multiple options. But how about the one that I mentioned above?

helix-core/src/main/java/org/apache/helix/HelixManager.java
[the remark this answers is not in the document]
jiajunwang: Optional is not looking good here.
It might be overkill, but how about we define a "private" SessionAwareHelixManager interface? Which extends HelixManager but add getSessionIdIfLead() method.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1967,18 +1937,19 @@ public void asyncSetData(final String path, Object datat, final int version,
           new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
       return;
     }
-    doAsyncSetData(path, data, version, startT, cb);
+    doAsyncSetData(path, data, version, startT, cb, parseExpectedSessionId(datat));
   }
 
   private void doAsyncSetData(final String path, byte[] data, final int version, final long startT,
```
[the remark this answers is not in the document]
pkuwm: We may do it later if we need the expected session. I'd like to make this diff minimal.

helix-core/src/main/java/org/apache/helix/model/Message.java
pkuwm: Actually the existing code is already messing :(

helix-core/src/main/java/org/apache/helix/HelixProperty.java
pkuwm: I thought about this one before. I think it is fine. The existing constructors are already kind of messing.. We can not still converge the other constructors to this one.

helix-core/src/main/java/org/apache/helix/HelixManager.java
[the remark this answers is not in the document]
pkuwm: I thought about the private interface. My thought is that this private interface will increase more work to identify which places will need this SessionAwareHelixManager and construct it. And it doesn't converge. Eg. if there is only one place that uses the sessionAwareHelixManager and only one place checking the session, it's fine.
But if there is mixed usage, we may need to use `instaceof` to check, which doesn't look nice.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
kaisun2000: @pkuwm , here is the thing. If say when pause signal znode is added, the under old session S0, the event is put into zkclient queue. Then a new session establish for S1, would the controller under S1 for this logic, would not pause of the old event. 

The question is that can you double check if the new session, controller would get another event of pausing due to controller cache refresh? This is to make sure pause will not be lost immediately followed by a controller change.

kaisun2000 (approved):
Overall it looks good to me. 

Raise one question about Paul signal added immediately with a lead controller switch caae? Can we double check. 

If this is indeed a concern. Let us file another ticket to track it.

Also resolve JJ's comment, it seems minor at this point.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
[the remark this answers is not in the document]
pkuwm: We'll track the work in another issue: https://github.com/apache/helix/issues/1434

helix-core/src/main/java/org/apache/helix/HelixManager.java
```diff
@@ -419,6 +419,17 @@ void addExternalViewChangeListener(org.apache.helix.ExternalViewChangeListener l
...
   Long getSessionStartTime();
 
+  /**
+   * Checks whether the cluster manager is leader and returns the session ID associated to the
+   * connection of cluster data store, if and only if it is leader.
+   *
+   * @return {@code String} session ID is a valid string if the cluster manager is leader.
+   * Otherwise, returns null.
+   */
+  default String getSessionIdIfLead() {
```
jiajunwang, line 429: If you think my previous suggestion about the SessionAwareHelixManager interface is too much, then please still keep the Optional<String> here.

jiajunwang (approved):
LGTM!

helix-core/src/main/java/org/apache/helix/HelixProperty.java
```diff
@@ -177,6 +178,15 @@ public HelixProperty(ZNRecord record, String id) {
         _record.getEphemeralOwner());
   }
 
+  /*
+   * Only used by Message which needs to be session-aware.
+   */
+  protected HelixProperty(SessionAwareZNRecord record, String id) {
+    _record = record;
```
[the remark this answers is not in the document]
pkuwm: Synced offline. Updated.

kaisun2000:
@pkuwm, @jiajunwang,

Sync-ed with @pkuwm offline. As I raise, there is another race condition in this code. This is not introduced by this pull request, it exists all the time. 

Basically, it is like this:
>Say, pause signal is added, the old lead controller see this one and the callback is in zkclient queue. the new lead controller to be also see it, but the generic controller ignored it as it is not leader yet.
then new controller becomes leader due to old controller session expiration.
will new controller in this case not pausing?

Let us sync about shall we fix it here, or in another pull, or what should we do? I will create an issue to track it.

pkuwm:
> @pkuwm, @jiajunwang,
> 
> Sync-ed with @pkuwm offline. As I raised, there seems to be another race condition in this code. This is not introduced by this pull request, it exists all the time.
> 
> Basically, it is like this:
> 
> > Say, pause signal is added, the old lead controller see this one and the callback is in zkclient queue. the new lead controller to be also see it, but the generic controller ignored it as it is not leader yet.
> > then new controller becomes leader due to old controller session expiration.
> > will new controller in this case not pausing?
> 
> Let us sync about shall we fix it here, or in another pull, or what should we do? I will create an issue to track it.
> 
> #1442 is tracking this newly raised issue.

@kaisun2000 Thanks for creating the issue for tracking it. If it is an issue, we should get it in another PR. Let's discuss.

pkuwm:
This PR is ready to be merged, approved by @kaisun2000 @jiajunwang 

There is a case that after controller leader switches, pipelines of both old Helix controller leader and new leader are running in parallel.
Different assignment decisions are sent to different participants so there are double masters for a single partition.

The commit addresses this issue by blocking non-leader controller to send messages to ZK.

Issue #1067, opened 2020-06-05
Improve message generation logic to increase rate of relay (P2P) message


zhangmeng916:
In Helix design, relay message (P2P message) is leveraged to improve mastership handoff latency. However, from our customer's using experience, we found that a large percent of messages that should be sent as relay messages are not, which caused high mastership handoff latency. The reason is that in message generation phase, we hold on the generation of SLAVE -> MASTER message to the future master if there is already a pending message for that instance on the same partition. But in many cases, the pending message is already processed, and the current state of that partition is also updated, and only the message itself has not been deleted yet. In these scenarios, we should not wait on the pending message, instead, the new message from S->M should be generated, and then in message selection phase the two messages will be combined as a relay message.

PR #1068, opened 2020-06-05
Remove waiting on message deletion if current state is already updated

zhangmeng916:
### Issues

(#1067 )

### Description

This PR remove the waiting on existing pending message if the message is already processed and current state has already been updated. This would help generate message earlier and increase the rate of P2P message.

### Tests

Two tests are heavily modified:
TestP2PMessagesAvoidDuplicatedMessage
TestP2PStateTransitionMessages

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestRebalancePipeline.testMsgTriggeredRebalance:203 expected:<true> but was:<false>
[ERROR]   TestTaskNumAttempts.testTaskNumAttemptsWithDelay:76 expected:<2> but was:<1>
[ERROR]   TestTaskRebalancer.timeouts » ThreadTimeout Method org.testng.internal.TestNGM...
[INFO] 
[ERROR] Tests run: 1145, Failures: 4, Errors: 0, Skipped: 1
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:27 h
[INFO] Finished at: 2020-06-22T12:14:24-07:00
[ERROR] 

helix-rest
[INFO] Tests run: 162, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 36.34 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 162, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  40.589 s
[INFO] Finished at: 2020-06-09T19:12:40-07:00

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
@@ -220,7 +220,8 @@ public abstract class MessageGenerationPhase extends AbstractBaseStage {
                 generateCancellationMessageForPendingMessage(desiredState, currentState, nextState, pendingMessage,
                     manager, resource, partition, sessionIdMap, instanceName, stateModelDef,
                     cancellationMessage, isCancellationEnabled);
-          } else {
+          }
+          if (pendingMessage == null || currentState.equalsIgnoreCase(pendingMessage.getToState())) {
             // Create new state transition message
             message = createStateTransitionMessage(manager, resource, partition.getPartitionName(),
                 instanceName, currentState, nextState, sessionIdMap.get(instanceName),
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
@@ -221,7 +221,10 @@ public abstract class MessageGenerationPhase extends AbstractBaseStage {
                     manager, resource, partition, sessionIdMap, instanceName, stateModelDef,
                     cancellationMessage, isCancellationEnabled);
           }
-          if (pendingMessage == null || currentState.equalsIgnoreCase(pendingMessage.getToState())) {
+          // We will generate new message if pending message is null or current state equals
+          // pending message's toState (no cancellation message should be generated in this case)
+          if (pendingMessage == null || (message == null &&
+              currentState.equalsIgnoreCase(pendingMessage.getToState()))) {
             // Create new state transition message
             message = createStateTransitionMessage(manager, resource, partition.getPartitionName(),
                 instanceName, currentState, nextState, sessionIdMap.get(instanceName),
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
@@ -225,6 +225,13 @@ public abstract class MessageGenerationPhase extends AbstractBaseStage {
           // pending message's toState (no cancellation message should be generated in this case)
           if (pendingMessage == null || (message == null &&
               currentState.equalsIgnoreCase(pendingMessage.getToState()))) {
+            if (pendingMessage != null && message == null && currentState
+                .equalsIgnoreCase(pendingMessage.getToState())) {
+              logger.info("Ignore the pending message for resource %s partition %s instance "
+                      + "%s since the current state %s equals toState of pending message.",
+                  resource.getResourceName(), partition.getPartitionName(), instanceName,
+                  currentState);
+            }
             // Create new state transition message
             message = createStateTransitionMessage(manager, resource, partition.getPartitionName(),
                 instanceName, currentState, nextState, sessionIdMap.get(instanceName),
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -220,7 +220,8 @@ private void generateMessage(final Resource resource, final BaseControllerDataPr
                 generateCancellationMessageForPendingMessage(desiredState, currentState, nextState, pendingMessage,
                     manager, resource, partition, sessionIdMap, instanceName, stateModelDef,
                     cancellationMessage, isCancellationEnabled);
-          } else {
+          }
+          if (pendingMessage == null || currentState.equalsIgnoreCase(pendingMessage.getToState())) {
```
jiajunwang, line 224: When "pendingMessage != null" but "currentState.equalsIgnoreCase(pendingMessage.getToState())", the cancellation message will be replaced with the new ST message right? I think it works. But would it be better to send both of the cancellation message and the new one?

If we do this, will the participant discards the cancellation message, or proceed with both of them? It would be safer from Helix perspective since the pending message will be removed for sure with cancellation. But it could be complicated from the participant perspective since I think we might need some sync control on the participant side.

zhangmeng916: Actually if looking at the generateCancellationMessageForPendingMessage logic, there is

`else if (currentState.equalsIgnoreCase(pendingState)) {
        LogUtil.logInfo(logger, _eventId,
            "Message hasn't been removed for " + instanceName + " to transit " + resource
                .getResourceName() + "." + partition.getPartitionName() + " to "
                + pendingState + ", desiredState: " + desiredState + ", isRelay: " + pendingMessage.isRelayMessage());
      }`

So actually when currentState equals pendingState, there is no cancellation message generated.

jiajunwang: Then it means the decision of canceling or not is purely in the generateCancellationMessageForPendingMessage(). And we don't have the option to send both of the messages for now.

In this case, it would be cleaner to do the following, I guess.
```
if (pendingMessage != null) {
            message =
                generateCancellationMessageForPendingMessage(....);
}
if (message == null) {
            // Create new state transition message
            message = createStateTransitionMessage(...);
}
```

jiajunwang: I was thinking if there is a way to avoid the same condition being checked inside generateCancellationMessageForPendingMessage and here. But it seems that there is no easy way.

However, to avoid the possible unexpected overwrite (for instance, if we change the private method logic later), I think we can change as following,

Option one,
`if (message == null && (pendingMessage == null || currentState.equalsIgnoreCase(pendingMessage.getToState()))) {`
Which means if there is no cancellation message and the the pending message has been done then we generate new ST.

Option two,
Remove generateCancellationMessageForPendingMessage and move all if-else conditions back to the caller methods. Because you do need to handle those conditions with different logic. That seems to be a clearer logic.

zhangmeng916: GenerateCancellationMessageForPendingMessage is used in different places. Doesn't look good to move that function back. We can guard it with "message == null" for now.

jiajunwang: I think the other place will never hit the other 2 conditions. But not 100% sure.
Anyway, both options work.

helix-core/src/test/java/org/apache/helix/integration/messaging/TestP2PNoDuplicatedMessage.java
```diff
@@ -172,7 +172,7 @@ public void testP2PStateTransitionEnabled() {
     }
 
     double ratio = ((double) p2pTrigged) / ((double) total);
-    Assert.assertTrue(ratio > 0.6, String
+    Assert.assertTrue(ratio == 1, String
```
jiajunwang, line 175: Impressive!

jiajunwang: nit, you can use Assert.assertEquals(p2pTrigged, total) now.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -225,6 +225,13 @@ private void generateMessage(final Resource resource, final BaseControllerDataPr
           // pending message's toState (no cancellation message should be generated in this case)
           if (pendingMessage == null || (message == null &&
```
jiajunwang, line 226: What happened to this condition?
Aren't they duplicate?

zhangmeng916:
This PR is ready to merge, approved by @dasahcc 
Final commit message:
Remove the waiting logic in message generation phase on message deletion if current state is already updated. This will help increase the rate of P2P message during mastership handoff.

PR #1072, opened 2020-06-08
Validate data write size limit in ZkClient

### LINKED ISSUE #1071: Znode write size limit in zkclient is not consistent with the one in znrecord serializers
There is configurable write size limit for znrecord serializer. But there is also a data size limit check in raw zkclient:
```
private void checkDataSizeLimit(byte[] data) {
    if (data != null && data.length > ZNRecord.SIZE_LIMIT) {
```
If configurable write size limit for znrecord serializer is different from the default size 1 MB, especially it is set to larger than 1 MB, writing won't work because it would be blocked by this default size check in zkclient.

We need to fix it. A good idea is to make both consistently same. But the system property `zk.serializer.znrecord.write.size.limit.bytes` is for a **ZNRecord serializer**, not sure if it is fine to put it at ZkClient level, as it will apply to all other serializers, as well. We may discuss this.

pkuwm:
### Issues

Fixes #1071 

### Description

If configurable write size limit for znrecord serializer is different from the default size 1 MB, especially it is set to larger than 1 MB, writing won't work because it would be blocked by this default size check in zkclient.

This PR also reads both `jute.maxbuffer` config and znrecord serializer write size limit, and validates the correctness of data size limit.

### Tests

- testDataSizeGreaterThanLimit
- testDataSizeLessThanLimit
- testInvalidWriteSizeLimitConfig

In zookeeper-api module.
```
[INFO] Tests run: 50, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 64.326 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 50, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:08 min
[INFO] Finished at: 2020-10-22T17:17:17-07:00
```

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -101,6 +101,7 @@ public class ZkClient implements Watcher {
   public final long _uid;
 
   // ZNode write size limit in bytes.
+  // TODO: use ZKConfig#JUTE_MAXBUFFER once bumping up ZK to 3.5.2+
   private static final int WRITE_SIZE_LIMIT =
       Integer.getInteger(ZkSystemPropertyKeys.JUTE_MAXBUFFER, ZNRecord.SIZE_LIMIT);
 
@@ -2505,7 +2506,6 @@ public class ZkClient implements Watcher {
     int serializerSize = ZNRecordUtil.getSerializerWriteSizeLimit();
     LOG.info("ZNRecord serializer write size limit: {}; ZkClient write size limit: {}",
         serializerSize, WRITE_SIZE_LIMIT);
-    // ZNRecord serializer write size limit should not be set greater than size limit in ZkClient
     if (serializerSize > WRITE_SIZE_LIMIT) {
       throw new IllegalStateException("ZNRecord serializer write size limit " + serializerSize
           + " is greater than ZkClient size limit " + WRITE_SIZE_LIMIT);
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -35,10 +35,11 @@ import java.util.concurrent.atomic.AtomicLong;
 import javax.management.JMException;
 
 import org.apache.helix.zookeeper.api.client.ChildrenSubscribeResult;
-import org.apache.helix.zookeeper.datamodel.SessionAwareZNRecord;
 import org.apache.helix.zookeeper.constant.ZkSystemPropertyKeys;
+import org.apache.helix.zookeeper.datamodel.SessionAwareZNRecord;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.exception.ZkClientException;
+import org.apache.helix.zookeeper.util.ZNRecordUtil;
 import org.apache.helix.zookeeper.zkclient.annotation.PreFetchChangedData;
 import org.apache.helix.zookeeper.zkclient.callback.ZkAsyncCallMonitorContext;
 import org.apache.helix.zookeeper.zkclient.callback.ZkAsyncCallbacks;
@@ -98,6 +99,10 @@ public class ZkClient implements Watcher {
 
   private static AtomicLong UID = new AtomicLong(0);
   public final long _uid;
+
+  // ZNode write size limit in bytes.
+  private static final int WRITE_SIZE_LIMIT =
+      Integer.getInteger(ZkSystemPropertyKeys.JUTE_MAXBUFFER, ZNRecord.SIZE_LIMIT);
 
   private final IZkConnection _connection;
   private final long _operationRetryTimeoutInMillis;
@@ -213,6 +218,7 @@ public class ZkClient implements Watcher {
     }
 
     _uid = UID.getAndIncrement();
+    validateWriteSizeLimitConfig();
 
     _connection = zkConnection;
     _pathBasedZkSerializer = zkSerializer;
@@ -718,7 +724,7 @@ public class ZkClient implements Watcher {
     long startT = System.currentTimeMillis();
     try {
       final byte[] dataBytes = dataObject == null ? null : serialize(dataObject, path);
-      checkDataSizeLimit(dataBytes);
+      checkDataSizeLimit(path, dataBytes);
 
       final String actualPath = retryUntilConnected(
           () -> getExpectedZookeeper(expectedSessionId).create(path, dataBytes, acl, mode));
@@ -1878,7 +1884,7 @@ public class ZkClient implements Watcher {
     long startT = System.currentTimeMillis();
     try {
       final byte[] data = serialize(datat, path);
-      checkDataSizeLimit(data);
+      checkDataSizeLimit(path, data);
       final Stat stat = (Stat) retryUntilConnected(new Callable<Object>() {
         @Override
         public Object call() throws Exception {
@@ -2006,12 +2012,15 @@ public class ZkClient implements Watcher {
     });
   }
 
-  private void checkDataSizeLimit(byte[] data) {
-    if (data != null && data.length > ZNRecord.SIZE_LIMIT) {
-      LOG.error(
-          "Data size larger than 1M, will not write to zk. Data (first 1k): " + new String(data)
-              .substring(0, 1024));
-      throw new ZkClientException("Data size larger than 1M");
+  private void checkDataSizeLimit(String path, byte[] data) {
+    if (data == null) {
+      return;
+    }
+
+    if (data.length > WRITE_SIZE_LIMIT) {
+      throw new ZkClientException("Data size of path " + path
+          + " is greater than write size limit "
+          + WRITE_SIZE_LIMIT + " bytes");
     }
   }
 
@@ -2480,8 +2489,7 @@ public class ZkClient implements Watcher {
 
     if (stat.getNumChildren() > NUM_CHILDREN_LIMIT) {
       LOG.error("Failed to get children for path {} because of connection loss. "
-              + "Number of children {} exceeds limit {}, aborting retry.", path,
-          stat.getNumChildren(),
+              + "Number of children {} exceeds limit {}, aborting retry.", path, stat.getNumChildren(),
           NUM_CHILDREN_LIMIT);
       // MarshallingErrorException could represent transport error: exceeding the
       // Jute buffer size. So use it to exit retry loop and tell that zk is not able to
@@ -2490,6 +2498,17 @@ public class ZkClient implements Watcher {
     } else {
       LOG.debug("Number of children {} is less than limit {}, not exiting retry.",
           stat.getNumChildren(), NUM_CHILDREN_LIMIT);
+    }
+  }
+
+  private void validateWriteSizeLimitConfig() {
+    int serializerSize = ZNRecordUtil.getSerializerWriteSizeLimit();
+    LOG.info("ZNRecord serializer write size limit: {}; ZkClient write size limit: {}",
+        serializerSize, WRITE_SIZE_LIMIT);
+    // ZNRecord serializer write size limit should not be set greater than size limit in ZkClient
+    if (serializerSize > WRITE_SIZE_LIMIT) {
+      throw new IllegalStateException("ZNRecord serializer write size limit " + serializerSize
+          + " is greater than ZkClient size limit " + WRITE_SIZE_LIMIT);
     }
   }
 }
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/constant/ZkSystemPropertyKeys.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/constant/ZkSystemPropertyKeys.java
@@ -60,4 +60,7 @@ public class ZkSystemPropertyKeys {
    */
   public static final String ZK_AUTOSYNC_ENABLED =
       "zk.zkclient.autosync.enabled";
+
+  /** System property key for jute.maxbuffer */
+  public static final String JUTE_MAXBUFFER = "jute.maxbuffer";
 }
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/ZNRecordUtil.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/ZNRecordUtil.java
@@ -50,7 +50,7 @@ public class ZNRecordUtil {
 
   /**
    * Returns ZNRecord serializer write size limit in bytes. If size limit is configured to be less
-   * than or equal to 0, the default value will be used instead.
+   * than or equal to 0, the default value {@link ZNRecord#SIZE_LIMIT} will be used instead.
    */
   public static int getSerializerWriteSizeLimit() {
     Integer writeSizeLimit =
--- a/zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/TestHelper.java
+++ b/zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/TestHelper.java
@@ -114,6 +114,14 @@ public class TestHelper {
     return fullClassName.substring(fullClassName.lastIndexOf('.') + 1);
   }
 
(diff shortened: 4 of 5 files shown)


helix-core/src/test/java/org/apache/helix/manager/zk/TestRawZkClient.java
```diff
@@ -819,4 +820,45 @@ public void testAsyncWriteOperations() {
       zkClient.delete("/tmp/asyncOversize");
     }
   }
+
+  @Test(expectedExceptions = ZkMarshallingError.class,
```
narendly, line 824: Thanks for adding this test! One favor to ask - we are actually in the process of moving this class to zookeeper-api (https://github.com/apache/helix/pull/1070).

Do you think we could hold off on this change until that is committed first to avoid merge conflicts?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1839,12 +1848,15 @@ protected void doRetry() {
...
-              .substring(0, 1024));
-      throw new ZkClientException("Data size larger than 1M");
+  private void checkDataSizeLimit(String path, byte[] data) {
+    if (data == null) {
+      return;
+    }
+
+    int sizeLimit = ZNRecordUtil.getSerializerWriteSizeLimit();
+    if (data.length > sizeLimit) {
+      throw new ZkMarshallingError(
```
narendly, line 1858: I agree that this is the right thing to do. Thanks for catching this and fixing this.

One concern I have is with the change of the exception type. I am not sure if changing it to ZkMarshallingError is the right thing to do because

1. ZkMarshallingError is usually thrown in ZkSerializer implementations.
2. Changing Exception types may cause backward-compatibility issues.

What do you think?

pkuwm: These questions are valid and natural ones which I also thought about.
1. The reason I change it is I would like to make it consistent with serializer. In helix zkclient, the data size check is right after serializing. `ZkMarshallingError` could also indicate the serialized data still exceeds the limit. So I feel `ZkMarshallingError` is more accurate than `ZkClientException`.
2. Actually before version 1.0.0, the exception type is `HelixException`. So changing to `ZkClientException` already breaks backward-compatibility if it breaks. Considering recent 1.0.0 zookeeper-api module users(helix-core 1.0.0 is deprecated), if a user explicitly uses this raw zkclient and catches this runtime exception, that may break. But I am not sure if that really exists in 1.0.0 usage, considering the APIs don't explicitly signature `ZkClientException`.

Though having these thoughts, I don't have a strong preference: we could keep the original exception type. If we would like to change the type, may be it is a good idea to do it in 1.0.1 to avoid future backward compatibility issue.

narendly: To give you more context, ZkMarshallingError is something that was inherited from 101tec.zkclient library and explicitly thrown for ZkSerializer implementations. Since this is our ZkClient API-level check outside the serializer, I think ZkClientException. 

Also, we've decided to convert all HelixExceptions to ZkClientException, so I think ZkClientException would be more appropriate.

dasahcc: +1. Let's not change the exception type of this.

pkuwm: Thanks for the input. Resolved.

jiajunwang:
I'm not convinced that this is a good idea.

For the non-Helix user, ZNRecord is not meaningful. How do you plan to document this config and let them configure ZK_SERIALIZER_**ZNRECORD**_WRITE_SIZE_LIMIT_BYTES for the general ZK usage?
Secondly, we shall not block the write in the ZkClient if the write size is smaller than the jute.maxbuffer configuration. That is not very meaningful. On the other hand, if we have a large ZK_SERIALIZER_ZNRECORD_WRITE_SIZE_LIMIT_BYTES configured, but the jute.maxbuffer is small, what we shall do? The write will still fail.

My suggestion,
- The raw ZkClient shall check size based on jute.maxbuffer, and maybe we leave some buffer like the default limit is current doing.
- If jute.maxbuffer is not configured, then we apply the default value according to the ZK document.

pkuwm:
> I'm not convinced that this is a good idea.
> 
> For the non-Helix user, ZNRecord is not meaningful. How do you plan to document this config and let them configure ZK_SERIALIZER_**ZNRECORD**_WRITE_SIZE_LIMIT_BYTES for the general ZK usage?
> Secondly, we shall not block the write in the ZkClient if the write size is smaller than the jute.maxbuffer configuration. That is not very meaningful. On the other hand, if we have a large ZK_SERIALIZER_ZNRECORD_WRITE_SIZE_LIMIT_BYTES configured, but the jute.maxbuffer is small, what we shall do? The write will still fail.
> 
> My suggestion,
> 
> * The raw ZkClient shall check size based on jute.maxbuffer, and maybe we leave some buffer like the default limit is current doing.
> * If jute.maxbuffer is not configured, then we apply the default value according to the ZK document.

@jiajunwang Thanks for the comments. I have the same concern about using the same config at zkclient level. This is why I would like to have it discussed at #1071 and confirmed: 
> A good idea is to make both consistently same. But the system property zk.serializer.znrecord.write.size.limit.bytes is for a ZNRecord serializer, not sure if it is fine to put it at ZkClient level, as it will apply to all other serializers, as well. We may discuss this.

I have one thought if we use `jute.maxbuff` value here. If `jute.maxbuff` is set, considering client and server's values are the same. zkclient might be able to write but fail to read, due to the extra fields in ZK server. This is why we want to leave some buffer. 

But it may be difficult to determine the buffer size. Say we leave 5 KB as buffer, the value would be maxbuffer - 5 KB.  But the value of ZK_SERIALIZER_**ZNRECORD**_WRITE_SIZE_LIMIT_BYTES might fall in the range of [maxbuffer, maxbuffer - 5 KB]. Even serializer check passes but it could be blocked by this check maxbuffer - 5 KB.

If we are using jute.maxbuffer for this check, we have to document to make it clear  ZK_SERIALIZER_ZNRECORD_WRITE_SIZE_LIMIT_BYTES should be less or equal to maxbuffer - 5KB.

If possible, I would prefer to just have one config `zkclient.znode.write.size.limit` at ZkClient level, instead of znrecord serializer. So it satisfies all cases. Let's discuss further.

jiajunwang:
1. jute.maxbuffer is a Zookeeper lib configuration, I don't think we need to add additional doc about it.
2. Regarding how to determine the buffer, I think we can just do what we are doing now. Keeping the behavior the same.
3. To avoid confusion, we can do the configuration check when the ZkClient is constructed. Just fail the constructing if the ZK_SERIALIZER_ZNRECORD_WRITE_SIZE_LIMIT_BYTES is configured larger than the ZkClient write limitation. With a clear Exception, users will understand what to do.

pkuwm:
> * jute.maxbuffer is a Zookeeper lib configuration, I don't think we need to add additional doc about it.
> * Regarding how to determine the buffer, I think we can just do what we are doing now. Keeping the behavior the same.
> * To avoid confusion, we can do the configuration check when the ZkClient is constructed. Just fail the constructing if the ZK_SERIALIZER_ZNRECORD_WRITE_SIZE_LIMIT_BYTES is configured larger than the ZkClient write limitation. With a clear Exception, users will understand what to do.

@jiajunwang Thanks for the input. Size margin would be using the current margin: default 1 MB - ZNRecord size limit (1000 KB) = 24 KB.  I've updated the PR.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -73,6 +84,10 @@
   private static Logger LOG = LoggerFactory.getLogger(ZkClient.class);
   private static long MAX_RECONNECT_INTERVAL_MS = 30000; // 30 seconds
 
+  // Default value for system property jute.maxbuffer
+  // It specifies the maximum size of the data that can be stored in a znode.
+  private static final int DEFAULT_JUTE_MAXBUFFER = 0xfffff;
```
jiajunwang, line 89: I'm not sure if we can directly use the value ClientCnxn.packetLen. It is public and static.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2216,4 +2236,24 @@ private void validateCurrentThread() {
...
+    // ZNRecord serializer write size limit should not be set greater than size limit in ZkClient
+    if (serializerSize > zkClientSize) {
+      throw new IllegalStateException("ZNRecord serializer write size limit: " + serializerSize
+          + " is greater than size limit in ZkClient: " + zkClientSize);
+    }
+  }
+
+  private int getWriteSizeLimit() {
+    // The size margin left in default SIZE_LIMIT
+    int sizeMargin = DEFAULT_JUTE_MAXBUFFER - ZNRecord.SIZE_LIMIT;
```
jiajunwang, line 2252: This is awkward. How about just create a constant for margin and deprecate the SIZE_LIMIT?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2216,4 +2236,24 @@ private void validateCurrentThread() {
       throw new IllegalArgumentException("Must not be done in the zookeeper event thread.");
     }
   }
+
+  private void validateWriteSizeLimitConfig() {
+    int serializerSize = ZNRecordUtil.getSerializerWriteSizeLimit();
+    int zkClientSize = getWriteSizeLimit();
+    // ZNRecord serializer write size limit should not be set greater than size limit in ZkClient
+    if (serializerSize > zkClientSize) {
+      throw new IllegalStateException("ZNRecord serializer write size limit: " + serializerSize
```
jiajunwang, line 2245: With compression on, the serializer may always return a much smaller package than the input size.
So do we have to fail it here?

It is debatable.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: ZNRecordSerializer is also using the SIZE_LIMIT to determine the actual size limit. Maybe we still need it. Otherwise, we still have `sizeLimit = JUTE_MAXBUFFER - MARGIN`. I understand this calculation is awkward, but just it is just a local variable. Or we could add one more constant for margin so we don't have to calculate it again here.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: Answer is no. That one is different: it's for the response packet size and it's 4 MB. We are looking at server side jute.maxbuffer that is by default 1 MB in `BinaryInputArchive`. It does no harm to define our own private constant and the default value 1 MB is known.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: In this case, we could add one more check for compression on/off. If compression is turned on, we allow serializerSize > zkClientSize. Agreed?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: The main concern is that you introduced some internal logic that is not configurable or commented directly where the variables are defined. Who will be able to tell the margin is derived by the default max buffer size and the default size limit?
First, this logic is not reasonable, why the margin is based on the default configs? Secondly, it introduces more chaos to our code.

jiajunwang: Just add a MARGIN value please.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: What if ZK changes the default max buffer in some new versions? Are we going to change this number accordingly? So my suggestion is that,
1. We fine the default value in Zookeeper code.
2. if 1 is not possible, then we can just use the current SIZE_LIMIT for the default value.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2216,4 +2236,24 @@ private void validateCurrentThread() {
...
+    }
+  }
+
+  private int getWriteSizeLimit() {
+    // The size margin left in default SIZE_LIMIT
+    int sizeMargin = DEFAULT_JUTE_MAXBUFFER - ZNRecord.SIZE_LIMIT;
+
+    // jute.maxbuffer is not directly used as size limit in ZkClient.
+    // Instead, leave some margin as ZNRecord.SIZE_LIMIT and use (jute.maxbuffer - sizeMargin)
+    return Integer.getInteger(ZkSystemPropertyKeys.JUTE_MAXBUFFER, DEFAULT_JUTE_MAXBUFFER)
```
jiajunwang, line 2256: So according to what I comment before, the logic here can be,
if (value of ZkSystemPropertyKeys.JUTE_MAXBUFFER exits) {
  return value of ZkSystemPropertyKeys.JUTE_MAXBUFFER - MARGIN;
} else {
  return SIZE_LIMIT;
}

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: Let's just keep it strict for now.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: Then you run into the ZK server side buffer issue. The write might be done successfully, but the read will fail.
I think we agreed to keep the buffer until we bump up the ZK version to a newer one and the buffer issue should be fixed in that version.

pkuwm: @jiajunwang For default value 1000KB in ZNRecord SIZE_LIMIT, we still leave a margin that is 24 KB. But for the value that users config `jute.maxbuffer`, we should honor it as what is done in ZNRecordSerializer. We don't leave a margin in ZNRecordSerializer, right? So if user configs jute.maxbuffer, we expect them to understand it well. Say they already consider a margin in the value jute.maxbuffer, we should not(users don't expect that, either) leave another redundant margin. 

Say a user sets `jute.maxbuff` to 975KB that they think it already leaves enough room as margin, we don't want to deduct it again 975 - 24 = 951KB as the real value. It would also surprise user: if data is 960KB that user expects it to be written successfully, but it actually fails because of the extra margin.

So for the custom value if a user sets`jute.maxbuff`, I don't think we should deduct the value with the 24KB margin.

lei-xia:
Are we still planning to merge this PR in?  What is pending here?

pkuwm:
> Are we still planning to merge this PR in?  What is pending here?

@lei-xia Yes. It was not urgent so it was left for a while. We will still need this to make it more accurate and consistent with znrecord size limit config. Will polish the PR and get it reviewed this week.

jiajunwang (approved):
LGTM!
Please address the minor comments.

pkuwm:
Thanks, @jiajunwang, for the review.

The PR is ready to be merged, approved by @jiajunwang 

If configurable write size limit for znrecord serializer is different from the default size 1 MB, especially it is set to larger than 1 MB, writing won't work because it would be blocked by this default size check in zkclient. This commit makes ZkClient to validate the data write size limit.

PR #803, opened 2020-02-25
Remove workflow garbage collection

alirezazamani:
### Issues
- [x] My PR addresses the following Helix issues and references them in the PR title:
Fixes #802.

### Description
- [x] Here are some details about my PR, including screenshots of any UI changes:

There is a race condition between workflow garbage collection and cache refresh which cause problems for Helix task scheduler. This PR removed the garbage collection.

- [x] The following is the result of the "mvn test" command on the appropriate module:

Results :

Failed tests: 
  TestResourceChangeDetector.testRemoveInstance:284 » ZkClient Failed to delete ...
org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceNonRack.testLackEnoughInstances(org.apache.helix.integration.rebalancer.CrushRebalancers.TestCrushAutoRebalanceNonRack)
  Run 1: PASS
  Run 2: TestCrushAutoRebalanceNonRack.testLackEnoughInstances:250 » ZkClient Failed to...

  TestWorkflowTermination.testWorkflowJobFail:251->verifyWorkflowCleanup:257 expected:<true> but was:<false>

Tests run: 1081, Failures: 3, Errors: 0, Skipped: 4

[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:11 h
[INFO] Finished at: 2020-02-24T15:49:43-08:00
[INFO] ------------------------------------------------------------------------

These tests passed when I run them individually.

### Commits

- [x] My commits all reference appropriate Apache Helix GitHub issues in their subject lines, and I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Code Quality
- [x] My diff has been formatted using helix-style.xml

PR #1076, opened 2020-06-09
Recover Workflow Garbage Collection Logic

### LINKED ISSUE #1075: Recover Workflow Garbage Collection Logic
Previously, the workflow garbage collection logic was removed in fear of race conditions (PR: #803 ). As we see value in bringing back workflow garbage collection, we are getting rid of the race conditions and bringing back the logic. 

Specifically, two items need to be completed:
1. Avoid direct dependency of garbage collection stage on the cache. We can instead deliver the relevant content of the cache into the stage during sync time, and use it during async time. This way the race condition with cache is no longer an issue. 
2. We are getting rid of an unnecessary JobConfig write that serves no purpose to avoid possible race conditions.

NealSun96:
### Issues

Fixes #1075

### Description

Previously, the workflow garbage collection logic was removed in fear of race conditions (PR: #803 ). As we see value in bringing back workflow garbage collection, we are getting rid of the race conditions and bringing back the logic. 

Specifically, two items need to be completed:
1. Avoid direct dependency of garbage collection stage on the cache. We can instead deliver the relevant content of the cache into the stage during sync time, and use it during async time. This way the race condition with cache is no longer an issue. 
2. We are getting rid of an unnecessary JobConfig write that serves no purpose to avoid possible race conditions.  

### Tests

testGetExpiredJobsFromCache, testWorkflowContextGarbageCollection(recovered from old PR), testWorkflowGarbageCollection

```
[ERROR] Tests run: 1153, Failures: 3, Errors: 0, Skipped: 1, Time elapsed: 4,609.951 s <<< FAILURE! - in TestSuite
[ERROR] testEnableCompressionResource(org.apache.helix.integration.TestEnableCompression)  Time elapsed: 218.35 s  <<< FAILURE!
        at org.apache.helix.integration.TestEnableCompression.testEnableCompressionResource(TestEnableCompression.java:117)

[ERROR] testStateTransitionTimeOut(org.apache.helix.integration.paticipant.TestStateTransitionTimeoutWithResource)  Time elapsed: 36.446 s  <<< FAILURE!
        at org.apache.helix.integration.paticipant.TestStateTransitionTimeoutWithResource.testStateTransitionTimeOut(TestStateTransitionTimeoutWithResource.java:171)

[ERROR] testHappyPathExtOpZkCacheBaseDataAccessor(org.apache.helix.manager.zk.TestZkCacheAsyncOpSingleThread)  Time elapsed: 1.191 s  <<< FAILURE!
java.lang.AssertionError: zkCache doesn't match data on Zk expected:<true> but was:<false>
        at org.apache.helix.manager.zk.TestZkCacheAsyncOpSingleThread.testHappyPathExtOpZkCacheBaseDataAccessor(TestZkCacheAsyncOpSingleThread.java:214)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestZkCacheAsyncOpSingleThread.testHappyPathExtOpZkCacheBaseDataAccessor:214 zkCache doesn't match data on Zk expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1153, Failures: 3, Errors: 0, Skipped: 1
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:16 h
[INFO] Finished at: 2020-07-14T11:10:05-07:00
```
Rerun
```
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 51.087 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  56.554 s
[INFO] Finished at: 2020-07-14T11:14:48-07:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
@@ -41,6 +41,6 @@ public enum AttributeName {
   LastRebalanceFinishTimeStamp,
   ControllerDataProvider,
   STATEFUL_REBALANCER,
-  WORKFLOW_CONFIG_MAP,
-  RESOURCE_CONTEXT_MAP
+  WORKFLOWS_TO_BE_DELETED,
+  EXPIRED_JOBS_MAP
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
@@ -40,5 +40,7 @@ public enum AttributeName {
   PipelineType,
   LastRebalanceFinishTimeStamp,
   ControllerDataProvider,
-  STATEFUL_REBALANCER
+  STATEFUL_REBALANCER,
+  WORKFLOW_CONFIG_MAP,
+  RESOURCE_CONTEXT_MAP
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
@@ -42,7 +42,7 @@ public enum AttributeName {
   ControllerDataProvider,
   STATEFUL_REBALANCER,
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
-  TO_BE_DELETED_WORKFLOWS,
+  TO_BE_PURGED_WORKFLOWS,
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
-  TO_BE_DELETED_JOBS_MAP
+  TO_BE_PURGED_JOBS_MAP
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
@@ -42,7 +42,7 @@ public enum AttributeName {
   ControllerDataProvider,
   STATEFUL_REBALANCER,
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
-  WORKFLOWS_TO_BE_DELETED,
+  TO_BE_DELETED_WORKFLOWS,
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
-  EXPIRED_JOBS_MAP
+  TO_BE_DELETED_JOBS_MAP
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
@@ -41,6 +41,8 @@ public enum AttributeName {
   LastRebalanceFinishTimeStamp,
   ControllerDataProvider,
   STATEFUL_REBALANCER,
+  // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
   WORKFLOWS_TO_BE_DELETED,
+  // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
   EXPIRED_JOBS_MAP
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
@@ -1,5 +1,6 @@
 package org.apache.helix.controller.stages;
 
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
@@ -28,23 +29,10 @@ public class TaskGarbageCollectionStage extends AbstractAsyncBaseStage {
 
   @Override
   public void process(ClusterEvent event) throws Exception {
-    WorkflowControllerDataProvider dataProvider =
-        event.getAttribute(AttributeName.ControllerDataProvider.name());
-    event.addAttribute(AttributeName.WORKFLOW_CONFIG_MAP.name(),
-        dataProvider.getWorkflowConfigMap());
-    event.addAttribute(AttributeName.RESOURCE_CONTEXT_MAP.name(), dataProvider.getContexts());
-
-    super.process(event);
-  }
-
-  @Override
-  public void execute(ClusterEvent event) {
-    Map<String, WorkflowConfig> workflowConfigMap =
-        event.getAttribute(AttributeName.WORKFLOW_CONFIG_MAP.name());
-    Map<String, ZNRecord> resourceContextMap =
-        event.getAttribute(AttributeName.RESOURCE_CONTEXT_MAP.name());
+    // Use main thread to compute what jobs need to be purged, and what workflows need to be gc'ed.
+    // This is to avoid race conditions since the cache will be modified. After this work, then the
+    // async work will happen.
     HelixManager manager = event.getAttribute(AttributeName.helixmanager.name());
-
     if (manager == null) {
       LOG.warn(
           "HelixManager is null for event {}({}) in cluster {}. Skip TaskGarbageCollectionStage.",
@@ -52,26 +40,69 @@ public class TaskGarbageCollectionStage extends AbstractAsyncBaseStage {
       return;
     }
 
-    Set<WorkflowConfig> existingWorkflows = new HashSet<>(workflowConfigMap.values());
-    for (WorkflowConfig workflowConfig : existingWorkflows) {
-      // clean up the expired jobs if it is a queue.
+    WorkflowControllerDataProvider dataProvider =
+        event.getAttribute(AttributeName.ControllerDataProvider.name());
+    for (Map.Entry<String, ZNRecord> entry : dataProvider.getContexts().entrySet()) {
+      WorkflowConfig workflowConfig = dataProvider.getWorkflowConfig(entry.getKey());
       if (workflowConfig != null && (!workflowConfig.isTerminable() || workflowConfig
           .isJobQueue())) {
-        String workflowId = workflowConfig.getWorkflowId();
-        if (resourceContextMap.get(workflowId) != null) {
-          try {
-            TaskUtil.purgeExpiredJobs(workflowId, workflowConfig,
-                new WorkflowContext(resourceContextMap.get(workflowId)), manager,
-                _rebalanceScheduler);
-          } catch (Exception e) {
-            LOG.warn("Failed to purge job for workflow {}!", workflowId, e);
+        WorkflowContext workflowContext = dataProvider.getWorkflowContext(entry.getKey());
+        long purgeInterval = workflowConfig.getJobPurgeInterval();
+        long currentTime = System.currentTimeMillis();
+        if (purgeInterval > 0
+            && workflowContext.getLastJobPurgeTime() + purgeInterval <= currentTime) {
+          // Find jobs that are ready to be purged
+          Set<String> expiredJobs =
+              TaskUtil.getExpiredJobsFromCache(dataProvider, workflowConfig, workflowContext);
+          if (!expiredJobs.isEmpty()) {
+            dataProvider.setExpiredJobsForWorkflow(workflowConfig.getWorkflowId(), expiredJobs);
           }
-        } else {
-          LOG.warn("Workflow {} context does not exist!", workflowId);
+          scheduleNextJobPurge(workflowConfig.getWorkflowId(), currentTime, purgeInterval,
+              _rebalanceScheduler, manager);
+        }
+      } else if (workflowConfig == null) {
+        if (entry.getValue() != null && entry.getValue().getId()
+            .equals(TaskUtil.WORKFLOW_CONTEXT_KW)) {
+          // Find workflows that need to be purged
+          dataProvider.addToBeDeletedWorkflow(entry.getKey());
         }
       }
     }
 
-    TaskUtil.workflowGarbageCollection(workflowConfigMap, resourceContextMap, manager);
+    super.process(event);
+  }
+
+  @Override
+  public void execute(ClusterEvent event) {
+    HelixManager manager = event.getAttribute(AttributeName.helixmanager.name());
+    if (manager == null) {
+      LOG.warn(
+          "HelixManager is null for event {}({}) in cluster {}. Skip TaskGarbageCollectionStage async execution.",
+          event.getEventId(), event.getEventType(), event.getClusterName());
+      return;
+    }
+
+    Map<String, Set<String>> expiredJobsMap = new HashMap<>();
+    Set<String> toBeDeletedWorkflows = new HashSet<>();
+
+    for (Map.Entry<String, Set<String>> entry : expiredJobsMap.entrySet()) {
+      try {
+        TaskUtil.purgeExpiredJobs(entry.getKey(), entry.getValue(), manager,
+            _rebalanceScheduler);
+      } catch (Exception e) {
+        LOG.warn("Failed to purge job for workflow {}!", entry.getKey(), e);
+      }
+    }
+
+    TaskUtil.workflowGarbageCollection(toBeDeletedWorkflows, manager);
+  }
+
+  private static void scheduleNextJobPurge(String workflow, long currentTime, long purgeInterval,
+      RebalanceScheduler rebalanceScheduler, HelixManager manager) {
+    long nextPurgeTime = currentTime + purgeInterval;
+    long currentScheduledTime = rebalanceScheduler.getRebalanceTime(workflow);
+    if (currentScheduledTime == -1 || currentScheduledTime > nextPurgeTime) {
+      rebalanceScheduler.scheduleRebalance(manager, workflow, nextPurgeTime);
+    }
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
@@ -63,12 +63,10 @@ public class TaskGarbageCollectionStage extends AbstractAsyncBaseStage {
           scheduleNextJobPurge(workflowConfig.getWorkflowId(), currentTime, purgeInterval,
               _rebalanceScheduler, manager);
         }
-      } else if (workflowConfig == null) {
-        if (entry.getValue() != null && entry.getValue().getId()
-            .equals(TaskUtil.WORKFLOW_CONTEXT_KW)) {
-          // Find workflows that need to be purged
-          workflowsToBeDeleted.add(entry.getKey());
-        }
+      } else if (workflowConfig == null && entry.getValue() != null && entry.getValue().getId()
+          .equals(TaskUtil.WORKFLOW_CONTEXT_KW)) {
+        // Find workflows that need to be purged
+        workflowsToBeDeleted.add(entry.getKey());
       }
     }
     event.addAttribute(AttributeName.EXPIRED_JOBS_MAP.name(),
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
@@ -47,7 +47,7 @@ public class TaskGarbageCollectionStage extends AbstractAsyncBaseStage {
 
     if (manager == null) {
       LOG.warn(
-          " HelixManager is null for event {}({}) in cluster {}. Skip TaskGarbageCollectionStage.",
+          "HelixManager is null for event {}({}) in cluster {}. Skip TaskGarbageCollectionStage.",
           event.getEventId(), event.getEventType(), event.getClusterName());
       return;
     }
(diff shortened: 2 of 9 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
```diff
@@ -23,34 +27,53 @@ public AsyncWorkerType getAsyncWorkerType() {
...
+        event.getAttribute(AttributeName.WORKFLOW_CONFIG_MAP.name());
+    Map<String, ZNRecord> resourceContextMap =
+        event.getAttribute(AttributeName.RESOURCE_CONTEXT_MAP.name());
     HelixManager manager = event.getAttribute(AttributeName.helixmanager.name());
 
-    if (dataProvider == null || manager == null) {
+    if (manager == null) {
       LOG.warn(
-          "ResourceControllerDataProvider or HelixManager is null for event {}({}) in cluster {}. Skip TaskGarbageCollectionStage.",
+          " HelixManager is null for event {}({}) in cluster {}. Skip TaskGarbageCollectionStage.",
```
narendly, line 50: Why are we starting with a space?

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
```diff
@@ -23,34 +27,53 @@ public AsyncWorkerType getAsyncWorkerType() {
   }
 
   @Override
-  public void execute(ClusterEvent event) {
+  public void process(ClusterEvent event) throws Exception {
     WorkflowControllerDataProvider dataProvider =
         event.getAttribute(AttributeName.ControllerDataProvider.name());
+    event.addAttribute(AttributeName.WORKFLOW_CONFIG_MAP.name(),
+        dataProvider.getWorkflowConfigMap());
+    event.addAttribute(AttributeName.RESOURCE_CONTEXT_MAP.name(), dataProvider.getContexts());
```
narendly, line 35: How would this affect the memory footprint of each event?

Also, is a soft copy okay here?

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
```diff
@@ -23,34 +27,53 @@ public AsyncWorkerType getAsyncWorkerType() {
...
       LOG.warn(
-          "ResourceControllerDataProvider or HelixManager is null for event {}({}) in cluster {}. Skip TaskGarbageCollectionStage.",
+          " HelixManager is null for event {}({}) in cluster {}. Skip TaskGarbageCollectionStage.",
           event.getEventId(), event.getEventType(), event.getClusterName());
       return;
     }
 
-    Set<WorkflowConfig> existingWorkflows =
-        new HashSet<>(dataProvider.getWorkflowConfigMap().values());
+    Set<WorkflowConfig> existingWorkflows = new HashSet<>(workflowConfigMap.values());
```
narendly, line 55: If we assumed soft copy was okay from above, then why are we creating a new set? Also this is not a hard copy either is it?

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
```diff
@@ -23,34 +27,53 @@ public AsyncWorkerType getAsyncWorkerType() {
...
-        try {
-          TaskUtil.purgeExpiredJobs(workflowConfig.getWorkflowId(), workflowConfig,
-              dataProvider.getWorkflowContext(workflowConfig.getWorkflowId()), manager,
-              _rebalanceScheduler);
-        } catch (Exception e) {
-          LOG.warn(String.format("Failed to purge job for workflow %s with reason %s",
-              workflowConfig.getWorkflowId(), e.toString()));
+        String workflowId = workflowConfig.getWorkflowId();
+        if (resourceContextMap.containsKey(workflowId)
+            && resourceContextMap.get(workflowId) != null) {
```
narendly, line 62: Aren't these two checks same checks? Why are we doing the same thing twice?

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
```diff
@@ -23,34 +27,53 @@ public AsyncWorkerType getAsyncWorkerType() {
...
+          try {
+            TaskUtil.purgeExpiredJobs(workflowId, workflowConfig,
+                new WorkflowContext(resourceContextMap.get(workflowId)), manager,
+                _rebalanceScheduler);
+          } catch (Exception e) {
+            LOG.warn(String.format("Failed to purge job for workflow %s with reason %s", workflowId,
+                e.toString()));
```
narendly, line 69: Use parameterized logging.

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
```diff
@@ -23,34 +27,53 @@ public AsyncWorkerType getAsyncWorkerType() {
...
+          }
+        } else {
+          LOG.warn(String.format("Workflow %s context does not exist!", workflowId));
```
narendly, line 72: Use parameterized logging.

helix-core/src/test/java/org/apache/helix/integration/task/TestWorkflowContextWithoutConfig.java
```diff
@@ -103,6 +103,73 @@ public void testWorkflowContextWithoutConfig() throws Exception {
...
+    // expired.
+    boolean workflowExpired = TestHelper.verify(() -> {
+      WorkflowContext wCtx = _driver.getWorkflowContext(workflowName);
+      WorkflowConfig wCfg = _driver.getWorkflowConfig(workflowName);
+      IdealState idealState = _admin.getResourceIdealState(CLUSTER_NAME, workflowName);
+      return (wCtx == null && wCfg == null && idealState == null);
+    }, TestHelper.WAIT_DURATION);
+    Assert.assertTrue(workflowExpired);
+
+
```
narendly, line 141: Redundant line

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -1037,6 +1037,53 @@ public static void purgeExpiredJobs(String workflow, WorkflowConfig workflowConf
...
   }
 
+  /**
+   * The function that loops through the all existing workflow contexts and removes IdealState and
+   * workflow context of the workflow whose workflow config does not exist.
+   * @param workflowConfigMap
+   * @param resourceContextMap
+   * @param manager
+   */
+  public static void workflowGarbageCollection(final Map<String, WorkflowConfig> workflowConfigMap,
```
jiajunwang, line 1047: Is this tool related to the TaskGarbageCollectionStage?
It would be better to split the PR if they are not strictly related.

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
jiajunwang: I don't think a reference to the map object itself here is OK. My understanding is that you want to persist in the cached map before it is modified in the later stages. In this case, you need to copy the map, not just pass the reference.

Also, if this change passes our current test, then it may indicate the test does not cover the race condition case. To justify that this PR really fixes the race condition, please add the corresponding test case which fails because of the race condition without the fix.

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
```diff
@@ -23,34 +27,53 @@ public AsyncWorkerType getAsyncWorkerType() {
   }
 
   @Override
-  public void execute(ClusterEvent event) {
+  public void process(ClusterEvent event) throws Exception {
     WorkflowControllerDataProvider dataProvider =
         event.getAttribute(AttributeName.ControllerDataProvider.name());
+    event.addAttribute(AttributeName.WORKFLOW_CONFIG_MAP.name(),
+        dataProvider.getWorkflowConfigMap());
```
jiajunwang, line 34: Just curious, why are we adding all the workflow/resource contexts here?
IMHO, a better design would be after all workflows, jobs being processed, add the rest unprocessed workflows, jobs to the async GC thread. Will that be better?

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
NealSun96: This is a very good point for reviewing. First, my mistake: this should be a deep copy. Now, with the concerns raised here, we have a couple options:

1. We still do deep copies. I'm only copying the maps that are necessary for this stage to run, which is the 2 maps for WorkflowConfig and Context. 
Pro: Doesn't change existing logic
Con: Uses more memory. @narendly I'd like to know how concerned are you about the memory usage for them? 
2. We do what @jiajunwang suggests; we only pass in the ones that need to be GC'ed. 
Pro: Less memory usage.
Con: More time spent in the main thread. It's actually not a fast thing to do because `getExpiredJobs()` contains Zk read operations. 
3. @dasahcc proposed another approach: the event is like a trigger, and the garbage collection stage figures out what to GC by itself. The stage performs 2 Zk reads to get WorkflowConfigs and WorkflowContexts. Imo 2 Zk reads are not many comparing to the existing Zk reads this stage does (per job level). 

Feel free to let me know what you guys think. 

@jiajunwang About the test comment: I agree with what you said 100% - there should have been a test. The race condition was introduced due to cache refresh, but whatever approach we take, this stage will now be completely unrelated to the cache. Should we still write a test that involves cache refresh()? I don't think it would make sense NOW because this stage is no longer related to it; it would make sense before.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
NealSun96: This is the logic I'm recovering with this PR; it's also the logic that @alirezazamani deleted in the linked PR in the description. :)

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
NealSun96: I have also explained it in an earlier comment: the logic of this stage is to loop through every workflow and figure out what to purge. We could do what you said before this stage, but that adds delay to the main pipeline.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
jiajunwang: You are not answering the question. Sorry but I don't have that context.
So is it related to the stage code, please?

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
jiajunwang: So which way do you think would be better? And what is the additional latency we are talking about here?

So what I proposed is that we finish the main rebalance calculation first. During that computing, we know what workflow/job is no longer alive. Then we record them in the TaskCache. Then at the end of the pipeline, pass that list to the GC stage which will remove thing async. I think it would be helpful because,
1. no additional calculation to check if an object needs to be purged, since the rebalance pipeline will have to check for this anyway.
2. no parallel logic to determine if a job or workflow needs to be purged separately. So less conflict, and fewer race conditions.

There might be TF logic prevents us from doing that, but please investigate and we can talk if there are any blocks.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
NealSun96: I see the confusion: this function is called on line 75 in the stage, please see above. 

I was trying to give you the context: this function was deleted in https://github.com/apache/helix/pull/803, and recovering it is the sole purpose of this PR.

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
NealSun96: I understand your proposal better now, and after looking at it carefully, we can ditch the Zk reads and preprocess lightweight logic. In the preprocess section of the stage, we can loop through all WorkflowContext and figure out 1. what workflows need to be garbage collected, and 2. what jobs need to be purged. All work can be done based on cache.

For your point 1 you did mention that we can do the calculation in "rebalance pipeline" (I assume is the synonym for TaskSchedulingStage". I doing the calculation in GarbageCollectionStage for code clarity, and the overhead isn't much because it's 1 loop through workflows using in-memory operations. What do you think?

jiajunwang: > I prefer doing the calculation in GarbageCollectionStage for code clarity, and the overhead isn't much because it's 1 loop through workflows using in-memory operations. What do you think?

As long as there is no duplicate logic, I'm fine. The concern is that in TaskSchedulingStage, we may have the logic to loop all the tasks for the non-removable ones, then in the GarbageCollectionStage, we loop all the tasks for the removable ones. That would be concerning because of the duplicate logic and potential conflicts.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
jiajunwang: Got you.

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
jiajunwang: I agree with what you replied about the test. Let's address the other comments.

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
jiajunwang: On my second thought, we should do the check in the async GC thread. The reason is that we may have a job with the same name re-created. So it is possible that when TaskSchedulingStage is running, JOBA should be GCed, but when it is the time to GC, JOBA has already been recreated.
To prevent problems in this kind of scenario, I think we should do the check inside GC thread. And also delete the jobs inside the GC thread only after the check.

NealSun96: So you're right that looping is done twice, but there is no exact duplication for sure. My point is that TaskSchedulingStage is doing a lot of things, to a point where I think it's better for the code base if we allow the other stage to do work just to improve clarity. Is the delay of running a loop a matter of concern? If so we should do it in the TaskSchedulingStage loop. 

About your second thought: 1. even if we do the check in async thread, same job name is still an issue because the operation is not atomic. It could be recreated just before deletion; 2. if this is a concern for jobs, this should also be a concern for workflow? If we need to do the check for workflow in async thread, we face the same problem again.

jiajunwang: > About your second thought: 1. even if we do the check in async thread, same job name is still an issue because the operation is not atomic. It could be recreated just before deletion; 2. if this is a concern for jobs, this should also be a concern for workflow? If we need to do the check for workflow in async thread, we face the same problem again.

Probably easier to check and avoid the race conditions when the logics are in the same block. But I do think this is an issue that we want to solve when you claim to address race condition. Are you implying we do this in the future?

NealSun96: @jiajunwang Thank you for the offline discussion! 

Conclusion: 1. we will do preprocessing in GarbageCollectionStage. We will sacrifice a bit of performance for better code clarity. All garbage collection logic is better to be in 1 class. 
2. There is no good way to deal with the situation described in this thread (same job name). We are not expanding the scope of this PR to cover that, and therefore it will be left as a TODO item for now.

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
NealSun96: Closing this thread as deep copy is no longer a concern anymore with the new approach @jiajunwang proposed.

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
NealSun96: Outdated

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
```diff
@@ -40,5 +40,7 @@
   PipelineType,
   LastRebalanceFinishTimeStamp,
   ControllerDataProvider,
-  STATEFUL_REBALANCER
+  STATEFUL_REBALANCER,
+  WORKFLOW_CONFIG_MAP,
```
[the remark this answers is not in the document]
NealSun96: Outdated

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
```diff
@@ -40,5 +40,9 @@
   PipelineType,
   LastRebalanceFinishTimeStamp,
   ControllerDataProvider,
-  STATEFUL_REBALANCER
+  STATEFUL_REBALANCER,
+  // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
+  WORKFLOWS_TO_BE_DELETED,
+  // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
+  EXPIRED_JOBS_MAP
```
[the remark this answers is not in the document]
NealSun96: Could you elaborate please?

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
```diff
@@ -23,34 +29,89 @@ public AsyncWorkerType getAsyncWorkerType() {
...
+        long currentTime = System.currentTimeMillis();
+        if (purgeInterval > 0
+            && workflowContext.getLastJobPurgeTime() + purgeInterval <= currentTime) {
+          // Find jobs that are ready to be purged
+          Set<String> expiredJobs =
+              TaskUtil.getExpiredJobsFromCache(dataProvider, workflowConfig, workflowContext);
+          if (!expiredJobs.isEmpty()) {
+            expiredJobsMap.put(workflowConfig.getWorkflowId(), expiredJobs);
+          }
+          scheduleNextJobPurge(workflowConfig.getWorkflowId(), currentTime, purgeInterval,
```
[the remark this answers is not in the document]
NealSun96: This line is moved from the original purgeExpiredJob and moved here. The reason is that by triggering, we no longer need to pass `purgeInterval` to the async thread. I will elaborate on this in another comment.

I think it's necessary, right? The goal of this is to trigger a purge even if there is no event, I believe.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -977,72 +1011,69 @@ public static boolean isJobStarted(String job, WorkflowContext workflowContext)
...
+      HelixDataAccessor accessor = manager.getHelixDataAccessor();
+      List<String> resourceConfigs =
+          accessor.getChildNames(accessor.keyBuilder().resourceConfigs());
+      if (resourceConfigs.size() > 0) {
+        RebalanceUtil.scheduleOnDemandPipeline(manager.getClusterName(), 0L);
+      } else {
+        LOG.warn("No resource config to trigger rebalance for clean up contexts for" + expiredJobs);
       }
     }
-    setNextJobPurgeTime(workflow, currentTime, purgeInterval, rebalanceScheduler, manager);
```
NealSun96, line 1037: This function is moved from here to `TaskGarbageCollectionStage`'s main thread. The reason for the move is that purgeInterval no longer needs to be passed to the async thread. I believe the trade-off is worth it because the frequency isn't critical. FYI: @dasahcc

NealSun96: That's a good point, but I think you may have gotten this backwards? `setNextJobPurgeTime` should be a part of the pipeline, so it makes more sense to be in the stage instead of the function, right? If there is ever a REST endpoint that triggers the function, we don't want to schedule a rebalance.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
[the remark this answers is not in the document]
NealSun96: Ok, renaming.

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
```diff
@@ -23,34 +29,87 @@ public AsyncWorkerType getAsyncWorkerType() {
...
       if (workflowConfig != null && (!workflowConfig.isTerminable() || workflowConfig
           .isJobQueue())) {
-        try {
-          TaskUtil.purgeExpiredJobs(workflowConfig.getWorkflowId(), workflowConfig,
-              dataProvider.getWorkflowContext(workflowConfig.getWorkflowId()), manager,
-              _rebalanceScheduler);
-        } catch (Exception e) {
-          LOG.warn(String.format("Failed to purge job for workflow %s with reason %s",
-              workflowConfig.getWorkflowId(), e.toString()));
+        WorkflowContext workflowContext = dataProvider.getWorkflowContext(entry.getKey());
```
alirezazamani, line 52: Can we have workflow config without having context? Do we need to do null check?

alirezazamani:
Overall looks good to me. Just to make sure, can you rebase with master branch, build the project and run the tests again and report the result?

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
NealSun96: I don't believe it's necessary as `entry` come from `dataProvider.getContexts().entrySet()`

alirezazamani: Ok makes sense.

NealSun96:
@alirezazamani Done so and the results have been updated.

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
NealSun96: Resolved offline.

NealSun96:
Final commit message:
## Recover Workflow Garbage Collection Logic ##
Added back workflow garbage collection logic; changed garbage collection stage to calculate jobs/workflows to purge in main thread, and do garbage collection in the async thread.

PR #1077, opened 2020-06-09
Add close method

mgao0:
### Issues

Currently there is no close method to close the lock and the underlying zk client connection. It could result to zk client leak.

### Description

Added a close() method to the interface and implementation of the lock, where the zk client will be closed.

### Tests

[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  54.480 s
[INFO] Finished at: 2020-06-09T11:20:35-07:00

### Commits

### Code Quality

- [ ] My diff has been formatted using helix-style.xml 


helix-lock/src/main/java/org/apache/helix/lock/DistributedLock.java
```diff
@@ -49,4 +49,9 @@
    * false if the user is not the lock owner or the lock doesn't have a owner
    */
   boolean isCurrentOwner();
+
+  /**
+   * Close the lock and release resources
+   */
+  void close();
```
jiajunwang, line 56: I suggest making this interface AutoClosable, then you implement close() naturally.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -115,6 +113,11 @@ public boolean isCurrentOwner() {
         .getTimeout());
   }
 
+  @Override
+  public void close() {
+    _baseDataAccessor.close();
```
jiajunwang, line 118: You don't call unlock() here?

helix-lock/src/test/java/org/apache/helix/lock/helix/TestZKHelixNonblockingLock.java
```diff
@@ -71,9 +73,14 @@ public void beforeMethod() {
     Assert.assertFalse(_gZkClient.exists(_lockPath));
   }
 
+  @AfterSuite
+  public void afterSuite() throws IOException {
+    _lock.close();
```
jiajunwang, line 78: Add a test case about close() to validate:
1. lock released properly. The releases information should be right (if we have any).
2. the connection has been ended.

helix-lock/src/main/java/org/apache/helix/lock/DistributedLock.java
zhangmeng916, line 56: So we would expect lock users to explicitly call this one? This is a bit unintuitive, when put together with the other APIs that are intended for lock purpose only.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -54,8 +53,8 @@
    * @param lockMsg the reason for having this lock
```
zhangmeng916, line 53: Can you fix the typo "lcok" in this file?

helix-lock/src/main/java/org/apache/helix/lock/DistributedLock.java
mgao0: As discussed offline, to avoid unexpected close of the lock, we will not implement AutoClosable interface for now.

mgao0: Discussed offline, for now we will let user to explicitly call the close method, since we cannot determine all the use cases where we should close the lock right now. After we have a clearer picture of all the features, we can try to do this for our users.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
mgao0: Discussed offline. Won't unlock the lock inside close(), but will check to make sure the lock is unlocked when the user calls close().

helix-lock/src/test/java/org/apache/helix/lock/helix/TestZKHelixNonblockingLock.java
mgao0: I added two tests. One is trying to close the lock when it's unlocked, and one is trying to close it when it's locked.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
mgao0: Updated. Thanks for checking!

mgao0:
Final commit message: Add a close method to Helix lock interface and current implementation to avoid ZK connection leakage

Issue #1097, opened 2020-06-16
The instance may not be assigned with the replica as expected with some customized state model


jiajunwang:
A regression was introduced by PR #986.

The PR tried to prioritize the preference list to avoid unnecessary top state transitions. However, there was a bug in the prioritizing logic and if one participant is skipped due to low priority, it won't be picked up again during the calculating. As a result, this participant won't be assigned with any replica even it is originally in the preference list.

This only happens if the state model has been customized so it is multiple top states and there is an intermediate state with expected count -1 between the top state and the other state.

PR #1098, opened 2020-06-16
Fix the issue that the instance may not be assigned a replica as expected.

jiajunwang:
### Issues

#1097 

### Description

This is to fix a regression that was introduced by PR #986 
The PR tried to prioritize the preference list to avoid unnecessary top state transitions. However, there was a bug in the prioritizing logic and if one participant is skipped due to low priority, it won't be picked up again during the calculating. As a result, this participant won't be assigned with any replica even it is originally in the preference list.
This only happens if the state model has been customized so it is multiple top states and there is an intermediate state with expected count -1 between the top state and the other states.

This fix will ensure the skipped participant being checked again until it gets the assignment.

### Tests

TestAbstractRebalancer.java
Added test data that simulate a customized state model fits the problem statement.

[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestClusterVerifier.testResourceSubset:225 expected:<false> but was:<true>
[INFO] 
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time: 01:25 h
[INFO] Finished at: 2020-06-17T18:59:29-07:00

Rerun

[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 45.908 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 52.290 s
[INFO] Finished at: 2020-06-18T11:40:16-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
@@ -428,9 +428,9 @@ public abstract class AbstractRebalancer<T extends BaseControllerDataProvider> i
         // Note that in either case, the proposed instance is considered to be assigned with a state
         // by now.
         if (!assignedInstances.add(proposedInstance)) {
-          // Sanity check to ensure there is no unexpected double assignment.
-          throw new HelixException(String
-              .format("The proposed instance %s has already been assigned before.", proposedInstance));
+          throw new AssertionError(String
+              .format("The proposed instance %s has been already assigned before.",
+                  proposedInstance));
         }
       }
     }
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -364,17 +365,20 @@ public abstract class AbstractRebalancer<T extends BaseControllerDataProvider> i
     liveAndEnabled.removeAll(disabledInstancesForPartition);
 
     // Sort the instances based on replicas' state priority in the current state
-    List<String> sortedPreferenceList = new ArrayList<>(preferenceList);
-    sortedPreferenceList.sort(new StatePriorityComparator(currentStateMap, stateModelDef));
+    List<String> currentStatePrioritizedList = new ArrayList<>(preferenceList);
+    currentStatePrioritizedList.retainAll(liveAndEnabled);
+    currentStatePrioritizedList.sort(new StatePriorityComparator(currentStateMap, stateModelDef));
+    Iterator<String> currentStatePrioritizedInstanceIter = currentStatePrioritizedList.iterator();
 
     // Assign the state to the instances that appear in the preference list.
     for (String state : statesPriorityList) {
       int stateCount =
           getStateCount(state, stateModelDef, liveAndEnabled.size(), preferenceList.size());
-      for (String instance : preferenceList) {
+      for (int i = 0; i < preferenceList.size(); i++) {
         if (stateCount <= 0) {
           break; // continue assigning for the next state
         }
+        String instance = preferenceList.get(i);
         if (assigned.contains(instance) || !liveAndEnabled.contains(instance)) {
           continue; // continue checking for the next available instance
         }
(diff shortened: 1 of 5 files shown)


helix-core/src/test/java/org/apache/helix/controller/rebalancer/TestAbstractRebalancer.java
```diff
@@ -56,11 +56,11 @@ public void testComputeBestPossibleState(String comment, String stateModelName,
             new IdealState("test"), new ClusterConfig("TestCluster"), partition,
             MonitoredAbnormalResolver.DUMMY_STATE_RESOLVER);
 
-    Assert.assertEquals(bestPossibleMap, expectedBestPossibleMap);
+    Assert.assertTrue(bestPossibleMap.equals(expectedBestPossibleMap));
```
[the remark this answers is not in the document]
jiajunwang: This is an Assert bug. assertEquals does not work for mapping or list. In general, it is doing left join. We want inner join : )

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
```diff
@@ -385,10 +389,19 @@ public static int getStateCount(String state, StateModelDefinition stateModelDef
           // If the desired state is the top state, but the instance cannot be transited to the
           // top state in one hop, try to keep the top state on current host or a host with a closer
           // state.
```
dasahcc, line 391: We can also additional check for single top state since multi top state even does not require these operations

jiajunwang: I thought about it, but if we have a state model which requires 2 top states, and replica is 4, do we still want this improvement? My answer is yes. I don't think we want to lose the universality.
In addition, it may not really prevent all the issues. For example, if the replica is 1, without a complete fix, I guess there will still be issues. So we should fix it and fix it for all cases.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
```diff
@@ -385,10 +389,19 @@ public static int getStateCount(String state, StateModelDefinition stateModelDef
...
-          for (String currentStatePrioritizedInstance : sortedPreferenceList) {
-            if (!assigned.contains(currentStatePrioritizedInstance) && liveAndEnabled
-                .contains(currentStatePrioritizedInstance)) {
+          while (currentStatePrioritizedInstanceIter.hasNext()) {
+            // Note that it is safe to check the prioritized instance items only once here.
+            // Since the only possible condition when we don't use an instance in this list is that
+            // it has been assigned with some state. And this is not revertable in this method. So
+            // checking it one more time later will only waste time.
+            String currentStatePrioritizedInstance = currentStatePrioritizedInstanceIter.next();
```
dasahcc, line 397: Maybe I am wrong. But could that be a scenario of:
preference list = 1, 2, 3
current state = 4, 5, 6

Then we keep assign ONLINE to 4,5,6. There will not be any movement. For DelayedAutoRebalance we may dont have it. But for default AutoRebalancer Or SemiAuto, we may use this for AbstractRebalancer then there will not be any movement.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
```diff
@@ -385,10 +389,19 @@ public static int getStateCount(String state, StateModelDefinition stateModelDef
...
+            if (!assigned.contains(currentStatePrioritizedInstance)) {
```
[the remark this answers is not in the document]
jiajunwang: I moved this logic above. Inactive nodes have been removed from the currentStatePrioritizedInstanceList. This shortens the instance list and saves some computing.

Note that we may don't want to do it for the preferencelist since it is not deepcopy.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
```diff
@@ -385,10 +389,19 @@ public static int getStateCount(String state, StateModelDefinition stateModelDef
...
               proposedInstance = currentStatePrioritizedInstance;
+              // If we find a different instance for the partition placement, then we need to check
+              // the same instance again or it will not be assigned with any partitions.
+              if (!proposedInstance.equals(instance)) {
+                i--;
```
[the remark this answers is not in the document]
jiajunwang: This is a trade-off. 
I agree queue would be much simpler. And with the queue, I can do more optimization.
However, since this calculation will be run many times. Creating a queue here might prolong the rebalance time a lot. Let me see if I can profile this.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
jiajunwang: currentStatePrioritizedInstance != current state instances.
In your case, the currentStatePrioritizedInstance will be 1,2,3.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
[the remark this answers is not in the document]
jiajunwang: I tried it, with some enhancement, the performance is actually slightly better.
The only downside is that we introduced many more changes to make it clearer. Please review the new change.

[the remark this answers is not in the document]
jiajunwang: One thing to note here, offer does not work. Adding the item to the back will dramatically change the preference list order, that is not what we want.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/AbstractRebalancer.java
```diff
@@ -352,60 +356,137 @@ public static int getStateCount(String state, StateModelDefinition stateModelDef
...
           bestPossibleStateMap.put(proposedInstance, state);
           stateCount--;
         }
-        assigned.add(proposedInstance);
+        // Note that in either case, the proposed instance is considered to be assigned with a state
+        // by now.
+        if (!assignedInstances.add(proposedInstance)) {
+          throw new AssertionError(String
+              .format("The proposed instance %s has been already assigned before.",
+                  proposedInstance));
```
dasahcc, line 433: This should not happen, right. You do the assignedInstances both for peek of the queue and after adjust instance.

jiajunwang: Yeah, this is for the sanity check. If something goes wrong, I hope it fails early. This could be very hard to debug. And if we have some other customized SM defs, it might not be covered by the unit test. Adding this assertion will help us identify issues much easier.

jiajunwang:

jiajunwang:
Accidentally hit "revert". Will discard the reverting branch. The Master is fine.

PR #1104, opened 2020-06-18
Enhance flaky test TestRemoveUserCbHandlerOnPathRemoval

### LINKED ISSUE #1102: Enhance TestZkCallbackHandlerLeak
testRemoveUserCbHandlerOnPathRemoval has a flaky test

Assert
        .assertEquals(watchPaths.get("existWatches").size(), 2,
            "Should have 2 exist-watches: CURRENTSTATE/{oldSessionId} and CURRENTSTATE/{oldSessionId}/TestDB0");

kaisun2000:
### Issues

Fixes #1102 
TestRemoveUserCbHandlerOnPathRemoval has a flaky test

### Description

`Assert
.assertEquals(watchPaths.get("existWatches").size(), 2,
"Should have 2 exist-watches: CURRENTSTATE/{oldSessionId} and CURRENTSTATE/{oldSessionId}/TestDB0");`
is flaky. 

### Tests

This test case pass. This pull only touch test case logic.

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/integration/TestZkCallbackHandlerLeak.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/TestZkCallbackHandlerLeak.java
@@ -605,6 +605,17 @@ public class TestZkCallbackHandlerLeak extends ZkUnitTestBase {
         "Should have 1 data-watches: MESSAGES");
     Assert.assertEquals(watchPaths.get("childWatches").size(), 1,
         "Should have 1 child-watches: MESSAGES");
+
+    // In this test participant0 also register to its own cocurrent state with a callbackhandler
+    // When session expiration happens, the current state parent path would also changes. However,
+    // an exists watch would still be installed by event pushed to ZkCLient event thread by
+    // fireAllEvent() children even path on behalf of old session callbackhandler. By the time this
+    // event gets invoked, the old session callbackhandler was removed, but the event would still
+    // install a exist watch for old session.
+    // The closest similar case in production is that router/controller has an session expiration at
+    // the same time as participant.
+    // Currently there are many places to register watch in Zookeeper over the evolution of Helix
+    // and ZkClient. We plan for further simplify the logic of watch installation next.
     result = TestHelper.verify(()-> {
       Map<String, List<String>> wPaths = ZkTestHelper.getZkWatch(participantToExpire.getZkClient());
       return wPaths.get("existWatches").size() == 1;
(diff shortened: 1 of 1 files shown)


helix-core/src/test/java/org/apache/helix/integration/TestZkCallbackHandlerLeak.java
```diff
@@ -605,9 +605,12 @@ public void onStateChange(String instanceName, List<CurrentState> statesInfo,
...
         "Should have 1 child-watches: MESSAGES");
-    Assert
-        .assertEquals(watchPaths.get("existWatches").size(), 2,
-            "Should have 2 exist-watches: CURRENTSTATE/{oldSessionId} and CURRENTSTATE/{oldSessionId}/TestDB0");
+    result = TestHelper.verify(()-> {
+      Map<String, List<String>> wPaths = ZkTestHelper.getZkWatch(participantToExpire.getZkClient());
+      return wPaths.get("existWatches").size() == 1;
+    }, 10000);
+    Assert.assertTrue(result,
+        "Should have 1 exist-watches: CURRENTSTATE/{oldSessionId}");
```
jiajunwang, line 613: Please correct me if I am wrong.
When the participant handles the new session, this folder of the old session will be removed, right? With the new change, should the watcher be removed as well when the folder being deleted?

kaisun2000: This test is confusing. Note here it is a participant0, register and currentstate/{sessionid} callbackhandler of itself. 

This does not normally happen in production,

The only case this can happen is that controller have CB for participant 1. At the same time controller and participant 1 both have session expiration. 

Back to the question, yes, the old session will be removed from participant point of view. But from router (who subscribe to callbackhandler), each existing callbackhandler of router would be called via lambda from zkclient event queue inserted by fireAllEvent(). 

` fireChildChangedEvents(entry.getKey(), entry.getValue(), true);`

The lambda would install this watch. 

Note, this is normally not a problem as we don't have such usage case. Only if both participant and controller/router have session expiration at the same time. 

Later, we plan to make it clear the responsibility of who should install watcher. We can simplify the logic here.

jiajunwang: Make sense. Thanks for the explanation. Could you please add the critical points of the explanation to the comment of the code? I think we will need to improve this further, and for sure the test case needs to be updated accordingly. Then with this additional info, it would easier to understand what shall we expect in this scenario later.

kaisun2000: Good point, added the comment.

kaisun2000:

    

> Fix flaky test testRemoveUserCbHandlerOnPathRemoval # #1102
>     
>     TestRemoveUserCbHandlerOnPathRemoval has a flaky assert. Addres
>     the assertion in this diff.

Issue #767, opened 2020-02-15
Assignment metadata missing for the WAGED rebalancer in stress test.


jiajunwang:
Please see the attachment, the baseline node is empty. I observed this multiple times in the stress test.
![image](https://user-images.githubusercontent.com/1803880/74579475-abb50780-4f4f-11ea-9c06-a718e327cc81.png)

I suspect this is about the GC thread.
1. ZkBucketDataAccessor.GC_THREAD should be created as a single thread pool instead of creating by newScheduledThreadPool() which could have more threads.
2. To ensure the order of the clean tasks, we'd better leverage a queue instead of relying on the TTL.

jiajunwang:
@narendly Could you please take a look at this issue? We will need to fix it before we can put Waged rebalancer into large scale usage.

NealSun96:
@jiajunwang Taking over this issue.

jiajunwang:
There is another concurrent modification issue found.

1. Regular GC task fails due to concurrent list modification and the stale versions are not removed at all.
2. If, by coincident, there is newer version in the list other then the current version, then because of the modification of the list inside the loop, the final element (the newer version) won't be filtered but being left in the to-be-removed list. Then the GC task removes the most recent version. For example,
  a) Input, current version "2"
  b) Children = [1, 2, 3]
  c) The task avoids checking "2", so the list for loop is: [1, 3]
  d) When check "1", it is removed from the list. So the list becomes [3]. Then the loop ends, because the first item has already been looped from the for iteration perspective.
  e) The version to be removed is "3"!

PR #1107, opened 2020-06-20
Fix ZkBucketDataAccessor failure due to concurrent modification.

jiajunwang:
### Issues

#767 

### Description

Concurrent modification causes two issues.
1. Regular GC task fails due to concurrent list modification and the stale versions are not removed at all.
2. If, by coincident, there is a newer version in the list other then the current version, then because of the modification of the list inside the loop, the final element (the newer version) won't be filtered but being left in the to-be-removed list. Then the GC task removes the most recent version. For example,
  a) Input, current version "2"
  b) Children = [1, 2, 3]
  c) The task avoids checking "2", so the list for loop is: [1, 3]
  d) When checking "1", it is removed from the list. So the list becomes [3]. Then the loop ends, because the first item has already been looped from the for iteration perspective.
  e) The version to be removed is "3"!

This PR fixes the issue by avoiding concurrent modifications. Also, it simplifies the logic so as to reduce the pending GC tasks.
The test is also updated accordingly.

### Tests

TestZkBucketDataAccessor

[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestJobQueueCleanUp.testJobQueueAutoCleanUp:111 Sets differ: expected [testJobQueueAutoCleanUp_JOB8, testJobQueueAutoCleanUp_JOB7, testJobQueueAutoCleanUp_JOB6, testJobQueueAutoCleanUp_JOB5, testJobQueueAutoCleanUp_JOB9] but got [testJobQueueAutoCleanUp_JOB4, testJobQueueAutoCleanUp_JOB8, testJobQueueAutoCleanUp_JOB7, testJobQueueAutoCleanUp_JOB6, testJobQueueAutoCleanUp_JOB5, testJobQueueAutoCleanUp_JOB9]
[INFO]
[ERROR] Tests run: 1145, Failures: 1, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time: 01:19 h
[INFO] Finished at: 2020-06-19T19:27:46-07:00

Rerun

[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 15.293 s - in org.apache.helix.integration.task.TestJobQueueCleanUp
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 21.621 s
[INFO] Finished at: 2020-06-19T21:53:23-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
@@ -25,9 +25,9 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.TimerTask;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
 import com.google.common.collect.ImmutableMap;
@@ -69,18 +69,19 @@ public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
       Executors.newSingleThreadScheduledExecutor();
 
   private final int _bucketSize;
-  private final long _versionTTL;
+  private final long _versionTTLms;
   private ZkSerializer _zkSerializer;
   private RealmAwareZkClient _zkClient;
   private ZkBaseDataAccessor<byte[]> _zkBaseDataAccessor;
+  private ScheduledFuture _gcTaskFuture = null;
 
   /**
    * Constructor that allows a custom bucket size.
    * @param zkAddr
    * @param bucketSize
-   * @param versionTTL in ms
+   * @param versionTTLms in ms
    */
-  public ZkBucketDataAccessor(String zkAddr, int bucketSize, long versionTTL) {
+  public ZkBucketDataAccessor(String zkAddr, int bucketSize, long versionTTLms) {
     if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED)) {
       try {
         // Create realm-aware ZkClient.
@@ -114,7 +115,7 @@ public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
     _zkBaseDataAccessor = new ZkBaseDataAccessor<>(_zkClient);
     _zkSerializer = new ZNRecordJacksonSerializer();
     _bucketSize = bucketSize;
-    _versionTTL = versionTTL;
+    _versionTTLms = versionTTLms;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBucketDataAccessor.java
```diff
@@ -126,8 +133,17 @@ public void testMultipleWrites() throws Exception {
...
     Assert.assertTrue(TestHelper.verify(() -> {
       List<String> children = _zkBaseDataAccessor.getChildNames(PATH, AccessOption.PERSISTENT);
-      return children.size() == 3;
-    }, 60 * 1000L));
+      return children.size() == 3 && children.containsAll(ImmutableList
+          .of(LAST_SUCCESSFUL_WRITE_KEY, LAST_WRITE_KEY,
+              new Long(lastSuccessfulWriteVer).toString()));
+    }, VERSION_TTL_MS * 2));
+
+    // Wait one more TTL to ensure that the
```
narendly, line 141: ensure that the...?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -329,32 +330,34 @@ public void close() {
...
   }
 
-  private void updateGCTimer(String rootPath, String currentVersion) {
-    TimerTask gcTask = new TimerTask() {
-      @Override
-      public void run() {
+  private synchronized void updateGCTimer(String rootPath, long currentVersion) {
+    if (_gcTaskFuture != null) {
+      _gcTaskFuture.cancel(false);
+    }
```
narendly, line 336: Do we really need to cancel? What if that causes incomplete deletion of stale versions? I think we should let it queue up?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBucketDataAccessor.java
jiajunwang: Good catch. I missed this part.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
jiajunwang: This cancel will only cancel the pending tasks. For the pending tasks that are queued up, they will only cause extra children list read. The later operation will be the same. And if there is risk of incomplete deleting, then that concern lives no matter we let it queued up or not. That issue (if exists) should be addressed separately.

jiajunwang:

PR #1109, opened 2020-06-22
ZkClient should not keep retrying getChildren() due to large number of children

### LINKED ISSUE #962: ZkClient keeps retrying connecting infinitely if getChildren's packet size exceeds 4MB
### Problem
For ZkClient's getChildren() operation, if there are a large number of children and the response packet size exceeds `jute.maxbuffer` default value 4MB on zk client side, ZkClient will get a `ConnectionLossException` and keep retrying connecting to ZK. The consequence is, the infinite retry may cause heavy GC on ZK server and kill ZK server.

### Related Logs
**zkCli to reproduce the exception**

```
Exception in thread "main" org.apache.zookeeper.KeeperException$ConnectionLossException: KeeperErrorCode = ConnectionLoss for /CLUSTER-1/CONFIGS/RESOURCE
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:99)
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:51)
	at org.apache.zookeeper.ZooKeeper.getChildren(ZooKeeper.java:1532)
	at org.apache.zookeeper.ZooKeeper.getChildren(ZooKeeper.java:1560)
	at org.apache.zookeeper.ZooKeeperMain.processZKCmd(ZooKeeperMain.java:731)
	at org.apache.zookeeper.ZooKeeperMain.processCmd(ZooKeeperMain.java:599)
	at org.apache.zookeeper.ZooKeeperMain.executeLine(ZooKeeperMain.java:371)
	at org.apache.zookeeper.ZooKeeperMain.run(ZooKeeperMain.java:331)
```

**In application logs**:
```
2020/03/03 17:55:19.979 WARN [ClientCnxn] [HelixTaskExecutor-message_handle_thread-SendThread(localhost:2181)] [helix] [] Session 0x270a12ba21102df for server localhost/127.0.0.1:2181, unexpected error, closing socket connection and attempting reconnect
java.io.IOException: Packet len4198500 is out of range!
        at org.apache.zookeeper.ClientCnxnSocket.readLength(ClientCnxnSocket.java:113)
        at org.apache.zookeeper.ClientCnxnSocketNIO.doIO(ClientCnxnSocketNIO.java:79)
        at org.apache.zookeeper.ClientCnxnSocketNIO.doTransport(ClientCnxnSocketNIO.java:366)
        at org.apache.zookeeper.ClientCnxn$SendThread.run(ClientCnxn.java:1145)

```

### Related Code in Helix ZkClient
https://github.com/apache/helix/blob/108dfc6c9b48a89b1c9804f0c2b77c2572d623a8/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java#L1449-L1481

### Potential Solutions

- Catch the particular error `Packet len4198500 is out of range` in ZkClient and stop retrying. However, the particular error is not thrown to ZkClient, instead, only a general `ConnectionLossException` is thrown, so native zookeeper code has to be changed to throw such error for ZkClient to catch.
- Add retry loop policy to ZkClient so retrying connecting could be limited by time or number of retries.
- Stat check numChildren on connection loss. If numChildren exceeds the limit, exit retry loop.
- ZK server should fail to create a child, if after the creation, packet length exceeds jute.maxbuffer for listing children.
- Returning children in pagination: https://issues.apache.org/jira/browse/ZOOKEEPER-2260
- More...

pkuwm:
### Issues

Fixes #962 

### Description

For ZkClient's getChildren() operation, if there are a large number of children and the response packet size exceeds jute.maxbuffer default value 4MB on zk client side, ZkClient will get a ConnectionLossException and keep retrying connecting to ZK. The consequence is, the infinite retry may cause heavy GC on ZK server and kill ZK server.

We could consider getting number of children in the path and check if it exceeds the limit. If it exceeds the children limit, there is no need to retry and then exist retry.

The default children limit is set for helix messages length using UUID. Packet length of 100K messages doesn't exceed 4 MB.

This PR is the first phase to fix the issue: a workaround to exit retry loop. Once we have a better way to solve the problem on ZK server side, we should remove this check.

### Tests

- testGetChildrenOnLargeNumChildren

```
zookeeper-api

[INFO] Tests run: 38, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 54.839 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 38, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  58.964 s
[INFO] Finished at: 2020-07-22T16:12:28-07:00

helix-core

[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestCustomizedViewAggregation.testCustomizedViewAggregation:389->validateAggregationSnapshot:233 expected:<true> but was:<false>
[ERROR]   TestRebalanceRunningTask.testFixedTargetTaskAndEnabledRebalanceAndNodeAdded:330 expected:<true> but was:<false>
[INFO]
[ERROR] Tests run: 1145, Failures: 3, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:23 h
[INFO] Finished at: 2020-07-22T17:39:02-07:00


[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 6.157 s - in org.apache.helix.integration.task.TestRebalanceRunningTask
[INFO]
[INFO] Results:
[INFO]

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 7.442 s - in org.apache.helix.integration.TestCustomizedViewAggregation
[INFO]
[INFO] Results:
[INFO]
[INFO]


[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 11.025 s - in org.apache.helix.integration.task.TestJobQueueCleanUp
[INFO]
[INFO] Results:
[INFO]
```

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -1017,7 +1017,7 @@ public class ZkClient implements Watcher {
             // TODO: remove this check once we have a better way to exit infinite retry
             ++connectionLossRetryCount;
             if (connectionLossRetryCount >= 3) {
-              checkNumChildren(path);
+              checkNumChildrenLimit(path);
               connectionLossRetryCount = 0;
             }
 
@@ -2254,8 +2254,8 @@ public class ZkClient implements Watcher {
     }
   }
 
-  private void checkNumChildren(String path) throws KeeperException, InterruptedException {
-    Stat stat = ((ZkConnection) getConnection()).getZookeeper().exists(path, false);
+  private void checkNumChildrenLimit(String path) throws KeeperException {
+    Stat stat = getStat(path);
     if (stat == null) {
       return;
     }
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -74,8 +74,9 @@ public class ZkClient implements Watcher {
 
   private static final long MAX_RECONNECT_INTERVAL_MS = 30000; // 30 seconds
 
-  // If number of children exceeds this limit,
-  // getChildren() should not retry on connection loss.
+  // If number of children exceeds this limit, getChildren() should not retry on connection loss.
+  // This is a workaround for exiting retry on connection loss because of large number of children.
+  //TODO: remove it once we have a better way to exit retry for this case
   private static final int NUM_CHILDREN_LIMIT;
 
   private final IZkConnection _connection;
@@ -1000,7 +1001,8 @@ public class ZkClient implements Watcher {
 
   protected List<String> getChildren(final String path, final boolean watch) {
     long startT = System.currentTimeMillis();
-    // Need one element array to change value of this final variable.
+    // This variable needs to be accessed in inner class Callable so it needs to be final.
+    // And need one element array to change value of this final variable.
     final int[] connectionLossRetryCount = {0};
     try {
       List<String> children = retryUntilConnected(new Callable<List<String>>() {
@@ -1021,15 +1023,22 @@ public class ZkClient implements Watcher {
               // This is a workaround to check numChildren to have a chance to exit retry loop.
               // TODO: remove this check once we have a better way to exit infinite retry
               Stat stat = getStat(path);
-              if (stat != null && stat.getNumChildren() > NUM_CHILDREN_LIMIT) {
-                LOG.error("Failed to get children for path {} because number of children {} "
-                        + "exceeds limit {}, aborting retry.", path, stat.getNumChildren(),
-                    NUM_CHILDREN_LIMIT);
-                // There is not an accurate KeeperException for the purpose.
-                // MarshallingErrorException could represent transport error,
-                // so use it to exit retry loop and tell that zk is not able to
-                // transport the data because packet length is too large.
-                throw new KeeperException.MarshallingErrorException();
+              if (stat != null) {
+                if (stat.getNumChildren() > NUM_CHILDREN_LIMIT) {
+                  LOG.error("Failed to get children for path {} because number of children {} "
+                          + "exceeds limit {}, aborting retry.", path, stat.getNumChildren(),
+                      NUM_CHILDREN_LIMIT);
+                  // There is not an accurate KeeperException for the purpose.
+                  // MarshallingErrorException could represent transport error,
+                  // so use it to exit retry loop and tell that zk is not able to
+                  // transport the data because packet length is too large.
+                  throw new KeeperException.MarshallingErrorException();
+                } else {
+                  // No need to do stat again for next connection loss.
+                  connectionLossRetryCount[0] = 0;
+                  LOG.info("Number of children {} is less than limit {}, not exiting retry.",
+                      stat.getNumChildren(), NUM_CHILDREN_LIMIT);
+                }
               }
             }
 
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -1007,23 +1007,22 @@ public class ZkClient implements Watcher {
           try {
             return getConnection().getChildren(path, watch);
           } catch (ConnectionLossException e) {
-            ++connectionLossRetryCount;
+            // Issue: https://github.com/apache/helix/issues/962
+            // Connection loss might be caused by an excessive number of children.
+            // Infinitely retrying connecting may cause high GC in ZK server and kill ZK server.
+            // This is a workaround to check numChildren to have a chance to exit retry loop.
             // Allow retrying 3 times before checking stat checking number of children,
             // because there is a higher possibility that connection loss is caused by other
-            // factors such as network connectivity, connected ZK node could not serve
-            // the request, session expired, etc.
+            // factors such as network connectivity, session expired, etc.
+            // TODO: remove this check once we have a better way to exit infinite retry
+            ++connectionLossRetryCount;
             if (connectionLossRetryCount >= 3) {
-              // Issue: https://github.com/apache/helix/issues/962
-              // Connection loss might be caused by an excessive number of children.
-              // Infinitely retrying connecting may cause high GC in ZK server and kill ZK server.
-              // This is a workaround to check numChildren to have a chance to exit retry loop.
-              // TODO: remove this check once we have a better way to exit infinite retry
-              Stat stat = getStat(path);
+              Stat stat = ((ZkConnection) getConnection()).getZookeeper().exists(path, false);
               if (stat != null) {
                 if (stat.getNumChildren() > NUM_CHILDREN_LIMIT) {
-                  LOG.error("Failed to get children for path {} because number of children {} "
-                          + "exceeds limit {}, aborting retry.", path, stat.getNumChildren(),
-                      NUM_CHILDREN_LIMIT);
+                  LOG.error("Failed to get children for path {} because of connection loss. "
+                          + "Number of children {} exceeds limit {}, aborting retry.", path,
+                      stat.getNumChildren(), NUM_CHILDREN_LIMIT);
                   // MarshallingErrorException could represent transport error: exceeding the
                   // Jute buffer size. So use it to exit retry loop and tell that zk is not able to
                   // transport the data because packet length is too large.
@@ -1555,14 +1554,12 @@ public class ZkClient implements Watcher {
         } catch (Exception e) {
           throw ExceptionUtil.convertToRuntimeException(e);
         }
-
         // before attempting a retry, check whether retry timeout has elapsed
         if (System.currentTimeMillis() - operationStartTime > _operationRetryTimeoutInMillis) {
           throw new ZkTimeoutException("Operation cannot be retried because of retry timeout ("
               + _operationRetryTimeoutInMillis + " milli seconds). Retry was caused by "
               + retryCauseCode);
         }
-        LOG.warn("Retrying operation, caused by {}", retryCauseCode);
       }
     } finally {
       if (_monitor != null) {
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -70,8 +70,13 @@ import org.slf4j.LoggerFactory;
  * WARN: Do not use this class directly, use {@link org.apache.helix.zookeeper.impl.client.ZkClient} instead.
  */
 public class ZkClient implements Watcher {
-  private static Logger LOG = LoggerFactory.getLogger(ZkClient.class);
-  private static long MAX_RECONNECT_INTERVAL_MS = 30000; // 30 seconds
+  private static final Logger LOG = LoggerFactory.getLogger(ZkClient.class);
+
+  private static final long MAX_RECONNECT_INTERVAL_MS = 30000; // 30 seconds
+
+  // If number of children exceeds this limit,
+  // getChildren() should not retry on connection loss.
+  private static final int NUM_CHILDREN_LIMIT;
 
   private final IZkConnection _connection;
   private final long _operationRetryTimeoutInMillis;
@@ -102,6 +107,17 @@ public class ZkClient implements Watcher {
   // To automatically retry the async operation, we need a separate thread other than the
   // ZkEventThread. Otherwise the retry request might block the normal event processing.
   protected final ZkAsyncRetryThread _asyncCallRetryThread;
+
+  static {
+    // 100K is specific for helix messages which use UUID.
+    // It makes messages packet length just below 4 MB.
+    // Tested and calculated: response packet length: 4 * 1024 * 1024 bytes.
+    // UUID string length: 36 bytes. Each child adds 4 bytes in packet.
+    // Estimate num children: (4 * 1024 * 1024) / (36 + 4) ~= 104.8K
+    // Set it here for unit test to use reflection to change value
+    // because compilers optimize constants by replacing them inline.
+    NUM_CHILDREN_LIMIT = 100 * 1000;
+  }
 
   private class IZkDataListenerEntry {
     final IZkDataListener _dataListener;
@@ -984,11 +1000,43 @@ public class ZkClient implements Watcher {
 
   protected List<String> getChildren(final String path, final boolean watch) {
     long startT = System.currentTimeMillis();
+    // Need one element array to change value of this final variable.
+    final int[] connectionLossRetryCount = {0};
     try {
       List<String> children = retryUntilConnected(new Callable<List<String>>() {
         @Override
         public List<String> call() throws Exception {
-          return getConnection().getChildren(path, watch);
+          try {
+            return getConnection().getChildren(path, watch);
+          } catch (ConnectionLossException e) {
+            ++connectionLossRetryCount[0];
+            // Allow retrying 3 times before checking stat checking number of children,
+            // because there is a higher possibility that connection loss is caused by other
+            // factors such as network connectivity, connected ZK node could not serve
+            // the request, session expired, etc.
+            if (connectionLossRetryCount[0] >= 3) {
+              // Issue: https://github.com/apache/helix/issues/962
+              // Connection loss might be caused by an excessive number of children.
+              // Infinitely retrying connecting may cause high GC in ZK server and kill ZK server.
+              // This is a workaround to check numChildren to have a chance to exit retry loop.
+              // TODO: remove this check once we have a better way to exit infinite retry
+              Stat stat = getStat(path);
+              if (stat != null && stat.getNumChildren() > NUM_CHILDREN_LIMIT) {
+                LOG.error("Failed to get children for path {} because number of children {} "
+                        + "exceeds limit {}, aborting retry.", path, stat.getNumChildren(),
+                    NUM_CHILDREN_LIMIT);
+                // There is not an accurate KeeperException for the purpose.
+                // MarshallingErrorException could represent transport error,
+                // so use it to exit retry loop and tell that zk is not able to
+                // transport the data because packet length is too large.
+                throw new KeeperException.MarshallingErrorException();
+              }
+            }
+
+            // Re-throw the ConnectionLossException for retryUntilConnected() to catch
+            // and retry the operation.
+            throw e;
+          }
         }
       });
       record(path, null, startT, ZkClientMonitor.AccessType.READ);
@@ -1503,12 +1551,14 @@ public class ZkClient implements Watcher {
         } catch (Exception e) {
           throw ExceptionUtil.convertToRuntimeException(e);
         }
+
         // before attempting a retry, check whether retry timeout has elapsed
         if (System.currentTimeMillis() - operationStartTime > _operationRetryTimeoutInMillis) {
           throw new ZkTimeoutException("Operation cannot be retried because of retry timeout ("
               + _operationRetryTimeoutInMillis + " milli seconds). Retry was caused by "
               + retryCauseCode);
         }
+        LOG.warn("Retrying operation, caused by {}", retryCauseCode);
       }
     } finally {
       if (_monitor != null) {
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -1011,9 +1011,9 @@ public class ZkClient implements Watcher {
             // Connection loss might be caused by an excessive number of children.
             // Infinitely retrying connecting may cause high GC in ZK server and kill ZK server.
             // This is a workaround to check numChildren to have a chance to exit retry loop.
-            // Allow retrying 3 times before checking stat checking number of children,
-            // because there is a higher possibility that connection loss is caused by other
-            // factors such as network connectivity, session expired, etc.
+            // Check numChildren stat every other 3 connection loss, because there is a higher
+            // possibility that connection loss is caused by other factors such as network
+            // connectivity, session expired, etc.
             // TODO: remove this check once we have a better way to exit infinite retry
             ++connectionLossRetryCount;
             if (connectionLossRetryCount >= 3) {
@@ -1028,7 +1028,6 @@ public class ZkClient implements Watcher {
                   // transport the data because packet length is too large.
                   throw new KeeperException.MarshallingErrorException();
                 } else {
-                  // No need to do stat again for next connection loss.
                   connectionLossRetryCount = 0;
                   LOG.info("Number of children {} is less than limit {}, not exiting retry.",
                       stat.getNumChildren(), NUM_CHILDREN_LIMIT);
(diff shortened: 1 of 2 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -103,6 +108,17 @@
   // ZkEventThread. Otherwise the retry request might block the normal event processing.
   protected final ZkAsyncRetryThread _asyncCallRetryThread;
 
+  static {
```
lei-xia, line 111: Please add a TODO item here to state this is just a workaround fix because zookeeper does not return the failure reason if it fails to read a large children list.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -984,11 +1000,43 @@ private void fireAllEvents() {
...
+              Stat stat = getStat(path);
+              if (stat != null && stat.getNumChildren() > NUM_CHILDREN_LIMIT) {
+                LOG.error("Failed to get children for path {} because number of children {} "
+                        + "exceeds limit {}, aborting retry.", path, stat.getNumChildren(),
+                    NUM_CHILDREN_LIMIT);
+                // There is not an accurate KeeperException for the purpose.
+                // MarshallingErrorException could represent transport error,
+                // so use it to exit retry loop and tell that zk is not able to
+                // transport the data because packet length is too large.
+                throw new KeeperException.MarshallingErrorException();
```
lei-xia, line 1032: This exception is confusing, we can inherit a new ZookeeperException here if needed.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -984,11 +1000,43 @@ private void fireAllEvents() {
 
   protected List<String> getChildren(final String path, final boolean watch) {
     long startT = System.currentTimeMillis();
+    // Need one element array to change value of this final variable.
+    final int[] connectionLossRetryCount = {0};
```
lei-xia, line 1004: ? this pattern looks very confusing,

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: I added a TODO at `getChildren()` so when we remove that code part at getChildren(), we will also remove this private LIMIT. To make it clear, I also add a TODO for `NUM_CHILDREN_LIMIT`.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: This variable needs to be accessed in inner class Callable so it needs to be final.
And since it is final, we could not increment the value if we make a an integer. So we need this one element array to change value of this final variable. I understand it looks a bit ugly, but since we are removing the code part in a few months once we have a better way in zk server, I wanted to make it simple without creating a wrapper class.

pkuwm:
> Suggest change "TODO: remove this check once we have a better way to exit infinite retry " to "TODO: remove this check once the new Zookeeper client containing children pagination (link to the ZK issue) is released".

@lei-xia Thanks for the suggestion. My thought is ZK might not have to add the feature for children pagination. But it may have other ways such as:
- Not allowing creating a child if it is created, packet length will exceed jute.maxbuffer when listing children
- Throw a different exception to indicate the error so zkclient could exit retry

So I don't want to add a specific solution which is not yet determined in this comment. Anyway we will keep it in mind and on our TODO list. What do you think?

pkuwm:
> why not just merge this with line 79? private static final int NUM_CHILDREN_LIMIT = xxx;

Purpose is for unit test. I explain it in comment: Set it here for unit test to use reflection to change value because compilers optimize constants by replacing them inline. If we don't do it like this, we would not be able to unit test getChildren() unless we create > 100K children.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
lei-xia: Can we use wrapped integer?

lei-xia: Also, connectionLossRetryCount seems only used within callable, why not define within the callable function?

pkuwm: Considering clean code, I think we could implement a GetChildrenCallable class that has the counter internally, so we could get rid of the one element array. It will look better

pkuwm: Resolved. I was thinking in the lambda way that we could not define the private member in callable class but actually we could when we new a callable class.. Kind of missing the old stuff.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: @lei-xia `KeeperException` is an abstract class and it is not flexible for us to extend to have a new exception code. Say if we create a new exception type 
```
PacketOutOfLenException extends KeeperException {
   PacketOutOfLenException(code) {
      super(code);
   }
}
```
If we use a different code other than existing defined code, KeeperException won't be able to recognize it and will throw `IllegalArgumentException(Invalid exception code)`. I don't think this what we need.

Comparing all the exception types in KeeperException, I think `MarshallingErrorException` is the most close one that represents the failure reason: failed at transport layer.

When ZkClient throws `MarshallingErrorException`, we'll have error log `"Failed to get children for path {} because number of children {} exceeds limit {}, aborting retry."` and know the error is because of large children listing.

What do you think?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -984,11 +997,51 @@ private void fireAllEvents() {
...
+              if (stat != null) {
+                if (stat.getNumChildren() > NUM_CHILDREN_LIMIT) {
+                  LOG.error("Failed to get children for path {} because number of children {} "
+                          + "exceeds limit {}, aborting retry.", path, stat.getNumChildren(),
+                      NUM_CHILDREN_LIMIT);
+                  // There is not an accurate KeeperException for the purpose.
+                  // MarshallingErrorException could represent transport error,
+                  // so use it to exit retry loop and tell that zk is not able to
+                  // transport the data because packet length is too large.
+                  throw new KeeperException.MarshallingErrorException();
```
jiajunwang, line 1031: BadArgumentsException looks better than this one.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -103,6 +109,13 @@
   // ZkEventThread. Otherwise the retry request might block the normal event processing.
   protected final ZkAsyncRetryThread _asyncCallRetryThread;
 
+  static {
```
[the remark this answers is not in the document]
pkuwm: I don't want to change the original init way for MAX_RECONNECT_INTERVAL_MS.
Like I comment, in terms of NUM_CHILDREN_LIMIT, the purpose is for unit test. I explain it in comment: Set it here for unit test to use reflection to change value because compilers optimize constants by replacing them inline. If we don't do it like this, we would not be able to unit test getChildren() unless we create > 100K children.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: I considered both. `BadArgumentsException` was the one I tried to use, but realized that `MarshallingErrorException` could represent a transport error and mostly mean data exceeds jute.maxbuffer. Here is an example: https://tgockel.github.io/zookeeper-cpp/classzk_1_1marshalling__error.html#details
`An error occurred while marshalling data. The most common cause of this is exceeding the Jute buffer size – meaning the transaction was too large`
So I decided to choose MarshallingError.  BadArgumentsException for me sounds more like the API is misused, eg, invalid path that has extra slash "/path/".

jiajunwang: Sounds good. So just mention this point instead of saying there is no accurate exception? It seems to be accurate.
Another option is to throw an HelixException, or some new zkclient exception that we invent. Either way works, I think.

pkuwm: OK. Let me just remove that line of comment to avoid confusion :)

Considering backward compatibility, if a `KeeperException` is thrown for the operation, it will be converted to `ZkException`. If user catches `ZkException` and handles it, there is extra impact, their app still runs.

If we throw a ZkException (we don't throw HelixException since we've separate zk module from helix), it will be converted to a RuntimeException, which may cause ZkClient users terminated if they don't catch RuntimeException. 

Or an option is: we change `retryUntilConnected` to also catch `ZkException`. With this change, we could throw `ZkException` with a message indicating "Large children error". I am not sure if we would like to change `retryUntilConnected` considering this is a workaround and later we remove the logic. What do you think about this option? @jiajunwang

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -984,11 +997,50 @@ private void fireAllEvents() {
...
       List<String> children = retryUntilConnected(new Callable<List<String>>() {
+        private int connectionLossRetryCount = 0;
+
         @Override
         public List<String> call() throws Exception {
-          return getConnection().getChildren(path, watch);
+          try {
+            return getConnection().getChildren(path, watch);
+          } catch (ConnectionLossException e) {
+            ++connectionLossRetryCount;
```
dasahcc, line 1010: Do we need this? If it is connection loss, @jiajunwang 's improvement for retry at zk event thread layer can handle that, right?

pkuwm: @dasahcc My purpose is to avoid stat check for each connection loss, because large children is just a very tiny possibility. And high possibility for network issue/session expiry. Then we could avoid the extra stat call in most of time.

// Allow retrying 3 times before checking stat checking number of
//  because there is a higher possibility that connection loss is caused by other
// factors such as network connectivity, connected ZK node could not serve
// the request, session expired, etc.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: `If user catches ZkException and handles it, there is extra impact, their app still runs.`

Do you mean there is **no extra impact?**

Overall, I am a little bit confused. I think ZkException is what we (IOItech) convert from KeeperException to. Upper layer sees ZkException, they don't deal with KeeperException (which is from Zookeeper client).

In this case, as long as ZkClient event thread can catch this ZkException and don't terminate, that is what we want, right?

Am I missing something?

pkuwm: @kaisun2000 Yeah you caught it. I meant there is **no** extra impact. Sorry for the confusion.

Yes, as I mentioned, there are two options:
1.  throw `KeeperException.MarshallingErrorException`. This will be converted by retryUntilConnected to ZkException. Benefit is we don't change retryUntilConnected.
2. throw ZkException. Benefit is we could add message in the exception. But we have to change retryUntilConnected to catch ZkException and rethrow it. Considering this is a workaround and later we remove the logic, I use `KeeperException.MarshallingErrorException` here. What do you think? @kaisun2000

kaisun2000: As discussed offline. Here is the basic principle inherited from IOITech
this layer, we see KeeperException,
user see ZkException,
retryUntilConnected, throw ZkException and use see ZkException. 

Here is an idea:
So maybe we can just add another type of ZkException say TooManyChildrenZkException and convert this from retryUntilConnect to this new ZkException. User of ZkClient can handle this TooManyChidrenZkException with whatever logic they want.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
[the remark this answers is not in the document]
pkuwm: @jiajunwang No, the test will not work, because compilers optimize it inline:
```
getChildren() {
  if (stat.getNumChildren() > 100 * 1000) { }
}
```

We would like to avoid that and keep the variable. So the static block is used.

In the version 3.4.13 of zk we use, the `packetLen` also declares in this way.
```
public static final int packetLen;

  static {
    if (LOG.isDebugEnabled()) {
      LOG.debug("zookeeper.disableAutoWatchReset is " + disableAutoWatchReset);
    }

    packetLen = Integer.getInteger("jute.maxbuffer", 4194304);
  }
```

FYI, `packetLen` is removed in newer version of ZK.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: Adding @jiajunwang for extra opinions.
I understand that and I also thought about it. My take is since this is only a workaround not for long term, I would like to just make it simple.
If later we would like to make it nicer because ZK gives a different error code, we could have a sub ZkException like `ZkMarshallingError` (we already have this exception defined).

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -799,4 +802,66 @@ public void testAsyncWriteOperations() {
...
     }
   }
+
+  /*
+   * Tests getChildren() when there are an excessive number of children and connection loss happens,
+   * the operation should terminate and exit retry loop.
+   */
+  @Test
+  public void testGetChildrenOnLargeNumChildren() throws Exception {
+    // Default packetLen is 4M. It is static final and initialized
```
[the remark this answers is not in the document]
pkuwm: I don't think it is necessary, because anyway it checks jute.maxbuffer. What you mean 10K children nodes are for 4 MB, it is still defined by jute.maxbuffer. We adjust this value to smaller one for testing so only 5 children are needed. Either 100K or 5, it is behaving the same, why would we bother to create 10K children, right? :)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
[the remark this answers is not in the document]
jiajunwang: Thanks for the explanation.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1503,12 +1555,14 @@ private void acquireEventLock() {
...
           throw ExceptionUtil.convertToRuntimeException(e);
         }
+
         // before attempting a retry, check whether retry timeout has elapsed
         if (System.currentTimeMillis() - operationStartTime > _operationRetryTimeoutInMillis) {
           throw new ZkTimeoutException("Operation cannot be retried because of retry timeout ("
               + _operationRetryTimeoutInMillis + " milli seconds). Retry was caused by "
               + retryCauseCode);
         }
+        LOG.warn("Retrying operation, caused by {}", retryCauseCode);
```
jiajunwang, line 1565: I think we removed this because many clients complaining that this retry log floods their log file. Please make it debug or don't add it.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -984,11 +997,50 @@ private void fireAllEvents() {
...
+            if (connectionLossRetryCount >= 3) {
+              // Issue: https://github.com/apache/helix/issues/962
+              // Connection loss might be caused by an excessive number of children.
+              // Infinitely retrying connecting may cause high GC in ZK server and kill ZK server.
+              // This is a workaround to check numChildren to have a chance to exit retry loop.
+              // TODO: remove this check once we have a better way to exit infinite retry
+              Stat stat = getStat(path);
+              if (stat != null) {
+                if (stat.getNumChildren() > NUM_CHILDREN_LIMIT) {
+                  LOG.error("Failed to get children for path {} because number of children {} "
```
jiajunwang, line 1024: Let's make the comment clear.
"Failed to get children for path {} because of connection loss. Stop retrying because the number of children exceeds......"

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -984,11 +997,50 @@ private void fireAllEvents() {
...
+                          + "exceeds limit {}, aborting retry.", path, stat.getNumChildren(),
+                      NUM_CHILDREN_LIMIT);
+                  // MarshallingErrorException could represent transport error: exceeding the
+                  // Jute buffer size. So use it to exit retry loop and tell that zk is not able to
+                  // transport the data because packet length is too large.
+                  throw new KeeperException.MarshallingErrorException();
+                } else {
+                  // No need to do stat again for next connection loss.
+                  connectionLossRetryCount = 0;
```
jiajunwang, line 1033: I think you might want to keep checking the children's number on any retries after the first 3. Or, what if the children are increased during the retry, right?

So just make the logic like,
if (count < 3 || getChildrenCount() < limit) {
  count++;
  retry;
} else {
  exist;
}

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -984,11 +997,50 @@ private void fireAllEvents() {
...
+            // because there is a higher possibility that connection loss is caused by other
+            // factors such as network connectivity, connected ZK node could not serve
+            // the request, session expired, etc.
+            if (connectionLossRetryCount >= 3) {
+              // Issue: https://github.com/apache/helix/issues/962
+              // Connection loss might be caused by an excessive number of children.
+              // Infinitely retrying connecting may cause high GC in ZK server and kill ZK server.
+              // This is a workaround to check numChildren to have a chance to exit retry loop.
+              // TODO: remove this check once we have a better way to exit infinite retry
+              Stat stat = getStat(path);
```
jiajunwang, line 1021: This is a recursive retryUntilConnect call inside. I think it shall work. But it would be better if we can avoid this recursive call. Basically what you need to do is,
1. wait until connected.
2. call the native exits() call for the state.

The benefit is mainly avoiding potential issues if we have retryuntilconnect inside another retry.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: Done.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: I also thought about it. It would have nested retryUntilConnected. The benefit is it would retry to get the stat if there is connection loss for getStat(). If we use native exists() and there is connection loss, it would not retry and instead go back to getChildren() again.
The potential issue could be the retry timeout, because new retryUntilConnected() has a new start time. So getState() would potentially retry longer than the configured retry timeout for getChildren().
I don't have a strong preference. Seems you prefer native exits(), I'll change it.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: If you change it to debug, we shall be able to see it during the debug practice (by changing the log4j attribute). So Let's do debug for now.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: > Your logic here is only allow retrying 3 times, after 3 retries, exit retry loop.

Could you please check my logic carefully. It will keep retrying if the ChildrenCount is good. It is "||", not "&&". It is doing almost the same as your logic. getChildrenCount() won't be called if the count is smaller than 3.
This is helping you to simplify the code.

pkuwm: > I think you might want to keep checking the children's number on any retries after the first 3. Or, what if the children are increased during the retry, right?

Yes, it will still check numChildren after the first 3 retries. The strategy is check numChildren every other 3 retries.
Your code checks numChildren every time after the first 3 retries, right? But actually I don't think it is necessary to check every time after 3 retries. I would like to just check numChildren every other 3 retries. So maybe your code could not really simplify the logic. I'll sync up with you.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2216,4 +2253,25 @@ private void validateCurrentThread() {
       throw new IllegalArgumentException("Must not be done in the zookeeper event thread.");
     }
   }
+
+  private void checkNumChildren(String path) throws KeeperException, InterruptedException {
+    Stat stat = ((ZkConnection) getConnection()).getZookeeper().exists(path, false);
```
jiajunwang, line 2258: I just noticed this issue. So you only check on connection loss exception. What if when you call this exists, the connection has not been recovered yet? This is very possible since there is no other operation or waits between the error and this exists() call.

In this case, the check will throw an exception, and the retry until connect call will also return unexpectedly.

I think we need to wait until connected here.

pkuwm: @jiajunwang Good catch! I used to call `getStat()` considering reconnect. By somehow I changed it to `exists()` which is not supposed to. I believe `getStat()` is what we need: wait until connected and monitoring. Regarding nested retry until connected, it is fine because if getStat keeps getting connection loss and retrying, even we don't use getStat(), it'll be retrying as well in getChildren. 
Reverted to `getStat()`

pkuwm:
```
For ZkClient's getChildren() operation, if there are a large number of children and the response packet size exceeds jute.maxbuffer default value 4MB on zk client side, ZkClient will get a ConnectionLossException and keep retrying connecting to ZK. The consequence is, the infinite retry may cause heavy GC on ZK server and kill ZK server.

This commit implements a workaround to exit retry loop for getChildren() if a large number of children cause connection loss. 
```

PR #1112, opened 2020-06-22
Merge watcher branch to master

### LINKED ISSUE #1034: Fix leaking CallbackHandler and leaking Zk path in the Zookeeper server
Recently we identified two class of leaking Zk path in Zookeeper server

1. Leaking current state CallbackHandler and subsequently if the leaked CallbackHandler current state instance session timeout, the Zk path of old current state parent path  would be leaked on Zookeeper server. This would also result in "dangling" CallbackHandler which would leak memory and issue additional zookeeper read not necessary.

2. Leaking current state znode path. This is due to deletion of the current state znode is async to installing watcher in various thread in ZkClient, callbackhandler etc. 

The current short term fix would address these two issues.

There are couple of other optimization that can be done in term of zookeeper access. We will put them as long term, and address them separately.

kaisun2000:
This is to fix #1034
Current watcher short term fix is completed. Please merge it to apache/master


jiajunwang:
This does not look right. Have you rebased to the master in your branch before creating this PR? Hint, you might need to resolve some conflict and clean up the branch so it is a clean merge PR.

kaisun2000:
I thought you would do the merge to master from apache/watcher branch. If I rebase apache/watcher to apache/master, we need to review it quickly or master may move again.

kaisun2000:
Created pull request https://github.com/apache/helix/pull/1115 as I can't change the comparison branches here. Closing this one. Can you have a look at 1115?

Issue #1120, opened 2020-06-24
Need to clean up the WAGED persisted assignment state when no resources are on WAGED


jiajunwang:
During the migration test, we find an issue that the WAGED persisted assignment state is not cleaned up even when the resources are configured not to use the WAGED rebalancer. This causes a potential issue that if the cluster is migrated to WAGED and rolled back, then migrated again, the persisted state will impact the assignment so the first migration and second migration will have very different results.

And the side effect is that the assignment persistent state will never be cleaned up, which is a potential waste of resources.

PR #1123, opened 2020-06-26
Cleanup the persisted assignment state if no resource is on WAGED rebalancer.

jiajunwang:
### Issues

#1120

### Description

This is to prevent the WAGED rebalancer reads stale assignment records from the previous rebalance pipeline.
For example,
1. Resource A was the only resource. And it is rebalanced by WAGED, then we have a persisted assignment for A.
2. Resource A was reconfigured to using DelayedRebalancer, then we stop the WAGED rebalancer since there is no more resource using WAGED. So the persisted records are still in ZK.
3. Resource A is recreated and using WAGED again. In this case, the previous persisted assignment is no longer valid. We should treat A as a brand new resource instead of considering the stale assignment record.

Moreover, this change will help to clean up the ZK persisted data if no resource is using WAGED.

### Tests

TestAssignmentMetadataStore, TestWagedRebalancer

[INFO] Results:
[INFO] 
[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 1149, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time: 01:25 h
[INFO] Finished at: 2020-06-26T00:00:39-07:00

Rerun

[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 14.508 s - in org.apache.helix.integration.task.TestJobQueueCleanUp
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 23.728 s
[INFO] Finished at: 2020-06-26T00:01:34-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -27,10 +27,10 @@ import java.util.Map;
 import org.apache.helix.BucketDataAccessor;
 import org.apache.helix.HelixException;
 import org.apache.helix.HelixProperty;
-import org.apache.helix.zookeeper.datamodel.ZNRecord;
-import org.apache.helix.manager.zk.ZNRecordJacksonSerializer;
 import org.apache.helix.manager.zk.ZkBucketDataAccessor;
 import org.apache.helix.model.ResourceAssignment;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.apache.helix.zookeeper.datamodel.serializer.ZNRecordJacksonSerializer;
 import org.apache.helix.zookeeper.zkclient.exception.ZkNoNodeException;
 import org.apache.helix.zookeeper.zkclient.serialize.ZkSerializer;
 
@@ -70,7 +70,7 @@ public class AssignmentMetadataStore {
         _globalBaseline = splitAssignments(baseline);
       } catch (ZkNoNodeException ex) {
         // Metadata does not exist, so return an empty map
-        _globalBaseline = Collections.emptyMap();
+        _globalBaseline = new HashMap<>();
       }
     }
     return _globalBaseline;
@@ -85,7 +85,7 @@ public class AssignmentMetadataStore {
         _bestPossibleAssignment = splitAssignments(baseline);
       } catch (ZkNoNodeException ex) {
         // Metadata does not exist, so return an empty map
-        _bestPossibleAssignment = Collections.emptyMap();
+        _bestPossibleAssignment = new HashMap<>();
       }
     }
     return _bestPossibleAssignment;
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -70,7 +70,7 @@ protected AssignmentMetadataStore(BucketDataAccessor bucketDataAccessor, String
         _globalBaseline = splitAssignments(baseline);
       } catch (ZkNoNodeException ex) {
         // Metadata does not exist, so return an empty map
-        _globalBaseline = Collections.emptyMap();
+        _globalBaseline = new HashMap<>();
```
pkuwm, line 73: New elements need to add to this empty map?

jiajunwang: Yes

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/TestAssignmentMetadataStore.java
```diff
@@ -28,74 +29,50 @@
...
+        new ZkBucketDataAccessor(_manager.getMetadataStoreConnectionString(), DEFAULT_BUCKET_SIZE,
+            Integer.MAX_VALUE), _manager.getClusterName());
   }
 
   @AfterClass
   public void afterClass() {
     if (_store != null) {
       _store.close();
     }
+    _baseAccessor.remove("/" + CLUSTER_NAME + "/ASSIGNMENT_METADATA", AccessOption.PERSISTENT);
```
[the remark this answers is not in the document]
jiajunwang: Yeah, I think we can just remove the cluster.

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/TestAssignmentMetadataStore.java
```diff
@@ -107,55 +84,103 @@ public void afterClass() {
...
+    Assert.assertTrue(bestPossibleVersionCount > 0);
+
+    _store.clearAssignmentMetadata();
+
+    // 1. cache is cleaned up
+    Assert.assertEquals(_store._bestPossibleAssignment, Collections.emptyMap());
+    Assert.assertEquals(_store._globalBaseline, Collections.emptyMap());
+    // 2. refresh the cache and then read from ZK again to ensure the persisted assignments is empty
+    _store.reset();
+    Assert.assertEquals(_store.getBaseline(), Collections.emptyMap());
```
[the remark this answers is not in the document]
jiajunwang: No difference. Same thing.

helix-core/src/test/java/org/apache/helix/controller/rebalancer/waged/TestAssignmentMetadataStore.java
```diff
@@ -172,13 +197,17 @@ public void testAssignmentCache() {
...
+   *
    * @param metadataType
    * @return
    */
   private List<String> getExistingVersionNumbers(String metadataType) {
     List<String> children = _baseAccessor
         .getChildNames("/" + CLUSTER_NAME + "/ASSIGNMENT_METADATA/" + metadataType,
             AccessOption.PERSISTENT);
+    if (children == null) {
+      children = Collections.EMPTY_LIST;
```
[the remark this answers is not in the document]
jiajunwang: It doesn't hurt. What if we want to add something to the list in the future?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
kaisun2000, line 73: This change is non-essential, right?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -95,53 +95,56 @@ protected AssignmentMetadataStore(BucketDataAccessor bucketDataAccessor, String
...
-    HelixProperty combinedAssignments =
-        combineAssignments(BEST_POSSIBLE_KEY, bestPossibleAssignment);
+    HelixProperty combinedAssignments = combineAssignments(key, newAssignment);
     try {
-      _dataAccessor.compressedBucketWrite(_bestPossiblePath, combinedAssignments);
+      _dataAccessor.compressedBucketWrite(path, combinedAssignments);
     } catch (IOException e) {
       // TODO: Improve failure handling
-      throw new HelixException("Failed to persist BestPossibleAssignment!", e);
+      throw new HelixException(
```
kaisun2000, line 141: What is going to catch this one?

kaisun2000:
One high level question. `public synchronized boolean persistBaseline` or bestPossible, the synchronized is to make sure that with in a process, the persisting to ZK is guarded. Is there a case that in a single controller java process, there will be more than one waged pipeline running? My understanding is that for CRUSH full auto, we have per cluster controller pipeline. But for waged, they is only one for all clusters? Am I wrong here?

kaisun2000:
My follow-up question is actually that there seems still some potential inconsistency issue here not address across controllers over different machine. 

The pattern is similar to this https://github.com/apache/helix/pull/1066. 

If one controller session expired, and super controller assign another controller to take care the cluster. When first controller come up, his controller pipeline may still run and persists his baseline and best possible again. How do we guard this across java process race condition?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
jiajunwang: It is. In this PR, I changed the logic so the map will be kept. It needs to be modifiable.
Another option is to introduce more logic to handle the corner cases, which I don't prefer.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
jiajunwang: The rebalancer will do it.

jiajunwang:
> One high level question. `public synchronized boolean persistBaseline` or bestPossible, the synchronized is to make sure that with in a process, the persisting to ZK is guarded. Is there a case that in a single controller java process, there will be more than one waged pipeline running? My understanding is that for CRUSH full auto, we have per cluster controller pipeline. But for waged, they is only one for all clusters? Am I wrong here?

You are wrong.
1. Both WAGED and the traditional rebalancers are for each cluster.
2. synchronized is not a must if we make the assumption that it will only be used in the WAGED rebalancer. And we are not going to change the WAGED rebalancer.
3. Having this to protect is a nice-to-have thing. So do you suggest we remove it? I think it won't hurt.

jiajunwang:
> My follow-up question is actually that there seems still some potential inconsistency issue here not addressed, namely across controllers over different machine.
> 
> The pattern is similar to this #1066.
> 
> If one controller session expired, and super controller assign another controller to take care the cluster. When first controller come up, his controller pipeline may still run and persists his baseline and best possible again. How do we guard this across java process race condition?
> 
> This is a pattern of similar problems. We should think a little bit more generic and address them all long term.

Please note that we have an MVCC implemented in the bucket accessor to ensure cross node consistency.

kaisun2000:
> You are wrong.
> 
> 1. Both WAGED and the traditional rebalancers are for each cluster.
> 2. synchronized is not a must if we make the assumption that it will only be used in the WAGED rebalancer. And we are not going to change the WAGED rebalancer.
> 3. Having this to protect is a nice-to-have thing. So do you suggest we remove it? I think it won't hurt.

So the invariant is that each cluster has its own copy of persisted baseline and best-possible. If this is clear, removing sync seems to be safe.

jiajunwang:
> > You are wrong.
> > 
> > 1. Both WAGED and the traditional rebalancers are for each cluster.
> > 2. synchronized is not a must if we make the assumption that it will only be used in the WAGED rebalancer. And we are not going to change the WAGED rebalancer.
> > 3. Having this to protect is a nice-to-have thing. So do you suggest we remove it? I think it won't hurt.
> 
> So the invariant is that each cluster has its own copy of persisted baseline and best-possible. If this is clear, removing sync seems to be safe.

But it won't hurt. And IMO, the assumption that we won't change WAGED logic in the future is weak. I prefer to keep it.

kaisun2000:
> But it won't hurt. And IMO, the assumption that we won't change WAGED logic in the future is weak. I prefer to keep it.

As an aside, I recall waged has an advantage as allocating over all clusters to instance. So this is not in this version?

jiajunwang:
"allocating over all clusters to instance" what does that mean? Could you please give me an example?


Cheers,
-Jiajun

________________________________

kaisun2000:
> Please note that we have an MVCC implemented in the bucket accessor to ensure cross node consistency.

I will try to examine the bucket accessor a little bit more carefully. Here is the question:

Say controller A is for resource R allocation. Then controller A session expires, controller B take care of resource B. Later when A is back, for a while A and B would run at the same time. When serializing say baseline for resource R,  A wins over B, is it going to be a problem for this resource as B would later be controller for R.

jiajunwang:
Whichever wins is fine. Since the next pipeline running will take the cached result and keep calculating based on there.

  *   If A wins, then A keeps calculating based on its local state.
  *   If B wins, then because A cache the result after write, it will still calculate based on its local state. And the next pipeline will overwrite B's output. And we are good.
In this case, we don't really need B's result.


Cheers,
-Jiajun

________________________________

kaisun2000:
I see. Thx for the explanation.

jiajunwang:
Thanks for the review : )


Cheers,
-Jiajun

________________________________

jiajunwang:

PR #1124, opened 2020-06-26
Add stale message map to improve P2P message rate

zhangmeng916:
### Issues

(#1067 )

### Description

This PR adds the stale message map in Instance Message cache and updates it in every current state computation stage. For those stale message, we do not put it in pending messages, so that it won't block the next state transition on that partition.

### Tests

helix-core
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestRoutingTableProviderPeriodicRefresh.testPeriodicRefresh:192 expected:<7> but was:<6>
[ERROR]   TestWorkflowTimeout.testWorkflowRunningTime:66 » Helix Workflow "testWorkflowR...
[INFO]
[ERROR] Tests run: 1153, Failures: 2, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:17 h
[INFO] Finished at: 2020-07-21T10:44:32-07:00

helix-rest:
[INFO] Tests run: 163, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 31.269 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  35.057 s
[INFO] Finished at: 2020-07-14T23:21:11-07:00

helix-rest:
[INFO] Tests run: 163, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 31.265 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  35.137 s
[INFO] Finished at: 2020-07-15T11:25:19-07:00

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -153,11 +153,17 @@ public class InstanceMessagesCache {
         "END: InstanceMessagesCache.refresh(), {} of Messages read from ZooKeeper. took {} ms. ",
         newMessageKeys.size(), (System.currentTimeMillis() - startTime));
 
+    refreshStaleMessageCache();
     return true;
   }
 
   public Map<String, Map<String, Message>> getStaleMessageCache() {
     return _staleMessageCache;
+  }
+
+  public void addStaleMessage(String instanceName, Message staleMessage) {
+    _staleMessageCache.putIfAbsent(instanceName, new HashMap<>());
+    _staleMessageCache.get(instanceName).putIfAbsent(staleMessage.getMsgId(), staleMessage);
   }
 
   /**
@@ -523,6 +529,22 @@ public class InstanceMessagesCache {
     }
     _relayMessageCache.get(instanceName).put(relayMessage.getId(), relayMessage);
     _relayHostMessageCache.put(relayMessage.getMsgId(), hostMessage);
+  }
+
+  // filter stale message cache by message cache to remove deleted messages
+  private void refreshStaleMessageCache() {
+    LOG.info("Start to refresh stale message cache");
+    for (String instanceName : _staleMessageCache.keySet()) {
+      for (String messageId : _staleMessageCache.get(instanceName).keySet()) {
+        if (!_messageCache.getOrDefault(instanceName, Collections.emptyMap())
+            .containsKey(messageId)) {
+          _staleMessageCache.get(instanceName).remove(messageId);
+          if (_staleMessageCache.get(instanceName).size() == 0) {
+            _staleMessageCache.remove(instanceName);
+          }
+        }
+      }
+    }
   }
 
   @Override public String toString() {
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -156,7 +156,7 @@ public class InstanceMessagesCache {
   public Map<String, Map<String, Message>> getStaleMessageCache() {
     return _staleMessageCache;
   }
-  
+
   /**
    * Refresh relay message cache by updating relay messages read from ZK, and remove all expired relay messages.
    */
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -152,27 +152,18 @@ public class InstanceMessagesCache {
     LOG.info(
         "END: InstanceMessagesCache.refresh(), {} of Messages read from ZooKeeper. took {} ms. ",
         newMessageKeys.size(), (System.currentTimeMillis() - startTime));
+
+    refreshStaleMessageCache();
+
     return true;
   }
 
-  // filter stale message cache by message cache to remove deleted messages
   public Map<String, Map<String, Message>> getStaleMessageCache() {
-    for (String instanceName : _staleMessageCache.keySet()) {
-      for (String messageId : _staleMessageCache.get(instanceName).keySet()) {
-        if (!_messageCache.getOrDefault(instanceName, Collections.emptyMap())
-            .containsKey(messageId)) {
-          _staleMessageCache.get(instanceName).remove(messageId);
-          if (_staleMessageCache.get(instanceName).size() == 0) {
-            _staleMessageCache.remove(instanceName);
-          }
-        }
-      }
-    }
     return _staleMessageCache;
   }
 
-  public void setStaleMessageCache(Map<String, Map<String, Message>> staleMessageMap) {
-    _staleMessageCache = staleMessageMap;
+  public void cacheStaleMessageCache(Map<String, Map<String, Message>> staleMessageMap) {
+    _staleMessageCache = new HashMap<>(staleMessageMap);
   }
   /**
    * Refresh relay message cache by updating relay messages read from ZK, and remove all expired relay messages.
@@ -537,6 +528,21 @@ public class InstanceMessagesCache {
     }
     _relayMessageCache.get(instanceName).put(relayMessage.getId(), relayMessage);
     _relayHostMessageCache.put(relayMessage.getMsgId(), hostMessage);
+  }
+
+  // filter stale message cache by message cache to remove deleted messages
+  private void refreshStaleMessageCache() {
+    for (String instanceName : _staleMessageCache.keySet()) {
+      for (String messageId : _staleMessageCache.get(instanceName).keySet()) {
+        if (!_messageCache.getOrDefault(instanceName, Collections.emptyMap())
+            .containsKey(messageId)) {
+          _staleMessageCache.get(instanceName).remove(messageId);
+          if (_staleMessageCache.get(instanceName).size() == 0) {
+            _staleMessageCache.remove(instanceName);
+          }
+        }
+      }
+    }
   }
 
   @Override public String toString() {
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
@@ -157,8 +158,17 @@ public class InstanceMessagesCache {
     return true;
   }
 
+  @VisibleForTesting
   public Map<String, Map<String, Message>> getStaleMessageCache() {
     return _staleMessageCache;
+  }
+
+  public Set<Message> getStaleMessagesByInstance(String instanceName) {
+    Map<String, Message> staleMessageMap = _staleMessageCache.get(instanceName);
+    if (staleMessageMap != null) {
+      return new HashSet<>(staleMessageMap.values());
+    }
+    return Collections.emptySet();
   }
 
   public void addStaleMessage(String instanceName, Message staleMessage) {
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -531,20 +531,25 @@ public class InstanceMessagesCache {
     _relayHostMessageCache.put(relayMessage.getMsgId(), hostMessage);
   }
 
-  // filter stale message cache by message cache to remove deleted messages
+  // filter stale message cache by message cache to remove already deleted messages
   private void refreshStaleMessageCache() {
     LOG.info("Start to refresh stale message cache");
+    Map<String, Set<String>> toRemoveMessages = new HashMap<>();
     for (String instanceName : _staleMessageCache.keySet()) {
       for (String messageId : _staleMessageCache.get(instanceName).keySet()) {
         if (!_messageCache.getOrDefault(instanceName, Collections.emptyMap())
             .containsKey(messageId)) {
-          _staleMessageCache.get(instanceName).remove(messageId);
-          if (_staleMessageCache.get(instanceName).size() == 0) {
-            _staleMessageCache.remove(instanceName);
-          }
+          toRemoveMessages.computeIfAbsent(instanceName, k -> new HashSet<>()).add(messageId);
         }
       }
     }
+
+    toRemoveMessages.entrySet().stream().forEach(entry -> {
+      entry.getValue().stream().forEach(id -> _staleMessageCache.get(entry.getKey()).remove(id));
+      if (_staleMessageCache.get(entry.getKey()).size() == 0) {
+        _staleMessageCache.remove(entry.getKey());
+      }
+    });
   }
 
   @Override public String toString() {
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -157,6 +157,9 @@ public class InstanceMessagesCache {
     return _staleMessageCache;
   }
 
+  public void setStaleMessageCache(Map<String, Map<String, Message>> staleMessageMap) {
+    _staleMessageCache = staleMessageMap;
+  }
   /**
    * Refresh relay message cache by updating relay messages read from ZK, and remove all expired relay messages.
    */
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -155,7 +155,19 @@ public class InstanceMessagesCache {
     return true;
   }
 
+  // filter stale message cache by message cache to remove deleted messages
   public Map<String, Map<String, Message>> getStaleMessageCache() {
+    for (String instanceName : _staleMessageCache.keySet()) {
+      for (String messageId : _staleMessageCache.get(instanceName).keySet()) {
+        if (!_messageCache.getOrDefault(instanceName, Collections.emptyMap())
+            .containsKey(messageId)) {
+          _staleMessageCache.get(instanceName).remove(messageId);
+          if (_staleMessageCache.get(instanceName).size() == 0) {
+            _staleMessageCache.remove(instanceName);
+          }
+        }
+      }
+    }
     return _staleMessageCache;
   }
 
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -55,6 +55,8 @@ public class InstanceMessagesCache {
   // <instance -> {<MessageId, Message>}>
   private Map<String, Map<String, Message>> _messageCache = Maps.newHashMap();
 
+  private Map<String, Map<String, Message>> _staleMessageCache = Maps.newHashMap();
+
   // maintain a set of valid pending P2P messages.
   // <instance -> {<MessageId, Message>}>
   private Map<String, Map<String, Message>> _relayMessageCache = Maps.newHashMap();
@@ -151,6 +153,13 @@ public class InstanceMessagesCache {
     return true;
   }
 
+  public Map<String, Map<String, Message>> getStaleMessageCache() {
+    return _staleMessageCache;
+  }
+
+  public void setStaleMessageCache(Map<String, Map<String, Message>> staleMessageMap) {
+    _staleMessageCache = staleMessageMap;
+  }
   /**
    * Refresh relay message cache by updating relay messages read from ZK, and remove all expired relay messages.
    */
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -55,6 +55,8 @@ public class InstanceMessagesCache {
   // <instance -> {<MessageId, Message>}>
   private Map<String, Map<String, Message>> _messageCache = Maps.newHashMap();
 
+  // maintain a cache of stale messages
+  // <instance -> {<MessageId, Message>}>
   private Map<String, Map<String, Message>> _staleMessageCache = Maps.newHashMap();
 
   // maintain a set of valid pending P2P messages.
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -156,10 +156,7 @@ public class InstanceMessagesCache {
   public Map<String, Map<String, Message>> getStaleMessageCache() {
     return _staleMessageCache;
   }
-
-  public void setStaleMessageCache(Map<String, Map<String, Message>> staleMessageMap) {
-    _staleMessageCache = staleMessageMap;
-  }
+  
   /**
    * Refresh relay message cache by updating relay messages read from ZK, and remove all expired relay messages.
    */
--- a/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
@@ -153,8 +153,6 @@ public class InstanceMessagesCache {
         "END: InstanceMessagesCache.refresh(), {} of Messages read from ZooKeeper. took {} ms. ",
         newMessageKeys.size(), (System.currentTimeMillis() - startTime));
 
-    refreshStaleMessageCache();
-
     return true;
   }
 
@@ -162,9 +160,6 @@ public class InstanceMessagesCache {
     return _staleMessageCache;
   }
 
-  public void cacheStaleMessageCache(Map<String, Map<String, Message>> staleMessageMap) {
-    _staleMessageCache = new HashMap<>(staleMessageMap);
-  }
   /**
    * Refresh relay message cache by updating relay messages read from ZK, and remove all expired relay messages.
    */
@@ -528,21 +523,6 @@ public class InstanceMessagesCache {
     }
     _relayMessageCache.get(instanceName).put(relayMessage.getId(), relayMessage);
     _relayHostMessageCache.put(relayMessage.getMsgId(), hostMessage);
-  }
-
-  // filter stale message cache by message cache to remove deleted messages
-  private void refreshStaleMessageCache() {
-    for (String instanceName : _staleMessageCache.keySet()) {
-      for (String messageId : _staleMessageCache.get(instanceName).keySet()) {
-        if (!_messageCache.getOrDefault(instanceName, Collections.emptyMap())
-            .containsKey(messageId)) {
-          _staleMessageCache.get(instanceName).remove(messageId);
-          if (_staleMessageCache.get(instanceName).size() == 0) {
-            _staleMessageCache.remove(instanceName);
-          }
-        }
-      }
-    }
   }
 
   @Override public String toString() {
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -544,6 +544,13 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     return _instanceMessagesCache.getMessages(instanceName);
   }
 
+  public void setStaleMessages(Map<String, Map<String, Message>> staleMessageMap) {
+    _instanceMessagesCache.setStaleMessageCache(staleMessageMap);
+  }
+
+  public Map<String, Map<String, Message>> getStaleMessages() {
+    return _instanceMessagesCache.getStaleMessageCache();
+  }
   /**
    * Provides a list of current outstanding pending relay messages on a given instance.
    * @param instanceName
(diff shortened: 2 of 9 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -56,11 +57,17 @@
  */
 public class CurrentStateComputationStage extends AbstractBaseStage {
   private static Logger LOG = LoggerFactory.getLogger(CurrentStateComputationStage.class);
+  private boolean isTaskFrameworkPipeline = false;
```
jiajunwang, line 60: nit, _ isTaskFrameworkPipeline

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -544,6 +544,13 @@ public synchronized void setLiveInstances(List<LiveInstance> liveInstances) {
     return _instanceMessagesCache.getMessages(instanceName);
   }
 
+  public void setStaleMessages(Map<String, Map<String, Message>> staleMessageMap) {
```
jiajunwang, line 547: This is never called? Or there are some other changes that you haven't pushed?

zhangmeng916: Yeah, it's never called. Originally I used it, but then I simplified the code, and it's not used any more. I've removed.

helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
```diff
@@ -151,6 +153,13 @@ public boolean refresh(HelixDataAccessor accessor, Map<String, LiveInstance> liv
     return true;
   }
 
+  public Map<String, Map<String, Message>> getStaleMessageCache() {
+    return _staleMessageCache;
+  }
+
+  public void setStaleMessageCache(Map<String, Map<String, Message>> staleMessageMap) {
```
jiajunwang: Not used?

jiajunwang: If no set is called, how the items in the cache being updated?

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -236,6 +267,11 @@ private void updateCurrentStates(LiveInstance instance, Collection<CurrentState>
     }
   }
 
+  private void setStaleMessage(CurrentStateOutput currentStateOutput, String instanceName,
```
jiajunwang, line 270: Can we just call "currentStateOutput.setStaleMessage(instanceName, message);" in the caller so we don't need this method?

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -191,6 +213,15 @@ private void updatePendingMessages(LiveInstance instance, Collection<Message> pe
     }
   }
 
+  private boolean isStaleMessage(Message message, String currentState) {
+    if (_isTaskFrameworkPipeline || currentState == null || message.getFromState() == null
```
jiajunwang, line 217: Introducing _isTaskFrameworkPipeline to isStaleMessage is confusing. I would prefer do this,
"if (!_isTaskFrameworkPipeline && isStaleMessage(...)) {set stale message}"

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -344,6 +367,13 @@ private void schedulePendingMessageCleanUp(
             if (accessor.removeProperty(msg.getKey(accessor.keyBuilder(), instanceName))) {
               LogUtil.logInfo(logger, _eventId, String
                   .format("Deleted message %s from instance %s", msg.getMsgId(), instanceName));
+              if (staleMessageMap != null && staleMessageMap.containsKey(msg.getTgtName())
+                  && staleMessageMap.get(msg.getTgtName()).containsKey(msg.getMsgId())) {
```
jiajunwang, line 371: nit, could be simpler by
if (staleMessageMap != null && staleMessageMap.getOrDefault(msg.getTgtName(), EmptyMap()).remove(msg.getMsgId())) {
  if (staleMessageMap.get(msg.getTgtName()).size() == 0) {
                  staleMessageMap.remove(msg.getTgtName());
                }
}

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -371,6 +401,14 @@ private boolean shouldCleanUpPendingMessage(Message pendingMsg, String currentSt
     }
   }
 
+  private boolean shouldCleanUpStaleMessage(Message staleMsg, Long currentStateTransitionEndTime) {
+    if (staleMsg == null) {
```
jiajunwang, line 405: Can staleMsg be null?

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -344,6 +367,13 @@ private void schedulePendingMessageCleanUp(
             if (accessor.removeProperty(msg.getKey(accessor.keyBuilder(), instanceName))) {
               LogUtil.logInfo(logger, _eventId, String
                   .format("Deleted message %s from instance %s", msg.getMsgId(), instanceName));
+              if (staleMessageMap != null && staleMessageMap.containsKey(msg.getTgtName())
```
jiajunwang, line 370: I think staleMessageMap won't be null.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -188,14 +192,24 @@ private void generateMessage(final Resource resource, final BaseControllerDataPr
 
         if (pendingMessage != null && shouldCleanUpPendingMessage(pendingMessage, currentState,
             currentStateOutput.getEndTime(resourceName, partition, instanceName))) {
-          LogUtil.logInfo(logger, _eventId, String.format(
-              "Adding pending message %s on instance %s to clean up. Msg: %s->%s, current state of resource %s:%s is %s",
-              pendingMessage.getMsgId(), instanceName, pendingMessage.getFromState(),
-              pendingMessage.getToState(), resourceName, partition, currentState));
-          if (!pendingMessagesToCleanUp.containsKey(instanceName)) {
-            pendingMessagesToCleanUp.put(instanceName, new HashMap<String, Message>());
+          String logMsg = String.format(
```
jiajunwang, line 195: Better to move this logic into the logAndAddToCleanUp() method too.
This will increase the parameters, but the code here would be much simpler.

jiajunwang: This also helps to reduce the duplicate code that logs the stale messages.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
zhangmeng916: I think it won't be. It's got instantiated in `CurrentStateOutput`

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
zhangmeng916: Normally it shouldn't. But I'm not 100% sure. I saw pendingMessage has similar check. Want to keep it for safe.

helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
zhangmeng916: So it's used by this way:
Map<String, Map<String, Message>> existingStaleMessages = cache.getStaleMessages();
currentStateOutput.setStaleMessageMap(existingStaleMessages);

So when the currentStateOutput's stale message gets updated, the instance cache should also be updated, right? as it's just shallow copy. Let me know if you think differently.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
zhangmeng916: Please take a look at the updated one. I believe it's safe.

helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
jiajunwang: This is not a good style. Better to make it explicitly updated.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
jiajunwang: Both versions are safe. Just too verbose.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
jiajunwang: I personally prefer to fail early. If it should not, then we should not check. The concern is that overprotected code will hide potential issues.
For example, if some bugs make it always null, then you will always get false as a result. But it would be better that we get an NPE, then the bug will be identified the first time.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateOutput.java
```diff
@@ -238,6 +256,16 @@ public Message getPendingMessage(String resourceName, Partition partition, Strin
     return getStateMessage(resourceName, partition, instanceName, _pendingMessageMap);
   }
 
+  public Set<Message> getStaleMessages(String instanceName) {
```
jiajunwang, line 259: nit, maybe call it getStaleMessagesByInstance would help to differentiate it from getStaleMessageMap

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -77,8 +81,9 @@ protected void processEvent(ClusterEvent event, ResourcesStateMap resourcesState
     BaseControllerDataProvider cache = event.getAttribute(AttributeName.ControllerDataProvider.name());
     Map<String, Resource> resourceMap =
         event.getAttribute(AttributeName.RESOURCES_TO_REBALANCE.name());
-    Map<String, Map<String, Message>> pendingMessagesToCleanUp = new HashMap<>();
     CurrentStateOutput currentStateOutput = event.getAttribute(AttributeName.CURRENT_STATE.name());
+
+    Map<String, Map<String, Message>> pendingMessagesToCleanUp = new HashMap<>();
```
jiajunwang, line 86: Takes me a while to figure it out. So we shall call it messagesToCleanup now, right?

jiajunwang: Same for the names of the other parameters that are pointing to this same map.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -335,15 +362,21 @@ private void addGeneratedMessageToMap(final Message message,
...
-      @Override public Object call() {
+      @Override
+      public Object call() {
         for (Map.Entry<String, Map<String, Message>> entry : pendingMessagesToPurge.entrySet()) {
           String instanceName = entry.getKey();
           for (Message msg : entry.getValue().values()) {
             if (accessor.removeProperty(msg.getKey(accessor.keyBuilder(), instanceName))) {
               LogUtil.logInfo(logger, _eventId, String
                   .format("Deleted message %s from instance %s", msg.getMsgId(), instanceName));
+              staleMessageMap.getOrDefault(msg.getTgtName(), Collections.emptyMap())
```
jiajunwang, line 375: So based on what I commented before, if we calculate from scratch in every pipeline, we won't need to maintain this map here, right? Or we will have some usage side this stage that we will need to update the content of the map at runtime?

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -371,6 +404,11 @@ private boolean shouldCleanUpPendingMessage(Message pendingMsg, String currentSt
     }
   }
 
+  private boolean shouldCleanUpStaleMessage(Long currentStateTransitionEndTime) {
```
jiajunwang, line 407: nit, I think we can just put this logic to the caller. Since there is just one caller, and the logic is simple. The additional method layers will increase latency a little bit you know.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -77,15 +84,19 @@ public void process(ClusterEvent event) throws Exception {
...
       updateCurrentStates(instance, currentStateMap.values(), currentStateOutput, resourceMap);
+
+      Map<String, Map<String, Message>> existingStaleMessages = cache.getStaleMessages();
+      currentStateOutput.setStaleMessageMap(existingStaleMessages);
+      // update pending messages
+      Map<String, Message> messages = cache.getMessages(instanceName);
+      Map<String, Message> relayMessages = cache.getRelayMessages(instanceName);
+      updatePendingMessages(instance, messages.values(), currentStateOutput,
+          relayMessages.values(), resourceMap, existingStaleMessages);
+      cache.setStaleMessages(currentStateOutput.getStaleMessageMap());
```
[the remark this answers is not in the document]
zhangmeng916: The purpose of the cache is not for performance, but to keep all the historical stale messages. For example, assume a partition goes through M->S->M->S, and the first M->S message is not deleted even during the second M->S transition, if we only look at pending messages only, the first M->S message would be considered as non stale message, which causes confusion. With the cache, the stale messages would not be checked again.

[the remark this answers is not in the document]
jiajunwang: Then we need to define how the historical stale messages records are removed from the cache, or we will need to reboot the controller periodically, right?
Another concern is that if leadership switch happen, then you will still lose track of the first M->S message. So this is not a complete fix.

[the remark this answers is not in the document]
zhangmeng916: The stale message is periodically cleaned up the same way as `pending message`, and the logic is in message generation phase. 
If controller switch happens, the in memory data is just lost right? Then all stale messages are automatically cleaned up. Any concern?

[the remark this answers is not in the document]
jiajunwang: I got the first point. Should be fine. Thanks for the explanation.

Regarding the leadership switch, the new leader won't know your cached value. So it will treat the previous message ( "first M->S message") as not stale. In general, if you have any logic depends on the cache, then the leadership switch will be troublesome.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
zhangmeng916: currentStateOutput is already updated with latest stale messages, and accordingly, the cache is updated. So this method is actually redundant, and that's why I used to remove it. But I think from your previous comment, you'd like to an explicit `set` operation?

jiajunwang: You mean we can just remove it? The comment here is just saying that this private method is not necessary : )

But regarding the set, yes, I think we need explicit set. Implicit set always introduces bugs.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
zhangmeng916: We do need for maintaining the history from previous pipelines. This could help better reduce stale messages.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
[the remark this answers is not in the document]
zhangmeng916: I see what you mean. I was talking about the case that message is gone from zk, but still in cache. So if the message hasn't been deleted from ZK yet, then yes, it'll stay there, and may cause confusion. I'm not sure whether this would cause a logic error. But I cannot think of other option unless we mark it in ZK.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
zhangmeng916: I mean this two lines:
updatePendingMessages(instance, messages.values(), relayMessages.values(), currentStateOutput,
          resourceMap);
      cache.setStaleMessages(currentStateOutput.getStaleMessageMap());

The second one is redundant. currentStateOutput is already updated in the function. But seems you think it's better to have the second one?

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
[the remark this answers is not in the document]
zhangmeng916: I think about it, I don't think it'll cause any problem even when controller switches. The message won't be processed if the target state is not the ideal state. The only problem is that the code won't automatically remove it for us.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
jiajunwang: This is related to my other comment. In general, we shall not pass a modifiable map ref directly. Especially if the passed ref will be used to update. And given that you make a shallow copy instead of just pass the ref, then we will need both.

helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
```diff
@@ -151,6 +155,25 @@ public boolean refresh(HelixDataAccessor accessor, Map<String, LiveInstance> liv
     return true;
   }
 
+  // filter stale message cache by message cache to remove deleted messages
+  public Map<String, Map<String, Message>> getStaleMessageCache() {
+    for (String instanceName : _staleMessageCache.keySet()) {
```
jiajunwang, line 160: I think you can do this when the _messageCache is refreshed in refresh() method.
1. changing values in a get method is not a good style. We want to be able to call get any time, but if it has a side effect then the callers need to review the detail of the method every time if they want to use it.
2. This is for cleaning up and prevent memory leakage only, so clean up when refresh is done at the beginning of the pipeline is enough, I think.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -77,15 +87,20 @@ public void process(ClusterEvent event) throws Exception {
...
-      Map<String, Message> messages = cache.getMessages(instanceName);
-      Map<String, Message> relayMessages = cache.getRelayMessages(instanceName);
-      updatePendingMessages(instance, messages.values(), currentStateOutput, relayMessages.values(), resourceMap);
-
       // update current states.
       Map<String, CurrentState> currentStateMap = cache.getCurrentState(instanceName,
           instanceSessionId);
       updateCurrentStates(instance, currentStateMap.values(), currentStateOutput, resourceMap);
+
+      Map<String, Map<String, Message>> existingStaleMessages =
```
jiajunwang, line 95: nit, can we follow the same style that how the other 2 maps are used here?
Basically passing the existingStaleMessages to updatePendingMessages too instead of set to the currentStateOutput here?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -544,6 +544,14 @@ public synchronized void setLiveInstances(List<LiveInstance> liveInstances) {
     return _instanceMessagesCache.getMessages(instanceName);
   }
 
+  public Map<String, Map<String, Message>> getStaleMessages() {
+    return _instanceMessagesCache.getStaleMessageCache();
+  }
+
+  public void setStaleMessages(Map<String, Map<String, Message>> staleMessageMap) {
```
jiajunwang, line 551: nit, cacheStaleMessages so all the updating message cache methods are named in the same way?

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -335,15 +363,21 @@ private void addGeneratedMessageToMap(final Message message,
...
-      @Override public Object call() {
+      @Override
+      public Object call() {
         for (Map.Entry<String, Map<String, Message>> entry : pendingMessagesToPurge.entrySet()) {
           String instanceName = entry.getKey();
           for (Message msg : entry.getValue().values()) {
             if (accessor.removeProperty(msg.getKey(accessor.keyBuilder(), instanceName))) {
               LogUtil.logInfo(logger, _eventId, String
                   .format("Deleted message %s from instance %s", msg.getMsgId(), instanceName));
+              staleMessageMap.getOrDefault(msg.getTgtName(), Collections.emptyMap())
```
jiajunwang, line 376: Now I have a better understanding of your change. And I think this cleanup is not necessary.
The stale message map in the current state output object is re-calculated from the cache every time. So as long as we clean the cache properly, we don't need to care about the output map. Please correct me if I missed anything.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
zhangmeng916: Actually I found in `InstanceMessageCache`, more are named with `cache`, so I changed in that file.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
zhangmeng916: yep, I agree. It'll go away later anyway. Shouldn't be a concern to leave it a little longer.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateOutput.java
```diff
@@ -165,6 +168,19 @@ public void setPendingRelayMessage(String resourceName, Partition partition, Str
     setStateMessage(resourceName, partition, instanceName, message, _pendingRelayMessageMap);
   }
 
+  public Map<String, Map<String, Message>> getStaleMessageMap() {
+    return _staleMessageMap;
+  }
+
+  public void setStaleMessageMap(Map<String, Map<String, Message>> staleMessageMap) {
+    _staleMessageMap = staleMessageMap;
```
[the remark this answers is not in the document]
zhangmeng916: The this one should also be changed?  cache.setStaleMessages(existingStaleMessages);
Is this a rule we have?

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -188,14 +196,21 @@ private void generateMessage(final Resource resource, final BaseControllerDataPr
...
-              pendingMessage.getToState(), resourceName, partition, currentState));
-          if (!pendingMessagesToCleanUp.containsKey(instanceName)) {
-            pendingMessagesToCleanUp.put(instanceName, new HashMap<String, Message>());
+          logAndAddToCleanUp(messagesToCleanUp, pendingMessage, instanceName, resourceName,
+              partition, currentState, PENDING_MESSAGE);
+        }
+
+        for (Message staleMessage : staleMessages) {
+          if (staleMessage == null) {
+            logger.warn("Should not contain a stale message as null");
```
jiajunwang, line 205: Is it really possible that we have a null element in the return set?

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -102,15 +114,22 @@ public void process(ClusterEvent event) throws Exception {
...
+      Collection<Message> pendingRelayMessages,
+      Map<String, Map<String, Message>> existingStaleMessages,
+      CurrentStateOutput currentStateOutput, Map<String, Resource> resourceMap) {
     String instanceName = instance.getInstanceName();
     String instanceSessionId = instance.getEphemeralOwner();
 
     // update all pending messages
     for (Message message : pendingMessages) {
+      // ignore existing stale messages
+      if (existingStaleMessages.getOrDefault(instanceName, Collections.emptyMap())
```
jiajunwang, line 126: We can call currentStateOutput.getStaleMessagesByInstance() to get the message set as the input, right?

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -129,7 +148,14 @@ private void updatePendingMessages(LiveInstance instance, Collection<Message> pe
         String partitionName = message.getPartitionName();
         Partition partition = resource.getPartition(partitionName);
         if (partition != null) {
-          setMessageState(currentStateOutput, resourceName, partition, instanceName, message);
+          String currentState = currentStateOutput.getCurrentState(resourceName, partition,
+              instanceName);
+          if (_isTaskFrameworkPipeline || !isStaleMessage(message, currentState)) {
+            setMessageState(currentStateOutput, resourceName, partition, instanceName, message);
+          } else {
+            existingStaleMessages.putIfAbsent(instanceName, new HashMap<>());
```
jiajunwang, line 156: As mentioned above, I prefer this to be an explicit set method, addStaleMessage()

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -544,6 +544,19 @@ public synchronized void setLiveInstances(List<LiveInstance> liveInstances) {
     return _instanceMessagesCache.getMessages(instanceName);
   }
 
+  public Map<String, Map<String, Message>> getStaleMessages() {
```
jiajunwang, line 547: So this one will return the map for modifying. I still don't like this implicit map/list ref for modifying purposes.
As I checked the current usage, I think you can replace it with 2 explicit modifying methods,
1. addStaleMessage(instance, message)
2. removeStaleMessage(instance, message) -- or maybe always clean up when the cache is refreshed, so you don't even need this method?

As for supporting test, we can make it protected so only visible to test. Or keeping this method but always return shallow copy of the map only for the tests.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
zhangmeng916: CurrentStateOutput currently does not have stale messages in it. It's not necessary because every class has access to cache.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
zhangmeng916: Shouldn't be possible. Removed.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -544,6 +544,27 @@ public synchronized void setLiveInstances(List<LiveInstance> liveInstances) {
     return _instanceMessagesCache.getMessages(instanceName);
   }
 
+  /**
+   * This function is supposed to be only used by testing purpose for safety. For "get" usage,
+   * please use getStaleMessagesByInstance.
+   */
+  public Map<String, Map<String, Message>> getStaleMessages() {
```
jiajunwang, line 551: Let's add a @VisibleForTesting tag here

helix-core/src/main/java/org/apache/helix/common/caches/InstanceMessagesCache.java
```diff
@@ -148,9 +152,20 @@ public boolean refresh(HelixDataAccessor accessor, Map<String, LiveInstance> liv
     LOG.info(
         "END: InstanceMessagesCache.refresh(), {} of Messages read from ZooKeeper. took {} ms. ",
         newMessageKeys.size(), (System.currentTimeMillis() - startTime));
+
+    refreshStaleMessageCache();
     return true;
   }
 
+  public Map<String, Map<String, Message>> getStaleMessageCache() {
```
jiajunwang, line 160: I think we have this method mainly for supporting the test. Otherwise, we can have the getStaleMessageForInstance() here, right?

My suggestion is that we should not change our business logic in any way because of the test requirement. So I will say, let's move the getStaleMessageForInstance() implementation here. And leave this method as @VisibleForTesting only.

jiajunwang: Or, if you can make the test to check InstanceMessagesCache only, that would be even better. But that is optional.

zhangmeng916: Tried to only use `getStaleMessageCache()` in this class. But it's difficult to change the test to check this class directly.  So keep the function in two classes.

zhangmeng916:

Final commit message:
Add stale message cache to keep processed but not deleted messages. Controller will not need to wait for these messages tp be gone before sending more messages to participants.
This change will improve the rate of P2P messages and help reduce mastership handoff latency.

PR #1127, opened 2020-06-28
Fix validation logic in GenericApiBuilders and add usage tests

### LINKED ISSUE #1121: Fix GenericZkHelixApiBuilder and GenericBaseDataAccessorBuilder
protected RealmAwareZkClient createZkClient(RealmAwareZkClient.RealmMode realmMode,
      RealmAwareZkClient.RealmAwareZkClientConfig clientConfig, String zkAddress) {
    switch (realmMode) {
      case MULTI_REALM:
        try {
          return new FederatedZkClient(connectionConfig,
              clientConfig.setZkSerializer(new ZNRecordSerializer()));
        } catch (IOException | InvalidRoutingDataException | IllegalStateException e) {
          throw new HelixException("Failed to create FederatedZkClient!", e);
        }
      case SINGLE_REALM:
        // Create a HelixZkClient: Use a SharedZkClient because ClusterSetup does not need to do
        // ephemeral operations
        return SharedZkClientFactory.getInstance()
            .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
                clientConfig.createHelixZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
      default:
        throw new HelixException("Invalid RealmMode given: " + realmMode);
    }
  }

the RealmMode set in RealmAwareZkConnectionConfig is not being honored.

narendly:
### Issues

Fixes #1121 

### Description

Some of the validation logic in GenericBaseDataAccessorBuilder and GenericZkHelixApiBuilder were confusing and there were no usage tests. This commit reviews the consistency between validate methods of BaseDataAccessor APIs and writes test cases to demonstrate how they should be used.

Also, connection timeout setting wasn't being honored in the createZkClient() methods, so this bug was fixed as well.

createZkConnectionConfig() was removed from RealmAwareZkClient because the intended purpose of this method isn't clear and was not being used anywhere else.

### Tests

TestMultiZkHelixJavaApis

zookeeper-api:

```
[INFO] Tests run: 37, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 51.491 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 37, Failures: 0, Errors: 0, Skipped: 0
```

helix-core:

```
[ERROR] Failures: 
[ERROR]   TestEnableCompression.testEnableCompressionResource:116 » ThreadTimeout Method...
[ERROR]   TestAutoRebalanceWithDisabledInstance.testDisableEnableInstanceAutoRebalance:69 expected:<true> but was:<false>
[ERROR]   TestClusterInMaintenanceModeWhenReachingMaxPartition.testDisableCluster:119 expected:<true> but was:<false>
[ERROR]   TestTaskPerformanceMetrics.testTaskPerformanceMetrics:128 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1149, Failures: 4, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
```
Ran the failures individually:
```
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 50.52 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  56.059 s
[INFO] Finished at: 2020-07-01T12:05:50-07:00
```

helix-rest:
```
[INFO] Tests run: 163, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 33.513 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
```

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/GenericBaseDataAccessorBuilder.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/GenericBaseDataAccessorBuilder.java
@@ -99,18 +99,19 @@ public class GenericBaseDataAccessorBuilder<B extends GenericBaseDataAccessorBui
         break;
 
       case SINGLE_REALM:
+        HelixZkClient.ZkConnectionConfig helixZkConnectionConfig =
+            new HelixZkClient.ZkConnectionConfig(zkAddress)
+                .setSessionTimeout(connectionConfig.getSessionTimeout());
         if (_zkClientType == ZkClientType.DEDICATED) {
           // If DEDICATED, then we use a dedicated HelixZkClient because we must support ephemeral
           // operations
           zkClient = DedicatedZkClientFactory.getInstance()
-              .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
-                  clientConfig.createHelixZkClientConfig());
+              .buildZkClient(helixZkConnectionConfig, clientConfig.createHelixZkClientConfig());
         } else {
           // if SHARED: Use a SharedZkClient because ZkBaseDataAccessor does not need to
           // do ephemeral operations.
           zkClient = SharedZkClientFactory.getInstance()
-              .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress),
-                  clientConfig.createHelixZkClientConfig());
+              .buildZkClient(helixZkConnectionConfig, clientConfig.createHelixZkClientConfig());
           zkClient
               .waitUntilConnected(HelixZkClient.DEFAULT_CONNECTION_TIMEOUT, TimeUnit.MILLISECONDS);
         }
@@ -123,23 +124,25 @@ public class GenericBaseDataAccessorBuilder<B extends GenericBaseDataAccessorBui
 
   /**
    * Validate ZkClientType based on RealmMode.
+   * If ZkClientType is DEDICATED or SHARED, the realm mode must be SINGLE-REALM.
+   * If ZkClientType is FEDERATED, the realm mode must be MULTI-REALM.
    * @param zkClientType
    * @param realmMode
    */
   private void validateZkClientType(ZkClientType zkClientType,
       RealmAwareZkClient.RealmMode realmMode) {
-    boolean isZkClientTypeSet = zkClientType != null;
-    // If ZkClientType is set, RealmMode must either be single-realm or not set.
-    if (isZkClientTypeSet && realmMode == RealmAwareZkClient.RealmMode.MULTI_REALM) {
-      throw new HelixException("ZkClientType cannot be set on multi-realm mode!");
+    if (realmMode == null) {
+      throw new HelixException("Cannot validate ZkClient type: realmMode is null!");
     }
-    // If ZkClientType is not set and realmMode is single-realm, default to SHARED
-    if (!isZkClientTypeSet && realmMode == RealmAwareZkClient.RealmMode.SINGLE_REALM) {
-      zkClientType = ZkClientType.SHARED;
-    }
-    if (realmMode == RealmAwareZkClient.RealmMode.SINGLE_REALM
-        && zkClientType == ZkClientType.FEDERATED) {
-      throw new HelixException("FederatedZkClient cannot be set on single-realm mode!");
+    // If ZkClientType is DEDICATED or SHARED, the realm mode cannot be multi-realm.
+    // If ZkClientType is FEDERATED, the realm mode cannot be single-realm.
+    if (((zkClientType == ZkClientType.DEDICATED || zkClientType == ZkClientType.SHARED)
+        && realmMode == RealmAwareZkClient.RealmMode.MULTI_REALM) || (
+        zkClientType == ZkClientType.FEDERATED
+            && realmMode == RealmAwareZkClient.RealmMode.SINGLE_REALM)) {
+      throw new HelixException(
+          "Invalid combination of ZkClientType and RealmMode: ZkClientType is " + zkClientType
+              .name() + " and realmMode is " + realmMode.name());
     }
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/GenericBaseDataAccessorBuilder.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/GenericBaseDataAccessorBuilder.java
@@ -132,7 +132,8 @@ public class GenericBaseDataAccessorBuilder<B extends GenericBaseDataAccessorBui
   private void validateZkClientType(ZkClientType zkClientType,
       RealmAwareZkClient.RealmMode realmMode) {
     if (realmMode == null) {
-      throw new HelixException("Cannot validate ZkClient type: realmMode is null!");
+      throw new HelixException(
+          "GenericBaseDataAccessorBuilder: Cannot validate ZkClient type! RealmMode is null!");
     }
     // If ZkClientType is DEDICATED or SHARED, the realm mode cannot be multi-realm.
     // If ZkClientType is FEDERATED, the realm mode cannot be single-realm.
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/GenericBaseDataAccessorBuilder.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/GenericBaseDataAccessorBuilder.java
@@ -82,22 +82,11 @@ public class GenericBaseDataAccessorBuilder<B extends GenericBaseDataAccessorBui
     switch (realmMode) {
       case MULTI_REALM:
         try {
-          if (_zkClientType == ZkClientType.DEDICATED) {
-            // Use a realm-aware dedicated zk client
-            zkClient = DedicatedZkClientFactory.getInstance()
-                .buildZkClient(connectionConfig, clientConfig);
-          } else if (_zkClientType == ZkClientType.SHARED) {
-            // Use a realm-aware shared zk client
-            zkClient =
-                SharedZkClientFactory.getInstance().buildZkClient(connectionConfig, clientConfig);
-          } else {
-            zkClient = new FederatedZkClient(connectionConfig, clientConfig);
-          }
-        } catch (IOException | InvalidRoutingDataException | IllegalStateException e) {
+          zkClient = new FederatedZkClient(connectionConfig, clientConfig);
+        } catch (IOException | InvalidRoutingDataException e) {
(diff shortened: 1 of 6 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/GenericBaseDataAccessorBuilder.java
```diff
@@ -123,23 +124,26 @@ protected RealmAwareZkClient createZkClient(RealmAwareZkClient.RealmMode realmMo
...
    * @param zkClientType
    * @param realmMode
    */
   private void validateZkClientType(ZkClientType zkClientType,
       RealmAwareZkClient.RealmMode realmMode) {
-    boolean isZkClientTypeSet = zkClientType != null;
-    // If ZkClientType is set, RealmMode must either be single-realm or not set.
-    if (isZkClientTypeSet && realmMode == RealmAwareZkClient.RealmMode.MULTI_REALM) {
-      throw new HelixException("ZkClientType cannot be set on multi-realm mode!");
+    if (realmMode == null) {
```
dasahcc, line 134: Do we have some default mode for it? Shall we make it default behavior as single realm?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/ZkClientType.java
```diff
@@ -25,18 +25,24 @@
    * creation, callback functionality, and session management. But note that this is more
    * resource-heavy since it creates a dedicated ZK connection so should be used sparingly only
    * when the aforementioned features are needed.
+   *
+   * Valid on SINGLE_REALM only.
    */
   DEDICATED,
```
jiajunwang, line 31: In this case, shall we merge the realm option into the enum item's fields? Then we don't need to check since the user cannot configure this option.

narendly: @jiajunwang 
The user does have the ability to select the ZkClientType. Before ZooScalability, we allowed users to select the ZkClientType (Dedicated or Shared) when creating a BaseDataAccessor. We are doing this to keep things backward-compatible. See the following examples (copy-pasted from the test cases added as part of this PR) below:

 ```
   // Create ZkBaseDataAccessor, multi-realm, dedicated ZkClient (should fail)
    try {
      accessor = zkBaseDataAccessorBuilder.setRealmMode(RealmAwareZkClient.RealmMode.MULTI_REALM)
          .setZkClientType(ZkClientType.DEDICATED).setZkAddress(firstZkAddress)
          .setRealmAwareZkConnectionConfig(connectionConfigBuilder.build()).build();
      Assert.fail("MULTI_REALM and DEDICATED ZkClientType are an invalid combination!");
    } catch (HelixException e) {
      // Expected
    }

     // Create ZkBaseDataAccessor, multi-realm, shared ZkClient (should fail)
    try {
      accessor = zkBaseDataAccessorBuilder.setRealmMode(RealmAwareZkClient.RealmMode.MULTI_REALM)
          .setZkClientType(ZkClientType.SHARED).setZkAddress(firstZkAddress)
          .setRealmAwareZkConnectionConfig(connectionConfigBuilder.build()).build();
      Assert.fail("MULTI_REALM and SHARED ZkClientType are an invalid combination!");
    } catch (HelixException e) {
      // Expected
    }

     // Create ZkBaseDataAccessor, multi-realm, federated ZkClient, ZkAddress set (should fail)
    try {
      accessor = zkBaseDataAccessorBuilder.setRealmMode(RealmAwareZkClient.RealmMode.MULTI_REALM)
          .setZkClientType(ZkClientType.FEDERATED).setZkAddress(firstZkAddress)
          .setRealmAwareZkConnectionConfig(connectionConfigBuilder.build()).build();
      Assert.fail("MULTI_REALM and FEDERATED ZkClientType do not connect to one ZK!");
    } catch (HelixException e) {
      // Expected
    }


```

Comment (narendly) [reply zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/ZkClientType.java:31]: So to answer your question, because the user may want to choose which ZkClientType to use, it's appropriate that we leave this enum as a separate public enum.

Comment (jiajunwang) [reply zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/ZkClientType.java:31]: > The user does have the ability to select the ZkClientType.

Sure. My question is that once a user selects a certain type, do they have to specify a RealmMode in addition? Since based on your comment, it seems to be pure overhead. Because if they select DEDICATED, then the real mode has to be SINGLE_REALM.
I know this is for backward compatibility, so shall we just fill the RealmMode for the user when they configure the client using ZkClientType? In this case, we don't need the validation logic.

Comment (narendly) [reply zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/ZkClientType.java:31]: @jiajunwang 

That is actually something I've considered and the original implementation was somewhat doing that. But I realized that this type of "backfill" or doing this under the surface was more confusing to the user and adds more difficulty in the usage of the Builder API. We should make sure we require the users to explicitly set the types/configs, otherwise we run the risk of doing things without the user knowing and confusing them.

Comment (jiajunwang) [reply zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/ZkClientType.java:31]: So I have several questions after reviewing more code. Maybe you would like to clarify the plan?

1. The default builder build() method can actually create a client with MultiRealm enabled for DEDICATED client. However, we disable this option by enforcing the validation logic. Was it intentionally done?

2. If this is the case, then the actual build() code can be simplified, right? Since half of the branches won't be called forever.

3. 
Regarding what you replied,

> more confusing to the user and adds more difficulty in the usage of the Builder API

I guess what you meant is that if there are 2 ways of configuring the client and these 2 options are all available in one builder, it would be confusing that which one to use.

I think we discussed this topic before. To make it clearer, my suggestion would be,
1. The new builder does not allow ZkClientType config. Assume this is for backward compatibility. So we should not introduce to the new APIs.
2. For the older APIs (deprecated ones), as long as the type is passed in, we auto-fill the RealmMode based on the type. Since you mentioned in the comment that the mapping is one to one, this is doable.
3. In this case, there won't be a confusing part, IMO. And we won't need these validation logics everywhere. This helps with future maintenance work greatly.

Not sure if you agree, please feel free to schedule a discussion. Maybe I missed some usage.

BTW, if you are concerned about the already published version, then we can deprecate the ZkClientType setting method in the builders to suggest users not using it for now. The validation part can be left in the code until we make the API clean later.

Comment (narendly) [reply zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/ZkClientType.java:31]: 1. I think you're referring to 

```
      case MULTI_REALM:
        try {
          if (_zkClientType == ZkClientType.DEDICATED) {
            // Use a realm-aware dedicated zk client
            zkClient = DedicatedZkClientFactory.getInstance()
                .buildZkClient(connectionConfig, clientConfig);
          } else if (_zkClientType == ZkClientType.SHARED) {
            // Use a realm-aware shared zk client
            zkClient =
                SharedZkClientFactory.getInstance().buildZkClient(connectionConfig, clientConfig);
          } else {
            zkClient = new FederatedZkClient(connectionConfig, clientConfig);
          }
        } catch (IOException | InvalidRoutingDataException | IllegalStateException e) {
          throw new HelixException("Not able to connect on multi-realm mode.", e);
        }
        break;
```

Reviewing this logic again, I believe this was a bug. Multi-realm means it has to use FederatedZkClient only because it has to be able to talk to multiple ZKs. I've fixed this. Good catch!

2. I've removed some code in the createZkClient() method.

3. We can still support ZkClientType after fixing the bug. Like you said, ZkClientType is not set, we default to SHARED (original behavior).

Comment (jiajunwang) [reply zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/ZkClientType.java:31]: Given the current situation that user still has an option to choose between dedicated/shared clients when the realm mode is single realm, I agree that we still need to keep the methods in the builder.

File: helix-core/src/main/java/org/apache/helix/manager/zk/GenericBaseDataAccessorBuilder.java [hunk 3]
```diff
@@ -82,38 +82,28 @@ protected RealmAwareZkClient createZkClient(RealmAwareZkClient.RealmMode realmMo
...
-                .buildZkClient(connectionConfig, clientConfig);
-          } else if (_zkClientType == ZkClientType.SHARED) {
-            // Use a realm-aware shared zk client
-            zkClient =
-                SharedZkClientFactory.getInstance().buildZkClient(connectionConfig, clientConfig);
-          } else {
-            zkClient = new FederatedZkClient(connectionConfig, clientConfig);
-          }
-        } catch (IOException | InvalidRoutingDataException | IllegalStateException e) {
+          zkClient = new FederatedZkClient(connectionConfig, clientConfig);
```
dasahcc, line 85: If you think this is a bug we should fix, then we need to make clear definition of what is single realm and what is multi realm. My original thought was if you are using sharding key to determine the ZK address then it is mutli realm.

Then for dedicated zkclient, and shared zkclient, they can still use sharding key for searching zk address. Then it will be meaningless for dedicated/shared Zkclient to have the sharding key and they will only accept the ZK address. This conflict with what we have changed.

narendly: > If you think this is a bug we should fix, then we need to make clear definition of what is single realm and what is multi realm. My original thought was if you are using sharding key to determine the ZK address then it is mutli realm.

"Multi-realm" means it needs to be able to talk to different realms. "Single-realm" means it only connects to one realm.
If you're using a sharding key to locate which ZK to talk to, then it's "realm-aware", not multi-realm.

> Then for dedicated zkclient, and shared zkclient, they can still use sharding key for searching zk address. Then it will be meaningless for dedicated/shared Zkclient to have the sharding key and they will only accept the ZK address. This conflict with what we have changed.

RealmAware DedicatedZkClient and SharedZkClient do not require the ZK address. They instead require the ZK path sharding key and only connect to the ZK realm that the given ZK path sharding key maps to. Again, they fall under the single-realm use case.

@dasahcc Does this clarify?

dasahcc: Sounds good. We need to clear deliver this to users. Then for the change. I am totally fine with it.

PR #1129, opened 2020-06-29
Validate instance topology configuration before let it comes online 

xyuanlu:
### Issues

#1042

### Description

Helix does not check if an instance's topology config is defined correctly when the instance is added or joins cluster. This will cause Helix to fail the rebalance. This change adds the sanity check when
1. Instance joins cluster
2. User update InstanceConfig using RestApi
3. Controller prepares the topology structure for rebalance <-- Currently we only check here

This change also centralized the logic for topology sanity check. 
There are 2 behavior change for AssignableNode.computeFaultZone. The new logic is the same as in class Topology. It is better to have a centralized util function for both places. The two behavior changes are:
1. Originally, if clusterConfig.topology is null and Instance zoneId is none, AssignableNode.computeFaultZone will use instance's name while the logic in Topology class throws exception. 
2. Originally, if clusterConfig.topology is not null but clusterConfig.faultZoneType is null, AssignableNode.computeFaultZone will throw exception while the logic in Topology class uses the last valid type in clusterConfig.topology.

This changes makes the behavior consistent and use the logic in Topology class.

### Tests

TestParticipantManager.testValidateDeltaInstanceConfigForUpdate
TestPerInstanceAccessor.simpleIntegrationTestNeg

```
Failures:                                                                                                                                                 
[ERROR] org.apache.helix.integration.TestDisablePartition.testDisableFullAuto(org.apache.helix.integration.TestDisablePartition)                                  
[ERROR]   Run 1: TestDisablePartition.testDisableFullAuto:202 expected:<OFFLINE> but was:<LEADER>                                                                 
[ERROR]   Run 2: TestDisablePartition.testDisableFullAuto:114 » ZkClient Failed to delete /Test...                                                                
[INFO]
[ERROR]   TestNoThrottleDisabledPartitions.testDisablingTopStateReplicaByDisablingInstance:98 expected:<false> but was:<true>                                     
[INFO]                                                                                                                                                            
[ERROR] Tests run: 1162, Failures: 4, Errors: 0, Skipped: 1
[INFO]                                                                                                                    

[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 29.329 s - in org.apache.helix.controller.changedetector.TestResourceChangeDetector
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 20.066 s - in org.apache.helix.integration.TestDisablePartition                
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 26.942 s - in org.apache.helix.integration.TestNoThrottleDisabledPartitions
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 11.773 s - in org.apache.helix.integration.controller.TestControllerLeadershipChange
```

### Commits

### Documentation (Optional)

NA

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
@@ -981,10 +981,6 @@ public class ConfigAccessor {
     if (overwrite) {
       ZKUtil.createOrReplace(_zkClient, zkPath, newInstanceConfig.getRecord(), true);
     } else {
-      /* If overwrite is false, the new records will be appended to fields in current instanceConfig.
-       * We only need to do the check if Domain or ZoneId is updated to new value.
-       */
-      //validateTopologyFieldsInInstanceConfig(clusterName, newInstanceConfig.getRecord().getSimpleFields());
       ZKUtil.createOrUpdate(_zkClient, zkPath, newInstanceConfig.getRecord(), true, true);
     }
   }
@@ -1017,22 +1013,16 @@ public class ConfigAccessor {
 
   /**
    * Validate if the topology related settings (Domain or ZoneId) in the given instanceConfig
-   * is valid and aligns with current clusterConfig.
-   * This function should be called when instance added to cluster or caller updates instanceConfig
-   * by overwrites the current config with a new one.
+   * are valid and align with current clusterConfig.
+   * This function should be called when instance added to cluster or caller updates instanceConfig.
    *
-   * @param
-   * @param instanceName
-   * @param instanceConfig
    * @throws IllegalArgumentException
    */
   public static boolean validateTopologySettingInInstanceConfig(ClusterConfig clusterConfig,
       String instanceName, InstanceConfig instanceConfig) throws IllegalArgumentException{
-    //ClusterConfig clusterConfig = getClusterConfig(clusterName);
+    //IllegalArgumentException will be thrown here if the input is not valid.
     Topology.computeInstanceTopologyMap(clusterConfig, instanceName, instanceConfig,
         false /*earlyQuitForFaultZone*/);
     return true;
   }
-
-
 }
--- a/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
@@ -20,14 +20,17 @@ package org.apache.helix;
  */
 
 import java.io.IOException;
+import java.security.InvalidParameterException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 
+import org.apache.helix.controller.rebalancer.topology.Topology;
 import org.apache.helix.manager.zk.GenericZkHelixApiBuilder;
 import org.apache.helix.manager.zk.ZKUtil;
 import org.apache.helix.model.ClusterConfig;
@@ -369,7 +372,6 @@ public class ConfigAccessor {
     } else {
       update.setMapField(mapKey, keyValueMap);
     }
-
     ZKUtil.createOrMerge(_zkClient, zkPath, update, true, true);
   }
 
@@ -945,23 +947,23 @@ public class ConfigAccessor {
    * replaced with the value of the same field in given config if it presents. If there is new field
    * in given config but not in current config, the field will be added into the current config..
    * The list fields and map fields will be replaced as a single entry.
-   * The current Cluster config will be replaced with the given clusterConfig. WARNING: This is not
+   * The current instanceConfig will be replaced with the given instanceConfig. WARNING: This is not
    * thread-safe or concurrent updates safe.
    * *
    *
    * @param clusterName
    * @param instanceName
-   * @param instanceConfig
+   * @param newInstanceConfig
    *
    * @return
    */
   public void updateInstanceConfig(String clusterName, String instanceName,
-      InstanceConfig instanceConfig) {
-    updateInstanceConfig(clusterName, instanceName, instanceConfig, false);
+      InstanceConfig newInstanceConfig) {
+    updateInstanceConfig(clusterName, instanceName, newInstanceConfig, false);
   }
 
   private void updateInstanceConfig(String clusterName, String instanceName,
-      InstanceConfig instanceConfig, boolean overwrite) {
+      InstanceConfig newInstanceConfig, boolean overwrite) {
     if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
       throw new HelixException("fail to setup config. cluster: " + clusterName + " is NOT setup.");
     }
@@ -976,11 +978,14 @@ public class ConfigAccessor {
           "updateInstanceConfig failed. Given InstanceConfig does not already exist. instance: "
               + instanceName);
     }
-
     if (overwrite) {
-      ZKUtil.createOrReplace(_zkClient, zkPath, instanceConfig.getRecord(), true);
+      ZKUtil.createOrReplace(_zkClient, zkPath, newInstanceConfig.getRecord(), true);
     } else {
-      ZKUtil.createOrUpdate(_zkClient, zkPath, instanceConfig.getRecord(), true, true);
+      /* If overwrite is false, the new records will be appended to fields in current instanceConfig.
+       * We only need to do the check if Domain or ZoneId is updated to new value.
+       */
+      //validateTopologyFieldsInInstanceConfig(clusterName, newInstanceConfig.getRecord().getSimpleFields());
+      ZKUtil.createOrUpdate(_zkClient, zkPath, newInstanceConfig.getRecord(), true, true);
     }
   }
 
@@ -1009,4 +1014,25 @@ public class ConfigAccessor {
               _zkAddress), false);
     }
   }
+
+  /**
+   * Validate if the topology related settings (Domain or ZoneId) in the given instanceConfig
+   * is valid and aligns with current clusterConfig.
+   * This function should be called when instance added to cluster or caller updates instanceConfig
+   * by overwrites the current config with a new one.
+   *
+   * @param
+   * @param instanceName
+   * @param instanceConfig
+   * @throws IllegalArgumentException
+   */
+  public static boolean validateTopologySettingInInstanceConfig(ClusterConfig clusterConfig,
+      String instanceName, InstanceConfig instanceConfig) throws IllegalArgumentException{
+    //ClusterConfig clusterConfig = getClusterConfig(clusterName);
+    Topology.computeInstanceTopologyMap(clusterConfig, instanceName, instanceConfig,
+        false /*earlyQuitForFaultZone*/);
+    return true;
+  }
+
+
 }
--- a/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
@@ -20,17 +20,14 @@ package org.apache.helix;
  */
 
 import java.io.IOException;
-import java.security.InvalidParameterException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 
-import org.apache.helix.controller.rebalancer.topology.Topology;
 import org.apache.helix.manager.zk.GenericZkHelixApiBuilder;
 import org.apache.helix.manager.zk.ZKUtil;
 import org.apache.helix.model.ClusterConfig;
@@ -372,6 +369,7 @@ public class ConfigAccessor {
     } else {
       update.setMapField(mapKey, keyValueMap);
     }
+
     ZKUtil.createOrMerge(_zkClient, zkPath, update, true, true);
   }
 
@@ -963,7 +961,7 @@ public class ConfigAccessor {
   }
 
   private void updateInstanceConfig(String clusterName, String instanceName,
-      InstanceConfig newInstanceConfig, boolean overwrite) {
+      InstanceConfig instanceConfig, boolean overwrite) {
     if (!ZKUtil.isClusterSetup(clusterName, _zkClient)) {
       throw new HelixException("fail to setup config. cluster: " + clusterName + " is NOT setup.");
     }
@@ -978,10 +976,11 @@ public class ConfigAccessor {
           "updateInstanceConfig failed. Given InstanceConfig does not already exist. instance: "
               + instanceName);
     }
+
     if (overwrite) {
-      ZKUtil.createOrReplace(_zkClient, zkPath, newInstanceConfig.getRecord(), true);
+      ZKUtil.createOrReplace(_zkClient, zkPath, instanceConfig.getRecord(), true);
     } else {
-      ZKUtil.createOrUpdate(_zkClient, zkPath, newInstanceConfig.getRecord(), true, true);
+      ZKUtil.createOrUpdate(_zkClient, zkPath, instanceConfig.getRecord(), true, true);
     }
   }
 
@@ -1010,5 +1009,4 @@ public class ConfigAccessor {
               _zkAddress), false);
     }
   }
-
 }
--- a/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
@@ -953,13 +953,13 @@ public class ConfigAccessor {
    *
    * @param clusterName
    * @param instanceName
-   * @param newInstanceConfig
+   * @param instanceConfig
    *
    * @return
    */
   public void updateInstanceConfig(String clusterName, String instanceName,
-      InstanceConfig newInstanceConfig) {
-    updateInstanceConfig(clusterName, instanceName, newInstanceConfig, false);
+      InstanceConfig instanceConfig) {
+    updateInstanceConfig(clusterName, instanceName, instanceConfig, false);
   }
 
   private void updateInstanceConfig(String clusterName, String instanceName,
@@ -1011,18 +1011,4 @@ public class ConfigAccessor {
     }
   }
 
-  /**
-   * Validate if the topology related settings (Domain or ZoneId) in the given instanceConfig
-   * are valid and align with current clusterConfig.
-   * This function should be called when instance added to cluster or caller updates instanceConfig.
-   *
-   * @throws IllegalArgumentException
-   */
-  public static boolean validateTopologySettingInInstanceConfig(ClusterConfig clusterConfig,
-      String instanceName, InstanceConfig instanceConfig) throws IllegalArgumentException{
-    //IllegalArgumentException will be thrown here if the input is not valid.
-    Topology.computeInstanceTopologyMap(clusterConfig, instanceName, instanceConfig,
-        false /*earlyQuitForFaultZone*/);
-    return true;
-  }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
@@ -247,17 +247,11 @@ public class Topology {
   }
 
   /**
-   * This function returns a LinkedHashMap<String, String> object representing
-   * the topology path for an instance.
-   * LinkedHashMap is used here since the order of the path needs to be preserved
-   * when creating the topology tree.
-   *
    * @param clusterTopologyKeyDefaultValue  a LinkedHashMap where keys are cluster topology path and
    *                                       values are their corresponding default value. The entries
    *                                        are ordered by ClusterConfig.topology setting.
    * @param faultZoneForEarlyQuit   this flag is set to true only if caller wants the path
    *                                to faultZone instead the whole path for the instance.
-   * @return an LinkedHashMap object representing the topology path for the input instance.
    */
   private static LinkedHashMap<String, String> computeInstanceTopologyMapHelper(
       boolean isTopologyAwareEnabled, String instanceName, InstanceConfig instanceConfig,
@@ -317,6 +311,19 @@ public class Topology {
     return instanceTopologyMap;
   }
 
+  /**
+   * This function returns a LinkedHashMap<String, String> object representing
+   * the topology path for an instance.
+   * LinkedHashMap is used here since the order of the path needs to be preserved
+   * when creating the topology tree.
+   *
+   * @param clusterConfig         clusterConfig of the given cluster.
+   * @param instanceName          name of the instance.
+   * @param instanceConfig        instanceConfig to be checked.
+   * @param earlyQuitForFaultZone Set to true if we only need the path till faultZone.
+   *
+   * @return an LinkedHashMap object representing the topology path for the input instance.
+   */
   public static LinkedHashMap<String, String> computeInstanceTopologyMap(
       ClusterConfig clusterConfig, String instanceName, InstanceConfig instanceConfig,
       boolean earlyQuitForFaultZone) throws IllegalArgumentException {
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
@@ -23,9 +23,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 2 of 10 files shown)


helix-rest/src/test/java/org/apache/helix/rest/server/TestPerInstanceAccessor.java
```diff
@@ -463,5 +464,53 @@ public void testValidateWeightForInstance()
...
+    // set domain to an invalid value
+    record.getSimpleFields()
+        .put(InstanceConfig.InstanceConfigProperty.DOMAIN.name(), "InvalidDomainValue");
+    entity =
+        Entity.entity(OBJECT_MAPPER.writeValueAsString(record), MediaType.APPLICATION_JSON_TYPE);
+    // Updating using an invalid domain value should return a non-OK response
+    new JerseyUriRequestBuilder(
+        "clusters/{}/instances/{}/configs?command=update&doSanityCheck=true")
+        .expectedReturnStatusCode(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode())
+        .format(CLUSTER_NAME, INSTANCE_NAME).post(this, entity);
```
kaisun2000, line 512: No validation of post results?

xyuanlu: TFTR.
Result is validated at line 511.

helix-rest/src/test/java/org/apache/helix/rest/server/TestPerInstanceAccessor.java
```diff
@@ -463,5 +464,53 @@ public void testValidateWeightForInstance()
     // Must have the results saying they are all valid (true) because capacity keys are set
     // in ClusterConfig
     node.iterator().forEachRemaining(child -> Assert.assertTrue(child.getBooleanValue()));
+    System.out.println("End test :" + TestHelper.getTestMethodName());
+  }
+
+  @Test(dependsOnMethods = "testValidateWeightForInstance")
```
kaisun2000, line 470: If I understand correctly, we are testing changing instance config domain field via rest. Here, the validation of config is actually executed from rest server side, right?

Can we add a small comment about the purpose of this test?

helix-rest/src/test/java/org/apache/helix/rest/server/TestPerInstanceAccessor.java
kaisun2000: I see.
`expectedReturnStatusCode` if get not expected code, what will happen? Did it throw an exception?

Basically, will the 'mvn test' report the test as failure as the `Assert` failure will do?

xyuanlu: Yes it will.

helix-rest/src/test/java/org/apache/helix/rest/server/TestPerInstanceAccessor.java
xyuanlu: This is correct. I will update.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PerInstanceAccessor.java
```diff
@@ -542,4 +558,23 @@ public Response getHealthReportsOnInstance(
...
+    if (isDelete) {
+      for (Map.Entry<String, String> entry : newInstanceConfig.getRecord().getSimpleFields()
+          .entrySet()) {
+        originalInstanceConfigCopy.getRecord().getSimpleFields().remove(entry.getKey());
+      }
+    } else {
+      originalInstanceConfigCopy.getRecord().update(newInstanceConfig.getRecord());
+    }
+
+    return ConfigAccessor
```
dasahcc, line 576: Is this sanity check only for topology? If yes, let's name it as topoogy check for the perInstanceAccessor, because the sanity check should also involve the "HELIX_PORT" something like that in the simple field.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -945,23 +947,23 @@ public void setInstanceConfig(String clusterName, String instanceName,
...
    * @param instanceName
-   * @param instanceConfig
+   * @param newInstanceConfig
    *
    * @return
    */
   public void updateInstanceConfig(String clusterName, String instanceName,
-      InstanceConfig instanceConfig) {
-    updateInstanceConfig(clusterName, instanceName, instanceConfig, false);
+      InstanceConfig newInstanceConfig) {
```
lei-xia, line 961: It should be okey to leave it "instanceConfig" instead of "newInstanceConfig", otherwise, you need to rename this parameter in all other "set" or "update" related methods to keep them consistent, which I do not recommend though.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -1009,4 +1010,19 @@ public ConfigAccessor build() {
...
   }
+
+  /**
+   * Validate if the topology related settings (Domain or ZoneId) in the given instanceConfig
+   * are valid and align with current clusterConfig.
+   * This function should be called when instance added to cluster or caller updates instanceConfig.
+   *
+   * @throws IllegalArgumentException
+   */
+  public static boolean validateTopologySettingInInstanceConfig(ClusterConfig clusterConfig,
```
lei-xia, line 1021: Seems this method is not used by ConfigAccessor itself?  My suggestion is put this method into InstanceConfig (or ClusterConfig) and it does not have to be static (it can be called like myInstanceConfig.validateTopologySetting(String clusterTopologyString).

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -43,11 +43,11 @@
  */
 public class Topology {
   private static Logger logger = LoggerFactory.getLogger(Topology.class);
+
   public enum Types {
-    ROOT,
-    ZONE,
-    INSTANCE
+    ROOT, ZONE, INSTANCE
```
lei-xia, line 48: Is this our new style to format a enum types?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -187,26 +187,31 @@ private void joinCluster() {
...
+        String domain = cloudInstanceInformation
+            .get(CloudInstanceInformation.CloudInstanceField.FAULT_DOMAIN.name()) + _instanceName;
+        instanceConfig = HelixUtil.composeInstanceConfig(_instanceName);
+        instanceConfig.setDomain(domain);
       }
+      _helixAdmin.addInstance(_clusterName, instanceConfig);
+    } else {
+      ConfigAccessor.validateTopologySettingInInstanceConfig(
+          _configAccessor.getClusterConfig(_clusterName), _instanceName,
+          _configAccessor.getInstanceConfig(_clusterName,_instanceName));
```
lei-xia, line 211: Should this check before line 207?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PerInstanceAccessor.java
```diff
@@ -287,6 +288,7 @@ public Response getInstanceConfig(@PathParam("clusterId") String clusterId,
   @Path("configs")
   public Response updateInstanceConfig(@PathParam("clusterId") String clusterId,
       @PathParam("instanceName") String instanceName, @QueryParam("command") String commandStr,
+      @DefaultValue("false") @QueryParam("doSanityCheck") boolean doSanityCheck,
```
lei-xia, line 291: My suggestion is to set doSanityCheck's default value being true, i.e, we should always validate the instance config unless it is explicitly override to not doing so.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PerInstanceAccessor.java
```diff
@@ -308,19 +310,33 @@ public Response updateInstanceConfig(@PathParam("clusterId") String clusterId,
...
+
+    if (doSanityCheck && (command == Command.delete || command == Command.update)) {
+      try {
+        validateDeltaInstanceConfigForUpdate(clusterId, instanceName, configAccessor,
+            instanceConfig, command == Command.delete);
+      } catch (IllegalArgumentException ex) {
+        LOG.error(
+            String.format("Error in update instance config for instance: %s", instanceName),
+            ex);
+        return serverError(ex);
```
lei-xia, line 322: If possible, let us return some useful error information to the caller, for example, where is the format error, such as "Invalid topology setting", etc.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -230,16 +220,57 @@ private boolean isInstanceEnabled(ClusterConfig clusterConfig, String instanceNa
...
+   *                                  topology keys. The set will remain empty if topology aware is
+   *                                  not enabled or this cluster uses zone instead of domains.
+   * @param defaultDomainPathValues   out parameter. Map to be populated for all default path keys.
+   *                                  The map will remain empty if topology aware is not enabled or
+   *                                  this cluster uses zone instead of domains.
+   * @return lastValidType in clusterConfig.topology
+   */
+  private static String populateClusterTopologySetting(ClusterConfig clusterConfig,
+      LinkedHashSet<String> clusterTopologyKeys /*OUT*/,
+      Map<String, String> defaultDomainPathValues /*OUT*/) {
```
lei-xia, line 236: Using input parameters to populate output values is generally not a best practice, maybe fine given this is just private method. But let us see if we find a better way to return these values.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: I think auto formatter changes this..

helix-rest/src/test/java/org/apache/helix/rest/server/TestPerInstanceAccessor.java
xyuanlu: Updated.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PerInstanceAccessor.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
xyuanlu: Updated,

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
xyuanlu: The InstanceConfig is autogenerated for that branch before line 207. Do we also need to check that as well?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PerInstanceAccessor.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -204,8 +172,9 @@ private Node createClusterTree() {
       InstanceConfig insConfig = _instanceConfigMap.get(instanceName);
       try {
         LinkedHashMap<String, String> instanceTopologyMap =
-            computeInstanceTopologyMap(_clusterConfig.isTopologyAwareEnabled(), instanceName,
-                insConfig, _clusterTopologyKeys);
+            computeInstanceTopologyMapHelper(_clusterConfig.isTopologyAwareEnabled(), instanceName,
```
NealSun96, line 175: I believe you meant to call `computeInstanceTopologyMap()` with a false in the end here, instead of the concrete function itself?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -230,20 +199,73 @@ private boolean isInstanceEnabled(ClusterConfig clusterConfig, String instanceNa
...
   /**
    * This function returns a LinkedHashMap<String, String> object representing
    * the topology path for an instance.
    * LinkedHashMap is used here since the order of the path needs to be preserved
    * when creating the topology tree.
    *
+   * @param clusterTopologyKeyDefaultValue  a LinkedHashMap where keys are cluster topology path and
+   *                                       values are their corresponding default value. The entries
+   *                                        are ordered by ClusterConfig.topology setting.
+   * @param faultZoneForEarlyQuit   this flag is set to true only if caller wants the path
```
NealSun96, line 258: This comment should be for `computeInstanceTopologyMap()`? That public function needs an entire javadoc also.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: TFTR. I think computeInstanceTopologyMap for users outside this function. They need to call 'computeInstanceTopologyMap' to construct the 'clusterTopologyConfig' (line 323) first and then build the instance topology path. Here in Topology it self, the 'clusterTopologyConfig' is computed in the constructor. So we need another layer of abstraction. 
I am open to discussion.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
NealSun96: I see. Typically if a concrete function involves a null value, such as this case, it usually isn't called directly by outside callers and is only called by its wrapper. I'm not 100% certain with the convention here so I'll leave it to you to decide. Thank you for explaining!

helix-core/src/main/java/org/apache/helix/model/InstanceConfig.java
```diff
@@ -637,4 +638,19 @@ public static InstanceConfig toInstanceConfig(String instanceId) {
...
+
+  /**
+   * Validate if the topology related settings (Domain or ZoneId) in the given instanceConfig
+   * are valid and align with current clusterConfig.
+   * This function should be called when instance added to cluster or caller updates instanceConfig.
+   *
+   * @throws IllegalArgumentException
+   */
+  public boolean validateTopologySettingInInstanceConfig(ClusterConfig clusterConfig,
+      String instanceName) throws IllegalArgumentException{
```
pkuwm, line 650: Usually runtime exception is not recommended to put in method signature. Having it in java doc is good enough.

helix-core/src/test/java/org/apache/helix/integration/manager/TestParticipantManager.java
```diff
@@ -114,6 +118,50 @@ public void simpleIntegrationTest() throws Exception {
...
+
+
+    String instanceName = "localhost_12918";
+    HelixManager participant =
+        new ZKHelixManager(clusterName, instanceName , InstanceType.PARTICIPANT, ZK_ADDR);
+    participant.getStateMachineEngine().registerStateModelFactory("MasterSlave",
+        new MockMSModelFactory());
+    // We are expecting an IllegalArgumentException since the domain is not set.
+    try {
+      participant.connect();
```
pkuwm, line 151: Assert fail here to protect the test. Otherwise if expecting exception is not throw, the test will still pass. It should fail.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PerInstanceAccessor.java
```diff
@@ -308,20 +309,34 @@ public Response updateInstanceConfig(@PathParam("clusterId") String clusterId,
...
     InstanceConfig instanceConfig = new InstanceConfig(record);
     ConfigAccessor configAccessor = getConfigAccessor();
+
     try {
+      /*
+       * Even if the instance is disabled, non-valid instance topology config will cause rebalance
+       * failure. We are doing the check whenever user updates InstanceConfig.
+       */
+      if (command == Command.delete || command == Command.update) {
+        validateDeltaTopologySettingInInstanceConfig(clusterId, instanceName, configAccessor,
```
pkuwm, line 319: I suggest changing the method signature a bit: `boolean isDelete` -> `Command command`. You could put if..else check in this private method.
An nice article to help with clean such boolean parameter : https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PerInstanceAccessor.java
```diff
@@ -308,20 +309,34 @@ public Response updateInstanceConfig(@PathParam("clusterId") String clusterId,
...
+          HelixConfigScope instanceScope =
+              new HelixConfigScopeBuilder(HelixConfigScope.ConfigScopeProperty.PARTICIPANT)
+                  .forCluster(clusterId).forParticipant(instanceName).build();
+          configAccessor.remove(instanceScope, record);
+          break;
+        default:
+          return badRequest(String.format("Unsupported command: %s", command));
       }
+    } catch (IllegalArgumentException ex) {
+      LOG.error(String.format("Invalid topology setting for Instance : %s. Fail the config update",
```
pkuwm, line 337: Could make advantage of parameterized logging :)
`LOG.error("Invalid topology setting for Instance : {}", instanceName, ex);`

helix-core/src/main/java/org/apache/helix/ConfigAccessor.java
```diff
@@ -20,14 +20,17 @@
  */
 
 import java.io.IOException;
+import java.security.InvalidParameterException;
```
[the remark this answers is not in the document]
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -289,6 +310,29 @@ private boolean isInstanceEnabled(ClusterConfig clusterConfig, String instanceNa
...
+   * @param clusterConfig         clusterConfig of the given cluster.
+   * @param instanceName          name of the instance.
+   * @param instanceConfig        instanceConfig to be checked.
+   * @param earlyQuitForFaultZone Set to true if we only need the path till faultZone.
+   *
+   * @return an LinkedHashMap object representing the topology path for the input instance.
+   */
+  public static LinkedHashMap<String, String> computeInstanceTopologyMap(
+      ClusterConfig clusterConfig, String instanceName, InstanceConfig instanceConfig,
+      boolean earlyQuitForFaultZone) throws IllegalArgumentException {
```
pkuwm, line 328: I suggest not throwing `IllegalArgumentException` in the method signature because: an unchecked exception does not need to declare in the method signature. `throws` keyword is used for handling checked exception. There are some good articles about this, eg: https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#throwstag

I suggest put this info in javadoc like this 
```
@throws IllegalArugmentException  when it is thrown
```

xyuanlu: TFTR. Updated.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PerInstanceAccessor.java
dasahcc, line 319: Why we dont call this in the switch case? It will make the code more clear. By doing that, you can let this validateDeltaTopologySettingInInstanceConfig, return a Config if it is valid. Otherwise, return null. 

So the following code will based on the return result to do the write to ZK.

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -187,26 +187,28 @@ private void joinCluster() {
...
+        CloudInstanceInformation cloudInstanceInformation = getCloudInstanceInformation();
+        String domain = cloudInstanceInformation
+            .get(CloudInstanceInformation.CloudInstanceField.FAULT_DOMAIN.name()) + _instanceName;
+        instanceConfig = HelixUtil.composeInstanceConfig(_instanceName);
+        instanceConfig.setDomain(domain);
       }
+      _helixAdmin.addInstance(_clusterName, instanceConfig);
     }
+    _configAccessor.getInstanceConfig(_clusterName, _instanceName)
+        .validateTopologySettingInInstanceConfig(_configAccessor.getClusterConfig(_clusterName),
```
dasahcc, line 210: Why we dont validate it before we write to ZK?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -187,25 +187,31 @@ private void joinCluster() {
...
-        }
+        LOG.info(_instanceName + " is auto-registering cluster: " + _clusterName);
+        CloudInstanceInformation cloudInstanceInformation = getCloudInstanceInformation();
+        String domain = cloudInstanceInformation
+            .get(CloudInstanceInformation.CloudInstanceField.FAULT_DOMAIN.name()) + _instanceName;
+        instanceConfig = HelixUtil.composeInstanceConfig(_instanceName);
+        instanceConfig.setDomain(domain);
       }
+      instanceConfig
+          .validateTopologySettingInInstanceConfig(_configAccessor.getClusterConfig(_clusterName),
```
zhangmeng916, line 208: If the autoRegistration is false, but the autoJoin is true, then the instance config will be populated without zone information. In this case, will the validation fail? If it fails, meaning that we don't allow customers to use autoJoin if it's rack aware. Just make sure this is the behavior you want.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PerInstanceAccessor.java
```diff
@@ -310,20 +311,34 @@ public Response updateInstanceConfig(@PathParam("clusterId") String clusterId,
...
+           * The new instanceConfig will be merged with existing one.
+           * Even if the instance is disabled, non-valid instance topology config will cause rebalance
+           * failure. We are doing the check whenever user updates InstanceConfig.
+           */
+          validateDeltaTopologySettingInInstanceConfig(clusterId, instanceName, configAccessor,
+              instanceConfig, command);
+          configAccessor.updateInstanceConfig(clusterId, instanceName, instanceConfig);
+          break;
+        case delete:
+          validateDeltaTopologySettingInInstanceConfig(clusterId, instanceName, configAccessor,
```
dasahcc, line 328: Do we need a if statement here?

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
zhangmeng916: Discussed offline. This validation here means that we won't support a customer to use zone based topology and "auto join" function (no auto registration) at the same time. Participant will terminate in this case. This is different from current requirement, but logically correct.

xyuanlu: Thanks for the update Meng. I would like to add a point. 
I think if the autoRegistration is false, we only support Instance autoJoin if cluster level topology aware rebalance is set up false. If the cluster level topology aware is not false, then the following validation will fail (and vice versa).

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/PerInstanceAccessor.java
xyuanlu: validateDeltaTopologySettingInInstanceConfig will throw exception if the input is not valid. So If I understand it correctly, no need to have a if statement here. Maybe a comment is needed to avoid confusion?

dasahcc: Then let's modify the function definition and internal logic. It returns boolean value will be useless.

xyuanlu: TFTR. Updated.

xyuanlu:

Validate instance topology configuration before let it comes online

Helix does not check if an instance's topology config is defined correctly when the instance is added or joins cluster. This will cause Helix to fail the rebalance. This change adds the sanity check when
- Instance joins cluster as liveInstance
- User update InstanceConfig using RestApi
- Controller prepares the topology structure for rebalance <-- Currently we only check here
This change also did come code clean up and centralized the logic for topology sanity check.

pkuwm:
@xyuanlu Since the CI test didn't finish, could you help run the tests and after that I'll help merge? Thanks.

PR #1142, opened 2020-07-07
Respect Maximum Number Of Attempts for the tasks

### LINKED ISSUE #957: MaxAttemptsPerTask is not being respected in several scenarios
In the scenario of targeted tasks, when master moves to new instance, task should correctly move to new instance and number of attempts for that task should increase as well. However, in this scenario, the MaxAttemptsPerTask is not being respected and tasks gets assigned to new instance regardless of MaxAttemptsPerTask.

### LINKED ISSUE #1141: Possible quota leak for running tasks
In the Task Framework logic, If a task is running and job goes to TIMED_OUT, FAILING, FAILED and ABORTED state, controller will send RUNNIG to DROPPED message. However, this is not desirable because task need to go to ABORTED state, the quota needs to be release by controller and then the task can be DROPPED. Also the same issue can cause the job the job be stuck in FAILING state because context of the task will be always in running state and the job will never go to FAILED state.

### LINKED ISSUE #1143: _existsLiveInstanceOrCurrentStateChange flag in cache
There is flag in the cache which is called _existsLiveInstanceOrCurrentStateChange which is mainly used by Task Framework logic. This flag is used for the TF logic to check if task's target partition has been moved to new instance or not. However, there is a possibility of the race condition here. Since CurrentState and Message are existed in two different folders and updated separately, if cache refresh happens in between, we might loose the notification once the target partition has moved. (This theory has been proved using a test and probably this can be a reason for some existing flaky tests for targeted jobs). Also if currentState is changed and we still have a pending message for partition, we do not make any decision for this partition/task. To resolve this issue, we might want to also consider the message change for this flag as well. So the code can be something like this:
```
  private void refreshClusterStateChangeFlags(Set<HelixConstants.ChangeType> propertyRefreshed) {
    _existsLiveInstanceOrCurrentStateChange =
        _propertyDataChangedMap.get(HelixConstants.ChangeType.CURRENT_STATE).getAndSet(false)
            || _propertyDataChangedMap.get(HelixConstants.ChangeType.MESSAGE).getAndSet(false)
            || propertyRefreshed.contains(HelixConstants.ChangeType.CURRENT_STATE)
            || propertyRefreshed.contains(HelixConstants.ChangeType.LIVE_INSTANCE);
  }
```

alirezazamani:
### Issues
Fixes #957 
Fixes #1141
Fixes #1143  

### Description

In this PR, several scheduling parts have been changed which enforces the scheduler to respect maximum number of attempts for the tasks.

Also, it has been observed that when a task being dropped and scheduled again, max number of attempts is not being respected. In this PR, further checks is added to not schedule the tasks again once we reach its maximum number of attempts.

Moreover, there is race condition for _existsLiveInstanceOrCurrentStateChange flag which has been fixed in this PR.

Note: Several of the tests have been changed. Specially the tests with strict MaxNumberOfAttempts. These tests need to be changed because they are mostly related to the targeted jobs. Once we start the participants and targeted partition bounces between the participants, the tasks needs to be reassigned and number of attempts for the task will increase. We haven't noticed it before since we haven't been respecting MaxNumberOfAttempts for the DROPPED tasks. In other word, previously controller increases the task number of attempts without actually respecting the fact that the task has reached maximum number of attempts.

### Tests
TestMaxNumberOfAttemptsMasterSwitch

helix-core:
```
[ERROR] Failures: 
[ERROR]   TestWorkflowTermination.testWorkflowPausedTimeout:170->verifyWorkflowCleanup:257 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1150, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:21 h
[INFO] Finished at: 2020-07-09T18:42:36-07:00
```
The failed test succeeded when run individually.
mvn test -Dtest="TestWorkflowTermination"
```
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 36.198 s - in org.apache.helix.integration.task.TestWorkflowTermination
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  41.610 s
[INFO] Finished at: 2020-07-09T19:06:56-07:00
[INFO] ------------------------------------------------------------------------``
```
helix-rest:
```
[INFO] Tests run: 159, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 42.166 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  47.540 s
[INFO] Finished at: 2020-07-09T19:08:19-07:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
+++ b/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
@@ -75,9 +75,9 @@ public abstract class AbstractTaskDispatcher {
       Set<Integer> skippedPartitions, WorkflowControllerDataProvider cache,
       Map<String, Set<Integer>> tasksToDrop) {
 
-    // If a job is in the following states and contains running tasks, all of the running tasks
-    // should go to the ABORTED states.
-    Set<TaskState> jobStatesForRunningTaskToAbortedState =
+    // If a job is in one of the following states and its tasks are in RUNNING states, the tasks
+    // will be aborted.
+    Set<TaskState> jobStatesForAbortingTasks =
         new HashSet<>(Arrays.asList(TaskState.TIMING_OUT, TaskState.TIMED_OUT, TaskState.FAILING,
             TaskState.FAILED, TaskState.ABORTED));
 
@@ -192,7 +192,7 @@ public abstract class AbstractTaskDispatcher {
         switch (currState) {
         case RUNNING: {
           TaskPartitionState nextState = TaskPartitionState.RUNNING;
-          if (jobStatesForRunningTaskToAbortedState.contains(jobState)) {
+          if (jobStatesForAbortingTasks.contains(jobState)) {
             nextState = TaskPartitionState.TASK_ABORTED;
           } else if (jobTgtState == TargetState.STOP) {
             nextState = TaskPartitionState.STOPPED;
@@ -753,7 +753,7 @@ public abstract class AbstractTaskDispatcher {
   // 2- Task has just gone to TIMED_OUT, ERROR or DROPPED states and has reached to its
   // maxNumberAttempts
   // These tasks determine whether the job needs to FAILED or not.
-  protected static void addGiveUpPartitions(Set<Integer> set, JobContext ctx,
+  protected static void addGivenUpPartitions(Set<Integer> set, JobContext ctx,
       Iterable<Integer> pIds, JobConfig cfg) {
     for (Integer pId : pIds) {
       if (isTaskGivenup(ctx, cfg, pId)) {
--- a/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
+++ b/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
@@ -544,8 +544,8 @@ public abstract class AbstractTaskDispatcher {
       Set<Integer> allPartitions, final long currentTime, Collection<String> liveInstances) {
 
     // See if there was LiveInstance change and cache LiveInstances from this iteration of pipeline
-    boolean existsLiveInstanceOrCurrentStateChange =
-        cache.getExistsLiveInstanceOrCurrentStateChange();
+    boolean existsLiveInstanceOrCurrentStateOrMessageChangeChange =
+        cache.getExistsLiveInstanceOrCurrentStateOrMessageChange();
 
     // The excludeSet contains the set of task partitions that must be excluded from consideration
     // when making any new assignments.
@@ -572,7 +572,8 @@ public abstract class AbstractTaskDispatcher {
     Set<Integer> partitionsToRetryOnLiveInstanceChangeForTargetedJob = new HashSet<>();
     // If the job is a targeted job, in case of live instance change, we need to assign
     // non-terminal tasks so that they could be re-scheduled
-    if (!TaskUtil.isGenericTaskJob(jobCfg) && existsLiveInstanceOrCurrentStateChange) {
+    if (!TaskUtil.isGenericTaskJob(jobCfg)
+        && existsLiveInstanceOrCurrentStateOrMessageChangeChange) {
       // This job is a targeted job, so FixedAssignmentCalculator will be used
       // There has been a live instance change. Must re-add incomplete task partitions to be
       // re-assigned and re-scheduled
@@ -608,7 +609,8 @@ public abstract class AbstractTaskDispatcher {
     }
 
     // If this is a targeted job and if there was a live instance change
-    if (!TaskUtil.isGenericTaskJob(jobCfg) && existsLiveInstanceOrCurrentStateChange) {
+    if (!TaskUtil.isGenericTaskJob(jobCfg)
+        && existsLiveInstanceOrCurrentStateOrMessageChangeChange) {
       // Drop current jobs only if they are assigned to a different instance, regardless of
       // the jobCfg.isRebalanceRunningTask() setting
       dropRebalancedRunningTasks(tgtPartitionAssignments, currentInstanceToTaskAssignments, paMap,
--- a/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
+++ b/helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
@@ -185,17 +185,13 @@ public abstract class AbstractTaskDispatcher {
         switch (currState) {
         case RUNNING: {
           TaskPartitionState nextState = TaskPartitionState.RUNNING;
-          if (jobState == TaskState.TIMING_OUT) {
+          if (jobState == TaskState.TIMING_OUT || jobState == TaskState.TIMED_OUT
+              || jobState == TaskState.FAILING || jobState == TaskState.FAILED
+              || jobState == TaskState.ABORTED) {
             nextState = TaskPartitionState.TASK_ABORTED;
           } else if (jobTgtState == TargetState.STOP) {
             nextState = TaskPartitionState.STOPPED;
-          } else if (jobState == TaskState.ABORTED || jobState == TaskState.FAILED
-              || jobState == TaskState.FAILING || jobState == TaskState.TIMED_OUT) {
-            // Drop tasks if parent job is not in progress
-            paMap.put(pId, new PartitionAssignment(instance, TaskPartitionState.DROPPED.name()));
-            break;
           }
-
           paMap.put(pId, new PartitionAssignment(instance, nextState.name()));
           assignedPartitions.get(instance).add(pId);
           if (LOG.isDebugEnabled()) {
@@ -560,7 +556,7 @@ public abstract class AbstractTaskDispatcher {
       excludeSet.addAll(assignedSet);
     }
     addCompletedTasks(excludeSet, jobCtx, allPartitions);
-    addGiveupPartitions(excludeSet, jobCtx, allPartitions, jobCfg);
+    addPartitionsReachedMaximumRetries(excludeSet, jobCtx, allPartitions, jobCfg);
     excludeSet.addAll(skippedPartitions);
     Set<Integer> partitionsWithDelay = TaskUtil.getNonReadyPartitions(jobCtx, currentTime);
     excludeSet.addAll(partitionsWithDelay);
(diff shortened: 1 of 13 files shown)


helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -185,17 +185,13 @@ public void updatePreviousAssignedTasksStatus(
         switch (currState) {
         case RUNNING: {
           TaskPartitionState nextState = TaskPartitionState.RUNNING;
-          if (jobState == TaskState.TIMING_OUT) {
+          if (jobState == TaskState.TIMING_OUT || jobState == TaskState.TIMED_OUT
+              || jobState == TaskState.FAILING || jobState == TaskState.FAILED
+              || jobState == TaskState.ABORTED) {
```
narendly, line 190: Would it be cleaner to define a set of enums (states) where we should mark the tasks aborted? Then we could do a simple exists() check. 

Or, it seems that if we check for != condition, we won't have to list out so many states here?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -560,7 +556,7 @@ protected void handleAdditionalTaskAssignment(
       excludeSet.addAll(assignedSet);
     }
     addCompletedTasks(excludeSet, jobCtx, allPartitions);
-    addGiveupPartitions(excludeSet, jobCtx, allPartitions, jobCfg);
+    addPartitionsReachedMaximumRetries(excludeSet, jobCtx, allPartitions, jobCfg);
```
narendly, line 559: Thanks for renaming!

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -745,7 +743,7 @@ protected void scheduleForNextTask(String job, JobContext jobCtx, long now) {
     }
   }
 
-  // add all partitions that have been tried maxNumberAttempts
+  // add all partitions that are given up
```
narendly, line 746: What is the distinction between "given up" and partitions that exceeded the # of maximum attempts?

helix-core/src/test/java/org/apache/helix/integration/task/TestDeleteJobFromJobQueue.java
```diff
@@ -44,10 +44,10 @@ public void testForceDeleteJobFromJobQueue() throws InterruptedException {
     // Create two jobs: job1 will complete fast, and job2 will be stuck in progress (taking a long
     // time to finish). The idea is to force-delete a stuck job (job2).
     JobConfig.Builder jobBuilder = JobConfig.Builder.fromMap(WorkflowGenerator.DEFAULT_JOB_CONFIG)
-        .setMaxAttemptsPerTask(1).setWorkflow(jobQueueName)
+        .setWorkflow(jobQueueName)
```
narendly, line 47: Could you explain further why this change is needed? If we add more nodes, the db partitions will get shifted. Are we adding nodes while the workflow is in progress and before its tasks complete? In that case, the shuffling of tasks (getting dropped and rescheduled onto another instance) makes sense and we should increment # of attempts.

Since the focus of this test doesn't seem to be # of attempts, do you think we should add a new test that tests for this? We want to make sure that for targeted tasks, if we add more nodes and if that causes db partitions to be moved, then we should observe the tasks get dropped and retried with an increase in the # of attempts. One thing to watch out for here is "rebalanceRunningTask" - what role should that parameter play?

narendly: I just realized that you did add a new test below, TestMaxNumberOfAttemptsMasterSwitch. Great job!

Now, if we could reason through how the "rebalanceRunningTask" config should (or should not) apply here, we're set :)

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
narendly: Also just to confirm, this is to make sure that task states that are stuck in RUNNING get updated to ABORTED?

alirezazamani: Done. I create a list for these states. Yes and also make sure task goes to ABORTED state and we release quota.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: Thanks for the comment. I added comments for more explanation. 
Also I noticed the given up tasks has been already added in job dispatcher. So we do not need to add the tasks again. However, we want to have a sanity check to filter all of the tasks, regarding of their states, if they reached their maximum number of attempts. Because it has been observed that sometimes, the state of the task might be INIT or DROPPED or other states (if we have pending messages) and we miss max number of attempts check for those tasks.

helix-core/src/test/java/org/apache/helix/integration/task/TestDeleteJobFromJobQueue.java
alirezazamani: I reverted this change for this test. However, we still need to increase number of attempts for some test that we stop the participant and start them again. 
For example, in the TestForceDeleteWorkflow.testDeleteCompletedWorkflowForcefully:
In super.beforeClass() we start the participant, however, once the test started, we stop those participants and start them again. I have observed that if we start the workflow very close to the time that we start the participant one by one (kind of like rolling upgrade), Master might jump several times between the nodes. Hence if we set maxNumberOfAttempts to 1, then the test will fail. Since we just started to respect maxNumberOfAttempts in this PR.

About your second comment: Yeah the newly added test checks maxNumberOfAttempts for targetJobChange. For "rebalanceRunningTask", I checked other tests and I saw we this logic is already covered in "TestRebalanceRunningTask" extensively for different scenarios. So if rebalance running task is set and we need to drop the task in one instance and run in other one (Send INIT->RUNNING), then this is new scheduling and we are respecting max number of attempts. So rebalanceRunningTask is happening until we reach maxNumberOfAttempts. In other word, once a task reaches maxNumberOfAttempts, it won't be scheduled again. So no new INIT->RUNNING once we reach maxNumberOfAttempts.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: ;-)

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -74,6 +75,12 @@ public void updatePreviousAssignedTasksStatus(
       Set<Integer> skippedPartitions, WorkflowControllerDataProvider cache,
       Map<String, Set<Integer>> tasksToDrop) {
 
+    // If a job is in the following states and contains running tasks, all of the running tasks
+    // should go to the ABORTED states.
```
narendly, line 79: Nit: 

If a job is in one of the following states and its tasks are in RUNNING states, the tasks will be aborted.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -74,6 +75,12 @@ public void updatePreviousAssignedTasksStatus(
...
+    Set<TaskState> jobStatesForRunningTaskToAbortedState =
```
narendly, line 80: I think calling it `jobStatesForAbortingTasks` does the trick and is shorter. What do you think?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -745,8 +748,12 @@ protected void scheduleForNextTask(String job, JobContext jobCtx, long now) {
...
   }
 
-  // add all partitions that have been tried maxNumberAttempts
-  protected static void addGiveupPartitions(Set<Integer> set, JobContext ctx,
+  // Add all partitions/tasks that are cannot be retried. These tasks are:
+  // 1- Task is in ABORTED or ERROR state.
+  // 2- Task has just gone to TIMED_OUT, ERROR or DROPPED states and has reached to its
+  // maxNumberAttempts
+  // These tasks determine whether the job needs to FAILED or not.
+  protected static void addGiveUpPartitions(Set<Integer> set, JobContext ctx,
```
narendly, line 756: Typo: `addGivenUpPartitions`

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: Fixed. Thanks

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: Done :-)

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: Yeah I wasn't happy with previous name as well. Change it as you suggested. Thanks.

alirezazamani:
> Please address the comments. Approving because the pending issues are minor, style-related issues. Also, please rebase from master and run the tests before marking it as ready for being merged. Great work!

@narendly Thank you for reviewing. I addressed your final comments as well. Let me do final checks and several round of testing. After that I will put final message for merging. Thanks

alirezazamani:

Final commit message:
Respect maximum number of attempts for the tasks

In this commit, several scheduling parts have been changed in order to
enforce the scheduler to respect maximum number of attempts for
the tasks.

Also, it has been observed that when a task being dropped and
scheduled again, max number of attempts is not being respected.
in this commit, further checks is added to avoid schedule the
tasks again once we reach its maximum number of attempts.

Issue #1148, opened 2020-07-14
ZKHelixManager.disconnect() will block forever if ZK connection is not SyncConnected.


jiajunwang:
If there is no ZooKeeper nor ZooKeeper connection, calling ZKHelixManager.disconnect() will block forever.

Example stack:
	  at sun.misc.Unsafe.park(Unsafe.java:-1)
	  at java.util.concurrent.locks.LockSupport.parkUntil(LockSupport.java:237)
	  at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.awaitUntil(AbstractQueuedSynchronizer.java:2072)
	  at org.I0Itec.zkclient.ZkClient.waitForKeeperState(ZkClient.java:636)
	  at org.I0Itec.zkclient.ZkClient.waitUntilConnected(ZkClient.java:619)
	  at org.I0Itec.zkclient.ZkClient.waitUntilConnected(ZkClient.java:615)
	  at org.I0Itec.zkclient.ZkClient.retryUntilConnected(ZkClient.java:679)
	  at org.apache.helix.manager.zk.ZkClient.getChildren(ZkClient.java:210)
	  at org.I0Itec.zkclient.ZkClient.getChildren(ZkClient.java:409)
	  at org.apache.helix.manager.zk.CallbackHandler.subscribeForChanges(CallbackHandler.java:293)
	  at org.apache.helix.manager.zk.CallbackHandler.invoke(CallbackHandler.java:163)
	  at org.apache.helix.manager.zk.CallbackHandler.reset(CallbackHandler.java:408)
	  at org.apache.helix.manager.zk.ZKHelixManager.resetHandlers(ZKHelixManager.java:726)
	  at org.apache.helix.manager.zk.ZKHelixManager.disconnect(ZKHelixManager.java:532)

PR #1149, opened 2020-07-14
Refine the ZkHelixManager.disconnect() method so it won't be blocked when the ZK connection breaks.

jiajunwang:
### Issues

#1148 

### Description

The issue was that if ZkHelixManager.disconnect() is called when the ZK connection is not synConnected, the disconnect() call will hang and never return. This usually blocks Helix users' code when they try to clean up and finalize after ZK connection becomes unrecoverable.
This PR introduces a temporary thread to finish the cleanup work which relies on the ZK connection (or application code). This thread will be interrupted if ZK connection breaks before or during the disconnect process.

### Tests

TestZkConnectionLost.testDisconnectWhenConnectionBreak()

helix-core

[INFO] Tests run: 1152, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,613.998 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1152, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS

helix-rest

[INFO] Tests run: 163, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 46.568 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.Timer;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import javax.management.JMException;
 
 import com.google.common.collect.Sets;
@@ -816,11 +817,12 @@ public class ZKHelixManager implements HelixManager, IZkStateListener {
        */
       _messagingService.getExecutor().shutdown();
 
-      cleanupCallbackHandlers();
-    } catch (InterruptedException e) {
-      LOG.warn(
-          "The callback handler cleanup has been interrupted. Some callback handlers might not be reset properly. Continue to finish the other HelixMananger disconnect tasks.",
-          e);
+      if (!cleanupCallbackHandlers()) {
+        LOG.warn(
+            "The callback handler cleanup has been cleanly done. "
+                + "Some callback handlers might not be reset properly. "
+                + "Continue to finish the other Helix Mananger disconnect tasks.");
+      }
     } finally {
       GenericHelixController controller = _controller;
       if (controller != null) {
@@ -859,8 +861,12 @@ public class ZKHelixManager implements HelixManager, IZkStateListener {
   /**
    * The callback handler cleanup operations that require an active ZkClient connection.
    * If ZkClient is not connected, Helix Manager shall skip the cleanup.
+   *
+   * @return true if the cleanup has been done successfully.
    */
-  private void cleanupCallbackHandlers() throws InterruptedException {
+  private boolean cleanupCallbackHandlers() {
+    AtomicBoolean cleanupDone = new AtomicBoolean(false);
+
     if (_zkclient.waitUntilConnected(_waitForConnectedTimeout, TimeUnit.MILLISECONDS)) {
       // Create a separate thread for executing cleanup task to avoid forever retry.
       Thread cleanupThread = new Thread(String
@@ -878,6 +884,8 @@ public class ZKHelixManager implements HelixManager, IZkStateListener {
           if (participantManager != null) {
             participantManager.disconnect();
           }
+
+          cleanupDone.set(true);
         }
       };
 
@@ -915,7 +923,6 @@ public class ZKHelixManager implements HelixManager, IZkStateListener {
           cleanupThread.join();
         } catch (InterruptedException ex) {
           cleanupThread.interrupt();
-          throw ex;
         }
       } finally {
         _zkclient.unsubscribeStateChanges(stateListener);
@@ -924,6 +931,8 @@ public class ZKHelixManager implements HelixManager, IZkStateListener {
       LOG.warn(
           "ZkClient is not connected to the Zookeeper. Skip the cleanup work that requires accessing Zookeeper.");
     }
+
+    return cleanupDone.get();
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


helix-core/src/test/java/org/apache/helix/integration/TestZkConnectionLost.java
```diff
@@ -103,6 +104,28 @@ public void afterClass() throws Exception {
...
+    Thread testThread = new Thread("Testing HelixManager disconnect") {
+      @Override
+      public void run() {
+        _controller.disconnect();
+        disconnected.set(true);
+      }
+    };
+    try {
+      testThread.start();
+      testThread.join(5000);
```
narendly, line 120: Instead of giving a hard-coded timeout value, do you think we could use a CountDownLatch for this so the test doesn't depend on timing?

jiajunwang: I don't think so. Maybe I can make it a longer timeout. But if the logic is not working as expected, then the thread will never finish. So I'm not sure how the CountDownLatch can help here. Maybe you can elaborate on your idea?

helix-core/src/test/java/org/apache/helix/integration/TestZkConnectionLost.java
```diff
@@ -39,15 +40,15 @@
...
-import org.apache.helix.zookeeper.impl.factory.SharedZkClientFactory;
 import org.apache.helix.task.JobConfig;
 import org.apache.helix.task.JobQueue;
 import org.apache.helix.task.TaskState;
 import org.apache.helix.task.WorkflowContext;
 import org.apache.helix.tools.ClusterSetup;
 import org.apache.helix.tools.ClusterVerifiers.BestPossibleExternalViewVerifier;
 import org.apache.helix.tools.ClusterVerifiers.ZkHelixClusterVerifier;
+import org.apache.helix.zookeeper.api.client.HelixZkClient;
+import org.apache.helix.zookeeper.impl.factory.SharedZkClientFactory;
```
narendly, line 51: Why are these imports needed?

jiajunwang: Not added in this PR. This is auto re-ordering when I apply the latest format.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -866,6 +858,83 @@ public void disconnect() {
     }
   }
 
+  /**
+   * The callback handler cleanup operations that require an active ZkClient connection.
```
dasahcc, line 862: Why this require an active zkclient? CallbackHandler are in memory object, is there anything we need to connection to ZK?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -866,6 +858,83 @@ public void disconnect() {
...
+          if (participantManager != null) {
+            participantManager.disconnect();
+          }
+
+          cleanupDone.set(true);
+        }
+      };
+
+      // Define the state listener to terminate the cleanup thread when the ZkConnection breaks.
+      IZkStateListener stateListener = new IZkStateListener() {
```
dasahcc, line 893: It is complicated to create a listener for subscription. Shall we use some simple way to see whether the connection is alive? For example, just have an exist call for the cluster since it is in HelixManager?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
jiajunwang: 1. we subscribe, and the subscription requires ZK connection.
2. Some of our Helix callback handler implementations will read ZK even for finalizing.
3. There is no way to prevent any customer-defined handler logic to access ZK for finalizing.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
jiajunwang: `just have an exist call for the cluster since it is in HelixManager?` This is already done.
However, this won't prevent if the connection breaks during the cleanup process.
For example, we have 10 callback handlers to reset, the connection breaks at the 5th one. Then the check before this cleanup work won't work. To achieve the goal, an alternative way is keeping checking the connection state in a loop. But that is a bad design compare of using the listener.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -866,6 +858,83 @@ public void disconnect() {
...
+        }
+      } finally {
+        _zkclient.unsubscribeStateChanges(stateListener);
+      }
+    } else {
+      LOG.warn(
+          "ZkClient is not connected to the Zookeeper. Skip the cleanup work that requires accessing Zookeeper.");
+    }
+
+    return cleanupDone.get();
```
dasahcc, line 935: Will the call be hanging here as well? If there is no connection, cleanupThread got interrupted, but no one set this value to be true.

jiajunwang: We are not waiting for it. It will just return whatever inside. This is a non-blocking call.

jiajunwang: The result might be false if the clean up was not done completely due to disconnection.

jiajunwang:

PR #1154, opened 2020-07-15
Add initial callback when adding routing table listener

### LINKED ISSUE #1090: Add init function when adding a routing table listener
Currently in Helix controller logic, when a listener is added, the init() function will be called to ensure the first time invoke. However, the similar logic is missing in adding routing table provider listener. The problem caused by this missing function is that Helix customers would need to manual trigger the callback in order to get the initial value or they need to wait until the first changing event arrives. We should add an init function when adding a routing table listener to make sure our customers will get a triggering in the beginning.

zhangmeng916:
### Issues

Fixes #1090 
### Description

This PR adds an initial callback when user adds a listener to Helix routing table. This initial callback will ensure uses are able to perform operations, e.g. read routing table data or some other actions, immediately after they add the listener instead of waiting for the first data change to trigger the callback.

### Tests

- [X] The following tests are modified for this issue:

TestRoutingTableProvider.java

helix-core:
[INFO]
[INFO] Results:
[INFO]
[ERROR] Failures:
[INFO]
[ERROR] Tests run: 1209, Failures: 1, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:12 h
[INFO] Finished at: 2020-09-24T14:47:28-07:00

Local test passed. 
### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
@@ -422,19 +422,19 @@ public class RoutingTableProvider
    * Add RoutingTableChangeListener with user defined context
    * @param routingTableChangeListener
    * @param context user defined context
-   * @param doInit whether to trigger the initial callback during adding listener
+   * @param isTriggerCallback whether to trigger the initial callback during adding listener
    */
   public void addRoutingTableChangeListener(
-      final RoutingTableChangeListener routingTableChangeListener, Object context, boolean doInit) {
+      final RoutingTableChangeListener routingTableChangeListener, Object context,
+      boolean isTriggerCallback) {
     _routingTableChangeListenerMap.put(routingTableChangeListener, new ListenerContext(context));
-    logger.info(
-        "Attach RoutingTableProviderChangeListener {}. The doInit value of this listener is {}",
-        routingTableChangeListener.getClass().getName(), doInit);
-    if (doInit) {
+    logger.info("Attach RoutingTableProviderChangeListener {}.",
+        routingTableChangeListener.getClass().getName(), isTriggerCallback);
+    if (isTriggerCallback) {
+      logger.info("The isTriggerCallback value for the listener is true");
       final NotificationContext periodicRefreshContext = new NotificationContext(_helixManager);
       periodicRefreshContext.setType(NotificationContext.Type.PERIODIC_REFRESH);
-      _routerUpdater.queueEvent(periodicRefreshContext, ClusterEventType.PeriodicalRebalance,
-          null);
+      _routerUpdater.queueEvent(periodicRefreshContext, ClusterEventType.PeriodicalRebalance, null);
       }
   }
 
--- a/helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
@@ -431,32 +431,11 @@ public class RoutingTableProvider
         "Attach RoutingTableProviderChangeListener {}. The doInit value of this listener is {}",
         routingTableChangeListener.getClass().getName(), doInit);
     if (doInit) {
-      if (_sourceDataTypeMap.isEmpty()) {
-        routingTableChangeListener.onRoutingTableChange(getRoutingTableSnapshot(), context);
-      } else {
-        for (PropertyType propertyType : _sourceDataTypeMap.keySet()) {
-          switch (propertyType) {
-            case EXTERNALVIEW:
-            case TARGETEXTERNALVIEW:
-            case CURRENTSTATES:
-              routingTableChangeListener
-                  .onRoutingTableChange(getRoutingTableSnapshot(propertyType), context);
-              break;
-            case CUSTOMIZEDVIEW:
-              for (String customizedStateType : _sourceDataTypeMap
-                  .getOrDefault(PropertyType.CUSTOMIZEDVIEW, Collections.emptyList())) {
-                routingTableChangeListener.onRoutingTableChange(
-                    getRoutingTableSnapshot(PropertyType.CUSTOMIZEDVIEW, customizedStateType),
-                    context);
-              }
-              break;
-            default:
-              logger.warn("Unsupported source data type: {}, stop triggering callback!",
-                  propertyType);
-          }
-        }
+      final NotificationContext periodicRefreshContext = new NotificationContext(_helixManager);
+      periodicRefreshContext.setType(NotificationContext.Type.PERIODIC_REFRESH);
+      _routerUpdater.queueEvent(periodicRefreshContext, ClusterEventType.PeriodicalRebalance,
+          null);
       }
-    }
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
@@ -294,6 +294,10 @@ public class RoutingTableProvider
    * @param sourceDataTypeMap
    */
   private void validateSourceDataTypeMap(Map<PropertyType, List<String>> sourceDataTypeMap) {
+    if (sourceDataTypeMap == null) {
+      throw new IllegalArgumentException(
+          "The sourceDataTypeMap of Routing Table Provider should not be null");
+    }
     for (PropertyType propertyType : sourceDataTypeMap.keySet()) {
       if (propertyType.equals(PropertyType.CUSTOMIZEDVIEW)
           && sourceDataTypeMap.get(propertyType).size() == 0) {
@@ -302,10 +306,10 @@ public class RoutingTableProvider
       }
       if (!propertyType.equals(PropertyType.CUSTOMIZEDVIEW)
           && sourceDataTypeMap.get(propertyType).size() != 0) {
-        logger.error("Type has been used in addition to the propertyType {} !",
-            propertyType.name());
-        throw new HelixException(
-            String.format("Type %s has been used in addition to the propertyType %s !",
+        logger
+            .error("Type has been used in addition to the propertyType {} !", propertyType.name());
+        throw new HelixException(String
+            .format("Type %s has been used in addition to the propertyType %s !",
                 sourceDataTypeMap.get(propertyType), propertyType.name()));
       }
     }
@@ -426,7 +430,7 @@ public class RoutingTableProvider
     logger.info(
         "Attach RoutingTableProviderChangeListener {}. The doInit value of this listener is {}",
         routingTableChangeListener.getClass().getName(), doInit);
-    if (doInit && _sourceDataTypeMap != null) {
+    if (doInit) {
       if (_sourceDataTypeMap.isEmpty()) {
         routingTableChangeListener.onRoutingTableChange(getRoutingTableSnapshot(), context);
       } else {
--- a/helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
@@ -423,8 +423,9 @@ public class RoutingTableProvider
   public void addRoutingTableChangeListener(
       final RoutingTableChangeListener routingTableChangeListener, Object context, boolean doInit) {
     _routingTableChangeListenerMap.put(routingTableChangeListener, new ListenerContext(context));
-    logger.info("Attach RoutingTableProviderChangeListener {}",
-        routingTableChangeListener.getClass().getName());
+    logger.info(
+        "Attach RoutingTableProviderChangeListener {}. The doInit value of this listener is {}",
+        routingTableChangeListener.getClass().getName(), doInit);
     if (doInit && _sourceDataTypeMap != null) {
       if (_sourceDataTypeMap.isEmpty()) {
         routingTableChangeListener.onRoutingTableChange(getRoutingTableSnapshot(), context);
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
```diff
@@ -413,6 +413,31 @@ public void addRoutingTableChangeListener(
     _routingTableChangeListenerMap.put(routingTableChangeListener, new ListenerContext(context));
     logger.info("Attach RoutingTableProviderChangeListener {}",
         routingTableChangeListener.getClass().getName());
+    if (_sourceDataTypeMap.isEmpty()) {
```
dasahcc, line 416: This is an initial call when adding listener. The only concern is returning empty data should be a good signal for user? This changes behavior. Some users are relying on the snapshot content. Shall we have a different API that give the option to let user choose it?

zhangmeng916: I see your concern. There're several customers that wanted this feature. However, it's possible to cause unnecessary callback to other customers. Maybe we have a boolean value in the function, and let customers to determine their usage?

dasahcc: Let's not change the API. The other customers can still use this API. But let's have another API and old API call the new API with input value as false (for first time refresh) to guarantee the backward compatible.

zhangmeng916: So the users that would like the initial callback will explicitly call the new API?

jiajunwang: FYI, in Helix Manager, adding the listener will cause the listener to get a "free" callback when the callback handler is init. So the requirement here is doing the same for router.

Given that saying, if the routing table has not been updated (maybe during the reading) then we return an empty snapshot. In this case, if the router still notifies the listener once the update is done, then we are still good. However, based on the code here, I think we might have a race condition that this notification could be sent later than the proper one. Then user would get the wrong information.
So we need a concurrency control here.

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
```diff
@@ -413,6 +413,31 @@ public void addRoutingTableChangeListener(
...
+      routingTableChangeListener.onRoutingTableChange(getRoutingTableSnapshot(), context);
+    } else {
+      for (PropertyType propertyType : _sourceDataTypeMap.keySet()) {
```
jiajunwang, line 419: As Junkai mentioned, I think we need a new interface for the listeners anyway.
The current one won't tell the provider which type the listener wants to listen to. So I think we just add a new API so as to add the listeners with the right type.
In this case, the old API needs to be deprecated and if user add their listen using this API, we shall just assign it to all the valid types.

zhangmeng916: Why would we need the type as input? We already put the type in _sourceDataTypeMap when users initialize the routing table provider, right? So we can trigger callback with required type of snapshot.

jiajunwang: For example, this constructor allows the user to create a map of types.
```
  public RoutingTableProvider(HelixManager helixManager, Map<PropertyType, List<String>> sourceDataTypeMap) {
    this(helixManager, sourceDataTypeMap, true, DEFAULT_PERIODIC_REFRESH_INTERVAL);
  }
```
But we don't have a way for the user to specify the listener for different types. So one listener will be triggered multiple times if multiple types are specified.

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
dasahcc: > So the users that would like the initial callback will explicitly call the new API?

Yes.

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
zhangmeng916: I see your point, so right now although the routing table provider only initialized with certain source data types, users will still get callback on all snapshot change, and they need to filter based on the type to see whether the change is they want. But I think this change is beyond the scope of this PR. It's always a problem after we have multiple types, all the Zookeeper change will trigger callback, if we want to change, we will need to distinguish between different listeners, and the change also needs to be made in `notifyRoutingTableChange` when we receive a change. 
For this PR only, we do trigger callback based on types, and user won't get unused callback for those types they don't care. For more general type support for callback, we'll need more changes. Thoughts?

jiajunwang: I agree this is out of this PR's scope. My main concern is that this PR adds one more place that we need to modify later. It would be great if we have only one method to trigger all onRoutingTableChange calls.

zhangmeng916: Yeah, the onRoutingTableChange is called in both add listener and whenever there is any change triggers it. These two won't be able to be merged, right?

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
```diff
@@ -403,16 +403,55 @@ public RoutingTableSnapshot getRoutingTableSnapshot(PropertyType propertyType, S
...
+   * @param context user defined context
+   * @param doInit whether to trigger the initial callback during adding listener
+   */
+  public void addRoutingTableChangeListener(
+      final RoutingTableChangeListener routingTableChangeListener, Object context, boolean doInit) {
     _routingTableChangeListenerMap.put(routingTableChangeListener, new ListenerContext(context));
     logger.info("Attach RoutingTableProviderChangeListener {}",
         routingTableChangeListener.getClass().getName());
+    if (doInit) {
+      if (_sourceDataTypeMap.isEmpty()) {
```
narendly, line 429: If `_sourceDataTypeMap ` is not properly initialized in the constructor (if this method gets called before the object is fully initialized), this might throw an NPE. I think we should add a null check here on the map. This will guard against potential concurrency issues (that would be pretty hard to debug but might happen regardless).

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
```diff
@@ -403,16 +403,55 @@ public RoutingTableSnapshot getRoutingTableSnapshot(PropertyType propertyType, S
...
+    addRoutingTableChangeListener(routingTableChangeListener, context, false);
+  }
+
+  /**
+   * Add RoutingTableChangeListener with user defined context
+   * @param routingTableChangeListener
+   * @param context user defined context
+   * @param doInit whether to trigger the initial callback during adding listener
+   */
+  public void addRoutingTableChangeListener(
```
narendly, line 423: This is a high-level question. Have we considered doing init as the default behavior? Here are the reasons I can think off the top of my head:

1. It doesn't hurt to init (what are the bad side effects of reading in the beginning? If the user wishes to add a listener, that means the user is ready to read and be notified). In other words, I can't think of any negative consequences of this behavior change or any potential backward-compatibility issues that might break things.

2. This greatly reduces verbosity in the code. Simple is better.

zhangmeng916: @dasahcc had some concerns that if users are not expecting an empty snapshot, and the initial callback provides them such, they may not have a way to handle it yet.

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
zhangmeng916: I added the null check. Just to make sure, you're concerning the race condition between add routing table listener and the routing table provider construction?

zhangmeng916: @pkuwm I think it makes sense. @narendly could you please let us know when it is possible that adding the listener happens before routing table provider is fully initialized.

narendly: You could read this article: https://stackoverflow.com/questions/24367472/is-an-object-created-first-and-then-its-constructor-executed. Pay attention to point 4 - although taking a closer look, I don't think this leaks itself as `this`. We actually see this in the commit: https://github.com/apache/helix/commit/ff0485bdb65d1051ca63ecc97c5b864aee932465

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
narendly: I'm not quite seeing that - applications should program in a way that expects some data in the snapshot, right? Their logic shouldn't try to "time it" in their callback logic since that's never predictable.

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
zhangmeng916: Thanks for the information. Didn't know this before.

pkuwm: @zhangmeng916  Then are we checking null for sourceMap in validateSourceDataTypeMap?

narendly:
@zhangmeng916 
Please follow the pull request description guidelines. It's recommended that we use the following keywords:

close
closes
closed
fix
fixes
fixed
resolve
resolves
resolved

when referring to the issue number. This is clearly outlined in our guidelines so that when the PR gets closed, the issue gets linked and closed as well. See https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
narendly: Sorry if my comment wasn't clear. What I was saying was since `RoutingTableProvider` doesn't give its own reference and doesn't implement `RoutingTableChangeListener`, we probably don't have to worry about this and we shouldn't have to add a null check here. I think it should be okay as is :)

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
zhangmeng916: let's quickly sync offline.

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
```diff
@@ -403,16 +407,35 @@ public RoutingTableSnapshot getRoutingTableSnapshot(PropertyType propertyType, S
...
+      final RoutingTableChangeListener routingTableChangeListener, Object context, boolean doInit) {
     _routingTableChangeListenerMap.put(routingTableChangeListener, new ListenerContext(context));
-    logger.info("Attach RoutingTableProviderChangeListener {}",
-        routingTableChangeListener.getClass().getName());
+    logger.info(
+        "Attach RoutingTableProviderChangeListener {}. The doInit value of this listener is {}",
```
jiajunwang, line 431: nit, we can split "The doInit value of this listener is {}" part. And move it into the "if" code block below. Might be cleaner.

jiajunwang (commented):
LGTM!
Please ensure the test passes. Currently, the test result seems to be not so good.

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
```diff
@@ -403,16 +407,35 @@ public RoutingTableSnapshot getRoutingTableSnapshot(PropertyType propertyType, S
...
+  public void addRoutingTableChangeListener(
+      final RoutingTableChangeListener routingTableChangeListener, Object context,
+      boolean isTriggerCallback) {
     _routingTableChangeListenerMap.put(routingTableChangeListener, new ListenerContext(context));
-    logger.info("Attach RoutingTableProviderChangeListener {}",
-        routingTableChangeListener.getClass().getName());
+    logger.info("Attach RoutingTableProviderChangeListener {}.",
+        routingTableChangeListener.getClass().getName(), isTriggerCallback);
+    if (isTriggerCallback) {
+      logger.info("The isTriggerCallback value for the listener is true");
```
jiajunwang, line 434: What I meant is that we can log something like "Force triggering a callback for the new listener" here. So the log does not care about the name of the parameter "isTriggerCallback".

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
```diff
@@ -403,16 +407,35 @@ public RoutingTableSnapshot getRoutingTableSnapshot(PropertyType propertyType, S
...
+   * @param isTriggerCallback whether to trigger the initial callback during adding listener
+   */
+  public void addRoutingTableChangeListener(
+      final RoutingTableChangeListener routingTableChangeListener, Object context,
+      boolean isTriggerCallback) {
     _routingTableChangeListenerMap.put(routingTableChangeListener, new ListenerContext(context));
-    logger.info("Attach RoutingTableProviderChangeListener {}",
-        routingTableChangeListener.getClass().getName());
+    logger.info("Attach RoutingTableProviderChangeListener {}.",
+        routingTableChangeListener.getClass().getName(), isTriggerCallback);
```
jiajunwang, line 432: nit, remove the isTriggerCallback that is not used in the log.

jiajunwang (approved):
LGTM! 2 minor comments.

zhangmeng916:
Final commit message:
Add initial callback when adding routing table listener.

PR #1180, opened 2020-07-27
Remove race condition in TestTaskSchedulingTwoCurrentStates

### LINKED ISSUE #1179: TestTaskSchedulingTwoCurrentStates is unstable
TestTaskSchedulingTwoCurrentStates is unstable because controller and the test modifying the IdealState at the same time. The controller follows the Full-Auto strategy and the test is trying to use Semi-Auto.

alirezazamani:
### Issues
Fixes #1179 

### Description

In this PR, TestTaskSchedulingTwoCurrentStates has been stabilized by creating SEMI-AUTO resource and update the preferenceList to avoid race condition that happens when test and controller update the preferenceList at the same time.

### Tests

helix-core:
```
[INFO] Tests run: 1155, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,800.964 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1155, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:20 h
[INFO] Finished at: 2020-07-27T14:59:27-07:00
```

helix-rest:
```
[INFO] Tests run: 163, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 43.032 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  48.283 s
[INFO] Finished at: 2020-07-27T15:07:54-07:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestTaskSchedulingTwoCurrentStates.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestTaskSchedulingTwoCurrentStates.java
@@ -51,6 +51,7 @@ import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.impl.client.ZkClient;
 import org.apache.zookeeper.data.Stat;
 import org.testng.Assert;
+import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 import com.google.common.collect.ImmutableMap;
@@ -91,6 +92,11 @@ public class TestTaskSchedulingTwoCurrentStates extends TaskTestBase {
           new TaskStateModelFactory(_participants[i], taskFactoryReg));
       _participants[i].syncStart();
     }
+  }
+
+  @AfterClass()
+  public void afterClass() throws Exception {
+    super.afterClass();
   }
 
   @Test
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestTaskSchedulingTwoCurrentStates.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestTaskSchedulingTwoCurrentStates.java
@@ -27,14 +27,10 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.helix.AccessOption;
 import org.apache.helix.HelixDataAccessor;
-import org.apache.helix.HelixManagerFactory;
-import org.apache.helix.InstanceType;
 import org.apache.helix.PropertyKey;
 import org.apache.helix.TestHelper;
 import org.apache.helix.ZkTestHelper;
 import org.apache.helix.integration.manager.MockParticipantManager;
-import org.apache.helix.manager.zk.ZKHelixDataAccessor;
-import org.apache.helix.model.ClusterConfig;
 import org.apache.helix.model.CurrentState;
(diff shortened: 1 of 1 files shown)


helix-core/src/test/java/org/apache/helix/integration/task/TestTaskSchedulingTwoCurrentStates.java
```diff
@@ -65,7 +60,7 @@
  * sent when there are two CurrentStates.
  */
 public class TestTaskSchedulingTwoCurrentStates extends TaskTestBase {
-  private static final String DATABASE = WorkflowGenerator.DEFAULT_TGT_DB;
```
kaisun2000, line 68: Is it possible the old way of using WorkflowGenerator.DEFAULT_TGT_DB would have conflict, if multiple test using the same "WorkflowGenerator.DEFAULT_TGT_DB"?

alirezazamani: The main reason is that in before class method, "TestDB" is created as Full-Auto and in this test we are trying to change it to Semi-Auto which causes race condition between controller and test updating IdealState. So if want to have Semi-Auto resource, we need to create it separately (with different name) and leverage it in this test.

kaisun2000: I think the better way to fix it is not call super.beforeClass() in line 72/77. This way, we won't create the cluster/full auto etc and participant that we need to stop immediately. 

Instead, just creates cluster, controller and participant in this before test. What is your take?

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskSchedulingTwoCurrentStates.java
```diff
@@ -65,7 +60,7 @@
...
+  private static final String DATABASE = "TestDB_" + TestHelper.getTestClassName();
```
kaisun2000, line 63: Another thing is that for all the resources created in 
```
@BeforeClass
  public void beforeClass() throws Exception {
```

We should tear them down. The participant threads, the controller thread should be stopped with syncStop. Maybe we should also clean up the cluster path etc in Zookeeper.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskSchedulingTwoCurrentStates.java
alirezazamani: Your point is somehow valid. But I still think this is better option because this way we are saving many duplicated code which will be hard to maintain and most of the code is written in the parent class. Instead of that, as what I proposed in this PR, we can separate the DATABASE name and create new one which includes this class name and will be unique. We have same behavior for most of the Task Framework related test and workflow names have been derived from class name. This way the tests is easier to maintain and more understandable.

alirezazamani:

Final commit message:
Remove race condition in TestTaskSchedulingTwoCurrentStates

In this commit, TestTaskSchedulingTwoCurrentStates has been stabilized
by creating SEMI-AUTO resource and update the preferenceList to avoid
race condition that happens when test and controller update the
preferenceList at the same time.

PR #1183, opened 2020-07-28
Add HelixManager constructor with RealmAwareZkConnectionConfig

### LINKED ISSUE #1182: HelixManager constructor that accepts RealmAwareZkConnectionConfig
Need a way to create HelixManager with custom zk connection config.

narendly:
### Issues

Fixes #1182

### Description

Previously, there wasn't a way to use HelixManager with a custom routing data ZK connection config. This new way of constructing HelixManager allows users to use HelixManager with a custom RealmAwareZkConnectionConfig.

### Tests

testZKHelixManager

zookeeper-api:
```
[ERROR] Failures: 
[ERROR]   TestRawZkClient.testGetChildrenOnLargeNumChildren:908 Should not successfully get children.
[INFO] 
[ERROR] Tests run: 39, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
```
This is a known failure due to https://github.com/apache/helix/pull/1186

helix-core:

```
[ERROR] Failures: 
[ERROR]   TestZKCallback.testInvocation:188
[INFO] 
[ERROR] Tests run: 1156, Failures: 2, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE

```
For the two failures, the tests were run individually.
```
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 28.056 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
```

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
@@ -19,13 +19,6 @@ package org.apache.helix;
  * under the License.
  */
 
-/**
- * factory that creates cluster managers
- *
- * for zk-based cluster managers, the getZKXXX(..zkClient) that takes a zkClient parameter
- *   are intended for session expiry test purpose
- */
-
 import org.apache.helix.manager.zk.HelixManagerStateListener;
 import org.apache.helix.manager.zk.ZKHelixManager;
 import org.slf4j.Logger;
@@ -34,6 +27,10 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Obtain one of a set of Helix cluster managers, organized by the backing system.
+ * factory that creates cluster managers
+ *  *
+ *  * for zk-based cluster managers, the getZKXXX(..zkClient) that takes a zkClient parameter
+ *  *   are intended for session expiry test purpose
  */
 public final class HelixManagerFactory {
   private static final Logger LOG = LoggerFactory.getLogger(HelixManagerFactory.class);
--- a/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
@@ -27,10 +27,7 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Obtain one of a set of Helix cluster managers, organized by the backing system.
- * factory that creates cluster managers
- *  *
- *  * for zk-based cluster managers, the getZKXXX(..zkClient) that takes a zkClient parameter
- *  *   are intended for session expiry test purpose
+ * factory that creates cluster managers.
  */
 public final class HelixManagerFactory {
   private static final Logger LOG = LoggerFactory.getLogger(HelixManagerFactory.class);
--- a/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
@@ -28,7 +28,6 @@ package org.apache.helix;
 
 import org.apache.helix.manager.zk.HelixManagerStateListener;
 import org.apache.helix.manager.zk.ZKHelixManager;
-import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -69,8 +68,8 @@ public final class HelixManagerFactory {
   }
 
   /**
-   * Construct a ZkHelixManager using the HelixManagerProperty instance given. If a proper
-   * ZkConnectionConfig. HelixManagerProperty given must contain a valid ZkConnectionConfig.
+   * Construct a ZkHelixManager using the HelixManagerProperty instance given.
+   * HelixManagerProperty given must contain a valid ZkConnectionConfig.
    * @param clusterName
    * @param instanceName
    * @param type
@@ -86,12 +85,13 @@ public final class HelixManagerFactory {
   }
 
   /**
-   * Construct a ZkHelixManager using the HelixManagerProperty instance given. If a proper
-   * ZkConnectionConfig is given in HelixManagerProperty, zkAddr field will be overriden.
+   * Construct a ZkHelixManager using the HelixManagerProperty instance given.
+   * NOTE: if both zkAddr and a valid ZkConnectionConfig are given in HelixManagerProperty, the
+   * instantiation will fail - only one is required.
    * @param clusterName
    * @param instanceName
    * @param type
-   * @param zkAddr will be overriden if a valid ZkConnectionConfig is given in helixManagerProperty
+   * @param zkAddr
    * @param stateListener
    * @param helixManagerProperty
    * @return
--- a/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
@@ -77,23 +77,4 @@ public final class HelixManagerFactory {
     return new ZKHelixManager(clusterName, instanceName, type, null, stateListener,
         helixManagerProperty);
   }
-
-  /**
-   * Construct a ZkHelixManager using the HelixManagerProperty instance given.
-   * NOTE: if both zkAddr and a valid ZkConnectionConfig are given in HelixManagerProperty, the
-   * instantiation will fail - only one is required.
-   * @param clusterName
-   * @param instanceName
-   * @param type
-   * @param zkAddr
-   * @param stateListener
-   * @param helixManagerProperty
-   * @return
-   */
-  public static HelixManager getZKHelixManager(String clusterName, String instanceName,
-      InstanceType type, String zkAddr, HelixManagerStateListener stateListener,
-      HelixManagerProperty helixManagerProperty) {
-    return new ZKHelixManager(clusterName, instanceName, type, zkAddr, stateListener,
-        helixManagerProperty);
-  }
 }
--- a/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
@@ -25,10 +25,13 @@ package org.apache.helix;
  * for zk-based cluster managers, the getZKXXX(..zkClient) that takes a zkClient parameter
  *   are intended for session expiry test purpose
  */
+
 import org.apache.helix.manager.zk.HelixManagerStateListener;
 import org.apache.helix.manager.zk.ZKHelixManager;
+import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 
 /**
  * Obtain one of a set of Helix cluster managers, organized by the backing system.
@@ -65,4 +68,38 @@ public final class HelixManagerFactory {
     return new ZKHelixManager(clusterName, instanceName, type, zkAddr, stateListener);
   }
 
+  /**
+   * Construct a ZkHelixManager using the HelixManagerProperty instance given. If a proper
+   * ZkConnectionConfig. HelixManagerProperty given must contain a valid ZkConnectionConfig.
+   * @param clusterName
+   * @param instanceName
+   * @param type
+   * @param stateListener
+   * @param helixManagerProperty must contain a valid ZkConnectionConfig
+   * @return
+   */
+  public static HelixManager getZKHelixManager(String clusterName, String instanceName,
+      InstanceType type, HelixManagerStateListener stateListener,
+      HelixManagerProperty helixManagerProperty) {
+    return new ZKHelixManager(clusterName, instanceName, type, null, stateListener,
+        helixManagerProperty);
+  }
+
+  /**
+   * Construct a ZkHelixManager using the HelixManagerProperty instance given. If a proper
+   * ZkConnectionConfig is given in HelixManagerProperty, zkAddr field will be overriden.
+   * @param clusterName
+   * @param instanceName
+   * @param type
+   * @param zkAddr will be overriden if a valid ZkConnectionConfig is given in helixManagerProperty
+   * @param stateListener
+   * @param helixManagerProperty
+   * @return
+   */
+  public static HelixManager getZKHelixManager(String clusterName, String instanceName,
+      InstanceType type, String zkAddr, HelixManagerStateListener stateListener,
+      HelixManagerProperty helixManagerProperty) {
+    return new ZKHelixManager(clusterName, instanceName, type, zkAddr, stateListener,
+        helixManagerProperty);
+  }
 }
--- a/helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
@@ -49,54 +49,82 @@ public class HelixManagerProperty {
   @Deprecated
   public HelixManagerProperty(Properties helixManagerProperties, CloudConfig cloudConfig) {
     _helixCloudProperty = new HelixCloudProperty(cloudConfig);
-    setVersion(helixManagerProperties.getProperty(SystemPropertyKeys.HELIX_MANAGER_VERSION));
-    setHealthReportLatency(
+    _version = helixManagerProperties.getProperty(SystemPropertyKeys.HELIX_MANAGER_VERSION);
+    _healthReportLatency = Long.parseLong(
         helixManagerProperties.getProperty(SystemPropertyKeys.PARTICIPANT_HEALTH_REPORT_LATENCY));
   }
 
-  public HelixManagerProperty() {
+  private HelixManagerProperty(String version, long healthReportLatency,
+      HelixCloudProperty helixCloudProperty,
+      RealmAwareZkClient.RealmAwareZkConnectionConfig zkConnectionConfig,
+      RealmAwareZkClient.RealmAwareZkClientConfig zkClientConfig) {
+    _version = version;
+    _healthReportLatency = healthReportLatency;
+    _helixCloudProperty = helixCloudProperty;
+    _zkConnectionConfig = zkConnectionConfig;
+    _zkClientConfig = zkClientConfig;
   }
 
   public HelixCloudProperty getHelixCloudProperty() {
     return _helixCloudProperty;
   }
 
-  public void setHelixCloudProperty(HelixCloudProperty helixCloudProperty) {
-    _helixCloudProperty = helixCloudProperty;
-  }
-
   public String getVersion() {
     return _version;
-  }
-
-  public void setVersion(String version) {
-    _version = version;
   }
 
   public long getHealthReportLatency() {
     return _healthReportLatency;
   }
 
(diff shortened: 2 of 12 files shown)


helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -20,54 +20,82 @@
...
    */
+  @Deprecated
   public HelixManagerProperty(Properties helixManagerProperties, CloudConfig cloudConfig) {
     _helixCloudProperty = new HelixCloudProperty(cloudConfig);
     setVersion(helixManagerProperties.getProperty(SystemPropertyKeys.HELIX_MANAGER_VERSION));
     setHealthReportLatency(
         helixManagerProperties.getProperty(SystemPropertyKeys.PARTICIPANT_HEALTH_REPORT_LATENCY));
   }
 
+  public HelixManagerProperty() {
```
jiajunwang, line 57: I think the plan is only allowing the property construct through HelixPropertyFactory. Shall we just make this constructor package-private? Also the other set methods.

@zhangmeng916 Please correct me if this is not the plan.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -20,54 +20,82 @@
...
 
   /**
+   * ** Deprecated - HelixManagerProperty should be a general property/config object used for
+   * HelixManager creation, not tied only to Properties or CloudConfig **
+   *
    * Initialize Helix manager property with default value
    * @param helixManagerProperties helix manager related properties input as a map
    * @param cloudConfig cloudConfig read from Zookeeper
    */
+  @Deprecated
```
jiajunwang, line 49: I agree with this change.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -225,9 +226,24 @@ public ZKHelixManager(String clusterName, String instanceName, InstanceType inst
       String zkAddress, HelixManagerStateListener stateListener,
       HelixManagerProperty helixManagerProperty) {
```
jiajunwang, line 227: helixManagerProperty.RealmAwareZkConnectionConfig now contains zk connection information that possibly conflicts with zkAddress. This is dangerous.
Actually, I already see one potential bug in the method getMetadataStoreConnectionString. Which always returns the zkAddress only without checking the connectionConfig.

Please ensure we are only recording realmAwareZkConnectionConfig. Let's try to remove _zkAddress for good. Or at least update this field according to the realmAwareZkConnectionConfig input.

Other configurations such as sessionTimeout are in the same situation.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
narendly: I don't like that idea - it loses flexibility. What is the benefit of making this package-private?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
narendly: The documentation makes it clear that zkAddress given will be overriden if a valid ZkConnectionConfig is supplied.

We cannot remove zkAddress as that would break backward-compatibility.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
jiajunwang: So all the config read/processing work is in HelixPropertyFactory. In this case, we won't have unexpected overwrite or misusage of the property object. The flexibility should be in HelixPropertyFactory only.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
jiajunwang: Ideally, we should deprecate this one but create a constructor with clean parameters.

Besides a real concern is that if one passes a wrong or invalid zkAddress with this constructor (since he or she assumes it will be overridden by ZkConnectionConfig), then getMetadataStoreConnectionString will return this invalid value too. And the WAGED rebalancer won't work.
A slightly less serious issue would be, most of our logs are using zkAddress, so it would be harder for us to debug.

narendly: @jiajunwang Then we could play it safe by adding a check so that we do not allow instantiation of a ZkHelixManager instance when both zkAddress and ZkConnectionConfig are set. How does that sound?

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
narendly: Then that in and of itself is a verbose way of doing things. At that point, we should eliminate and just add a HelixProperty Builder that allows creation of HelixManagerProperty instance with final fields.

zhangmeng916: The original plan was to let users only get the manager property through `getHelixManagerProperty` with default value, and then allow them to make modification to override the fields in it. I remember there was some concern about users accidentally changing it. We can revisit the assumption though.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -20,54 +20,82 @@
...
   public HelixManagerProperty(Properties helixManagerProperties, CloudConfig cloudConfig) {
     _helixCloudProperty = new HelixCloudProperty(cloudConfig);
     setVersion(helixManagerProperties.getProperty(SystemPropertyKeys.HELIX_MANAGER_VERSION));
     setHealthReportLatency(
```
zhangmeng916, line 53: fyi, the reason that we have a couple of fields like "version", "latency" here is just to give an example of how to migrate system properties to Helix manager property. A "TODO" should have been added here for completeness in future work.

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
```diff
@@ -65,4 +68,38 @@ public static HelixManager getZKHelixManager(String clusterName, String instance
     return new ZKHelixManager(clusterName, instanceName, type, zkAddr, stateListener);
   }
 
+  /**
+   * Construct a ZkHelixManager using the HelixManagerProperty instance given. If a proper
+   * ZkConnectionConfig. HelixManagerProperty given must contain a valid ZkConnectionConfig.
```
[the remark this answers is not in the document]
narendly: Thanks. Will fix!

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
```diff
@@ -65,4 +68,38 @@ public static HelixManager getZKHelixManager(String clusterName, String instance
...
+        helixManagerProperty);
+  }
+
+  /**
+   * Construct a ZkHelixManager using the HelixManagerProperty instance given. If a proper
+   * ZkConnectionConfig is given in HelixManagerProperty, zkAddr field will be overriden.
+   * @param clusterName
+   * @param instanceName
+   * @param type
+   * @param zkAddr will be overriden if a valid ZkConnectionConfig is given in helixManagerProperty
```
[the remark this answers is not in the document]
narendly: Let us add a check for validation.

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
```diff
@@ -65,4 +64,39 @@ public static HelixManager getZKHelixManager(String clusterName, String instance
...
+   * @param instanceName
+   * @param type
+   * @param zkAddr
+   * @param stateListener
+   * @param helixManagerProperty
+   * @return
+   */
+  public static HelixManager getZKHelixManager(String clusterName, String instanceName,
+      InstanceType type, String zkAddr, HelixManagerStateListener stateListener,
+      HelixManagerProperty helixManagerProperty) {
```
jiajunwang, line 98: Can we get rid of this constructor by allowing setup zkAddr in the HelixManagerProperty or in the RealmAwareZkConnectionConfig?

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -33,7 +33,7 @@
 import org.slf4j.LoggerFactory;
 
 
-/**
+/***
```
jiajunwang, line 36: extra *

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
jiajunwang: In addition, if we do so, then the HelixPropertyFactory can call the property builder to build a property even for the existing getHelixProperty API, right?
So it would be clean,
1. Callers only call factory methods to get property.
2. The factory calls the property builder to get the new object.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -131,6 +131,7 @@
   private int _reportLatency;
 
   protected RealmAwareZkClient _zkclient;
+  private RealmAwareZkClient.RealmAwareZkConnectionConfig _realmAwareZkConnectionConfig;
```
jiajunwang, line 134: As I pinged you in slack, if we can let RealmAwareZkConnectionConfig backward supports _zkAddress, then we can safe tons of validations and branches code in this class.
Moreover, we can simplify the code following this principle,
1. new APIs only accept RealmAwareZkConnectionConfig, we don't use zkAddress directly.
2. for older API, we create default RealmAwareZkConnectionConfig based on the zkAddress. So we don't have 2 possibilities in the internal code implementations.

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
```diff
@@ -19,19 +19,18 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
 /**
  * Obtain one of a set of Helix cluster managers, organized by the backing system.
+ * factory that creates cluster managers
+ *  *
+ *  * for zk-based cluster managers, the getZKXXX(..zkClient) that takes a zkClient parameter
+ *  *   are intended for session expiry test purpose
```
zhangmeng916, line 33: This description is not quite easy to follow.

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -88,8 +88,9 @@ private CloudConfig getCloudConfig(String zkAddress, String clusterName) {
     CloudConfig cloudConfig;
     RealmAwareZkClient dedicatedZkClient = null;
     try {
-      if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED)) {
-        // If the multi ZK config is enabled, use multi-realm mode with DedicatedZkClient
+      if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED) || zkAddress == null) {
```
zhangmeng916, line 91: Seems we will need to follow this patten: 
`if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED) || zkAddr == null)`
 in a couple of different places. Is there a better way to represent? The first one is self explanatory, but the second one does not look straightforward. The doc for explanation is also in another file.

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
narendly: @zhangmeng916 

Objectively, this comment is not very helpful to me - one tip when giving reviews & comments: either make your comment concrete by clarifying what's not easy to follow, or suggest an alternative?

With that said, I'm just going to remove the second block since it doesn't apply here.

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
narendly: I have the comment explaining the behavior. Please suggest if you have another way to achieve this otherwise.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
narendly: That goes against the design of "RealmAware"ZkClient/configs. They are not tied to a single zk, and therefore we should not have a ZK field.

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
narendly: RealmAware zk config should not have a field called ZkAddress. That's against the design of RealmAware API.

narendly: As to whether we should "move" `zkAddress` into helixManagerProperty - this idea is not ideal. It is what we have now and we should keep it in the constructor for backward-compatibility, and giving users multiple ways to input `zkAddress` will undoubtedly confuse users. 

So I think the right thing to do here is to keep zkAddress in HelixManager/HelixManagerFactory as a parameter. It won't continue to increase getZkHelixManager() methods. All future configs that are needed to generate an instance of HelixManager could go into `helixManagerProperty`.

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
zhangmeng916: Could you make it in a function? So for example, if later you have other conditions, you'll still need to add it in all related places?

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
narendly: Update: the factory and builder are all serving different purposes and might be useful to have in the future. We decided to keep both.

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
jiajunwang: I think we shall not be blocked on this zkAddress argument. It can be discussed later.

How about this, to ensure we are not introducing more confusing, we don't this specific method. So for this class, either you give a zkAddress to get HM or you give a HelixManagerProperty to get the HM.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -111,7 +111,7 @@
   public static final int DEFAULT_MAX_DISCONNECT_THRESHOLD = 600; // Default to be a large number
   private static final int DEFAULT_WAIT_CONNECTED_TIMEOUT = 10 * 1000;  // wait until connected for up to 10 seconds.
 
-  protected final String _zkAddress;
+  protected String _zkAddress;
```
jiajunwang, line 114: No need to remove final right?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
jiajunwang, line 134: Can we make this field final too? Since it is an input config.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1464,4 +1471,48 @@ private RealmAwareZkClient resolveZkClient(HelixZkClientFactory zkClientFactory,
...
+   * Check that not both zkAddress and ZkConnectionConfig are set.
+   * If zkAddress is not given and ZkConnectionConfig is given, check that ZkConnectionConfig has
+   * a ZK path sharding key set because HelixManager must work on single-realm mode.
+   * @param zkAddress
+   * @param helixManagerProperty
+   */
+  private void validateZkConnectionSettings(String zkAddress,
+      HelixManagerProperty helixManagerProperty) {
+    if (helixManagerProperty != null && helixManagerProperty.getZkConnectionConfig() != null) {
+      if (zkAddress != null && !zkAddress.isEmpty()) {
```
jiajunwang, line 1485: Let's be strict here, I think even empty string is not allowed. The zkAddress shall be completely untouched.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1464,4 +1471,48 @@ private RealmAwareZkClient resolveZkClient(HelixZkClientFactory zkClientFactory,
...
+   * Resolve ZK connection info for logging purposes.
+   * @return
+   */
+  private String getZkConnectionInfo() {
+    String zkConnectionInfo;
+    if (_zkAddress == null) {
+      if (_helixManagerProperty != null && _helixManagerProperty.getZkConnectionConfig() != null) {
+        zkConnectionInfo = _helixManagerProperty.getZkConnectionConfig().toString();
+      } else {
+        zkConnectionInfo = "ZkAddr and ZkConnectionConfig are null!";
```
jiajunwang, line 1511: nit, I think just return "None" is enough. Since otherwise, the log string will look strange.

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
narendly: Have you considered: what if a user needs to provide both `zkAddr` and `helixManagerProperty`?

At any rate, we could add this constructor later when we absolutely need to. For the sake of keeping this moving forward, I'll remove this `getZkHelixManager` method for now.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
narendly: Now that we've changed the design for HelixManager constructors to always accept zkAddress, we could bring the `final` keyword back.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
narendly: It is not given via ZkHelixManager's constructor, meaning it could be null. It comes from HelixManagerProperty as an optional field. So we cannot make this final.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
narendly: Sure, removing `&& !zkAddress.isEmpty()`.

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
jiajunwang: My thought is the same as what you described. We can always add it later if needed.

jiajunwang: It seems this comment has been marked as resolved without finishing the code change. Are you still working on this change?

narendly:
This PR is ready to be merged

PR #1187, opened 2020-07-29
Add latency metric support for customized view aggregation

### LINKED ISSUE #1214: Add latency metric support for customized view aggregation
As customized view aggregation is a new feature added to Helix controller, it is hard to predict its impact to the performance of our controller, and the performance of this pipeline is hard to estimate.
We need to support for latency metric reporting between the time when user updates customized state for a partition using CustomizedStateProvider, and the time this change shows up in aggregated customized view and written to zk. With that, we can 1. compare this latency with other pipeline latencies, so we know if this newly added feature saturate our controller or not 2. get an understanding of the pipeline performance 3. we can better monitor this feature and diagnose any problems in the future

mgao0:
### Issues

Fixes #1214

### Description

This PR adds support for latency metric reporting between the time when user updates customized state for a partition using CustomizedStateProvider, and the time this change shows up in aggregated customized view and written to zk. With that, we can 1. compare this latency with other pipeline latencies, so we know if this newly added feature saturate our controller or not 2. get an understanding of the pipeline performance 3. better monitor this feature and diagnose any problems in the future

Added: 
1. TestCustomizedViewMonitor
2. TestCustomizedViewStage -> testLatencyMetricReporting

Modified:
1. TestCustomizedStateUpdate
2. TestCustomizedViewAggregation

[INFO]
[INFO] Results:
[INFO]
[WARNING] Tests run: 1168, Failures: 0, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:13 h
[INFO] Finished at: 2020-08-13T19:47:48-07:00

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
@@ -27,7 +27,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import javax.management.JMException;
 
 import org.apache.helix.HelixConstants;
 import org.apache.helix.HelixDataAccessor;
@@ -38,7 +37,6 @@ import org.apache.helix.common.caches.CustomizedViewCache;
 import org.apache.helix.common.caches.PropertyCache;
 import org.apache.helix.controller.LogUtil;
 import org.apache.helix.controller.pipeline.Pipeline;
-import org.apache.helix.controller.stages.ClusterEvent;
 import org.apache.helix.controller.stages.MissingTopStateRecord;
 import org.apache.helix.model.CustomizedState;
 import org.apache.helix.model.CustomizedStateConfig;
@@ -46,7 +44,6 @@ import org.apache.helix.model.CustomizedView;
 import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.ResourceAssignment;
-import org.apache.helix.monitoring.mbeans.CustomizedViewMonitor;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -82,7 +79,6 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
   // Maintain a set of all ChangeTypes for change detection
   private Set<HelixConstants.ChangeType> _refreshedChangeTypes;
   private Set<String> _aggregationEnabledTypes = new HashSet<>();
-  private Map<String, CustomizedViewMonitor> _customizedViewMonitors = new HashMap<>();
 
   // CrushEd strategy needs to have a stable partition list input. So this cached list persist the
   // previous seen partition lists. If the members in a list are not modified, the old list will be
@@ -441,17 +437,6 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
   public void clearMonitoringRecords() {
     _missingTopStateMap.clear();
     _lastTopStateLocationMap.clear();
-  }
-
-  public synchronized CustomizedViewMonitor getOrCreateCustomizedViewMonitor(ClusterEvent event)
-      throws JMException {
-    String clusterName = event.getClusterName();
-    if (_customizedViewMonitors.get(clusterName) == null) {
-      CustomizedViewMonitor monitor = new CustomizedViewMonitor(clusterName);
-      _customizedViewMonitors.put(clusterName, monitor);
-      monitor.register();
-    }
-    return _customizedViewMonitors.get(clusterName);
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import javax.management.JMException;
 
 import org.apache.helix.HelixConstants;
 import org.apache.helix.HelixDataAccessor;
@@ -37,6 +38,7 @@ import org.apache.helix.common.caches.CustomizedViewCache;
 import org.apache.helix.common.caches.PropertyCache;
 import org.apache.helix.controller.LogUtil;
 import org.apache.helix.controller.pipeline.Pipeline;
+import org.apache.helix.controller.stages.ClusterEvent;
 import org.apache.helix.controller.stages.MissingTopStateRecord;
 import org.apache.helix.model.CustomizedState;
 import org.apache.helix.model.CustomizedStateConfig;
@@ -44,6 +46,7 @@ import org.apache.helix.model.CustomizedView;
 import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.ResourceAssignment;
+import org.apache.helix.monitoring.mbeans.CustomizedViewMonitor;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -79,6 +82,7 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
   // Maintain a set of all ChangeTypes for change detection
   private Set<HelixConstants.ChangeType> _refreshedChangeTypes;
   private Set<String> _aggregationEnabledTypes = new HashSet<>();
+  private Map<String, CustomizedViewMonitor> _customizedViewMonitors = new HashMap<>();
 
   // CrushEd strategy needs to have a stable partition list input. So this cached list persist the
   // previous seen partition lists. If the members in a list are not modified, the old list will be
@@ -437,6 +441,17 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
   public void clearMonitoringRecords() {
     _missingTopStateMap.clear();
     _lastTopStateLocationMap.clear();
+  }
+
+  public synchronized CustomizedViewMonitor getOrCreateCustomizedViewMonitor(ClusterEvent event)
+      throws JMException {
+    String clusterName = event.getClusterName();
+    if (_customizedViewMonitors.get(clusterName) == null) {
+      CustomizedViewMonitor monitor = new CustomizedViewMonitor(clusterName);
+      _customizedViewMonitors.put(clusterName, monitor);
+      monitor.register();
+    }
+    return _customizedViewMonitors.get(clusterName);
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateComputationStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateComputationStage.java
@@ -86,6 +86,9 @@ public class CustomizedStateComputationStage extends AbstractBaseStage {
           customizedStateOutput
               .setCustomizedState(customizedStateType, resourceName, partition, instanceName,
                   customizedState.getState(partitionName));
+          customizedStateOutput
+              .setStartTime(customizedStateType, resourceName, partition, instanceName,
+                  String.valueOf(customizedState.getStartTime(partitionName)));
         }
       }
     }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateComputationStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateComputationStage.java
@@ -32,7 +32,6 @@ import org.apache.helix.model.Resource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
 public class CustomizedStateComputationStage extends AbstractBaseStage {
   private static Logger LOG = LoggerFactory.getLogger(CustomizedStateComputationStage.class);
 
@@ -85,10 +84,8 @@ public class CustomizedStateComputationStage extends AbstractBaseStage {
         if (partition != null) {
           customizedStateOutput
               .setCustomizedState(customizedStateType, resourceName, partition, instanceName,
-                  customizedState.getState(partitionName));
-          customizedStateOutput
-              .setStartTime(customizedStateType, resourceName, partition, instanceName,
-                  String.valueOf(customizedState.getStartTime(partitionName)));
+                  customizedState.getState(partitionName),
+                  customizedState.getStartTime(partitionName));
         }
       }
     }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
@@ -24,32 +24,63 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.helix.model.CustomizedState;
 import org.apache.helix.model.Partition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 public class CustomizedStateOutput {
+  private static Logger LOG = LoggerFactory.getLogger(CustomizedStateOutput.class);
+
   // stateType -> (resourceName -> (Partition -> (instanceName -> customizedState)))
   private final Map<String, Map<String, Map<Partition, Map<String, String>>>> _customizedStateMap;
+  // stateType -> (resourceName -> (Partition -> (instanceName -> startTime)))
+  private final Map<String, Map<String, Map<Partition, Map<String, String>>>> _startTimeMap;
 
   public CustomizedStateOutput() {
     _customizedStateMap = new HashMap<>();
+    _startTimeMap = new HashMap<>();
   }
 
   public void setCustomizedState(String stateType, String resourceName, Partition partition,
       String instanceName, String state) {
-    if (!_customizedStateMap.containsKey(stateType)) {
-      _customizedStateMap
-          .put(stateType, new HashMap<String, Map<Partition, Map<String, String>>>());
+    setCustomizedStateProperty(CustomizedState.CustomizedStateProperty.CURRENT_STATE, stateType,
+        resourceName, partition, instanceName, state);
+  }
+
+  public void setStartTime(String stateType, String resourceName, Partition partition,
+      String instanceName, String state) {
+    setCustomizedStateProperty(CustomizedState.CustomizedStateProperty.START_TIME, stateType,
+        resourceName, partition, instanceName, state);
+  }
+
+  private void setCustomizedStateProperty(CustomizedState.CustomizedStateProperty propertyName,
+      String stateType, String resourceName, Partition partition, String instanceName,
+      String state) {
+    Map<String, Map<String, Map<Partition, Map<String, String>>>> mapToUpdate;
+    switch (propertyName) {
+      case CURRENT_STATE:
+        mapToUpdate = _customizedStateMap;
+        break;
+      case START_TIME:
+        mapToUpdate = _startTimeMap;
+        break;
+      default:
+        LOG.error(
+            "The customized state property is not supported, could not update customized state output.");
+        return;
     }
-    if (!_customizedStateMap.get(stateType).containsKey(resourceName)) {
-      _customizedStateMap.get(stateType)
-          .put(resourceName, new HashMap<Partition, Map<String, String>>());
+    if (!mapToUpdate.containsKey(stateType)) {
+      mapToUpdate.put(stateType, new HashMap<String, Map<Partition, Map<String, String>>>());
     }
-    if (!_customizedStateMap.get(stateType).get(resourceName).containsKey(partition)) {
-      _customizedStateMap.get(stateType).get(resourceName)
-          .put(partition, new HashMap<String, String>());
+    if (!mapToUpdate.get(stateType).containsKey(resourceName)) {
+      mapToUpdate.get(stateType).put(resourceName, new HashMap<Partition, Map<String, String>>());
     }
-    _customizedStateMap.get(stateType).get(resourceName).get(partition).put(instanceName, state);
+    if (!mapToUpdate.get(stateType).get(resourceName).containsKey(partition)) {
+      mapToUpdate.get(stateType).get(resourceName).put(partition, new HashMap<String, String>());
+    }
+    mapToUpdate.get(stateType).get(resourceName).get(partition).put(instanceName, state);
   }
 
   /**
@@ -59,8 +90,32 @@ public class CustomizedStateOutput {
    * @return
    */
   public Map<String, Map<Partition, Map<String, String>>> getCustomizedStateMap(String stateType) {
-    if (_customizedStateMap.containsKey(stateType)) {
-      return Collections.unmodifiableMap(_customizedStateMap.get(stateType));
+    return getCustomizedStateProperty(CustomizedState.CustomizedStateProperty.CURRENT_STATE,
+        stateType);
+  }
+
+  public Map<String, Map<Partition, Map<String, String>>> getStartTimeMap(String stateType) {
+    return getCustomizedStateProperty(CustomizedState.CustomizedStateProperty.START_TIME,
+        stateType);
+  }
+
+  private Map<String, Map<Partition, Map<String, String>>> getCustomizedStateProperty(
+      CustomizedState.CustomizedStateProperty propertyName, String stateType) {
+    Map<String, Map<String, Map<Partition, Map<String, String>>>> readFromMap;
+    switch (propertyName) {
+      case CURRENT_STATE:
+        readFromMap = _customizedStateMap;
+        break;
+      case START_TIME:
+        readFromMap = _startTimeMap;
+        break;
+      default:
+        LOG.error(
+            "The customized state property is not supported, could not read from customized state output.");
+        return Collections.emptyMap();
+    }
+    if (readFromMap.containsKey(stateType)) {
+      return Collections.unmodifiableMap(readFromMap.get(stateType));
     }
     return Collections.emptyMap();
   }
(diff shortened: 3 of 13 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -24,32 +24,63 @@
...
+    Map<String, Map<String, Map<Partition, Map<String, String>>>> mapToUpdate;
+    switch (propertyName) {
+      case CURRENT_STATE:
+        mapToUpdate = _customizedStateMap;
+        break;
+      case START_TIME:
+        mapToUpdate = _startTimeMap;
+        break;
+      default:
+        LOG.error(
```
pkuwm, line 70: Often, the only use for the ERROR level within a certain application is when a valuable business use case cannot be completed due to technical issues or a bug. 
I'd prefer not to use this logging level too generously because that would add too much noise to the logs and reduce the significance of a single ERROR event.

The WARN level should be used when something bad happened, but the application still has the chance to heal itself or the issue can wait a day or two to be fixed.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -24,32 +24,63 @@
...
+      mapToUpdate.get(stateType).put(resourceName, new HashMap<Partition, Map<String, String>>());
     }
-    if (!_customizedStateMap.get(stateType).get(resourceName).containsKey(partition)) {
-      _customizedStateMap.get(stateType).get(resourceName)
-          .put(partition, new HashMap<String, String>());
+    if (!mapToUpdate.get(stateType).get(resourceName).containsKey(partition)) {
+      mapToUpdate.get(stateType).get(resourceName).put(partition, new HashMap<String, String>());
     }
-    _customizedStateMap.get(stateType).get(resourceName).get(partition).put(instanceName, state);
+    mapToUpdate.get(stateType).get(resourceName).get(partition).put(instanceName, state);
```
pkuwm, line 83: We could make it simpler and cleaner by adopting `computeIfAbsent`:
```
mapToUpdate.computeIfAbsent(stateType, k -> new HashMap<>())
  .computeIfAbsent(resourceName, k -> new HashMap<>())
  .computeIfAbsent(partition, k -> new HashMap<>())
  .put(instanceName, state);
```

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
```diff
@@ -152,6 +161,31 @@ private void computeCustomizedStateView(final Resource resource, final String st
...
+  private void asyncReportLatency(ExecutorService threadPool, CustomizedViewMonitor monitor,
+      List<CustomizedView> updatedCustomizedViews, Map<String, CustomizedView> curCustomizedViews,
+      Map<String, Map<Partition, Map<String, String>>> updatedStartTimestamps,
+      boolean[] updateSuccess, long curTime) {
+    AbstractBaseStage.asyncExecute(threadPool, () -> {
+      try {
+        monitor.reportLatency(updatedCustomizedViews, curCustomizedViews, updatedStartTimestamps,
+            updateSuccess, curTime);
+      } catch (Exception e) {
+        LOG.error("Failed to report UpdateToAggregationLatency metric.", e);
```
pkuwm, line 186: I suggest `warn` level.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,190 @@
...
+import com.google.common.collect.Maps;
+import org.apache.helix.model.CustomizedView;
+import org.apache.helix.model.Partition;
+import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMBeanProvider;
+import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMetric;
+import org.apache.helix.monitoring.mbeans.dynamicMBeans.HistogramDynamicMetric;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testng.collections.Lists;
+
```
pkuwm, line 24: The latest helix-style doesn't format two empty lines. Maybe good to update helix-style in your IntelliJ.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,190 @@
...
+
+public class CustomizedViewMonitor extends DynamicMBeanProvider {
+  private static Logger LOG = LoggerFactory.getLogger(CustomizedViewMonitor.class);
```
pkuwm, line 27: `static final`

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,190 @@
+package org.apache.helix.monitoring.mbeans;
+
```
pkuwm, line 2: Apache License?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,190 @@
...
+  /**
+   * Find updated customized states and report the aggregation latency of each customized state
+   * @param updatedCustomizedViews Customized views that have been updated, obtained from CustomizedStateOutput
+   * @param curCustomizedViews Current customized view values from the CustomizedViewCache
+   * @param updatedStartTimestamps All customized state START_TIME property values from CustomizedStateOutput
+   * @param updateSuccess If the customized view update to ZK is successful or not
+   * @param endTime The timestamp when the new customized view is updated to ZK
+   */
+  public void reportLatency(List<CustomizedView> updatedCustomizedViews,
```
pkuwm, line 75: It seems we need to consider concurrency?

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
mgao0: Good to know. Thanks! I changed all the ERROR level to warning since it's not significant issue

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
mgao0: These events are put into a thread pool for async execution, but the parameters we put in are all hard copied so they won't interfere with each other. So each event is independent.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
```diff
@@ -152,6 +163,33 @@ private void computeCustomizedStateView(final Resource resource, final String st
     if (curCustomizedView == null || !curCustomizedView.getRecord().equals(view.getRecord())) {
       // Add customized view to the list which will be written to ZK later.
       updatedCustomizedViews.add(view);
+      updatedStartTimestamps.put(resourceName,
+          customizedStateOutput.getResourceStartTimeMap(stateType, resourceName));
     }
   }
+
+  private CustomizedViewMonitor getOrCreateMonitor(ClusterEvent event) throws JMException {
```
jiajunwang, line 171: Once we put this map into the cache object, concurrent control is necessary for this method.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
```diff
@@ -97,7 +103,8 @@ public void execute(final ClusterEvent event) throws Exception {
       for (Resource resource : resourceMap.values()) {
         try {
           computeCustomizedStateView(resource, stateType, customizedStateOutput, curCustomizedViews,
-              updatedCustomizedViews);
+              updatedCustomizedViews, updatedStartTimestamps);
+          Map<String, CustomizedView> curCustomizedViewsCopy = new HashMap<>(curCustomizedViews);
```
zhangmeng916, line 107: As you never use this variable in other places, just move it to line 120.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
```diff
@@ -107,8 +114,11 @@ public void execute(final ClusterEvent event) throws Exception {
           }
           // add/update customized-views from zk and cache
           if (updatedCustomizedViews.size() > 0) {
-            dataAccessor.setChildren(keys, updatedCustomizedViews);
+            boolean[] success = dataAccessor.setChildren(keys, updatedCustomizedViews);
             cache.updateCustomizedViews(stateType, updatedCustomizedViews);
+            asyncReportLatency(cache.getAsyncTasksThreadPool(), getOrCreateMonitor(event),
+                new ArrayList<>(updatedCustomizedViews), curCustomizedViewsCopy,
```
zhangmeng916, line 120: Have you run the whole integration tests? I'm a bit concerned about the extra memory cost. Do you need to whole curCustomizedViewCopy? The ones that are corresponding to the updatedViews should be fine, right? as you only need those timestamps to compare.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,151 @@
...
+              try {
+                long timestamp = Long.parseLong(partitionStartTimeMap.get(instanceName));
+                if (timestamp > 0) {
+                  collectedTimestamps.add(timestamp);
+                } else {
+                  LOG.warn(
+                      "Failed to find customized state update time stamp for reos, the number should be positive.");
+                }
+              } catch (NumberFormatException e) {
+                LOG.warn("Error occurs while parsing customized state update time stamp");
```
zhangmeng916, line 142: -> timestamp, one word.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,151 @@
...
+
+    for (int i = 0; i < updatedCustomizedViews.size(); i++) {
+      if (!updateSuccess[i]) {
+        continue;
+      }
+      CustomizedView newCV = updatedCustomizedViews.get(i);
+      String resourceName = newCV.getResourceName();
+      CustomizedView oldCV =
+          curCustomizedViews.getOrDefault(resourceName, new CustomizedView(resourceName));
+      if (newCV.getRecord().equals(oldCV.getRecord())) {
```
zhangmeng916, line 111: Is this still necessary? The updated customized views should have been compared and filtered.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,151 @@
...
+          for (Map.Entry<String, String> stateMapEntry : newStateMap.entrySet()) {
+            String instanceName = stateMapEntry.getKey();
+            if (!stateMapEntry.getValue().equals(oldStateMap.get(instanceName))) {
+              try {
+                long timestamp = Long.parseLong(partitionStartTimeMap.get(instanceName));
+                if (timestamp > 0) {
+                  collectedTimestamps.add(timestamp);
+                } else {
+                  LOG.warn(
+                      "Failed to find customized state update time stamp for reos, the number should be positive.");
```
zhangmeng916, line 139: typo?

helix-core/src/test/java/org/apache/helix/controller/stages/TestCustomizedViewStage.java
```diff
@@ -112,4 +110,60 @@ public void testCachedCustomizedViews() throws Exception {
...
+    runStage(event, new ResourceComputationStage());
+    runStage(event, new CustomizedStateComputationStage());
+    runStage(event, new CustomizedViewAggregationStage());
+
+    ObjectName objectName = new ObjectName(String
+        .format("%s:%s=%s", MonitorDomainNames.CustomizedView.name(), "Cluster", clusterName));
+    ObjectInstance monitor = _server.getObjectInstance(objectName);
+    Assert.assertNotNull(monitor);
+    TestHelper.verify(() -> (long) _server.getAttribute(objectName,
+        CustomizedViewMonitor.UPDATE_TO_AGGREGATION_LATENCY_GAUGE + ".Max") == 0,
```
zhangmeng916, line 160: Can you explain what is verified as 0?

mgao0: Good catch. I made a mistake here, it should be not equal to 0, which means there is a latency data point reported.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateComputationStage.java
```diff
@@ -86,6 +86,9 @@ private void updateCustomizedStates(String instanceName, String customizedStateT
           customizedStateOutput
               .setCustomizedState(customizedStateType, resourceName, partition, instanceName,
                   customizedState.getState(partitionName));
+          customizedStateOutput
+              .setStartTime(customizedStateType, resourceName, partition, instanceName,
+                  String.valueOf(customizedState.getStartTime(partitionName)));
```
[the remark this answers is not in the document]
mgao0: Good suggestion. I'll keep the long type, and save the two conversions of types for saving to and  retrieving from the map.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateComputationStage.java
```diff
@@ -86,6 +86,9 @@ private void updateCustomizedStates(String instanceName, String customizedStateT
           customizedStateOutput
               .setCustomizedState(customizedStateType, resourceName, partition, instanceName,
                   customizedState.getState(partitionName));
+          customizedStateOutput
```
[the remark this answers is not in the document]
mgao0: I added another parameter to setCustomizedState() and now this method can set both fields.
Originally I didn't want to change the interface of setCustomizedState() because it is a public method. But I think your comment also makes sense since we are going to keep this start time field, and this is the only place setCustomizedState()  is called.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -24,32 +24,58 @@
...
 import org.apache.helix.model.Partition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 public class CustomizedStateOutput {
+  private static final Logger LOG = LoggerFactory.getLogger(CustomizedStateOutput.class);
+
   // stateType -> (resourceName -> (Partition -> (instanceName -> customizedState)))
   private final Map<String, Map<String, Map<Partition, Map<String, String>>>> _customizedStateMap;
```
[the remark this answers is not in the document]
mgao0: Discussed offline, I tried the private class design and it didn't work ideally. I think we need a better and more comprehensive design to improve this. For now, let's just keep them separated.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedStateOutput.java
```diff
@@ -24,32 +24,58 @@
...
+  public void setStartTime(String stateType, String resourceName, Partition partition,
+      String instanceName, String state) {
+    setCustomizedStateProperty(CustomizedState.CustomizedStateProperty.START_TIME, stateType,
+        resourceName, partition, instanceName, state);
+  }
+
+  private void setCustomizedStateProperty(CustomizedState.CustomizedStateProperty propertyName,
+      String stateType, String resourceName, Partition partition, String instanceName,
+      String state) {
+    Map<String, Map<String, Map<Partition, Map<String, String>>>> mapToUpdate;
```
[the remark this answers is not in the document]
mgao0: I updated this part. I was trying to simplify the code but found it makes things more complicated. Now I just separate the two methods, as they do not really cause a lot duplicated code.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
mgao0: I was trying to be safe. But you are right, it's not necessary. Removed it.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
mgao0: See the next reply.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
mgao0: Discussed offline, current curCustomizedViewCopy is a whole copy of the entire set of customized views. We do not need that, so I will only save the curCustomizedView of what's changed, so we won't need to be concerned about memory usage. So curCustomizedViewCopy will be initialized as an empty map, and when we compare the customized views, if it's changed, we put the old customized view into this map, for later use to find out which partition state changed.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
```diff
@@ -112,14 +115,18 @@ public void execute(final ClusterEvent event) throws Exception {
       }
       // add/update customized-views from zk and cache
       if (updatedCustomizedViews.size() > 0) {
-        dataAccessor.setChildren(keys, updatedCustomizedViews);
+        boolean[] success = dataAccessor.setChildren(keys, updatedCustomizedViews);
```
zhangmeng916, line 118: Can we log somewhere how many updates are failed? It's useful information.

helix-core/src/test/java/org/apache/helix/controller/stages/TestCustomizedViewStage.java
```diff
@@ -112,4 +110,60 @@ public void testCachedCustomizedViews() throws Exception {
...
+    ObjectName objectName = new ObjectName(String
+        .format("%s:%s=%s", MonitorDomainNames.AggregatedView.name(), "Cluster", clusterName));
+    ObjectInstance monitor = _server.getObjectInstance(objectName);
+    Assert.assertNotNull(monitor);
+    TestHelper.verify(() -> (long) _server.getAttribute(objectName,
+        CustomizedViewMonitor.UPDATE_TO_AGGREGATION_LATENCY_GAUGE + ".Max") != 0,
+        TestHelper.WAIT_DURATION);
+
+    if (manager.isConnected()) {
+      manager.disconnect(); // For DummyClusterManager, this is not necessary
```
zhangmeng916, line 164: You commented as not necessary?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,145 @@
...
+    List<Long> collectedTimestamps = new ArrayList<>();
+
+    for (int i = 0; i < updatedCustomizedViews.size(); i++) {
+      if (!updateSuccess[i]) {
+        continue;
+      }
+      CustomizedView newCV = updatedCustomizedViews.get(i);
+      String resourceName = newCV.getResourceName();
+      CustomizedView oldCV =
+          curCustomizedViews.getOrDefault(resourceName, new CustomizedView(resourceName));
```
zhangmeng916, line 110: Is getOrDefault necessary? You already new one in aggregation stage.

mgao0: I feel it's safer. It's just a check so it doesn't add any burden. Although it's unlikely, but this is a public method so it theoretically can be called somewhere else other than aggregation stage where the handling is not done the same.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
mgao0: I added a logging in CustomizedViewMonitor because it is already looping through this boolean[] so it's the most convenient place to add it.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,149 @@
...
+  public DynamicMBeanProvider register() throws JMException {
+    List<DynamicMetric<?, ?>> attributeList = new ArrayList<>();
+    attributeList.add(_updateToAggregationLatencyGauge);
+    doRegister(attributeList, MBEAN_DESCRIPTION, getMBeanName());
+    return this;
+  }
+
+  private ObjectName getMBeanName() throws MalformedObjectNameException {
+    return new ObjectName(String
+        .format("%s:%s=%s", MonitorDomainNames.AggregatedView.name(), "Cluster", _clusterName));
```
jiajunwang, line 69: Besides cluster name, can we also have customized View as part of the key? Maybe type = customized view? This is mainly for future extension.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  void recordUpdateToAggregationLatency(long latency) {
+    if (_updateToAggregationLatencyGauge != null) {
+      _updateToAggregationLatencyGauge.updateValue(latency);
+    }
+  }
+
+  /**
+   * Find updated customized states and report the aggregation latency of each customized state
```
jiajunwang, line 84: Please mention a little bit more about the parameter usage here. Basically we only update the timestamp if the state has been changed, right?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -79,6 +82,7 @@
   // Maintain a set of all ChangeTypes for change detection
   private Set<HelixConstants.ChangeType> _refreshedChangeTypes;
   private Set<String> _aggregationEnabledTypes = new HashSet<>();
+  private Map<String, CustomizedViewMonitor> _customizedViewMonitors = new HashMap<>();
```
jiajunwang, line 85: On my second thought, let's follow the convention (refer to _clusterStatusMonitor) and put it in GenericHelixController. What do you think?

The problem with the cache is that this object won't be available if we change to pass the snapshot later.

In general, we should always use the event to carry the monitors. And GenericHelixController is one of the easiest options.

mgao0: As we discussed, I put CustomizedViewMonitor as a field in ClusterStatusMonitor, so now in CustomizedViewAggregationStage, we'll just get the CustomizedViewMonitor from the event's ClusterStatusMonitor attribute.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
```diff
@@ -88,7 +93,9 @@ public void execute(final ClusterEvent event) throws Exception {
     // update customized view
     for (String stateType : customizedStateOutput.getAllStateTypes()) {
       List<CustomizedView> updatedCustomizedViews = new ArrayList<>();
+      Map<String, Map<Partition, Map<String, Long>>> updatedStartTimestamps = new HashMap<>();
       Map<String, CustomizedView> curCustomizedViews = new HashMap<>();
+      Map<String, CustomizedView> curCustomizedViewsCopy = new HashMap<>();
```
jiajunwang, line 98: As we discussed, since this stage is already async stage, we don't need to introduce these complicities, we can make the report logic synchronizely done in this stage.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/CustomizedViewMonitor.java
```diff
@@ -0,0 +1,154 @@
...
+   * Latency reporting excludes: updating a customized state that is the same as previous customized state,
+   * deleting a customized state
+   * @param updatedCustomizedViews Customized views that have been updated, obtained from CustomizedStateOutput
+   * @param curCustomizedViews Current customized view values from the CustomizedViewCache
+   * @param updatedStartTimestamps All customized state START_TIME property values from CustomizedStateOutput.
+   * START_TIME field is automatically updated when a customized state for a partition is updated by CustomizedStateProvider.
+   * @param updateSuccess If the customized view update to ZK is successful or not
+   * @param endTime The timestamp when the new customized view is updated to ZK
+   */
+  public void reportLatency(List<CustomizedView> updatedCustomizedViews,
```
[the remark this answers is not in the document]
mgao0: I discussed with @zhangmeng916 , we think global latency is good enough for our purpose. So we won't use per-resource latency since it can cause complexities and won't bring additional benefit for us in the foreseeable future.

helix-core/src/main/java/org/apache/helix/controller/stages/CustomizedViewAggregationStage.java
```diff
@@ -152,6 +162,68 @@ private void computeCustomizedStateView(final Resource resource, final String st
...
+      CustomizedView newCV = updatedCustomizedViews.get(i);
+      String resourceName = newCV.getResourceName();
+
+      if (!updateSuccess[i]) {
+        LOG.warn("Customized views are not updated successfully for resource {} on cluster {}",
+            resourceName, clusterName);
+        continue;
+      }
+
+      CustomizedView oldCV =
```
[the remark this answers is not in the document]
mgao0: If the resource is newly created, the oldCV won't have the object, that's why I initialize an empty CustomizedView object here, otherwise it will return a null and will complicate the following comparisons.
Yes, we decided not to put a start timestamp when deleting a state, thus there's no need to check for a removed resource too.

mgao0:
Final commit message:
Add latency metric support for customized view aggregation

Issue #1189, opened 2020-07-30
Rest API to force remove the live instance ZNode of a zombie instances.


jiajunwang:
Zombie instance here means the participant that is still connected with the ZK but it refuses to do any work. In practice, we find it is hard to deal with these instances unless we manually remove their live instance node from the ZK.
It would be better if the Helix rest support such an API to give admins better support.

PR #1190, opened 2020-07-30
Add Helix rest Zookeeper delete API to allow removing ephemeral ZNode

jiajunwang:
### Issues

#1189

### Description

Add a new Helix rest API in the ZookeeperAccessor for deleting an ephemeral ZNode.

Note that before we have ACL/audit support in the Helix rest, allowing raw ZK write operation is dangerous.
This API is introduced prematurely for resolving the issue of "zombie" participant (the instance has an active zk connection, but refuse to do any work). Currently, the existence of such a node may block the normal state transitions and then impact the cluster's availability. This PR restricts that only an ephemeral node can be deleted to minimize the risk.

### Tests

TestZooKeeperAccessor.testDelete()

helix-rest

[INFO] Tests run: 164, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 49.55 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 164, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 55.836 s
[INFO] Finished at: 2020-07-30T15:40:28-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
@@ -21,6 +21,7 @@ package org.apache.helix.rest.server.resources.zookeeper;
 
 import java.util.List;
 import java.util.Map;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
@@ -58,8 +59,7 @@ public class ZooKeeperAccessor extends AbstractResource {
     getBinaryData,
     getStringData,
     getChildren,
-    getStat,
-    delete
+    getStat
   }
 
   @GET
@@ -75,8 +75,7 @@ public class ZooKeeperAccessor extends AbstractResource {
         (ServerContext) _application.getProperties().get(ContextPropertyKeys.SERVER_CONTEXT.name());
     _zkBaseDataAccessor = _serverContext.getByteArrayZkBaseDataAccessor();
 
-    // Need to prepend a "/" since JAX-RS regex removes it
-    path = "/" + path;
+    path = prependPath(path);
 
     // Check that the path supplied is valid
     if (!ZkValidationUtil.isPathValid(path)) {
@@ -96,13 +95,31 @@ public class ZooKeeperAccessor extends AbstractResource {
         return getChildren(_zkBaseDataAccessor, path);
       case getStat:
         return getStat(_zkBaseDataAccessor, path);
-      case delete:
-        return delete(_zkBaseDataAccessor, path);
       default:
         String errMsg = "Unsupported command: " + commandStr;
         LOG.error(errMsg);
         return badRequest(errMsg);
     }
+  }
+
+  @DELETE
+  @Path("{path: .+}")
+  public Response delete(@PathParam("path") String path) {
+    // Lazily initialize ZkBaseDataAccessor
+    ServerContext _serverContext =
+        (ServerContext) _application.getProperties().get(ContextPropertyKeys.SERVER_CONTEXT.name());
+    _zkBaseDataAccessor = _serverContext.getByteArrayZkBaseDataAccessor();
+
+    path = prependPath(path);
+
+    // Check that the path supplied is valid
+    if (!ZkValidationUtil.isPathValid(path)) {
+      String errMsg = "The given path is not a valid ZooKeeper path: " + path;
+      LOG.info(errMsg);
+      return badRequest(errMsg);
+    }
+
+    return delete(_zkBaseDataAccessor, path);
   }
 
   /**
@@ -207,22 +224,32 @@ public class ZooKeeperAccessor extends AbstractResource {
    * @return The delete result and the operated path.
    */
   private Response delete(BaseDataAccessor zkBaseDataAccessor, String path) {
-    // TODO: Remove this restriction once we have audit and ACL for the API calls.
-    // TODO: This method is added pre-maturely to support removing the live instance of a zombie
-    // TODO: instance. It is risky to allow all deleting requests before audit and ACL are done.
     Stat stat = zkBaseDataAccessor.getStat(path, AccessOption.PERSISTENT);
-    if (stat != null && stat.getEphemeralOwner() <= 0) {
+    if (stat == null) {
+      return notFound();
+    } else if (stat.getEphemeralOwner() <= 0) {
+      // TODO: Remove this restriction once we have audit and ACL for the API calls.
+      // TODO: This method is added pre-maturely to support removing the live instance of a zombie
+      // TODO: instance. It is risky to allow all deleting requests before audit and ACL are done.
       throw new WebApplicationException(Response.status(Response.Status.FORBIDDEN)
-          .entity(String.format("Deleting a non ephemeral node is not allowed", path)).build());
+          .entity(String.format("Deleting a non ephemeral node is not allowed. Path %s.", path))
+          .build());
     }
 
-    Boolean ret = zkBaseDataAccessor.remove(path, AccessOption.PERSISTENT);
-    Map<String, String> result =
-        ImmutableMap.of(ZooKeeperCommand.delete.name(), ret.toString(), PATH_STR, path);
-    return JSONRepresentation(result);
+    if (zkBaseDataAccessor.remove(path, AccessOption.PERSISTENT)) {
+      return OK();
+    } else {
+      throw new WebApplicationException(Response.status(Response.Status.INTERNAL_SERVER_ERROR)
+          .entity(String.format("Failed to delete %s.", path)).build());
+    }
   }
 
   private ZooKeeperCommand getZooKeeperCommandIfPresent(String command) {
     return Enums.getIfPresent(ZooKeeperCommand.class, command).orNull();
+  }
+
+  private String prependPath(String path) {
+    // Need to prepend a "/" since JAX-RS regex removes it
+    return "/" + path;
   }
 }
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
@@ -47,7 +47,7 @@ import org.slf4j.LoggerFactory;
  * ZooKeeperAccessor provides methods for accessing ZooKeeper resources (ZNodes).
  * It provides basic ZooKeeper features supported by ZkClient.
  */
-@Path("/zookeeper")
+@Path("/zookeeper{path: /.+}")
 public class ZooKeeperAccessor extends AbstractResource {
   private static final Logger LOG = LoggerFactory.getLogger(ZooKeeperAccessor.class.getName());
   private BaseDataAccessor<byte[]> _zkBaseDataAccessor;
@@ -61,7 +61,6 @@ public class ZooKeeperAccessor extends AbstractResource {
   }
 
   @GET
-  @Path("{path: .+}")
   public Response get(@PathParam("path") String path, @QueryParam("command") String commandStr) {
     ZooKeeperCommand cmd = getZooKeeperCommandIfPresent(commandStr);
     if (cmd == null) {
@@ -72,8 +71,6 @@ public class ZooKeeperAccessor extends AbstractResource {
     ServerContext _serverContext =
         (ServerContext) _application.getProperties().get(ContextPropertyKeys.SERVER_CONTEXT.name());
     _zkBaseDataAccessor = _serverContext.getByteArrayZkBaseDataAccessor();
-
-    path = prependPath(path);
 
     // Check that the path supplied is valid
     if (!ZkValidationUtil.isPathValid(path)) {
@@ -101,14 +98,11 @@ public class ZooKeeperAccessor extends AbstractResource {
   }
 
   @DELETE
-  @Path("{path: .+}")
   public Response delete(@PathParam("path") String path) {
     // Lazily initialize ZkBaseDataAccessor
     ServerContext _serverContext =
         (ServerContext) _application.getProperties().get(ContextPropertyKeys.SERVER_CONTEXT.name());
     _zkBaseDataAccessor = _serverContext.getByteArrayZkBaseDataAccessor();
-
-    path = prependPath(path);
 
     // Check that the path supplied is valid
     if (!ZkValidationUtil.isPathValid(path)) {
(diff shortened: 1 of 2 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -94,6 +96,8 @@ public Response get(@PathParam("path") String path, @QueryParam("command") Strin
         return getChildren(_zkBaseDataAccessor, path);
       case getStat:
         return getStat(_zkBaseDataAccessor, path);
+      case delete:
```
pkuwm, line 99: Would this go to a "DELETE" HTTP verb?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -52,12 +51,15 @@
...
+  private static final String PATH_STR = "path";
+
   public enum ZooKeeperCommand {
     exists,
     getBinaryData,
     getStringData,
     getChildren,
-    getStat
+    getStat,
+    delete
```
narendly, line 62: I think it would be smarter to use deleteEphemeral and rename your methods accordingly because it seems that it's  not the general delete you're trying to support.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -192,7 +196,29 @@ private Response getStat(BaseDataAccessor<byte[]> zkBaseDataAccessor, String pat
...
+    // TODO: Remove this restriction once we have audit and ACL for the API calls.
+    // TODO: This method is added pre-maturely to support removing the live instance of a zombie
+    // TODO: instance. It is risky to allow all deleting requests before audit and ACL are done.
+    Stat stat = zkBaseDataAccessor.getStat(path, AccessOption.PERSISTENT);
+    if (stat != null && stat.getEphemeralOwner() <= 0) {
+      throw new WebApplicationException(Response.status(Response.Status.FORBIDDEN)
+          .entity(String.format("Deleting a non ephemeral node is not allowed", path)).build());
+    }
+
+    Boolean ret = zkBaseDataAccessor.remove(path, AccessOption.PERSISTENT);
```
narendly, line 219: Would `zkBaseDataAccessor.remove(path, AccessOption.EPHEMERAL)` be an option?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -192,7 +196,29 @@ private Response getStat(BaseDataAccessor<byte[]> zkBaseDataAccessor, String pat
...
+   * @return The delete result and the operated path.
+   */
+  private Response delete(BaseDataAccessor zkBaseDataAccessor, String path) {
+    // TODO: Remove this restriction once we have audit and ACL for the API calls.
+    // TODO: This method is added pre-maturely to support removing the live instance of a zombie
+    // TODO: instance. It is risky to allow all deleting requests before audit and ACL are done.
+    Stat stat = zkBaseDataAccessor.getStat(path, AccessOption.PERSISTENT);
+    if (stat != null && stat.getEphemeralOwner() <= 0) {
+      throw new WebApplicationException(Response.status(Response.Status.FORBIDDEN)
+          .entity(String.format("Deleting a non ephemeral node is not allowed", path)).build());
```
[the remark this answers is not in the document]
jiajunwang: good catch!

helix-rest/src/test/java/org/apache/helix/rest/server/TestZooKeeperAccessor.java
```diff
@@ -192,4 +191,33 @@ public void testGetStat() throws IOException {
...
+      // 2. Create a ephemeral node. Delete shall be done successfully.
+      _testBaseDataAccessor.remove(deletePath, AccessOption.PERSISTENT);
+      _testBaseDataAccessor.create(deletePath, null, AccessOption.EPHEMERAL);
+      // Verify with the REST endpoint
+      String data = new JerseyUriRequestBuilder("zookeeper{}?command=delete").format(deletePath)
+          .isBodyReturnExpected(true).get(this);
+      Map<String, String> result = OBJECT_MAPPER.readValue(data, HashMap.class);
+      Assert.assertEquals(result.get("path"), deletePath);
+      Assert.assertEquals(result.get("delete"), new Boolean(true).toString());
+      Assert.assertFalse(_testBaseDataAccessor.exists(deletePath, AccessOption.PERSISTENT));
```
xyuanlu, line 217: Question here (may not related to your change).
Is the second param here in `exists(String path, int options)` useful? I did not find it being considered in any implementation of BaseDataAccessor.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
jiajunwang: That makes more sense, let me have a try.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
jiajunwang: I don't think we want "deleteEphemeral" eventually. As mentioned in the description, this is a premature feature that we add now for unblocking our users.

Alternatively, I tried to allow deleting live instance only. But that will pollute the ZookeeperAccessor API with Helix logics. So I discarded that idea.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
jiajunwang: The latter parameter is not used. Put EPHEMERAL or PERSISTENT is equally confusing. So let me just put the first enum item for now. I actually intend to avoid using EPHEMERAL since it indicates this parameter has some usage inside (but it is not true).

I think we need to discard this interesting API for good.

helix-rest/src/test/java/org/apache/helix/rest/server/TestZooKeeperAccessor.java
jiajunwang: No use at all. There is no such thing in ZK lib. I think we shall clean them up in a separate PR.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
narendly: I still don't agree - it seems that you are suggesting using `delete`, expecting its behavior to change in the future. That will bring about backward-compatibility issues and make the meaning of "delete" murky. 

It's fine if you need to provide another endpoint/API, but a good API design involves something that is 1) easy to do and 2) doing exactly what it's advertising to do. Do you see why it might be less desirable to add hidden assumptions to `delete`?

helix-rest/src/test/java/org/apache/helix/rest/server/TestZooKeeperAccessor.java
xyuanlu: Right. That confuses me as well. It seems many functions has a not-used 'int option' as input. May I suggest using a value consist with the node type as input? For example, we want to check if the newly created EPHEMERAL node is deleted, let use EPHEMERAL here to avoid further confusion.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
jiajunwang: I can see where you are coming from. Could you check the latest change that I have modified the method to use DELETE verb according to what Huizhi suggested? I think it is cleaner. However, in this case, we need some more parameters to separate the cases. And I think it might be overcomplicated.

@dasahcc and @pkuwm please also share your opinion since you also contributed to the Helix rest.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
jiajunwang: Let's don't touch the option for this PR. Let me create an issue for the unnecessary AccessOption.

pkuwm:
@jiajunwang By "zombie" participant, you meant the ephemeral node doesn't have any active zk connection/session, but it is not deleted by ZK?

jiajunwang:
> @jiajunwang By "zombie" participant, you meant the ephemeral node doesn't have any active zk connection/session, but it is not deleted by ZK?

Please read the description of PR, "the instance has an active zk connection, but refuse to do any work"

pkuwm:
> > @jiajunwang By "zombie" participant, you meant the ephemeral node doesn't have any active zk connection/session, but it is not deleted by ZK?
> 
> Please read the description of PR, "the instance has an active zk connection, but refuse to do any work"

Oh I read "**an active** zkconnection" as **inactive** zkconnection...

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -101,6 +100,26 @@ public Response get(@PathParam("path") String path, @QueryParam("command") Strin
...
 
+  @DELETE
+  @Path("{path: .+}")
+  public Response delete(@PathParam("path") String path) {
+    // Lazily initialize ZkBaseDataAccessor
+    ServerContext _serverContext =
+        (ServerContext) _application.getProperties().get(ContextPropertyKeys.SERVER_CONTEXT.name());
+    _zkBaseDataAccessor = _serverContext.getByteArrayZkBaseDataAccessor();
+
+    path = prependPath(path);
```
pkuwm, line 111: Actually if we set the root path as below, we don't need to prepend the root slash, which is not that clean.
```
@Path("/zookeeper{path: /.+}")
public class ZooKeeperAccessor extends AbstractResource {
```

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
pkuwm: My opinion is, 
1. use `DELETE` verb that is designed for the REST delete operation.
2. if we only want to support deleting ephemeral, document it well and return a clear response like:
```
    HTTP/1.1  404
    Content-Type: application/json
 
    {
      "message": "Deleting a non-ephemeral node is not supported/allowed",
      "path": "/a/b/c"
    }
```
And it is extensible if we want to support deleting persistent node in the future.

jiajunwang: This information is carried in the response entity as a string for now. I don't think we need to make it too structural (complicated) given it is a temporary restriction. And eventually, we do not have a clear standard for the response format now. So I would prefer holding on any more complex idea.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
jiajunwang: Neat, changed the code accordingly. There is more code like that. But I would not change them in this PR.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -196,6 +209,32 @@ private Response getStat(BaseDataAccessor<byte[]> zkBaseDataAccessor, String pat
...
+  /**
+   * Delete the ZNode at the given path if exists.
+   * @param zkBaseDataAccessor
+   * @param path
+   * @return The delete result and the operated path.
+   */
+  private Response delete(BaseDataAccessor zkBaseDataAccessor, String path) {
+    Stat stat = zkBaseDataAccessor.getStat(path, AccessOption.PERSISTENT);
+    if (stat == null) {
+      return notFound();
```
pkuwm, line 221: Could we add a msg to this as well: ("Path %s does not exist", path)? I think it gives a user a better idea. Otherwise the msg returned is unfriendly if we use `curl endpoint` in terminal.
```
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
<title>Error 404 </title>
</head>
<body>
<h2>HTTP ERROR: 404</h2>
<p>Problem accessing /admin/v2/zookeeper/aa. Reason:
<pre>    Not Found</pre></p>
<hr /><a href="http://eclipse.org/jetty">Powered by Jetty:// 9.4.12.v20180830</a><hr/>
</body>
</html>
```

VS
```
{
  "message" : "Path /aa does not exist",
  "status": 404
}
```

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -196,6 +209,32 @@ private Response getStat(BaseDataAccessor<byte[]> zkBaseDataAccessor, String pat
...
+    } else if (stat.getEphemeralOwner() <= 0) {
+      // TODO: Remove this restriction once we have audit and ACL for the API calls.
+      // TODO: This method is added pre-maturely to support removing the live instance of a zombie
+      // TODO: instance. It is risky to allow all deleting requests before audit and ACL are done.
+      throw new WebApplicationException(Response.status(Response.Status.FORBIDDEN)
+          .entity(String.format("Deleting a non-ephemeral node is not allowed.")).build());
+    }
+
+    if (zkBaseDataAccessor.remove(path, AccessOption.PERSISTENT)) {
+      return OK();
```
pkuwm, line 231: At least add a message `OK("Success")`?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/zookeeper/ZooKeeperAccessor.java
```diff
@@ -196,6 +209,32 @@ private Response getStat(BaseDataAccessor<byte[]> zkBaseDataAccessor, String pat
     return JSONRepresentation(result);
   }
 
+  /**
+   * Delete the ZNode at the given path if exists.
+   * @param zkBaseDataAccessor
+   * @param path
+   * @return The delete result and the operated path.
+   */
+  private Response delete(BaseDataAccessor zkBaseDataAccessor, String path) {
```
narendly, line 218: I'm not seeing how this is any different from using delete. This is no better than using `delete` for two different types of delete's - delete and deleteEphemeral.

Perhaps you could add a commandStr here to differentiate two different types of `delete`s, and when you want to add an endpoint for regular delete backed by ACL checks, then just implement that _if_ that becomes necessary?

My point was not about what kind of REST verb we should use - it's pretty clear we should use DELETE in this case. But it's more about following a good API design which, again, is something that is hard to misuse by not embedding hidden assumptions or TODOs that may cause a behavior change down the road.

jiajunwang: Discussed with Junkai in slack, his point is that we don't need the additional cmd layer for now.

jiajunwang:

PR #1194, opened 2020-07-31
Fix flaky test testGetChildrenOnLargeNumChildren

### LINKED ISSUE #1193: Flaky test testGetChildrenOnLargeNumChildren
The test uses reflection to change `ClientCnxn.packetLen`. It runs successfully until the commit where it is added. And after later commits are added, the test becomes flaky. Because `ClientCnxn.packetLen` value is not changed in zookeeper object that zkClient uses.

To fix it and make it simple, replace reflection with the naive way: create 110K children. To save the creation time, use zk transaction api to batch the znodes creation. Now the creation time is reduced from 170s to 4s.

pkuwm:
### Issues

Fixes #1193 

### Description

The test testGetChildrenOnLargeNumChildren becomes flaky after more commits are checked in.

### Tests

```
zookeeper-api

[INFO] Tests run: 38, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 62.807 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 38, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:07 min
[INFO] Finished at: 2020-07-30T17:10:34-07:00
```

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -21,14 +21,10 @@ import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
 import javax.management.JMException;
 
 import org.apache.helix.zookeeper.api.client.ChildrenSubscribeResult;
-import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.apache.helix.zookeeper.constant.ZkSystemPropertyKeys;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.exception.ZkClientException;
@@ -50,7 +46,6 @@ import org.apache.helix.zookeeper.zkclient.serialize.BasicZkSerializer;
 import org.apache.helix.zookeeper.zkclient.serialize.PathBasedZkSerializer;
 import org.apache.helix.zookeeper.zkclient.serialize.ZkSerializer;
 import org.apache.helix.zookeeper.zkclient.util.ExponentialBackoffStrategy;
-import org.apache.zookeeper.AsyncCallback;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.KeeperException.ConnectionLossException;
@@ -68,7 +63,6 @@ import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
 /**
  * "Native ZkClient": not to be used directly.
  *
@@ -83,11 +77,13 @@ public class ZkClient implements Watcher {
 
   // If number of children exceeds this limit, getChildren() should not retry on connection loss.
   // This is a workaround for exiting retry on connection loss because of large number of children.
+  // 100K is specific for helix messages which use UUID, making packet length just below 4 MB.
   // TODO: remove it once we have a better way to exit retry for this case
-  private static final int NUM_CHILDREN_LIMIT;
+  private static final int NUM_CHILDREN_LIMIT = 100 * 1000;
 
-
-  private static final String ZK_AUTOSYNC_ENABLED_DEFAULT = "true";
+  private static final boolean SYNC_ON_SESSION = Boolean.parseBoolean(
+      System.getProperty(ZkSystemPropertyKeys.ZK_AUTOSYNC_ENABLED, "true"));
+  private static final String SYNC_PATH = "/";
 
   private final IZkConnection _connection;
   private final long _operationRetryTimeoutInMillis;
@@ -118,19 +114,6 @@ public class ZkClient implements Watcher {
   // To automatically retry the async operation, we need a separate thread other than the
   // ZkEventThread. Otherwise the retry request might block the normal event processing.
   protected final ZkAsyncRetryThread _asyncCallRetryThread;
-
-  static {
-    // 100K is specific for helix messages which use UUID, making packet length just below 4 MB.
-    // Set it here for unit test to use reflection to change value
-    // because compilers optimize constants by replacing them inline.
-    NUM_CHILDREN_LIMIT = 100 * 1000;
-  }
-
-  private static final boolean SYNC_ON_SESSION = Boolean.parseBoolean(
-      System.getProperty(ZkSystemPropertyKeys.ZK_AUTOSYNC_ENABLED, ZK_AUTOSYNC_ENABLED_DEFAULT));
-  ;
-
-  private static final String SYNC_PATH = "/";
 
   private class IZkDataListenerEntry {
     final IZkDataListener _dataListener;
--- a/zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
+++ b/zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
@@ -875,8 +875,9 @@ public class TestRawZkClient extends ZkTestBase {
     for (int i = 0; i < 110; i++) {
       List<Op> ops = new ArrayList<>(1000);
       for (int j = 0; j < 1000; j++) {
-        String child = path + "/" + UUID.randomUUID().toString();
-        ops.add(Op.create(child, new byte[0], ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL));
+        String childPath = path + "/" + UUID.randomUUID().toString();
+        ops.add(
+            Op.create(childPath, new byte[0], ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL));
       }
       // Reduce total creation time by batch creating znodes
       _zkClient.multi(ops);
--- a/zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
+++ b/zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
@@ -20,8 +20,6 @@ package org.apache.helix.zookeeper.impl.client;
  */
 
 import java.lang.management.ManagementFactory;
-import java.lang.reflect.Field;
-import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
@@ -56,12 +54,13 @@ import org.apache.helix.zookeeper.zkclient.exception.ZkSessionMismatchedExceptio
 import org.apache.helix.zookeeper.zkclient.exception.ZkTimeoutException;
 import org.apache.helix.zookeeper.zkclient.metric.ZkClientMonitor;
 import org.apache.helix.zookeeper.zkclient.metric.ZkClientPathMonitor;
-import org.apache.zookeeper.ClientCnxn;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.KeeperException;
+import org.apache.zookeeper.Op;
 import org.apache.zookeeper.WatchedEvent;
 import org.apache.zookeeper.Watcher;
 import org.apache.zookeeper.Watcher.Event.KeeperState;
+import org.apache.zookeeper.ZooDefs;
 import org.apache.zookeeper.ZooKeeper;
 import org.apache.zookeeper.data.Stat;
 import org.testng.Assert;
@@ -69,7 +68,6 @@ import org.testng.AssertJUnit;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
-
 
 public class TestRawZkClient extends ZkTestBase {
   private final String TEST_TAG = "test_monitor";
@@ -571,8 +569,6 @@ public class TestRawZkClient extends ZkTestBase {
     // Wait until the ZkClient has got a new session.
     Assert.assertTrue(_zkClient.waitUntilConnected(1, TimeUnit.SECONDS));
 
-    final long originalSessionId = _zkClient.getSessionId();
-
     try {
       // Create node.
       _zkClient.create(path, data, CreateMode.PERSISTENT);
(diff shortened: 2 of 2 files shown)


zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -868,59 +864,38 @@ public void testAsyncWriteOperations() {
...
-      childrenLimitField.set(null, originChildrenLimit);
-      childrenLimitField.setAccessible(isChildrenLimitAccessible);
-
-      modifiersField.setAccessible(isModifierAccessible);
+      _zkClient.close();
+      _zkClient = new ZkClient(ZkTestBase.ZK_ADDR);
 
       Assert.assertTrue(TestHelper.verify(() -> {
-        _zkClient.deleteRecursively(path);
+        _zkClient.delete(path);
```
kaisun2000, line 898: can we delete parent path here without child node deleted?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -868,59 +864,38 @@ public void testAsyncWriteOperations() {
    * Tests getChildren() when there are an excessive number of children and connection loss happens,
    * the operation should terminate and exit retry loop.
    */
-  @Test
+  @Test(timeOut = 30L)
```
kaisun2000, line 867: why do we need this? is it 30sec?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -119,19 +115,6 @@
   // ZkEventThread. Otherwise the retry request might block the normal event processing.
   protected final ZkAsyncRetryThread _asyncCallRetryThread;
 
-  static {
```
kaisun2000, line 122: what is the difference of using this static block vs initalization at declaration?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: No, we cannot because there would be node not empty exception.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: `30 * 1000L ms`.

A good practice would be: if the test relies on an external system(zk here), it'd be good to have a timeout. Otherwise, if the connection hangs, the test also hangs. Eg., if the operation `multi()` hangs since it is sync, it hangs there forever?

Timeout makes the test process continue for other tests.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: If you read the comment: // Set it here for unit test to use reflection to change value	
    // because compilers optimize constants by replacing them inline.

kaisun2000: I see.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: Another thing is that in these 4 secs, is it possible another test would time out the session of _zkClient. Shall we use a dedicated ZkClient by creating one and later close it.

kaisun2000: also for safety, make it 60secs?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: My concern is this:

if using deleteRecursively(), it won't work because list all the children znode name would cause "marshalling" error as what happened in GCN, right?

if using delete as you do here, I think it would throw another exception as not empty. So this does not work here. Or am I missing something?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -868,59 +864,38 @@ public void testAsyncWriteOperations() {
...
-    String path = "/" + TestHelper.getTestMethodName();
-    // Create 5 children to make packet length of children exceed 150 bytes
+    // Create 110K children to make packet length of children exceed 4 MB
     // and cause connection loss for getChildren() operation
-    for (int i = 0; i < 5; i++) {
-      _zkClient.createPersistent(path + "/" + UUID.randomUUID().toString(), true);
+    String path = "/" + TestHelper.getTestMethodName();
+    int toCreateChildrenCount = 110;
+
+    _zkClient.createPersistent(path);
```
kaisun2000, line 874: How about let us construct a new zkclient here. Is it possible _zkClient would be session timeout by another test or thread?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -868,59 +864,38 @@ public void testAsyncWriteOperations() {
    * Tests getChildren() when there are an excessive number of children and connection loss happens,
    * the operation should terminate and exit retry loop.
    */
-  @Test
+  @Test(timeOut = 30 * 1000L)
```
kaisun2000, line 867: 60 secs?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: This _zkClient is created within this class. While this test is running, there is no other thread referencing this _zkClient to close it.

I considered using another zkClient. But it just adds one more connection to so for this case there will be two connections at the same time. I believe reuse this one is good enough.

I ran it in my laptop as well, 30s is good enough. If it's more than 30s, something is wrong and it should timeout to fail.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: Replied above. Won't be closed by another thread. It is created in this class, not from a base class.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: delete() returns boolean. Should handle `ZkException` here as I've removed deleteRecursively().

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: You laptop runs on SSD. Desktop or TMC machines can be slower. Leave some room for error is safe here IMO.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: So in this case, it would always throw exception, right? 

Then why bother. Just leave the znode there anyway.

pkuwm: As a good practice, a unit test should cleanup itself and avoid other tests.
It won't always close. If you carefully understand the code, you'll find the tips of using ephemeral nodes and closing the zkClient to delete the ephemeral nodes for cleanup.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: 30s has already good room for this test. In my tests, it completes in ~5s. If it really exceeds 30s, maybe 60s don't even help.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: My question is that is this exception always thrown? I believe it would always be thrown. Can you verify?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: Can you the whole test under helix in desktop to make sure this 30s is safe?

pkuwm: I did zookeeper-api in both desktop and laptop. Is this what you meant?

kaisun2000: No, run all test from helix. That is what TMC is doing I guess. This is to simulate that test environment. We don't want that you fix this one but later it turned out this test still not working over there. Note the cost over there is way higher than changing it here. 

In fact, I noticed in our test, there are failed test due to Timed out from time to time. I don't think it is a good idea to add a time out here. 

By the way, can you re-open this comment thread.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: Can we also re-open this comment thread?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: I actually didn't resolve this thread... I intended to keep it open for discussion.

Running all helix tests is not necessary. This is a separate module. Even running in the test machine, tests in different modules are independent.

Timeout is good practice. Just an example. https://github.com/apache/zookeeper/blob/fe940cdd8fb23ba09684cefb73233d570f4a20fa/zookeeper-server/src/test/java/org/apache/zookeeper/server/ZooKeeperServerMainTest.java#L165

How could you ensure 60s would not timeout then? Then how about 10 minutes, 1 hour?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: What do you mean by `always thrown`?

kaisun2000: `String path = "/" + TestHelper.getTestMethodName()` is path.

Now at the time this line of code `_zkClient.delete(path);` , there are a lot of znode underlying this `path`. So the delete should always fail, right?  Or am I understanding this wrong?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: Yes, then remove this timeout. The test has a default timeout 300sec. In my opinion, relying on 300sec default is much safer. Testing under your own laptop may not fit a lot other environment.

See this:

[ERROR] testLostZkConnection(org.apache.helix.integration.TestZkConnectionLost) Time elapsed: 300.015 s <<< FAILURE!
org.testng.internal.thread.ThreadTimeoutException: Method org.testng.internal.TestNGMethod.testLostZkConnection() didn't finish within the time-out 300000

pkuwm: Please let me remind you that the default timeout 300sec is set only for helix-core's tests, not in zookeeper-api.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: @kaisun2000 I've replied you: they are ephemeral znodes.
> It won't always throw exception. If you carefully understand the code, you'll find the tips of using ephemeral nodes and closing the zkClient to delete the ephemeral nodes for cleanup. The exception handling here is just in case zk server doesn't remove ephemeral nodes yet.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -868,60 +864,41 @@ public void testAsyncWriteOperations() {
...
-    for (int i = 0; i < 5; i++) {
-      _zkClient.createPersistent(path + "/" + UUID.randomUUID().toString(), true);
+    String path = "/" + TestHelper.getTestMethodName();
+
+    _zkClient.createPersistent(path);
+
+    for (int i = 0; i < 110; i++) {
+      List<Op> ops = new ArrayList<>(1000);
+      for (int j = 0; j < 1000; j++) {
+        String child = path + "/" + UUID.randomUUID().toString();
```
narendly, line 878: Nit: childPath

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -868,60 +864,41 @@ public void testAsyncWriteOperations() {
...
+        ops.add(Op.create(child, new byte[0], ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL));
+      }
+      // Reduce total creation time by batch creating znodes
+      _zkClient.multi(ops);
     }
 
     try {
       _zkClient.getChildren(path);
-      Assert.fail("Should not successfully get children.");
+      Assert.fail("Should not successfully get children because of connection loss.");
```
narendly, line 887: Nit: let's say this a little more clearly:

Something on the lines of "Should fail because listing 110K znodes with UUID names should cause packet length exceeded error, causing a connection loss."

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -868,60 +864,41 @@ public void testAsyncWriteOperations() {
...
-
-      modifiersField.setAccessible(isModifierAccessible);
+      _zkClient.close();
+      _zkClient = new ZkClient(ZkTestBase.ZK_ADDR);
 
       Assert.assertTrue(TestHelper.verify(() -> {
-        _zkClient.deleteRecursively(path);
-        return !_zkClient.exists(path);
+        try {
+          return _zkClient.delete(path);
```
narendly, line 898: Shouldn't you delete recursively? Wouldn't this fail because of children nodes?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: I see. Did not notice this ephemeral node. So closing the _zkClient delete the nodes. The second delete is just to delete the parent path. This looks good.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: They are ephemeral nodes and closing the zkClient deletes the ephemeral nodes for cleanup.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: where is the place to config the 300secs. Shall we just configure the same in zooscalability-api?

kaisun2000:
Other than timeout which may not be conservative enough, it looks good. The idea of using transaction and ephemeral node is brilliant.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
pkuwm: It is in test config file: `testng.xml`

Configuring the timeout for test suit for zookeeper-api is out of scope of this PR. If we want to do that, I would prefer to do it in another PR.

pkuwm:
```
testGetChildrenOnLargeNumChildren becomes flaky after more commits are checked in because of reflection doesn't work as expected. This commit fixes it by replacing reflection with creating 110K children for the test.
```

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
narendly: I think this is fine for testing purposes, but here are some scenarios we can think about:

1. doesn't the packet len error cause the session to be terminated? Or does your fix change that behavior? In this case if my assumption is true, the client that created the ephemeral nodes will lose its session and the ephemerals tied to the session will be gone anyway, so the nodes will be gone before zkclient.close()?

2. this is a rare scenario, but sometimes we observe in-memory local clients lose connection for some unknown reason. What impact would that have on the test?

If we really wanted to be right and thorough, I would write persistent nodes, and check that the numChildren field comes out to 110K from Stats with an Assert statement, and then do the explicit deletion at the end. This would guard against any potential client-side glitches and failures.

PR #1199, opened 2020-07-31
Resolve duplicate classes in different helix jars

### LINKED ISSUE #1198: Duplicate classes in exported jars
There are duplicate classes at runtime since 1.0+, because of multiple modules packaging. Two different JARs contain the same package but with different contents. Eg, both helix-core.jar and metrics-common.jar contains `SensorNameProvider.class`. And at tests runtime, we have a check to detect duplicate classes and this causes tests failure.

```
:application-war:treatyOfVersailles FAILED |  
-- | --
  | check war for conflicts within application.war
found 148 duplicates
--
found org/apache/helix/msdcommon/util/ZkValidationUtil.class inside [WEB-INF/lib/helix-core-1.0.1.3.jar, WEB-INF/lib/helix-common-1.0.1.3.jar, WEB-INF/lib/metadata-store-directory-common-1.0.1.3.jar] |  
-- | --
  | found org/apache/helix/SystemPropertyKeys.class inside [WEB-INF/lib/helix-core-1.0.1.3.jar, WEB-INF/lib/helix-common-1.0.1.3.jar] |  
  | found org/apache/helix/ZNRecord.class inside [WEB-INF/lib/helix-core-1.0.1.3.jar, WEB-INF/lib/helix-common-1.0.1.3.jar] |  
  | found org/apache/helix/ZNRecordDelta.class inside [WEB-INF/lib/helix-core-1.0.1.3.jar, WEB-INF/lib/helix-common-1.0.1.3.jar] |  
  | found org/apache/helix/zookeeper/api/client/ChildrenSubscribeResult.class inside [WEB-INF/lib/helix-core-1.0.1.3.jar, WEB-INF/lib/helix-common-1.0.1.3.jar, WEB-INF/lib/zookeeper-api-1.0.1.3.jar] |  
  | found org/apache/helix/zookeeper/api/client/HelixZkClient$ZkClientConfig.class inside [WEB-INF/lib/helix-core-1.0.1.3.jar, WEB-INF/lib/helix-common-1.0.1.3.jar, WEB-INF/lib/zookeeper-api-1.0.1.3.jar] |  
  | found org/apache/helix/zookeeper/api/client/HelixZkClient$ZkConnectionConfig.class inside [WEB-INF/lib/helix-core-1.0.1.3.jar, WEB-INF/lib/helix-common-1.0.1.3.jar, WEB-INF/lib/zookeeper-api-1.0.1.3.jar]
```

The problem is, when we package the helix jars, the dependent subproject's classes are packaged into helix-core.jar: all classes in module helix-core and its dependencies match `org.apache.helix.*` are packaged to helix-core.jar. Then there are duplicate classes from metrics-common.jar/zookeeper-api.jar/helix-core.jar

Though helix doesn't have duplicate classes in the same package, it is still not a good practice to package the same class into two different jars that have dependency relation.

It is necessary to clean up the duplicate classes. And small refactoring work is needed to fix this issue.

pkuwm:
### Issues

Fixes #1198 

### Description

There are duplicate classes at runtime since 1.0+, because of the module separation. Two different JARs contain the same package but with different contents. Eg, both helix-core.jar and metrics-common.jar contains SensorNameProvider.class. And at tests runtime, we have a check to detect duplicate classes and this causes tests failure.

Though helix doesn't have duplicate classes in the same package, it is still not a good practice to package the same class into two different jars that have dependency relation.

It is necessary to clean up the duplicate classes. And small refactoring work is needed to fix this issue, so with these different package paths, we could cleanly avoid exporting classes of metrics-common to helix-core.jar.

**Refactor change**:
Dedicated path for metrics-common: org.apache.helix.monitoring.XXX -> org.apache.helix.monitoring.**common**.XXX

image

### Tests

### Commits

### Documentation (Optional)

### Code Quality


helix-core/src/main/java/org/apache/helix/InstanceType.java
```diff
@@ -22,7 +22,7 @@
 import java.util.Arrays;
 import java.util.List;
 
-import org.apache.helix.monitoring.mbeans.MonitorDomainNames;
+import org.apache.helix.monitoring.common.mbeans.MonitorDomainNames;
```
narendly, line 25: I think the title of this PR/commit is a bit misleading - are we removing any classes?

I think instead of moving dual-existent classes so that they have different import paths, we should identify which classes both exist in both metrics-common and helix-core, and just remove the ones in helix-core. They were initially left in helix-core, but we should remove them clean since they're causing issues. This way, we won't have to modify any package names and the issue we get from duplicate classes at runtime will no longer be there.

pkuwm: Thanks for pointing out the misleading title. Changed it to resolved.

The classes in metrics-common have the same package path org.apache.helix.monitoring as those classes in helix-core. The challenge is, with existing bundle plugin/strategy in helix, there is no way to exclude the metric-common's classes in helix-core.jar.

Even we could change the bundle plugin to support class filtering, I still don't think it is clean if we still put same package classes in different jars. 

The right way for different modules/jars should have different package paths, eg.:
metrics-common: org.apache.helix.monitoring.common
helix-core: org.apache.helix.monitoring

With these different package paths, we could cleanly avoid exporting classes of metrics-common to helix-core.jar.
This is the reason why the PR would like to differentiate the package paths for different modules/jars.

pkuwm:
> The movement of the classes is logically not what we intended to do. We do want them in the common package. Let's think about better way to do it. But glad to know there are not many classes impacted.

@jiajunwang 
For current structure, helix-common is actually redundant as only helix-core depends on helix-common. And helix-common has the same package paths with helix-core. That being said, it is totally fine to just merge helix-common to helix-core. With current structure, I don't see a value that we have helix-common but only adds redundancy.

The only movement of classes is in metrics-common. The purpose is to clearly give a dedicated package path for metrics-common and cleanly avoid exporting classes of metrics-common to helix-core.jar.

I 100% agree that we should bring up a good plan to clean up the structure. This is what I would like to push forward to make helix code structure clear. And I believe you do, as well, and you possibly have an idea to restructure code base. :)

helix-core/src/main/java/org/apache/helix/InstanceType.java
narendly: Why don't we just remove the conflicting classes from helix-core? That is what my 2nd paragraph is saying. If you just change the package structure for the conflicting classes in metrics-common, you'd be masking the problem (the duality still exists) instead of solving the problem.

pkuwm: I guess you missed something about the problem. There are no conflicting classes in helix-core. All classes are unique/only in helix codebase. The problem is, when we package the helix jars, the dependent subproject's classes are packaged into helix-core.jar: all classes in module helix-core and its dependencies match `org.apache.helix.*` are packaged to helix-core.jar. Then there are duplicate classes from metrics-common.jar/zookeeper-api.jar/helix-core.jar

narendly:
> > The movement of the classes is logically not what we intended to do. We do want them in the common package. Let's think about better way to do it. But glad to know there are not many classes impacted.
> 
> @jiajunwang
> For current structure, helix-common is actually redundant as only helix-core depends on helix-common. And helix-common has the same package paths with helix-core. That being said, it is totally fine to just merge helix-common to helix-core. With current structure, I don't see a value that we have helix-common but only adds redundancy.
> 


It was an explicit design choice when we separated helix-common from helix-core. We shouldn't consider merging it back into helix-core as that would further complicate the module structure.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterEventMonitor.java
```diff
@@ -26,10 +26,10 @@
...
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.SlidingTimeWindowArrayReservoir;
-import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMBeanProvider;
-import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMetric;
-import org.apache.helix.monitoring.mbeans.dynamicMBeans.HistogramDynamicMetric;
-import org.apache.helix.monitoring.mbeans.dynamicMBeans.SimpleDynamicMetric;
+import org.apache.helix.monitoring.common.mbeans.dynamicMBeans.DynamicMBeanProvider;
+import org.apache.helix.monitoring.common.mbeans.dynamicMBeans.DynamicMetric;
+import org.apache.helix.monitoring.common.mbeans.dynamicMBeans.HistogramDynamicMetric;
+import org.apache.helix.monitoring.common.mbeans.dynamicMBeans.SimpleDynamicMetric;
```
narendly, line 32: The original package naming was intentional in order to not cause existing applications to have to change import statements (code change). Why is this needed? Are users importing multiple modules at the same time?

helix-core/src/main/java/org/apache/helix/InstanceType.java
narendly: So what you're claiming here is when maven packages all of them into one helix-core.jar, that jar ends up having multiple copies of the same class with the same package name?

The description in the issue https://github.com/apache/helix/issues/1198 is still not too clear - could you explain what the failure is exactly?

pkuwm: > The problem is, when we package the helix jars, the dependent subproject's classes are packaged into helix-core.jar: all classes in module helix-core and its dependencies match org.apache.helix.* are packaged to helix-core.jar. Then there are duplicate classes from metrics-common.jar/zookeeper-api.jar/helix-core.jar
I believe this is clear. And also I've pasted the failure log message.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterEventMonitor.java
pkuwm: When users import helix-core, since helix-core depends on helix-common, helix-common.jar and its transitive dependencies are also imported.
I understand the purpose of not changing packaging naming, but packaging dependencies into helix-core is not well handled, which causes the duplicate class issue.

pkuwm:
> > > The movement of the classes is logically not what we intended to do. We do want them in the common package. Let's think about better way to do it. But glad to know there are not many classes impacted.
> 
> > 
> 
> > @jiajunwang
> 
> > For current structure, helix-common is actually redundant as only helix-core depends on helix-common. And helix-common has the same package paths with helix-core. That being said, it is totally fine to just merge helix-common to helix-core. With current structure, I don't see a value that we have helix-common but only adds redundancy.
> 
> > 
> 
> 
> 
> 
> 
> It was an explicit design choice when we separated helix-common from helix-core. We shouldn't consider merging it back into helix-core as that would further complicate the module structure.

Then could you give a better proposal solution for the issue?

helix-core/src/main/java/org/apache/helix/InstanceType.java
narendly: Thanks for pasting the log message, but this doesn't seem like it's coming from Maven. Is this coming from some other build tool like gradle? Since this detail is missing in the issue/PR description, could you please add this detail?

I think this could be solved by tweaking the ivy file like you did here - do you think we could try to resolve the conflict without changing package names/moving packages?

narendly: Also, this part was a little misleading:

> And at tests runtime, we have a check to detect duplicate classes and this causes tests failure.

I wanted to check if this test is referring to some test happening outside of the Helix maven project? Could you confirm?

PR #1215, opened 2020-08-04
Remove the unnecessary checks to allow cleaning up map/list fields in the ClusterConfig and InstanceConfig.

### LINKED ISSUE #627: ClusterConfig/InstanceConfig set methods does not allow removal of keys for map and list fields
```
  /**
   * Set the required Instance Capacity Keys.
   * @param capacityKeys
   */
  public void setInstanceCapacityKeys(List<String> capacityKeys) {
    if (capacityKeys == null || capacityKeys.isEmpty()) {
      throw new IllegalArgumentException("The input instance capacity key list is empty.");
    }
    _record.setListField(ClusterConfigProperty.INSTANCE_CAPACITY_KEYS.name(), capacityKeys);
  }
```

I remember leaving a comment on the PR that added this that we need to allow the user to be able to set empty capacityKeys. @jiajunwang

jiajunwang:
### Issues

Resolves #627 

### Description

The null check or empty check disallow class users to clean up the corresponding fields.
This change remove these checkes in follow methods,
- ClusterConfig.setDefaultCapacityMap
- ClusterConfig.setGlobalRebalancePreference
- ClusterConfig.setAbnormalStateResolverMap
- InstanceConfig.setInstanceCapacityMap

### Tests

Modify the existing tests to cover the new change.

helix-core
[ERROR] Failures:
[ERROR]   TestAutoRebalance.testAutoRebalance:177 expected:<true> but was:<false>
[INFO]
[ERROR] Tests run: 1151, Failures: 1, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time: 01:17 h
[INFO] Finished at: 2020-08-04T00:12:37-07:00

Rerun
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 19.694 s
[INFO] Finished at: 2020-08-04T11:03:57-07:00

helix-rest
[INFO] Tests run: 164, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 53.047 s
[INFO] Finished at: 2020-08-04T11:05:11-07:00

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
+++ b/helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
@@ -870,15 +870,16 @@ public class ClusterConfig extends HelixProperty {
    *                    If null, the resolver map item will be removed from the config.
    */
   public void setAbnormalStateResolverMap(Map<String, String> resolverMap) {
-    if (resolverMap != null && resolverMap.values().stream()
-        .anyMatch(className -> className == null)) {
-      throw new IllegalArgumentException(
-          "Invalid Abnormal State Resolver Map definition. Class name cannot be empty.");
-    }
     if (resolverMap == null) {
       _record.getMapFields().remove(ClusterConfigProperty.ABNORMAL_STATES_RESOLVER_MAP.name());
+    } else {
+      if (resolverMap.values().stream()
+          .anyMatch(className -> className == null || className.isEmpty())) {
+        throw new IllegalArgumentException(
+            "Invalid Abnormal State Resolver Map definition. Class name cannot be empty.");
+      }
+      _record.setMapField(ClusterConfigProperty.ABNORMAL_STATES_RESOLVER_MAP.name(), resolverMap);
     }
-    _record.setMapField(ClusterConfigProperty.ABNORMAL_STATES_RESOLVER_MAP.name(), resolverMap);
   }
 
   public Map<String, String> getAbnormalStateResolverMap() {
--- a/helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
+++ b/helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
@@ -873,10 +873,14 @@ public class ClusterConfig extends HelixProperty {
     if (resolverMap == null) {
       _record.getMapFields().remove(ClusterConfigProperty.ABNORMAL_STATES_RESOLVER_MAP.name());
     } else {
-      if (resolverMap.values().stream()
-          .anyMatch(className -> className == null || className.isEmpty())) {
+      if (resolverMap.entrySet().stream().anyMatch(e -> {
+        String stateModelDefName = e.getKey();
+        String className = e.getValue();
+        return stateModelDefName == null || stateModelDefName.isEmpty() || className == null
+            || className.isEmpty();
+      })) {
         throw new IllegalArgumentException(
-            "Invalid Abnormal State Resolver Map definition. Class name cannot be empty.");
+            "Invalid Abnormal State Resolver Map definition. StateModel definition name and the resolver class name cannot be empty.");
       }
       _record.setMapField(ClusterConfigProperty.ABNORMAL_STATES_RESOLVER_MAP.name(), resolverMap);
     }
--- a/helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
+++ b/helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
@@ -709,9 +709,6 @@ public class ClusterConfig extends HelixProperty {
    * @param capacityKeys
    */
   public void setInstanceCapacityKeys(List<String> capacityKeys) {
-    if (capacityKeys == null || capacityKeys.isEmpty()) {
-      throw new IllegalArgumentException("The input instance capacity key list is empty.");
-    }
     _record.setListField(ClusterConfigProperty.INSTANCE_CAPACITY_KEYS.name(), capacityKeys);
   }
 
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -709,9 +709,6 @@ public void enableP2PMessage(boolean enabled) {
    * @param capacityKeys
    */
   public void setInstanceCapacityKeys(List<String> capacityKeys) {
-    if (capacityKeys == null || capacityKeys.isEmpty()) {
-      throw new IllegalArgumentException("The input instance capacity key list is empty.");
-    }
```
kaisun2000, line 714: What is the meaning of setting to 'null' and empty? What is the difference between  'null' and empty?

jiajunwang: Both empty the list. Null will remove the key. Empty list will set an empty list to the key. Both are fine for our logic.

kaisun2000: How about add an comment to let user know the intention. This would save further clarification etc,

jiajunwang: I was wrong, it actually just put a null value there. This is not what I want to do. Let me change it to the right behavior. Good catch!

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -788,39 +791,43 @@ public void setDefaultPartitionWeightMap(Map<String, Integer> weightDataMap)
   private void setDefaultCapacityMap(ClusterConfigProperty capacityPropertyType,
       Map<String, Integer> capacityDataMap) throws IllegalArgumentException {
     if (capacityDataMap == null) {
-      throw new IllegalArgumentException("Default capacity data is null");
+      _record.getMapFields().remove(capacityPropertyType.name());
+    } else {
+      Map<String, String> data = new HashMap<>();
+      capacityDataMap.entrySet().stream().forEach(entry -> {
+        if (entry.getValue() < 0) {
```
kaisun2000, line 798: If the value string is not valid integer representation, will this throw other Exception? If so, do you want to catch it and convert to IllegalArgumentException?

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -859,13 +866,18 @@ public boolean isGlobalRebalanceAsyncModeEnabled() {
 
   /**
    * Set the abnormal state resolver class map.
+   * @param resolverMap - the resolver map
+   *                    If null, the resolver map item will be removed from the config.
    */
   public void setAbnormalStateResolverMap(Map<String, String> resolverMap) {
-    if (resolverMap.values().stream()
-        .anyMatch(className -> className == null || className.isEmpty())) {
+    if (resolverMap != null && resolverMap.values().stream()
```
kaisun2000, line 873: I am a little bit confused. 

`anyMatch(className -> className == null))` means if there is a className (key) in resolverMap, we throw exception. 

Then how about value in resolverMap, can it be null? What does it mean then?

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: The value is not a string, it is Integer.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: It is searching through the value set. The stream() is not done to the key set. I think you missed the first part.

But it is a good catch that I should not remove "className.isEmpty()" check.

kaisun2000: Right, we are saying value can't be empty or 'null'. However, don't forget about key, the key in resovlerMap is potentially can be null too, right? Shall we also throw an exception there?

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -859,14 +866,20 @@ public boolean isGlobalRebalanceAsyncModeEnabled() {
...
-          "Invalid Abnormal State Resolver Map definition. Class name cannot be empty.");
+    if (resolverMap == null) {
+      _record.getMapFields().remove(ClusterConfigProperty.ABNORMAL_STATES_RESOLVER_MAP.name());
+    } else {
+      if (resolverMap.values().stream()
+          .anyMatch(className -> className == null || className.isEmpty())) {
+        throw new IllegalArgumentException(
+            "Invalid Abnormal State Resolver Map definition. Class name cannot be empty.");
+      }
+      _record.setMapField(ClusterConfigProperty.ABNORMAL_STATES_RESOLVER_MAP.name(), resolverMap);
```
kaisun2000, line 881: resolverMap can also contain key as null. Do we want to throw exception of this case?

jiajunwang: I see what you mean now. Make sense. Let me add the check for keys too.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -859,14 +866,24 @@ public boolean isGlobalRebalanceAsyncModeEnabled() {
...
+      _record.getMapFields().remove(ClusterConfigProperty.ABNORMAL_STATES_RESOLVER_MAP.name());
+    } else {
+      if (resolverMap.entrySet().stream().anyMatch(e -> {
+        String stateModelDefName = e.getKey();
+        String className = e.getValue();
+        return stateModelDefName == null || stateModelDefName.isEmpty() || className == null
+            || className.isEmpty();
+      })) {
+        throw new IllegalArgumentException(
+            "Invalid Abnormal State Resolver Map definition. StateModel definition name and the resolver class name cannot be empty.");
```
kaisun2000, line 883: nit. 

StateModel definition name can't be null, resolver class name cannot be empty can't be null or empty.

kaisun2000: We have a test case for this?

jiajunwang: Yes, the test has been added. Let me modify the log string.

jiajunwang:

PR #1218, opened 2020-08-04
Fix ClusterManager thread leak 

### LINKED ISSUE #1216: Identify and fix leaked threads of ClusterManager in our unit test
There are cases of leaking ClusterManager and its subclasses. The leaked threads can be in hundreds, Note, relying on GC of the ClusterManager object does not mean the thread is shutdown. For example TestNoDoubleAssign.

kaisun2000:
### Issues

fix #1216 

### Description

TestNoDoubleAssign relies on randomness of creating a lot of participant or
ClusterManager. Due to race condition of stopping/starting participants, the
Test leaks hundreds ClusterManager thread and ZkClient threads associated.

### Tests

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/integration/rebalancer/DelayedAutoRebalancer/TestDelayedAutoRebalance.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/rebalancer/DelayedAutoRebalancer/TestDelayedAutoRebalance.java
@@ -49,7 +49,7 @@ public class TestDelayedAutoRebalance extends ZkTestBase {
   protected static final int PARTITIONS = 5;
   // TODO: remove this wait time once we have a better way to determine if the rebalance has been
   // TODO: done as a reaction of the test operations.
-  protected static final int DEFAULT_REBALANCE_PROCESSING_WAIT_TIME = 1000;
+  protected static final int DEFAULT_REBALANCE_PROCESSING_WAIT_TIME = 3000;
 
   protected final String CLASS_NAME = getShortClassName();
   protected final String CLUSTER_NAME = CLUSTER_PREFIX + "_" + CLASS_NAME;
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
@@ -42,7 +42,7 @@ import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
 public class TestNoDoubleAssign extends TaskTestBase {
-  private static final int THREAD_COUNT = 1;
+  private static final int THREAD_COUNT = 10;
   private static final long CONNECTION_DELAY = 100L;
   private static final long POLL_DELAY = 50L;
   private static final String TASK_DURATION = "200";
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
@@ -155,16 +155,14 @@ public class TestNoDoubleAssign extends TaskTestBase {
    * Randomly causes Participants to lost connection temporarily.
    */
   private void breakConnection() {
-    // Note, THREAD_COUNT == 1 is a must to avoid leaking ClusterManager (participant).
-    // Otherwise, multiple thread has race of stopping and starting the same slot of index
-    // This would cause a vast amount (hundreds to thousand) leaking threads.
-    // The design of this test itself releys on randomness too, which is also not a good idea.
-    _executorServiceConnection = Executors.newSingleThreadScheduledExecutor();
+    _executorServiceConnection = Executors.newScheduledThreadPool(THREAD_COUNT);
     _executorServiceConnection.scheduleAtFixedRate(() -> {
       // Randomly pick a Participant and cause a transient connection issue
-      int participantIndex = RANDOM.nextInt(_numNodes);
-      stopParticipant(participantIndex);
-      startParticipant(participantIndex);
+      synchronized (this){
+        int participantIndex = RANDOM.nextInt(_numNodes);
+        stopParticipant(participantIndex);
+        startParticipant(participantIndex);
+      }
     }, 0L, CONNECTION_DELAY, TimeUnit.MILLISECONDS);
   }
 }
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
@@ -42,7 +42,7 @@ import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
 public class TestNoDoubleAssign extends TaskTestBase {
-  private static final int THREAD_COUNT = 10;
+  private static final int THREAD_COUNT = 1;
   private static final long CONNECTION_DELAY = 100L;
   private static final long POLL_DELAY = 50L;
   private static final String TASK_DURATION = "200";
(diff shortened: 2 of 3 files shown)


helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
```diff
@@ -42,7 +42,7 @@
 import org.testng.annotations.Test;
 
 public class TestNoDoubleAssign extends TaskTestBase {
-  private static final int THREAD_COUNT = 10;
+  private static final int THREAD_COUNT = 1;
```
pkuwm, line 45: Since thread count is 1, I think `newSingleThreadExecutor` is what you need?

helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
```diff
@@ -155,11 +155,16 @@ private void pollForDoubleAssign() {
    * Randomly causes Participants to lost connection temporarily.
    */
   private void breakConnection() {
+    // Note, send to THREAD_COUNT == 1 is a must to avoid leaking ClusterManager (participant).
```
jiajunwang, line 158: I suspect THREAD_COUNT > 1 is the test requirement here. But I'm not sure. @narendly, could you please comment here?

helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
```diff
@@ -155,11 +155,16 @@ private void pollForDoubleAssign() {
...
+    // Otherwise, multiple thread has race of stopping and starting the same slot of index
+    // this would cause a vast amount hundreds to thousand leaking threads.
+    // The design of this test itself relys on randomness too, which is also not a good idea.
     _executorServiceConnection = Executors.newScheduledThreadPool(THREAD_COUNT);
     _executorServiceConnection.scheduleAtFixedRate(() -> {
       // Randomly pick a Participant and cause a transient connection issue
       int participantIndex = RANDOM.nextInt(_numNodes);
-      _participants[participantIndex].disconnect();
+      // _participants[participantIndex].disconnect();
```
pkuwm, line 166: Remove this instead of commenting it out?

helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
kaisun2000: This test relies on randomness. Also, the do not really wait for enough time to do breakConnection() etc. However, if we wait, there will be even a lot more leaking. Let us make an agreement that the design of the test is out of scope of this pull request. First stabilize the test. Will use singleThreadExecutor for now,

helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
kaisun2000: changed to use newSingleThreadExecutor for break connection

helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
mgao0: @kaisun2000 Can you help me understand why THREAD_COUNT == 1 is a must to avoid leaking ClusterManager (participant)? Does it cause leakage even we shut down the executor after test?
In my understanding, the thread count there doesn't necessarily mean that this executor will use this number of threads at most. This number is for the core thread number, so more like a minimum number concept. If there's a  need, the executor will allocate more threads to handle the tasks even though THREAD_COUNT is specified as 1.

kaisun2000: Sure. Let me explain:

Assume we have two thread A, B. Both of them choose slot 0. The sequence of action is 

A stop 0
B stop 0
A start 0
B start 0

Now _participants[0] hold the reference of clutermanager started by B.  clustermanager started by A is not referenced by any one any more. They are leaked.

mgao0: Thanks for explaining @kaisun2000 . It makes sense.

jiajunwang: @kaisun2000 if that's the only reason, then we can just add a concurrent control on the stop, start block?

Moreover, I believe startParticipant will auto close it if one object is already running. We can also add synchronize there.

kaisun2000: something like this? I can give it a try. But it basically boils down to single thread case anyway. 

```
_executorServiceConnection.scheduleAtFixedRate(() -> {
      // Randomly pick a Participant and cause a transient connection issue
     synchronized (_executorServiceConnection) {
      int participantIndex = RANDOM.nextInt(_numNodes);
      stopParticipant(participantIndex);
      startParticipant(participantIndex);
      }
  }, 0L, CONNECTION_DELAY, TimeUnit.MILLISECONDS);

```

kaisun2000: > I believe startParticipant will auto close it if one object is already running. We can also add synchronize there.

Yes, it has. But over there still boils down to single thread execution. It does not increase parallism in this case. The above code seems to be clear in terms that we wan to stress that cleaning the resource by stopping it. Or we need to think about object life cycle.

helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
lei-xia, line 45: Is THREAD_COUNT still needed? If not, why not delete it?

helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
```diff
@@ -155,11 +155,15 @@ private void pollForDoubleAssign() {
    * Randomly causes Participants to lost connection temporarily.
    */
   private void breakConnection() {
-    _executorServiceConnection = Executors.newScheduledThreadPool(THREAD_COUNT);
+    // Note, THREAD_COUNT == 1 is a must to avoid leaking ClusterManager (participant).
```
lei-xia, line 158: Please update the comment.

helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
kaisun2000: pollForDoubleAssign() also use this one.

This test has a bad design. Aside from the race condition which cause leak,  you will see that breakConnection() and pollForDoubleAssign() does not really reliably reproduce the problem. 

See the other discussion with JJ, the proposed fix is:

```
_executorServiceConnection.scheduleAtFixedRate(() -> {
      // Randomly pick a Participant and cause a transient connection issue
     synchronized (_executorServiceConnection) {
      int participantIndex = RANDOM.nextInt(_numNodes);
      stopParticipant(participantIndex);
      startParticipant(participantIndex);
      }
  }, 0L, CONNECTION_DELAY, TimeUnit.MILLISECONDS);
```

Let me know what is your take? Since the proposed fix from you and JJ are different.

helix-core/src/test/resources/log4j.properties
```diff
@@ -16,7 +16,11 @@
 # specific language governing permissions and limitations
 # under the License.
 #
+
+# Set root logger level to DEBUG and its only appender to R.
```
jiajunwang, line 20: Remove the unnecessary comments?

helix-core/src/test/java/org/apache/helix/integration/rebalancer/DelayedAutoRebalancer/TestDelayedAutoRebalance.java
```diff
@@ -49,7 +49,7 @@
   protected static final int PARTITIONS = 5;
   // TODO: remove this wait time once we have a better way to determine if the rebalance has been
   // TODO: done as a reaction of the test operations.
-  protected static final int DEFAULT_REBALANCE_PROCESSING_WAIT_TIME = 1000;
+  protected static final int DEFAULT_REBALANCE_PROCESSING_WAIT_TIME = 3000;
```
jiajunwang, line 52: Are we sure this is causing the trouble?
Based on what I tested before, 1 second is already a very long time. If with 1 second, the test is still not stable, then it might be other problems.

Let me be clear,
1. Changing this time just prolong the test, not a big deal.
2. I'm more concerned that this won't really make the tests more stable.

helix-core/src/test/java/org/apache/helix/integration/task/TestNoDoubleAssign.java
```diff
@@ -158,9 +158,11 @@ private void breakConnection() {
     _executorServiceConnection = Executors.newScheduledThreadPool(THREAD_COUNT);
     _executorServiceConnection.scheduleAtFixedRate(() -> {
       // Randomly pick a Participant and cause a transient connection issue
-      int participantIndex = RANDOM.nextInt(_numNodes);
-      _participants[participantIndex].disconnect();
-      startParticipant(participantIndex);
+      synchronized (this){
```
jiajunwang, line 161: In this case, shall we lock all the start/stop callers?
I think a better way is to lock inside the start and stop methods. Then add a reset method in addition which requests the same lock.

PR #1256, opened 2020-08-12
Do not ignore any map or list field key change in the Change Detector.

### LINKED ISSUE #1255: WAGED Change Detector ignores IS list field key changes
**Describe the bug**
There is a regression in the WAGED Change Detector. The detector ignores IS list field key changes. In this case, the rebalancer won't trigger the required rebalance. As a result, the newly added resource or new partitions won't be assigned automatically. 

**To Reproduce**
Call HelixAdmin.addResource() to add a new resource.
Wait for a while, until the first rebalance is done.
Call HelixAdmin.rebalance() for the new resource. Ideally, the admin will update the IS with a fully extended partition list. And the rebalancer should assign the partitions. However, due to the regression, this 2nd change will not trigger the baseline rebalance. So the new partitions won't be assigned.

**Expected behavior**
After HelixAdmin.rebalance() called, all the partitions of this new resource should be assigned shortly, and automatically.

**Additional context**
Impacted Version: 1.0.1

This bug only impacts WAGED managed resources.

As a workaround, manually trigger rebalance (for example, touching the IdealState by changing the NUM_PARTITIONS and then changing it back), or reset the helix controller to trigger a refresh baseline rebalance.

jiajunwang:
### Issues

Fixes #1255 

### Description

This PR extends the scope of the non-trimmable elements that are defined in the ResourceChangeDetector. With this change, any key changes in the Map fields or List fields won't be ignored by the change detector. Instead of just fixing the issue that modifying the IS preference list does not trigger automatic rebalance, this change will also help to prevent other potential bugs with the same root cause.

Note this change might increase the rebalancer running in some circumstances. However, given a very explicit allowlist (or blocklist) in the change detector is very expensive to maintain, the current solution trades off the potential performance impact for simplicity.

### Tests

Modified the related tests:
1. TestWagedRebalancer.testRebalanceOnChanges
2. TestHelixPropoertyTimmer

[INFO] Tests run: 1163, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,626.399 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1163, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 01:17 h
[INFO] Finished at: 2020-08-12T00:29:45-07:00

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/changedetector/trimmer/HelixPropertyTrimmer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/changedetector/trimmer/HelixPropertyTrimmer.java
@@ -19,7 +19,7 @@ package org.apache.helix.controller.changedetector.trimmer;
  * under the License.
  */
 
-import java.util.List;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/controller/changedetector/trimmer/HelixPropertyTrimmer.java
```diff
@@ -57,38 +57,49 @@
    * @param originalProperty
    */
   protected ZNRecord doTrim(T originalProperty) {
+    ZNRecord originalZNRecord = originalProperty.getRecord();
```
narendly, line 60: This is a high-level comment: 

Since this is causing an issue for the admin.rebalance method (which essentially just populates the mapFields in the IdealState), my immediate instinct would be that this change could be contained in IdealStateTrimmer. 

This way, we contain the scope of changes for IdealStates only. Is there any difficulty with overriding `doTrim()` in IdealStateTrimmer instead of making the change and applying it generally in HelixPropertyTrimmer?

jiajunwang: Good point. So this is also my first try. FYI, I also tried to extend the field type to include LIST_FIELD_KEY_ONLY. But then I find these 2 solutions are too complicated.

It is correct that putting this logic into the IdealStateTrimmer will immediately fix the problem without too much change. The complex is that in this way, it would be hard to reason the overall logic of all trimmers. Some of them care about values, the others do not. Of course, we can add comments. But gradually, it will become harder to understand and maintain.

So I go back and think about why we need these trimmers. Do we need to trim soo much information? I think we don't. The rebalancer will perform well even we detect all keys change (no value, just key's change) for all the properties. And the logic is very clean and concentrate:
1. Keys are not ignored
2. Values are ignored accordingly and this is specified by the getNonTrimmableFields() method.

Then I firstly tried to not trim all keys (including the simple fields). But then I realize it is not optimal, since the simple fields' keys are tightly bounded with the values. And finally, I have the current proposal.

Overall, the main reason for this trade-off is to make the trim logic less diverse and easier to maintain.

helix-core/src/main/java/org/apache/helix/controller/changedetector/trimmer/HelixPropertyTrimmer.java
```diff
@@ -57,38 +57,49 @@
...
-        originalProperty).entrySet()) {
+    Map<FieldType, Set<String>> nonTrimmableFields = getNonTrimmableFields(originalProperty);
+
+    // Ensure the keys of all map fields and list fields are preserved even after the trim.
+    // The keys of list and map fields are relatively stable and contain important information. So
+    // they should not be trimmed.
+    originalZNRecord.getMapFields().keySet().stream()
+        .forEach(key -> trimmedZNRecord.setMapField(key, Collections.EMPTY_MAP));
+    originalZNRecord.getListFields().keySet().stream()
+        .forEach(key -> trimmedZNRecord.setListField(key, Collections.EMPTY_LIST));
```
NealSun96, line 72: This part of code may come as a surprise to future developers that extend this class. When `getNonTrimmableFields()` is provided for overriding, it's natural to think `getNonTrimmableFields()` provides full control on "what fields to keep". This logic here, while necessary, may become debug overhead. It's also hard to change this behavior in the future if it stays in `doTrim()` like so. 

I suggest creating another function such as `getNonTrimmableKeysOnly()` and implement it under `HelixPropertyTrimmer.java` to preserve MapField and ListField keys by default; in `doTrim()`, we call that function and preserve keys only. The extended classes may choose to override it (and for now, they shouldn't). This is a clearer approach to creating this abstract class.

helix-core/src/main/java/org/apache/helix/controller/changedetector/trimmer/HelixPropertyTrimmer.java
jiajunwang: BTW, the ideal solution is to separate the ZK paths for input and output. Then we can simplify or even remove the trim logic as desired.

helix-core/src/main/java/org/apache/helix/controller/changedetector/trimmer/HelixPropertyTrimmer.java
jiajunwang: Good idea. It helps to clean up the code.

jiajunwang:
This PR is approved and ready to be merged.

Re-run the test, all passed.

Issue #1268, opened 2020-08-14
Fix flaky TestEnableCompression hangs(timeout) often in github


kaisun2000:
Log:
>2020-08-12T00:08:59.7823040Z [ERROR] testEnableCompressionResource(org.apache.helix.integration.TestEnableCompression)  Time elapsed: 301.309 s  <<< FAILURE!
2020-08-12T00:08:59.7832929Z org.testng.internal.thread.ThreadTimeoutException: Method org.testng.internal.TestNGMethod.testEnableCompressionResource() didn't finish within the time-out 300000
2020-08-12T00:08:59.7840902Z         at org.apache.helix.integration.TestEnableCompression.testEnableCompressionResource(TestEnableCompression.java:116)
2020-08-12T00:08:59.7841208Z 

>2020-08-12T12:18:59.8288893Z ---------- Test Class TestPauseSignal thread leakage detected! ---------------
2020-08-12T12:19:04.7889521Z START testEnableCompressionResource at Wed Aug 12 12:19:04 UTC 2020
2020-08-12T12:19:04.7896157Z START TestEnableCompression_testEnableCompressionResource at Wed Aug 12 12:19:04 UTC 2020
2020-08-12T12:24:05.2712624Z END testEnableCompressionResource at Wed Aug 12 12:24:05 UTC 2020, took: 300482ms.
2020-08-12T12:24:05.2713094Z AfterClass:TestEnableCompression afterclass of ZkTestBase called!
2020-08-12T12:24:06.3960081Z TestEnableCompression has active threads cnt:838
2020-08-12T12:24:06.4334615Z zookeeper server threads has 4 thread

Test in intellij normally finished in 30 sec. 

hangs `boolean result = verifier.verify(120000L);`      // ------> hanges here.

```
public void testEnableCompressionResource() throws Exception {
    String className = TestHelper.getTestClassName();
    String methodName = TestHelper.getTestMethodName();
    String clusterName = className + "_" + methodName;

    System.out.println("START " + clusterName + " at " + new Date(System.currentTimeMillis()));

    MockParticipantManager[] participants = new MockParticipantManager[5];
    // ClusterSetup setupTool = new ClusterSetup(ZK_ADDR);
    int numNodes = 10;
    TestHelper.setupCluster(clusterName, ZK_ADDR, 12918, // participant port
        "localhost", // participant name prefix
        "TestDB", // resource name prefix
        0, // no resources, will be added later
        0, // partitions per resource
        numNodes, // number of nodes
[...]

PR #1295, opened 2020-08-20
fix TestRawZkClient unstableness

### LINKED ISSUE #1294: Fix flaky test TestRawZkClient.testZkClientMonitor()
LOG:
>2020-08-19T06:04:57.4143926Z [ERROR] testZkClientMonitor(org.apache.helix.zookeeper.impl.client.TestRawZkClient)  Time elapsed: 0.036 s  <<< FAILURE!
2020-08-19T06:04:57.4147601Z java.lang.AssertionError: expected:<0> but was:<1>
2020-08-19T06:04:57.4154957Z 	at org.apache.helix.zookeeper.impl.client.TestRawZkClient.testZkClientMonitor(TestRawZkClient.java:288)
2020-08-19T06:04:57.4155636Z

See the following code, there is a race condition here. connect() would cause a process() callback in zookeeper object eventThread.. If the _monitor is not constructed, the state changed value would be 0, otherwise, it would be 1.

github is slow, the value is 1.
```
 protected ZkClient(IZkConnection zkConnection, int connectionTimeout, long operationRetryTimeout,
      PathBasedZkSerializer zkSerializer, String monitorType, String monitorKey,
      String monitorInstanceName, boolean monitorRootPathOnly) {
    if (zkConnection == null) {
      throw new NullPointerException("Zookeeper connection is null!");
    }

    _uid = UID.getAndIncrement();

    _connection = zkConnection;
    _pathBasedZkSerializer = zkSerializer;
    _operationRetryTimeoutInMillis = operationRetryTimeout;

    _asyncCallRetryThread = new ZkAsyncRetryThread(zkConnection.getServers());
    _asyncCallRetryThread.start();
    LOG.info("ZkClient created with _uid {}, _asyncCallRetryThread id {}, stack {}", _uid, _asyncCallRetryThread.getId(),
        Arrays.asList(Thread.currentThread().getStackTrace()));

    connect(connectionTimeout, this);

    // initiate monitor
    try {
      if (monitorKey != null && !monitorKey.isEmpty() && monitorType != null && !monitorType
          .isEmpty()) {
        _monitor =
            new ZkClientMonitor(monitorType, monitorKey, monitorInstanceName, monitorRootPathOnly,
                _eventThread);
        _monitor.register();
      } else {
        LOG.info("ZkClient monitor key or type is not provided. Skip monitoring.");
      }
    } catch (JMException e) {
      LOG.error("Error in creating ZkClientMonitor", e);
    }
  }
```

kaisun2000:
### Issues

Fix #1294 

### Description

ZkClient connect to ZooKeeper client first before monitor bean
initialization. If the monitor is not fully constructed, the
test would pass. Otherwise, as in github, it would not. Fix this
issue by properly construct the object.

### Tests

None

In Old connection New session
After session expiry sessionId= 72277739317428225
Shut down zookeeper at port 2127 in thread main
Starting ZK server at localhost:2358
[INFO] Tests run: 44, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 701.476 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 44, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  11:47 min
[INFO] Finished at: 2020-08-19T18:11:02-07:00
ksun-mn1:zookeeper-api ksun$ git status 

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/integration/TestPartitionMovementThrottle.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/TestPartitionMovementThrottle.java
@@ -215,7 +215,7 @@ public class TestPartitionMovementThrottle extends ZkStandAloneCMTestBase {
     DelayedTransition.setDelay(20);
     DelayedTransition.enableThrottleRecord();
 
-    // start another 3 nodes 
+    // start another 3 nodes
     for (int i = NODE_NR - 3; i < NODE_NR; i++) {
       _participants[i].syncStart();
     }
--- a/helix-core/src/test/java/org/apache/helix/integration/TestPartitionMovementThrottle.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/TestPartitionMovementThrottle.java
@@ -215,7 +215,7 @@ public class TestPartitionMovementThrottle extends ZkStandAloneCMTestBase {
     DelayedTransition.setDelay(20);
     DelayedTransition.enableThrottleRecord();
 
-    // start another 3 nodes
+    // start another 3 nodes 
     for (int i = NODE_NR - 3; i < NODE_NR; i++) {
       _participants[i].syncStart();
     }
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -126,8 +126,6 @@ public class ZkClient implements Watcher {
   private PathBasedZkSerializer _pathBasedZkSerializer;
   private ZkClientMonitor _monitor;
 
-  private AtomicBoolean _syncDone = new AtomicBoolean(false);
-
   // To automatically retry the async operation, we need a separate thread other than the
   // ZkEventThread. Otherwise the retry request might block the normal event processing.
   protected final ZkAsyncRetryThread _asyncCallRetryThread;
@@ -236,6 +234,7 @@ public class ZkClient implements Watcher {
     }
 
     connect(connectionTimeout, this);
+
   }
 
   public List<String> subscribeChildChanges(String path, IZkChildListener listener) {
@@ -1279,10 +1278,6 @@ public class ZkClient implements Watcher {
         });
   }
 
-  @VisibleForTesting
-  public boolean getSyncStatus() {
-    return _syncDone.get();
-  }
 
   /*
    *  Note, issueSync takes a ZooKeeper (client) object and pass it to doAsyncSync().
@@ -1308,7 +1303,6 @@ public class ZkClient implements Watcher {
 
     KeeperException.Code code = KeeperException.Code.get(callbackHandler.getRc());
     if (code == KeeperException.Code.OK) {
-      _syncDone.set(true);
       LOG.info("sycnOnNewSession with sessionID {} async return code: {} and proceeds", sessionId,
           code);
       return true;
@@ -1327,7 +1321,6 @@ public class ZkClient implements Watcher {
 
     if (SYNC_ON_SESSION) {
       final ZooKeeper zk = ((ZkConnection) getConnection()).getZookeeper();
-      _syncDone.set(false);
       _eventThread.send(new ZkEventThread.ZkEvent("Sync call before new session event of session " + sessionId,
           sessionId) {
         @Override
@@ -2157,18 +2150,22 @@ public class ZkClient implements Watcher {
 
       IZkConnection zkConnection = getConnection();
       _eventThread = new ZkEventThread(zkConnection.getServers());
-      _eventThread.start();
-
-      LOG.debug("ZkClient created with _uid {}, _eventThread {}", _uid, _eventThread.getId());
 
       if (_monitor != null) {
         _monitor.setAndInitZkEventThreadMonitor(_eventThread);
-        try {
-          _monitor.register();
-        } catch (JMException e) {
-          LOG.error("Error in creating ZkClientMonitor {}!", e);
-        }
       }
+
+      try {
+        if (_monitor != null) {
+          _monitor.register();
+        }
+      } catch (JMException e){
+        LOG.error("Error in creating ZkClientMonitor", e);
+      }
+
+      _eventThread.start();
+
+      LOG.debug("ZkClient created with _uid {}, _eventThread {}", _uid, _eventThread.getId());
 
       if (isManagingZkConnection()) {
         zkConnection.connect(watcher);
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -35,6 +35,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.management.JMException;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.helix.zookeeper.api.client.ChildrenSubscribeResult;
 import org.apache.helix.zookeeper.constant.ZkSystemPropertyKeys;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
@@ -124,11 +125,6 @@ public class ZkClient implements Watcher {
   private volatile boolean _closed;
   private PathBasedZkSerializer _pathBasedZkSerializer;
   private ZkClientMonitor _monitor;
-
-  final private String _monitorKey;
-  final private String _monitorType;
-  final private String _monitorInstanceName;
-  final private boolean _monitorRootPathOnly;
 
   private AtomicBoolean _syncDone = new AtomicBoolean(false);
 
@@ -230,10 +226,14 @@ public class ZkClient implements Watcher {
     _asyncCallRetryThread.start();
     LOG.debug("ZkClient created with _uid {}, _asyncCallRetryThread id {}", _uid, _asyncCallRetryThread.getId());
 
-    _monitorType = monitorType;
-    _monitorKey = monitorKey;
-    _monitorInstanceName = monitorInstanceName;
-    _monitorRootPathOnly = monitorRootPathOnly;
+    if (monitorKey != null && !monitorKey.isEmpty() && monitorType != null && !monitorType
+        .isEmpty()) {
+      _monitor =
+          new ZkClientMonitor(monitorType, monitorKey, monitorInstanceName, monitorRootPathOnly,
+              _eventThread);
+    } else {
+      LOG.info("ZkClient monitor key or type is not provided. Skip monitoring.");
+    }
 
     connect(connectionTimeout, this);
   }
@@ -1279,6 +1279,7 @@ public class ZkClient implements Watcher {
         });
   }
 
+  @VisibleForTesting
   public boolean getSyncStatus() {
     return _syncDone.get();
   }
@@ -2160,19 +2161,13 @@ public class ZkClient implements Watcher {
 
       LOG.debug("ZkClient created with _uid {}, _eventThread {}", _uid, _eventThread.getId());
 
-      // initiate monitor
-      try {
-        if (_monitorKey != null && !_monitorKey.isEmpty() && _monitorType != null && !_monitorType
-            .isEmpty()) {
-          _monitor =
-              new ZkClientMonitor(_monitorType, _monitorKey, _monitorInstanceName, _monitorRootPathOnly,
-                  _eventThread);
+      if (_monitor != null) {
+        _monitor.setAndInitZkEventThreadMonitor(_eventThread);
+        try {
           _monitor.register();
-        } else {
-          LOG.info("ZkClient monitor key or type is not provided. Skip monitoring.");
+        } catch (JMException e) {
+          LOG.error("Error in creating ZkClientMonitor {}!", e);
         }
-      } catch (JMException e) {
-        LOG.error("Error in creating ZkClientMonitor", e);
       }
 
       if (isManagingZkConnection()) {
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -31,6 +31,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.management.JMException;
 
@@ -123,6 +124,13 @@ public class ZkClient implements Watcher {
   private volatile boolean _closed;
   private PathBasedZkSerializer _pathBasedZkSerializer;
   private ZkClientMonitor _monitor;
+
+  final private String _monitorKey;
+  final private String _monitorType;
+  final private String _monitorInstanceName;
+  final private boolean _monitorRootPathOnly;
+
+  private AtomicBoolean _syncDone = new AtomicBoolean(false);
 
   // To automatically retry the async operation, we need a separate thread other than the
   // ZkEventThread. Otherwise the retry request might block the normal event processing.
@@ -222,22 +230,12 @@ public class ZkClient implements Watcher {
     _asyncCallRetryThread.start();
     LOG.debug("ZkClient created with _uid {}, _asyncCallRetryThread id {}", _uid, _asyncCallRetryThread.getId());
 
-    connect(connectionTimeout, this);
+    _monitorType = monitorType;
+    _monitorKey = monitorKey;
+    _monitorInstanceName = monitorInstanceName;
+    _monitorRootPathOnly = monitorRootPathOnly;
 
-    // initiate monitor
-    try {
-      if (monitorKey != null && !monitorKey.isEmpty() && monitorType != null && !monitorType
-          .isEmpty()) {
-        _monitor =
-            new ZkClientMonitor(monitorType, monitorKey, monitorInstanceName, monitorRootPathOnly,
-                _eventThread);
-        _monitor.register();
-      } else {
-        LOG.info("ZkClient monitor key or type is not provided. Skip monitoring.");
-      }
-    } catch (JMException e) {
-      LOG.error("Error in creating ZkClientMonitor", e);
-    }
+    connect(connectionTimeout, this);
   }
 
   public List<String> subscribeChildChanges(String path, IZkChildListener listener) {
@@ -1281,6 +1279,10 @@ public class ZkClient implements Watcher {
         });
   }
 
+  public boolean getSyncStatus() {
+    return _syncDone.get();
+  }
+
   /*
    *  Note, issueSync takes a ZooKeeper (client) object and pass it to doAsyncSync().
    *  The reason we do this is that we want to ensure each new session event is preceded with exactly
@@ -1305,6 +1307,7 @@ public class ZkClient implements Watcher {
 
     KeeperException.Code code = KeeperException.Code.get(callbackHandler.getRc());
     if (code == KeeperException.Code.OK) {
+      _syncDone.set(true);
       LOG.info("sycnOnNewSession with sessionID {} async return code: {} and proceeds", sessionId,
           code);
       return true;
@@ -1323,6 +1326,7 @@ public class ZkClient implements Watcher {
 
     if (SYNC_ON_SESSION) {
       final ZooKeeper zk = ((ZkConnection) getConnection()).getZookeeper();
+      _syncDone.set(false);
       _eventThread.send(new ZkEventThread.ZkEvent("Sync call before new session event of session " + sessionId,
           sessionId) {
         @Override
@@ -2155,6 +2159,21 @@ public class ZkClient implements Watcher {
       _eventThread.start();
 
       LOG.debug("ZkClient created with _uid {}, _eventThread {}", _uid, _eventThread.getId());
+
+      // initiate monitor
+      try {
+        if (_monitorKey != null && !_monitorKey.isEmpty() && _monitorType != null && !_monitorType
+            .isEmpty()) {
+          _monitor =
+              new ZkClientMonitor(_monitorType, _monitorKey, _monitorInstanceName, _monitorRootPathOnly,
+                  _eventThread);
+          _monitor.register();
+        } else {
+          LOG.info("ZkClient monitor key or type is not provided. Skip monitoring.");
+        }
+      } catch (JMException e) {
+        LOG.error("Error in creating ZkClientMonitor", e);
+      }
 
       if (isManagingZkConnection()) {
         zkConnection.connect(watcher);
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -31,11 +31,9 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.management.JMException;
 
-import com.google.common.annotations.VisibleForTesting;
 import org.apache.helix.zookeeper.api.client.ChildrenSubscribeResult;
 import org.apache.helix.zookeeper.constant.ZkSystemPropertyKeys;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
@@ -235,6 +233,13 @@ public class ZkClient implements Watcher {
 
     connect(connectionTimeout, this);
 
+    try {
+      if (_monitor != null) {
+        _monitor.register();
+      }
+    } catch (JMException e){
+      LOG.error("Error in creating ZkClientMonitor", e);
+    }
   }
 
   public List<String> subscribeChildChanges(String path, IZkChildListener listener) {
@@ -2152,15 +2157,10 @@ public class ZkClient implements Watcher {
       _eventThread = new ZkEventThread(zkConnection.getServers());
 
       if (_monitor != null) {
-        _monitor.setAndInitZkEventThreadMonitor(_eventThread);
-      }
-
-      try {
-        if (_monitor != null) {
-          _monitor.register();
+        boolean result = _monitor.setAndInitZkEventThreadMonitor(_eventThread);
+        if (!result) {
+          LOG.error("register _eventThread monitor failed due to an existing one");
         }
-      } catch (JMException e){
-        LOG.error("Error in creating ZkClientMonitor", e);
       }
 
       _eventThread.start();
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientMonitor.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientMonitor.java
@@ -38,9 +38,13 @@ import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMetric;
 import org.apache.helix.monitoring.mbeans.dynamicMBeans.SimpleDynamicMetric;
 import org.apache.helix.monitoring.mbeans.exception.MetricException;
 import org.apache.helix.zookeeper.zkclient.ZkEventThread;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 public class ZkClientMonitor extends DynamicMBeanProvider {
+  private static final Logger LOG = LoggerFactory.getLogger(ZkClientMonitor.class);
+
   public static final String MONITOR_TYPE = "Type";
   public static final String MONITOR_KEY = "Key";
   protected static final String MBEAN_DESCRIPTION = "Helix Zookeeper Client Monitor";
@@ -83,8 +87,22 @@ public class ZkClientMonitor extends DynamicMBeanProvider {
     _expiredSessionCounter = new SimpleDynamicMetric("ExpiredSessionCounter", 0l);
     _dataChangeEventCounter = new SimpleDynamicMetric("DataChangeEventCounter", 0l);
     _outstandingRequestGauge = new SimpleDynamicMetric("OutstandingRequestGauge", 0l);
+
     if (zkEventThread != null) {
-      _zkEventThreadMetric = new ZkThreadMetric(zkEventThread);
+      boolean result = setAndInitZkEventThreadMonitor(zkEventThread);
+      if (!result) {
+        LOG.error("register zkEventThreadMonitor failed due to an existing one.");
+      }
+    }
+
+    for (ZkClientPathMonitor.PredefinedPath path : ZkClientPathMonitor.PredefinedPath.values()) {
+      // If monitor root path only, check if the current path is Root.
+      // Otherwise, add monitors for every path.
+      if (!_monitorRootOnly || path.equals(ZkClientPathMonitor.PredefinedPath.Root)) {
+        _zkClientPathMonitorMap.put(path,
+            new ZkClientPathMonitor(path, _monitorType, _monitorKey, _monitorInstanceName)
+        );
+      }
     }
   }
 
@@ -96,9 +114,12 @@ public class ZkClientMonitor extends DynamicMBeanProvider {
             (monitorKey + (monitorInstanceName == null ? "" : "." + monitorInstanceName)));
   }
 
-  public void setAndInitZkEventThreadMonitor(ZkEventThread zkEventThread) {
+  public synchronized boolean setAndInitZkEventThreadMonitor(ZkEventThread zkEventThread) {
     if (_zkEventThreadMetric == null) {
       _zkEventThreadMetric = new ZkThreadMetric(zkEventThread);
+      return true;
+    } else {
+      return false;
     }
   }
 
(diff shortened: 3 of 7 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -124,6 +125,13 @@
   private PathBasedZkSerializer _pathBasedZkSerializer;
   private ZkClientMonitor _monitor;
 
+  final private String _monitorKey;
```
jiajunwang, line 128: These fields are not helpful to the ZkClient.
Could you try to change the ZkClientMonitor instead so as to avoid these fields? What I am thinking is that, allowing the _eventThread sub-monitor to be added later after the ZkClientMonitor has been constructed. So it can be done in the connect() by referring to _monitor.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/TestHelper.java
```diff
@@ -32,7 +32,7 @@
 
 public class TestHelper {
   private static final Logger LOG = LoggerFactory.getLogger(TestHelper.class);
-  public static final long WAIT_DURATION = 20 * 1000L; // 20 seconds
+  public static final long WAIT_DURATION = 60 * 1000L; // 20 seconds
```
jiajunwang, line 35: 1. the comment is not updated accordingly.
2. Could you please justify the reason we need to triple the wait duration?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -285,18 +285,26 @@ public void testZkClientMonitor()
...
-    zkClient.exists(TEST_ROOT);
+    // Note, we need to wait here for the reason that doAsyncSync() blocks only the zkClient event thread. The main
+    // thread of zkClient would issue exits(TEST_ROOT) without blocking. The return of doAsyncSync() would be asyc
+    // to main thread. doAsyncSync() is a source of 1 read and main thread exists(TEST_ROOT) would be another.
+    TestHelper.verify(()->{
+      return ((org.apache.helix.zookeeper.zkclient.ZkClient)zkClient).getSyncStatus();
+    }, TestHelper.WAIT_DURATION);
+
     Assert.assertEquals((long) beanServer.getAttribute(rootname, "ReadCounter"), 1);
+    //Assert.assertEquals((long) beanServer.getAttribute(rootname, "ReadTotalLatencyCounter"), 1);
```
jiajunwang, line 302: What happen to these 2 checks?

jiajunwang: Do we really need this check? If we ensure after exists(), the read is 2, then we shall be good, right?
Note that to check this read counter one by one, we need to add one method just for this purpose. I think it is an overkill and it is not necessary.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1281,6 +1279,10 @@ protected void doRetry() throws Exception {
         });
   }
 
+  public boolean getSyncStatus() {
```
jiajunwang, line 1282: I have multiple questions.
1. If this is just for test, please make it package-private.
2. Do we really need to check the read count one by one increase?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2156,6 +2160,21 @@ public void connect(final long maxMsToWaitUntilConnected, Watcher watcher)
 
       LOG.debug("ZkClient created with _uid {}, _eventThread {}", _uid, _eventThread.getId());
 
+      // initiate monitor
+      try {
+        if (_monitorKey != null && !_monitorKey.isEmpty() && _monitorType != null && !_monitorType
+            .isEmpty()) {
+          _monitor =
+              new ZkClientMonitor(_monitorType, _monitorKey, _monitorInstanceName, _monitorRootPathOnly,
```
pkuwm, line 2168: We should consider closing the monitor to avoid leakage. If the later part of code zk connection timeout and throws exception, the monitor should also be closed.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/TestHelper.java
kaisun2000: Github zookeeper is slow. See #1268 for profiling result. Using 60 secs does not really hurt anything, as we still poll the condition every 500ms. 

Otherwise, the test can be flaky.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: You are right. line 302 and 303 can/should be removed.  The time value is indeterministic. 

line 301 is fine.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: My thinking is that ZkClientMonitor does not hold any session object, nor thread. Let GC deal with it would be fine. Or do you see otherwise?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: Seems you have concern about this one by increase it one by one https://github.com/apache/helix/pull/1295/files#r474338328

Let me share my thinking here. 

The doAsyncSync() callback finishing is async to main thread here. It can increase the read counter anytime.  Thus, we need TestHelper.verify to wait for this read from sync(). Here, delete line 301  assert is also ok, just wait for 307 having read counter as 2. Same reason, assert 301 is 1 is fine. 

In sum, either way it is fine. But without TestHelper.verify waiting is not fine.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: 1/ Will make this package level by removing public.

2/ See my comment https://github.com/apache/helix/pull/1295/files#r474351453. See if this address the concern.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
lei-xia: Why we move this into connect()? what problem we try to solve here?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
jiajunwang: 2 ways are different because you will need to add code for supporting the check. If you think both ways of checking are fine, let's do the simpler one.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -222,22 +226,16 @@ protected ZkClient(IZkConnection zkConnection, int connectionTimeout, long opera
...
-      } else {
-        LOG.info("ZkClient monitor key or type is not provided. Skip monitoring.");
-      }
-    } catch (JMException e) {
-      LOG.error("Error in creating ZkClientMonitor", e);
+    if (monitorKey != null && !monitorKey.isEmpty() && monitorType != null && !monitorType
+        .isEmpty()) {
+      _monitor =
+          new ZkClientMonitor(monitorType, monitorKey, monitorInstanceName, monitorRootPathOnly,
+              _eventThread);
```
jiajunwang, line 233: Let's give it a more complete change.
1. The constructor will not require _eventThread anymore, it will be null anyway.
2. setAndInitZkEventThreadMonitor shall create the event thread monitor and call updateAttributesInfo() to update the attribute. Re-register the whole monitor is wrong, you will end up either having 2 mbeans or the 2nd register will fail.

jiajunwang: Synced with Kai offline, the 2nd point is fine since the register will be postpone to the connect. Given this, I would prefer making the registration logic closer to the monitor construct logic.

So shall we pass the ZkClientMonitor into connect() to explicitly update it in the call. Then register the monitor after the connect(), but still in the constructor.

pseudo code:

_monitor = new ZkClientMonitor();
// Comment that we need to update the _monitor in the event lock to avoid missing any records about event thread.
connect(connectTimeout, this, _monitor);
_monitor.register()

kaisun2000: This does not seem to work. 

First  it is not necessary to pass _monitor into `connect(connectTimeout, this, _monitor);` as _monitor is instance variable and connect is method. Connect can see it anyway.

Second, the gist to avoid all these racing issue is that we must make sure _monitor is fully constructed and also register-ed before ZkClient either get callback from ZK.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/ZkTestBase.java
```diff
@@ -143,7 +143,11 @@ protected ZkServer startZkServer(final String zkAddress) {
     int port = Integer.parseInt(zkAddress.substring(zkAddress.lastIndexOf(':') + 1));
     ZkServer zkServer = new ZkServer(dataDir, logDir, defaultNameSpace, port);
     System.out.println("Starting ZK server at " + zkAddress);
-    zkServer.start();
+    try {
+      zkServer.start();
+    } catch (Exception e) {
```
pkuwm, line 148: What's the reason to catch exception and print? I guess you want to have the stacktrace for debugging. But please be noted that if exception is thrown, this change will return a zkServer that's not started. I don't think this is what you expected, is it?
I guest keeping it to throw exception so the test would not proceed.

kaisun2000: You are right. Let me re-throw the exception to prevent the test from proceeding.

Without printing it out here, the problem is that Intellij will not print out the exception here. Thus, it seems to be a mystery as why all the test is skipped.

kaisun2000: Revert this part.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -285,18 +285,24 @@ public void testZkClientMonitor()
...
+    // Note, we need to wait here for the reason that doAsyncSync() blocks only the zkClient event thread. The main
+    // thread of zkClient would issue exits(TEST_ROOT) without blocking. The return of doAsyncSync() would be asyc
+    // to main thread. doAsyncSync() is a source of 1 read and main thread exists(TEST_ROOT) would be another.
+    TestHelper.verify(()->{
+      return ((org.apache.helix.zookeeper.zkclient.ZkClient)zkClient).getSyncStatus();
+    }, TestHelper.WAIT_DURATION);
+
     Assert.assertEquals((long) beanServer.getAttribute(rootname, "ReadCounter"), 1);
+
+    zkClient.exists(TEST_ROOT);
```
pkuwm, line 303: Nit: If we need to check exits(), we should assert true? I actually don't think it is necessary as below already checks. But if really check, assert true is needed.
```
if (!_zkClient.exists(TEST_ROOT)) {
       _zkClient.createPersistent(TEST_ROOT, true);
}
```

kaisun2000: This is not to assert exists() return true. This is to make sure the next `Assert.assertEquals((long) beanServer.getAttribute(rootname, "ReadCounter"), 2);` would be true. Namely this trigger a read and get recorded by bean.

This is not functionality added by me. It is basically line 298 in previous version.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: 1. This is to make the code cleaner. It is a good practice for module the code that the private method does not refer to the class private fields directly.

2. The thread monitor is the only one impacted, so if you do the set there, there should be no problem. Mix everything together is chaotic. Ideally, I would suggest splitting the ZK client monitor and ZK thread monitor.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/TestHelper.java
jiajunwang: "the comment is not updated accordingly."

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: Sorry that I cannot make the link work. Maybe because code has been changed. Could you please copy-paste your comment here? I still think this is not necessary.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
jiajunwang: "But without TestHelper.verify waiting is not fine." why is that? You can TestHelper.verify to wait until the count is 2.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: See the discussion above about why think we can't get rid of this getSyncStatus().

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: Unfortunately, you can't or you have race condition in this case, if my understanding is right. Let me illustrate:

doAsycSync(), when the success code returns, it would increase by read counter one from the asyncThread in the ZkClient. 

Note, the main thread running `zkClient.exist(Test_Root )` would increase read counter synchronously from main thread.

The problem is that the doAsynSync() increase to read counter is totally async to the main thread. It is the case that after main thread finish `zkClient.exist(Test_Root )`, the doAsyncSync() would finish for sure. Recent github test actually illustrate this point. Previous we don't realize as it did not happen.

In fact, theoretically, the doAsyncSync() increase can happen a lot later too. 

Let me know if there is other ways you think that we can make the test stable while also no need to use getSyncStatus().

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/TestHelper.java
kaisun2000: Good point, will update this one. Let us first resolve the bigger issue about how to initialize and how to make test stable.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: Maybe I mis-understand what do you mean. Can you draw the code a little bit more specific? 

from the above pseduo code

```
_monitor = new ZkClientMonitor();
// Comment that we need to update the _monitor in the event lock to avoid missing any records about event thread.
connect(connectTimeout, this, _monitor);
_monitor.register()
```

construction of ZkMonitor and ZkThreadMonitor and their registration must be done before the first doAsyncSync() is sent out, or you have race condition of missing some read/write.

In the above pseudo code, `_monitor.register()` is too late, the first sycn() is already out and complete earlier than this register().  This violate the above assertion. 

I may not fully understand your idea. Let us be more specific. As long as it works, I can change it to that way.

jiajunwang: The registration does not need to happen prior. Even not registered, the metric can record data.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
jiajunwang: Let's sync up tomorrow. I didn't get why waiting for the read count = 2 won't work.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: I test it. It seems not true. We need to register the bean in connect() under the lock. Otherwise, depends on timing, you may lose one read.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2152,6 +2150,19 @@ public void connect(final long maxMsToWaitUntilConnected, Watcher watcher)
 
       IZkConnection zkConnection = getConnection();
       _eventThread = new ZkEventThread(zkConnection.getServers());
+
+      if (_monitor != null) {
+        _monitor.setAndInitZkEventThreadMonitor(_eventThread);
+      }
+
+      try {
```
kaisun2000, line 2158: Note, It seems crucial to register here. Or depending on timing, we may lose read of the first sync().

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: removed.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: changed to waiting for 2

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: I debugged with your code, so the ZkClientPathMonitor objects are currently only created in the register() method. So that's the reason we lose the count if register later.

There is a simple fix, we can instantiate all the ZkClientPathMonitors in the ZkClientMonitor constructor. Like this,
    for (ZkClientPathMonitor.PredefinedPath path : 
```
ZkClientPathMonitor.PredefinedPath.values()) {
      // If monitor root path only, check if the current path is Root.
      // Otherwise, add monitors for every path.
      if (!_monitorRootOnly || path.equals(ZkClientPathMonitor.PredefinedPath.Root)) {
        _zkClientPathMonitorMap.put(path,
            new ZkClientPathMonitor(path, _monitorType, _monitorKey, _monitorInstanceName));
      }
    }
```
Then in the register(), just register all the ZkClientPathMonitor objects that are in the _zkClientPathMonitorMap.

The good thing is that it seems to be the right thing to do.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientMonitor.java
```diff
@@ -96,6 +96,12 @@ public static ObjectName getObjectName(String monitorType, String monitorKey,
             (monitorKey + (monitorInstanceName == null ? "" : "." + monitorInstanceName)));
   }
 
+  public void setAndInitZkEventThreadMonitor(ZkEventThread zkEventThread) {
+    if (_zkEventThreadMetric == null) {
+      _zkEventThreadMetric = new ZkThreadMetric(zkEventThread);
```
jiajunwang, line 101: nit, throw Exception if the _zkEventThreadMetric has already been set or the monitor object has already been registered. Otherwise, we may see an unexpected monitor value in the final result.

jiajunwang: BTW, you can check for registration status by checking _objectName with a synchronized lock on the monitor object.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -285,18 +285,20 @@ public void testZkClientMonitor()
     Assert.assertTrue(beanServer.isRegistered(idealStatename));
 
     Assert.assertEquals((long) beanServer.getAttribute(name, "DataChangeEventCounter"), 0);
-    Assert.assertEquals((long) beanServer.getAttribute(name, "StateChangeEventCounter"), 0);
+    Assert.assertEquals((long) beanServer.getAttribute(name, "StateChangeEventCounter"), 1);
     Assert.assertEquals((long) beanServer.getAttribute(name, "ExpiredSessionCounter"), 0);
     Assert.assertEquals((long) beanServer.getAttribute(name, "OutstandingRequestGauge"), 0);
-    // account for doAsyncSync()
-    Assert.assertEquals((long) beanServer.getAttribute(name, "TotalCallbackCounter"), 1);
```
jiajunwang, line 292: Is this value still validatable?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -31,9 +31,11 @@
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
```
jiajunwang, line 34: no need anymore?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -31,9 +31,11 @@
...
 import java.util.concurrent.atomic.AtomicLong;
 import javax.management.JMException;
 
+import com.google.common.annotations.VisibleForTesting;
```
jiajunwang, line 38: With the latest proposal, this one can be removed too.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -285,18 +285,20 @@ public void testZkClientMonitor()
...
 
-    // Test exists
-    Assert.assertEquals((long) beanServer.getAttribute(rootname, "ReadCounter"), 0);
```
jiajunwang, line 295: Actually, can we wait here before "zkClient.exists(TEST_ROOT);" for ReadCounter == 1. Then wait here for 2? So we ensure exist call will increase one counter too.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -285,18 +285,20 @@ public void testZkClientMonitor()
...
-    Assert.assertEquals((long) beanServer.getAttribute(rootname, "ReadTotalLatencyCounter"), 0);
-    Assert.assertEquals((long) beanServer.getAttribute(rootname, "ReadLatencyGauge.Max"), 0);
```
jiajunwang, line 297: If you wait for the counter in the previous line, then here we can still check for

    Assert.assertTrue((long) beanServer.getAttribute(rootname, "ReadTotalLatencyCounter") >= 0);
    Assert.assertTrue((long) beanServer.getAttribute(rootname, "ReadLatencyGauge.Max") >= 0);	    

right?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -285,18 +285,20 @@ public void testZkClientMonitor()
...
+
+    // Assert.assertEquals((long) beanServer.getAttribute(rootname, "ReadCounter"), 2);
+    // wait for both doAysncSync() finish and zkClient.exists(Test_ROOT) finish from 2 different threads.
+    // The condition would be ReadCounter is 2.
+    boolean verifyResult = TestHelper.verify(()->{
+       return (long) beanServer.getAttribute(rootname, "ReadCounter") == 2;
+    }, TestHelper.WAIT_DURATION);
+    Assert.assertTrue(verifyResult, " did not see 2 read yet");
+
     Assert.assertTrue((long) beanServer.getAttribute(rootname, "ReadTotalLatencyCounter") >= 0);
```
jiajunwang, line 302: I propose that we record the ReadTotalLatencyCounter that we get before exists() call. Then we check if this counter is equal to the previous value or is increased here.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -285,18 +285,20 @@ public void testZkClientMonitor()
...
     Assert.assertTrue((long) beanServer.getAttribute(rootname, "ReadLatencyGauge.Max") >= 0);
```
jiajunwang, line 303: If we checked "ReadLatencyGauge.Max" before, then this one can be ignored.

jiajunwang (commented):
I took a deeper look at the test logic. Since this might be the most important test in our code base, let's be careful to not lose any coverage.

For more details, we can discuss in next week's meeting.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientMonitor.java
kaisun2000: What type of exception would be good here?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: fix.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: fixed.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientMonitor.java
kaisun2000: Let me put a return value as false and let the calling layer log error

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: fixed.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: fixed.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: right.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
kaisun2000: no, not realiably I guess.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
kaisun2000: implemeted this way

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientMonitor.java
```diff
@@ -38,9 +38,13 @@
 import org.apache.helix.monitoring.mbeans.dynamicMBeans.SimpleDynamicMetric;
 import org.apache.helix.monitoring.mbeans.exception.MetricException;
 import org.apache.helix.zookeeper.zkclient.ZkEventThread;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 public class ZkClientMonitor extends DynamicMBeanProvider {
+  private static final Logger LOG = LoggerFactory.getLogger(ZkClientMonitor.class);
```
jiajunwang, line 46: _logger exists in the super class. There is no need to add a new one here.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientMonitor.java
```diff
@@ -109,12 +135,8 @@ public DynamicMBeanProvider register() throws JMException {
...
         getObjectName(_monitorType, _monitorKey, _monitorInstanceName));
     for (ZkClientPathMonitor.PredefinedPath path : ZkClientPathMonitor.PredefinedPath.values()) {
-      // If monitor root path only, check if the current path is Root.
-      // Otherwise, add monitors for every path.
-      if (!_monitorRootOnly || path.equals(ZkClientPathMonitor.PredefinedPath.Root)) {
-        _zkClientPathMonitorMap.put(path,
-            new ZkClientPathMonitor(path, _monitorType, _monitorKey, _monitorInstanceName)
-                .register());
+      if (_zkClientPathMonitorMap.get(path) != null)  {
+        _zkClientPathMonitorMap.get(path).register();
```
jiajunwang, line 139: nit, it could be simpler to be:

_zkClientPathMonitorMap.values().stream().foreach(...);

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientPathMonitor.java
```diff
@@ -32,6 +32,9 @@
 import org.apache.helix.monitoring.mbeans.dynamicMBeans.DynamicMetric;
 import org.apache.helix.monitoring.mbeans.dynamicMBeans.HistogramDynamicMetric;
 import org.apache.helix.monitoring.mbeans.dynamicMBeans.SimpleDynamicMetric;
+import org.apache.helix.zookeeper.zkclient.ZkClient;
```
jiajunwang, line 35: Unnecessary?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientMonitor.java
kaisun2000: fixed.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientPathMonitor.java
kaisun2000: removed.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientMonitor.java
kaisun2000: There are many ways to write a loop and now we have stream. 

Still, I will just change it anyway.

jiajunwang (approved):
LGTM! Please address the last comment.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientPathMonitor.java
```diff
@@ -34,6 +34,7 @@
 import org.apache.helix.monitoring.mbeans.dynamicMBeans.SimpleDynamicMetric;
 
 
+
```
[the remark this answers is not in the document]
kaisun2000: removed.

kaisun2000:
This diff is approved. Please help to merge

>ZkClient connects to ZooKeeper client first before monitor bean
initialization and registration. This cause race condition that fails
metrics test.Fix this issue by properly construct the object.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientMonitor.java
```diff
@@ -96,7 +96,7 @@ public ZkClientMonitor(String monitorType, String monitorKey, String monitorInst
 
     for (ZkClientPathMonitor.PredefinedPath path : ZkClientPathMonitor.PredefinedPath.values()) {
       // If monitor root path only, check if the current path is Root.
-      // Otherwise, add monitors for every path.
+      // Otherwise, add monitors for every  path.
```
jiajunwang, line 99: nit, but could you please revert this "touch" change?

jiajunwang:
@kaisun2000 Please run helix-core tests since the change impacts almost all logics.

jiajunwang:
@kaisun2000 you can rerun the test job in Github UI. There is no need to touch code for a re-run.

pkuwm:
I've triggered rerun the workflow. 
FYI, @kaisun2000 This is a button for us to **Re-run jobs**.
image

kaisun2000:
@jiajunwang, @pkuwm, as we looked, my github page does not have this UI button to re-run. Other than that, any other concerns? Or let us merge this in.

pkuwm:
> @jiajunwang, @pkuwm, as we looked, my github page does not have this UI button to re-run. Other than that, any other concerns? Or let us merge this in.

@kaisun2000 I've helped re-run all the jobs.

helix-core/src/test/java/org/apache/helix/integration/TestPartitionMovementThrottle.java
```diff
@@ -215,7 +215,7 @@ public void testANYtypeThrottle() throws InterruptedException {
     DelayedTransition.setDelay(20);
     DelayedTransition.enableThrottleRecord();
 
-    // start another 3 nodes
+    // start another 3 nodes
```
jiajunwang, line 218: Please remove the extra space before merge.

kaisun2000: removed.

kaisun2000:
@jiajunwang, removed the extra space to trigger run from my side. Please help to merge.

Issue #447, opened 2019-09-03
Propose a Trie cluster structure to replace ClusterModel/ClusterContext/OptimalAssignment, etc


i3wangyi:
I give another thought about current Helix's problem statement and existing data models (*ClusterModel*: stores all assignable nodes and replicas, historic best assignments, *ClusterContext*: mostly tracks the metadata of cluster model) and I feel like a tree (or forest) based data structure would best describe and model the problem.  Before I dive into details or implementations of this approach, I'd like to share with you @narendly @jiajunwang and get some early feedback.

##Why tree structure?
It's actually the same concept used in Topology class implementation. The topology of {cluster -> zone -> instance -> partitions} ideally is a N-ary tree, e.g
```
class Node:
    def __init__(self, val):
         self.children = []
         self.val = val
```
or a company's orgchart. 

Given this analogy, the problem can be simply abstracted as 
**Find the best leaf nodes (partitions) assignments so that n-1 nodes (instances) is even distributed in weights**

Further, to support fast query/insert/update, a Trie data structure would be even better since it has a better performance finding the node. Why? If we've created a search query like **"/cluster/{zoneId}/{instanceId}/{PartitionId}"** a Trie could find the node in O(depth of trie) = O(1) since tree depth is fixed = 4, whereas in a simple N-ary tree one have to recursively search all its children until finding the node, time complexity is O(N).

Because Trie supports fast query, we can easily assign one replica node to an instance node in O(1) time. Vice versa, it's the same performance of reverting an assignment.
```
class Trie:
     def __init__(self, val):
         self.children = dict() # id -> next level Trie
         self.val = val
```

With the Trie definition, we can store additional properties (dynamic and static) in Trie class. For example, an instance Trie node can have properties: instance config, instance weight, id etc. A partition Trie node can have properties: partition state, weight, resource, id, etc. 
What's more interesting is we could achieve fine-grained leveled hard/soft constraints. For example, a fault zone Trie node could have a hard constraint as property: the same partitions of different states cannot be located in the same zone trie node.  
```
class ZoneTrie:
    def assign(self, node, partition):
        if (check against all zone level hard constraints passed):
               return self.children[node].assign(node, partition) # check next instance level constraint
        return False
```

One additional feature would benefit us in unit test/benchmark test is the Trie structure can be easily converted to/from a json format string
```
{ 
  "zone id<0>": {
       "instance id <50>": {
           "partition id <0>": {
              "resource": "testDB"
              "state":  "Master"
            }
        },
[...]

PR #1307, opened 2020-08-24
Add TrieClusterTopology for retrieving hierarchical topology

### LINKED ISSUE #1306: Support retrieving hierarchical cluster topology
Helix cluster topology allows users to define multiple levels of domains. There is no restriction on the number of levels or the name of a specific domain. However, we do not have an API to retrieve the hierarchical cluster topology and the instances under each domain.

Implement an API to retrieve hierarchical cluster topology so that users can easily get all the instances for a certain domain.

zhangmeng916:
### Issues

Fixed #1306 

### Description

This PR implements a cluster topology using a trie. The trie is constructed in a way that each non end node represents a domain, and each end node represents an instance. This data structure could help retrieve the topology under any given domain efficiently. 

### Tests

TestTrieClusterTopology.java

helix-core:
[WARNING] Tests run: 1206, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 3,971.319 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[WARNING] Tests run: 1206, Failures: 0, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:06 h
[INFO] Finished at: 2020-09-23T23:08:37-07:00

helix-rest:
known issue: https://github.com/apache/helix/issues/1336
[ERROR] Failures: 
[ERROR]   TestClusterAccessor.testUpdateConfigFields:167 » IndexOutOfBounds Index: 0, Si...
[INFO] 
[ERROR] Tests run: 168, Failures: 1, Errors: 0, Skipped: 25
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  31.376 s
[INFO] Finished at: 2020-09-23T23:49:39-07:00

(Before CI test pass, please copy & paste the result of "mvn test")

### Documentation (Optional)

### Commits

### Code Quality


jiajunwang:
Seems to be similar to the current Topology.java design. Why we need a new class?
If anything missed in Topology.java, we can add new methods there to read the tree and return the desired information.

zhangmeng916:
> Seems to be similar to the current Topology.java design. Why we need a new class?
> If anything missed in Topology.java, we can add new methods there to read the tree and return the desired information.

This class is purely to collect the instance configs and return to the users in the format they want. Helix has no logic involved in this part and we do not do any calculation. We considered topology, and the conclusion is that it will make both more complicated for now. Topology is mainly used for rebalancing purpose, and Helix adds default value if there is anything missing, and also Topology actually only supports two levels, as shown here:   

public enum Types {
    ROOT,
    ZONE,
    INSTANCE
  }

Although other levels keys are inserted in the tree, they're ignored during further calculation. This guideline is hard coded everywhere through that class. If we're going to change that, it's basically rewriting it.
Another big issue is that Topology is tightly coupled with `faultZoneType` in different logic. But the new API totally ignore that field. 
Although in the long run we may consolidate two classes, the prerequisite will be topology can support real multiple level (more than 2) hierarchy without hardcoding anything.
Trie data structure is more suitable for this use case regarding simpleness and efficiency. This class will only be used when users query domain information for the cluster or for any specific domain.

xyuanlu:
Minor correction here. 
I think the enum Type mentioned above is used only when cluster level clusterConfig.Topology is not defined. Otherwise, we still parse the topology definition in cluster config.

In function getClusterTopologySetting:
```
private static ClusterTopologyConfig getClusterTopologySetting(ClusterConfig clusterConfig) {
    ClusterTopologyConfig clusterTopologyConfig = new ClusterTopologyConfig();
       ...
      if (topologyDef != null) {
        // We parse "/zone/rack/sub-rack/instance" into  ["zone", "rack", "sub-rack", "instance"].
      } else {
        // Use default cluster topology definition, i,e. /root/zone/instance
        clusterTopologyConfig.endNodeType = Types.INSTANCE.name();
        clusterTopologyConfig.faultZoneType = Types.ZONE.name();
      }
      ...
  }
```


In my understanding, current Topology.java has similar functions like validateAndOrderDomain and the constructor  TrieClusterTopology. 
Topology::getClusterTopologySetting returns a parsed list of cluster topology as the constructor here. 
There are similar logic of validating the instance topology config aligns with cluster's config. 

IMHO, I personally prefer reusing current code.

zhangmeng916:
> getClusterTopologySetting

The big problem I have using Topology is that the logic handling is different. If I'm going to use it, I'll need to introduce a lot of switch cases, as the main purpose for Topology is for rebalance instead of API response to users. Here're a few examples:
1. if topology is null, it does
clusterTopologyConfig.endNodeType = Types.INSTANCE.name();
clusterTopologyConfig.faultZoneType = Types.ZONE.name();
But I'll throw an exception and return.

2. When an instance topology misses a value, it also fills the default value, but I will need to throw an exception. 

3. When you find an instance topology is not defined, you'll throw an exception of terminating rebalance. This is not something I should have in the log and return to users.

4. With current clusterTopologyConfig.topologyKeyDefaultValue's linkedhashmap, it is not as easy/efficient as a trie to return the topology under each domain. 

In all, the logic of Topology is more specific for rebalancing usage. If I change it, it will become a complicated class that needs to adapt to different usage. I think the correct way in the future should be:
Have a base class that returns the raw data in cluster config and instance config, and different classes extend it for different usage.

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
```diff
@@ -0,0 +1,244 @@
...
+    removeInvalidInstanceConfig(instanceConfigMap);
+    Map<String, Map<String, String>> instanceDomainsMap = new HashMap<>();
+    instanceConfigMap.entrySet().forEach(
+        entry -> instanceDomainsMap.put(entry.getKey(), entry.getValue().getDomainAsMap()));
+
+    for (Map.Entry<String, Map<String, String>> entry : instanceDomainsMap.entrySet()) {
+      TrieNode curNode = _rootNode;
+      String path = "";
+      for (int i = 0; i < _topologyKeys.length; i++) {
+        String key = _topologyKeys[i] + CONNECTOR + entry.getValue().get(_topologyKeys[i]);
```
xyuanlu, line 208: Thanks for the detailed reply.

I think the previous 3 points can be achieved by adding a flag and using non static 'DEFAULT_DOMAIN_PREFIX'. 
For the point 4, 
Please correct me if I am wrong, line 288 in constructTrie will also add a "default" node similar to 'clusterTopologyKeyDefaultValue'. Maybe we could read the tree and bypass default node. 


In my own opinion, I think the question would be, do we construct topology tree in class Topology for these instances? If yes, its better to read from the existing tree. If not, I think it's ok to have another orthogonal data structure representing the hierarchy.

jiajunwang:
> > Seems to be similar to the current Topology.java design. Why we need a new class?
> > If anything missed in Topology.java, we can add new methods there to read the tree and return the desired information.
> 
> This class is purely to collect the instance configs and return to the users in the format they want. Helix has no logic involved in this part and we do not do any calculation. We considered topology, and the conclusion is that it will make both more complicated for now. Topology is mainly used for rebalancing purpose, and Helix adds default value if there is anything missing, and also Topology actually only supports two levels, as shown here:
> 
> public enum Types {
> ROOT,
> ZONE,
> INSTANCE
> }
> 
> Although other levels keys are inserted in the tree, they're ignored during further calculation. This guideline is hard coded everywhere through that class. If we're going to change that, it's basically rewriting it.
> Another big issue is that Topology is tightly coupled with `faultZoneType` in different logic. But the new API totally ignore that field.
> Although in the long run we may consolidate two classes, the prerequisite will be topology can support real multiple level (more than 2) hierarchy without hardcoding anything.
> Trie data structure is more suitable for this use case regarding simpleness and efficiency. This class will only be used when users query domain information for the cluster or for any specific domain.

As we discussed, let's define the detailed requirement first and then refine the java class design.
But to answer some of you comment here, my premature reply would be,
1. I think adding default value logic needs to be delivered to the user as well. They need to understand how Helix locates an instance. It is a Helix API eventually. If the rest API tells the user that this node has no definition, but internally, it is rebalanced as in the default fault zone, then it would be very confusing.
2. Topology.java supports all possible Domain definition, not only 2 levels. Although the other layers are not used by the rebalancer, you can get the full tree structure from Topology.java if simply adding a method that traverses the tree like you what you are doing in this new class. I'm not aware of the necessity of rewriting.
3. faultZoneType is purely optional.
4. I agree that the trie data structure is more preferred. But it seems not a must for this requirement. If it is a normal tree, the get methods can still be developed, right?

lei-xia:
Are we going to add HelixAdmin API changes to this PR?

zhangmeng916:
> Are we going to add HelixAdmin API changes to this PR?

I plan to add in next PR. This one will have complete logic for trie topology. Admin API should be straightforward.

zhangmeng916:
> Are we going to add HelixAdmin API changes to this PR?

@lei-xia I tried to add Helix admin related code in one PR. It would double the PR size, considering all the test cases. I feel it'll make the PR too huge for review purpose. Still think we should separate them. One is for the underlying logic, and the other for user facing API. What do you think?

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  /**
+   * Return the topology of a cluster as a map. The key of the map is the first level of
+   * domain, and the value is a set of string that represents the path to each end node in that
+   * domain. E.g., assume the topology is defined as /group/zone/rack/instance, the result may be {
+   * ["group_0": {"zone_0/rack_0/instance_0", "zone_1/rack_1/instance_1"}], ["group_1": {"zone_1
+   * /rack_1/instance_1", "zone_1/rack_1/instance_2"}]}
+   */
+  public Map<String, Set<String>> getClusterTopology() {
```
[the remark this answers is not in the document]
zhangmeng916: This is the API by default gets the full topology of the cluster. The key of the map will be the first level of domain defined in topology, e.g. zone, faultdomain, etc.

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
```diff
@@ -0,0 +1,244 @@
...
+      }
+    }
+  }
+
+  private static class TrieNode {
+    // A mapping between trie key and children nodes.
+    private Map<String, TrieNode> _children;
+
+    // the complete path/prefix leading to the current node.
+    private final String _path;
```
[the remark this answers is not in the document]
zhangmeng916: This API is for the common case where instances are returned for the fault zone type:
  public Map<String, List<String>> getInstancesUnderFaultZone() {
    return getTopologyUnderDomainType(_faultZoneType);
  }
The name of the instance is implicitly embedded in the `path` field of the Trie node. e.g. the path could be like zone:myzone0/instance:myinstance0, so 'myinstance0' is the instance name.
I didn't see the necessity of marking leaf nodes. Do you think it'll save time in querying? Because we always need to traverse to the leaves anyway.

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
```diff
@@ -0,0 +1,340 @@
...
+    }
+    toRemoveConfig.forEach(entry -> instanceConfigMap.remove(entry));
+  }
+
+  /**
+   * Constructs a trie based on the provided instance config map. It loops through all instance
+   * configs and constructs the trie in a top down manner.
+   */
+  private void constructTrie(Map<String, InstanceConfig> instanceConfigMap) {
+    removeInvalidInstanceConfig(instanceConfigMap);
```
jiajunwang, line 288: How are the callers supposed to know which instance has been removed because the configuration is invalid?
In addition, the removal is directly modifying the input map. So it could dangerous since it may change the caller's map implicitly.

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
```diff
@@ -0,0 +1,340 @@
...
+      }
+    }
+    return results;
+  }
+
+  private void removeInvalidInstanceConfig(Map<String, InstanceConfig> instanceConfigMap) {
+    Set<String> toRemoveConfig = new HashSet<>();
+    for (String instanceName : instanceConfigMap.keySet()) {
+      Map<String, String> domainAsMap = instanceConfigMap.get(instanceName).getDomainAsMap();
+      if (domainAsMap.isEmpty()) {
```
jiajunwang, line 265: nit, this check is not necessary. The else condition will ensure an instance with an empty list to be removed.

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
```diff
@@ -0,0 +1,340 @@
...
+  private final String _faultZoneType;
+
+  public TrieClusterTopology(final List<String> liveNodes,
+      final Map<String, InstanceConfig> instanceConfigMap, ClusterConfig clusterConfig) {
+    if (instanceConfigMap == null || !instanceConfigMap.keySet().containsAll(liveNodes)) {
+      throw new HelixException(String.format("Config for instances %s is not found!",
+          instanceConfigMap == null ? liveNodes : liveNodes.removeAll(instanceConfigMap.keySet())));
+    }
+
+    String topologyDef = clusterConfig.getTopology();
```
jiajunwang, line 60: What if topology awareness is disabled on this cluster? Shall we still throw Exception?

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
```diff
@@ -0,0 +1,340 @@
...
+    if (topologyDef == null) {
+      throw new HelixException(String.format("The topology of cluster %s is empty!",
+          clusterConfig.getClusterName()));
+    }
+    // A list of all keys in cluster topology, e.g., a cluster topology defined as
+    // /group/zone/rack/host will return ["group", "zone", "rack", "host"].
+    _topologyKeys = Arrays.asList(topologyDef.trim().split(DELIMITER)).stream()
+        .filter(str -> !str.isEmpty()).collect(Collectors.toList()).toArray(new String[0]);
```
jiajunwang, line 68: I think we are not requiring such strict formatting elsewhere. For example, spaces are ignored.
So maybe you should also trim here so as to ensure the logic is consistent.

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
```diff
@@ -0,0 +1,340 @@
...
+    if (_topologyKeys.length == 0) {
+      throw new HelixException(String.format("The topology of cluster %s is not correctly defined",
+          clusterConfig.getClusterName()));
+    }
+    _faultZoneType = clusterConfig.getFaultZoneType();
+    _rootNode = new TrieNode(new HashMap<>(), "", "ROOT");
```
jiajunwang, line 74: Can we let constructTrie() return a root which is constructed by the private method internally?
This helps to simplify the code a little bit.

jiajunwang (commented):
To follow up on our previous discussion about why not reusing the existing Topology.java, my thought is that we shall try to reuse it if no conflict logic.
1. The validation is necessary. And we should keep one set of check logic instead of multiple sets to ensure consistency.
2. Although the detailed requirements are not confirmed yet, your new methods are too much for the existing class. But this is not a good reason for creating a new tree class. We can create a new class that contains all the traverse logic to read the tree generated by the Topology.java.
3. Separating the tree constructing logic and the traverse logic (which is more related to the business and more changeable) is a cleaner design anyway.

There are some other detailed comments I noted within the code.

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
zhangmeng916: Please refer to the updated PR. We do not need to topology awareness to be enabled for API purpose.

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
zhangmeng916: Please see the updated PR. I add an invalid config list that customers can retrieve.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -420,6 +420,24 @@ void addStateModelDef(String clusterName, String stateModelDef, StateModelDefini
...
+   * @param clusterName
+   */
+  Map<String, List<String>> getInstancesUnderFaultZone(String clusterName);
+
+  /**
+   * Get all the instances whose domain config is not valid
+   * @param clusterName
+   */
+  List<String> getInvalidInstances(String clusterName);
+
```
jiajunwang, line 440: Didn't we agree to add one method that returns ClusterTopology object here?

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
```diff
@@ -0,0 +1,247 @@
...
+import org.apache.helix.model.ClusterTrie;
+import org.apache.helix.model.InstanceConfig;
+import org.apache.helix.model.TrieNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class ClusterTopology {
+  private static Logger logger = LoggerFactory.getLogger(ClusterTopology.class);
+  private static final String DELIMITER = "/";
```
jiajunwang, line 42: Can we just refer to the DELIMITER and CONNECTOR in ClusterTrie?

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
zhangmeng916: If we return an object, users will still need to parse it by themselves. It's a lot of work if you look at the functions in ClusterTrie. I don't think we want to have this requirement on them. Or you mean users need to call the functions inside ClusterTrie? then they need to know all the available functions? Please let me know.

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
zhangmeng916: I trim the topologyDef, do you mean I need to trim something else?

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
zhangmeng916: There're some offline discussion, and the conclusion is that trie is a good fit for cluster topology, as also mentioned here: https://github.com/apache/helix/issues/447
and should apply to both rebalance and user facing APIs, meaning that we only need one class for both uses. 
In this PR, we will create a generic trie structure with validation and construction logic. Topology APIs will use this structure and do whatever traverse needed. Laster, `topology` logic will also use this trie structure and implement the functions it need. The new trie structure is not fundamentally different from the tree in `topology` now, but with a clearer organization/validation logic and present in a more generic way. `topology` should be able to use it, if something mismatch, we can always come back to adjust the structure.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
zhangmeng916: Discussed offline, and we will only keep one API in HelixAdmin. Updated.

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
```diff
@@ -0,0 +1,248 @@
...
+    });
+    return results;
+  }
+
+  /**
+   * Return all the paths from a TrieNode as a set.
+   * @param node the node from where to collect all the nodes' paths.
+   * @return All the paths under the node.
+   */
+  private Set<String> getPathUnderNode(TrieNode node) {
```
jiajunwang, line 190: Regardless of what I mentioned above, I think you can merge getPathUnderNode() and getStartNodes() to avoid duplicate logic. Some minor change is required, of course. But I think its a better option.

High-level, you need a method that searches for multiple nodes under a certain scope. And another method that gets a certain node (getStartNode).

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
```diff
@@ -0,0 +1,248 @@
...
+      } else {
+        for (TrieNode child : node.getChildren().values()) {
+          nodeStack.push(child);
+        }
+      }
+    }
+    return resultMap;
+  }
+
+  private TrieNode getStartNode(LinkedHashMap<String, String> domain) {
```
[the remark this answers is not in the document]
jiajunwang: And, just curious, where does the "start" come from? The leaf node is a special case. But here the logic is generic, so it is just getNode(), right?

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
```diff
@@ -0,0 +1,248 @@
...
+  /**
+   * Return the topology under a certain domain as a map. The key of the returned map is the next
+   * level domain, and the value is a list of string that represents the path to each end node in
+   * that domain.
+   * @param domain A map defining the domain name and its value, e.g. {["group": "1"], ["zone",
+   *               "2"]}
+   * @return the topology under the given domain, e.g. {["/group:1/zone:2/rack:0": {"/host:0",
+   * "/host:1"}, ["/group:1/zone:2/rack:1": {"/host:2", "/host:3"}]}
+   */
+  private Map<String, List<String>> getTopologyUnderDomain(Map<String, String> domain) {
```
jiajunwang, line 95: We may not need these 2 functions for supporting the public methods.
I think what we want is a query method that accepts a Map of domain type and value pairs. However, the trick is that the value could be a wildcard "*" or just put null. Then we just need to maintain one set of logic. And for sure, no duplicate logic.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,150 @@
...
+
+  public  String getFaultZoneType() {
+    return _faultZoneType;
+  }
+
+  public List<String> getInvalidInstances() {
+    return _invalidInstances;
+  }
+
+  private void removeInvalidInstanceConfig(Map<String, InstanceConfig> instanceConfigMap) {
```
jiajunwang, line 74: Better to put _topologyKeys to the parameter. This is a better style since the private methods will not depend on each other.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,150 @@
...
+      liveNodesCopy.addAll(liveNodes);
+      throw new HelixException(String.format("Config for instances %s is not found!",
+          instanceConfigMap == null ? liveNodes
+              : liveNodesCopy.removeAll(instanceConfigMap.keySet())));
+    }
+  }
+
+  // Note that we do not validate whether topology-aware is enabled or fault zone type is
+  // defined, as they do not block the construction of the trie
+  private void validateClusterConfig(ClusterConfig clusterConfig) {
```
jiajunwang, line 107: nit, just return the _topologyKeys as the return value. So we don't modify the object field secretly inside a "validate" method. Moreover, maybe call it getTopologyDef() would be more straightforward.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,150 @@
...
+      throw new HelixException(String.format("The topology of cluster %s is not correctly defined",
+          clusterConfig.getClusterName()));
+    }
+  }
+
+  /**
+   * Constructs a trie based on the provided instance config map. It loops through all instance
+   * configs and constructs the trie in a top down manner.
+   */
+  private TrieNode constructTrie(Map<String, InstanceConfig> instanceConfigMap) {
```
jiajunwang, line 127: Same here, better to put the _topologyKeys into the parameter list.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,150 @@
...
+                instanceName));
+            _invalidInstances.add(instanceName);
+            break;
+          }
+        }
+      } catch (IllegalArgumentException e) {
+        _invalidInstances.add(instanceName);
+      }
+    }
+    _invalidInstances.forEach(entry -> instanceConfigMap.remove(entry));
```
jiajunwang, line 91: instanceConfigMap.keyset().removeAll(_invalidInstances)

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
jiajunwang: In general, private methods are better to be "independent" with the object. So that maintenance of the class would be much easier.

Suggest to change to
private List<String> getInvalidIstanceConfig(Map<String, InstanceConfig> instanceConfigMap);

Then we can do instanceConfigMap.keyset().removeAll(_invalidInstances) in the caller where the object fields are updated in a centralized place.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,150 @@
...
+  private String[] _topologyKeys;
+  private String _faultZoneType;
+  private List<String> _invalidInstances = new ArrayList<>();
+
+  public ClusterTrie(final List<String> liveNodes,
+      final Map<String, InstanceConfig> instanceConfigMap, ClusterConfig clusterConfig) {
+    validateInstanceConfig(liveNodes, instanceConfigMap);
+    validateClusterConfig(clusterConfig);
+    _faultZoneType = clusterConfig.getFaultZoneType();
+    _rootNode = constructTrie(instanceConfigMap);
```
jiajunwang, line 54: So to summary the following comments, my suggestion would be,

validateInstanceConfig(liveNodes, instanceConfigMap);
_topologyKeys = getTopologyKeys(clusterConfig);
_faultZoneType = clusterConfig.getFaultZoneType();
_invalidInstances = getInvalidInstances(instanceConfigMap);
instanceConfigMap.keyset().removeAll(_invalidInstances);
_rootNode = constructTrie(instanceConfigMap, _topologyKeys);

Would this be cleaner and easier to maintain for the long term?

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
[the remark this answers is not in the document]
zhangmeng916: "start" means where we start to collect the children node information. E.g., if we have the input of (["group":"0"], ["zone":"1"]), then we know that we should get all the children information from node /group:0/zone:1, basically this is the start point. 
And if we have an input as "zone" domainType, then we need to find all the starting nodes for that domainType, and then return all the children. e.g. they can be /group:0/zone:0, /group:0/zone:1, /group:1/zone:0, /group:1/zone:1 etc.

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
zhangmeng916: I can merge these two, but I think it'll be a bit messy to combine the two cases, one is to find all nodes under a certain domain map, basically a branch of the trie, and the other is to find all nodes under a certain domain type, basically all branches of the trie.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,150 @@
...
+
+    for (Map.Entry<String, Map<String, String>> entry : instanceDomainsMap.entrySet()) {
+      TrieNode curNode = rootNode;
+      String path = "";
+      for (int i = 0; i < _topologyKeys.length; i++) {
+        String key = _topologyKeys[i] + CONNECTOR + entry.getValue().get(_topologyKeys[i]);
+        path = path + DELIMITER + key;
+        TrieNode nextNode = curNode.getChildren().get(key);
+        if (nextNode == null) {
+          nextNode = new TrieNode(new HashMap<>(), path, _topologyKeys[i]);
```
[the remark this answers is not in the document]
zhangmeng916: The reason is that we don't know whether there're sub-domains with the same name, e.g. there can be /group:0/zone:0, /group:1/zone:0, /group:2/zone:0, we'll need full path to distinguish between them

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
zhangmeng916: I tried with a null value for domainType, but feel the logic is still very difficult to merge. 
If you look at the two functions:
///
private TrieNode getStartNode(LinkedHashMap<String, String> domainMap) {
    TrieNode curNode = _trieClusterTopology.getRootNode();
    TrieNode nextNode;
    for (Map.Entry<String, String> entry : domainMap.entrySet()) {
      nextNode = curNode.getChildren().get(entry.getKey() + CONNECTOR + entry.getValue());
      if (nextNode == null) {
        throw new IllegalArgumentException(String
            .format("The input domain %s does not have the value %s", entry.getKey(),
                entry.getValue()));
      }
      curNode = nextNode;
    }
    return curNode;
  }

  private List<TrieNode> getStartNodes(String domainType) {
    List<TrieNode> results = new ArrayList<>();
    TrieNode curNode = _trieClusterTopology.getRootNode();
    Deque<TrieNode> nodeStack = new ArrayDeque<>();
    nodeStack.push(curNode);
    while (!nodeStack.isEmpty()) {
      curNode = nodeStack.pop();
      if (curNode.getNodeKey().equals(domainType)) {
        results.add(curNode);
      } else {
        for (TrieNode child : curNode.getChildren().values()) {
          nodeStack.push(child);
        }
      }
    }
    return results;
  }
///
They're not easy to combine, since one is to find a specific node in the trie, while the other is to find all nodes for a certain level across the trie. Thoughts?

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,155 @@
...
+  private String[] getTopologyDef(ClusterConfig clusterConfig) {
+    String[] topologyDef;
+    String topologyDefInConfig = clusterConfig.getTopology();
+    if (topologyDefInConfig == null) {
+      throw new HelixException(String.format("The topology of cluster %s is empty!",
+          clusterConfig.getClusterName()));
+    }
+    // A list of all keys in cluster topology, e.g., a cluster topology defined as
+    // /group/zone/rack/host will return ["group", "zone", "rack", "host"].
+    topologyDef = Arrays.asList(topologyDefInConfig.trim().split(DELIMITER)).stream()
```
jiajunwang, line 119: The same example that I gave Xiaoyuan for the Topology.java change,

"/group/ zone/ rack /host", do we accept this one? With your current logic, I think it will return "["group", " zone", " rack ", "host"]". Please mind the space in the keys. It will cause a mismatch in the later algorithm.

helix-core/src/main/java/org/apache/helix/model/TrieClusterTopology.java
jiajunwang: Commented in your new code. Let me resolve this one.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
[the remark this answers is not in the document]
jiajunwang: I don't get the reason. With your example, it should be very simple to distinguish since these 3 zone:0 nodes are under different parent nodes. And while you are traversing, the algorithm can construct the full path based on parents' information. You can even keep a stack with paths for this purpose.

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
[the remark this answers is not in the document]
jiajunwang: I understand the logic, I'm just asking about the name. Since this method can also be used to get the leaf, I feel getNode() is simpler and more generic. But no strong preference here.

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
jiajunwang: Yeah, I see. Let's keep the current design for now.

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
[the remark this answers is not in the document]
zhangmeng916: Yeah, I agree. Changed the name.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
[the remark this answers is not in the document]
zhangmeng916: Do you suggest to replace the _path in TrieNode with parent node field? I feel the _path is easier for traversal. Is the concern mainly on the memory usage?

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
zhangmeng916: I see what you mean. I trimmed spaces after split. I think space should be allowed. Updated PR.

helix-core/src/main/java/org/apache/helix/model/TrieNode.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public String getPath() {
+    return _path;
+  }
+
+  public String getNodeKey() {
+    return _nodeKey;
+  }
+
+  public void addChildrenMap(Map <String, TrieNode> children) {
```
jiajunwang, line 53: I guess this is not necessary?

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+    return invalidInstances;
+  }
+
+  // Note that we do not validate whether topology-aware is enabled or fault zone type is
+  // defined, as they do not block the construction of the trie
+  private String[] getTopologyDef(ClusterConfig clusterConfig) {
+    String[] topologyDef;
+    String topologyDefInConfig = clusterConfig.getTopology();
+    if (topologyDefInConfig == null) {
```
jiajunwang, line 113: Better to add one more check that, if the topologyDefInConfig does not start with DELIMITER, then it is invalid.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,156 @@
...
+      throw new HelixException(String.format("The topology of cluster %s is empty!",
+          clusterConfig.getClusterName()));
+    }
+    // A list of all keys in cluster topology, e.g., a cluster topology defined as
+    // /group/zone/rack/host will return ["group", "zone", "rack", "host"].
+    topologyDef =
+        Arrays.asList(topologyDefInConfig.trim().split(DELIMITER)).stream().map(str -> str.trim())
```
jiajunwang, line 120: nit, the first tirm becomes unnecessary with the later one in the map().

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,156 @@
...
+  private TrieNode constructTrie(Map<String, InstanceConfig> instanceConfigMap,
+      final String[] topologyKeys) {
+    TrieNode rootNode = new TrieNode("", "ROOT");
+    Map<String, Map<String, String>> instanceDomainsMap = new HashMap<>();
+    instanceConfigMap.entrySet().forEach(
+        entry -> instanceDomainsMap.put(entry.getKey(), entry.getValue().getDomainAsMap()));
+
+    for (Map.Entry<String, Map<String, String>> entry : instanceDomainsMap.entrySet()) {
+      TrieNode curNode = rootNode;
+      String path = "";
```
jiajunwang, line 142: StringBuilder for the performance.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
```diff
@@ -0,0 +1,156 @@
...
+      for (int i = 0; i < topologyKeys.length; i++) {
+        String key = topologyKeys[i] + CONNECTOR + entry.getValue().get(topologyKeys[i]);
+        path = path + DELIMITER + key;
+        TrieNode nextNode = curNode.getChildren().get(key);
+        if (nextNode == null) {
+          nextNode = new TrieNode(path, topologyKeys[i]);
```
jiajunwang, line 148: I guess you just need the full path for the leaf node.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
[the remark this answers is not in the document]
jiajunwang: It's both memory and additional computing for splitting/merging the path String.
If we just put minimum info into trie, then there is no need to merge Strings. And there is no additional memory usage. Then when you read the tree, you don't need to split based on the ":" again. So I feel there would be a lot of saving on both the write and read side.

To be specific, I think the node needs: 1. children map<domain value, TrieNode>, 2. domain Type (optional, since the recorded topology def list can be used to determine the type), 3. domain value (optional too, duplicate with the childrm map key, but could be convient).
Then when you read, you can fill the type according to the depth of the tree and collect the domain value based on the children map key.
The matter is to avoid String operations in which we create new String objects.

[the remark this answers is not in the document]
zhangmeng916: I understand your point. I tried a bit with removing path variable. The problem is that I will need to compose the path when I try to return the results. e.g. if the domain given is "zone":"0", I find the trie node and all its children, and when I return, I'll need to return "group:0/zone:0" as the full path, so I still need to build the string with parent information, there is no other way to know the domains above it. So the path is actually very convenient. What do you think?

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
zhangmeng916: Please see the above comments. Actually have the full path stored in each trie node help save a lot of efforts during query. We don't need to store parents, and don't need to compose the full path when return the map.

helix-core/src/main/java/org/apache/helix/model/ClusterTrie.java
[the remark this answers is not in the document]
jiajunwang: So it either String operation on the read side or on the write side. Maybe only write side would be better since we definitely have more reads.
Let's keep the current design for now. It is an internal method so we can optimize if necessary.

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
```diff
@@ -0,0 +1,248 @@
...
+      if (curNode.getDomainType().equals(domain)) {
+        results.add(curNode);
+      } else {
+        for (TrieNode child : curNode.getChildren().values()) {
+          nodeStack.push(child);
+        }
+      }
+    }
+    return results;
+  }
```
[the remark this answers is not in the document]
jiajunwang: Any thoughts about this comment, please?

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
```diff
@@ -0,0 +1,249 @@
...
+      if (curNode.getNodeKey().equals(domainType)) {
+        results.add(curNode);
+      } else {
+        for (TrieNode child : curNode.getChildren().values()) {
+          nodeStack.push(child);
+        }
+      }
+    }
+    return results;
+  }
```
jiajunwang, line 239: Shall we move them to ClusterTrie? They are useful to read the tree for all reader classes, right?

jiajunwang (approved):
Overall looks great to me.
One last comment, please revisit, thanks.

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
zhangmeng916: Agree. Moved to cluster trie.

helix-core/src/main/java/org/apache/helix/api/topology/ClusterTopology.java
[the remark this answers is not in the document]
zhangmeng916: Yep, I've moved them to clustertrie.

zhangmeng916:
Final commit message:
Add TrieClusterTopology for retrieving hierarchical topology in a cluster.
- Add TrieNode class to define a node in the trie
- Add ClusterTrie class to handle the construction, validation and retrieval of nodes/paths in the trie.
- Add ClusterTopology class to provide different APIs for users to retrieve cluster topology information.
- Add APIs in HelixAdmin to retrieve ClusterTopology of a specific cluster.

jiajunwang:
> This PR is ready to merge. Approved by @jiajunwang
> Final commit message:
> Add TrieClusterTopology for retrieving hierarchical topology in a cluster.
> 
> * Add TrieNode class to define a node in the trie
> * Add ClusterTrie class to handle the construction, validation and retrieval of nodes/paths in the trie.
> * Add ClusterTopology class to provide different APIs for users to retrieve cluster topology information.
> * Add APIs in HelixAdmin to retrieve ClusterTopology of a specific cluster.

Test result, please.

zhangmeng916:
> > This PR is ready to merge. Approved by @jiajunwang
> > Final commit message:
> > Add TrieClusterTopology for retrieving hierarchical topology in a cluster.
> > 
> > * Add TrieNode class to define a node in the trie
> > * Add ClusterTrie class to handle the construction, validation and retrieval of nodes/paths in the trie.
> > * Add ClusterTopology class to provide different APIs for users to retrieve cluster topology information.
> > * Add APIs in HelixAdmin to retrieve ClusterTopology of a specific cluster.
> 
> Test result, please.

@jiajunwang tested and updated.

PR #1328, opened 2020-08-27
Enhance ZkHelixVerifier and its related subclass to take a default wait period

### LINKED ISSUE #1321: Refactor Test BestPossibleExternalViewVerifier
#526 can be mitigated by adding a fixed waiting (cool down) time period in the verifier. However, we don't want to do it in  BestPossibleExternalViewVerifier as  this class is used in production code. Thus, we will derive a sub class which would have this cool down.

Also, enhance some logging in BestPossibleExternalViewVerifier

kaisun2000:
### Issues

Resolve #1321 

### Description

Due to #526, currently we have random sleep() in the test before verifyByPolling(). 
This cause test failure in different environment. Here we enhance ZKHelixVerifier
family with an option in builder to add a waiting period. To keep it backward 
compatible, without enable this waiting option, the code would behave exactly the
same as before.

### Tests

None

Running 

### Documentation (Optional)

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
```diff
@@ -224,12 +224,15 @@ public boolean verifyByZkCallback(long timeout) {
   @Override
   protected synchronized boolean verifyState() {
     try {
+      LOG.debug("Verifier start verifyState at {}", System.currentTimeMillis());
```
lei-xia, line 227: Do you need System.currentTimeMillis()?  Will the log4j print out the timestamp?

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
```diff
@@ -245,13 +248,17 @@ protected synchronized boolean verifyState() {
           return false;
         }
       }
+      
+      LOG.debug("Verifier finished live instances at {}", System.currentTimeMillis());
```
lei-xia, line 252: "finished retrieving live instances"

kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
kaisun2000: good point. Removed time stamp for all of them.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+  /**
+   * Deprecated - please use the Builder to construct this class.
+   * @param zkAddr
+   * @param clusterName
+   * @param resources
+   * @param errStates
+   * @param expectLiveInstances
+   */
+  @Deprecated
```
jiajunwang, line 28: Why are we adding new "Deprecated" methods?

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
```diff
@@ -0,0 +1,133 @@
...
+
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class TestBestPossibleExternalViewVerifier extends BestPossibleExternalViewVerifier {
```
jiajunwang, line 11: License and a comment for the class, please.

jiajunwang: Moreover, my suggestion is that we put it to the test paths, not in the main.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public TestBestPossibleExternalViewVerifier.Builder setZkClient(RealmAwareZkClient zkClient) {
+      _zkClient = zkClient;
+      return this;
+    }
+  }
+
+  @Override
+  public boolean verifyByPolling(long timeout, long period) {
```
jiajunwang, line 125: There are more verify methods, I think we shall add COOL_DOWN to all of them. Otherwise, this verifier will behave inconsistently.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
```diff
@@ -234,10 +237,14 @@ protected boolean verifyByCallback(long timeout, List<ClusterVerifyTrigger> trig
         if (!success) {
           // make a final try if timeout
           success = verifyState();
+          if (!success) {
+            LOG.error("verifyByCallback failed due to timeout, with stack trace {}",
+                Arrays.asList(Thread.currentThread().getStackTrace()));
```
jiajunwang, line 242: Stack trace shall be debug log.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
```diff
@@ -0,0 +1,133 @@
...
+   * @param expectLiveInstances
+   */
+  @Deprecated
+  public TestBestPossibleExternalViewVerifier(RealmAwareZkClient zkClient, String clusterName,
+      Set<String> resources, Map<String, Map<String, String>> errStates,
+      Set<String> expectLiveInstances) {
+    super(zkClient, clusterName, resources, errStates, expectLiveInstances);
+  }
+
+  public static class Builder extends ZkHelixClusterVerifier.Builder<TestBestPossibleExternalViewVerifier.Builder> {
```
jiajunwang, line 49: I had a quick try, I think will work fine. You don't need to copy-paste all code.

  public static class Builder extends BestPossibleExternalViewVerifier.Builder {
    public Builder(String clusterName) {
      super(clusterName);
    }
  }

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
jiajunwang, line 11: TestBestPossibleExternalViewVerifier => BestPossibleExternalViewVerifierWithCoolDown?

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Set;
+
+import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class TestBestPossibleExternalViewVerifier extends BestPossibleExternalViewVerifier {
+  private static Logger LOG = LoggerFactory.getLogger(TestBestPossibleExternalViewVerifier.class);
+  private static int COOL_DOWN = 2 * 1000;
```
jiajunwang, line 13: Make it configurable?

jiajunwang (commented):
Many tests (especially WAGED tests) are using StrictMatchExternalViewVerifier. We need to do the same for that too.

Overall, my suggestion is still to add waiting to the specific test case where need it. This generic cooling down logic does not look reasonable to me. The cooldown shall be with the test logic.

One alternative way is that we add a parameter to the verify method that you can specify a cooldown in the caller logic. Could you please consider this solution?

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
kaisun2000: Good.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
kaisun2000:
```
verify()
verify(long timeout)
verifyByZkCallback()
```

all eventually boils down to `verifyByZkCallback(long timeout)`

`verifyByPolling()` boils down to `verifyByPolling(DEFAULT_TIMEOUT, DEFAULT_PERIOD)`.

So we only need to add cool down here.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
kaisun2000: fixed.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
kaisun2000: done.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
kaisun2000: done.

kaisun2000: changed the path and name. added license,

kaisun2000:
> Many tests (especially WAGED tests) are using StrictMatchExternalViewVerifier. We need to do the same for that too.
> 
> Overall, my suggestion is still to add waiting to the specific test case where need it. This generic cooling down logic does not look reasonable to me. The cooldown shall be with the test logic.
> 
> One alternative way is that we add a parameter to the verify method that you can specify a cooldown in the caller logic. Could you please consider this solution?

I did a search of places using `verifyByPolling()`. There are between 150 to 200 places. 

So shall we have another one called as the following?

```
public boolean verifyByPolling(long timeout, long period, long cooldown) {

```

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
jiajunwang: The timestamp seems not to be removed.

And I really don't like these outputs even as debug. I think they are only useful when debugging. And we shall not add them to the repo.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
```diff
@@ -105,7 +105,7 @@ public BestPossibleExternalViewVerifier(RealmAwareZkClient zkClient, String clus
     _dataProvider = new ResourceControllerDataProvider();
   }
 
-  private BestPossibleExternalViewVerifier(RealmAwareZkClient zkClient, String clusterName,
+  protected BestPossibleExternalViewVerifier(RealmAwareZkClient zkClient, String clusterName,
```
jiajunwang, line 108: This breaks the builder design pattern a little bit.
I'm thinking if we have a better way.

What if we add a configuration to the Verifier Builder called "waitUntilVerify"? It can be a generic configuration for all verifiers even production ones.
1. It is optional, so no backward compatibility issue.
2. It potentially helps our customers too, because they may face the same verify timing issue.
3. It requires the same amount of change as your current design. But no extra class introduced.

Please consider adding it to ZkHelixClusterVerifier.Builder, so all the verifier gets it.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
```diff
@@ -234,6 +236,9 @@ protected boolean verifyByCallback(long timeout, List<ClusterVerifyTrigger> trig
         if (!success) {
           // make a final try if timeout
           success = verifyState();
+          if (!success) {
+            LOG.error("verifyByCallback failed due to timeout");
```
jiajunwang, line 240: nit, keep the same log string style?
"LOG.error("verifier timeout out with timeout {}", timeout);"

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
```diff
@@ -19,6 +19,7 @@
  * under the License.
  */
 
+import java.util.Arrays;
```
jiajunwang, line 22: Remove?

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/TestBestPossibleExternalViewVerifier.java
jiajunwang: But you didn't change "verifyByZkCallback", right?

kaisun2000: I think we don't need to change verifyByZkCallback. To me, it won't help. What is your take?

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
kaisun2000: fixed.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
kaisun2000: refactored.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
```diff
@@ -335,6 +352,11 @@ public B setZkAddr(String zkAddress) {
       return setZkAddress(zkAddress);
     }
 
+    public B setWaitTillVerify(int waitPeriod) {
+      _coolDown = waitPeriod;
```
jiajunwang, line 356: Let's just call it "waitPeriodTillVerify. Cooldown is not an obvious name.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
jiajunwang, line 227: As we discussed, please remove the debug logs if not absolutely necessary for long term usage.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
```diff
@@ -187,6 +190,14 @@ public boolean verifyByZkCallback() {
    * @return
    */
   public boolean verifyByPolling(long timeout, long period) {
```
jiajunwang, line 192: I believe you can and should add the same wait for verifyByZkCallback too.
The logic of verifyByZkCallback is that
1. subscribe to changes.
2. no matter if change happens, verify the state for once.
3. if subscription results in some notification, then verify in parallel.
4. any of the paths succeed, then the verify succeeds.

In this case, we shall wait until the first subscribe to ensure the waitTillVerify config takes effect. This won't cause any problem with verifying logic. Since if the change happens before we subscribe, we are still fine because the first default verify check will be done once at least.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
kaisun2000: removed all

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
kaisun2000: changed

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
kaisun2000: make sense.

jiajunwang (approved):
LGTM!

kaisun2000:
This diff is approve, please help to merge into trunk.

>TestBestPossibleExternalViewVerifier would have a default COOL_DOWN period
before verifyByPolling. Due to #526, currently we have randow sleep in the
test before verifyByPolling. We intend to use TestBestPossibeVerifier
instead in our unit test. While keep using BestPossibleExternalViewVerifier
in production code.

jiajunwang:
> This diff is approve, please help to merge into trunk.
> 
> > TestBestPossibleExternalViewVerifier would have a default COOL_DOWN period
> > before verifyByPolling. Due to #526, currently we have randow sleep in the
> > test before verifyByPolling. We intend to use TestBestPossibeVerifier
> > instead in our unit test. While keep using BestPossibleExternalViewVerifier
> > in production code.

@kaisun2000 please update the comment, it has diverged from the code change quite a lot. Please do take care of the final comment message, once it is in, we don't have a way to modify except reverting the PR.

kaisun2000:
Good point. I did not realize the previous commit message not applying anymore. Let me update it.

kaisun2000:
This diff is approved, please help to merge.

> Due to #526, currently we have random sleep() in the test before verifyByPolling().
This cause test failure in different environment. Here we enhance ZKHelixVerifier
family with an option in builder to add a waiting period. To keep it backward
compatible, without enable this waiting option, the code would behave exactly the
same as before.

PR #1344, opened 2020-09-03
Reduce unnecessary getChildren call in subscribeForChanges

### LINKED ISSUE #1343: Unnecessary getChildren call in subscribeForChanges
In CallbackHanlder's `subscribeForChanges()`, children names need to be fetched. `subscribeChildChange(path, callbackType)` does actually have the children list, but it doesn't return, and later `getChildren()` is called again. It wastes one zk call: not only wasting time in handling a callback, but also adds more read pressure to zk server.

We could use the return result from subscribeChildChange() and save one `getChildren()` call.

pkuwm:
### Issues

Fixes #1343 

### Description

In CallbackHanlder's subscribeForChanges(), children names need to be fetched. subscribeChildChange(path, callbackType) does actually have the children list, but it doesn't return, and later getChildren() is called again. It wastes one zk call: not only wasting time in handling a callback, but also adds more read pressure to zk server.

We could use the return result from subscribeChildChange() and save one getChildren() call.

This PR also adds event types logging and removes two logs. When debugging, we can actually use the event types to determine the code logic. No need to print those two logs, which saves logging overhead when there are intensive callbacks.

### Tests

```
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestCardDealingAdjustmentAlgorithmV2.testAlgorithmConstructor:127 expected:<9> but was:<6>
[ERROR]   Run 1: TestCardDealingAdjustmentAlgorithmV2.testComputeMappingForDifferentReplicas:273 Total movements: 4 != expected 8, replica: 1
[ERROR]   Run 2: TestCardDealingAdjustmentAlgorithmV2.testComputeMappingForDifferentReplicas:273 Total movements: 4 != expected 8, replica: 2
[ERROR]   Run 3: TestCardDealingAdjustmentAlgorithmV2.testComputeMappingForDifferentReplicas:273 Total movements: 14 != expected 21, replica: 3
[INFO]   Run 4: PASS
[INFO]   Run 5: PASS
[INFO]
[INFO]
[ERROR] Tests run: 1204, Failures: 2, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:12 h
[INFO] Finished at: 2020-09-24T16:30:00-07:00


[INFO] Tests run: 27, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.483 s - in org.apache.helix.controller.strategy.crushMapping.TestCardDealingAdjustmentAlgorithmV2
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 27, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  7.198 s
[INFO] Finished at: 2020-09-24T17:35:52-07:00
```
### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -538,7 +538,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
     }
   }
 
-  private void subscribeChildChange(String path, NotificationContext.Type callbackType) {
+  private List<String> subscribeChildChange(String path, NotificationContext.Type callbackType) {
     if (callbackType == NotificationContext.Type.INIT
         || callbackType == NotificationContext.Type.CALLBACK) {
       if (logger.isDebugEnabled()) {
@@ -559,12 +559,15 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
       if (!childrenSubscribeResult.isInstalled()) {
         logger.info("CallbackHandler {} subscribe data path {} failed!", _uid, path);
       }
+      return childrenSubscribeResult.getChildren();
     } else if (callbackType == NotificationContext.Type.FINALIZE) {
       logger.info("CallbackHandler{}, {} unsubscribe child-change. path: {}, listener: {}",
           _uid ,_manager.getInstanceName(), path, _listener);
 
       _zkClient.unsubscribeChildChanges(path, this);
     }
+
+    return null;
   }
 
   private void subscribeDataChange(String path, NotificationContext.Type callbackType) {
@@ -597,23 +600,21 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
 
   private void subscribeForChanges(NotificationContext.Type callbackType, String path,
       boolean watchChild) {
-    logger.info("CallbackHandler {} Subscribing changes listener to path: {}, type: {}, listener: {}",
-        _uid, path, callbackType, _listener);
+    logger.info("CallbackHandler {} subscribing changes listener to path: {}, callback type: {}, "
+            + "event types: {}, listener: {}, watchChild: {}",
+        _uid, path, callbackType, _eventTypes, _listener, watchChild);
 
     long start = System.currentTimeMillis();
     if (_eventTypes.contains(EventType.NodeDataChanged)
         || _eventTypes.contains(EventType.NodeCreated)
         || _eventTypes.contains(EventType.NodeDeleted)) {
-      logger.info("CallbackHandler{} Subscribing data change listener to path: {}", _uid, path);
+      logger.info("CallbackHandler {} subscribing data change listener to path: {}", _uid, path);
       subscribeDataChange(path, callbackType);
     }
 
     if (_eventTypes.contains(EventType.NodeChildrenChanged)) {
-      logger.info("CallbackHandler{}, Subscribing child change listener to path: {}", _uid, path);
-      subscribeChildChange(path, callbackType);
+      List<String> children = subscribeChildChange(path, callbackType);
       if (watchChild) {
-        logger.info("CallbackHandler{}, Subscribing data change listener to all children for path: {}", _uid, path);
-
         try {
           switch (_changeType) {
             case CURRENT_STATE:
@@ -633,11 +634,10 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
                 if (bucketSize > 0) {
                   // subscribe both data-change and child-change on bucketized parent node
                   // data-change gives a delete-callback which is used to remove watch
-                  subscribeChildChange(childPath, callbackType);
+                  List<String> bucketizedChildNames = subscribeChildChange(childPath, callbackType);
                   subscribeDataChange(childPath, callbackType);
 
                   // subscribe data-change on bucketized child
-                  List<String> bucketizedChildNames = _zkClient.getChildren(childPath);
                   if (bucketizedChildNames != null) {
                     for (String bucketizedChildName : bucketizedChildNames) {
                       String bucketizedChildPath = childPath + "/" + bucketizedChildName;
@@ -651,10 +651,14 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
               break;
             }
             default: {
-              List<String> childNames = _zkClient.getChildren(path);
-              if (childNames != null) {
-                for (String childName : childNames) {
-                  String childPath = path + "/" + childName;
+              // When callback type is FINALIZE, subscribeChildChange doesn't
+              // get children list. We need to read children to unsubscribe data change listeners.
+              if (callbackType == Type.FINALIZE) {
+                children = _zkClient.getChildren(path);
+              }
+              if (children != null) {
+                for (String child : children) {
+                  String childPath = path + "/" + child;
                   subscribeDataChange(childPath, callbackType);
                 }
               }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -538,6 +538,11 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
     }
   }
 
+  /*
+   * If callback type is INIT or CALLBACK, subscribes child change listener to the path
+   * and returns the path's children names. The children list might be null when the path
+   * doesn't exist or callback type is other than INIT/CALLBACK.
+   */
   private List<String> subscribeChildChange(String path, NotificationContext.Type callbackType) {
     if (callbackType == NotificationContext.Type.INIT
         || callbackType == NotificationContext.Type.CALLBACK) {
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -533,7 +533,7 @@ public void invoke(NotificationContext changeContext) throws Exception {
     }
   }
 
-  private void subscribeChildChange(String path, NotificationContext.Type callbackType) {
+  private List<String> subscribeChildChange(String path, NotificationContext.Type callbackType) {
     if (callbackType == NotificationContext.Type.INIT
```
xyuanlu, line 537: I feel like we could create an empty list here. So we won’t return a null and no need to do the nullptr check later.

pkuwm: I thought about it. There is still a possibility that `childrenSubscribeResult.getChildren()` returns `null`. The caller still needs to do the `null` check. So I just make it consistent with `zkClient.subscribeChildChanges` and let the caller determine the behavior.

I also thought about using `Optional`, but it doesn't seem that would add too much value. And `Optional` will add more overhead than `null`. The original code also does null check, so for now I'd keep the change minimum. Later we will still have a better design and hopefully will make the code cleaner.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -646,10 +645,9 @@ private void subscribeForChanges(NotificationContext.Type callbackType, String p
               break;
             }
             default: {
-              List<String> childNames = _zkClient.getChildren(path);
-              if (childNames != null) {
-                for (String childName : childNames) {
-                  String childPath = path + "/" + childName;
+              if (children != null) {
```
xyuanlu, line 648: There is a slightly behavior difference here. Originally before this change, if type is Finalize, we still iterate trough all node and do `subscribeDataChange `. 
Now we do not return a children list when finalize (line 565) so we do not do `subscribeDataChange` for those nodes.  
This behavior difference cause unit test `TestListenerCallbackBatchMode` fail on my side. Did you run in to this problem?
Thx.

pkuwm: Yeah, I was aware of the failed test and the code in `subscribeChildChange()` for `NotificationContext.Type.FINALIZE`.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang: Is it possible that you just return ChildrenSubscribeResult?
Then if _isInstalled is false, no one will read the children. Otherwise, if _isInstalled, then the list should never be null.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -592,8 +595,8 @@ private void subscribeForChangesAsyn(NotificationContext.Type callbackType, Stri
 
   private void subscribeForChanges(NotificationContext.Type callbackType, String path,
       boolean watchChild) {
-    logger.info("Subscribing changes listener to path: {}, type: {}, listener: {}", path,
-        callbackType, _listener);
+    logger.info("Subscribing changes listener to path: {}, callback type: {}, event types: {}, "
```
jiajunwang, line 598: For debug, please add new fields to the end of the string. And don't change the existing content if they are not hurt too much. Otherwise, the dev needs to switch the code version back and forth to double-check before debugging the older version deployment.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -604,11 +607,8 @@ private void subscribeForChanges(NotificationContext.Type callbackType, String p
     }
 
     if (_eventTypes.contains(EventType.NodeChildrenChanged)) {
-      logger.info("Subscribing child change listener to path: {}", path);
```
jiajunwang, line 607: I personally think these 2 info logs are still desired. Especially we are changing the subscription mechanism.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -646,10 +645,14 @@ private void subscribeForChanges(NotificationContext.Type callbackType, String p
               break;
             }
             default: {
-              List<String> childNames = _zkClient.getChildren(path);
-              if (childNames != null) {
-                for (String childName : childNames) {
-                  String childPath = path + "/" + childName;
+              // When callback type is FINALIZE or PERIODIC_REFRESH, subscribeChildChange doesn't
+              // get children list. We need to read children to unsubscribe data change listeners.
+              if (callbackType != Type.INIT && callbackType != Type.CALLBACK) {
```
jiajunwang, line 650: As I mentioned above, if the return value is ChildrenSubscribeResult, then you just need to check "_isInstalled" for the following logic.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -646,10 +645,14 @@ private void subscribeForChanges(NotificationContext.Type callbackType, String p
...
-              if (childNames != null) {
-                for (String childName : childNames) {
-                  String childPath = path + "/" + childName;
+              // When callback type is FINALIZE, subscribeChildChange doesn't
+              // get children list. We need to read children to unsubscribe data change listeners.
+              if (callbackType == Type.FINALIZE) {
+                children = _zkClient.getChildren(path);
+              }
+              if (children != null) {
+                for (String child : children) {
```
kaisun2000, line 654: Parallel this using Asyc should bring more performance gain, right? why don't we do it here?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -554,12 +554,15 @@ private void subscribeChildChange(String path, NotificationContext.Type callback
       if (!childrenSubscribeResult.isInstalled()) {
         logger.info("CallbackHandler {} subscribe data path {} failed!", this, path);
       }
+      return childrenSubscribeResult.getChildren();
```
kaisun2000, line 557: I think the main issue here is that line 551
`_zkClient.subscribeChildChanges(path, this, callbackType != Type.INIT);` we have two code path.

if callbackType is init, the parent path of callback handler may not be created. Thus, the return would be null, or empty? Please verify.

The potentially issue is that note later at `default`, we re-use the return children list from here. 
Note, original code retrieve children one more time and it may retrieve node by then. Here, it may not.

This is different from previous logic. Not sure if this would bring a bug or not.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
pkuwm: I also thought about it. But from the code, if `_isInstalled == true`, it is still possible that the children list is null: 
```
        try {
          return getChildren(path, true);
        } catch (ZkNoNodeException e) {
          // ignore, the "exists" watch will listen for the parent node to appear
          LOG.info("watchForChilds path not existing:{} skipWatchingNodeNoteExist: {}",
              path, skipWatchingNonExistNode);
        }
        return null;
```
So we could not determine if children list is null or not based on `_isInstalled`.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
pkuwm: Replied above.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -646,10 +652,14 @@ private void subscribeForChanges(NotificationContext.Type callbackType, String p
...
             default: {
-              List<String> childNames = _zkClient.getChildren(path);
-              if (childNames != null) {
-                for (String childName : childNames) {
-                  String childPath = path + "/" + childName;
+              // When callback type is FINALIZE, subscribeChildChange doesn't
+              // get children list. We need to read children to unsubscribe data change listeners.
+              if (callbackType == Type.FINALIZE) {
+                children = _zkClient.getChildren(path);
+              }
```
jiajunwang, line 659: As we discussed, I think this check is not necessary. The later check for children != null is good enough.
If the node was not created and then create during this callback processing, then we will have another callback to process the event since the watcher has been added.

pkuwm: Moved the logic to subscribeChildChange: for FINALIZE, it also returns the children.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -538,7 +538,12 @@ public void invoke(NotificationContext changeContext) throws Exception {
     }
   }
 
-  private void subscribeChildChange(String path, NotificationContext.Type callbackType) {
+  /*
+   * If callback type is INIT or CALLBACK, subscribes child change listener to the path
+   * and returns the path's children names. The children list might be null when the path
+   * doesn't exist or callback type is other than INIT/CALLBACK/FINALIZE.
```
jiajunwang, line 544: The current logic will return a list even the type is not INIT/CALLBACK/FINALIZE, right? Although I guess it is expected.
1. Please fix the comment.
2. The method name becomes misleading somehow with this change. I guess it should be called getChildrenAndSubscribeChildChange. Obviously, not a good name. But feel free to change if you have a better idea. This is just a nice to have thing.

jiajunwang (approved):

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
pkuwm: Resolved comment.
Since we also have `subscribeChildChanges()` in ZkClient that returns children, so I'd like to keep it as it is to be consistent. Besides, I don't have a better name for this and I think it is OK to keep it.

pkuwm:


In CallbackHanlder's subscribeForChanges(), children names need to be fetched. subscribeChildChange(path, callbackType) does actually have the children list, but it doesn't return, and later getChildren() is called again. It wastes one zk call: not only wasting time in handling a callback, but also adds more read pressure to zk server.

This commit uses the return result from subscribeChildChange() and saves getChildren() call.

PR #1359, opened 2020-09-11
Don't output the best possible state calculate result if the result is not valid.

### LINKED ISSUE #1358: Resource metric for the rebalance state could be overwritten unexpectedly
### Describe the bug
The rebalance state metric for each resource is the way Helix provides to identify if a resource is being rebalanced normally. We identify an issue that may lead this metric to report wrong information.
In detail, even though the bestpossiblestate calculation fails, the state might remain to be NORMAL.

### To Reproduce
1. Use DelayedRebalance since it may not throw an exception even the rebalance is not possible.
2. Try to rebalance a resource with no available instances.
3. Check the ResourceMonitor attribute RebalanceState, it would be showing BEST_POSSIBLE_CAL_FAILED and then be overwritten to NORMAL by the later pipeline stage.

### Expected behavior
The BEST_POSSIBLE_CAL_FAILED state should remain in the ResourceMonitor until enough instances appear and the resource is normally rebalanced.

jiajunwang:
### Issues

Fixes #1358 

### Description

The invalid result may mislead the following stages in the pipeline and cause problems. For example, this change fixes the issue that resource monitor rebalance state may be inaccurately reported.

### Tests

Modified the existing test to cover change.

[INFO] Results:
[INFO] 
[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 1200, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time: 01:14 h
[INFO] Finished at: 2020-09-10T17:51:36-07:00

Rerun

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 8.305 s - in org.apache.helix.integration.task.TestDeleteJobFromJobQueue
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 14.473 s
[INFO] Finished at: 2020-09-10T17:56:04-07:00

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
@@ -347,36 +347,39 @@ public class BestPossibleStateCalcStage extends AbstractBaseStage {
       LogUtil.logError(logger, _eventId, "Error computing assignment for resource " + resourceName
           + ". no rebalancer found. rebalancer: " + rebalancer + " mappingCalculator: "
           + mappingCalculator);
-    }
-
-    if (rebalancer != null && mappingCalculator != null) {
-      ResourceAssignment partitionStateAssignment = null;
+    } else {
       try {
         HelixManager manager = event.getAttribute(AttributeName.helixmanager.name());
         rebalancer.init(manager);
         idealState =
             rebalancer.computeNewIdealState(resourceName, idealState, currentStateOutput, cache);
 
-        output.setPreferenceLists(resourceName, idealState.getPreferenceLists());
+        // Check if calculation is done successfully
+        if (!checkBestPossibleStateCalculation(idealState)) {
+          LogUtil.logWarn(logger, _eventId,
+              "The calculated idealState is not valid, resource: " + resourceName);
+          return false;
+        }
 
         // Use the internal MappingCalculator interface to compute the final assignment
         // The next release will support rebalancers that compute the mapping from start to finish
-        partitionStateAssignment = mappingCalculator
+        ResourceAssignment partitionStateAssignment = mappingCalculator
             .computeBestPossiblePartitionState(cache, idealState, resource, currentStateOutput);
 
         if (partitionStateAssignment == null) {
           LogUtil.logWarn(logger, _eventId,
-              "PartitionStateAssignment is null, resource: " + resourceName);
+              "The calculated partitionStateAssignment is null, resource: " + resourceName);
           return false;
         }
 
+        // Set the calculated result to the output.
+        output.setPreferenceLists(resourceName, idealState.getPreferenceLists());
         for (Partition partition : resource.getPartitions()) {
           Map<String, String> newStateMap = partitionStateAssignment.getReplicaMap(partition);
           output.setState(resourceName, partition, newStateMap);
         }
 
-        // Check if calculation is done successfully
-        return checkBestPossibleStateCalculation(idealState);
+        return true;
       } catch (HelixException e) {
         // No eligible instance is found.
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -347,36 +347,39 @@ private boolean computeSingleResourceBestPossibleState(ClusterEvent event,
...
-        partitionStateAssignment = mappingCalculator
+        ResourceAssignment partitionStateAssignment = mappingCalculator
             .computeBestPossiblePartitionState(cache, idealState, resource, currentStateOutput);
 
         if (partitionStateAssignment == null) {
           LogUtil.logWarn(logger, _eventId,
-              "PartitionStateAssignment is null, resource: " + resourceName);
+              "The calculated partitionStateAssignment is null, resource: " + resourceName);
           return false;
         }
```
narendly, line 373: Question: if `checkBestPossibleStateCalculation(idealState)` is true, will we ever get a null `partitionStateAssignment`?

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -347,36 +347,39 @@ private boolean computeSingleResourceBestPossibleState(ClusterEvent event,
...
 
+        // Set the calculated result to the output.
+        output.setPreferenceLists(resourceName, idealState.getPreferenceLists());
```
narendly, line 376: Is this to make sure if we ever get a null value for partitionStateAssignment, we don't set anything in the output? 

I wonder if this might cause a behavior change. Before this change, did `output` get populated with some preferenceList value (maybe the one from the previous, last successful pipeline)? After this change, if the `computeBestPossiblePartitionState` returns a `null`, then output will start having nothing, which is a different behavior from before this change?

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
jiajunwang: This depends on the mapping calculator implementation. There is no guarantee. So we'd better keep the check.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
jiajunwang: Yeah, this is a behavior change. But the original behavior is the problem here. If any value is set in the output, then the later stage will just use without any doubt. This is very dangerous.

I checked later stages like intermediate state calc, and message generating, they can handle a null output gracefully. But they cannot handle the output with the wrong information gracefully.=

kaisun2000: It seems to me that the original code, `return checkBestPossibleStateCalculation(idealState);` would still return false. Does it feed the wrong info in `output` to later stage? 

The main thing is that the new code fixed the issue of missing metric for invalid idealState.

jiajunwang: The original code returns false while the output is filled with an empty mapping. However, this empty mapping is misleading.

kaisun2000 (approved):
Appoved as there seems concerns are addressed.

jiajunwang:
This PR has been approved and ready to be merged.

PR #1362, opened 2020-09-14
Move ST message to state validation from executing phase to scheduling phase.

### LINKED ISSUE #1361: Move ST message to state validation from executing phase to scheduling phase.
Currently we do all the state validation (the from state should be same current state and to state should be different to current state) for state transition message when the message is being executed in HelixStateTransitionHandler.

We should move the to state validation to the scheduling stage because of the following reasons:
1. It is a no-op operation, thus we should not create a thread to execute it.
2. The lag between scheduling and executing could add additional latency for controller and eventually application latency.

xyuanlu:
### Issues

Resolve #1361

### Description

Currently we do all the state validation (the from state should be same current state and to state should be different to current state) for state transition message when the message is being executed in HelixStateTransitionHandler.

We should move the to state validation to the scheduling stage because of the following reasons:
1. It is a no-op operation, thus we should not create a thread to execute it.
2. The lag between scheduling and executing could add additional latency for controller and eventually application latency.

### Tests

TestStateTransitionHandlerFactory.testStaledMessage

```
[INFO] Results:                                                                                                                                                                                                                                                                                        
[INFO]                                                                                                                                                                                                                                                                                                 
[ERROR] Failures:                                                                                                                                                                                                                                                                                      
[ERROR]   TestHelixAdminCli.testDeactivateCluster:604 » ZkClient Failed to delete /TestH...                                                                                                                                                                                                            
[INFO]                                                                                                                                                                                                                                                                                                 
[ERROR] Tests run: 1213, Failures: 2, Errors: 0, Skipped: 2                                                                                                                                                                                                                                            
[INFO]                                                                                                                                                                                                                                                                                                 
[INFO] BUILD FAILURE                                                                                                                                                                                                                                                                                   
[INFO] Total time:  01:07 h                                                                                                                                                                                                                                                                            
[INFO] Finished at: 2020-10-06T21:42:24-07:00                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                           
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 28.41 s - in org.apache.helix.tools.TestHelixAdminCli
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 60.272 s - in org.apache.helix.integration.TestPauseSignal


```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -112,7 +112,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
     // Set start time right before invoke client logic
     _currentStateDelta.setStartTime(_message.getPartitionName(), System.currentTimeMillis());
 
-    Exception err = isMessageStaled(false /*inSchedulerCheck*/);
+    Exception err = validateStaleMessage(false /*inSchedulerCheck*/);
 
     if (err != null) {
       _statusUpdateUtil
@@ -443,7 +443,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
   }
 
   // Verify the fromState and current state of the stateModel.
-  public Exception isMessageStaled(boolean inSchedulerCheck) {
+  public Exception validateStaleMessage(boolean inSchedulerCheck) {
     String fromState = _message.getFromState();
     String toState = _message.getToState();
     String partitionName = _message.getPartitionName();
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -69,6 +69,16 @@ public class HelixStateTransitionHandler extends MessageHandler {
     }
   }
 
+  public static class StaleMessageValidateResult {
+    public boolean isValid;
+    public Exception exception;
+
+    StaleMessageValidateResult(Exception exp) {
+      exception = exp;
+      isValid = exception == null;
+    }
+  }
+
   private static final Logger logger = LoggerFactory.getLogger(HelixStateTransitionHandler.class);
   private final StateModel _stateModel;
   StatusUpdateUtil _statusUpdateUtil;
@@ -112,12 +122,13 @@ public class HelixStateTransitionHandler extends MessageHandler {
     // Set start time right before invoke client logic
     _currentStateDelta.setStartTime(_message.getPartitionName(), System.currentTimeMillis());
 
-    Exception err = staleMessageValidator();
-    if (err != null) {
+    StaleMessageValidateResult err = staleMessageValidator();
+    if (!err.isValid) {
       _statusUpdateUtil
-          .logError(_message, HelixStateTransitionHandler.class, err.getMessage(), _manager);
-      logger.error(err.getMessage());
-      throw err;
+          .logError(_message, HelixStateTransitionHandler.class, err.exception.getMessage(),
+              _manager);
+      logger.error(err.exception.getMessage());
+      throw err.exception;
     }
 
     // Reset the REQUESTED_STATE property if it exists.
@@ -441,7 +452,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
   }
 
   // Verify the fromState and current state of the stateModel.
-  public Exception staleMessageValidator() {
+  public StaleMessageValidateResult staleMessageValidator() {
     String fromState = _message.getFromState();
     String toState = _message.getToState();
     String partitionName = _message.getPartitionName();
@@ -464,7 +475,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
           "Current state of stateModel does not match the fromState in Message, CurrentState: %s, Message: %s->%s, Partition: %s, from: %s, to: %s",
           state, fromState, toState, partitionName, _message.getMsgSrc(), _message.getTgtName()));
     }
-    return err;
+    return new StaleMessageValidateResult(err);
   }
 
   @Override
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -107,17 +107,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
         + _message.getToState() + ", relayedFrom: " + _message.getRelaySrcHost());
 
     HelixDataAccessor accessor = _manager.getHelixDataAccessor();
-
     String partitionName = _message.getPartitionName();
-    String fromState = _message.getFromState();
-    String toState = _message.getToState();
-
-    // Verify the fromState and current state of the stateModel
-    // getting current state from state model will provide most up-to-date
-    // current state. In case current state is null, partition is in initial
-    // state and we are setting it in current state
-    String state = _stateModel.getCurrentState() != null ? _stateModel.getCurrentState()
-        : _currentStateDelta.getState(partitionName);
 
     // Set start time right before invoke client logic
     _currentStateDelta.setStartTime(_message.getPartitionName(), System.currentTimeMillis());
@@ -452,12 +442,19 @@ public class HelixStateTransitionHandler extends MessageHandler {
 
   }
 
+  // Verify the fromState and current state of the stateModel.
   public Exception isMessageStaled() {
-    String state = _stateModel.getCurrentState() != null ? _stateModel.getCurrentState()
-        : _currentStateDelta.getState(_message.getPartitionName());
     String fromState = _message.getFromState();
     String toState = _message.getToState();
     String partitionName = _message.getPartitionName();
+
+
+    // state in _currentStateDelta uses current state from state model. It has the
+    // most up-to-date. current state. In case currentState in stateModel is null,
+    // partition is in initial state and we using it as current state.
+    // Defined in HelixStateMachineEngine.
+    String state =  _currentStateDelta.getState(partitionName);
+
 
     //String err = null;
 
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -113,7 +113,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
     _currentStateDelta.setStartTime(_message.getPartitionName(), System.currentTimeMillis());
 
     try {
-      validateStaleMessage(false /*isPreCheck*/);
+      validateStaleMessage();
     } catch (Exception err) {
       _statusUpdateUtil
           .logError(_message, HelixStateTransitionHandler.class, err.getMessage(), _manager);
@@ -443,7 +443,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
   }
 
   // Verify the fromState and current state of the stateModel.
-  private void validateStaleMessage (boolean isPreCheck) throws Exception {
+  public void validateStaleMessage() throws Exception {
     String fromState = _message.getFromState();
     String toState = _message.getToState();
     String partitionName = _message.getPartitionName();
@@ -459,7 +459,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
       throw new HelixDuplicatedStateTransitionException(String
           .format("Partition %s current state is same as toState (%s->%s) from message.",
               partitionName, fromState, toState));
-    } else if (!isPreCheck && fromState != null && !fromState.equals("*") && !fromState
+    } else if (fromState != null && !fromState.equals("*") && !fromState
         .equalsIgnoreCase(state)) {
       // If current state is neither toState nor fromState in message, there is a problem
       throw new HelixStateMismatchException(String.format(
@@ -468,9 +468,6 @@ public class HelixStateTransitionHandler extends MessageHandler {
     }
   }
 
-  public void precheckForStaleMessage() throws Exception {
-    validateStaleMessage(true /*isPreCheck*/);
-  }
 
   @Override
   public void onTimeout() {
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -122,18 +122,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
     // Set start time right before invoke client logic
     _currentStateDelta.setStartTime(_message.getPartitionName(), System.currentTimeMillis());
 
-    Exception err = null;
-    if (toState.equalsIgnoreCase(state)) {
-      // To state equals current state, we can just ignore the message
-      err = new HelixDuplicatedStateTransitionException(
-          String.format("Partition %s current state is same as toState (%s->%s) from message.",
-              partitionName, fromState, toState));
-    } else if (fromState != null && !fromState.equals("*") && !fromState.equalsIgnoreCase(state)) {
-      // If current state is neither toState nor fromState in message, there is a problem
-      err = new HelixStateMismatchException(String.format(
-          "Current state of stateModel does not match the fromState in Message, CurrentState: %s, Message: %s->%s, Partition: %s, from: %s, to: %s",
-          state, fromState, toState, partitionName, _message.getMsgSrc(), _message.getTgtName()));
-    }
+    Exception err = isMessageStaled();
 
     if (err != null) {
       _statusUpdateUtil.logError(_message, HelixStateTransitionHandler.class, err.getMessage(),
@@ -461,6 +450,31 @@ public class HelixStateTransitionHandler extends MessageHandler {
       _stateModel.rollbackOnError(_message, _notificationContext, error);
     }
 
+  }
+
+  public Exception isMessageStaled() {
+    String state = _stateModel.getCurrentState() != null ? _stateModel.getCurrentState()
+        : _currentStateDelta.getState(_message.getPartitionName());
+    String fromState = _message.getFromState();
+    String toState = _message.getToState();
+    String partitionName = _message.getPartitionName();
+
+    //String err = null;
+
+    Exception err = null;
+    if (toState.equalsIgnoreCase(state)) {
+      // To state equals current state, we can just ignore the message
+      err = new HelixDuplicatedStateTransitionException(
+          String.format("Partition %s current state is same as toState (%s->%s) from message.",
+              partitionName, fromState, toState));
+    } else if (fromState != null && !fromState.equals("*") && !fromState.equalsIgnoreCase(state)) {
+      // If current state is neither toState nor fromState in message, there is a problem
+      err = new HelixStateMismatchException(String.format(
+          "Current state of stateModel does not match the fromState in Message, CurrentState: %s, Message: %s->%s, Partition: %s, from: %s, to: %s",
+          state, fromState, toState, partitionName, _message.getMsgSrc(), _message.getTgtName()));
+    }
+
+    return err;
   }
 
   @Override
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -38,10 +38,6 @@ import org.apache.helix.NotificationContext;
 import org.apache.helix.NotificationContext.MapKey;
 import org.apache.helix.PropertyKey;
 import org.apache.helix.PropertyKey.Builder;
-import org.apache.helix.zookeeper.datamodel.ZNRecord;
-import org.apache.helix.zookeeper.datamodel.ZNRecordBucketizer;
-import org.apache.helix.zookeeper.datamodel.ZNRecordDelta;
-import org.apache.helix.zookeeper.datamodel.ZNRecordDelta.MergeOperation;
 import org.apache.helix.model.CurrentState;
 import org.apache.helix.model.Message;
 import org.apache.helix.model.Message.Attributes;
@@ -50,6 +46,10 @@ import org.apache.helix.participant.statemachine.StateModelFactory;
 import org.apache.helix.participant.statemachine.StateModelParser;
 import org.apache.helix.participant.statemachine.StateTransitionError;
 import org.apache.helix.util.StatusUpdateUtil;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.apache.helix.zookeeper.datamodel.ZNRecordBucketizer;
+import org.apache.helix.zookeeper.datamodel.ZNRecordDelta;
+import org.apache.helix.zookeeper.datamodel.ZNRecordDelta.MergeOperation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -463,6 +442,35 @@ public void onError(Exception e, ErrorCode code, ErrorType type) {
 
   }
 
+  // Verify the fromState and current state of the stateModel.
+  public Exception isMessageStaled(boolean inSchedulerCheck) {
```
zhangmeng916, line 446: minor: change isMessageStaled to isMessageStale. We didn't on purpose stale the message.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -463,6 +442,35 @@ public void onError(Exception e, ErrorCode code, ErrorType type) {
...
+    // Defined in HelixStateMachineEngine.
+    String state = _currentStateDelta.getState(partitionName);
+
+    Exception err = null;
+    if (toState.equalsIgnoreCase(state)) {
+      // To state equals current state, we can just ignore the message
+      err = new HelixDuplicatedStateTransitionException(String
+          .format("Partition %s current state is same as toState (%s->%s) from message.",
+              partitionName, fromState, toState));
+    } else if (!inSchedulerCheck && fromState != null && !fromState.equals("*") && !fromState
```
zhangmeng916, line 463: Besides the current_state equals to_state case, is this also a case that we can move to scheduling phase instead of execution phase? That could help reduce even more threads usage.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -919,9 +912,25 @@ public void onMessage(String instanceName, List<Message> messages,
...
-                message.getResourceName(), message.getPartitionName(), msg.getMsgId(), String.valueOf(msg.isRelayMessage()),
-                msg.getReadTimeStamp(), System.currentTimeMillis(), message.getFromState(),
-                message.getToState()));
+                message.getResourceName(), message.getPartitionName(), msg.getMsgId(),
+                String.valueOf(msg.isRelayMessage()), msg.getReadTimeStamp(),
+                System.currentTimeMillis(), message.getFromState(), message.getToState()));
+          }
+          if (createHandler instanceof HelixStateTransitionHandler) {
+            // We only check to state if there is no ST task scheduled/executing.
+            Exception err = ((HelixStateTransitionHandler) createHandler).isMessageStaled(true /*inSchedulerCheck*/);
```
zhangmeng916, line 921: The return type is confusing. `isMessageStaled` implies a boolean value as return. You can change the function name to something like validateStaleMessage, etc, or change the return type.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -980,9 +980,10 @@ public TaskState pollForJobState(String workflowName, String jobName, long timeo
         && System.currentTimeMillis() < st + timeout);
 
     if (ctx == null || !allowedStates.contains(ctx.getJobState(jobName))) {
-      throw new HelixException(
-          String.format("Workflow \"%s\" context is null or job \"%s\" is not in states: %s",
-              workflowName, jobName, allowedStates));
+      String cur = ctx == null ? "null" : ctx.getJobState(jobName).toString();
+      throw new HelixException(String.format(
+          "Workflow \"%s\" context is null or job \"%s\" is not in states: %s, cur state is: %s",
+          workflowName, jobName, allowedStates, cur));
```
zhangmeng916, line 986: Is this just a log improvement? The logic is not impacted by this PR's change, right?

helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
```diff
@@ -377,6 +414,55 @@ public void testDuplicatedMessage() throws InterruptedException {
...
+    changeContext.setChangeType(HelixConstants.ChangeType.MESSAGE);
+    executor.onMessage(instanceName, msgList, changeContext);
+
+    Thread.sleep(200);
+
+    // The message should be ignored since toState is the same as current state.
+    Assert.assertEquals(dataAccessor.getChildValues(keyBuilder.messages(instanceName), true).size(),
+        0);
+
+    System.out.println("END TestHelixTaskExecutor.testDuplicatedMessage()");
```
zhangmeng916, line 463: The test name is wrong here.

helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
```diff
@@ -377,6 +414,55 @@ public void testDuplicatedMessage() throws InterruptedException {
...
+    Assert.assertEquals(dataAccessor.getChildValues(keyBuilder.messages(instanceName), true).size(),
+            nMsgs);
+
+    changeContext.setChangeType(HelixConstants.ChangeType.MESSAGE);
+    executor.onMessage(instanceName, msgList, changeContext);
+
+    Thread.sleep(200);
+
+    // The message should be ignored since toState is the same as current state.
+    Assert.assertEquals(dataAccessor.getChildValues(keyBuilder.messages(instanceName), true).size(),
```
zhangmeng916, line 460: How soon is the message gets deleted from instance?

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
xyuanlu: Yes it is a log improvement. I personally find the new log more helpful when debugging. Since we are comparing two states, printing out the expected state and actually state could be more useful.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: TFTR. This case is a bit different than the above one. The handling for these two cases are different. 

For the first case when toState == currentState, we just treat the task as succeeded and finish task. There is no difference between doing the no-op when executing and bypass the task at scheduling phase.  

In the second case when currentState != fromState, executor will set the task as failed and report state transition error. Ir will also update ZK.

helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
xyuanlu: In scheduling phase, HelixUtil.removeMessageFromZK will be called as soon as we hit the HelixDuplicatedStateTransitionException exception. There is no thread queueing in between. So I think it should be the IO delay time.

zhangmeng916: ok, that's fine. Asked cause I saw the 200ms sleep, just to confirm it immediately happens.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -919,9 +912,25 @@ public void onMessage(String instanceName, List<Message> messages,
...
-                message.getResourceName(), message.getPartitionName(), msg.getMsgId(), String.valueOf(msg.isRelayMessage()),
-                msg.getReadTimeStamp(), System.currentTimeMillis(), message.getFromState(),
-                message.getToState()));
+                message.getResourceName(), message.getPartitionName(), msg.getMsgId(),
+                String.valueOf(msg.isRelayMessage()), msg.getReadTimeStamp(),
+                System.currentTimeMillis(), message.getFromState(), message.getToState()));
+          }
+          if (createHandler instanceof HelixStateTransitionHandler) {
+            // We only check to state if there is no ST task scheduled/executing.
+            Exception err = ((HelixStateTransitionHandler) createHandler).validateStaleMessage(true /*inSchedulerCheck*/);
```
jiajunwang, line 921: What will be wrong if we do validateStaleMessage(false) here? If it works, then we can have a common logic, right?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -919,9 +912,25 @@ public void onMessage(String instanceName, List<Message> messages,
...
+            if (err != null) {
+              throw err;
```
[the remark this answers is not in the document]
xyuanlu: We do continue when catch the exception (line 957). We could change the message if I think if the wording is not accurate in the catch block.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -919,9 +912,25 @@ public void onMessage(String instanceName, List<Message> messages,
...
+            }
+          }
+          if (stateTransitionHandlers.containsKey(messageTarget)) {
```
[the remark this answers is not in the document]
xyuanlu: Please correct me if I am wrong. I think if we have an old O->S and a new S->M with current state is S. We do want to ignore the O->S and continue with S->M.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -463,6 +442,35 @@ public void onError(Exception e, ErrorCode code, ErrorType type) {
 
   }
 
+  // Verify the fromState and current state of the stateModel.
+  public Exception validateStaleMessage(boolean inSchedulerCheck) {
```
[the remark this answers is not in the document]
xyuanlu: TFTR. Will update.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
xyuanlu: The handling for these two cases are different.

For the first case when toState == currentState, we just treat the task as succeeded and finish task. There is no difference between doing the no-op when executing and bypass the task at scheduling phase.

In the second case when currentState != fromState, executor will set the task as failed and report state transition error. Ir will also update ZK.

Having same logic causes a test fail. Controller keeps sending many 'INIT->RUNNING' message when currentState is TASK_ERROR if we don't report task error. Currently task error is reported in HelixStateTransitionHandler.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -463,6 +442,36 @@ public void onError(Exception e, ErrorCode code, ErrorType type) {
...
+    // partition is in initial state and we using it as current state.
+    // Defined in HelixStateMachineEngine.
+    String state = _currentStateDelta.getState(partitionName);
+
+    if (toState.equalsIgnoreCase(state)) {
+      // To state equals current state, we can just ignore the message
+      throw new HelixDuplicatedStateTransitionException(String
+          .format("Partition %s current state is same as toState (%s->%s) from message.",
+              partitionName, fromState, toState));
+    } else if (!isPreCheck && fromState != null && !fromState.equals("*") && !fromState
```
dasahcc, line 462: Why we need this precheck flag? Two checks in onMessage and HelixTask execution validation logic should be same

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -463,6 +442,36 @@ public void onError(Exception e, ErrorCode code, ErrorType type) {
...
+
+    // state in _currentStateDelta uses current state from state model. It has the
+    // most up-to-date. current state. In case currentState in stateModel is null,
+    // partition is in initial state and we using it as current state.
+    // Defined in HelixStateMachineEngine.
+    String state = _currentStateDelta.getState(partitionName);
+
+    if (toState.equalsIgnoreCase(state)) {
+      // To state equals current state, we can just ignore the message
+      throw new HelixDuplicatedStateTransitionException(String
```
dasahcc, line 459: This is not duplicate state transition. Could be some STs are old in pending queue just not match what target state we have now.

xyuanlu: If it already been processed (in READ or UNPROCESSABLE state) then we would ignore the ST message before this check. (line 856 in HelixTaskExecutor)

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: TFTR. 
Originally in HelixStateTransitionHandler, handling for these two cases are different.
For the first case when toState == currentState, we just treat the task as succeeded and finish task. There is no difference between doing the no-op when executing and bypass the task at scheduling phase.

In the second case when currentState != fromState, executor will set the task as failed and report state transition error. Ir will also update ZK.

Having same logic causes a test fail. Controller keeps sending many 'INIT->RUNNING' message when currentState is TASK_ERROR if we don't report task error. Currently task error is reported in HelixStateTransitionHandler.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
dasahcc: What I am trying to say is not use HelixDuplicatedStateTransitionException, because there could be other cases for this scenario. If we check log, we may get confused.

xyuanlu: Thanks for the follow up. 
In my understanding, this message must be a new message generated by controller, maybe caused by reading the old currentState etc.Since this is a new message, doesn't this count as HelixDuplicatedStateTransition?

helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
```diff
@@ -232,11 +248,29 @@ public HelixTaskResult handleMessage() throws InterruptedException {
...
+        if (_testIsMessageStaled) {
+          super.precheckForStaleMessage();
+        }
+      }
     }
 
     @Override
     public MessageHandler createHandler(Message message, NotificationContext context) {
-      return new TestStateTransitionMessageHandler(message, context);
+      if (message.getResourceName()!="testStaledMessageResource") {
```
pkuwm, line 262: `equals()` should be used to check the string names are the same.
`!=` checks if they are the **same** object: same memory address, etc..

helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
```diff
@@ -212,18 +213,33 @@ public TestStateTransitionHandlerFactory(String msgType, long delay) {
...
+      public HelixTaskResult handleMessage() {
         HelixTaskResult result = new HelixTaskResult();
         _processedMsgIds.put(_message.getMsgId(), _message.getMsgId());
         if (_delay > 0) {
           System.out.println("Sleeping..." + _delay);
-          Thread.sleep(_delay);
+          try{
+            Thread.sleep(_delay);
+          } catch (Exception e) {
+            assert (false);
```
pkuwm, line 241: Can you help understand why we need to assert instead of throwing an `InterruptedException`?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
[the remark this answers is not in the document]
zhangmeng916: So I think the logic changed here. Previously we always discard the later one (so it's possible we choose the stale one, which is the earlier one). But now we do a filtering first based on current state, and if the message is stale, we will directly delete it, and there won't be duplicated message exception at all. Is this the logic you think?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
dasahcc: For the test you talked about, I think we may need to understand why we keep sending message INIT -> RUNNING. Because once the message has been DROPPED, controller should refresh current state and get the TASK_ERROR state. Then we should have some message like TASK_ERROR -> INIT or something, if we miss it, we should fix the logic to make it more robust. And I am OK to not fix it in this PR. But if that's the case, let's propose another PR for that. Otherwise, this is not a complete scenario. For regular resource management, it is not a complete filtering logic.

helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
xyuanlu: I think the reason is that the base handleMessage does not throw exception. The try-catch block here is only for handle exceptions in test.

helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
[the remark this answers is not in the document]
xyuanlu: Yes. We do a filtering first based on current state, and if the message is stale, we will directly delete it. (and we will throw exception as well, but not at the execution stage as before)
Although there is no existing bug complaining 2 valid new messages at the same time, I think we should not assume there will be no duplicated messages so we should still keep the 3rd check.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: I had an offline sync with Ali. He has submit a PR to fix this issue (#1390). I changed my PR accordingly. 
Currently my PR contains his patch.  Will rebase after his PR merged.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: I had an offline sync with Ali. He has submit a PR to fix this issue (#1390). I changed my PR accordingly. We now do the same validation for scheduling stage and execution stage.  
Currently my PR contains his patch. Will rebase after his PR merged.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -919,9 +912,25 @@ public void onMessage(String instanceName, List<Message> messages,
...
+            // If there are 2 messages in same batch about same partition's state transition,
+            // the later one is discarded
+            Message duplicatedMessage = stateTransitionHandlers.get(messageTarget)._message;
+            throw new HelixException(String.format(
```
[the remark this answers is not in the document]
xyuanlu: Update.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
xyuanlu: I had an offline sync with Ali. He has submit a PR to fix this issue (#1390). I changed my PR accordingly. We now do the same validation for scheduling stage and execution stage.  
Currently my PR contains his patch. Will rebase after his PR merged.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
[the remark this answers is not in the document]
jiajunwang: try-catch is relatively expensive. And it might be misused if the caught exception is too wildly defined. So my suggestion is to handle the error case with regular logic as long as it is possible.

helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
pkuwm: I don't think we need this change, as throwing an `InterruptedException` is good enough and cleaner to fail the test, unless the upstream caller catches and swallows the `InterruptedException`

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -460,7 +438,33 @@ public void onError(Exception e, ErrorCode code, ErrorType type) {
       StateTransitionError error = new StateTransitionError(type, code, e);
       _stateModel.rollbackOnError(_message, _notificationContext, error);
     }
+  }
+
+  // Verify the fromState and current state of the stateModel.
+  public Exception staleMessageValidator() {
```
pkuwm, line 444: `staleMessageValidator` sounds more like a class/variable name. How about changing the method name to a verb style: `validateStaleMessage()` (it does something)?

helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
xyuanlu: Please correct me if I am wrong. 
For the `InterruptedException`, we need to either catch it or add a declaration in method signature. Since the HelixStateTransitionHandler.handleMessage() does not throw exception, we could only do a try-catch here.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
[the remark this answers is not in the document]
xyuanlu: Updated.

dasahcc (approved):
Overall, LGTM. Good job!

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: In my understanding, validateXXX is a function throws exception and returns void or boolean. Since I have a function that returns an exception, I am not sure about the name validateXXX.

xyuanlu:

Final commit message:
Move ST message to state validation from executing phase to scheduling phase.

Move ST message to state validation to scheduling phase.because of the following reasons:
It is a no-op operation, thus we should not create a thread to execute it.
The lag between scheduling and executing could add additional latency for controller and eventually application latency.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
jiajunwang: nit, it would be more graceful to define a ValidationResult class and contain a boolean (pass or not) and an optional Exception. Especially as a public method. Is there a chance we can make it protected or package private?

xyuanlu: We are not using the boolean return value. May I ask is this for a cleaner interface purpose for further usage?

xyuanlu: Updated.

xyuanlu:
This PR is ready to be merged. Approved by @dasahcc 

Final commit message:
Move ST message to state validation from executing phase to scheduling phase.

Move ST message to state validation to scheduling phase.because of the following reasons:
It is a no-op operation, thus we should not create a thread to execute it.
The lag between scheduling and executing could add additional latency for controller and eventually application latency.

PR #1357, opened 2020-09-10
Add HelixRestObjectNameFactory to create namespaced ObjectName

pkuwm:
### Issues

- [x] My PR addresses the following Helix issues and references them in the PR description:

Resolves #1356 

### Description

- [x] Here are some details about my PR, including screenshots of any UI changes:

We need helix rest service metrics for each namespace. To achieve this, the namespace info needs to be contained in the MBean's object name, eg: `domain:name=getClusters,namespace=myNamespace,type=counters`. But DropWizard's default ObjectNameFactory doesn't have the namespace parameter.

This PR implements a custom object name factory and add the namespace to object name.

### Tests

- [x] The following tests are written for this issue:

TestHelixRestObjectNameFactory

- [x] The following is the result of the "mvn test" command on the appropriate module:

```
[INFO] Tests run: 167, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 43.615 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 167, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  49.751 s
[INFO] Finished at: 2020-09-11T00:08:46-07:00
[INFO] ------------------------------------------------------------------------
```

Final run:
```
[INFO] Tests run: 168, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 39.041 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 168, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  45.121 s
[INFO] Finished at: 2020-09-21T18:16:05-07:00
[INFO] ------------------------------------------------------------------------
```
### Documentation (Optional)

- In case of new functionality, my PR adds documentation in the following wiki page:

(Link the GitHub wiki you added)

### Commits

- My commits all reference appropriate Apache Helix GitHub issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
[...]

PR #1377, opened 2020-09-18
Add JmxReporter for reporting rest jmx metrics

### LINKED ISSUE #1376: JmxReporter for helix rest metrics
### Description
Existing helix rest monitoring is using default jersey mbean metrics. But there are too many mbeans. It is consuming too much memory. And we also like to monitor helix rest service availability so we need metrics like response status code metrics. To achieve this, we need a lightweight jmx reporter that can fit our requirement.

pkuwm:
### Issues

Resolves #1376 

### Description

We need a jmx reporter that is lightweight and easy to use for helix rest. The PR implements a `JmxReporter` class that is based on Dropwizard Metrics lib's [`JmxReporter`](https://github.com/dropwizard/metrics/blob/master/metrics-core/src/main/java/io/dropwizard/metrics/JmxReporter.java) class.

Having this class in helix gives us flexibility to reduce mbean attributes if we want to, by removing the metric methods in `JmxReporter`.

Some changes are applied:
* Add `withKeyProperties()` to builder so we can add more key properties to an object name: https://github.com/apache/helix/blob/3299fb4cdf77f62b546e4c33bffd86b5a646cf51/helix-rest/src/main/java/org/apache/helix/rest/metrics/JmxReporter.java#L126-L129
* Remove the `ObjectNameFactory` and put the simple logic to the internal `createName()` and its callers to pass in the parameters.
https://github.com/apache/helix/blob/3299fb4cdf77f62b546e4c33bffd86b5a646cf51/helix-rest/src/main/java/org/apache/helix/rest/metrics/JmxReporter.java#L713-L714
* Fix and add unit tests

### Tests

* TestJmxReporter

```
[INFO] Tests run: 175, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 39.013 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 175, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  45.352 s
[INFO] Finished at: 2020-09-17T21:16:58-07:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-rest/pom.xml
+++ b/helix-rest/pom.xml
@@ -149,6 +149,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-jersey2</artifactId>
+      <version>${metrics-jersey2.version}</version>
+    </dependency>
+
+    <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-all</artifactId>
       <scope>test</scope>
--- a/helix-rest/src/main/java/org/apache/helix/rest/metrics/JmxReporter.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metrics/JmxReporter.java
@@ -57,6 +57,8 @@ import org.slf4j.LoggerFactory;
  * <a href="https://github.com/dropwizard/metrics/blob/master/metrics-core/src/main/java/io/
  * dropwizard/metrics/JmxReporter.java">JmxReporter</a>
  * </pre>
+ * <p>
+ * It is recommended to be only used within Helix Rest.
  */
 public class JmxReporter implements Reporter, Closeable {
   /**
@@ -710,6 +712,12 @@ public class JmxReporter implements Reporter, Closeable {
       }
     }
 
+    /*
+     * Create an object name that must have "type" and "name" keys.
+     * And flexibly, extra keys can be added within keyProperties.
+     * Eg. object name: "org.apache.helix.rest:name=requests-count,type=counters,
+     * namespace=myNamespace"
+     */
     private ObjectName createName(String type, String name,
         Hashtable<String, String> keyProperties) {
       keyProperties.put(REST_JMX_TYPE_KEY, type);
--- a/helix-rest/src/main/java/org/apache/helix/rest/metrics/JmxReporter.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/metrics/JmxReporter.java
@@ -1,0 +1,815 @@
+package org.apache.helix.rest.metrics;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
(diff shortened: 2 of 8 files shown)


helix-rest/pom.xml
```diff
@@ -148,6 +148,12 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-jersey2</artifactId>
```
jiajunwang, line 153: Where is the new ivy def for metrics-jersey2? Do we need that too?

helix-rest/src/main/java/org/apache/helix/rest/metrics/JmxReporter.java
```diff
@@ -0,0 +1,815 @@
...
+/**
+ * A reporter which listens for new metrics and exposes them as namespaced MBeans.
+ * <p>
+ * The code is largely based on the {@code JmxReporter} class of the dropwizard metrics library
+ * <pre>See
+ * <a href="https://github.com/dropwizard/metrics/blob/master/metrics-core/src/main/java/io/
+ * dropwizard/metrics/JmxReporter.java">JmxReporter</a>
+ * </pre>
+ */
+public class JmxReporter implements Reporter, Closeable {
```
jiajunwang, line 61: How many codes do we need to change?
1. is it possible we extend the original JmxReporter?
2. is it possible you highlight the changed logic?
This file is not quite reviewable.

helix-rest/pom.xml
pkuwm: I thought about it. I don't see `jersey-server` is added to ivy file. Instead, we add the dependency to our MP wrapper. So I thought we could also add `metrics-jersey2` to the MP directly. Otherwise, we would need to add all the dependencies in pom to ivy.

helix-rest/src/main/java/org/apache/helix/rest/metrics/JmxReporter.java
pkuwm: 1. It is not possible that we extend the original JmxReporter, because it has internal classes that are private.
2. I put the the change description in the PR description: 
Some changes are applied:
* Add withKeyProperties() to builder so we can add more key properties to an object name
* Remove the ObjectNameFactory and put the simple logic to the internal createName()
* Fix and add unit tests

But let me also add some comments to highlight our change.

helix-rest/pom.xml
jiajunwang: As long as you are ensured that anyone without accessing "MP" and using ivy can see the same feature, then it should be fine.
But from what you described, I feel it is a potential problem.

helix-rest/src/main/java/org/apache/helix/rest/metrics/JmxReporter.java
jiajunwang: Based on what you described, I have a simpler alternative design that might work.

We implement our own ObjectNameFactory instead of JmxReporter. Let's call it NamespaceObjectNameFactory. It will be initialized with the namespace. Then you can put the namespace to the ObjectName whenever the reporter calls CreateName method.

helix-rest/pom.xml
pkuwm: As long as users are not using ivy, it is fine because we've defined all the required dependencies in pom. If they are using ivy without adding dependencies, it's been a problem because dependencies like `jersey-server` are not even in rest ivy.

helix-rest/src/main/java/org/apache/helix/rest/metrics/JmxReporter.java
pkuwm: I understand if we only want to customize the object name, we don’t have to add Jmx reporter. I was thinking we firstly implement our object name factory. Of course, like you said, we can put the key properties in the constructor: `ObjectNameFactory(HashTable<String, String>())`. And later if we want to remove those attributes, we add jmx reporter.
It really depends on whether if we need to remove some attributes. We can do it now or later. I don’t have a strong preference, but if we need to remove the attributes, we need to have the JmxReporter.

pkuwm:
We'll firstly use a customized object name factory to insert namespace key to object name: #1357
We'll put this in backlog. If we need to remove attributes and also need this JmxReporter for further integration work, we then add this class.

PR #1383, opened 2020-09-20
Add Metrics for External Custom Health Check Requests

### LINKED ISSUE #1382: Add metrics for custom rest client request
### Description
Custom rest client interacts with participant side to query for its health checks. Currently there is no metrics for the query requests. We'd like to have metrics for the requests.

### Expected
Metrics needed:
Total and failed requests;
Request latency;
Response content to determine health?

pkuwm:
### Issues

Resolves #1382 

### Description

Custom rest client interacts with participant side to query for its health checks. Currently there is no metrics for the query requests. We'd like to have metrics for the requests.

Here's a design trade-off:

1) Add a new API `getInstanceStoppableCheck(baseUrl, customPayloads, namespace)` in interface `CustomRestClient`. We need the namespace name when recording metrics in `CustomRestClient`. Since CustomRestClient is a singleton instance, we can not inject the namespace to the class. Having the namespace as a parameter in the API is an option. The benefit is we can accurately record the http request latency, and it is more scalable if we have more places calling the API, we don't need to do anything. Cons: change existing interface to add the new API. Though it is an interface, we may not expect external users to use it. So it should be fine to change the interface.
2) Record the metrics in the upstream caller outside `CustomRestClient`, eg.  `InstanceService` where `getInstanceStoppableCheck()` is called, we record the metrics. Pros: we don't need to change existing interface `CustomRestClient`. Cons: Latency of http request is more than the actual number because other stuff like json response parsing is also involved in `CustomRestClient`; and it is not scalable because we have to record metrics in each place where `CustomRestClient. getInstanceStoppableCheck()` is called.

This PR implements option 2), because there are not many places that call `CustomRestClient` APIs, and it doesn't look good that pass the namespace as a parameter just for metrics rather than logic related.

image

### Tests

```
[INFO] Tests run: 168, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 59.766 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 168, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:05 min
[INFO] Finished at: 2020-09-28T17:43:10-07:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
@@ -31,12 +31,18 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
+
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.SharedMetricRegistries;
+import com.codahale.metrics.Timer;
 import org.apache.helix.HelixProperty;
 import org.apache.helix.PropertyKey;
 import org.apache.helix.manager.zk.ZKHelixDataAccessor;
 import org.apache.helix.model.RESTConfig;
 import org.apache.helix.rest.client.CustomRestClient;
 import org.apache.helix.rest.client.CustomRestClientFactory;
+import org.apache.helix.rest.server.service.InstanceService;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -54,12 +60,27 @@ public class HelixDataAccessorWrapper extends ZKHelixDataAccessor {
   public static final String IS_HEALTHY_KEY = "IS_HEALTHY";
   public static final String EXPIRY_KEY = "EXPIRE";
 
+  // Metric names for custom partition check
+  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_TOTAL =
+      MetricRegistry.name(InstanceService.class, "custom_partition_check_http_requests_total");
+  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_ERROR_TOTAL = MetricRegistry
+      .name(InstanceService.class, "custom_partition_check_http_requests_error_total");
+  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUEST_DURATION =
+      MetricRegistry.name(InstanceService.class, "custom_partition_check_http_request_duration");
+
   private final Map<PropertyKey, HelixProperty> _propertyCache = new HashMap<>();
   private final Map<PropertyKey, List<String>> _batchNameCache = new HashMap<>();
+  private String _namespace;
   protected CustomRestClient _restClient;
 
   public HelixDataAccessorWrapper(ZKHelixDataAccessor dataAccessor) {
     super(dataAccessor);
+    _restClient = CustomRestClientFactory.get();
+  }
+
+  public HelixDataAccessorWrapper(ZKHelixDataAccessor dataAccessor, String namespace) {
+    super(dataAccessor);
+    _namespace = namespace;
     _restClient = CustomRestClientFactory.get();
   }
 
@@ -166,12 +187,18 @@ public class HelixDataAccessorWrapper extends ZKHelixDataAccessor {
 
   private Map<String, Boolean> getHealthStatusFromRest(String instance, List<String> partitions,
       RESTConfig restConfig, Map<String, String> customPayLoads) {
-    try {
+    MetricRegistry metrics = SharedMetricRegistries.getOrCreate(_namespace);
+    Counter requestsTotal = metrics.counter(CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_TOTAL);
+    Counter errorRequestsTotal = metrics.counter(CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_ERROR_TOTAL);
+    try (final Timer.Context timer = metrics.timer(CUSTOM_PARTITION_CHECK_HTTP_REQUEST_DURATION)
+        .time()) {
+      requestsTotal.inc();
       return _restClient.getPartitionStoppableCheck(restConfig.getBaseUrl(instance), partitions,
           customPayLoads);
     } catch (IOException e) {
       LOG.error("Failed to get partition status on instance {}, partitions: {}", instance,
           partitions, e);
+      errorRequestsTotal.inc();
       return Collections.emptyMap();
     }
   }
--- a/helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
@@ -60,12 +60,10 @@ public class HelixDataAccessorWrapper extends ZKHelixDataAccessor {
   public static final String EXPIRY_KEY = "EXPIRE";
 
   // Metric names for custom partition check
-  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_TOTAL =
-      MetricRegistry.name(InstanceService.class, "custom_partition_check_http_requests_total");
   private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_ERROR_TOTAL = MetricRegistry
       .name(InstanceService.class, "custom_partition_check_http_requests_error_total");
-  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUEST_DURATION =
-      MetricRegistry.name(InstanceService.class, "custom_partition_check_http_request_duration");
+  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_DURATION =
+      MetricRegistry.name(InstanceService.class, "custom_partition_check_http_requests_duration");
 
   private final Map<PropertyKey, HelixProperty> _propertyCache = new HashMap<>();
   private final Map<PropertyKey, List<String>> _batchNameCache = new HashMap<>();
@@ -82,16 +80,13 @@ public class HelixDataAccessorWrapper extends ZKHelixDataAccessor {
     this(dataAccessor, CustomRestClientFactory.get(), HelixRestNamespace.DEFAULT_NAMESPACE_NAME);
   }
 
-  public HelixDataAccessorWrapper(ZKHelixDataAccessor dataAccessor, String namespace) {
-    this(dataAccessor, CustomRestClientFactory.get(), namespace);
-  }
-
+  @Deprecated
   public HelixDataAccessorWrapper(ZKHelixDataAccessor dataAccessor,
       CustomRestClient customRestClient) {
     this(dataAccessor, customRestClient, HelixRestNamespace.DEFAULT_NAMESPACE_NAME);
   }
 
-  private HelixDataAccessorWrapper(ZKHelixDataAccessor dataAccessor,
+  public HelixDataAccessorWrapper(ZKHelixDataAccessor dataAccessor,
       CustomRestClient customRestClient, String namespace) {
     super(dataAccessor);
     _restClient = customRestClient;
@@ -197,9 +192,9 @@ public class HelixDataAccessorWrapper extends ZKHelixDataAccessor {
   private Map<String, Boolean> getHealthStatusFromRest(String instance, List<String> partitions,
       RESTConfig restConfig, Map<String, String> customPayLoads) {
     MetricRegistry metrics = SharedMetricRegistries.getOrCreate(_namespace);
-    try (final Timer.Context timer = metrics.timer(CUSTOM_PARTITION_CHECK_HTTP_REQUEST_DURATION)
+    // Total requests metric is included as an attribute(Count) in timers
+    try (final Timer.Context timer = metrics.timer(CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_DURATION)
         .time()) {
-      metrics.counter(CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_TOTAL).inc();
       return _restClient.getPartitionStoppableCheck(restConfig.getBaseUrl(instance), partitions,
           customPayLoads);
     } catch (IOException e) {
--- a/helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
@@ -32,7 +32,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 
-import com.codahale.metrics.Counter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.SharedMetricRegistries;
 import com.codahale.metrics.Timer;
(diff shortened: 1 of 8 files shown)


helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
```diff
@@ -166,12 +187,18 @@ public HelixDataAccessorWrapper(ZKHelixDataAccessor dataAccessor, CustomRestClie
 
   private Map<String, Boolean> getHealthStatusFromRest(String instance, List<String> partitions,
       RESTConfig restConfig, Map<String, String> customPayLoads) {
-    try {
+    MetricRegistry metrics = SharedMetricRegistries.getOrCreate(_namespace);
+    Counter requestsTotal = metrics.counter(CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_TOTAL);
+    Counter errorRequestsTotal = metrics.counter(CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_ERROR_TOTAL);
+    try (final Timer.Context timer = metrics.timer(CUSTOM_PARTITION_CHECK_HTTP_REQUEST_DURATION)
```
zhangmeng916, line 193: In case of an exception, the time will be also counted in the request duration, right? Is this what in the design? I think this will make the duration statistic less accurate.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
```diff
@@ -231,14 +252,22 @@ private StoppableCheck performHelixOwnInstanceCheck(String clusterId, String ins
   private StoppableCheck performCustomInstanceCheck(String clusterId, String instanceName,
       String baseUrl, Map<String, String> customPayLoads) {
     LOG.info("Perform instance level client side health checks for {}/{}", clusterId, instanceName);
-    try {
-      return new StoppableCheck(
-          _customRestClient.getInstanceStoppableCheck(baseUrl, customPayLoads),
+    MetricRegistry metrics = SharedMetricRegistries.getOrCreate(_namespace);
+    Counter requestsTotal = metrics.counter(CUSTOM_INSTANCE_CHECK_HTTP_REQUESTS_TOTAL);
+    Counter errorRequestsTotal = metrics.counter(CUSTOM_INSTANCE_CHECK_HTTP_REQUESTS_ERROR_TOTAL);
```
zhangmeng916, line 257: nit: variable name to be consistent with metric name to avoid misuse.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
```diff
@@ -58,9 +63,18 @@
   private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
   private static final ExecutorService POOL = Executors.newCachedThreadPool();
 
+  // Metric names for custom instance check
+  private static final String CUSTOM_INSTANCE_CHECK_HTTP_REQUESTS_TOTAL =
```
zhangmeng916, line 67: More specifically, this is "health" check. Maybe you can make it more accurate in case you have other types of instance check metrics later and need to distinguish between them.
Same for partition health check.

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
```diff
@@ -54,15 +60,30 @@
...
+      .name(InstanceService.class, "custom_partition_check_http_requests_error_total");
+  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUEST_DURATION =
+      MetricRegistry.name(InstanceService.class, "custom_partition_check_http_request_duration");
+
   private final Map<PropertyKey, HelixProperty> _propertyCache = new HashMap<>();
   private final Map<PropertyKey, List<String>> _batchNameCache = new HashMap<>();
+  private String _namespace;
   protected CustomRestClient _restClient;
 
   public HelixDataAccessorWrapper(ZKHelixDataAccessor dataAccessor) {
```
zhangmeng916, line 76: Shall we mark this as deprecated?

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
```diff
@@ -54,15 +60,30 @@
...
+  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_TOTAL =
+      MetricRegistry.name(InstanceService.class, "custom_partition_check_http_requests_total");
+  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_ERROR_TOTAL = MetricRegistry
+      .name(InstanceService.class, "custom_partition_check_http_requests_error_total");
+  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUEST_DURATION =
+      MetricRegistry.name(InstanceService.class, "custom_partition_check_http_request_duration");
+
   private final Map<PropertyKey, HelixProperty> _propertyCache = new HashMap<>();
   private final Map<PropertyKey, List<String>> _batchNameCache = new HashMap<>();
+  private String _namespace;
```
jiajunwang, line 73: _namespace is not a must logic for the class, so I suggest adding a set method for it instead of adding a constructor.
If the namespace is not set, then just skip the metrics logic.

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
```diff
@@ -166,12 +187,18 @@ public HelixDataAccessorWrapper(ZKHelixDataAccessor dataAccessor, CustomRestClie
 
   private Map<String, Boolean> getHealthStatusFromRest(String instance, List<String> partitions,
       RESTConfig restConfig, Map<String, String> customPayLoads) {
-    try {
+    MetricRegistry metrics = SharedMetricRegistries.getOrCreate(_namespace);
+    Counter requestsTotal = metrics.counter(CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_TOTAL);
+    Counter errorRequestsTotal = metrics.counter(CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_ERROR_TOTAL);
```
jiajunwang, line 192: Please check my comments for the instance accessor, the same here. Please lazily get and simplify the code whenever possible.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
```diff
@@ -58,9 +63,18 @@
...
+      MetricRegistry.name(InstanceService.class, "custom_instance_check_http_requests_total");
+  private static final String CUSTOM_INSTANCE_CHECK_HTTP_REQUESTS_ERROR_TOTAL =
+      MetricRegistry.name(InstanceService.class, "custom_instance_check_http_requests_error_total");
+  private static final String CUSTOM_INSTANCE_CHECK_HTTP_REQUEST_DURATION =
+      MetricRegistry.name(InstanceService.class, "custom_instance_check_http_request_duration");
+
   private final HelixDataAccessorWrapper _dataAccessor;
```
jiajunwang, line 74: You can get namespace from dataAccessor? Please try to ensure there is only one source of truth.

jiajunwang: Or do we share the same accessor across multiple service implementation? If that's the case, we need to reconsider the design.

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
pkuwm: I was thinking other callers may use it. Within helix, only a mock class calls it. So I just changed it and deprecate this, assuming this constructor is supposed to be used within helix rest.

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
pkuwm: It won't make the duration statistic too bad, compared with network IO. CPU is really fast to process the exception. Actually we still need to use a `try... finally` block - same as this `try.. with..resource`.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
pkuwm: No :) You got it. The first PR was a draft. A constructor with namespace is passed in, I didn't push the change yet when you saw this. Updated.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
```diff
@@ -231,14 +252,22 @@ private StoppableCheck performHelixOwnInstanceCheck(String clusterId, String ins
...
       String baseUrl, Map<String, String> customPayLoads) {
     LOG.info("Perform instance level client side health checks for {}/{}", clusterId, instanceName);
-    try {
-      return new StoppableCheck(
-          _customRestClient.getInstanceStoppableCheck(baseUrl, customPayLoads),
+    MetricRegistry metrics = SharedMetricRegistries.getOrCreate(_namespace);
+    Counter requestsTotal = metrics.counter(CUSTOM_INSTANCE_CHECK_HTTP_REQUESTS_TOTAL);
+    Counter errorRequestsTotal = metrics.counter(CUSTOM_INSTANCE_CHECK_HTTP_REQUESTS_ERROR_TOTAL);
+
+    try (final Timer.Context timer = metrics.timer(CUSTOM_INSTANCE_CHECK_HTTP_REQUEST_DURATION)
```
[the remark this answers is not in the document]
pkuwm: Java 8 `try..with..resource` needs this variable, though we won't use it in the block :). Maybe a newer java supports it.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
pkuwm: This naming is based on the method/function. It is a generic interface api `getInstanceStoppableCheck`, so no matter it is a health check or else, as long as it calls this api, we record it. What do you think?

pkuwm (commented):
Thanks @zhangmeng916 @jiajunwang for the review! I've updated the PR to pass the tests and adjust the code.

helix-rest/src/main/java/org/apache/helix/rest/server/HelixRestServer.java
```diff
@@ -222,7 +222,6 @@ public void shutdown() {
       }
     }
     _jmxReporterList.forEach(JmxReporter::stop);
-    _jmxReporterList.clear();
```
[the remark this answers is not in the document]
pkuwm: Somehow happened when rebasing/merging code from another local branch. Not intended.

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
pkuwm: In helix rest logic, I think we need to pass the namespace to this wrapper so partition check metrics could be namespaced, otherwise, we won't be able to record namespaced metrics for this partition check.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
pkuwm: Resolved.

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
```diff
@@ -54,15 +60,30 @@
   public static final String IS_HEALTHY_KEY = "IS_HEALTHY";
   public static final String EXPIRY_KEY = "EXPIRE";
 
+  // Metric names for custom partition check
+  private static final String CUSTOM_PARTITION_CHECK_HTTP_REQUESTS_TOTAL =
+      MetricRegistry.name(InstanceService.class, "custom_partition_check_http_requests_total");
```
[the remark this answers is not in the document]
jiajunwang: This is a good design actually. But it conflicts with our current assumption that the metric name contains GUAGE or COUNTER keyword. Please suggest how you think we shall let them work together? Or shall we modify our current metrics to include this type field too?

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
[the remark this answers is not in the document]
jiajunwang: I see. It needs a local variable to do the trick. It looks strange but as long as it works, I'm good.

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
jiajunwang: High-levelly, I think we shall try our best to ensure that the metrics logic does not impact business logic. But the current implementation has 2 impacts at least.
1. More constructors, harder to use.
2. We enforce the business logic to have separate accessor wrapper for different namespaces. This does not make much sense if you consider the resource usage, object management, etc.
On the other hand, if we do not let the data accessor wrapper record the namespace, then the metrics can be recorded in the callers. This is a trade-off.

With the current implementation, I actually agree with put it inside the wrapper class. But we shall try to pass the namespace in a more direct way. For example, since we are enforcing the wrapper object serve for one namespace only, we can construct the wrapper object inside the InstanceServiceImpl. In this way, you are passing the namespace only once. And there is no way for the other devs to introduce inconsistent namespace parameters.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
jiajunwang: As commented above, passing the namespace twice in one line of code is not a good design. If you are convinced that the wrapper object should be limited so serve only one service impl, then let's move it inside.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
```diff
@@ -58,29 +62,44 @@
...
   }
 
-  public InstanceServiceImpl(HelixDataAccessorWrapper dataAccessor, ConfigAccessor configAccessor, boolean skipZKRead) {
-    this(dataAccessor,configAccessor);
-    this._skipZKRead = skipZKRead;
+  @Deprecated
+  public InstanceServiceImpl(HelixDataAccessorWrapper dataAccessor, ConfigAccessor configAccessor,
+      boolean skipZKRead) {
+    this(dataAccessor, configAccessor, skipZKRead, HelixRestNamespace.DEFAULT_NAMESPACE_NAME);
+  }
```
jiajunwang, line 88: I might be wrong, but helix-rest is not a module that we expect anyone to extend, so we shall be able to remove unnecessary methods directly.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
pkuwm: Moved. Thx.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
pkuwm: Since helix is open source, I prefer to follow a standard process: deprecate first and then remove it.

helix-rest/src/main/java/org/apache/helix/rest/common/HelixDataAccessorWrapper.java
pkuwm: This is a good suggestion. Previously I was not intended to change the constructor of `InstanceServiceImpl`. As you mentioned, helix rest is expected to be internal, so I think it is fine to change the signature. I have constructed the wrapper object inside the InstanceServiceImpl.

kaisun2000 (approved):
The diff generally looks ok to me. Approve for now.

There are some comments JJ, please sync with him.

alirezazamani (approved):
LGTM

pkuwm:
Thanks, all, for the review!

This PR is ready to be merged, approved by @kaisun2000 @alirezazamani 

```
Custom rest client interacts with participant side to query for its health checks. Currently there is no metrics for the query requests. We'd like to have metrics for the external http requests.

This commit adds http request latency, requests total and error http request total metrics for the custom rest client calls.
```

Issue #1393, opened 2020-09-23
fix TestFullAutoNodeTagging.testSafeAssignment


kaisun2000:
LOG 1235 after delayedAuto as default

>2020-09-23T03:14:06.1737876Z [ERROR] testSafeAssignment(org.apache.helix.integration.rebalancer.TestFullAutoNodeTagging)  Time elapsed: 32.957 s  <<< FAILURE!
2020-09-23T03:14:06.1739684Z java.lang.AssertionError: expected:<true> but was:<false>
2020-09-23T03:14:06.1741742Z 	at org.apache.helix.integration.rebalancer.TestFullAutoNodeTagging.testSafeAssignment(TestFullAutoNodeTagging.java:300)

```
      if (taggedNodes.contains(instanceName)) {
        // make sure that the best possible matches the external view
        Thread.sleep(500);
        boolean result =
            ClusterStateVerifier.verifyByZkCallback(new BestPossAndExtViewZkVerifier(ZK_ADDR,
                clusterName));
        Assert.assertTrue(result);

```

kaisun2000:
START TestFullAutoNodeTagging_testSafeAssignment at Tue Sep 22 22:01:39 PDT 2020
0    [HelixController-pipeline-default-TestFullAutoNodeTagging_testSafeAssignment-(f23303a1_DEFAULT)] ERROR org.apache.helix.controller.rebalancer.DelayedAutoRebalancer  - No instances or active instances available for resource TestDB0, allInstances: [localhost_12922, localhost_12924, localhost_12925, localhost_12920], liveInstances: [], activeInstances: []
24   [HelixController-pipeline-default-TestFullAutoNodeTagging_testSafeAssignment-(fadbd556_DEFAULT)] ERROR org.apache.helix.controller.rebalancer.DelayedAutoRebalancer  - No instances or active instances available for resource TestDB0, allInstances: [localhost_12922, localhost_12924, localhost_12925, localhost_12920], liveInstances: [], activeInstances: []
32   [HelixController-pipeline-default-TestFullAutoNodeTagging_testSafeAssignment-(f21f0ad2_DEFAULT)] ERROR org.apache.helix.controller.rebalancer.DelayedAutoRebalancer  - No instances or active instances available for resource TestDB0, allInstances: [localhost_12922, localhost_12924, localhost_12925, localhost_12920], liveInstances: [], activeInstances: []
[...]

Issue #1374, opened 2020-09-17
fix TestAutoRebalance.testAutoRebalance


kaisun2000:
LOG  976 touch 7
>TestAutoRebalance.testAutoRebalance:177 expected:<true> but was:<false>

>2020-09-17T11:27:23.8495920Z [ERROR] testAutoRebalance(org.apache.helix.integration.rebalancer.TestAutoRebalance)  Time elapsed: 10.815 s  <<< FAILURE!
2020-09-17T11:27:23.8497305Z java.lang.AssertionError: expected:<true> but was:<false>
2020-09-17T11:27:23.8499039Z 	at org.apache.helix.integration.rebalancer.TestAutoRebalance.testAutoRebalance(TestAutoRebalance.java:177)
2020-09-17T11:27:23.8500501Z

kaisun2000:
```
    Thread.sleep(100);
    result = ClusterStateVerifier.verifyByPolling(
        new ExternalViewBalancedVerifier(_gZkClient, CLUSTER_NAME, TEST_DB), 10000, 100);
    Assert.assertTrue(result);

```

Need to get rid of these legacy verifier.

kaisun2000:
LOG 1244 after default to delayedAutoRebalancer

>2020-09-23T03:25:49.9699881Z [ERROR] testAutoRebalance(org.apache.helix.integration.rebalancer.TestAutoRebalance)  Time elapsed: 31.639 s  <<< FAILURE!
2020-09-23T03:25:49.9701635Z java.lang.AssertionError: expected:<true> but was:<false>
2020-09-23T03:25:49.9703882Z 	at org.apache.helix.integration.rebalancer.TestAutoRebalance.testAutoRebalance(TestAutoRebalance.java:179)
2020-09-23T03:25:49.9706164Z

kaisun2000:
```
    Thread.sleep(TestHelper.DEFAULT_REBALANCE_PROCESSING_WAIT_TIME);
    result = ClusterStateVerifier.verifyByPolling(
        new ExternalViewBalancedVerifier(_gZkClient, CLUSTER_NAME, TEST_DB), 30000, 100);
    Assert.assertTrue(result);
```

kaisun2000:
```
    // setup storage cluster
    _gSetupTool.addCluster(CLUSTER_NAME, true);
    _gSetupTool.addResourceToCluster(CLUSTER_NAME, TEST_DB, _PARTITIONS, STATE_MODEL,
[...]

PR #1401, opened 2020-09-23
Mark AutoRebalancer as deprecated and convert the default test logic to not use AutoRebalancer.

### LINKED ISSUE #1392: fix TestControllerLiveLock.test
Since the use of DelayedRebalancer as default from JJ this one is not stable

>2020-09-23T03:28:13.6527533Z [ERROR] test(org.apache.helix.integration.controller.TestControllerLiveLock)  Time elapsed: 42.874 s  <<< FAILURE!
2020-09-23T03:28:13.6530302Z java.lang.AssertionError: expected:<true> but was:<false>
2020-09-23T03:28:13.6537988Z 	at org.apache.helix.integration.controller.TestControllerLiveLock.test(TestControllerLiveLock.java:91)
2020-09-23T03:28:13.6539693Z

jiajunwang:
# Issues

Resolves #1392, #1393, #1374.

### Description

Mark AutoRebalancer as deprecated and convert the default test logic to not use AutoRebalancer.
The only remaining AutoRebalancer tests should be the ones that are specifically testing the AutoRebalancer class.

### Tests

This PR is a pure test change.

[ERROR]   TestDeleteJobFromJobQueue.testForceDeleteJobFromJobQueue:73 » Helix Failed to ...
[INFO] 
[ERROR] Tests run: 1200, Failures: 1, Errors: 0, Skipped: 0

Rerun

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 9.284 s - in org.apache.helix.integration.task.TestDeleteJobFromJobQueue
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/TestHelper.java
+++ b/helix-core/src/test/java/org/apache/helix/TestHelper.java
@@ -40,15 +40,14 @@ import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.helix.PropertyKey.Builder;
+import org.apache.helix.controller.rebalancer.strategy.CrushEdRebalanceStrategy;
+import org.apache.helix.controller.rebalancer.strategy.RebalanceStrategy;
 import org.apache.helix.integration.manager.ZkTestManager;
 import org.apache.helix.manager.zk.CallbackHandler;
 import org.apache.helix.manager.zk.ZKHelixAdmin;
 import org.apache.helix.manager.zk.ZKHelixDataAccessor;
 import org.apache.helix.manager.zk.ZNRecordSerializer;
 import org.apache.helix.manager.zk.ZkBaseDataAccessor;
-import org.apache.helix.zookeeper.api.client.HelixZkClient;
-import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
-import org.apache.helix.zookeeper.impl.factory.SharedZkClientFactory;
 import org.apache.helix.model.CurrentState;
 import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.IdealState.RebalanceMode;
@@ -59,13 +58,16 @@ import org.apache.helix.model.StateModelDefinition.StateModelDefinitionProperty;
 import org.apache.helix.store.zk.ZNode;
 import org.apache.helix.tools.ClusterSetup;
 import org.apache.helix.util.ZKClientPool;
+import org.apache.helix.zookeeper.api.client.HelixZkClient;
+import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.apache.helix.zookeeper.impl.factory.SharedZkClientFactory;
 import org.apache.helix.zookeeper.zkclient.IDefaultNameSpace;
 import org.apache.helix.zookeeper.zkclient.IZkChildListener;
 import org.apache.helix.zookeeper.zkclient.IZkDataListener;
 import org.apache.helix.zookeeper.zkclient.ZkClient;
 import org.apache.helix.zookeeper.zkclient.ZkServer;
 import org.apache.helix.zookeeper.zkclient.exception.ZkNoNodeException;
-import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -287,7 +289,10 @@ public class TestHelper {
 
       for (int i = 0; i < resourceNb; i++) {
         String resourceName = resourceNamePrefix + i;
-        setupTool.addResourceToCluster(clusterName, resourceName, partitionNb, stateModelDef, mode.toString());
+        setupTool.addResourceToCluster(clusterName, resourceName, partitionNb, stateModelDef,
+            mode.name(),
+            mode == RebalanceMode.FULL_AUTO ? CrushEdRebalanceStrategy.class.getName()
+                : RebalanceStrategy.DEFAULT_REBALANCE_STRATEGY);
         if (doRebalance) {
           setupTool.rebalanceStorageCluster(clusterName, resourceName, replica);
         }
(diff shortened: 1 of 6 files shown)


helix-core/src/test/java/org/apache/helix/TestHelper.java
```diff
@@ -287,7 +289,10 @@ public static void setupCluster(String clusterName, String zkAddr, int startPort
 
       for (int i = 0; i < resourceNb; i++) {
         String resourceName = resourceNamePrefix + i;
-        setupTool.addResourceToCluster(clusterName, resourceName, partitionNb, stateModelDef, mode.toString());
+        setupTool.addResourceToCluster(clusterName, resourceName, partitionNb, stateModelDef,
+            mode.name(),
+            mode == RebalanceMode.FULL_AUTO ? CrushEdRebalanceStrategy.class.getName()
+                : RebalanceStrategy.DEFAULT_REBALANCE_STRATEGY);
```
kaisun2000, line 295: Here, RebalanceMode.FULL_AUTO, we will always use crushED, which means delayAutoRebalancer with CrushED

Before this change, it is AutoRebalancer with AutoRebalanceStrategy, both takes current state into consideration. They bestPossibleVerifier may still work, but the state mapping can be different from one run to another run

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestAutoRebalance.java
```diff
@@ -229,6 +236,16 @@ static boolean verifyBalanceExternalView(ZNRecord externalView, int partitionCou
     return partitionCount == totalCount;
   }
 
+  private void setupAutoRebalancer() {
+    HelixAdmin admin = _gSetupTool.getClusterManagementTool();
```
kaisun2000, line 240: This actually relies on the fact that 

```
 _gSetupTool.addResourceToCluster(CLUSTER_NAME, "MyDB2", _PARTITIONS, "MasterSlave",
        RebalanceMode.FULL_AUTO + "");
```

would add  idea state using DelayedAutoRebalancer + AutoRebalance strategy

here, we change the idea state to use autoRebalancer, thus eventually we end up with autoRebalancer + autoRebalance strategy.

The thing is that if later people change the gSetupTool.addResourceToCluster() similar to testhelper to have DelayedAutoRebalancer + crushEd, this test won't work. 

Shall we add a comment that we depends on  _gSetupTool.addResourceToCluster to use default AutoRebalanceStrategy, or just note here we need autoRebalancer + autoRebalance strategy.

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestAutoRebalancePartitionLimit.java
```diff
@@ -54,7 +56,13 @@ public void beforeClass() throws Exception {
     _gSetupTool.addCluster(CLUSTER_NAME, true);
 
     _gSetupTool.addResourceToCluster(CLUSTER_NAME, TEST_DB, 100, "OnlineOffline",
-        RebalanceMode.FULL_AUTO + "", 0, 25);
+        RebalanceMode.FULL_AUTO.name(), 0, 25);
+
+    IdealState idealState =
+        _gSetupTool.getClusterManagementTool().getResourceIdealState(CLUSTER_NAME, TEST_DB);
+    idealState.setRebalancerClassName(AutoRebalancer.class.getName());
```
kaisun2000, line 63: add a comment, we need autoRebalancer + autoRebalance strategy. here.

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestAutoRebalancePartitionLimit.java
```diff
@@ -204,10 +215,14 @@ public boolean verify() {
...
-          accessor.getProperty(keyBuilder.externalView(_resourceName)).getRecord(),
-          numberOfPartitions, masterValue, replicas, cache.getLiveInstances().size(),
-          cache.getIdealState(_resourceName).getMaxPartitionsPerInstance());
+      try {
+        return verifyBalanceExternalView(
+            accessor.getProperty(keyBuilder.externalView(_resourceName)).getRecord(),
+            numberOfPartitions, masterValue, replicas, cache.getLiveInstances().size(),
+            cache.getIdealState(_resourceName).getMaxPartitionsPerInstance());
+      } catch (Exception e) {
+        return false;
```
kaisun2000, line 224: add a log.error please. It is better with stack trace. This would be tremendous helpful if the test fails.

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestAutoRebalanceWithDisabledInstance.java
```diff
@@ -43,6 +45,7 @@ public void beforeClass() throws Exception {
     super.beforeClass();
     _gSetupTool.addResourceToCluster(CLUSTER_NAME, TEST_DB_2, _PARTITIONS, STATE_MODEL,
         RebalanceMode.FULL_AUTO.name(), CrushEdRebalanceStrategy.class.getName());
+    setupAutoRebalancer();
```
kaisun2000, line 48: Again add the comment that we need autoRebalancer + autoRebalance strategy as this is autoRebalance test

kaisun2000 (approved):

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestAutoRebalance.java
jiajunwang: This test is for testing AutoRebalancer, so no matter what's the default behavior change, it is the right test logic. If it fails, then the new code change shall be fixed. I think there is no necessity of adding comment. Because this change only makes the logic more explicit.

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestAutoRebalancePartitionLimit.java
jiajunwang: Sure. Let me note this is to ensure we are testing AutoRebalancer.

helix-core/src/test/java/org/apache/helix/integration/rebalancer/TestAutoRebalancePartitionLimit.java
jiajunwang: Let me add a debug log here.
1. It is expected. Since the EV might not be created when the verify starts. So if we output an error message, it only pollutes the log with no good reason.
2. It will be retried in the loop. It would be too verbose if we output on every checks.

PR #1413, opened 2020-09-25
Feature: Offline Node Timeout During Maintenance Mode

### LINKED ISSUE #1412: Offline Node Timeout During Maintenance Mode
During the maintenance mode, partitions are assigned to nodes based on current states. It is possible for certain offline nodes to come back to life during the maintenance mode, in which case Helix doesn't differentiate them from other live nodes, and will assign them partitions based on their previous current states; this type of operations lead to state transitions from OFFLINE to another live state for the assigned partitions. From the customer point of view, this may cause new data bootstraps: if a node containing certain partitions has been offline for a certain period of time, data bootstrap may be required when it becomes live again. That is against the customer expectation of the maintenance mode.

Offline node time-out feature: if a node has been offline for a certain period of time, the maintenance mode rebalancer will not assign partitions to it.

NealSun96:
### Issues

Fixes #1412

### Description

As described in the issue, we would like to avoid the scenario where nodes that have been offline for a while come back to life during maintenance mode. Solution: if a node has been offline for a certain period of time, the maintenance mode rebalancer will not assign partitions to it. 

To accomplish this, two aspects of changes are necessary:
1. A new customer-defined value is required to enable this feature: MAINTENANCE_NODE_TIME_OUT. This value represents "how long a node has been offline such that it should be timed out". This value is set in ClusterConfig, similar to DELAY_REBALANCE_TIME.
2. A new logic should be added to the refreshLiveInstances() function of BaseControllerDataProvider. When the maintenance mode is enabled, find the new live nodes, and filter out the timed-out nodes; the timed-out nodes are treated as not live throughout the pipeline. 

### Tests

Everything under: TestParticipantHistory.java, TestOfflineNodeTimeoutDuringMaintenanceMode.java
testGetMaintenanceOfflineNodeTimeOut(), testSetMaintenanceOfflineNodeTimeOut()

```
[WARNING] Tests run: 1222, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 3,963.658 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[WARNING] Tests run: 1222, Failures: 0, Errors: 0, Skipped: 1
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:06 h
[INFO] Finished at: 2020-10-07T12:10:45-07:00

```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -471,31 +471,12 @@ public class BaseControllerDataProvider implements ControlContextProvider {
    * are timed-out if they have been offline for a while before going live during maintenance mode.
    */
   public Map<String, LiveInstance> getLiveInstances() {
-    return getLiveInstances(true);
-  }
-
-  /**
-   * Returns the LiveInstances for each of the instances that are currently up and running
-   */
-  public Map<String, LiveInstance> getAllLiveInstances() {
-    return getLiveInstances(false);
-  }
-
-  /**
-   * Returns the LiveInstances for each of the instances that are currently up and running
-   * @param excludeTimeoutInstances - Only effective during maintenance mode. If true, filter out
-   *                                instances that are timed-out during maintenance mode; instances
-   *                                are timed-out if they have been offline for a while before going
-   *                                live during maintenance mode.
-   */
-  private Map<String, LiveInstance> getLiveInstances(boolean excludeTimeoutInstances) {
-    if (excludeTimeoutInstances && isMaintenanceModeEnabled()) {
+    if (isMaintenanceModeEnabled()) {
       return _liveInstanceExcludeTimedOutForMaintenance;
     }
 
     return _liveInstanceCache.getPropertyMap();
   }
-
 
   /**
    * Return the set of all instances names.
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -761,7 +761,8 @@ public class BaseControllerDataProvider implements ControlContextProvider {
    * been offline for longer than the user defined timeout window.
    * @param timeOutWindow - the timeout window; guaranteed to be non-negative
    */
-  private boolean isInstanceTimedOutDuringMaintenance(HelixDataAccessor accessor, String instance, long timeOutWindow) {
+  private boolean isInstanceTimedOutDuringMaintenance(HelixDataAccessor accessor, String instance,
+      long timeOutWindow) {
     PropertyKey.Builder keyBuilder = accessor.keyBuilder();
     PropertyKey propertyKey = keyBuilder.participantHistory(instance);
     ParticipantHistory history = accessor.getProperty(propertyKey);
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -780,7 +780,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
       long timeOutWindow) {
     ParticipantHistory history =
         accessor.getProperty(accessor.keyBuilder().participantHistory(instance));
-    List<Long> onlineTimestamps = history.getHistoryTimestampsAsMilliseconds();
+    List<Long> onlineTimestamps = history.getOnlineTimestampsAsMilliseconds();
     List<Long> offlineTimestamps = history.getOfflineTimestampsAsMilliseconds();
 
     onlineTimestamps.add(System.currentTimeMillis());
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -109,6 +109,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
   private Map<String, Map<String, Set<String>>> _disabledInstanceForPartitionMap = new HashMap<>();
   private Set<String> _disabledInstanceSet = new HashSet<>();
   private final Map<String, MonitoredAbnormalResolver> _abnormalStateResolverMap = new HashMap<>();
+  private Set<String> _timedOutInstanceDuringMaintenance = new HashSet<>();
 
   public BaseControllerDataProvider() {
     this(AbstractDataCache.UNKNOWN_CLUSTER, AbstractDataCache.UNKNOWN_PIPELINE);
@@ -253,9 +254,34 @@ public class BaseControllerDataProvider implements ControlContextProvider {
   private void refreshLiveInstances(final HelixDataAccessor accessor,
       Set<HelixConstants.ChangeType> refreshedType) {
     if (_propertyDataChangedMap.get(HelixConstants.ChangeType.LIVE_INSTANCE).getAndSet(false)) {
+      // Keep a copy of old live instances in case of maintenance mode
+      Map<String, LiveInstance> oldLiveInstances = getLiveInstances();
       _liveInstanceCache.refresh(accessor);
       _updateInstanceOfflineTime = true;
       refreshedType.add(HelixConstants.ChangeType.LIVE_INSTANCE);
+
+      // If maintenance mode is enabled and timeout window is specified, filter 'new' live nodes
+      // for timed-out nodes
+      long timeOutWindow = _clusterConfig.getMaintenanceOfflineNodeTimeOut();
+      if (timeOutWindow >= 0 && isMaintenanceModeEnabled()) {
+        for (String instance : _liveInstanceCache.getPropertyMap().keySet()) {
+          // For every 'new' live node, check if it's timed-out
+          if (!oldLiveInstances.containsKey(instance) && isInstanceTimedOutDuringMaintenance(
+              accessor, instance, timeOutWindow)) {
+            _timedOutInstanceDuringMaintenance.add(instance);
+          }
+        }
+
+        // Remove all timed-out nodes that were recorded in this maintenance duration
+        for (String instance : _timedOutInstanceDuringMaintenance) {
+          _liveInstanceCache.deletePropertyByName(instance);
+        }
+      }
+
+      // If maintenance mode has exited, clear cached timed-out nodes
+      if (!isMaintenanceModeEnabled()) {
+        _timedOutInstanceDuringMaintenance.clear();
+      }
     } else {
       LogUtil.logInfo(logger, getClusterEventId(), String
           .format("No live instance change for %s cluster, %s pipeline", _clusterName,
@@ -679,6 +705,8 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     List<String> offlineNodes =
         new ArrayList<>(_instanceConfigCache.getPropertyMap().keySet());
     offlineNodes.removeAll(_liveInstanceCache.getPropertyMap().keySet());
+    // The timed-out instances during maintenance are online and not included in liveInstanceCache
+    offlineNodes.removeAll(_timedOutInstanceDuringMaintenance);
     _instanceOfflineTimeMap = new HashMap<>();
 
     for (String instance : offlineNodes) {
@@ -726,6 +754,40 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     if (_clusterConfig != null && _clusterConfig.getDisabledInstances() != null) {
       _disabledInstanceSet.addAll(_clusterConfig.getDisabledInstances().keySet());
     }
+  }
+
+  /*
+   * Check if the instance is timed-out during maintenance mode. An instance is timed-out if it has
+   * been offline for longer than the user defined timeout window.
+   * @param timeOutWindow - the timeout window; guaranteed to be non-negative
+   */
+  private boolean isInstanceTimedOutDuringMaintenance(HelixDataAccessor accessor, String instance, long timeOutWindow) {
+    PropertyKey.Builder keyBuilder = accessor.keyBuilder();
+    PropertyKey propertyKey = keyBuilder.participantHistory(instance);
+    ParticipantHistory history = accessor.getProperty(propertyKey);
+
+    // Cannot use _instanceOfflineTimeMap. _instanceOfflineTimeMap is last updated in the previous
+    // pipeline execution; it is possible a new offline timestamp is updated after the previous
+    // pipeline execution, so we need the most updated timestamp.
+    long lastOfflineTime = history.getLastTimeInOfflineHistory();
+    // lastOfflineTime is only negative if there is no offline history or the time format is wrong.
+    // Since this instance is a 'new' live instance, not having offline history = first time created
+    // instance; during maintenance mode, no partition will be assigned to such a new instance,
+    // therefore it's okay to no time it out. The wrong format case shouldn't happen at all and will
+    // not be handled either.
+    if (lastOfflineTime < 0) {
+      return false;
+    }
+
+    long currentTime = System.currentTimeMillis();
+    if (currentTime - lastOfflineTime > timeOutWindow) {
+      LogUtil.logWarn(logger, getClusterEventId(), String.format(
+          "During maintenance mode, instance %s is timed-out due to its offline time. Current time: "
+              + "%s, Last offline time: %s, Timeout window: %s", instance, currentTime,
+          lastOfflineTime, timeOutWindow));
+      return true;
+    }
+    return false;
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -112,6 +112,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
   private final Map<String, MonitoredAbnormalResolver> _abnormalStateResolverMap = new HashMap<>();
   private Set<String> _timedOutInstanceDuringMaintenance = new HashSet<>();
   private Map<String, LiveInstance> _liveInstanceSnapshotForMaintenance = new HashMap<>();
+  private Map<String, LiveInstance> _liveInstanceExcludeTimedOutForMaintenance;
 
   public BaseControllerDataProvider() {
     this(AbstractDataCache.UNKNOWN_CLUSTER, AbstractDataCache.UNKNOWN_PIPELINE);
@@ -333,6 +334,10 @@ public class BaseControllerDataProvider implements ControlContextProvider {
         }
       }
     }
+    _liveInstanceExcludeTimedOutForMaintenance =
+        _liveInstanceCache.getPropertyMap().entrySet().stream()
+            .filter(e -> !_timedOutInstanceDuringMaintenance.contains(e.getKey()))
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
   }
 
   private void updateIdealRuleMap() {
@@ -472,20 +477,18 @@ public class BaseControllerDataProvider implements ControlContextProvider {
 
   /**
    * Returns the LiveInstances for each of the instances that are currently up and running
-   * @param filterTimedOutInstances - Only set true during maintenance mode. If true, filter out
+   * @param excludeTimeoutInstances - Only set true during maintenance mode. If true, filter out
    *                                instances that are timed-out during maintenance mode; instances
    *                                are timed-out if they have been offline for a while before going
    *                                live during maintenance mode.
    * @return
    */
-  public Map<String, LiveInstance> getLiveInstances(boolean filterTimedOutInstances) {
-    if (!filterTimedOutInstances) {
+  public Map<String, LiveInstance> getLiveInstances(boolean excludeTimeoutInstances) {
+    if (!excludeTimeoutInstances) {
       return _liveInstanceCache.getPropertyMap();
     }
 
-    return _liveInstanceCache.getPropertyMap().entrySet().stream()
-        .filter(e -> !_timedOutInstanceDuringMaintenance.contains(e.getKey()))
-        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    return _liveInstanceExcludeTimedOutForMaintenance;
   }
 
 
@@ -824,7 +827,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
       long tailOfflineTime;
       for (int i = historyOfflineList.size() - 1; i >= 0; i--) {
         tailOfflineTime =
-            ParticipantHistory.parseHistoryDateStringToLong(historyOfflineList.get(i));
+            ParticipantHistory.historyDateStringToLong(historyOfflineList.get(i));
         // Ignore bad format case
         if (tailOfflineTime != -1) {
           if (tailOfflineTime <= onlineTimestamps.get(0)) {
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -255,8 +255,11 @@ public class BaseControllerDataProvider implements ControlContextProvider {
   private void refreshLiveInstances(final HelixDataAccessor accessor,
       Set<HelixConstants.ChangeType> refreshedType) {
     if (_propertyDataChangedMap.get(HelixConstants.ChangeType.LIVE_INSTANCE).getAndSet(false)) {
-      // Keep a copy of old live instances in case of maintenance mode
-      Map<String, LiveInstance> oldLiveInstances = getLiveInstances();
+      // Keep a copy of old live instances for maintenance mode
+      Map<String, LiveInstance> oldLiveInstances = new HashMap<>();
+      if (isMaintenanceModeEnabled()) {
+        oldLiveInstances = getLiveInstances(true);
+      }
       _liveInstanceCache.refresh(accessor);
       _updateInstanceOfflineTime = true;
       refreshedType.add(HelixConstants.ChangeType.LIVE_INSTANCE);
@@ -723,8 +726,6 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     List<String> offlineNodes =
         new ArrayList<>(_instanceConfigCache.getPropertyMap().keySet());
     offlineNodes.removeAll(_liveInstanceCache.getPropertyMap().keySet());
-    // The timed-out instances during maintenance are online and not included in liveInstanceCache
-    offlineNodes.removeAll(_timedOutInstanceDuringMaintenance);
     _instanceOfflineTimeMap = new HashMap<>();
 
     for (String instance : offlineNodes) {
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -111,6 +111,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
   private Set<String> _disabledInstanceSet = new HashSet<>();
   private final Map<String, MonitoredAbnormalResolver> _abnormalStateResolverMap = new HashMap<>();
   private Set<String> _timedOutInstanceDuringMaintenance = new HashSet<>();
+  private Map<String, LiveInstance> _liveInstanceSnapshotForMaintenance = new HashMap<>();
 
   public BaseControllerDataProvider() {
     this(AbstractDataCache.UNKNOWN_CLUSTER, AbstractDataCache.UNKNOWN_PIPELINE);
@@ -255,30 +256,13 @@ public class BaseControllerDataProvider implements ControlContextProvider {
   private void refreshLiveInstances(final HelixDataAccessor accessor,
       Set<HelixConstants.ChangeType> refreshedType) {
     if (_propertyDataChangedMap.get(HelixConstants.ChangeType.LIVE_INSTANCE).getAndSet(false)) {
-      // Keep a copy of old live instances for maintenance mode
-      Map<String, LiveInstance> oldLiveInstances = new HashMap<>();
+      // Keep a snapshot of old live instances for maintenance mode
       if (isMaintenanceModeEnabled()) {
-        oldLiveInstances = getLiveInstances(true);
+        _liveInstanceSnapshotForMaintenance = getLiveInstances();
       }
       _liveInstanceCache.refresh(accessor);
       _updateInstanceOfflineTime = true;
       refreshedType.add(HelixConstants.ChangeType.LIVE_INSTANCE);
-
-      // If maintenance mode is enabled and timeout window is specified, filter 'new' live nodes
-      // for timed-out nodes
-      long timeOutWindow = -1;
-      if (_clusterConfig != null) {
-        timeOutWindow = _clusterConfig.getOfflineNodeTimeOutForMaintenanceMode();
-      }
-      if (timeOutWindow >= 0 && isMaintenanceModeEnabled()) {
-        for (String instance : _liveInstanceCache.getPropertyMap().keySet()) {
-          // For every 'new' live node, check if it's timed-out
-          if (!oldLiveInstances.containsKey(instance) && isInstanceTimedOutDuringMaintenance(
-              accessor, instance, timeOutWindow)) {
-            _timedOutInstanceDuringMaintenance.add(instance);
-          }
-        }
-      }
     } else {
       LogUtil.logInfo(logger, getClusterEventId(), String
           .format("No live instance change for %s cluster, %s pipeline", _clusterName,
@@ -329,6 +313,28 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     }
   }
 
+  private void timeoutNodesDuringMaintenance(final HelixDataAccessor accessor) {
+    // If maintenance mode is enabled and timeout window is specified, filter 'new' live nodes
+    // for timed-out nodes
+    long timeOutWindow = -1;
+    if (_clusterConfig != null) {
+      timeOutWindow = _clusterConfig.getOfflineNodeTimeOutForMaintenanceMode();
+    }
+    if (timeOutWindow >= 0 && isMaintenanceModeEnabled()) {
+      for (String instance : _liveInstanceCache.getPropertyMap().keySet()) {
+        // 1. Check timed-out cache and don't do repeated work;
+        // 2. Check for nodes that didn't exist in the last iteration, because it has been checked;
+        // 3. For all other nodes, check if it's timed-out.
+        // When maintenance mode is first entered, all nodes will be checked as a result.
+        if (!_timedOutInstanceDuringMaintenance.contains(instance)
+            && !_liveInstanceSnapshotForMaintenance.containsKey(instance)
+            && isInstanceTimedOutDuringMaintenance(accessor, instance, timeOutWindow)) {
+          _timedOutInstanceDuringMaintenance.add(instance);
+        }
+      }
+    }
+  }
+
   private void updateIdealRuleMap() {
     // Assumes cluster config is up-to-date
     if (_clusterConfig != null) {
@@ -361,6 +367,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     _stateModelDefinitionCache.refresh(accessor);
     _clusterConstraintsCache.refresh(accessor);
     updateMaintenanceInfo(accessor);
+    timeoutNodesDuringMaintenance(accessor);
 
     // TODO: once controller gets split, only one controller should update offline instance history
     updateOfflineInstanceHistory(accessor);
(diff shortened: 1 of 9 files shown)


helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -253,9 +254,34 @@ private void refreshIdealState(final HelixDataAccessor accessor,
...
       _updateInstanceOfflineTime = true;
       refreshedType.add(HelixConstants.ChangeType.LIVE_INSTANCE);
+
+      // If maintenance mode is enabled and timeout window is specified, filter 'new' live nodes
+      // for timed-out nodes
+      long timeOutWindow = _clusterConfig.getMaintenanceOfflineNodeTimeOut();
+      if (timeOutWindow >= 0 && isMaintenanceModeEnabled()) {
+        for (String instance : _liveInstanceCache.getPropertyMap().keySet()) {
+          // For every 'new' live node, check if it's timed-out
+          if (!oldLiveInstances.containsKey(instance) && isInstanceTimedOutDuringMaintenance(
```
dasahcc, line 269: I think it may not be the place to do that. We can do it in maintenance recovery stage. I cannot remember the details of how to move the last offline time to history and set -1. It happened at cache refresh as well. So I think it is better to make it after cache refresh fully complete.

NealSun96: It needs to be here. The reason is that we need the old snapshot; see line 258. Once refresh happens, the old snapshot will be lost, and we wouldn't be able to compute the 'new' live nodes delta. 

In terms of history update, it's updateOfflineInstanceHistory() down below. You can see that I also handled that case.

dasahcc: Then that may not be the snapshot we need. Because, this change could cause race condition and let old machine starting bootstrapping, which breaks our rule.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
```diff
@@ -298,6 +298,26 @@ public void testAsyncGlobalRebalanceOption() {
             false), true);
   }
 
+  @Test
+  public void testGetMaintenanceOfflineNodeTimeOut() {
```
jiajunwang, line 302: 1. We need a test to cover the get default value case.
2. Might be easier for you to combine the 2 new tests.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
```diff
@@ -40,7 +41,9 @@
   private static Logger LOG = LoggerFactory.getLogger(ParticipantHistory.class);
 
   private final static int HISTORY_SIZE = 20;
-  private enum ConfigProperty {
+  final static String HISTORY_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss:SSS";
```
jiajunwang, line 44: private?

jiajunwang: If required in the test, then either put it in HelixConstants or redefine a string in the test case (this helps to prevent if anyone changes the format that breaks the compatibility)

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: Could you elaborate why there would be race conditions? The cache refresh is synchronously computed in the pipeline main thread; rebalancing is calculated on the same pipeline.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -253,9 +254,34 @@ private void refreshIdealState(final HelixDataAccessor accessor,
...
+              accessor, instance, timeOutWindow)) {
+            _timedOutInstanceDuringMaintenance.add(instance);
+          }
+        }
+
+        // Remove all timed-out nodes that were recorded in this maintenance duration
+        for (String instance : _timedOutInstanceDuringMaintenance) {
+          _liveInstanceCache.deletePropertyByName(instance);
```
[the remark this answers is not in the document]
NealSun96: That's a good idea. Will try.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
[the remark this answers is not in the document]
NealSun96: `oldLiveInstances` is for performance only. Losing it during leadership switch is still fine.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -728,6 +756,41 @@ private void updateDisabledInstances() {
...
+    // therefore it's okay to no time it out. The wrong format case shouldn't happen at all and will
+    // not be handled either.
+    if (lastOfflineTime < 0) {
+      return false;
+    }
+
+    long currentTime = System.currentTimeMillis();
+    if (currentTime - lastOfflineTime > timeOutWindow) {
+      LogUtil.logWarn(logger, getClusterEventId(), String.format(
+          "During maintenance mode, instance %s is timed-out due to its offline time. Current time: "
```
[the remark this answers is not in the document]
NealSun96: Why is that? This calculation is only done during the maintenance mode.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -728,6 +756,41 @@ private void updateDisabledInstances() {
...
+    // Cannot use _instanceOfflineTimeMap. _instanceOfflineTimeMap is last updated in the previous
+    // pipeline execution; it is possible a new offline timestamp is updated after the previous
+    // pipeline execution, so we need the most updated timestamp.
+    long lastOfflineTime = history.getLastTimeInOfflineHistory();
+    // lastOfflineTime is only negative if there is no offline history or the time format is wrong.
+    // Since this instance is a 'new' live instance, not having offline history = first time created
+    // instance; during maintenance mode, no partition will be assigned to such a new instance,
+    // therefore it's okay to no time it out. The wrong format case shouldn't happen at all and will
+    // not be handled either.
+    if (lastOfflineTime < 0) {
```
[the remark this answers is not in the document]
NealSun96: This isn't the actual `LAST_OFFLINE_TIME` field in History; I can understand the confusion due to variable name. 

I implemented a new function called `getLastTimeInOfflineHistory()` where the latest date in `OFFLINE` is returned. In the case when that cannot be returned, -1 will be the error indicator; you could refer the comment above for more detail.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -728,6 +756,41 @@ private void updateDisabledInstances() {
...
+  private boolean isInstanceTimedOutDuringMaintenance(HelixDataAccessor accessor, String instance,
+      long timeOutWindow) {
+    PropertyKey.Builder keyBuilder = accessor.keyBuilder();
+    PropertyKey propertyKey = keyBuilder.participantHistory(instance);
+    ParticipantHistory history = accessor.getProperty(propertyKey);
+
+    // Cannot use _instanceOfflineTimeMap. _instanceOfflineTimeMap is last updated in the previous
+    // pipeline execution; it is possible a new offline timestamp is updated after the previous
+    // pipeline execution, so we need the most updated timestamp.
+    long lastOfflineTime = history.getLastTimeInOfflineHistory();
```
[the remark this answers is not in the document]
NealSun96: 1. Sounds good;
2. I will add that in the javadoc, what do you think? I don't think it should be strictly enforced.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -122,7 +122,13 @@
...
     // none of participants or the cluster define pool sizes,
     // TaskConstants.DEFAULT_TASK_THREAD_POOL_SIZE will be used to create pool sizes.
-    GLOBAL_TARGET_TASK_THREAD_POOL_SIZE
+    GLOBAL_TARGET_TASK_THREAD_POOL_SIZE,
+
+    // The time out window for offline nodes during maintenance mode; if an offline node has been
+    // offline for more than this specified time period, it's treated as offline for the rest of
+    // the maintenance mode's duration even when it comes online.
+    // The unit is milliseconds.
+    MAINTENANCE_OFFLINE_NODE_TIME_OUT
```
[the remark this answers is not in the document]
NealSun96: Sure, but wouldn't that be too long? I don't know the convention about variable length, so I took a shorter version.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
NealSun96: Sounds good.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
NealSun96: That's a great idea; I'll just redefine it in test.

dasahcc (approved):
Good job!

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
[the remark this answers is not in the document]
NealSun96: Come to think about it, it's not fine: if `oldLiveInstances` is lost, every node is subject to timeout check; however, the timeout check only regards the node's last offline time. That means if a node is online for 8 weeks and last offline time is 9 weeks ago, we will still mark it as timed_out. That would be incorrect. Let me think about this case and update. @jiajunwang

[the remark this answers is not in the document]
NealSun96: The new method should address the problem.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -728,6 +782,100 @@ private void updateDisabledInstances() {
...
+      long onlineTimestamp = onlineTimestamps.get(onlineTimestampIndex);
+      long offlineTimeStamp = offlineTimestamps.get(offlineTimestampIndex);
+
+      // If the offline timestamp isn't within this pair of online timestamp, continue
+      if (offlineTimeStamp > onlineTimestamp) {
+        continue;
+      }
+
+      // Check the largest offline window against the timeout window
+      if (onlineTimestamp - offlineTimeStamp > timeOutWindow) {
```
dasahcc, line 861: It has the edge case for first interval. It is the interval online time just before maintenance mode, but somehow we dont check whether the offline time is before maintenance.

Let's say if the instance is offline before maintenance we need the potential timeout = online time - Math.max(offlinetimestamp, maintenance entering time);

NealSun96: The first offline time is always after the first online time based on my parsing logic, see above. 

Therefore we always start checking by the second online time, and do (second online time - first offline time), for example.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
```diff
@@ -112,6 +129,22 @@ public long getLastOfflineTime() {
     return offlineTime;
   }
 
+  /**
+   * Get the time when this node last goes offline in history. If the node does not have offline
+   * history or contains invalid date as the last element, return -1.
+   *
+   * @return
+   */
+  public long getLastTimeInOfflineHistory() {
```
lei-xia, line 138: What is difference this w/ getLastOfflineTime?

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
```diff
@@ -153,16 +183,73 @@ private void updateOfflineHistory(long time) {
...
   @Override
   public boolean isValid() {
     return true;
   }
+
+  /**
+   * Parses a history date in string format to its millisecond representation.
+   * Returns -1 if parsing fails.
+   */
+  public static long parseHistoryDateStringToLong(String dateString) {
```
lei-xia, line 198: minor:  -> historyDateStringToLong?

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
NealSun96: `getLastOfflineTime` returns LAST_OFFLINE_TIME, which can be either a timestamp or a -1 for ONLINE. `getLastTimeInOfflineHistory` returns the last timestamp in OFFLINE (list field), which is always a timestamp.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -428,13 +463,32 @@ public synchronized void setIdealStates(List<IdealState> idealStates) {
...
+
+  /**
+   * Returns the LiveInstances for each of the instances that are currently up and running
+   * @param filterTimedOutInstances - Only set true during maintenance mode. If true, filter out
+   *                                instances that are timed-out during maintenance mode; instances
+   *                                are timed-out if they have been offline for a while before going
+   *                                live during maintenance mode.
+   * @return
+   */
+  public Map<String, LiveInstance> getLiveInstances(boolean filterTimedOutInstances) {
```
lei-xia, line 481: filterTimedOutInstances is a bit confusing, it could mean either the result contains only Timeouted Instances or the other way. Suggest to rename it as:  excludeTimeoutInstances.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -299,6 +306,33 @@ private void updateMaintenanceInfo(final HelixDataAccessor accessor) {
...
+    if (timeOutWindow >= 0 && isMaintenanceModeEnabled()) {
+      for (String instance : _liveInstanceCache.getPropertyMap().keySet()) {
+        // 1. Check timed-out cache and don't do repeated work;
+        // 2. Check for nodes that didn't exist in the last iteration, because it has been checked;
+        // 3. For all other nodes, check if it's timed-out.
+        // When maintenance mode is first entered, all nodes will be checked as a result.
+        if (!_timedOutInstanceDuringMaintenance.contains(instance)
+            && !_liveInstanceSnapshotForMaintenance.containsKey(instance)
+            && isInstanceTimedOutDuringMaintenance(accessor, instance, timeOutWindow)) {
+          _timedOutInstanceDuringMaintenance.add(instance);
```
lei-xia, line 332: There could be some issue with this if a long-offline instance (say InstanceA) came back just after the cluster enter maintenance mode and before the next pipeline started.  InstanceA will be in the _liveInstanceSnapshotForMaintenance set (see line 261), so it will skip the timeout check here.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
lei-xia, line 481: Can we cache the results somewhere (we may already have it?) instead of computing everytime this method is called?  I think gertLiveInstances are called in many places during one pipeline.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -428,13 +463,32 @@ public synchronized void setIdealStates(List<IdealState> idealStates) {
   }
 
   /**
-   * Returns the LiveInstances for each of the instances that are curretnly up and running
+   * Returns the LiveInstances for each of the instances that are currently up and running
    * @return
    */
   public Map<String, LiveInstance> getLiveInstances() {
```
lei-xia, line 469: In which places we will need  LiveInstances containing timeouted instance?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: For the first time it enters maintenance mode, there will be no snapshot is stored. It's because maintenance mode signal is processed after refreshLiveInstances(). This way, all nodes will be checked when we first enter maintenance mode.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: Makes sense.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: In short: everywhere else that isn't critical to "timing out" instances. 

The new `getLiveInstances(boolean)` is minimized to only one place: `MessageGenerationPhase`. This approach minimizes any impact to existing code while also making the "timed-out" instances not reachable from the controller; effectively, those instances will have no operations done on them. This decision is made after @jiajunwang 's comment that we shouldn't directly alter the live instance cache; we should only alter the getter. 

Please let me know if this change makes sense. I believe only changing `MessageGenerationPhase` is sufficient.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -89,7 +89,7 @@ protected void processEvent(ClusterEvent event, ResourcesStateMap resourcesState
           + ". Requires HelixManager|DataCache|RESOURCES|CURRENT_STATE|INTERMEDIATE_STATE");
     }
 
-    Map<String, LiveInstance> liveInstances = cache.getLiveInstances();
+    Map<String, LiveInstance> liveInstances = cache.getLiveInstances(true);
```
lei-xia, line 92: Could this create inconsistency since LivInstances used by all other places in the pipeline are different from here?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -728,6 +782,100 @@ private void updateDisabledInstances() {
...
+        tailOnlineTime =
+            ParticipantHistory.extractTimeFromSessionHistoryString(historyOnlineList.get(i));
+        // Ignore bad format case
+        if (tailOnlineTime != -1) {
+          onlineTimestamps.add(0, tailOnlineTime);
+          // We only need one online timestamp before maintenance mode starts
+          if (tailOnlineTime <= _maintenanceSignal.getTimestamp()) {
+            break;
+          }
+        }
```
lei-xia, line 814: Can we wrap up the logic here into a method in ParticipantHistory, such as getAllOnlineHistory()?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -728,6 +782,100 @@ private void updateDisabledInstances() {
...
+        tailOfflineTime =
+            ParticipantHistory.parseHistoryDateStringToLong(historyOfflineList.get(i));
+        // Ignore bad format case
+        if (tailOfflineTime != -1) {
+          if (tailOfflineTime <= onlineTimestamps.get(0)) {
+            break;
+          }
+          offlineTimestamps.add(0, tailOfflineTime);
+        }
+      }
```
lei-xia: Similar here, the logic should belong to ParticipantHistory, such as getAllOfflineHistory().

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -728,6 +782,100 @@ private void updateDisabledInstances() {
...
+    }
+
+    // At this point, onlineTimestamps contains at least 1 timestamp that's before maintenance mode;
+    // offlineTimestamps contains 0+ timestamp that's > the first online timestamp.
```
lei-xia, line 842: This may not be true, for example, a newly added host?  (I.e, our code should be able to handle these corner cases properly).

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
lei-xia: Do we really need to go through every pair? Can we just look at the pair with offline time before maintenance entrance and online time after it?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: This ties back to the previous 2 sections of algorithm: they are not just parsing the dates. The first section stops parsing as soon as the first timestamp that's before maintenance mode; if there's no such timestamp, it just returns false (for newly added host). The second section stops parsing as soon as the timestamp is just after the first online timestamp. This is also why these logic should not be a member logic to ParticipantHistory because they are not meant to be generic.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: Please see below.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: Please see below.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: Actually I'm rethinking this. There could be some external view issue. Will update.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
NealSun96: Duplicate discussion with the above comment.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: Now getLiveInstances will default to excluding timed-out ones.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: @lei-xia We need to handle the case of "offline for too long" during maintenance mode as well, right?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
lei-xia: _liveInstanceSnapshotForMaintenance will be refreshed in the beginning of every pipeline, it contains all liveInstances (including these that should be timeout-ed), right? Say, if an (long-offline) instance comes back after the first pipeline before the next pipeline, that instance will be included in the _liveInstanceSnapshotForMaintenance and won't be checked here?   I.e, is _liveInstanceSnapshotForMaintenance 
 always equal to getLiveInstances()?  If it is, what is point of keeping a separate cache?

NealSun96: No, `_liveInstanceSnapshotForMaintenance` is equal to `getLiveInstances()` from the *last* pipeline. If a long-offline instance coms back after the first pipeline and before the next pipeline, it will *not* be in the snapshot, and will therefore be checked by the timeout logic. Whether this instance fails the check, it will be stored either in the snapshot or in the "timed-out cache", and it will not be checked by the pipeline after.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -253,6 +257,10 @@ private void refreshIdealState(final HelixDataAccessor accessor,
   private void refreshLiveInstances(final HelixDataAccessor accessor,
       Set<HelixConstants.ChangeType> refreshedType) {
     if (_propertyDataChangedMap.get(HelixConstants.ChangeType.LIVE_INSTANCE).getAndSet(false)) {
+      // Keep a snapshot of old live instances for maintenance mode
+      if (isMaintenanceModeEnabled()) {
```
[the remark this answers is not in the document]
NealSun96: This is intentional. When maintenance mode is first started, the snapshot is left empty so that all nodes are checked.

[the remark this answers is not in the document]
NealSun96: Discussed offline. This variable is removed and replaced by the cached result of live instances excluding timed-out ones. That way the intention is more clear.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -299,6 +307,40 @@ private void updateMaintenanceInfo(final HelixDataAccessor accessor) {
...
+  }
+
+  private void timeoutNodesDuringMaintenance(final HelixDataAccessor accessor) {
+    // If maintenance mode is enabled and timeout window is specified, filter 'new' live nodes
+    // for timed-out nodes
+    long timeOutWindow = -1;
+    if (_clusterConfig != null) {
+      timeOutWindow = _clusterConfig.getOfflineNodeTimeOutForMaintenanceMode();
+    }
+    if (timeOutWindow >= 0 && isMaintenanceModeEnabled()) {
```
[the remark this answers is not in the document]
NealSun96: Done.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -299,6 +307,40 @@ private void updateMaintenanceInfo(final HelixDataAccessor accessor) {
...
+      for (String instance : _liveInstanceCache.getPropertyMap().keySet()) {
+        // 1. Check timed-out cache and don't do repeated work;
+        // 2. Check for nodes that didn't exist in the last iteration, because it has been checked;
```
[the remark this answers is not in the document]
NealSun96: Discussed offline: the snapshot variable is replaced and now it's okay for ZK to be disrupted at any point. `_timedOutInstanceDuringMaintenance` can be partially completed - other nodes will be checked again. The new snapshot variable (which is just `_liveInstanceExcludeTimedOutForMaintenance`) is computed after this step and will always store the correct truth of "what instances are checked"; it can be partially completed - other nodes will be checked again.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -428,13 +471,30 @@ public synchronized void setIdealStates(List<IdealState> idealStates) {
...
+
+  /**
+   * Returns the LiveInstances for each of the instances that are currently up and running
+   * @param excludeTimeoutInstances - Only effective during maintenance mode. If true, filter out
+   *                                instances that are timed-out during maintenance mode; instances
+   *                                are timed-out if they have been offline for a while before going
+   *                                live during maintenance mode.
+   * @return
+   */
+  public Map<String, LiveInstance> getLiveInstances(boolean excludeTimeoutInstances) {
```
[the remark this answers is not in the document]
NealSun96: Discussed offline: function is made private. It can be extended later on.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: Moved to ParticipantHistory

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: Moved to ParticipantHistory

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: The algorithm has been reworked to look less complicated. Its essence remains - all mentioned edge cases are taken care of.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: This has been changed. We now rely on the previous calculation result, which makes more sense when reading.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
```diff
@@ -64,6 +69,20 @@ public ParticipantHistory(ZNRecord znRecord) {
...
+    return _record.getListField(ConfigProperty.HISTORY.name());
+  }
+
+  /**
+   * @return The list field for OFFLINE
+   */
+  public List<String> getOffline() {
+    return _record.getListField(ConfigProperty.OFFLINE.name());
+  }
+
```
jiajunwang, line 85: I think you don't need these 2 methods to be public?

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
```diff
@@ -153,16 +185,99 @@ private void updateOfflineHistory(long time) {
...
+   */
+  private static long extractTimeFromSessionHistoryString(String sessionHistoryString) {
+    Map<String, String> sessionHistoryMap = parseSessionHistoryStringToMap(sessionHistoryString);
+    if (!sessionHistoryMap.containsKey(ConfigProperty.TIME.name())) {
+      return -1;
+    }
+    try {
+      return Long.parseLong(sessionHistoryMap.get(ConfigProperty.TIME.name()));
+    } catch (NumberFormatException e) {
+      return -1;
```
jiajunwang, line 256: Log the error here.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
```diff
@@ -153,16 +185,99 @@ private void updateOfflineHistory(long time) {
...
+    }
+  }
+
+  /**
+   * For each entry in History, return its millisecond timestamp; for timestamps that cannot be
+   * parsed, skip them.
+   */
+  public List<Long> getHistoryTimestampsAsMilliseconds() {
```
jiajunwang, line 264: Shall we just call it getOnlineTimestampsAsMilliseconds?

jiajunwang (approved):
In general looks good. I left several minor comments. Please address.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
NealSun96: They are public because 
1. we should provide accessors since now there are actual code usage of ParticipantHistory; and, 
2. an integration test calls these functions.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
NealSun96: I had this debate too, but I think in the context of ParticipantHistory, the field names should be respected. The field is HISTORY.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -333,6 +370,7 @@ public synchronized void refresh(HelixDataAccessor accessor) {
     _stateModelDefinitionCache.refresh(accessor);
     _clusterConstraintsCache.refresh(accessor);
     updateMaintenanceInfo(accessor);
+    timeoutNodesDuringMaintenance(accessor, _clusterConfig, _isMaintenanceModeEnabled);
```
lei-xia, line 373: Any specific reason we do not put this into refreshLiveInstances()?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -428,13 +466,37 @@ public synchronized void setIdealStates(List<IdealState> idealStates) {
...
+   * are timed-out if they have been offline for a while before going live during maintenance mode.
    */
   public Map<String, LiveInstance> getLiveInstances() {
+    return getLiveInstances(true);
+  }
+
+  /**
+   * Returns the LiveInstances for each of the instances that are currently up and running
+   */
+  public Map<String, LiveInstance> getAllLiveInstances() {
```
lei-xia, line 480: These two methods (getLiveInstances() and getAllLiveInstances() are very confusing to each other.  We would need a better naming here.  We have four type of instances:  All Instances, Live Instances, enabled/disabled Instances, timeout-ed instances.  We have a method called getEnabledLiveInstances(), maybe we can name this new method getNonTimeoutLiveInstances()?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -728,6 +790,58 @@ private void updateDisabledInstances() {
     }
   }
 
+  /*
+   * Check if the instance is timed-out during maintenance mode. An instance is timed-out if it has
+   * been offline for longer than the user defined timeout window.
+   * @param timeOutWindow - the timeout window; guaranteed to be non-negative
+   */
+  private boolean isInstanceTimedOutDuringMaintenance(HelixDataAccessor accessor, String instance,
```
lei-xia, line 798: minor:  isInstanceTimeout... -> shouldInstanceTimeout...

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
```diff
@@ -153,16 +185,101 @@ private void updateOfflineHistory(long time) {
...
+    return df.format(new Date(dateLong));
+  }
+
+  /**
+   * Parses the session entry map that has been converted to string back to a map.
+   * NOTE TO CALLER: This assumes the divider between entries is ", " and the divider between
+   * key/value is "="; if the string is malformed, parsing correctness is not guaranteed. Always
+   * check if a key is contained before using the key.
+   */
+  public static Map<String, String> parseSessionHistoryStringToMap(String sessionHistoryString) {
```
lei-xia, line 227: Let us follow the name convention of another method, such as sessionHistoryStringToMap()?

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
```diff
@@ -153,16 +185,101 @@ private void updateOfflineHistory(long time) {
...
+    }
+
+    return sessionHistoryMap;
+  }
+
+  /*
+   * Take a string session history entry and extract the TIME field out of it. Return -1 if the TIME
+   * field doesn't exist or if the TIME field cannot be parsed to a long.
+   */
+  private static long extractTimeFromSessionHistoryString(String sessionHistoryString) {
```
lei-xia, line 248: getTimeFromSessionHistory

helix-core/src/test/java/org/apache/helix/integration/controller/TestOfflineNodeTimeoutDuringMaintenanceMode.java
```diff
@@ -0,0 +1,242 @@
+package org.apache.helix.integration.controller;
```
lei-xia, line 1: Do we have an end-to-end integration test here, such as verify that a timeout-ed instance during maintenance mode will not get assigned any partitions (received any state transition messages)?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: Yes, because we want this step to be after updateMaintenanceInfo(), which makes the most logical sense.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: I'm just going to remove this function and the boolean flag, since during maintenance mode, we always want to return the non-timed-out instances.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: It's isInstanceTimedOut, which is grammatically correct. I'd prefer is over should.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
NealSun96: Ok.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
NealSun96: Ok.

helix-core/src/test/java/org/apache/helix/integration/controller/TestOfflineNodeTimeoutDuringMaintenanceMode.java
NealSun96: I will cover that in a CHO test.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
jiajunwang: But it is not clear to the caller what do you get.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
jiajunwang: 1. you have the methods which return the timestamps. The raw string is hard to use and we shall not return it.
2. integration test can rely on protected methods.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
NealSun96: I understand what you mean, and since I also want to call it getOnlineTimestamps, I made the change. 

Though I don't think the old naming is bad: callers get the History timestamps, as advertised. If they don't know what these are they should read the class. 😃

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
NealSun96: I disagree with this. For a HelixProperty class, if there is a need of a field, there should be an accessor. It is true that we have timestamp parsers now, but for example, HISTORY contains other informations too. Down the road, there may be other code that relies on other HISTORY information. What would happen then? An accessor will be built. Right now there is a need to these fields, and I don't see a reason to not create accessors. 

For integration test it's not in the same package, so I don't think package private (nor protected) can help. This is a minor point; there are workarounds. My main point is the previous paragraph.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
jiajunwang: A meaningful method name is always better than comments.
The best code does not need comments. Although it is almost impossible to achieve.

helix-core/src/main/java/org/apache/helix/model/ParticipantHistory.java
jiajunwang: "What would happen then?" Then we add other methods to access this information explicitly. The string return which requires additional parse is a bad idea in general.
1. What if we change the string format? These methods return string will be non-backward-compatible. And we don't want to restrict ourselves unnecessarily, right?
2. Like you add the getOnlineTimestamp now, we can add other methods that parse the HISTORY string and return the required information.
3. The necessity of testing does not require us to make it public. We can make the test in the same package. Or read the field directly from ZNRecord. There are many ways to achieve it and public is the most impactful and costly (maintenance cost) way. I think it is not good option.

helix-core/src/test/java/org/apache/helix/integration/controller/TestOfflineNodeTimeoutDuringMaintenanceMode.java
NealSun96: Updated.

NealSun96:
This PR is ready to be merged, approved by @dasahcc @jiajunwang 
Final commit message:
## Feature: Offline Node Timeout During Maintenance Mode  ##
We would like to avoid the scenario where nodes that have been offline for a while come back to life during maintenance mode. Solution: if a node has been offline for a certain period of time, the maintenance mode rebalancer will not assign partitions to it.

Issue #1414, opened 2020-09-25
Add a system config to disable bucketized feature.


xyuanlu:
**Is your feature request related to a problem? Please describe.**
As mentioned in issue #1291 and  #401, doing `getChildren()` in `subscribeForChanges()` to determine if a node is bucketized or not takes a lot time. It would be useful to have a system config to disable the feature for performance boost.

xyuanlu:
Mark this as won't fix. Problem solved by PR https://github.com/apache/helix/pull/1645

Issue #1291, opened 2020-08-18
Helix controller leadership switch is slow for large cluster


xyuanlu:
### Describe the bug
The controller switch time for large scale cluster is slow. It takes around ~60s for a standby controller to obtain leadership of a cluster with ~50 participants and ~700 resources. The service's availability is impacted under this period of time. 

### To Reproduce
This issue can be constantly reproduced in large cluster set up.

### Expected behavior
The ultimate goal would be reducing the leadership handoff time to ~10s.

### Additional context
The performance improvement is a long process. We should do small improvements one at a time and achieve the final goal step by step.

Issue #401, opened 2019-08-13
Replace org.codehaus.jackson with FasterXML/jackson


brunnsbe:
This issue is the same as https://issues.apache.org/jira/browse/HELIX-747 (it's a bit confusing which issue list should be used for this project as the documentation on https://cwiki.apache.org/confluence/display/HELIX/Contributor+Workflow doesn't seem to be up to date):

The current json lib Helix uses is out of date. We should consider replacing it with a well-maintained lib.

FasterXML/jackson is compatible with the current lib we used. So it could be a good candidate.

The old version of Jackson that Helix currently uses contains multiple CVEs:
CVE-2017-17485, CVE-2017-7525, CVE-2017-15095

brunnsbe:
I forked this repository and made a commit that replaces the old Jackson with the newer FasterXML/Jackson:
https://github.com/brunnsbe/helix/commit/ca36d5e51c00e0278fa8526581a9ed9081e619c1
I however have problems running all the tests throws an OutOfMemoryException:
```
...
[ERROR]   TestZkHelixPropertyStore.testSet » OutOfMemory unable to create new native thr...
[ERROR]   TestZkHelixPropertyStore.testSetInvalidPath » OutOfMemory unable to create new...
[ERROR]   TestZkHelixPropertyStore.testZkClientMonitor » OutOfMemory unable to create ne...
[ERROR]   TestZkHelixPropertyStore.testZkTriggeredCallback » OutOfMemory unable to creat...
[ERROR]   TestZkManagerWithAutoFallbackStore.testBasic » OutOfMemory unable to create ne...
[INFO]
[ERROR] Tests run: 866, Failures: 177, Errors: 0, Skipped: 103
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Apache Helix 0.9.1-SNAPSHOT:
[INFO]
[INFO] Apache Helix ....................................... SUCCESS [  2.061 s]
[INFO] Apache Helix :: Core ............................... FAILURE [  01:00 h]
[INFO] Apache Helix :: Admin Webapp ....................... SKIPPED
[INFO] Apache Helix :: Restful Interface .................. SKIPPED
[INFO] Apache Helix :: HelixAgent ......................... SKIPPED
[INFO] Apache Helix :: Recipes ............................ SKIPPED
[INFO] Apache Helix :: Recipes :: Rabbitmq Consumer Group . SKIPPED
[...]

PR #1415, opened 2020-09-26
Add a system config to disable bucketized feature

xyuanlu:
### Issues

 #1414
Add a system config to disable bucketized feature.

### Description

As mentioned in issue #1291 and #401, doing getChildren() in subscribeForChanges() to determine if a node is bucketized or not takes a lot time. It would be useful to have a system config to disable the feature for performance boost.

### Tests

NA

```
[WARNING] Tests run: 1210, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 4,076.809 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[WARNING] Tests run: 1210, Failures: 0, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:08 h
[INFO] Finished at: 2020-09-25T16:28:06-07:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -75,6 +75,7 @@ import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.monitoring.mbeans.HelixCallbackMonitor;
 import org.apache.helix.zookeeper.api.client.ChildrenSubscribeResult;
 import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
+import org.apache.helix.zookeeper.constant.ZkSystemPropertyKeys;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.zkclient.IZkChildListener;
 import org.apache.helix.zookeeper.zkclient.IZkDataListener;
@@ -141,6 +142,9 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
 
   // indicated whether this CallbackHandler is ready to serve event callback from ZkClient.
   private boolean _ready = false;
+
+  private static final boolean BUCKETIZE_ZNRECORD_ENABLED = Boolean
+      .parseBoolean(System.getProperty(ZkSystemPropertyKeys.ZK_BUCKETIZE_ZNRECORD_ENABLED, "true"));
 
   static {
     SubscribeChangeEventProcessor = new DedupEventProcessor<CallbackHandler, SubscribeChangeEvent>(
@@ -632,31 +636,34 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
             case CUSTOMIZED_VIEW:
             case TARGET_EXTERNAL_VIEW: {
               // check if bucketized
-              BaseDataAccessor<ZNRecord> baseAccessor = new ZkBaseDataAccessor<>(_zkClient);
-              List<ZNRecord> records = baseAccessor.getChildren(path, null, 0, 0, 0);
-              for (ZNRecord record : records) {
-                HelixProperty property = new HelixProperty(record);
-                String childPath = path + "/" + record.getId();
+              if (BUCKETIZE_ZNRECORD_ENABLED) {
+                BaseDataAccessor<ZNRecord> baseAccessor = new ZkBaseDataAccessor<>(_zkClient);
+                List<ZNRecord> records = baseAccessor.getChildren(path, null, 0, 0, 0);
+                for (ZNRecord record : records) {
+                  HelixProperty property = new HelixProperty(record);
+                  String childPath = path + "/" + record.getId();
 
-                int bucketSize = property.getBucketSize();
-                if (bucketSize > 0) {
-                  // subscribe both data-change and child-change on bucketized parent node
-                  // data-change gives a delete-callback which is used to remove watch
-                  List<String> bucketizedChildNames = subscribeChildChange(childPath, callbackType);
-                  subscribeDataChange(childPath, callbackType);
+                  int bucketSize = property.getBucketSize();
+                  if (bucketSize > 0) {
+                    // subscribe both data-change and child-change on bucketized parent node
+                    // data-change gives a delete-callback which is used to remove watch
+                    subscribeChildChange(childPath, callbackType);
+                    subscribeDataChange(childPath, callbackType);
 
-                  // subscribe data-change on bucketized child
-                  if (bucketizedChildNames != null) {
-                    for (String bucketizedChildName : bucketizedChildNames) {
-                      String bucketizedChildPath = childPath + "/" + bucketizedChildName;
-                      subscribeDataChange(bucketizedChildPath, callbackType);
+                    // subscribe data-change on bucketized child
+                    List<String> bucketizedChildNames = _zkClient.getChildren(childPath);
+                    if (bucketizedChildNames != null) {
+                      for (String bucketizedChildName : bucketizedChildNames) {
+                        String bucketizedChildPath = childPath + "/" + bucketizedChildName;
+                        subscribeDataChange(bucketizedChildPath, callbackType);
+                      }
                     }
+                  } else {
+                    subscribeDataChange(childPath, callbackType);
                   }
-                } else {
-                  subscribeDataChange(childPath, callbackType);
                 }
-              }
-              break;
+                break;
+              } // else case uses default branch.
             }
             default: {
               if (children != null) {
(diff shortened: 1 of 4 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/constant/ZkSystemPropertyKeys.java
```diff
@@ -36,6 +36,16 @@
...
 
+  /**
+   * Setting this property to true enables ZNode bucketization, an operation to divide a ZNRecord
+   * into specified buckets. This property applies to
+   * {@link org.apache.helix.zookeeper.datamodel.ZNRecordBucketizer}.
+   * <p>
+   * The default value is "true" (enabled).
+   */
+  public static final String ZK_BUCKETIZE_ZNRECORD_ENABLED =
+      "zk.bucketize.znrecord.enabled";
```
pkuwm, line 47: How about `zk.znrecord.bucketize.enabled` or `zk.znrecord.bucketize.enabled`(which is consistent with the class `ZNRecordBucketizer`)?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixDataAccessor.java
```diff
@@ -599,4 +610,12 @@ private int constructOptions(PropertyType type) {
       List<DataUpdater<ZNRecord>> updaters, int options) {
     return _baseDataAccessor.updateChildren(paths, updaters, options);
   }
+
+  private void validateBucketizedEnabled(String path, boolean isRead) {
+    if (!BUCKETIZE_ZNRECORD_ENABLED) {
+      throw new HelixMetaDataAccessException(
+          "Can't " + (isRead ? "read" : "write") + " bucktized ZNode " + path
+              + " because Bucktize feature is not enabled. Please check ZK system property 'zk.bucketize.znrecord.enabled' .");
```
pkuwm, line 618: Use `ZkSystemPropertyKeys.ZK_BUCKETIZE_ZNRECORD_ENABLED` instead of hard coding it?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixDataAccessor.java
```diff
@@ -427,8 +437,9 @@ public boolean removeProperty(PropertyKey key) {
 
             int bucketSize = property.getBucketSize();
             if (bucketSize > 0) {
-              // TODO: fix this if record.id != pathName
               String childPath = parentPath + "/" + record.getId();
+              validateBucketizedEnabled(childPath, true /*isRead*/);
```
pkuwm, line 441: I guess we don't need a specific comment for a parameter like this in Java? It is kind of impacting the code readability, what do you think? If you like, I suggest you put it end of line or before line.
```
validateBucketizedEnabled(childPath, true); // true - is read
```

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -632,31 +636,33 @@ private void subscribeForChanges(NotificationContext.Type callbackType, String p
...
+                      }
                     }
                   }
-                } else {
-                  subscribeDataChange(childPath, callbackType);
                 }
+                break;
               }
-              break;
+              // go to default branch if bucketized feature is not enabled.
```
pkuwm, line 665: This is kind of tricky: readability is not very straightforward and it is easy to make a mistake if later someone adds another case after this, it'll break the logic.

How about this: we wrap the check into a private method and make it not easy to make a mistake?
```
shouldReadBucketizedChildren(type) {
  Set<> types; // CURRENT_STATE, ..., TARGET_EXTERNAL_VIEW
  return types.contains(type) && BUCKETIZE_ZNRECORD_ENABLED;
}

// Then for the block
if (shouldReadBucketizedChildren(_changeType) {
  // bucketized subscribe block
} else {
  // default block
}
```

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixDataAccessor.java
```diff
@@ -143,6 +147,7 @@ public boolean createMaintenance(MaintenanceSignal maintenanceSignal) {
     case EXTERNALVIEW:
       // check if bucketized
       if (value.getBucketSize() > 0) {
+        validateBucketizedEnabled(path, false /*isRead*/);
```
pkuwm, line 150: I see that there are so many places that we have to validate this. It seems it is only using baseDataAccessor to read/write, can you help me understand why we don't put this check inside baseDataAccessor?

And we don't check `Map<String, ZNRecord> map = bucketizer.bucketize(lastCurState.getRecord());
` in `ParticipantManager`?

pkuwm (commented):
Thanks for the work, @xyuanlu

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixDataAccessor.java
xyuanlu: Are you suggesting we read the config every time? We could if the config need to be dynamically changed.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixDataAccessor.java
xyuanlu: I personally feel that when reading the code without compiler, having comment like this for boolean value is helpful.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixDataAccessor.java
xyuanlu: Please correct me if I am wrong. 
I think we could use the existing bucketized node handling branch if added in ZKHelixDataAccessor for minion code change. Checking in baceAccessor might require more code change and branching. 
Also we need to touch multiple places -- ZkBaseDataAccessor.get/setData ZkBaseDataAccessor.get/setChildren() and ZkBaseDataAccessor.update().

zookeeper-api/src/main/java/org/apache/helix/zookeeper/constant/ZkSystemPropertyKeys.java
xyuanlu: TFTR. Good suggestion. Will update.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixDataAccessor.java
pkuwm: Not reading the config every time. We use the config key to replace the hard coded string. The benefit is, it'll be easier for us to find the reference that uses the key. And if we refactor the name, we won't miss this place.
And actually the compiler usually will replace the constant variable "ZkSystemPropertyKeys.ZK_BUCKETIZE_ZNRECORD_ENABLED" with the string "zk.bucketize.znrecord.enabled".

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixDataAccessor.java
```diff
@@ -600,4 +611,13 @@ private int constructOptions(PropertyType type) {
       List<DataUpdater<ZNRecord>> updaters, int options) {
     return _baseDataAccessor.updateChildren(paths, updaters, options);
   }
+
+  private void validateBucketizedEnabled(String path, boolean isRead) {
+    if (!ZNRECORD_BUCKETIZE_ENABLED) {
+      throw new HelixMetaDataAccessException(
+          "Can't " + (isRead ? "read" : "write") + " bucktized ZNode " + path
+              + " because Bucktize feature is not enabled. Please check ZK system property: "
```
NealSun96, line 619: *Bucketized (2 occasions here)

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixDataAccessor.java
NealSun96: I agree with @pkuwm that this may not be scalable and adding it to baseDataAccessor. Future developers might not be aware of this function and nothing enforces them to add it, then there could be unprotected bucketized read/write.

NealSun96 (changes requested):

PR #1416, opened 2020-09-26
add REST API for cluster topology

### LINKED ISSUE #1405: Add REST API for cluster topology
Provide a few REST API for user to retrieve cluster topology information.

Add APIs in ClusterAccessor

**Additional context**

zhangmeng916:
### Issues

Fixed #1405 

### Description
This PR provide a few REST API for user to retrieve cluster topology information. The APIs are added in ClusterAccessor

### Tests

helix-rest:
[INFO] Tests run: 170, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 139.365 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 170, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  02:23 min
[INFO] Finished at: 2020-10-06T21:37:18-07:00
[INFO] --------------------------------------------------------------------
### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
@@ -453,9 +453,6 @@ public class ClusterAccessor extends AbstractHelixResource {
       topologyMap = admin.getClusterTopology(clusterId).getTopologyMap();
     } catch (HelixException ex) {
       return badRequest(ex.getMessage());
-    } catch (Exception ex) {
-      LOG.error("Failed to get cluster topology map from cluster {}.", clusterId, ex);
-      return serverError(ex);
     }
     return JSONRepresentation(topologyMap);
   }
@@ -471,9 +468,6 @@ public class ClusterAccessor extends AbstractHelixResource {
       faultZoneMap = admin.getClusterTopology(clusterId).getFaultZoneMap();
     } catch (HelixException ex) {
       return badRequest(ex.getMessage());
-    } catch (Exception ex) {
-      LOG.error("Failed to get cluster fault zone map from cluster {}", clusterId, ex);
-      return serverError(ex);
     }
     return JSONRepresentation(faultZoneMap);
   }
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
@@ -447,9 +447,6 @@ public class ClusterAccessor extends AbstractHelixResource {
   @GET
   @Path("{clusterId}/topologymap")
   public Response getClusterTopologyMap(@PathParam("clusterId") String clusterId) {
-    if (!doesClusterExist(clusterId)) {
-      return notFound(String.format("Cluster %s does not exist", clusterId));
-    }
     HelixAdmin admin = getHelixAdmin();
     Map<String, List<String>> topologyMap;
     try {
@@ -457,8 +454,7 @@ public class ClusterAccessor extends AbstractHelixResource {
     } catch (HelixException ex) {
       return badRequest(ex.getMessage());
     } catch (Exception ex) {
-      LOG.error("Failed to get cluster topology map from cluster {}. Exception: " + "{}.",
-          clusterId, ex);
+      LOG.error("Failed to get cluster topology map from cluster {}.", clusterId, ex);
       return serverError(ex);
     }
     return JSONRepresentation(topologyMap);
@@ -476,8 +472,7 @@ public class ClusterAccessor extends AbstractHelixResource {
     } catch (HelixException ex) {
       return badRequest(ex.getMessage());
     } catch (Exception ex) {
-      LOG.error("Failed to get cluster fault zone map from cluster {}. Exception: " + "{}.",
-          clusterId, ex);
+      LOG.error("Failed to get cluster fault zone map from cluster {}", clusterId, ex);
       return serverError(ex);
     }
     return JSONRepresentation(faultZoneMap);
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
@@ -442,6 +442,26 @@ public class ClusterAccessor extends AbstractHelixResource {
     return OK(objectMapper.writeValueAsString(clusterTopology));
   }
 
+  @ResponseMetered(name = HttpConstants.READ_REQUEST)
+  @Timed(name = HttpConstants.READ_REQUEST)
+  @GET
+  @Path("{clusterId}/topologymap")
+  public Response getClusterTopologyMap(@PathParam("clusterId") String clusterId) {
+    HelixAdmin admin = getHelixAdmin();
+    Map<String, List<String>> topologyMap = admin.getClusterTopology(clusterId).getTopologyMap();
+    return JSONRepresentation(topologyMap);
+  }
+
+  @ResponseMetered(name = HttpConstants.READ_REQUEST)
+  @Timed(name = HttpConstants.READ_REQUEST)
+  @GET
+  @Path("{clusterId}/faultzonemap")
+  public Response getClusterFaultZoneMap(@PathParam("clusterId") String clusterId) {
+    HelixAdmin admin = getHelixAdmin();
+    Map<String, List<String>> topologyMap = admin.getClusterTopology(clusterId).getTopologyMap();
+    return JSONRepresentation(topologyMap);
+  }
+
   @ResponseMetered(name = HttpConstants.WRITE_REQUEST)
   @Timed(name = HttpConstants.WRITE_REQUEST)
   @POST
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
@@ -457,8 +457,8 @@ public class ClusterAccessor extends AbstractHelixResource {
     } catch (HelixException ex) {
       return badRequest(ex.getMessage());
     } catch (Exception ex) {
-      LOG.error("Failed to get cluster topology map fro cluster {}. Exception: " + "{}.", clusterId,
-          ex);
+      LOG.error("Failed to get cluster topology map from cluster {}. Exception: " + "{}.",
+          clusterId, ex);
       return serverError(ex);
     }
     return JSONRepresentation(topologyMap);
@@ -469,9 +469,6 @@ public class ClusterAccessor extends AbstractHelixResource {
   @GET
   @Path("{clusterId}/faultzonemap")
   public Response getClusterFaultZoneMap(@PathParam("clusterId") String clusterId) {
-    if (!doesClusterExist(clusterId)) {
-      return notFound(String.format("Cluster %s does not exist", clusterId));
-    }
     HelixAdmin admin = getHelixAdmin();
     Map<String, List<String>> faultZoneMap;
     try {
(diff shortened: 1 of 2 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -442,6 +442,26 @@ public Response getClusterTopology(@PathParam("clusterId") String clusterId) thr
     return OK(objectMapper.writeValueAsString(clusterTopology));
   }
 
+  @ResponseMetered(name = HttpConstants.READ_REQUEST)
+  @Timed(name = HttpConstants.READ_REQUEST)
+  @GET
+  @Path("{clusterId}/topologymap")
+  public Response getClusterTopologyMap(@PathParam("clusterId") String clusterId) {
+    HelixAdmin admin = getHelixAdmin();
+    Map<String, List<String>> topologyMap = admin.getClusterTopology(clusterId).getTopologyMap();
```
pkuwm, line 451: From the internal logic, if the cluster doesn't exist, an exception will be thrown - then it causes a server error 500, which is not appropriate, because it is a client error. Can we check if the cluster exists or not - if not, returns 404 NOT_FOUND and appropriate message
```
{
 status_code: 404,
 message: cluster <clusterId> is not found
}
```

helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
```diff
@@ -125,6 +125,69 @@ public void testGetClusterTopology() {
...
+        OBJECT_MAPPER.readValue(topologyMapDef, new TypeReference<HashMap<String, Object>>() {
+        });
+    Assert.assertEquals(topologyMap.size(), 2);
+    Assert.assertTrue(topologyMap.get("/helixZoneId:zone0") instanceof List);
+    List<String> instances = (List<String>) topologyMap.get("/helixZoneId:zone0");
+    Assert.assertEquals(instances.size(), 5);
+
+    Assert.assertTrue(topologyMap.get("/helixZoneId:zone1") instanceof List);
+    instances = (List<String>) topologyMap.get("/helixZoneId:zone1");
+    Assert.assertEquals(instances.size(), 5);
```
pkuwm, line 168: Do you think it'd be better to also verify the content of `instances`?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -442,6 +442,29 @@ public Response getClusterTopology(@PathParam("clusterId") String clusterId) thr
...
+    return JSONRepresentation(topologyMap);
+  }
+
+  @ResponseMetered(name = HttpConstants.READ_REQUEST)
+  @Timed(name = HttpConstants.READ_REQUEST)
+  @GET
+  @Path("{clusterId}/faultzonemap")
+  public Response getClusterFaultZoneMap(@PathParam("clusterId") String clusterId) {
+    HelixAdmin admin = getHelixAdmin();
+    Map<String, List<String>> topologyMap = admin.getClusterTopology(clusterId).getTopologyMap();
```
pkuwm, line 464: Typo?  `getTopologyMap`  -> `getFaultZoneMap`.

And `getFaultZoneMap` throws an `IllegalArgumentException` when "The fault zone in cluster config is not defined" - it seems to be client error. Would you consider catching the exception and then returning a client error, instead of a server error. I think it'd be better to differentiate a client error and a server error.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -442,6 +442,29 @@ public Response getClusterTopology(@PathParam("clusterId") String clusterId) thr
...
+    }
+    HelixAdmin admin = getHelixAdmin();
+    Map<String, List<String>> topologyMap = admin.getClusterTopology(clusterId).getTopologyMap();
+    return JSONRepresentation(topologyMap);
+  }
+
+  @ResponseMetered(name = HttpConstants.READ_REQUEST)
+  @Timed(name = HttpConstants.READ_REQUEST)
+  @GET
+  @Path("{clusterId}/faultzonemap")
```
pkuwm, line 461: From the java code logic, it seems these 2 REST APIs actually return a topology which is represented by a map.
If so, I think the REST endpoint `{clusterId}/topology` is good and clear enough? We don't need the `map` to specify the data format.

And for fault zone, how about adding a query parameter `domain_type` based on the `/topology` endpoint: `/topology?domain_type=fault_zone`?  (correct me if I am not clear about it, as I just understand it by reading the code) Then it looks cleaner and is easier to read.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -442,6 +442,29 @@ public Response getClusterTopology(@PathParam("clusterId") String clusterId) thr
     return OK(objectMapper.writeValueAsString(clusterTopology));
   }
 
+  @ResponseMetered(name = HttpConstants.READ_REQUEST)
+  @Timed(name = HttpConstants.READ_REQUEST)
+  @GET
+  @Path("{clusterId}/topologymap")
+  public Response getClusterTopologyMap(@PathParam("clusterId") String clusterId) {
+    if (!doesClusterExist(clusterId)) {
```
pkuwm, line 450: Since this will add one more zk read, and `admin.getClusterTopology(clusterId)` actually reads the cluster, I am not sure how heavy this call is. I assume it is mostly used during deployment? 

I wonder if we can use the result in `admin.getClusterTopology`: if the cluster doesn't exist, an exception(`HelixException/NoClusterExistException`) is thrown. Rest code here catches it and returns 404. Then it saves one zk read.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
zhangmeng916: There is already a legacy `topology` REST API in this class, so we have to use a different one. Any other idea than `topologymap`?
Regarding the suggestion of fault zone query, what you said is actually my previous implementation. However, we then decided to only expose the most common use case, i.e., return the topology under fault zone as an API to our users. So this fault zone API is not a generic one that accepts any level of domain. Instead, it's a fixed one only for "fault zone type" defined in topology.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
zhangmeng916: I tried a bit. So here're the cases.
There are a few exceptions. 
cluster not set up.
fault zone is not set.
instance configs are not set.
topology is not set or wrongly set.
How are we going to categorize them? I think they're all client side exceptions. What response we would like to return?

jiajunwang:
Please fill the required content to the description.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -442,6 +442,50 @@ public Response getClusterTopology(@PathParam("clusterId") String clusterId) thr
...
+      return notFound(String.format("Cluster %s does not exist", clusterId));
+    }
+    HelixAdmin admin = getHelixAdmin();
+    Map<String, List<String>> faultZoneMap;
+    try {
+      faultZoneMap = admin.getClusterTopology(clusterId).getFaultZoneMap();
+    } catch (HelixException ex) {
+      return badRequest(ex.getMessage());
+    } catch (Exception ex) {
+      LOG.error("Failed to get cluster fault zone map fro cluster {}. Exception: " + "{}.",
```
pkuwm, line 482: Typo: **fro** -> **from**

`LOG.error("Failed to get cluster fault zone map fro cluster {}.", clusterId, ex);` to print out the stack trace?

Same in `getClusterTopologyMap `.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
pkuwm: I realized that this call is not a lightweight call. It has many calls to zk to check all required below paths exist. I think this is pretty heavy. And even the cluster exists at the usual case, this check makes the rest api call pretty slow to respond, which I don't think is expected, right? And actually `isClusterSetup` is called in `getClusterTopology` -> `getClusterConfig()` -> `getConfigZnRecord` -- multiple duplicate calls to zk just to check `isClusterSetup`.  So the following `HelixException` catch is good enough. No need to do the pre-check `if (!doesClusterExist(clusterId)) {`

```
    requiredPaths.add(PropertyPathBuilder.idealState(clusterName));
    requiredPaths.add(PropertyPathBuilder.clusterConfig(clusterName));
    requiredPaths.add(PropertyPathBuilder.instanceConfig(clusterName));
    requiredPaths.add(PropertyPathBuilder.resourceConfig(clusterName));
    requiredPaths.add(PropertyPathBuilder.propertyStore(clusterName));
    requiredPaths.add(PropertyPathBuilder.liveInstance(clusterName));
    requiredPaths.add(PropertyPathBuilder.instance(clusterName));
    requiredPaths.add(PropertyPathBuilder.externalView(clusterName));
    requiredPaths.add(PropertyPathBuilder.controller(clusterName));
    requiredPaths.add(PropertyPathBuilder.stateModelDef(clusterName));
    requiredPaths.add(PropertyPathBuilder.controllerMessage(clusterName));
    requiredPaths.add(PropertyPathBuilder.controllerError(clusterName));
    requiredPaths.add(PropertyPathBuilder.controllerStatusUpdate(clusterName));
    requiredPaths.add(PropertyPathBuilder.controllerHistory(clusterName));
```

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
zhangmeng916: Do you mean ex.getStackTrace()? I took a look, seems not a convention in our rest. Any specific reason?

zhangmeng916: updated.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -442,6 +442,42 @@ public Response getClusterTopology(@PathParam("clusterId") String clusterId) thr
...
+  public Response getClusterTopologyMap(@PathParam("clusterId") String clusterId) {
+    HelixAdmin admin = getHelixAdmin();
+    Map<String, List<String>> topologyMap;
+    try {
+      topologyMap = admin.getClusterTopology(clusterId).getTopologyMap();
+    } catch (HelixException ex) {
+      return badRequest(ex.getMessage());
+    } catch (Exception ex) {
+      LOG.error("Failed to get cluster topology map from cluster {}.", clusterId, ex);
+      return serverError(ex);
```
pkuwm, line 458: This `serverError` actually is not needed, because jersey framework will handle it. If there is any exception that is not handled, it'll return server error 500.

helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
```diff
@@ -125,6 +126,93 @@ public void testGetClusterTopology() {
   }
 
   @Test(dependsOnMethods = "testGetClusterTopology")
+  public void testGetClusterTopologyAndFaultZoneMap() throws IOException {
```
pkuwm, line 129: It'd be more complete if the bad response is also tested. We don't always have good response. Eg. the cluster is not setup -> Error 400 bad request.

pkuwm (approved):
Overall looks good to me. Left 2 comments that you should be able to quickly address

zhangmeng916:
This PR is ready to merge. Approved by @pkuwm 
Final commit message:
Add REST API for cluster topology

PR #1422, opened 2020-09-29
Fix currentState not being removed when pending message existed

### LINKED ISSUE #1421: Task currentState might not get dropped when pending message existed
### Describe the bug
Once the last task of a job finished, the controller marks the job as completed/failed and sends the COMPLETED, TIMED_OUT, etc. message for the last task. However, if pending messages existed, the job will be marked as finished, but the tasks do not receive the DROPPED message. In this case, the job will be removed from the DAG and the currentState will not be removed. 

### To Reproduce
Have a queue with several jobs. Each job should have multiple short-lived tasks. After you start the queue, you would notice that the task current states remain in the ZK.

### Expected behavior
The controller should not mark the job as COMPLETED or FAILED if the pending message still existed for the tasks

### Additional context
This issue appears as we are not relying on IdealState anymore, Previously, the legacy pipeline would DROP the tasks in the next pipelines

alirezazamani:
### Issues

Fixes #1421 

### Description

With this PR, we make sure that the controller sends DROPPED messages for the jobs that have just finished and their tasks have pending messages.

### Tests

TestDropCurrentStateRunningTask.testJobCurrentStateDroppedAfterCompletion

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestDisableCustomCodeRunner.test:233 expected:<false> but was:<true>
[INFO] 
[ERROR] Tests run: 1210, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:11 h
[INFO] Finished at: 2020-09-28T17:49:41-07:00
```
The failed test individually:
mvn test -Dtest="TestDisableCustomCodeRunner"
```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 8.996 s - in org.apache.helix.integration.TestDisableCustomCodeRunner
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  14.768 s
[INFO] Finished at: 2020-09-28T18:17:04-07:00
```

### Commits

### Code Quality


dasahcc (approved):

helix-core/src/test/java/org/apache/helix/integration/task/TestJobCurrentStateDroppedAfterCompletion.java
```diff
@@ -0,0 +1,102 @@
...
+    super.beforeClass();
+  }
+
+  @AfterClass
+  public void afterClass() throws Exception {
+    super.afterClass();
+  }
+
+  @Test
+  public void testJobCurrentStateDroppedAfterCompletion() throws Exception {
```
jiajunwang, line 54: nit, is it possible to add this test to some existing test class? Recreating the whole test environment is expensive. And I feel we are adding more test classes than necessary. It will soon become unmanageable.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -122,30 +122,37 @@ public void updatePreviousAssignedTasksStatus(
...
-          LOG.warn(
-              "Instance {} does not match the assigned participant for pId {} in the job context. Skipping task scheduling.",
-              instance, pId);
-          continue;
-        }
-
         // Check for pending state transitions on this (partition, instance). If there is a pending
         // state transition, we prioritize this pending state transition and set the assignment from
         // this pending state transition, essentially "waiting" until this pending message clears
+        // If there is a pending message, we should not continue to update the context because from
```
jiajunwang, line 131: This means even the pending message target state matches the current state, we won't proceed? I know it is safer. But since it rolls back this enhancement, it might increase latency. Or does the removed check that the target state matches the current state will introduce any bug?

helix-core/src/test/java/org/apache/helix/integration/task/TestJobCurrentStateDroppedAfterCompletion.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: This is more toward the functionality of our scheduling. Without this change, there is a whole in our scheduling decision. Before IS removal change, we were relying on IS to remove the last CS from the participants. However, removing IS revealed this issue. This PR basically handles these cases and helps CS to be cleaned up after the job is being completed. Also, we think about it logically, it makes sense to consider ST not complete when a pending message has not been deleted from the ZK.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -122,30 +122,37 @@ public void updatePreviousAssignedTasksStatus(
...
+        // controller prospective, state transition has not been completed yet if pending message
+        // still existed.
+        // If context gets updated here, controller might remove the job from RunTimeJobDAG which
+        // can cause the task's CurrentState not being removed when there is a pending message for
+        // that task.
         Message pendingMessage =
             currStateOutput.getPendingMessage(jobResource, new Partition(pName), instance);
-        if (pendingMessage != null && !pendingMessage.getToState().equals(currState.name())) {
```
NealSun96, line 140: It looks like it's just about the order of jobContext updating vs processing pending messages, if I understood you correctly. If that's the case, is there a reason why we removed the second section of the if statement here, as @jiajunwang mentioned?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -122,30 +122,37 @@ public void updatePreviousAssignedTasksStatus(
...
-          // If there is a pending message whose destination state is different from the current
-          // state, just make the same assignment as the pending message. This is essentially
-          // "waiting" until this state transition is complete
+        if (pendingMessage != null) {
           processTaskWithPendingMessage(pId, pName, instance, pendingMessage, jobState, currState,
               paMap, assignedPartitions);
           continue;
         }
```
NealSun96, line 144: General comment: how did the legacy pipeline help masking this problem before? Or, was it not related to the legacy pipeline, but related to how we delete CurrentStates? The legacy pipeline logic processes the inflight jobs again, but it shouldn't make a difference?

alirezazamani: It is related to legacy code. Previously, when a job is completed, failed, or timed out, IS would be deleted by the controller. In the next pipeline, since the Ideal state does not exist but the current state exists, we go to legacy code and drop the remaining current states.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: It is very complicated to explain:
1- The controller sends RUNNINg to COMPLETED message.
2- The participant marks the current state to be COMPLETED but the pending message has not been removed yet.
3- In the next pipeline (pending message toState is COMPLETED, currentState is COMPLETED), this if statement would not be satisfied. Hence, we decide COMPLETED to DROPPED without considering pending messages.
4- The controller marks the job as completed and removes it from the DAG (so we will never consider this job again unless the controller switch to etc).
5- Since the controller sees the pending message, in the message generation phase we do not consider a new decision which is COMPLETED to DROPPED and then CS will never be DROPPED.
The conclusion is that the controller should not consider state transition as done state transition until the pending message is gone from ZK.

NealSun96: Thanks for explaining. So it looks to be a problem with 2 parts:
1. Job context is updated while there are pending messages. This can cause jobs to be removed since it's marked as completed. If DROPPED message isn't correctly sent here, then the DROPPED message will never be sent because the job is deleted;
2. Current state gets updated but pending messages aren't removed (async problem). In that case, the old if statement would not catch the problem and will proceed as if the message doesn't exist. 
I suppose 2 is not a problem except when the job is about to be deleted; if the job isn't deleted, the pipeline can simply retry next time. 

For 1, I don't understand why the new code in `TaskSchedulingStage` doesn't catch it. Specifically:
```
    // Jobs that exist in current states but are missing corresponding JobConfigs or WorkflowConfigs
    // or WorkflowContexts need to be cleaned up. Note that restOfResources can only be jobs,
    // because workflow resources are created based on Configs only - workflows don't have
    // CurrentStates
    for (String resourceName : restOfResources.keySet()) {
      _workflowDispatcher.processJobForDrop(resourceName, currentStateOutput, output);
    }
```

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
NealSun96: Feel free to close this. My questions are combined at the previous comment.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: For 1, it is not related to the deletion of the job. The issue is one you mark the last task of the job in the jobContext as completed, then you remove this job from RunTimeJobDag in jobDispatcher. Then next pipeline you will not even see this job in your RuntimeJobDAG. Now there is no way to drop the current state until JobConfig existed. After JobConfig is removed, then the code you pasted should take care of it.

alirezazamani:

PR #1429, opened 2020-10-02
Add Zookeeper multi ops support in the Helix ZkClient.

### LINKED ISSUE #1428: MultiOp support in the Helix ZkClient
The native multi ops method does not fit Helix ZkClient so it is hard for us to use correctly. It relies on the caller to do serialize, validation, and monitor.

Introduce a new multiOp method in the Helix ZkClient which contains all the default Helix ZkClient enhanced features.

**Additional context**

jiajunwang:
### Issues

Resolves #1428 

### Description

The Helix ZkClient multi ops interfaces can be used in the same way as the native Zookeeper multi ops method. But the users are required to construct Helix MultiOp Objects for the additional Helix support such as monitoring, validation, and auto compress, etc.

### Tests

TestRawZkClient.testMultiOps 

[INFO] Tests run: 47, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/MultiOp.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/MultiOp.java
@@ -1,0 +1,144 @@
+package org.apache.helix.zookeeper.api.client;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.List;
+
+import org.apache.helix.zookeeper.zkclient.serialize.PathBasedZkSerializer;
+import org.apache.zookeeper.CreateMode;
+import org.apache.zookeeper.Op;
+import org.apache.zookeeper.ZooDefs;
+import org.apache.zookeeper.data.ACL;
+
+
+/**
+ * A wrapper class of org.apache.zookeeper.Op to support Helix interfaces and monitoring
+ * functionality.
+ */
+public abstract class MultiOp {
+  private final String _path;
+
+  private MultiOp(String path) {
+    _path = path;
+  }
+
+  /**
+   * @return the target path of the operation.
+   */
+  public String getPath() {
+    return _path;
+  }
+
+  /**
+   * Build a new org.apache.zookeeper.Op object for Zookeeper.multi call.
+   * @param serializer
+   * @return org.apache.zookeeper.Op object corresponding to the MultiOp object.
+   */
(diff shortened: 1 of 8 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
```diff
@@ -261,8 +261,11 @@ void asyncSetData(final String path, Object datat, final int version,
 
   long getCreationTime(String path);
 
+  @Deprecated
   List<OpResult> multi(final Iterable<Op> ops);
```
narendly, line 265: Is deprecating this is a necessity here? zookeeper-api is supposed to be a general-purpose client library as well (not just tailored for Helix usage), so I don't see the need for it quite yet. My opinion is that we should continue to maintain the interface defined in the original ZkClient.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/MultiOp.java
```diff
@@ -0,0 +1,144 @@
...
+  public String getPath() {
+    return _path;
+  }
+
+  /**
+   * Build a new org.apache.zookeeper.Op object for Zookeeper.multi call.
+   * @param serializer
+   * @return org.apache.zookeeper.Op object corresponding to the MultiOp object.
+   */
+  public abstract Op buildZkOp(PathBasedZkSerializer serializer);
```
narendly, line 54: Will ZkSerializer be supported?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/MultiOp.java
```diff
@@ -0,0 +1,144 @@
...
+import org.apache.helix.zookeeper.zkclient.serialize.PathBasedZkSerializer;
+import org.apache.zookeeper.CreateMode;
+import org.apache.zookeeper.Op;
+import org.apache.zookeeper.ZooDefs;
+import org.apache.zookeeper.data.ACL;
+
+
+/**
+ * A wrapper class of org.apache.zookeeper.Op to support Helix interfaces and monitoring
+ * functionality.
```
narendly, line 33: Where do we have the "Helix interfaces and monitoring functionality"?

jiajunwang:
Thanks for reviewing, @narendly.
I'm still doing more prototype work. Until I'm convinced that it will work end to end, I will temporarily suspend this PR's work. Let me close it for now to avoid confusion. I will re-open it later with some API design.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
jiajunwang: This operation does not fit the Helix ZkClient design since,
1. there is no retry until connecting support.
2. there is no metric support.
Keep supporting it, then we need to take it as a very special case. Currently, there is no large scale usage, so the metrics and caller logic looks fine. Otherwise, we will see issues here and there.
So my take is that we want to replace it with the new multiops API with full support.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/MultiOp.java
jiajunwang: I don't see any issues with the interfaces.
There is one thing I take as a TODO. If the requested paths are located on different realms, then we won't be able to do it. We may want to check and fail such a request.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/MultiOp.java
jiajunwang: Will be in my complete version. But that will still be a POC.

PR #1439, opened 2020-10-06
Implement addTask API

### LINKED ISSUE #1437: Add new tasks to the running (IN-PROGRESS) jobs
There has been a request to add the ability to dynamically add tasks to the running jobs.

The API can be added to the TaskDriver which allows users to add new tasks while the job is being run.

alirezazamani:
### Issues

Fixes #1437 

### Description

In this PR, addTask API has been implemented which adds a new task to running (IN-PROGRESS) jobs.

### Tests

TestAddTask

```
[INFO] Tests run: 1222, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,263.589 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1222, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:11 h
[INFO] Finished at: 2020-10-07T17:44:08-07:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
@@ -535,17 +535,21 @@ public class TaskDriver {
 
   /**
    * Add task to a running (IN-PROGRESS) job or a job which has not started yet. Timeout for this
-   * operation is default timeout which is 5 minutes. {@link TaskDriver#DEFAULT_TIMEOUT}
+   * operation is the default timeout which is 5 minutes. {@link TaskDriver#DEFAULT_TIMEOUT}
    * Note1: Task cannot be added if the job is in an illegal state. A job can accept
    * new task if the job is in-progress or it has not started yet.
    * Note2: The job can only be added to non-targeted jobs.
-   * Note3: The taskID for new task should be unique. If not, this API throws an exception.
-   * Note4: In case of timeout exception, it is user's responsibility to check whether the task has
-   * been successfully added or not.
+   * Note3: The taskID for the new task should be unique. If not, this API throws an exception.
+   * Note4: In case of timeout exception, it is the user's responsibility to check whether the task
+   * has been successfully added or not.
    * @param workflowName
    * @param jobName
    * @param taskConfig
-   * @throws Exception
+   * @throws Exception if there is an issue with the request or the operation. 1-
+   *           IllegalArgumentException will be thrown if the inputs are invalid. 2- HelixException
+   *           will be thrown if the job is not in the states to accept a new task or if there is
+   *           any issue in updating jobConfig. 3- TimeoutException will be thrown if the outcome of
+   *           the task addition is unknown and cannot be verified.
    */
   public void addTask(String workflowName, String jobName, TaskConfig taskConfig) throws Exception {
     addTask(workflowName, jobName, taskConfig, DEFAULT_TIMEOUT);
@@ -556,16 +560,20 @@ public class TaskDriver {
    * Note1: Task cannot be added if the job is in an illegal state. A job can accept
    * new task if the job is in-progress or it has not started yet.
    * Note2: The job can only be added to non-targeted jobs.
-   * Note3: The taskID for new task should be unique. If not, this API throws an exception.
-   * Note4: In case of timeout exception, it is user's responsibility to check whether the task has
-   * been successfully added or not.
+   * Note3: The taskID for the new task should be unique. If not, this API throws an exception.
+   * Note4: In case of timeout exception, it is the user's responsibility to check whether the task
+   * has been successfully added or not.
    * Note5: timeout is the time that this API checks whether the task has been successfully added or
    * not.
    * @param workflowName
    * @param jobName
    * @param taskConfig
    * @param timeoutMs
-   * @throws Exception
+   * @throws Exception if there is an issue with the request or the operation. 1-
+   *           IllegalArgumentException will be thrown if the inputs are invalid. 2- HelixException
+   *           will be thrown if the job is not in the states to accept a new task or if there is
+   *           any issue in updating jobConfig. 3- TimeoutException will be thrown if the outcome of
+   *           the task addition is unknown and cannot be verified.
    */
   public void addTask(String workflowName, String jobName, TaskConfig taskConfig, long timeoutMs)
       throws Exception {
--- a/helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
@@ -42,11 +42,9 @@ import org.apache.helix.manager.zk.ZKHelixAdmin;
 import org.apache.helix.manager.zk.ZKHelixDataAccessor;
 import org.apache.helix.manager.zk.ZkBaseDataAccessor;
 import org.apache.helix.model.ClusterConfig;
-import org.apache.helix.model.IdealState;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.LiveInstance;
 import org.apache.helix.model.ResourceConfig;
-import org.apache.helix.model.builder.CustomModeISBuilder;
 import org.apache.helix.store.HelixPropertyStore;
 import org.apache.helix.store.zk.ZkHelixPropertyStore;
 import org.apache.helix.util.HelixUtil;
@@ -524,6 +522,123 @@ public class TaskDriver {
       }
       throw new HelixException("Failed to enqueue job");
     }
+  }
+
+  /**
+   * Add task to a running (IN-PROGRESS) job with default timeout
+   * @param taskConfig
+   * @param workflowName
+   * @param jobName
+   * @throws Exception
+   */
+  public void addTask(TaskConfig taskConfig, String workflowName, String jobName) throws Exception {
+    addTask(taskConfig, workflowName, jobName, DEFAULT_TIMEOUT);
+  }
+
+  /**
+   * Add task to a running (IN-PROGRESS) job
+   * @param taskConfig
+   * @param workflowName
+   * @param jobName
+   * @param timeout
+   * @throws Exception
+   */
+  public void addTask(TaskConfig taskConfig, String workflowName, String jobName, long timeout)
+      throws Exception {
+    WorkflowConfig workflowConfig = TaskUtil.getWorkflowConfig(_accessor, workflowName);
+    long endTime = System.currentTimeMillis() + timeout;
+
+    if (workflowConfig == null) {
+      throw new HelixException("Workflow " + workflowName + " config does not exist!");
+    }
+
+    String nameSpaceJobName = TaskUtil.getNamespacedJobName(workflowName, jobName);
+    JobConfig jobConfig = TaskUtil.getJobConfig(_accessor, nameSpaceJobName);
+    if (jobConfig == null) {
+      throw new HelixException("Job " + nameSpaceJobName + " config does not exist!");
+    }
+
+    if (jobConfig.getTargetResource() != null) {
+      throw new HelixException(
+          "Job " + nameSpaceJobName + " is a targeted job. New task cannot be added to this job!");
+    }
+
+    WorkflowContext workflowContext = getWorkflowContext(workflowName);
+    if (workflowContext == null) {
+      throw new HelixException("Workflow " + workflowName + " context does not exist!");
+    }
+
+    if (taskConfig == null) {
+      throw new IllegalArgumentException("Task cannot be added because taskConfig is null!");
+    }
+
+    if (taskConfig.getId() == null) {
+      throw new HelixException(
+          "Task cannot be added because taskID is null!");
+    }
+
+    if (taskConfig.getCommand() == null && jobConfig.getCommand() == null) {
+      throw new HelixException(
+          "Task cannot be added because both of the job and task have null command!");
+    }
+
+    if (taskConfig.getCommand() != null && jobConfig.getCommand() != null) {
+      throw new HelixException(
+          "Task cannot be added because command existed for both of job and task!");
+    }
+
+    for (String taskEntry : jobConfig.getMapConfigs().keySet()) {
+      if (taskEntry.equals(taskConfig.getId())) {
+        throw new HelixException(
+            "Task cannot be added because another task with the same ID already exists!");
+      }
+    }
+
+    Set<TaskState> illegalJobStateForTaskAddition =
+        new HashSet<>(Arrays.asList(TaskState.TIMING_OUT, TaskState.TIMED_OUT, TaskState.FAILING,
+            TaskState.FAILED, TaskState.ABORTED, TaskState.COMPLETED, TaskState.STOPPED,
+            TaskState.STOPPING));
+
+    TaskState jobState = workflowContext.getJobState(nameSpaceJobName);
+
+    if (jobState == null) {
+      throw new HelixException(
+          "Task cannot be added because JobState " + nameSpaceJobName + " is null!");
+    }
+
+    if (illegalJobStateForTaskAddition.contains(jobState)) {
+      throw new HelixException("Job " + nameSpaceJobName
+          + " is in illegal state to accept new task. Job State is " + jobState);
+    }
+
+    DataUpdater<ZNRecord> updater = currentData -> {
+      if (currentData != null) {
+        currentData.setMapField(taskConfig.getId(), taskConfig.getConfigMap());
+      } else {
+        LOG.warn("JobConfig DataUpdater: Fails to update JobConfig. CurrentData is null.");
+      }
+      return currentData;
+    };
+
+    String path = _accessor.keyBuilder().resourceConfig(nameSpaceJobName).getPath();
+    boolean status = _accessor.getBaseDataAccessor().update(path, updater, AccessOption.PERSISTENT);
+    if (!status) {
+      LOG.error("Failed to add task to the job {}", nameSpaceJobName);
+      throw new HelixException("Failed to add task to the job");
+    }
+
+    String taskID = taskConfig.getId();
+    while (System.currentTimeMillis() <= endTime) {
+      JobContext jobContext = getJobContext(nameSpaceJobName);
+      for (Map.Entry<String, Integer> entry : jobContext.getTaskIdPartitionMap().entrySet()) {
+        if (entry.getKey().equals(taskID) && getWorkflowContext(workflowName)
+            .getJobState(nameSpaceJobName) == TaskState.IN_PROGRESS) {
+          return;
+        }
+      }
+      Thread.sleep(1000L);
+    }
+    throw new HelixException("An unexpected issue happened while task being added to the job!");
   }
 
   /**
--- a/helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
@@ -545,13 +545,13 @@ public class TaskDriver {
    * @param workflowName
    * @param jobName
    * @param taskConfig
-   * @throws Exception if there is an issue with the request or the operation. 1-
-   *           IllegalArgumentException will be thrown if the inputs are invalid. 2- HelixException
-   *           will be thrown if the job is not in the states to accept a new task or if there is
-   *           any issue in updating jobConfig. 3- TimeoutException will be thrown if the outcome of
-   *           the task addition is unknown and cannot be verified.
+   * @throws TimeoutException if the outcome of the task addition is unknown and cannot be verified
+   * @throws IllegalArgumentException if the inputs are invalid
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
     }
   }
 
+  /**
+   * Add task to a running (IN-PROGRESS) job with default timeout
```
lei-xia, line 528: Not necessarily a running job, people can add task to NOT_STARTED jobs too, right?

lei-xia: Add a task to a job, if a job is already completed, this operation has no effect?

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+
+  /**
+   * Add task to a running (IN-PROGRESS) job
+   * @param taskConfig
+   * @param workflowName
+   * @param jobName
+   * @param timeout
+   * @throws Exception
+   */
+  public void addTask(TaskConfig taskConfig, String workflowName, String jobName, long timeout)
```
lei-xia, line 546: I would suggest  to change the parameters order to this (String workflowName, String jobName, TaskConfig, timeout), just to follow the conventions in other methods in the TaskDriver.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+    }
+
+    if (jobConfig.getTargetResource() != null) {
+      throw new HelixException(
+          "Job " + nameSpaceJobName + " is a targeted job. New task cannot be added to this job!");
+    }
+
+    WorkflowContext workflowContext = getWorkflowContext(workflowName);
+    if (workflowContext == null) {
+      throw new HelixException("Workflow " + workflowName + " context does not exist!");
```
lei-xia, line 568: If the workflow has not started, the context does not exist, right?  But do we allow them to add a task to newly created workflow/job?

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+    if (taskConfig.getCommand() == null && jobConfig.getCommand() == null) {
+      throw new HelixException(
+          "Task cannot be added because both of the job and task have null command!");
+    }
+
+    if (taskConfig.getCommand() != null && jobConfig.getCommand() != null) {
+      throw new HelixException(
+          "Task cannot be added because command existed for both of job and task!");
+    }
+
```
lei-xia, line 589: Can we wrap this logic to a separate (private) method, such as validateTaskConfig(). Also, you may want to put this check at the beginning of the method (to avoid non-necessary ZK read if the input itself is not valid).

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+    Set<TaskState> illegalJobStateForTaskAddition =
+        new HashSet<>(Arrays.asList(TaskState.TIMING_OUT, TaskState.TIMED_OUT, TaskState.FAILING,
+            TaskState.FAILED, TaskState.ABORTED, TaskState.COMPLETED, TaskState.STOPPED,
+            TaskState.STOPPING));
+
+    TaskState jobState = workflowContext.getJobState(nameSpaceJobName);
+
+    if (jobState == null) {
```
lei-xia, line 604: Could the JobState be null when the job was just created (has not be scheduled by the controller)?

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+    if (illegalJobStateForTaskAddition.contains(jobState)) {
+      throw new HelixException("Job " + nameSpaceJobName
+          + " is in illegal state to accept new task. Job State is " + jobState);
+    }
+
+    DataUpdater<ZNRecord> updater = currentData -> {
+      if (currentData != null) {
+        currentData.setMapField(taskConfig.getId(), taskConfig.getConfigMap());
+      } else {
+        LOG.warn("JobConfig DataUpdater: Fails to update JobConfig. CurrentData is null.");
```
lei-xia, line 618: LOG.ERROR?

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+    String taskID = taskConfig.getId();
+    while (System.currentTimeMillis() <= endTime) {
+      JobContext jobContext = getJobContext(nameSpaceJobName);
+      for (Map.Entry<String, Integer> entry : jobContext.getTaskIdPartitionMap().entrySet()) {
+        if (entry.getKey().equals(taskID) && getWorkflowContext(workflowName)
+            .getJobState(nameSpaceJobName) == TaskState.IN_PROGRESS) {
+          return;
+        }
+      }
+      Thread.sleep(1000L);
```
lei-xia, line 639: What will happen if timeout is less than 1000ms?  You may want to adjust the sleep time here based on the given timeout?

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
lei-xia, line 546: Please add more javaDoc here to describe 1) the behaviors and expectations (task may not be added in different senarios), 2) what does the timeout here mean? 3) The possible Exceptions (and what does each mean) could be thrown.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+    }
+    throw new HelixException("An unexpected issue happened while task being added to the job!");
```
lei-xia, line 641: Should we throw a specific TimeoutException instead of a general exception. What is the caller expected to do if it is timeout, call another method to verify whether the task has been successfully added, or retry the addTask()?

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
 
+  /**
+   * Add task to a running (IN-PROGRESS) job with default timeout
+   * @param taskConfig
+   * @param workflowName
+   * @param jobName
+   * @throws Exception
+   */
+  public void addTask(TaskConfig taskConfig, String workflowName, String jobName) throws Exception {
+    addTask(taskConfig, workflowName, jobName, DEFAULT_TIMEOUT);
```
jiajunwang, line 535: The default timeout is 5 mins. It is pretty long. Shall we just provide one method that returns immediately and one that waits for a timeout?

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+    }
+
+    for (String taskEntry : jobConfig.getMapConfigs().keySet()) {
+      if (taskEntry.equals(taskConfig.getId())) {
+        throw new HelixException(
+            "Task cannot be added because another task with the same ID already exists!");
+      }
+    }
+
+    Set<TaskState> illegalJobStateForTaskAddition =
```
jiajunwang, line 597: Please define this as a private static field of the class.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +524,123 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+    String path = _accessor.keyBuilder().resourceConfig(nameSpaceJobName).getPath();
+    boolean status = _accessor.getBaseDataAccessor().update(path, updater, AccessOption.PERSISTENT);
+    if (!status) {
+      LOG.error("Failed to add task to the job {}", nameSpaceJobName);
+      throw new HelixException("Failed to add task to the job");
+    }
+
+    String taskID = taskConfig.getId();
+    while (System.currentTimeMillis() <= endTime) {
+      JobContext jobContext = getJobContext(nameSpaceJobName);
```
jiajunwang, line 632: It could be a trick issue, please verify.
It is possible that the caller pass accessor and propertyStore backed by different ZkClient. In this case, the accessor is used to update and the propertyStore is used to get context here. Given there might be a propagation latency on ZK server-side, this check here might be invalid. We need to set and read using the same ZkClient.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Added not started also. Yes if it is completed, then the add operation will be failed.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Yes. Added this behavior.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: This is the behaviour we are using in the other functions of the TaskDriver. I think we can stick to the same behaviour here as well. Let me know your thoughts.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Changed it to timeout exception.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Added the ability to add a task to a job that is not started. If the job is not started or context is null, we can add tasks to the job.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: I think the agreement in the design doc was to have a timeout to check if the task has been added to not. Also, we are exposing another API with a non-default timeout.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: I don't think it is the case here. Here is how we initialize TaskDriver
```
  public TaskDriver(HelixManager manager) {
    this(manager.getClusterManagmentTool(), manager.getHelixDataAccessor(),
        manager.getHelixPropertyStore(), manager.getClusterName());
  }
```
all of the input parameters are backed by the same manager.

jiajunwang: However, the following public constructor is not in this case.

  public TaskDriver(HelixAdmin admin, HelixDataAccessor accessor,
      HelixPropertyStore<ZNRecord> propertyStore, String clusterName) {

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -77,6 +76,11 @@
   /** Default time out for monitoring workflow or job state */
   private final static int DEFAULT_TIMEOUT = 5 * 60 * 1000; /* 5 mins */
 
+  /** The illegal job states for the jobs to accept new task */
+  private static Set<TaskState> illegalJobStatesForTaskAddition = new HashSet<>(
```
jiajunwang, line 80: nit, private final static... please
Also, in this case, the var name should be ILLEGAL_JOB_STATES_FOR_TASK_ADDITION (or TASKS_MODIFICATION?)

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +530,163 @@ public void enqueueJobs(final String queue, final List<String> jobs,
     }
   }
 
+  /**
+   * Add task to a running (IN-PROGRESS) job or the job which is not started yet. Timeout for this
+   * operation is default timeout
```
jiajunwang, line 535: "default timeout" is not clear enough.
Please mention the real timeout here. Maybe use java doc feature to link with the variable would be a good idea. But I'm not sure if it works fine. please have a try.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +530,163 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+   * Note2: The job can only be added to non-targeted jobs.
+   * Note3: The taskID for new task should be unique. If not, this API throws and exception.
+   * Note4: In case of timeout exception, it is user's responsibility to check whether the task has
+   * been successfully added ot not.
+   * Note5: timeout is the time that this API checks whether the task has been successfully added or
+   * not.
+   * @param workflowName
+   * @param jobName
+   * @param taskConfig
+   * @param timeout
```
jiajunwang, line 564: unit

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
jiajunwang: If there are other places, then please don't hardcode the time.
In addition, IMO, the timeout less than 1000ms does not make sense, so we shall reject the request. Obviously, we shall mention in the public API that timeout less than 1000ms means no wait, or just being rejected.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -28,6 +28,7 @@
 import java.util.Map;
 import java.util.Set;
 
+import java.util.concurrent.TimeoutException;
```
jiajunwang, line 31: nit, format the imports?

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

jiajunwang (approved):
LGTM!

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +533,177 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+   * Note3: The taskID for new task should be unique. If not, this API throws an exception.
+   * Note4: In case of timeout exception, it is user's responsibility to check whether the task has
+   * been successfully added or not.
+   * Note5: timeout is the time that this API checks whether the task has been successfully added or
+   * not.
+   * @param workflowName
+   * @param jobName
+   * @param taskConfig
+   * @param timeoutMs
+   * @throws Exception
```
pkuwm, line 568: Since Exception is pretty generic, I don't see a clear clue where it is from. Can you clarify in what case this Exception is thrown?

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +533,185 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+      throw new HelixException(String.format(
+          "Job %s is a targeted job. New task cannot be added to this job!", nameSpaceJobName));
+    }
+
+    if (taskConfig.getCommand() == null && jobConfig.getCommand() == null) {
+      throw new HelixException(
+          "Task cannot be added because both of the job and task have null command!");
+    }
+
+    if (taskConfig.getCommand() != null && jobConfig.getCommand() != null) {
```
NealSun96, line 657: nit: could combine into (taskConfig.getCommand() == null != jobConfig.getCommand() == null) and say "Command must exist in either job or task, not both".

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +533,185 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+   * Note3: The taskID for the new task should be unique. If not, this API throws an exception.
+   * Note4: In case of timeout exception, it is the user's responsibility to check whether the task
+   * has been successfully added or not.
+   * Note5: timeout is the time that this API checks whether the task has been successfully added or
+   * not.
+   * @param workflowName
+   * @param jobName
+   * @param taskConfig
+   * @param timeoutMs
+   * @throws Exception if there is an issue with the request or the operation. 1-
```
pkuwm, line 572: It doesn't seem a general `Exception` is thrown, right? Since this is an API, to make the API signature clearer for users to use, instead of throwing a general `Exception`, shall we just throw an accurate exception that may be thrown: `TimeoutException`? Then it looks much clearer and easy to understand. RuntimeException doesn't need to be in a method signature.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -526,6 +533,185 @@ public void enqueueJobs(final String queue, final List<String> jobs,
...
+    if (workflowContext == null || jobContext == null) {
+      // Workflow context or job context is null. It means job has not been started. Hence task can
+      // be added to the job
+      addTaskToJobConfig(workflowName, jobName, taskConfig, endTime);
+      return;
+    }
+
+    TaskState jobState = workflowContext.getJobState(nameSpaceJobName);
+
+    if (jobState == null) {
```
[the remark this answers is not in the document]
alirezazamani: removed.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Good suggestion. Changed.

alirezazamani:

PR #1451, opened 2020-10-08
testUpdateConfigFields failure

### LINKED ISSUE #1336: Flaky test testUpdateConfigFields
```
[ERROR] Tests run: 166, Failures: 1, Errors: 0, Skipped: 25, Time elapsed: 52.12 s <<< FAILURE! - in TestSuite
3249
[ERROR] testUpdateConfigFields(org.apache.helix.rest.server.TestClusterAccessor)  Time elapsed: 0.015 s  <<< FAILURE!
3250
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
3251
	at org.apache.helix.rest.server.TestClusterAccessor.testUpdateConfigFields(TestClusterAccessor.java:167)

[INFO] Results:

[ERROR] Failures: 
3257
[ERROR]   TestClusterAccessor.testUpdateConfigFields:167 » IndexOutOfBounds Index: 0, Si...

[ERROR] Tests run: 166, Failures: 1, Errors: 0, Skipped: 25

3262
[INFO] BUILD FAILURE
3263
```

kaisun2000:
### Issues

fix #1336 

### Description

   The test becomes flaky due to waged rebalancer would add INSTANCE_CAPACITY_KEYS
    list field after global calculation. To the test, this means sometime a
    "magic" INSTANCE_CAPACITY_KEYS would appear, depend on timing. Here we fix this
    issue by skip this key for the failed test.

### Tests

https://github.com/apache/helix/runs/1246249153?check_suite_focus=true
This test now pass github run too.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
+++ b/helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
@@ -226,7 +226,6 @@ public class TestClusterAccessor extends AbstractTestClass {
     System.out.println("Start test :" + TestHelper.getTestMethodName());
     String cluster = _clusters.iterator().next();
     ClusterConfig oldConfig = getClusterConfigFromRest(cluster);
-    System.out.println("config at beginning of testAddConfig:" + oldConfig.toString());
 
     ClusterConfig configDelta = new ClusterConfig(cluster);
     configDelta.getRecord().setSimpleField("newField", "newValue");
@@ -244,7 +243,6 @@ public class TestClusterAccessor extends AbstractTestClass {
     oldConfig.getRecord().update(configDelta.getRecord());
     Assert.assertEquals(newConfig, oldConfig,
         "cluster config from response: " + newConfig + " vs cluster config actually: " + oldConfig);
-    System.out.println("config at end testAddConfig:" + newConfig.toString());
     System.out.println("End test :" + TestHelper.getTestMethodName());
   }
 
@@ -254,8 +252,6 @@ public class TestClusterAccessor extends AbstractTestClass {
     String cluster = _clusters.iterator().next();
     ClusterConfig config = getClusterConfigFromRest(cluster);
 
-    System.out.println("new config:" + config.toString());
-
     ZNRecord record = config.getRecord();
 
     String key = record.getSimpleFields().keySet().iterator().next();
@@ -263,7 +259,15 @@ public class TestClusterAccessor extends AbstractTestClass {
     record.getSimpleFields().clear();
     record.setSimpleField(key, value + "--updated");
 
-    key = record.getListFields().keySet().iterator().next();
+    // skip INSTANCE_CAPACITY_KEYS which can be added by waged rebalancer in global run
+    Iterator<String> itListField = record.getListFields().keySet().iterator();
+    while (itListField.hasNext()) {
+      key = itListField.next();
+      if (key.equals("ListField1")) {
+        break;
+      }
+    }
+
     List<String> list = record.getListField(key);
     list.remove(0);
     list.add("newValue--updated");
@@ -302,7 +306,16 @@ public class TestClusterAccessor extends AbstractTestClass {
     record.getSimpleFields().clear();
     record.setSimpleField(simpleKey, value);
 
-    String listKey = record.getListFields().keySet().iterator().next();
+    String listKey = null;
+    // skip INSTANCE_CAPACITY_KEYS which can be added by waged rebalancer in global run
+    Iterator<String> itListField = record.getListFields().keySet().iterator();
+    while (itListField.hasNext()) {
+      listKey = itListField.next();
+      if (listKey.equals("ListField1")) {
+        break;
+      }
+    }
+    Assert.assertTrue(listKey != null);
     List<String> list = record.getListField(listKey);
     record.getListFields().clear();
     record.setListField(listKey, list);
--- a/helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
+++ b/helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
@@ -243,6 +243,7 @@ public class TestClusterAccessor extends AbstractTestClass {
     oldConfig.getRecord().update(configDelta.getRecord());
     Assert.assertEquals(newConfig, oldConfig,
         "cluster config from response: " + newConfig + " vs cluster config actually: " + oldConfig);
+    System.out.println("new config:" + newConfig.toString());
     System.out.println("End test :" + TestHelper.getTestMethodName());
   }
 
@@ -251,6 +252,8 @@ public class TestClusterAccessor extends AbstractTestClass {
     System.out.println("Start test :" + TestHelper.getTestMethodName());
     String cluster = _clusters.iterator().next();
     ClusterConfig config = getClusterConfigFromRest(cluster);
+    
+    System.out.println("new config:" + config.toString());
 
     ZNRecord record = config.getRecord();
 
--- a/helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
+++ b/helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
@@ -226,6 +226,7 @@ public class TestClusterAccessor extends AbstractTestClass {
     System.out.println("Start test :" + TestHelper.getTestMethodName());
     String cluster = _clusters.iterator().next();
     ClusterConfig oldConfig = getClusterConfigFromRest(cluster);
+    System.out.println("config at beginning of testAddConfig:" + oldConfig.toString());
 
     ClusterConfig configDelta = new ClusterConfig(cluster);
     configDelta.getRecord().setSimpleField("newField", "newValue");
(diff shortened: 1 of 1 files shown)


helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
```diff
@@ -243,6 +243,7 @@ public void testAddConfigFields() throws IOException {
     oldConfig.getRecord().update(configDelta.getRecord());
     Assert.assertEquals(newConfig, oldConfig,
         "cluster config from response: " + newConfig + " vs cluster config actually: " + oldConfig);
+    System.out.println("new config:" + newConfig.toString());
```
lei-xia, line 246: Why printout again if it is equal?

jiajunwang: +1

I see system out println quite a lot in your other PRs. If there is a specific reason, please state. Otherwise, please avoid it. Otherwise, it will distract the reviewers' attention and slow down the review process unnecessarily.

zhangmeng916: Is this PR for review? Looks like it's only for debugging purpose instead of fixing the test failure. @kaisun2000 Maybe you can mark it as WIP if it's only for testing.

kaisun2000: This is to figure out what goes wrong, as this test:

1/ does not fail in my local run

2/ fails some times in git hub run.

3/ start to fail recently.

Now the root cause is clear. Basically, the INSTANCE_CAPACITY_KEYS list field can appear in the test and fail the test as the test does not expect this field to "magically" appear, depends on timing. 

It turned out to be the case that Waged Rebalancer would add this field to the cluster config when they finished the first global run.  Depending on the global run timing, the can fail. 

See the detailed analysis in the following issue: https://github.com/apache/helix/issues/1336

jiajunwang: Please keep the debug log in your local environment.
Your debug logs have been reviewed by at least 3 engineers. And I consider this as an unnecessary overhead : (

helix-rest/src/test/java/org/apache/helix/rest/server/TestClusterAccessor.java
```diff
@@ -259,7 +259,15 @@ public void testUpdateConfigFields() throws IOException {
     record.getSimpleFields().clear();
     record.setSimpleField(key, value + "--updated");
 
-    key = record.getListFields().keySet().iterator().next();
+    // skip INSTANCE_CAPACITY_KEYS which can be added by waged rebalancer in global run
```
jiajunwang, line 262: I think an easier way is to reconfigure the dependencies so that all the tests in this class run sequentially.

The shortage of current solutions is that it mixed up the logic between different tests. And once we have more test cases, it may break again.

kaisun2000: Not sure I got what you mean here. Current testng.xml already let the test run sequentially. 

```
<suite name="Suite" parallel="false">
  <test name="Test" preserve-order="true">
    <packages>
      <package name="org.apache.helix.rest.*"/>
    </packages>
  </test>
</suite>
```

kaisun2000: Can you draw some code to illustrate the idea?

jiajunwang: As discussed offline, we agreed that by enforcing the order of test cases, we shall be able to avoid this failure easily without touching the test logic.

kaisun2000: @jiajunwang, I think your theory is not right. The following is dependency trail. 

>testEnableWagedRebalanceForAllResources ->
testActivateSuperCluster ->
testGetStateModelDef ->
testEnableDisableMaintenanceModeWithCustomFields ->
testGetMaintenanceHistory ->
testGetControllerLeadershipHistory ->
testEnableDisableMaintenanceMode ->
testGetClusterConfig ->
testEnableDisableCluster ->
testCreateDeleteCluster ->
testDeleteConfigFields ->
testUpdateConfigFields

kaisun2000: Let us just use the current approach as waged is newest one and would be here for a long time.  Further change may also break it.

jiajunwang: That was a guess, yeah. But the key is the current way is not the correct solution, IMHO.
If our conclusion was wrong, that means we still don't understand the root cause. I would say we need to clearly understand why the WAGED field was added without any related test running. Then we can fix it in the right way.

Do you have the github environment execute order so I can reproduce it?

kaisun2000: https://github.com/apache/helix/runs/1246249153?check_suite_focus=true

look at log in this link or any other test in rest.

kaisun2000: That said, this is not guess. See the analysis in #1336. Basically, the only code path is following:

>computeBestPossibleAssignment -->
globalRebalance -->
ResourceChangeDetector.updateSnapshots -->
ResourceChangeSnapshot(ResourceControllerDataProvider dataProvider,
boolean ignoreNonTopologyChange) -->
ClusterConfigTrimmer.getInstance().trimProperty(dataProvider.getClusterConfig()) -->
HelixPropertyTrimmer.doTrim -->
copyNonTrimmableInfo(originalZNRecord, trimmedZNRecord, getNonTrimmableFields(originalProperty),
false) -->
protected Map<FieldType, Set> getNonTrimmableFields(ClusterConfig property) {
return STATIC_TOPOLOGY_RELATED_FIELD_MAP;
} -->
private static final Map<FieldType, Set> STATIC_TOPOLOGY_RELATED_FIELD_MAP = ImmutableMap
.of(FieldType.SIMPLE_FIELD, ImmutableSet
.of(ClusterConfigProperty.TOPOLOGY.name(),
ClusterConfigProperty.FAULT_ZONE_TYPE.name(),
ClusterConfigProperty.TOPOLOGY_AWARE_ENABLED.name(),
ClusterConfigProperty.MAX_PARTITIONS_PER_INSTANCE.name()),
FieldType.LIST_FIELD, ImmutableSet
.of(ClusterConfigProperty.INSTANCE_CAPACITY_KEYS.name()),
FieldType.MAP_FIELD, ImmutableSet
.of(ClusterConfigProperty.DEFAULT_INSTANCE_CAPACITY_MAP.name(),
ClusterConfigProperty.DEFAULT_PARTITION_WEIGHT_MAP.name(),
ClusterConfigProperty.REBALANCE_PREFERENCE.name()));

note 

```
private void copyNonTrimmableInfo(ZNRecord originalZNRecord, ZNRecord trimmedZNRecord,
      Map<FieldType, Set<String>> nonTrimmableFields, boolean trimValue) {
    for (Map.Entry<FieldType, Set<String>> fieldEntry : nonTrimmableFields.entrySet()) {
      FieldType fieldType = fieldEntry.getKey();
      Set<String> fieldKeySet = fieldEntry.getValue();
      if (null == fieldKeySet || fieldKeySet.isEmpty()) {
        continue;
      }
      switch (fieldType) {
        case SIMPLE_FIELD:
          fieldKeySet.stream().forEach(fieldKey -> {
            if (originalZNRecord.getSimpleFields().containsKey(fieldKey)) {
              trimmedZNRecord.getSimpleFields().putIfAbsent(fieldKey,
                  trimValue ? null : originalZNRecord.getSimpleField(fieldKey));
            }
          });
          break;
        case LIST_FIELD:
          fieldKeySet.stream().forEach(fieldKey -> {
            if (originalZNRecord.getListFields().containsKey(fieldKey)) {
              trimmedZNRecord.getListFields().putIfAbsent(fieldKey,
                  trimValue ? **Collections.EMPTY_LIST** : originalZNRecord.getListField(fieldKey));
            }
          });
          break;

```
This would add an empty INSTANCE_CAPACITY_KEYS=[]

jiajunwang: I don't get it. computeBestPossibleAssignment shall not lead to any config change being written back to ZK. If so, then we need to fix this bug. Are you saying the trimmed result is written back to ZK?

kaisun2000: In this case, before `computeBestPossibleAssignment`, there is no key INSTANCE_CAPACITY_KEYS. After this call, it seems not such key, it would add INSTANCE_CAPACITY_KEYS=[].

It does not really trim anything, but still assume that INSTANCE_CAPACITY_KEYS is empty. Thus, it would explicitly write it out.

jiajunwang: 1. This logic will only be executed when WAGED is used for at least one resource.
2. This logic update in-memory object only, so could you please identify the "write it out" part? I think what you meant is a ZK accessor write? If so, then it would be a bug, we shall fix it.

PR #1452, opened 2020-10-08
thread leakage checker and memory usage reporter #1226

### LINKED ISSUE #1226: Checking Resource Leak at end of a test class
Recently unit test can not complete due to threads leaking and ZK session leaking. We identified the leaks and fixed them. In order for future test not to leak resources, we need a way to enforce test class would release/clean up resource created during test.

A check can/should be implemented in @afterclass in ZkTestBase.java.

kaisun2000:
### Issues

fix #1226

### Description

Add thread leakage checker and memory usage reporter. The two
utility would be invoke before and after test classes. The would
help to detect/monitor resource/memory usage of the unit test.
### Tests

2020-10-08T07:55:24.0931119Z 
2020-10-08T07:55:24.4922133Z [ERROR] Failures: 
2020-10-08T07:55:24.4923510Z [ERROR]   TestDisableCustomCodeRunner.test:236 expected:<false> but was:<true>
2020-10-08T07:55:24.4925729Z [ERROR]   TestClusterStatusMonitorLifecycle.testClusterStatusMonitorLifecycle:290 expected:<true> but was:<false>
2020-10-08T07:55:24.4928050Z [ERROR] Tests run: 1213, Failures: 2, Errors: 0, Skipped: 0
re-run would work.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
+++ b/helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
@@ -176,6 +176,8 @@ public class ThreadLeakageChecker {
       });
     });
 
+    // todo: We should make the following System.out as LOG.INfO once we achieve 0 thread leakage.
+    // todo: also the calling point of this method would fail the test
     // step 3: enforce checking policy
     boolean checkStatus = true;
     for (ThreadCategory threadCategory : ThreadCategory.values()) {
--- a/helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
+++ b/helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
@@ -60,6 +60,24 @@ public class ThreadLeakageChecker {
   private static final String[] TIMER_THRD_PATTERN = new String[]{"time"};
   private static final String[] TASKSTATEMODEL_THRD_PATTERN = new String[]{"TaskStateModel"};
 
+  /*
+   * The two threshold -- warning and limit, are mostly empirical.
+   *
+   * ZkServer, current version has only 4 threads. In case later version use more, we the limit to 100.
+   * The reasoning is that these ZkServer threads are not deemed as leaking no matter how much they have.
+   *
+   * ZkSession is the ZkClient and native Zookeeper client we have. ZkTestBase has 12 at starting up time.
+   * Thus, if there is more than that, it is the test code leaking ZkClient.
+   *
+   * ForkJoin is created by using parallel stream or similar Java features. This is out of our control.
+   * Similar to ZkServer. The limit is to 100 while keep a small _warningLimit.
+   *
+   * Timer should not happen. Setting limit to 2 not 0 mostly because even when you cancel the timer
+   * thread, it may take some not deterministic time for it to go away. So give it some slack here
+   *
+   * Also note, this ThreadLeakage checker depends on the fact that tests are running sequentially.
+   * Otherwise, the report is not going to be accurate.
+   */
   private static enum ThreadCategory {
     ZkServer("zookeeper server threads", 4, 100, ZKSERVER_THRD_PATTERN),
     ZkSession("zkclient/zooKeeper session threads", 12, 12, ZKSESSION_THRD_PATTERN),
--- a/helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
+++ b/helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
@@ -1,3 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
 package org.apache.helix;
 
 import java.util.ArrayList;
@@ -33,20 +52,20 @@ public class ThreadLeakageChecker {
     return Arrays.asList(Arrays.copyOf(threads, count));
   }
 
-  private static final String[] ZkServerThrdPattern =
+  private static final String[] ZKSERVER_THRD_PATTERN =
       {"SessionTracker", "NIOServerCxn", "SyncThread:", "ProcessThread"};
-  private static final String[] ZkSessionThrdPattern =
+  private static final String[] ZKSESSION_THRD_PATTERN =
       new String[]{"ZkClient-EventThread", "ZkClient-AsyncCallback", "-EventThread", "-SendThread"};
-  private static final String[] ForkJoinThrdPattern = new String[]{"ForkJoinPool"};
-  private static final String[] TimerThrdPattern = new String[]{"time"};
-  private static final String[] TaskStateModelThrdPattern = new String[]{"TaskStateModel"};
+  private static final String[] FORKJOIN_THRD_PATTERN = new String[]{"ForkJoinPool"};
+  private static final String[] TIMER_THRD_PATTERN = new String[]{"time"};
+  private static final String[] TASKSTATEMODEL_THRD_PATTERN = new String[]{"TaskStateModel"};
 
   private static enum ThreadCategory {
-    ZkServer("zookeeper server threads", 4, 100, ZkServerThrdPattern),
-    ZkSession("zkclient/zooKeeper session threads", 12, 12, ZkSessionThrdPattern),
-    ForkJoin("fork join pool threads", 2, 100, ForkJoinThrdPattern),
-    Timer("timer threads", 0, 2, TimerThrdPattern),
-    TaskStateModel("TaskStateModel threads", 0, 0, TaskStateModelThrdPattern),
+    ZkServer("zookeeper server threads", 4, 100, ZKSERVER_THRD_PATTERN),
+    ZkSession("zkclient/zooKeeper session threads", 12, 12, ZKSESSION_THRD_PATTERN),
+    ForkJoin("fork join pool threads", 2, 100, FORKJOIN_THRD_PATTERN),
+    Timer("timer threads", 0, 2, TIMER_THRD_PATTERN),
+    TaskStateModel("TaskStateModel threads", 0, 0, TASKSTATEMODEL_THRD_PATTERN),
     Other("Other threads", 0, 2, new String[]{""});
 
     private String _description;
--- a/helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
+++ b/helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
@@ -44,10 +44,10 @@ public class ThreadLeakageChecker {
   private static enum ThreadCategory {
     ZkServer("zookeeper server threads", 4, 100, ZkServerThrdPattern),
     ZkSession("zkclient/zooKeeper session threads", 12, 12, ZkSessionThrdPattern),
-    ForkJoin("fork join pool threads", 2, 10, ForkJoinThrdPattern),
+    ForkJoin("fork join pool threads", 2, 100, ForkJoinThrdPattern),
     Timer("timer threads", 0, 2, TimerThrdPattern),
     TaskStateModel("TaskStateModel threads", 0, 0, TaskStateModelThrdPattern),
-    Other("Other threads", 0, 3, new String[]{""});
+    Other("Other threads", 0, 2, new String[]{""});
 
     private String _description;
     private List<String> _pattern;
@@ -145,25 +145,25 @@ public class ThreadLeakageChecker {
       int limit = threadCategory.getLimit();
       int warningLimit = threadCategory.getWarningLimit();
 
-      Integer catThreadCnt = threadByCnt.get(threadCategory);
-      if (catThreadCnt != null) {
+      Integer categoryThreadCnt = threadByCnt.get(threadCategory);
+      if (categoryThreadCnt != null) {
         boolean dumpThread = false;
-        if (catThreadCnt > limit) {
+        if (categoryThreadCnt > limit) {
           checkStatus = false;
           System.out.println(
-              "Failure " + threadCategory.getDescription() + " has " + catThreadCnt + " thread");
+              "Failure " + threadCategory.getDescription() + " has " + categoryThreadCnt + " thread");
           dumpThread = true;
-        } else if (catThreadCnt > warningLimit) {
+        } else if (categoryThreadCnt > warningLimit) {
           System.out.println(
-              "Warning " + threadCategory.getDescription() + " has " + catThreadCnt + " thread");
+              "Warning " + threadCategory.getDescription() + " has " + categoryThreadCnt + " thread");
           dumpThread = true;
         } else {
-          System.out.println(threadCategory.getDescription() + " has " + catThreadCnt + " thread");
+          System.out.println(threadCategory.getDescription() + " has " + categoryThreadCnt + " thread");
         }
         if (!dumpThread) {
           continue;
         }
-        // print first 10 thread names
+        // print first 100 thread names
         int i = 0;
         for (Thread t : threadByCat.get(threadCategory)) {
           System.out.println(i + " thread:" + t.getName());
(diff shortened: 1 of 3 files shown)


helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
```diff
@@ -127,8 +129,22 @@
   protected Map<String, ClusterSetup> _clusterSetupMap = new HashMap<>();
   protected Map<String, BaseDataAccessor> _baseDataAccessorMap = new HashMap<>();
 
+  static public void reportPhysicalMemory() {
+    com.sun.management.OperatingSystemMXBean os = (com.sun.management.OperatingSystemMXBean)
+        java.lang.management.ManagementFactory.getOperatingSystemMXBean();
+    long physicalMemorySize = os.getTotalPhysicalMemorySize();
+    System.out.println("************ SYSTEM Physical Memory:"  + physicalMemorySize);
+
+    int MB = 1024 * 1024;
```
lei-xia, line 138: minor: int -> long?

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
```diff
@@ -710,7 +726,9 @@ protected Message createMessage(Message.MessageType type, String msgId, String f
   }
 
   @AfterClass
-  public void cleanupLiveInstanceOwners() {
+  public void cleanupLiveInstanceOwners() throws InterruptedException {
```
lei-xia, line 729: Will this method being called by all sub test classes?

lei-xia:
Are we targeting to output the memory usage after each test class? Will this PR make it happen?

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
```diff
@@ -0,0 +1,181 @@
...
+    ThreadGroup candidate = Thread.currentThread().getThreadGroup();
+    while (candidate.getParent() != null) {
+      candidate = candidate.getParent();
+    }
+    return candidate;
+  }
+
+  private static List<Thread> getAllThreads() {
+    ThreadGroup rootThreadGroup = getRootThreadGroup();
+    Thread[] threads = new Thread[32];
```
zhangmeng916, line 27: Isn't the 32 too small as initial value? We have a lot more than that right?

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private static final String[] ZkServerThrdPattern =
+      {"SessionTracker", "NIOServerCxn", "SyncThread:", "ProcessThread"};
+  private static final String[] ZkSessionThrdPattern =
+      new String[]{"ZkClient-EventThread", "ZkClient-AsyncCallback", "-EventThread", "-SendThread"};
+  private static final String[] ForkJoinThrdPattern = new String[]{"ForkJoinPool"};
+  private static final String[] TimerThrdPattern = new String[]{"time"};
+  private static final String[] TaskStateModelThrdPattern = new String[]{"TaskStateModel"};
+
+  private static enum ThreadCategory {
```
zhangmeng916, line 44: Can you give some explanation of the default value? Otherwise, it's a bit hard to follow. E.g., whey ForkJoin is default to 2?

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
```diff
@@ -0,0 +1,181 @@
...
+      });
+    });
+
+    // step 3: enforce checking policy
+    boolean checkStatus = true;
+    for (ThreadCategory threadCategory : ThreadCategory.values()) {
+      int limit = threadCategory.getLimit();
+      int warningLimit = threadCategory.getWarningLimit();
+
+      Integer catThreadCnt = threadByCnt.get(threadCategory);
```
zhangmeng916, line 148: nit: can you use more character for "category", catThreadCnt is not quit self explanatory.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+        if (!dumpThread) {
+          continue;
+        }
+        // print first 10 thread names
+        int i = 0;
+        for (Thread t : threadByCat.get(threadCategory)) {
+          System.out.println(i + " thread:" + t.getName());
+          i++;
+          if (i == 100) {
```
zhangmeng916, line 171: Comments said "10", but I think "100" makes more sense.

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
kaisun2000: All the ones using ZkTestBase

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
```diff
@@ -719,6 +737,17 @@ public void cleanupLiveInstanceOwners() {
       clientMap.clear();
     }
     _liveInstanceOwners.clear();
+
+    boolean status = false;
+    try {
+      status = ThreadLeakageChecker.afterClassCheck(testClassName);
```
zhangmeng916, line 743: Just to make sure, having this check here should cover all tests?

kaisun2000:
> Are we targeting to output the memory usage after each test class? Will this PR make it happen?

Currently, all the ones using ZkTestBase. Since the majority use  ZkTestBase, we have a good idea of the memory usage over time.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
kaisun2000: Good question. The two threshold are mostly empirical. 

ZkServer, our version has only 4 threads now. In case later version use more. I set the limit to 100. The reasoning is that these ZkServer threads are not deemed as leaking no matter how much they have.

ZkSession is the ZkClient and native Zookeeper client we have. ZkTestBase has 12 at starting up time. Thus, if there is more than that, it is the test code leaking ZkClient.

ForkJoin is created by using parallel stream or whatever Java features. This is out of our control, similar to ZkServer. Let me change the _limit to 100 while keep a small _warningLimit. (Will make change for this one).

Timer  should not happen. Setting limit to 2 not 0 mostly because even when you cancel the timer thread, it may take some not deterministic time for it to go away. So give it some slack here.

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
kaisun2000: Only those test derived from ZkTestBase. That is majority for now. We can open another ticket to track using then for all.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
kaisun2000: It does not matter. Note in line 30 ` threads = new Thread[threads.length * 2];`, we double the threads array if they are not large enough, exponentially.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
kaisun2000: changed the comments

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
kaisun2000: changed

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
kaisun2000: changed.

zhangmeng916 (approved):

kaisun2000:
The PR is approved, please help to merge in.

>Add thread leakage checker and memory usage reporter. The two
utility would be invoke before and after test classes. The would
help to detect/monitor resource/memory usage of the unit test.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
```diff
@@ -0,0 +1,181 @@
+package org.apache.helix;
+
```
pkuwm, line 2: @kaisun2000 Can you add Apache License ?

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
```diff
@@ -33,6 +33,7 @@
 import javax.management.MBeanServerConnection;
 import javax.management.ObjectName;
 
+import bsh.This;
```
pkuwm, line 36: Is this needed?

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
pkuwm, line 44: I didn't mean to be picky, but I suggest we still follow the constant style: `static final String[] THIS_IS_CONSTANT`

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
kaisun2000: added.

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
kaisun2000: removed,

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
kaisun2000: changed.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import org.apache.helix.common.ZkTestBase;
+
+
+public class ThreadLeakageChecker {
```
jiajunwang, line 35: Please comment on what is this checker doing.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
jiajunwang: Please put the comment into the class. Code reviewers in the future shall not depend on this Github PR to understand the logic.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
```diff
@@ -0,0 +1,200 @@
...
+    return Arrays.asList(Arrays.copyOf(threads, count));
+  }
+
+  private static final String[] ZKSERVER_THRD_PATTERN =
+      {"SessionTracker", "NIOServerCxn", "SyncThread:", "ProcessThread"};
+  private static final String[] ZKSESSION_THRD_PATTERN =
+      new String[]{"ZkClient-EventThread", "ZkClient-AsyncCallback", "-EventThread", "-SendThread"};
+  private static final String[] FORKJOIN_THRD_PATTERN = new String[]{"ForkJoinPool"};
+  private static final String[] TIMER_THRD_PATTERN = new String[]{"time"};
+  private static final String[] TASKSTATEMODEL_THRD_PATTERN = new String[]{"TaskStateModel"};
```
jiajunwang, line 61: Any thoughts on how we are going to maintain these hardcoded patterns?

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
jiajunwang: Another question, does it only work for sequentially run tests?
If so, please note in the comment.

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
```diff
@@ -710,7 +725,9 @@ protected Message createMessage(Message.MessageType type, String msgId, String f
   }
 
   @AfterClass
-  public void cleanupLiveInstanceOwners() {
+  public void cleanupLiveInstanceOwners() throws InterruptedException {
+    String testClassName = this.getShortClassName();
+    System.out.println("AfterClass:" + testClassName + " afterclass of ZkTestBase called!");
```
jiajunwang, line 730: Please remove this output. I think with no leakage found, we don't need this information.

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
```diff
@@ -719,6 +736,17 @@ public void cleanupLiveInstanceOwners() {
       clientMap.clear();
     }
     _liveInstanceOwners.clear();
+
+    boolean status = false;
+    try {
+      status = ThreadLeakageChecker.afterClassCheck(testClassName);
+    } catch (Exception e) {
+      System.out.println("ThreadLeakageChecker exception:" + e.getStackTrace());
```
jiajunwang, line 744: Why system out print instead of log? If I redirect the test log into a file for debugging, I won't see this information.

helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkHelixJavaApis.java
```diff
@@ -170,6 +171,9 @@ public void beforeClass() throws Exception {
 
   @AfterClass
   public void afterClass() throws Exception {
+    String testClassName = getClass().getSimpleName();
+    System.out.println("AfterClass: " + testClassName + " of TestMultiZkHelixJavaApis called.");
```
jiajunwang, line 175: Same comments as the ZkTestBase class.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
kaisun2000: As long as developers start to use them, they can easier break out more categories and make the changes accordingly. Eventually the baton is passed code contributors.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
kaisun2000: I think the checker by itself is thread safe. But currently if multiple test is running, checker won't be of any use for obvious reasons. Let me add an comment its usage depends on test running sequentially.

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
kaisun2000: added the comment

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
kaisun2000: added to the comment that we need to run tests sequentially to make this leakage check to be meaningful.

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
kaisun2000: removed

helix-core/src/test/java/org/apache/helix/common/ZkTestBase.java
kaisun2000: Good point. changed to log.error

helix-core/src/test/java/org/apache/helix/ThreadLeakageChecker.java
jiajunwang: How about new thread names or new threads that are not tracked by these patterns?
As a minimum requirement, I suggest tracking unknown threads as well. And a limitation should be added onto those threads as well. So if anyone changes the name, or introducing new threads that are leaking, the test will break. And they will be forced to modify this section.

jiajunwang (commented):
Still, 2 problems that I also mentioned in the other test improvement PRs.
1. What is the guideline for using a system print vs. log? Without confirming this point, I cannot review it effectively. I prefer log, BTW.
2. Why not throw an exception on leakage? If just print error, then we can just let the tool runs every night. There is no need to do it with every "mvn test".
For example, if people ignore the error, then we might figure it out weeks later. Then this check will be very less helpful.

kaisun2000:
> Still, 2 problems that I also mentioned in the other test improvement PRs.
> 
> 1. What is the guideline for using a system print vs. log? Without confirming this point, I cannot review it effectively. I prefer log, BTW.
> 2. Why not throw an exception on leakage? If just print error, then we can just let the tool runs every night. There is no need to do it with every "mvn test".
>    For example, if people ignore the error, then we might figure it out weeks later. Then this check will be very less helpful.

If we throw exception, current test won't work. I reduced the threads leakage from around 3000 to several hundreds at this point. Getting it to 0 is still some way to go.

kaisun2000:
Sync-ed with JJ offline. Once the thread leakage reduced to 0. We will enable the throw exception (failing test). Before that, we use system.out to log it out for people to examine without enable info level logging

jiajunwang (approved):

kaisun2000:
The PR is approved. Please help to merge to trunk

>Add thread leakage checker and memory usage reporter. The two
utility would be invoke before and after test classes. The would
help to detect/monitor resource/memory usage of the unit test.

PR #1455, opened 2020-10-08
Fix task related flaky test due to phantom read (WIP)

### LINKED ISSUE #1454: Fix task related unstable unit test
Currently selective update can result in "phantom read". For example, when job1 cfg is added and then workflow cfg containing job 1 is updated. The selective update may only see updated workflow cfg but not job 1 cfg. This would cause job1 cfg and ctx removed. THus hanging tests.

Currently we work around this issue by setting purge interval to -1 and enhance logging in task to make test stable.

kaisun2000:
### Issues

fix #1454 

### Description

Currently selective update can result in "phantom read". For
example, when job1 cfg is added and then workflow cfg
containing job 1 is updated. The selective update may only
see updated workflow cfg but not job 1 cfg. This would cause
job1 cfg and ctx removed. THus hanging tests.

Currently we work around this issue by setting purge interval
to -1 and enhance logging in task to make test stable.

### Tests

running

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
@@ -61,12 +61,14 @@ import org.apache.helix.task.TaskState;
 import org.apache.helix.task.TaskUtil;
 import org.apache.helix.task.WorkflowConfig;
 import org.apache.helix.task.WorkflowContext;
+import org.slf4j.LoggerFactory;
 import org.testng.Assert;
 
 /**
  * Static test utility methods.
  */
 public class TaskTestUtil {
+  private static final org.slf4j.Logger logger = LoggerFactory.getLogger(TaskTestUtil.class);
   public static final String JOB_KW = "JOB";
   private final static int _default_timeout = 2 * 60 * 1000; /* 2 mins */
 
@@ -173,8 +175,14 @@ public class TaskTestUtil {
         }
       }
     }
-    return maxRunningCount > 1 && (workflowConfig.isJobQueue() ? maxRunningCount <= workflowConfig
+    boolean retVal =  maxRunningCount > 1 && (workflowConfig.isJobQueue() ? maxRunningCount <= workflowConfig
         .getParallelJobs() : true);
+    if (!retVal) {
+      logger.error("maxRunningCount={}, workflowConfig.isJobQueue()={}, maxRunningCount={}, workflowConfig.getParallelJobs()={}, workFlowConfig={}, stack trace {}",
+          maxRunningCount, workflowConfig.isJobQueue(), maxRunningCount, workflowConfig.getParallelJobs(),
+          workflowConfig);
+    }
+    return retVal;
   }
 
   public static Date getDateFromStartTime(String startTime)
@@ -231,9 +239,17 @@ public class TaskTestUtil {
 
   public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
       int failureThreshold, int capacity) {
+    return buildJobQueue(jobQueueName, delayStart, failureThreshold, capacity, false);
+  }
+
+  public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
+      int failureThreshold, int capacity, boolean disablePurge) {
     WorkflowConfig.Builder workflowCfgBuilder = new WorkflowConfig.Builder(jobQueueName);
     workflowCfgBuilder.setExpiry(120000);
     workflowCfgBuilder.setCapacity(capacity);
+    if (disablePurge) {
+      workflowCfgBuilder.setJobPurgeInterval(-1);
+    }
 
     Calendar cal = Calendar.getInstance();
     cal.set(Calendar.MINUTE, cal.get(Calendar.MINUTE) + delayStart / 60);
@@ -245,6 +261,11 @@ public class TaskTestUtil {
       workflowCfgBuilder.setFailureThreshold(failureThreshold);
     }
     return new JobQueue.Builder(jobQueueName).setWorkflowConfig(workflowCfgBuilder.build());
+  }
+
+  public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
+      int failureThreshold, boolean disablePurge) {
+    return buildJobQueue(jobQueueName, delayStart, failureThreshold, 500, disablePurge);
   }
 
   public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
@@ -239,15 +239,15 @@ public class TaskTestUtil {
 
   public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
       int failureThreshold, int capacity) {
-    return buildJobQueue(jobQueueName, delayStart, failureThreshold, capacity, false);
+    return buildJobQueue(jobQueueName, delayStart, failureThreshold, capacity, true);
   }
 
   public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
-      int failureThreshold, int capacity, boolean disablePurge) {
+      int failureThreshold, int capacity, boolean doPurge) {
     WorkflowConfig.Builder workflowCfgBuilder = new WorkflowConfig.Builder(jobQueueName);
     workflowCfgBuilder.setExpiry(120000);
     workflowCfgBuilder.setCapacity(capacity);
-    if (disablePurge) {
+    if (!doPurge) {
       workflowCfgBuilder.setJobPurgeInterval(-1);
     }
 
@@ -264,8 +264,8 @@ public class TaskTestUtil {
   }
 
   public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
-      int failureThreshold, boolean disablePurge) {
-    return buildJobQueue(jobQueueName, delayStart, failureThreshold, 500, disablePurge);
+      int failureThreshold, boolean doPurge) {
+    return buildJobQueue(jobQueueName, delayStart, failureThreshold, 500, doPurge);
   }
 
   public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
@@ -68,7 +68,7 @@ import org.testng.Assert;
  * Static test utility methods.
  */
 public class TaskTestUtil {
-  private static final org.slf4j.Logger logger = LoggerFactory.getLogger(TaskTestUtil.class);
+  private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(TaskTestUtil.class);
   public static final String JOB_KW = "JOB";
   private final static int _default_timeout = 2 * 60 * 1000; /* 2 mins */
 
@@ -178,7 +178,7 @@ public class TaskTestUtil {
     boolean retVal =  maxRunningCount > 1 && (workflowConfig.isJobQueue() ? maxRunningCount <= workflowConfig
         .getParallelJobs() : true);
     if (!retVal) {
-      logger.error("maxRunningCount={}, workflowConfig.isJobQueue()={}, maxRunningCount={}, workflowConfig.getParallelJobs()={}, workFlowConfig={}, stack trace {}",
+      LOG.error("maxRunningCount={}, workflowConfig.isJobQueue()={}, maxRunningCount={}, workflowConfig.getParallelJobs()={}, workFlowConfig={}, stack trace {}",
           maxRunningCount, workflowConfig.isJobQueue(), maxRunningCount, workflowConfig.getParallelJobs(),
           workflowConfig);
     }
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestBatchAddJobs.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestBatchAddJobs.java
@@ -87,7 +87,7 @@ public class TestBatchAddJobs extends ZkTestBase {
   @AfterClass
   public void afterClass() {
     String testClassName = this.getShortClassName();
-    System.out.println("AfterClass: " + testClassName + " of TestBachAddJobs called.");
+    System.out.println("AfterClass: " + testClassName + " called.");
 
     for (SubmitJobTask submitJobTask : _submitJobTasks) {
       submitJobTask.interrupt();
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestBatchAddJobs.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestBatchAddJobs.java
@@ -86,6 +86,9 @@ public class TestBatchAddJobs extends ZkTestBase {
 
   @AfterClass
   public void afterClass() {
+    String testClassName = this.getShortClassName();
+    System.out.println("AfterClass: " + testClassName + " of TestBachAddJobs called.");
+
     for (SubmitJobTask submitJobTask : _submitJobTasks) {
       submitJobTask.interrupt();
       submitJobTask.cleanup();
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestDeleteJobFromJobQueue.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestDeleteJobFromJobQueue.java
@@ -69,11 +69,18 @@ public class TestDeleteJobFromJobQueue extends TaskTestBase {
     Assert
         .assertNotNull(_driver.getJobContext(TaskUtil.getNamespacedJobName(jobQueueName, "job2")));
 
-    // The following force delete for the job should go through without getting an exception
+    // force deletion can have Exception thrown as controller is writing to propertystore path too.
+    // https://github.com/apache/helix/issues/1406
+
+    // Thus, we stop pipeline to make sure there is not such race condition.
+    _gSetupTool.getClusterManagementTool().enableCluster(CLUSTER_NAME, false);
+    Thread.sleep(3000); // note this sleep is critical as it would take time for controller to stop
+
     _driver.deleteJob(jobQueueName, "job2", true);
 
     // Check that the job has been force-deleted (fully gone from ZK)
     Assert.assertNull(_driver.getJobConfig(TaskUtil.getNamespacedJobName(jobQueueName, "job2")));
     Assert.assertNull(_driver.getJobContext(TaskUtil.getNamespacedJobName(jobQueueName, "job2")));
+
   }
 }
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
@@ -19,10 +19,8 @@ package org.apache.helix.integration.task;
  * under the License.
  */
 
-import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Date;
 import java.util.List;
 import org.apache.helix.TestHelper;
 import org.apache.helix.integration.manager.ClusterControllerManager;
@@ -34,14 +32,16 @@ import org.apache.helix.task.TaskUtil;
 import org.apache.helix.task.Workflow;
 import org.apache.helix.task.WorkflowConfig;
 import org.testng.Assert;
-import org.testng.ITestContext;
-import org.testng.annotations.AfterClass;
-import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeClass;
-import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
 public class TestEnqueueJobs extends TaskTestBase {
+  
+  @BeforeClass
+  public void beforeClass() throws Exception {
+    setSingleTestEnvironment();
+    super.beforeClass();
+  }
 
   @Test
   public void testJobQueueAddingJobsOneByOne() throws InterruptedException {
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
@@ -19,8 +19,10 @@ package org.apache.helix.integration.task;
  * under the License.
  */
 
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import org.apache.helix.TestHelper;
 import org.apache.helix.integration.manager.ClusterControllerManager;
@@ -32,7 +34,11 @@ import org.apache.helix.task.TaskUtil;
 import org.apache.helix.task.Workflow;
 import org.apache.helix.task.WorkflowConfig;
 import org.testng.Assert;
+import org.testng.ITestContext;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
 public class TestEnqueueJobs extends TaskTestBase {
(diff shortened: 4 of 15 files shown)


helix-core/src/test/java/org/apache/helix/integration/task/TestBatchAddJobs.java
```diff
@@ -86,6 +86,9 @@ private String getPrefix(String job) {
 
   @AfterClass
   public void afterClass() {
+    String testClassName = this.getShortClassName();
+    System.out.println("AfterClass: " + testClassName + " of TestBachAddJobs called.");
```
jiajunwang, line 90: Humm, testClassName should be "TestBachAddJobs", right? So it will be "AfterClass: TestBachAddJobs of TestBachAddJobs called."?

Moreover, why we add this logic to every child test class instead of print in the parent class?

helix-core/src/test/java/org/apache/helix/integration/task/TestDeleteJobFromJobQueue.java
```diff
@@ -69,11 +69,18 @@ public void testForceDeleteJobFromJobQueue() throws InterruptedException {
     Assert
         .assertNotNull(_driver.getJobContext(TaskUtil.getNamespacedJobName(jobQueueName, "job2")));
 
-    // The following force delete for the job should go through without getting an exception
+    // force deletion can have Exception thrown as controller is writing to propertystore path too.
+    // https://github.com/apache/helix/issues/1406
+
+    // Thus, we stop pipeline to make sure there is not such race condition.
+    _gSetupTool.getClusterManagementTool().enableCluster(CLUSTER_NAME, false);
+    Thread.sleep(3000); // note this sleep is critical as it would take time for controller to stop
```
jiajunwang, line 77: I think you also emphasized that we shall avoid using Thread.sleep() in the tests. 3-second sleep is not guaranteed to work.
Can we use verifier with a timeout to wait until the pipeline stopped?

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -41,13 +47,39 @@
...
+
+  @AfterClass
+  @Override
+  public void afterClass() throws Exception {
+    //WorkflowConfig.enableJobPurge();
+    super.afterClass();
+  }
+
+  @BeforeMethod
+  public void beforeTest(Method testMethod, ITestContext testContext) {
```
jiajunwang, line 61: Please confirm if this has already been done by the parent class ZkTestBase.
I think this is duplicate logic.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -41,13 +47,39 @@
   public void beforeClass() throws Exception {
     setSingleTestEnvironment();
     super.beforeClass();
+    //WorkflowConfig.disableJobPurge();
```
jiajunwang, line 50: What is this?

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -41,13 +47,39 @@
...
+  }
+
+  @AfterClass
+  @Override
+  public void afterClass() throws Exception {
```
jiajunwang, line 55: I did a quick test, I think you can just remove it and the super.afterClass() will still be called, since you don't have any method with the same name to override it in that case.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -41,13 +47,39 @@
...
+
+  @BeforeMethod
+  public void beforeTest(Method testMethod, ITestContext testContext) {
+    long startTime = System.currentTimeMillis();
+    System.out.println("START " + testMethod.getName() + " at " + new Date(startTime));
+    testContext.setAttribute("StartTime", System.currentTimeMillis());
+  }
+
+  @AfterMethod
+  public void endTest(Method testMethod, ITestContext testContext) {
```
jiajunwang, line 68: Samething here, ZKTestBase has this method with same logic already.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -65,12 +97,13 @@ public void testJobQueueAddingJobsOneByOne() throws InterruptedException {
         TaskState.COMPLETED);
   }
 
-  @Test
+  @Test()
```
jiajunwang, line 100: It's OK, but why?

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -131,15 +165,16 @@ public void testJobSubmitGenericWorkflows() throws InterruptedException {
     _driver.pollForWorkflowState(workflowName, TaskState.COMPLETED);
   }
 
-  @Test
+  @Test()
```
jiajunwang, line 168: Same here

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -202,7 +237,7 @@ public void testQueueParallelJobs() throws InterruptedException {
     Assert.assertTrue(minFinishTime > maxStartTime);
   }
 
-  @Test
+  @Test()
```
jiajunwang, line 240: Same here

helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
```diff
@@ -231,9 +239,17 @@ public static Date getDateFromStartTime(String startTime)
 
   public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
       int failureThreshold, int capacity) {
+    return buildJobQueue(jobQueueName, delayStart, failureThreshold, capacity, false);
+  }
+
+  public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
+      int failureThreshold, int capacity, boolean disablePurge) {
```
jiajunwang, line 246: nit, but it would be more intuitive to be "boolean doPurge"

helix-core/src/test/java/org/apache/helix/integration/task/TestMaxNumberOfAttemptsMasterSwitch.java
```diff
@@ -71,6 +71,9 @@ public void beforeClass() throws Exception {
 
   @AfterClass
   public void afterClass() throws Exception {
+    String testClassName = this.getShortClassName();
```
jiajunwang, line 74: Instead of doing this, I suggest refactoring the afterClass() method in TaskSynchronizedTestBase or ZkTestBase. In this case, you don't have duplicate code. And all the child classes have it automatically.

helix-core/src/test/java/org/apache/helix/integration/task/TestQuotaBasedScheduling.java
```diff
@@ -116,6 +117,11 @@ public void beforeClass() throws Exception {
     _jobCommandMap = Maps.newHashMap();
   }
 
+  @AfterClass
+  public void afterClass() throws Exception {
```
jiajunwang, line 121: I think it is not necessary. The super afterClass will still be called if there is no class with the same name override it in the child class. Please have a try.

helix-core/src/test/java/org/apache/helix/integration/task/TestQuotaBasedScheduling.java
```diff
@@ -135,6 +141,7 @@ public void testSchedulingWithoutQuota() throws InterruptedException {
     Workflow.Builder workflowBuilder = new Workflow.Builder(workflowName);
     WorkflowConfig.Builder configBuilder = new WorkflowConfig.Builder(workflowName);
     configBuilder.setAllowOverlapJobAssignment(true);
+    configBuilder.setJobPurgeInterval(-1);
```
jiajunwang, line 144: Are you suggest we remove these changes one by one if later we fix the purge issue?
@NealSun96 , what is the ETA of your change, please? I think removing these changes will be troublesome. Unless we separate them into a single change and revert it completely later.

helix-core/src/test/java/org/apache/helix/integration/task/TestQuotaBasedScheduling.java
```diff
@@ -491,7 +502,7 @@ public void testThreadLeak() throws InterruptedException {
    * Tests quota-based scheduling for a job queue with different quota types.
    * @throws InterruptedException
    */
-  @Test(dependsOnMethods = "testSchedulingWithoutQuota")
+  @Test(dependsOnMethods = "testSchedulingWithoutQuota", enabled = true)
```
jiajunwang, line 505: Isn't enabled default to be true?

helix-core/src/test/java/org/apache/helix/integration/task/TestStuckTaskQuota.java
```diff
@@ -63,6 +63,9 @@ public void beforeClass() throws Exception {
 
   @AfterClass
   public void afterClass() throws Exception {
+    String testClassName = this.getShortClassName();
```
jiajunwang, line 66: As we can see, there are many changes like this. So let's do it in the parent class. Shall we?

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskAssignmentCalculator.java
```diff
@@ -102,6 +108,26 @@ public void beforeClass() throws Exception {
...
+  }
+
+  @AfterMethod
+  public void endTest(Method testMethod, ITestContext testContext) {
+    Long startTime = (Long) testContext.getAttribute("StartTime");
+    long endTime = System.currentTimeMillis();
+    System.out.println("END " + testMethod.getName() + " at " + new Date(endTime) + ", took: "
+        + (endTime - startTime) + "ms.");
+  }
+
```
jiajunwang, line 130: Same here, I think if the classes are defined carefully, then the parent class logic can cover these.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancer.java
```diff
@@ -43,9 +45,21 @@
...
+  @BeforeClass
+  public void beforeClass() throws Exception {
+    super.beforeClass();
+  }
+
+  @AfterClass
+  public void afterClass() throws Exception {
+    super.afterClass();
+  }
+
```
jiajunwang, line 62: Same here. Are you sure we need them?

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerParallel.java
```diff
@@ -44,6 +45,11 @@ public void beforeClass() throws Exception {
     super.beforeClass();
   }
 
+  @AfterClass
```
jiajunwang, line 48: Same here.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerStopResume.java
```diff
@@ -32,6 +32,7 @@
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
+import com.sun.corba.se.spi.orbutil.threadpool.Work;
```
jiajunwang, line 35: Humm, interesting package. What is the usage, please?

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerStopResume.java
```diff
@@ -51,20 +52,36 @@
...
 import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
 public class TestTaskRebalancerStopResume extends TaskTestBase {
   private static final Logger LOG = LoggerFactory.getLogger(TestTaskRebalancerStopResume.class);
   private static final String JOB_RESOURCE = "SomeJob";
 
+  @BeforeClass
```
jiajunwang, line 63: Same here

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerStopResume.java
```diff
@@ -159,13 +190,16 @@ public void stopAndResumeNamedQueue() throws Exception {
     verifyJobNotInQueue(queueName, namespacedJob2);
   }
 
-  @Test
+  @Test(enabled = true)
```
jiajunwang, line 193: default is true.

  /**
   * Whether methods on this class/method are enabled.
   */
  public boolean enabled() default true;

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerStopResume.java
```diff
@@ -413,25 +447,38 @@ public void stopAndDeleteQueue() throws Exception {
...
     JobConfig.Builder job2 = new JobConfig.Builder().setCommand(MockTask.TASK_COMMAND)
+        .setJobCommandConfigMap(Collections.singletonMap(MockTask.JOB_DELAY, "2000"))
         .setTargetResource(WorkflowGenerator.DEFAULT_TGT_DB).setTargetPartitionStates(slave);
     String job2Name = "slaveJob";
     LOG.info("Enqueuing job2: " + job2Name);
+    jobNames.add(job2Name);
+    jobBuilders.add(job2);
     _driver.enqueueJob(queueName, job2Name, job2);
 
+    //_driver.enqueueJobs(queueName, jobNames, jobBuilders);
```
jiajunwang, line 480: ???

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerStopResume.java
```diff
@@ -413,25 +447,38 @@ public void stopAndDeleteQueue() throws Exception {
...
     JobQueue qCfg = new JobQueue.Builder(queueName).fromMap(wfCfg.getResourceConfigMap()).build();
     _driver.createQueue(qCfg);
 
     // Enqueue 2 jobs
+    List<String> jobNames = new ArrayList<>();
+    List<JobConfig.Builder> jobBuilders = new ArrayList<>();
+
     Set<String> master = Sets.newHashSet("MASTER");
     JobConfig.Builder job1 = new JobConfig.Builder().setCommand(MockTask.TASK_COMMAND)
+        .setJobCommandConfigMap(Collections.singletonMap(MockTask.JOB_DELAY, "2000"))
```
jiajunwang, line 462: I'm not sure if we really need it. But looks harmless.
@alirezazamani @NealSun96 , could you please help to take another look?

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskSchedulingTwoCurrentStates.java
```diff
@@ -93,6 +93,8 @@ public void beforeClass() throws Exception {
 
   @AfterClass()
   public void afterClass() throws Exception {
+    String testClassName = this.getShortClassName();
```
jiajunwang, line 96: Same here, try to update parent class please.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskThrottling.java
```diff
@@ -50,6 +52,10 @@ public void beforeClass() throws Exception {
     super.beforeClass();
   }
 
+  @AfterClass
+  public void afterClass() throws Exception {
```
jiajunwang, line 56: Same here.

jiajunwang (changes requested):
I have concerns about this PR.
1. The way we disable purge is by setting the config everywhere. As we discussed, it is a temporary resolution. When we need to remove them, it would be messy.
2. This PR mixes other changes with the purge change. Meaning we cannot revert it to remove the temporary "fix".
3. I think we agreed on wait until Neal evaluate the ETA of a more graceful temp solution, do we have the conclusion yet?
4. I think we can refactor a little bit by leveraging parent class. The current PR introduces lots of duplicate logic.

helix-core/src/test/java/org/apache/helix/integration/task/TestDeleteJobFromJobQueue.java
kaisun2000: How do we verify the pipeline stops processing event? Note, you can't just rely on pause signal node is in the Zookeeper. 

This is similar to waitTillVerify(). It seems to me there is not good way to wait for some signal.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
kaisun2000: The should be removed. 

Previously, we tried to set `protected static final long DEFAULT_JOB_PURGE_INTERVAL = 30 * 60 * 1000; // default 30 minutes` to -1 in WorkflowConfig.java, by using this function. 

This has its drawback as people have concerns as it pollutes production code as:

1/ need to make DEFAULT_JOB_PURGE_INTERVAL not final.
2/ The WorkflowConfig.disableJobPurge needs to be public. Thus, not testOnly.

helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
```diff
@@ -61,12 +61,14 @@
...
 import org.apache.helix.task.WorkflowConfig;
 import org.apache.helix.task.WorkflowContext;
+import org.slf4j.LoggerFactory;
 import org.testng.Assert;
 
 /**
  * Static test utility methods.
  */
 public class TaskTestUtil {
+  private static final org.slf4j.Logger logger = LoggerFactory.getLogger(TaskTestUtil.class);
```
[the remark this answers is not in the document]
kaisun2000: Changed to LOG

helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
```diff
@@ -231,9 +239,17 @@ public static Date getDateFromStartTime(String startTime)
 
   public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
```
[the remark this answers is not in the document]
kaisun2000: There are other places to use this function. If we don't have this signature, that means probably a lot of  usage places are also changed. That is not essential change, make it hard for people to reason and may be later to revert.

kaisun2000:
> I have concerns about this PR.
> 
> 1. The way we disable purge is by setting the config everywhere. As we discussed, it is a temporary resolution. When we need to remove them, it would be messy.
> 2. This PR mixes other changes with the purge change. Meaning we cannot revert it to remove the temporary "fix".
> 3. I think we agreed on wait until Neal evaluate the ETA of a more graceful temp solution, do we have the conclusion yet?
> 4. I think we can refactor a little bit by leveraging parent class. The current PR introduces lots of duplicate logic.

3/ Last time we discussed, the conclusion is that let Neal work on read one more time before purging as a temp solution. I will help to validate by reproducing the phantom read problem in the debugger. Let us sync about this tomorrow then.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerStopResume.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerStopResume.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestQuotaBasedScheduling.java
kaisun2000: removed

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
kaisun2000: restored.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
kaisun2000: restored.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerStopResume.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
kaisun2000: Right. This should be removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
kaisun2000: Right. This should be removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
kaisun2000: These @beforeClass and @afterClass is due to the old way using WorkflowConfig.disableJobPurge(). Should be removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestMaxNumberOfAttemptsMasterSwitch.java
kaisun2000: Good point. Moved to TaskSynchronizedTestBase @afterclass

helix-core/src/test/java/org/apache/helix/integration/task/TestQuotaBasedScheduling.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestStuckTaskQuota.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskAssignmentCalculator.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancer.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerParallel.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerStopResume.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskSchedulingTwoCurrentStates.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskThrottling.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/task/TestBatchAddJobs.java
kaisun2000: Changed to  System.out.println("AfterClass: " + testClassName + " called.");

Now changed to print in parent class, as suggested.

helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
kaisun2000: Changed to do purge

helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
```diff
@@ -247,6 +263,11 @@ public static Date getDateFromStartTime(String startTime)
     return new JobQueue.Builder(jobQueueName).setWorkflowConfig(workflowCfgBuilder.build());
   }
 
+  public static JobQueue.Builder buildJobQueue(String jobQueueName, int delayStart,
```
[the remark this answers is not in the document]
kaisun2000: changed to do purge

helix-core/src/test/java/org/apache/helix/integration/task/TestEnqueueJobs.java
```diff
@@ -32,22 +34,23 @@
...
+import org.testng.ITestContext;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
 public class TestEnqueueJobs extends TaskTestBase {
 
-  @BeforeClass
```
kaisun2000, line 40: This should not be removed,

helix-core/src/test/java/org/apache/helix/integration/task/TestDeleteJobFromJobQueue.java
NealSun96: Is it better to retry `deleteJob` then? Disable the cluster and retry until no exception.

helix-core/src/test/java/org/apache/helix/integration/task/TestQuotaBasedScheduling.java
NealSun96: It's scheduled this sprint, and I can finish it as fast as mid-next week

helix-core/src/test/java/org/apache/helix/integration/task/TestTaskRebalancerStopResume.java
NealSun96: This is certainly harmless because we poll for completed. What is the reason why we need the delay, though?

helix-core/src/test/java/org/apache/helix/integration/task/TaskTestUtil.java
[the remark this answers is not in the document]
jiajunwang: This specific one has 3 callers. So please just change it.

The one which has many references outside is "public static JobQueue.Builder buildJobQueue(String jobQueueName) {". But it seems that you are not changing it. So I guess we can just change it without too much overhead.

helix-core/src/test/java/org/apache/helix/integration/task/TestBatchAddJobs.java
```diff
@@ -86,6 +86,9 @@ private String getPrefix(String job) {
 
   @AfterClass
   public void afterClass() {
+    String testClassName = this.getShortClassName();
```
jiajunwang, line 89: Print in the superclass ZkTestBase.java?

Issue #1453, opened 2020-10-08
Race condition cause Timer leak in GenericHelixController


xyuanlu:
### Describe the bug
During our Helix controller log analysis, we found that the number of periodic rebalance event trigged for a particular clusters increase every hour with no config change. This implies the periodic rebalance's Timer leakage issue.

This is the log analysis result we have. The cluster has  periodic rebalance rate set to 2400 millisecond. 

```
| cluster_name  | Time_interval start (duration = 1h)  | # of  periodic rebalance event scheduled|
|  CLUSTER_X   | 2020-10-02 18:00:00                      |          15               |
|  CLUSTER_X   | 2020-10-02 19:00:00                      |          15               |
| CLUSTER_X    | 2020-10-02 20:00:00                     |          22               |
|  CLUSTER_X   | 2020-10-02 21:00:00                      |          75               |
|  CLUSTER_X   | 2020-10-02 22:00:00                     |          97               |
```

PR #1456, opened 2020-10-09
Fix Periodic rebalancer Timer leak

xyuanlu:
### Issues

#1453 Race condition cause Timer leak in GenericHelixController

### Description

We found a periodic rebalance's Timer leakage issue during log analysis. In current startPeriodRebalance, two thread may interference with each other. This may result in one timer got canceled twice, two timers are created with one timer leaked. 

```
  void startPeriodRebalance(long period, HelixManager manager) {
    if (period != _timerPeriod) {
      logger.info("Controller starting periodical rebalance timer at period " + period);
      if (_periodicalRebalanceTimer != null) {
        _periodicalRebalanceTimer.cancel();                  <<<------ 
      } 
      _periodicalRebalanceTimer =
          new Timer("GenericHelixController_" + _clusterName + "_periodical_Timer", true);     <<<-----
     ......
  }
```
This PR changes Timer to use a SingleThreadScheduledExecutor and adds a synchronized block in start/stop PeriodRebalance.

### Tests

NA

```
[ERROR] Failures: 
[INFO]
[ERROR] Tests run: 1234, Failures: 1, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:23 h

```

### Commits

### Documentation (Optional)

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -174,12 +174,12 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
 
   /**
    * The executors that can periodically run the rebalancing pipeline. A
-   * SingleThreadScheduledExecutor will start if there is one resource group has the config to do
+   * SingleThreadScheduledExecutor will start if there is resource group that has the config to do
    * periodically rebalance.
    */
   private static final ScheduledExecutorService _periodicalRebalanceExecutor =
       Executors.newSingleThreadScheduledExecutor();
-  private ScheduledFuture _periodicRebalancerFutureTasks = null;
+  private ScheduledFuture _periodicRebalanceFutureTask = null;
   long _timerPeriod = Long.MAX_VALUE;
   private final Object _lock = new Object();
 
@@ -339,16 +339,16 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
       logger.info("Controller starting periodical rebalance timer at period " + period);
       ScheduledFuture lastScheduledFuture = null;
       synchronized (_lock) {
-        if (_periodicRebalancerFutureTasks!=null && !_periodicRebalancerFutureTasks.isCancelled()) {
-          lastScheduledFuture = _periodicRebalancerFutureTasks;
+        if (_periodicRebalanceFutureTask!=null && !_periodicRebalanceFutureTask.isCancelled()) {
+          lastScheduledFuture = _periodicRebalanceFutureTask;
         }
         _timerPeriod = period;
-        _periodicRebalancerFutureTasks = _periodicalRebalanceExecutor
+        _periodicRebalanceFutureTask = _periodicalRebalanceExecutor
             .scheduleAtFixedRate(new RebalanceTask(manager, ClusterEventType.PeriodicalRebalance),
                 _timerPeriod, _timerPeriod, TimeUnit.MILLISECONDS);
       }
       if (lastScheduledFuture != null) {
-        lastScheduledFuture.cancel(true /* mayInterruptIfRunning */);
+        lastScheduledFuture.cancel(false /* mayInterruptIfRunning */);
       }
     } else {
       logger.info("Controller already has periodical rebalance timer at period " + _timerPeriod);
@@ -361,8 +361,8 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
   void stopPeriodRebalance() {
     logger.info("Controller stopping periodical rebalance timer at period " + _timerPeriod);
     synchronized (_lock) {
-      if (_periodicRebalancerFutureTasks != null && !_periodicRebalancerFutureTasks.isCancelled()) {
-        _periodicRebalancerFutureTasks.cancel(true /* mayInterruptIfRunning */);
+      if (_periodicRebalanceFutureTask != null && !_periodicRebalanceFutureTask.isCancelled()) {
+        _periodicRebalanceFutureTask.cancel(false /* mayInterruptIfRunning */);
         _timerPeriod = Long.MAX_VALUE;
       }
     }
@@ -1307,9 +1307,6 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
   }
 
   public void shutdown() throws InterruptedException {
-    if (_periodicRebalancerFutureTasks != null) {
-      _periodicRebalancerFutureTasks.cancel(false);
-    }
     _periodicalRebalanceExecutor.shutdown();
     stopPeriodRebalance();
     shutdownOnDemandTimer();
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -34,6 +34,7 @@ import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
@@ -172,12 +173,15 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
   private boolean _inMaintenanceMode;
 
   /**
-   * The timer that can periodically run the rebalancing pipeline. The timer will start if there is
-   * one resource group has the config to use the timer.
+   * The executors that can periodically run the rebalancing pipeline. A
+   * SingleThreadScheduledExecutor will start if there is one resource group has the config to do
+   * periodically rebalance.
    */
-  Timer _periodicalRebalanceTimer = null;
+  private static final ScheduledExecutorService _periodicalRebalanceExecutor =
+      Executors.newSingleThreadScheduledExecutor();
+  private ScheduledFuture _periodicRebalancerFutureTasks = null;
   long _timerPeriod = Long.MAX_VALUE;
-
+  private final Object _lock = new Object();
 
   /**
    * The timer that triggers the on-demand rebalance pipeline.
@@ -333,15 +337,19 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
   void startPeriodRebalance(long period, HelixManager manager) {
     if (period != _timerPeriod) {
       logger.info("Controller starting periodical rebalance timer at period " + period);
-      if (_periodicalRebalanceTimer != null) {
-        _periodicalRebalanceTimer.cancel();
+      ScheduledFuture lastScheduledFuture = null;
+      synchronized (_lock) {
+        if (_periodicRebalancerFutureTasks!=null && !_periodicRebalancerFutureTasks.isCancelled()) {
+          lastScheduledFuture = _periodicRebalancerFutureTasks;
+        }
+        _timerPeriod = period;
+        _periodicRebalancerFutureTasks = _periodicalRebalanceExecutor
+            .scheduleAtFixedRate(new RebalanceTask(manager, ClusterEventType.PeriodicalRebalance),
+                _timerPeriod, _timerPeriod, TimeUnit.MILLISECONDS);
       }
-      _periodicalRebalanceTimer =
-          new Timer("GenericHelixController_" + _clusterName + "_periodical_Timer", true);
-      _timerPeriod = period;
-      _periodicalRebalanceTimer
-          .scheduleAtFixedRate(new RebalanceTask(manager, ClusterEventType.PeriodicalRebalance),
-              _timerPeriod, _timerPeriod);
+      if (lastScheduledFuture != null) {
+        lastScheduledFuture.cancel(true /* mayInterruptIfRunning */);
+      }
     } else {
       logger.info("Controller already has periodical rebalance timer at period " + _timerPeriod);
     }
@@ -352,11 +360,11 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
    */
   void stopPeriodRebalance() {
     logger.info("Controller stopping periodical rebalance timer at period " + _timerPeriod);
-    if (_periodicalRebalanceTimer != null) {
-      _periodicalRebalanceTimer.cancel();
-      _periodicalRebalanceTimer = null;
-      _timerPeriod = Long.MAX_VALUE;
-      logger.info("Controller stopped periodical rebalance timer at period " + _timerPeriod);
+    synchronized (_lock) {
+      if (_periodicRebalancerFutureTasks != null && !_periodicRebalancerFutureTasks.isCancelled()) {
+        _periodicRebalancerFutureTasks.cancel(true /* mayInterruptIfRunning */);
+        _timerPeriod = Long.MAX_VALUE;
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -172,11 +173,15 @@
...
+   * The executors that can periodically run the rebalancing pipeline. A
+   * SingleThreadScheduledExecutor will start if there is one resource group has the config to do
+   * periodically rebalance.
    */
-  Timer _periodicalRebalanceTimer = null;
+  private static final ScheduledExecutorService _periodicalRebalanceExecutor =
+      Executors.newSingleThreadScheduledExecutor();
+  private ScheduledFuture _periodicRebalancerFutureTasks = null;
   long _timerPeriod = Long.MAX_VALUE;
+  private final Object _lock = new Object();
```
jiajunwang, line 184: Just lock the _periodicalRebalanceExecutor?
&
Have you tried the atomic reference solution that we discussed? I think that one is easier.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -333,15 +338,19 @@ private void forceRebalance(HelixManager manager, ClusterEventType eventType) {
   void startPeriodRebalance(long period, HelixManager manager) {
     if (period != _timerPeriod) {
       logger.info("Controller starting periodical rebalance timer at period " + period);
-      if (_periodicalRebalanceTimer != null) {
-        _periodicalRebalanceTimer.cancel();
+      ScheduledFuture lastScheduledFuture = null;
+      synchronized (_lock) {
+        if (_periodicRebalancerFutureTasks!=null && !_periodicRebalancerFutureTasks.isCancelled()) {
```
jiajunwang, line 343: Is this check necessary?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -333,15 +338,19 @@ private void forceRebalance(HelixManager manager, ClusterEventType eventType) {
   void startPeriodRebalance(long period, HelixManager manager) {
     if (period != _timerPeriod) {
```
jiajunwang, line 339: I think this check needs to be included in the critical section too.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -333,15 +338,19 @@ private void forceRebalance(HelixManager manager, ClusterEventType eventType) {
...
+        if (_periodicRebalancerFutureTasks!=null && !_periodicRebalancerFutureTasks.isCancelled()) {
+          lastScheduledFuture = _periodicRebalancerFutureTasks;
+        }
+        _timerPeriod = period;
+        _periodicRebalancerFutureTasks = _periodicalRebalanceExecutor
+            .scheduleAtFixedRate(new RebalanceTask(manager, ClusterEventType.PeriodicalRebalance),
+                _timerPeriod, _timerPeriod, TimeUnit.MILLISECONDS);
+      }
+      if (lastScheduledFuture != null) {
+        lastScheduledFuture.cancel(true /* mayInterruptIfRunning */);
```
jiajunwang, line 352: This changes the logic. Keep it as it was, please.
Interrupt it may cause some unexpected result. One example, the newly scheduled rebalance is scheduled with a delay. If you cancel the previous one by interrupt, then there might be a very long period that no rebalance is triggered.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -352,11 +361,11 @@ void startPeriodRebalance(long period, HelixManager manager) {
...
   void stopPeriodRebalance() {
     logger.info("Controller stopping periodical rebalance timer at period " + _timerPeriod);
-    if (_periodicalRebalanceTimer != null) {
-      _periodicalRebalanceTimer.cancel();
-      _periodicalRebalanceTimer = null;
-      _timerPeriod = Long.MAX_VALUE;
-      logger.info("Controller stopped periodical rebalance timer at period " + _timerPeriod);
+    synchronized (_lock) {
+      if (_periodicRebalancerFutureTasks != null && !_periodicRebalancerFutureTasks.isCancelled()) {
+        _periodicRebalancerFutureTasks.cancel(true /* mayInterruptIfRunning */);
```
jiajunwang, line 366: Same here.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1299,6 +1308,10 @@ protected void checkLiveInstancesObservation(List<LiveInstance> liveInstances,
   }
 
   public void shutdown() throws InterruptedException {
+    if (_periodicRebalancerFutureTasks != null) {
+      _periodicRebalancerFutureTasks.cancel(false);
+    }
```
jiajunwang, line 1313: Why need to cancel the task while stopPeriodRebalance is called here?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -172,11 +173,15 @@
   private boolean _inMaintenanceMode;
 
   /**
-   * The timer that can periodically run the rebalancing pipeline. The timer will start if there is
-   * one resource group has the config to use the timer.
+   * The executors that can periodically run the rebalancing pipeline. A
+   * SingleThreadScheduledExecutor will start if there is one resource group has the config to do
+   * periodically rebalance.
```
kaisun2000, line 178: We totally only have one SingleThreadScheduledExecutor at most for each genericHelixController object, right? This sounds like per resource group has one. Can we make it clear the life cycle of SingleThreadScheduledExecutor is the same as genericHelixController?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1299,6 +1308,10 @@ protected void checkLiveInstancesObservation(List<LiveInstance> liveInstances,
   }
 
   public void shutdown() throws InterruptedException {
+    if (_periodicRebalancerFutureTasks != null) {
+      _periodicRebalancerFutureTasks.cancel(false);
+    }
+    _periodicalRebalanceExecutor.shutdown();
```
kaisun2000, line 1314: Does this shutdown kill the thread immediately? I think not. 

Shall we provide an shutDownNow which kill the thread immediately? That would be useful for tests.

kaisun2000:
One more thing I forget to mention before is this:

The current leaking is caused that two events are push to TaskRebalance thread and HelixRebalance thread at the same time. 
For this kind of changing the config time period change, in fact, we don't need to put two events to two different thread running task or helix separately. 

All the above is under assumption that periodic rebalance triggering thread (previous timer, now the  is shared by task and now _periodicalRebalanceExecutor. Is there a possibility later that we will use two different (timer)?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -172,11 +173,15 @@
...
-   * The timer that can periodically run the rebalancing pipeline. The timer will start if there is
-   * one resource group has the config to use the timer.
+   * The executors that can periodically run the rebalancing pipeline. A
+   * SingleThreadScheduledExecutor will start if there is one resource group has the config to do
+   * periodically rebalance.
    */
-  Timer _periodicalRebalanceTimer = null;
+  private static final ScheduledExecutorService _periodicalRebalanceExecutor =
+      Executors.newSingleThreadScheduledExecutor();
+  private ScheduledFuture _periodicRebalancerFutureTasks = null;
```
lei-xia, line 182: _periodicRebalancerFutureTasks -> _periodicRebalancerFutureTask?  Should there always be only 1 (periodical) task?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: Yea it does not kill the thread immediately. There is a ScheduledExecutorService.shutDownNow() that interrupt threads and we could use that in test I guess.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: Yea you are correct. shutDown will cancel tasks later. Updated.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: I think including this check in won't have any functional different compared to not including it. For case like 2 thread change value from v0->v1 and v0->v2 at the same time, the result is indeterministic in either way.  I prefer critical section like this since it excludes 2 log IOs.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
kaisun2000: `   _asyncTasksThreadPool.shutdownNow();
    try {
      _asyncTasksThreadPool.awaitTermination(EVENT_THREAD_JOIN_TIMEOUT, TimeUnit.MILLISECONDS);
    } catch (InterruptedException ex) {
      logger.warn("Timeout when terminating async tasks. Some async tasks are still executing.");
    }
`
see the above. Can we do something similar like the above?

xyuanlu: shutdownNow cancels already scheduled TimerTask and interrupt running TimerTask. For this  _periodicalRebalanceExecutor, scheduled TimerTask is canceled in stopPeriodRebalance() and if there is running one, it is interrupted in shutdownPipeline();

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
mgao0: Second JJ's comment. IMO since the set of this value is inside critical section, the read should be included too. Theoretically you could get a bad value.

xyuanlu: TFTR. I think you are referring to situations like two threads both try to change period to a same value. including this compare in the inside critical section or not won't have different result. 
For the case when we change period to different values, the result is indeterminable in either way. 
Please correct m if I am wrong. I think in conclusion, having the compare in or out of critical section will have the same behavior. So I prefer smallest  critical section in this case, especially we have two IOs here.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -172,13 +173,15 @@
...
   /**
-   * The timer that can periodically run the rebalancing pipeline. The timer will start if there is
-   * one resource group has the config to use the timer.
+   * The executors that can periodically run the rebalancing pipeline. A
+   * SingleThreadScheduledExecutor will start if there is resource group that has the config to do
+   * periodically rebalance.
    */
-  Timer _periodicalRebalanceTimer = null;
+  private static final ScheduledExecutorService _periodicalRebalanceExecutor =
+      Executors.newSingleThreadScheduledExecutor();
```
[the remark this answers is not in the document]
xyuanlu: Good suggestion. Will update.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: Please correct me if I am wrong. In my understanding, shutdownPipeline() guarantees no event processor thread is alive anymore.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: Yea no need to check here. Updated.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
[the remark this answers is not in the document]
xyuanlu: I think this is a valid suggestion. However, I did not find setRemoveOnCancelPolicy() for interface ScheduledExecutorService  we are using here..

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
mgao0: Yes shutdownPipeline() does shut down the event process thread, but when you call shutdownNow, it's the thread inside the executor is shut down. I think they are different threads.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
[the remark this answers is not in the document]
mgao0: This method is only available in ScheduledThreadPoolExecutor. Is it easy to switch to ScheduledThreadPoolExecutor?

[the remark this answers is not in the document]
xyuanlu: Thanks Molly! We had an offline discussion.  Since cancelled tasks will be removed from the work queue when delay elapses, and there is no setRemoveOnCancelPolicy() for interface ScheduledExecutorService. We think for our current workload, it is ok to keep the way it is now.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1307,6 +1315,11 @@ public void shutdown() throws InterruptedException {
     logger.info("Shutting down {} pipeline", Pipeline.Type.TASK.name());
     shutdownPipeline(_taskEventThread, _taskEventQueue);
 
+    if (!_periodicalRebalanceExecutor
```
dasahcc, line 1318: Could supurious wake up can impact this?  I am not sure why we would like to wait here? Are we worrying about the partial task running? If it wakes up and return false, it will kill thread immediately. There is no difference as we kill it at first beginning.

xyuanlu: TFTR. 

Do you think it's better to do a _periodicalRebalanceExecutor.shutdownNow() at beginning and then check if no thread is running? Or just do _periodicalRebalanceExecutor.shutdownNow() at beginning is enough?

jiajunwang: I think it is fine. But it would be easier for the reader to understand if you put all the related logic together.

jiajunwang (approved):
LGTM

Please ensure the remaining comments are addressed.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
xyuanlu: TFTR. Updated.

xyuanlu:
My PR is ready to be merged in. Approved by @jiajunwang 

Final commit message:
Fix Periodic rebalancer Timer leak
In current startPeriodRebalance, two thread may interference with each other. This may result in one timer got canceled twice, two timers are created with one timer leaked. This PR changes Timer to use a SingleThreadScheduledExecutor and adds a synchronized block in start/stop PeriodRebalance.

PR #1460, opened 2020-10-12
Silence Stacktrace from Stale Reply Messages

### LINKED ISSUE #1459: Stale Reply Messages Raised as Exceptions
### Describe the bug
With the current approach of AsyncCallbackService, if a reply message is stale and doesn't have a corresponding callback, an exception will be raised when creating a handler for it. 

### To Reproduce
Reconnect a participant with stale reply messages. 

### Expected behavior
Helix handling it gracefully without raising exceptions.

NealSun96:
### Issues

Fixes #1459 

### Description

If a reply message doesn't have a corresponding callback, AsyncCallbackService will raise an exception when creating a handler for it. Since stale reply messages may happen during participant reconnects, some customers prefer not seeing too verbose error messages since they don't require any actions of debugging. 

This PR adds a less verbose (no stacktrace) HelixException and raises it for this specific case. It may also be used in other occasions if necessary.  

### Tests

```
[ERROR] Tests run: 1213, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 4,830.531 s <<< FAILURE! - in TestSuite
[ERROR] testPeriodicRefresh(org.apache.helix.integration.spectator.TestRoutingTableProviderPeriodicRefresh)  Time elapsed: 2.031 s  <<< FAILURE!

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 1213, Failures: 1, Errors: 0, Skipped: 1
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:20 h
[INFO] Finished at: 2020-10-12T16:36:07-07:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-common/src/main/java/org/apache/helix/HelixException.java
+++ b/helix-common/src/main/java/org/apache/helix/HelixException.java
@@ -30,6 +30,16 @@ public class HelixException extends RuntimeException {
     super(message);
   }
 
+  /**
+   * Create a HelixException that can optionally turn off stack trace. Its other characteristics are
+   * the same as a HelixException with a message.
+   * @param message the detail message
+   * @param writableStackTrace whether or not the stack trace should be writable
+   */
+  public HelixException(String message, boolean writableStackTrace) {
+    super(message, null, false, writableStackTrace);
+  }
+
   public HelixException(Throwable cause) {
     super(cause);
   }
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/AsyncCallbackService.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/AsyncCallbackService.java
@@ -69,7 +69,7 @@ public class AsyncCallbackService implements MultiTypeMessageHandlerFactory {
               + " does not have correponding callback. Probably timed out already. Correlation id: "
               + correlationId;
       _logger.error(errorMsg);
-      throw new HelixException(errorMsg);
+      throw new HelixException(errorMsg, false);
     }
     _logger.info("Verified reply message " + message.getMsgId() + " correlation:" + correlationId);
   }


helix-core/src/main/java/org/apache/helix/messaging/handling/AsyncCallbackService.java
```diff
@@ -113,12 +104,19 @@ public HelixTaskResult handleMessage() throws InterruptedException {
...
+          if (callback.isDone()) {
+            _logger.info("Removing finished callback, correlationid:" + _correlationId);
+            _callbackMap.remove(_correlationId);
+          }
         }
+      } else {
+        String msg = "Message " + _message.getMsgId()
+            + " does not have correponding callback. Probably timed out already. Correlation id: "
+            + _correlationId;
+        _logger.warn(msg);
```
zhangmeng916, line 119: A question regarding the description, if the error is just logged here, meaning the users won't really know there is a mismatch. If the callback is not done, they won't be able to distinguish between this error and other errors. Is this the expected behavior?

NealSun96: You mean they wouldn't be able to distinguish "stale messages" vs "messages that don't have callbacks set up correctly"? I think these cases are logically the same and can't be easily distinguished code-level. If we are loosening up on one case, it's going to affect the other. 

On the other hand, callback registration is done in our code and naturally there should be a callback before sending messages that require replies. I imagine the previous check was there for cautionary purpose, not that messages are commonly created with incorrect correlation ids.

zhangmeng916: Actually I meant this missing ID exception and exceptions in the later functions in onReply. Do we need to consider distinguishing between them? like in your test `Assert.assertFalse(callback.isDone())`. Any other errors may happen in middle?

NealSun96: I'm not sure I follow: this is not a "try" block, this is an "if" block. There's no exception handling here.

zhangmeng916: All i would like to ask is whether removing the previous thrown exception will cause any information lost. If it's ok, then I'm fine.

NealSun96: I see what you mean now: you're saying in both cases, callback.isDone() == false, and it can't be distinguished whether the callback failed or there's no callback registered. 

If the callback fails due to some exception like you said, the task result will reflect that (setException(e), setInterrupted(true)). It's more difficult to tell the "callback not registered" case, but since customers don't want exceptions raised, there isn't any better way here.

zhangmeng916: Yep, that's what I meant. I think it's not a big concern.

zhangmeng916 (approved):

jiajunwang:
Hold on, please.

I'm trying to understand why the Exception is an issue here. If the reply message is stale, meaning we don't have enough information to handle it, fail the handler creation op seems to be a reasonable solution.

If the operation is not failed out, then the handling will be finished with a successful result. It could be misleading.

This basically means bad clients can send an invalid message response to the host and DDoS it. Since the host cannot reject any messages, it will allocate many resources. And this is concerning.

helix-core/src/main/java/org/apache/helix/messaging/handling/AsyncCallbackService.java
```diff
@@ -62,15 +62,6 @@ void verifyMessage(Message message) {
       throw new HelixException(errorMsg);
     }
```
jiajunwang, line 64: Alternatively, if the exception is not graceful, we can return null when createHandler() is triggered. This is much better than returning a handler that does nothing (but still consume system resources).
Of course, the callers of createHandler() need to be improved to tolerate the null return value.

jiajunwang (changes requested):

helix-core/src/main/java/org/apache/helix/messaging/handling/AsyncCallbackService.java
NealSun96: You brought up a very good point, although for our scenario, ZK read probably fails before an effective DOS. I understand this reminds you of some TCP DOS. :) 

If you could check out `HelixTaskExecutor`, at around line 900, you can see how we handle null return value for `createHandler()`. The current approach is to skip the message without cleaning it up. I could explore on changing this current approach (because we need to clean up those reply messages), but such a change will impact other message handlings as well (such as state transition messages).

kaisun2000:
As discussed, the main concern is that current way, additional thread created.

NealSun96:
Update: after group discussion, changed the strategy to silencing the exception stacktrace instead of changing behavior. This still satisfies customer needs.

helix-common/src/main/java/org/apache/helix/HelixException.java
```diff
@@ -30,6 +30,16 @@ public HelixException(String message) {
     super(message);
   }
 
+  /**
+   * Create a HelixException that can optionally turn off stack trace. Its other characteristics are
+   * the same as a HelixException with a message.
+   * @param message the detail message
+   * @param writableStackTrace whether or not the stack trace should be writable
+   */
+  public HelixException(String message, boolean writableStackTrace) {
```
kaisun2000, line 39: So RuntimeException can be configured not to have stack trace out.

kaisun2000 (approved):

helix-common/src/main/java/org/apache/helix/HelixException.java
NealSun96: Yes sir

NealSun96:
Final commit message:
## Silence Stacktrace from Stale Reply Messages  ##
This PR adds a less verbose (no stacktrace) HelixException and raises it for stale reply message cases.

Issue #1250, opened 2020-08-11
Fix flaky test TestRoutingTableProviderPeriodicRefresh


kaisun2000:
- The test purely relies on timing. Frequent source of failure. 

- The second problem is that when failed, it would leak threads in RoutingProvider which would clog the test output. (This is fixed by adding tearing down logic to @afterClass.

LOG:

>[ERROR] testPeriodicRefresh(org.apache.helix.integration.spectator.TestRoutingTableProviderPeriodicRefresh)  Time elapsed: 2.019 s  <<< FAILURE!
java.lang.AssertionError: expected:<7> but was:<6>
        at org.apache.helix.integration.spectator.TestRoutingTableProviderPeriodicRefresh.testPeriodicRefresh(TestRoutingTableProviderPeriodicRefresh.java:214)

Code:

```
public void testPeriodicRefresh() throws InterruptedException {
    // This test purely relies on sleep(). Frequent failure source, need refactor
    // Wait so that initial refreshes finish (not triggered by periodic refresh timer)
    Thread.sleep(1000L);

    // Test short refresh
    int prevRefreshCount = _routingTableProvider.getRefreshCount();
```

jiajunwang:
Close test unstable tickets since we have an automatic tracking mechanism https://github.com/apache/helix/pull/1757 now for tracking the most recent test issues.

PR #1464, opened 2020-10-13
Fix MaxCapacityUsageGauge value not updated

### LINKED ISSUE #1463: MaxCapacityUsageGauge value is always 0
### Describe the bug
From metrics reporting result in mbean server, we find that the MaxCapacityUsageGauge is always 0.

### To Reproduce
Use jconsole to check the MaxCapacityUsageGauge attribute metric, no matter what capacity value we change, the value is not updated.

### Expected behavior
MaxCapacityUsageGauge should be correctly updated according to the capacity configs.

### Additional context
The root cause is, the resourceToMonitor map is empty because the resources are filtered out.

pkuwm:
### Issues

Fixes #1463 

### Description

MaxCapacityUsageGauge value not updated because the resourcesToMonitor map is empty.

This PR fixes the bug and also add an integration test to protect the metrics reporting logic.

image

### Tests

testInstanceCapacityMetrics()

```
[INFO] Results:
[INFO]
[ERROR] Failures:
[INFO]
[ERROR] Tests run: 1214, Failures: 1, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:23 h
[INFO] Finished at: 2020-10-13T01:54:46-07:00
```

The failed test is a known one: https://github.com/apache/helix/issues/1250
@kaisun2000 is working on it.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/WagedValidationUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/WagedValidationUtil.java
@@ -24,7 +24,9 @@ import java.util.List;
 import java.util.Map;
 
 import org.apache.helix.HelixException;
+import org.apache.helix.controller.rebalancer.waged.WagedRebalancer;
 import org.apache.helix.model.ClusterConfig;
+import org.apache.helix.model.IdealState;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.ResourceConfig;
 
@@ -87,5 +89,17 @@ public class WagedValidationUtil {
           partitionCapacity.toString()));
     }
     return partitionCapacity;
+  }
+
+  /**
+   * Checks whether or not a resource has enabled WAGED rebalancer.
+   *
+   * @param idealState {@code IdealState} of the resource being checked.
+   * @return {@code true} if WAGED is enabled; otherwise, {@code false}.
+   */
+  public static boolean isWagedEnabled(IdealState idealState) {
+    return idealState != null
+        && idealState.getRebalanceMode().equals(IdealState.RebalanceMode.FULL_AUTO)
+        && WagedRebalancer.class.getName().equals(idealState.getRebalancerClassName());
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
@@ -34,6 +34,7 @@ import org.apache.helix.controller.dataproviders.WorkflowControllerDataProvider;
 import org.apache.helix.controller.pipeline.AbstractBaseStage;
 import org.apache.helix.controller.pipeline.StageException;
 import org.apache.helix.controller.rebalancer.util.ResourceUsageCalculator;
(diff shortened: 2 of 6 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/WagedValidationUtil.java
```diff
@@ -88,4 +90,16 @@
...
     return partitionCapacity;
   }
+
+  /**
+   * Checks whether or not a resource has enabled WAGED rebalancer.
+   *
+   * @param idealState {@code IdealState} of the resource being checked.
+   * @return {@code true} if WAGED is enabled; otherwise, {@code false}.
+   */
+  public static boolean isWagedEnabled(IdealState idealState) {
```
NealSun96, line 100: This could be used in `BestPossibleStateCalcStage.computeResourceBestPossibleStateWithWagedRebalancer` as well. Could you check for other similar usages? Replacing those logic with this function is a step up for cleanness.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -283,10 +284,10 @@ private void reportInstanceCapacityMetrics(ClusterStatusMonitor clusterStatusMon
     asyncExecute(dataProvider.getAsyncTasksThreadPool(), () -> {
       try {
         // ResourceToRebalance map also has resources from current states.
-        // Only use the resources in ideal states to parse all replicas.
+        // Only use the resources in ideal states that enable WAGED to parse all replicas.
         Map<String, IdealState> idealStateMap = dataProvider.getIdealStates();
         Map<String, Resource> resourceToMonitorMap = resourceMap.entrySet().stream()
-            .filter(idealStateMap::containsKey)
+            .filter(entry -> WagedValidationUtil.isWagedEnabled(idealStateMap.get(entry.getKey())))
```
NealSun96, line 290: Is this a behavior change (all resources -> waged only)?

NealSun96 (changes requested):

helix-core/src/main/java/org/apache/helix/controller/rebalancer/util/WagedValidationUtil.java
pkuwm: Ideally, replacing those places with this API should be done in another PR. Since you've pointed out that place, I just made the change. If there are more places, let's make the change in another PR.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
pkuwm: Ideally, we would report every resource. But currently for those non WAGED resources, validation won't pass and `HelixException` would be thrown. So we'll skip non WAGED resources here.

NealSun96 (approved):

pkuwm:
This PR is ready to be merged, approved by @NealSun96 @jiajunwang 

MaxCapacityUsageGauge value not updated because the resourcesToMonitor map is empty.

This commit fixes the bug and also adds an integration test to protect the metrics reporting logic.

PR #1468, opened 2020-10-15
Implement deleteTask API

### LINKED ISSUE #1467: Delete a tasks from the running (IN-PROGRESS) jobs or the jobs that have not started yet
There has been a request to implement a feature that gives us the ability to dynamically delete the tasks from a running job or a job that has not been started

The API can be added to the TaskDriver which allows users to delete a task while the job is being run or not started. On the controller side, the controller should be able to drop the current state and remove the task from the job context.

alirezazamani:
### Issues

Fixes #1467  

### Description

In this PR, the changes that needed to delete an existing task from a general job have been added to TaskDriver.
Also, the changes that needed on the controller side to drop the task from the instances and removing the task from the
context has been implemented.

### Tests

TestAddDeleteTask

```
[INFO] Tests run: 1242, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,182.218 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1242, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:26 h
[INFO] Finished at: 2020-10-31T15:48:43-07:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
+++ b/helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
@@ -168,8 +168,10 @@ public class PropertyPathBuilder {
     // RESOURCE
     addEntry(PropertyType.WORKFLOWCONTEXT, 2,
         "/{clusterName}/PROPERTYSTORE/TaskRebalancer/{workflowName}/Context"); // Old
-    // WorkflowContext
-    // path
+
+    // TODO: These are the current task framework related paths. In the future, if we decide to use
+    // a different structure such as a non-flatten ZNode structure, these paths need to be changed
+    // accordingly.
     addEntry(PropertyType.TASK_CONFIG_ROOT, 1, "/{clusterName}/CONFIGS/RESOURCE");
     addEntry(PropertyType.WORKFLOW_CONFIG, 2, "/{clusterName}/CONFIGS/RESOURCE/{workflowName}");
     addEntry(PropertyType.JOB_CONFIG, 3,
--- a/helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
+++ b/helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
@@ -37,6 +37,7 @@ import org.apache.helix.model.Message;
 import org.apache.helix.model.PauseSignal;
 import org.apache.helix.model.StateModelDefinition;
 import org.apache.helix.model.StatusUpdate;
+import org.apache.helix.task.TaskConstants;
 import org.apache.helix.task.WorkflowContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -175,11 +176,11 @@ public class PropertyPathBuilder {
     addEntry(PropertyType.JOB_CONFIG, 4,
         "/{clusterName}/CONFIGS/TASK/{workflowName}/{jobName}/{jobName}");
     addEntry(PropertyType.TASK_CONTEXT_ROOT, 1,
-        "/{clusterName}/PROPERTYSTORE/TaskFrameworkContext");
-    addEntry(PropertyType.WORKFLOW_CONTEXT, 2,
-        "/{clusterName}/PROPERTYSTORE/TaskFrameworkContext/{workflowName}/Context");
-    addEntry(PropertyType.JOB_CONTEXT, 3,
-        "/{clusterName}/PROPERTYSTORE/TaskFrameworkContext/{workflowName}/{jobName}/Context");
+        "/{clusterName}/PROPERTYSTORE" + TaskConstants.REBALANCER_CONTEXT_ROOT);
+    addEntry(PropertyType.WORKFLOW_CONTEXT, 2, "/{clusterName}/PROPERTYSTORE"
+        + TaskConstants.REBALANCER_CONTEXT_ROOT + "/{workflowName}/Context");
+    addEntry(PropertyType.JOB_CONTEXT, 3, "/{clusterName}/PROPERTYSTORE"
+        + TaskConstants.REBALANCER_CONTEXT_ROOT + "/{workflowName}" + "_" + "{jobName}/Context");
   }
   static Pattern pattern = Pattern.compile("(\\{.+?\\})");
 
--- a/helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
+++ b/helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
@@ -170,11 +170,10 @@ public class PropertyPathBuilder {
         "/{clusterName}/PROPERTYSTORE/TaskRebalancer/{workflowName}/Context"); // Old
     // WorkflowContext
     // path
-    addEntry(PropertyType.TASK_CONFIG_ROOT, 1, "/{clusterName}/CONFIGS/TASK");
-    addEntry(PropertyType.WORKFLOW_CONFIG, 3,
-        "/{clusterName}/CONFIGS/TASK/{workflowName}/{workflowName}");
-    addEntry(PropertyType.JOB_CONFIG, 4,
-        "/{clusterName}/CONFIGS/TASK/{workflowName}/{jobName}/{jobName}");
+    addEntry(PropertyType.TASK_CONFIG_ROOT, 1, "/{clusterName}/CONFIGS/RESOURCE");
+    addEntry(PropertyType.WORKFLOW_CONFIG, 2, "/{clusterName}/CONFIGS/RESOURCE/{workflowName}");
+    addEntry(PropertyType.JOB_CONFIG, 3,
+        "/{clusterName}/CONFIGS/RESOURCE/{workflowName}" + "_" + "{jobName}");
     addEntry(PropertyType.TASK_CONTEXT_ROOT, 1,
         "/{clusterName}/PROPERTYSTORE" + TaskConstants.REBALANCER_CONTEXT_ROOT);
     addEntry(PropertyType.WORKFLOW_CONTEXT, 2, "/{clusterName}/PROPERTYSTORE"
--- a/helix-core/src/main/java/org/apache/helix/task/FixedTargetTaskAssignmentCalculator.java
+++ b/helix-core/src/main/java/org/apache/helix/task/FixedTargetTaskAssignmentCalculator.java
@@ -22,6 +22,7 @@ package org.apache.helix.task;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -121,6 +122,17 @@ public class FixedTargetTaskAssignmentCalculator extends TaskAssignmentCalculato
           .addAll(getPartitionsForTargetPartition(targetPartition, currentTargets, taskCtx));
     }
     return taskPartitions;
+  }
+
+  @Override
+  public Set<Integer> getRemovedPartitions(JobConfig jobConfig, JobContext jobContext, Set<Integer> allPartitions) {
+    Set<Integer> deletedPartitions = new HashSet<>();
+    for (Integer partition : jobContext.getPartitionSet()) {
+      if (!allPartitions.contains(partition)) {
+        deletedPartitions.add(partition);
+      }
+    }
+    return deletedPartitions;
   }
 
   private static List<Integer> getPartitionsForTargetPartition(String targetPartition,
--- a/helix-core/src/main/java/org/apache/helix/task/FixedTargetTaskAssignmentCalculator.java
+++ b/helix-core/src/main/java/org/apache/helix/task/FixedTargetTaskAssignmentCalculator.java
@@ -32,6 +32,7 @@ import java.util.TreeSet;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
+import java.util.stream.Collectors;
 import org.apache.helix.controller.stages.CurrentStateOutput;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.Message;
@@ -126,13 +127,8 @@ public class FixedTargetTaskAssignmentCalculator extends TaskAssignmentCalculato
 
   @Override
   public Set<Integer> getRemovedPartitions(JobConfig jobConfig, JobContext jobContext, Set<Integer> allPartitions) {
-    Set<Integer> deletedPartitions = new HashSet<>();
-    for (Integer partition : jobContext.getPartitionSet()) {
-      if (!allPartitions.contains(partition)) {
-        deletedPartitions.add(partition);
-      }
-    }
-    return deletedPartitions;
+    return jobContext.getPartitionSet().stream().filter(partition -> !allPartitions.contains(partition)).collect(
+        Collectors.toSet());
   }
 
   private static List<Integer> getPartitionsForTargetPartition(String targetPartition,
--- a/helix-core/src/main/java/org/apache/helix/task/GenericTaskAssignmentCalculator.java
+++ b/helix-core/src/main/java/org/apache/helix/task/GenericTaskAssignmentCalculator.java
@@ -22,6 +22,7 @@ package org.apache.helix.task;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -35,6 +36,7 @@ import com.google.common.collect.Maps;
 import org.apache.helix.HelixException;
 import org.apache.helix.controller.stages.CurrentStateOutput;
 import org.apache.helix.model.IdealState;
+import org.apache.helix.model.Partition;
 import org.apache.helix.model.ResourceAssignment;
 import org.apache.helix.util.JenkinsHash;
 import org.slf4j.Logger;
@@ -93,6 +95,20 @@ public class GenericTaskAssignmentCalculator extends TaskAssignmentCalculator {
     List<String> allNodes = Lists.newArrayList(instances);
     ConsistentHashingPlacement placement = new ConsistentHashingPlacement(allNodes);
     return placement.computeMapping(jobCfg, jobContext, partitionNums, resourceId);
+  }
+
+  @Override
+  public Set<Integer> getRemovedPartitions(JobConfig jobConfig, JobContext jobContext, Set<Integer> allPartitions) {
+    // Get all partitions existed in the context
+    Set<Integer> deletedPartitions = new HashSet<>();
+    // Check whether the tasks have been deleted from jobConfig
+    for (Integer partition : jobContext.getPartitionSet()) {
+      String partitionID = jobContext.getTaskIdForPartition(partition);
+      if (!jobConfig.getTaskConfigMap().containsKey(partitionID)) {
+        deletedPartitions.add(partition);
+      }
+    }
+    return deletedPartitions;
   }
 
   private class ConsistentHashingPlacement {
--- a/helix-core/src/main/java/org/apache/helix/task/GenericTaskAssignmentCalculator.java
+++ b/helix-core/src/main/java/org/apache/helix/task/GenericTaskAssignmentCalculator.java
@@ -97,20 +97,6 @@ public class GenericTaskAssignmentCalculator extends TaskAssignmentCalculator {
     return placement.computeMapping(jobCfg, jobContext, partitionNums, resourceId);
   }
 
-  @Override
-  public Set<Integer> getRemovedPartitions(JobConfig jobConfig, JobContext jobContext, Set<Integer> allPartitions) {
-    // Get all partitions existed in the context
-    Set<Integer> deletedPartitions = new HashSet<>();
-    // Check whether the tasks have been deleted from jobConfig
-    for (Integer partition : jobContext.getPartitionSet()) {
-      String partitionID = jobContext.getTaskIdForPartition(partition);
-      if (!jobConfig.getTaskConfigMap().containsKey(partitionID)) {
-        deletedPartitions.add(partition);
-      }
-    }
-    return deletedPartitions;
-  }
-
   private class ConsistentHashingPlacement {
     private JenkinsHash _hashFunction;
     private ConsistentHashSelector _selector;
--- a/helix-core/src/main/java/org/apache/helix/task/JobContext.java
+++ b/helix-core/src/main/java/org/apache/helix/task/JobContext.java
@@ -316,9 +316,9 @@ public class JobContext extends HelixProperty {
 
   /**
    * Remove the task from the map field of job context.
-   * @param taskNumber
+   * @param partitionSeqNumber
    */
-  public void removeTask(int taskNumber) {
-    _record.getMapFields().remove(String.valueOf(taskNumber));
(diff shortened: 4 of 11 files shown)


helix-core/src/test/java/org/apache/helix/integration/task/TestAddDeleteTask.java
```diff
@@ -89,9 +100,16 @@ public void testAddJobMissing() throws Exception {
     } catch (IllegalArgumentException e) {
       // Helix Exception is expected because job config is missing
     }
+
+    try {
+      _driver.deleteTask(workflowName, jobName, task.getId());
+      Assert.fail("Exception is expected because job config is missing");
+    } catch (IllegalArgumentException e) {
+      // Helix Exception is expected because job config is missing
```
pkuwm, line 108: Helix exception? While illegal argument exception is caught.

alirezazamani: Good catch. Let me fix the comments.

helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
```diff
@@ -169,17 +170,16 @@
         "/{clusterName}/PROPERTYSTORE/TaskRebalancer/{workflowName}/Context"); // Old
     // WorkflowContext
     // path
-    addEntry(PropertyType.TASK_CONFIG_ROOT, 1, "/{clusterName}/CONFIGS/TASK");
-    addEntry(PropertyType.WORKFLOW_CONFIG, 3,
-        "/{clusterName}/CONFIGS/TASK/{workflowName}/{workflowName}");
-    addEntry(PropertyType.JOB_CONFIG, 4,
-        "/{clusterName}/CONFIGS/TASK/{workflowName}/{jobName}/{jobName}");
+    addEntry(PropertyType.TASK_CONFIG_ROOT, 1, "/{clusterName}/CONFIGS/RESOURCE");
```
jiajunwang, line 173: I think we introduced these "new" path patterns for the purpose of refactoring TF structure. While I believe this is still the future design, I agree that we want to make them work first. Can we add a TODO here?

helix-core/src/main/java/org/apache/helix/task/JobContext.java
```diff
@@ -313,4 +313,12 @@ public long getExecutionStartTime() {
     }
     return map;
   }
+
+  /**
+   * Remove the task from the map field of job context.
+   * @param taskNumber
+   */
+  public void removeTask(int taskNumber) {
```
jiajunwang, line 321: nit, taskSeqNumber?

jiajunwang: Moreover, according to the naming convention of this class, I guess we are using "partition" instead of "task".

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -478,4 +482,61 @@ private TaskAssignmentCalculator getAssignmentCalculator(JobConfig jobConfig,
     }
     return new FixedTargetTaskAssignmentCalculator(assignableInstanceManager);
   }
+
+  /**
+   * Find the tasks that have been removed from job config, ass them to tasksToDrop. If task's
```
jiajunwang, line 487: ass => add

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -238,6 +238,10 @@ private ResourceAssignment computeResourceMapping(String jobResource,
 
     updateInstanceToTaskAssignmentsFromContext(jobCtx, currentInstanceToTaskAssignments);
 
+    // Find the tasks that have been removed form the config, add them to TasksToDrop
+    handleDeletedTasks(jobResource, jobCfg, jobCtx, currentInstanceToTaskAssignments, tasksToDrop,
```
jiajunwang, line 242: I think the calculation should be done in the taskAssignmentCal, so that allPartitions won't contains the removed tasks at all. Then the following logic is simple and we only match the current assignment with the expected task list. Anything mismatch, then we add or remove accordingly.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -607,14 +615,101 @@ public void addTask(String workflowName, String jobName, TaskConfig taskConfig,
...
+    Map<String, TaskConfig> allTaskConfigs = jobConfig.getTaskConfigMap();
+    for (Map.Entry<String, TaskConfig> entry : allTaskConfigs.entrySet()) {
+      if (entry.getKey().equals(taskID)) {
+        taskConfig = entry.getValue();
+      }
+    }
+
+    validateConfigsForTaskModifications(workflowName, jobName, taskConfig);
+
+    WorkflowContext workflowContext = getWorkflowContext(workflowName);
```
jiajunwang, line 680: Simple code is what I like : ) It also applies to addTask().

```
if (workflowContext != null && jobContext != null (maybe remove the 2nd condition?)) {
 TaskState jobState = workflowContext.getJobState(nameSpaceJobName);
  if (jobState != null && ILLEGAL_JOB_STATES_FOR_TASK_MODIFICATION.contains(jobState)) {
      throw new HelixException("Job " + nameSpaceJobName
          + " is in illegal state for task deletion. Job State is " + jobState);
  }
}
deleteTaskFromJobConfig(workflowName, jobName, taskID, endTime);
```

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -699,6 +798,73 @@ private void addTaskToJobConfig(String workflowName, String jobName, TaskConfig
...
+   * A helper method which deletes an existing task from the job config and verifies if task is
+   * deleted from the context by the controller.
+   * @param workflowName
+   * @param jobName
+   * @param taskID
+   * @param endTime
+   * @throws InterruptedException
+   * @throws TimeoutException
+   */
+  private void deleteTaskFromJobConfig(String workflowName, String jobName, String taskID,
```
jiajunwang, line 811: I haven't reviewed these code completely. But my guess is that this method is different from addTaskToJobConfig() for only the updater part. All the rest are the same. Could you please merge?

helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
alirezazamani: Added a TODO in case we want to change it in the future. I think for performance reasons, we want to keep the structure as it is.

helix-core/src/main/java/org/apache/helix/task/JobContext.java
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
alirezazamani: Fixed.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
alirezazamani: The issue with moving it as part of taskAssignmentCal.getAllTaskPartitions is that the outcome of that function would have some effect on the job and workflow status. We need to consider all the tasks unless it is completely DROPPED from the current state, then we can make sure that currentState has been cleaned up and it will not cause quota leakage later. So I think it makes more sense to have the task in allPartitions, and when we are sure that currentState is DROPPED then we just remove the task from context. In the future pipeline then it will not have any effect because tasks would not show up in allPartitions.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Very good suggestion. Changed in both add and delete.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: So they seem to be similar but they are different. Aside from the logs and updaters which are different, in one of them we add a task to config and then check if it is being added to context by the controller by polling. In another one, we delete from config and make sure it is being deleted by the controller from the context by polling. Anyways following your comment I did a few simplifications on add API. Thanks.

jiajunwang: Logs can be changed to "failed to update the task from the job...", etc.
The updater can be changed to be a parameter.

As for the additional check you mentioned, is it necessary? Since the updater is done, task modification has to be finished in the ZK. If the rebalancer runs slow, then the timeout will fail the operation even the task has been added/removed correctly.
Even we do want to wait, it is still doable by passing a verifier-style callback and then we can merge the private methods.
Or, alternatively, you can check the jobContext in the upper layer method after calls this generic task modification method. Note that since you are calling the accessor 2 times for the jobcontext and workflowcontext, there is no guarantee that the result is consistent anyway. So moving the jobcontext check outside will not introduce more issues.

alirezazamani: Ok. Yeah, we do want to check before returning the call. However, I think your comment makes sense. Let me try that and see how it goes.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
jiajunwang: High-levelly, my motivation is to simplify the logic here. The dispatcher logic looks like a series of if-else blocks. And there are redundencies and very easy for devs to introduce bugs. I understand it cannot be done in one day. So what I suggested is just want to prevent us adding another if-else block to the already long list.

1. it seems that allPartitions is not used by "addGivenUpPartitions" which is after your new method. So no matter how you modify the allPartitions list, it will impact the existing use cases anyway.
The only exception is the empty target partitions check. If the user removed all the tasks, then I think it is OK we mark this job as failure. Or we just prevent user from deleting all tasks. This is not our target use case.

2. On my understanding, handleDeletedTasks is mainly for 3 tasks.
- Add the deleted tasks to tasksToDrop.
First of all, I do believe we need a generic logic to check for all dropping tasks in a single place. It is a major issue if we have this kind of logic everywhere in different methods. One example of the pain is that you actually need to change the comment of tasksToDrop to fit this new usage.
But that is nice to have for now.
Back to this case, it shall be easy to fill the tasksToDrop list immediately after we have the filtered allPartitions list, I think. Or you can let the taskAssignmentCal to do it. Based on it's name, I guess the original design is to let this class finish such a work.

- Remove the deleted tasks from allPartitions.
As I mentioned, this can be done in the taskAssignmentCal logic.

- Remove the deleted tasks from jobContext.
I haven't fully understand this part. But I don't see there is a way to remove a task from the jobContext before. So I guess the previous assumption is that a task can be cancelled but not removed directly. Then is it safe to allow removing the task now? It seems to trigger race condition. For example, if a task has been running, and it is requested to be removed at the same time. We deleted it from the context directly. What will happen? Is it safer just to cancel it?
I might not understand the workflow here completely. Feel free to set up a meeting so we can discuss about it.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Done. Merged the update part in one method.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
alirezazamani: As we discussed offline, move the detection part in taskAssignmentCalculator. Now it works for targeted jobs as well in case the target DB shrinks. Now the logic is more general.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -478,4 +485,52 @@ private TaskAssignmentCalculator getAssignmentCalculator(JobConfig jobConfig,
...
+              partition, jobName, instance);
+          if (!tasksToDrop.containsKey(instance)) {
+            tasksToDrop.put(instance, new HashSet<>());
+          }
+          tasksToDrop.get(instance).add(partition);
+
+          // If current state or pending message have not been removed yet, we should not
+          // delete the context and leave unclean currentState
+          String pName = pName(jobName, partition);
+          if (currStateOutput.getCurrentState(jobName, new Partition(pName), instance) != null
```
pkuwm, line 520: If `!hasCurrentState` is also added to this if condition, when it is true, we don't need to `new Partition()` for later checks?

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -478,4 +485,52 @@ private TaskAssignmentCalculator getAssignmentCalculator(JobConfig jobConfig,
     }
     return new FixedTargetTaskAssignmentCalculator(assignableInstanceManager);
   }
+
+  /**
+   * Add the removed task to tasksToDrop to drop its current state. If task's currentState and
+   * pending message have been removed, delete the task from job context.
+   * @param jobName
```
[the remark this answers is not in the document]
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
alirezazamani: No. In order to get currentState, we need to create a partition with pName. Later on for the deletion, we just need an integer/partition number to delete it from context.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -238,6 +238,10 @@ private ResourceAssignment computeResourceMapping(String jobResource,
 
     updateInstanceToTaskAssignmentsFromContext(jobCtx, currentInstanceToTaskAssignments);
 
+    // Find the tasks that have been removed form the config, add them to TasksToDrop
```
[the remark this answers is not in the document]
alirezazamani: Fixed.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -478,4 +482,61 @@ private TaskAssignmentCalculator getAssignmentCalculator(JobConfig jobConfig,
...
+      Map<String, SortedSet<Integer>> currentInstanceToTaskAssignments,
+      Map<String, Set<Integer>> tasksToDrop, CurrentStateOutput currStateOutput,
+      Set<Integer> allPartitions) {
+    if (TaskUtil.isGenericTaskJob(jobConfig)) {
+      // Get all partitions existed in the context
+      Set<Integer> contextPartitions = jobContext.getPartitionSet();
+      // Check whether the tasks have been deleted from jobConfig
+      for (Integer partition : contextPartitions) {
+        String partitionID = jobContext.getTaskIdForPartition(partition);
+        if (!jobConfig.getTaskConfigMap().containsKey(partitionID)) {
```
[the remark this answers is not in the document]
alirezazamani: I moved this logic to taskAssignmentCalculator class. However, we cannot avoid some of the loops because we need to get the partitionID using jobContext.getTaskIdForPartition(partition) and find the one which has been removed.
We basically find the partitionID in the context and see if such taskID existed in the config it not now.
```
  @Override
  public Set<Integer> getRemovedPartitions(JobConfig jobConfig, JobContext jobContext, Set<Integer> allPartitions) {
    // Get all partitions existed in the context
    Set<Integer> deletedPartitions = new HashSet<>();
    // Check whether the tasks have been deleted from jobConfig
    for (Integer partition : jobContext.getPartitionSet()) {
      String partitionID = jobContext.getTaskIdForPartition(partition);
      if (!jobConfig.getTaskConfigMap().containsKey(partitionID)) {
        deletedPartitions.add(partition);
      }
    }
    return deletedPartitions;
  }
```

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -478,4 +482,61 @@ private TaskAssignmentCalculator getAssignmentCalculator(JobConfig jobConfig,
...
+          boolean hasCurrentState = false;
+          for (Map.Entry<String, SortedSet<Integer>> instanceToPartitions : currentInstanceToTaskAssignments
```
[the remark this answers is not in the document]
alirezazamani: I think that would make sense. However, I changed the structure of the code. If we assume that there won't be many removed tasks, then my new code would make more sense. In the new structure, if there is no task being removed, we don't loop at all. So overall, if not much task is being removed (which is mostly the case), the current implementation would be faster.

NealSun96 (changes requested):

helix-core/src/main/java/org/apache/helix/task/FixedTargetTaskAssignmentCalculator.java
```diff
@@ -123,6 +124,17 @@ public FixedTargetTaskAssignmentCalculator(AssignableInstanceManager assignableI
...
 
+  @Override
+  public Set<Integer> getRemovedPartitions(JobConfig jobConfig, JobContext jobContext, Set<Integer> allPartitions) {
+    Set<Integer> deletedPartitions = new HashSet<>();
+    for (Integer partition : jobContext.getPartitionSet()) {
+      if (!allPartitions.contains(partition)) {
+        deletedPartitions.add(partition);
+      }
+    }
+    return deletedPartitions;
```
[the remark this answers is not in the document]
alirezazamani: Done.

helix-core/src/main/java/org/apache/helix/task/GenericTaskAssignmentCalculator.java
```diff
@@ -95,6 +97,20 @@
...
+  public Set<Integer> getRemovedPartitions(JobConfig jobConfig, JobContext jobContext, Set<Integer> allPartitions) {
+    // Get all partitions existed in the context
+    Set<Integer> deletedPartitions = new HashSet<>();
+    // Check whether the tasks have been deleted from jobConfig
+    for (Integer partition : jobContext.getPartitionSet()) {
+      String partitionID = jobContext.getTaskIdForPartition(partition);
+      if (!jobConfig.getTaskConfigMap().containsKey(partitionID)) {
+        deletedPartitions.add(partition);
+      }
+    }
```
dasahcc, line 110: We can save the loop operation by set operation as:

Set<Integer> deletedPartitions = new HashSet<>(jobContext.getPartitionSet());
deletedPartitions.removeAll(jobConfig.getTaskConfigMap().keySet());

helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
```diff
@@ -146,6 +147,20 @@ public ThreadCountBasedTaskAssignmentCalculator(TaskAssigner taskAssigner,
     return taskAssignment;
   }
 
+  @Override
+  public Set<Integer> getRemovedPartitions(JobConfig jobConfig, JobContext jobContext, Set<Integer> allPartitions) {
```
dasahcc, line 151: Since this is same implementation as generic one, can we have this implementation as default implement in abstract class? Then if we need difference implementation, let override it.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -585,36 +585,195 @@ public void addTask(String workflowName, String jobName, TaskConfig taskConfig,
...
 
-    TaskState jobState = workflowContext.getJobState(nameSpaceJobName);
+    String taskID = taskConfig.getId();
+    while (System.currentTimeMillis() <= endTime) {
+      jobContext =
+          _accessor.getProperty(_accessor.keyBuilder().jobContextZNode(workflowName, jobName));
+      workflowContext =
+          _accessor.getProperty(_accessor.keyBuilder().workflowContextZNode(workflowName));
+      if (jobContext.getTaskIdPartitionMap().containsKey(taskID)
+          && workflowContext.getJobState(nameSpaceJobName) == TaskState.IN_PROGRESS) {
```
dasahcc, line 638: Please make sure user understand it. If they have the delay scheduling between tasks or if there are dependencies, this add may fail. Also we do not do "rollback" for added tasks.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
```diff
@@ -585,36 +585,195 @@ public void addTask(String workflowName, String jobName, TaskConfig taskConfig,
...
     }
-    addTaskToJobConfig(workflowName, jobName, taskConfig, endTime);
+
+    while (System.currentTimeMillis() <= endTime) {
+      jobContext =
+          _accessor.getProperty(_accessor.keyBuilder().jobContextZNode(workflowName, jobName));
+      workflowContext =
+          _accessor.getProperty(_accessor.keyBuilder().workflowContextZNode(workflowName));
+      if (!jobContext.getTaskIdPartitionMap().containsKey(taskID)
+          && workflowContext.getJobState(nameSpaceJobName) == TaskState.IN_PROGRESS) {
```
dasahcc, line 759: This condition may not be correct. If this is the only task, job can be completed or something.

dasahcc (approved):
Overall logic looks good.

helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
alirezazamani: I think it is safe to keep it as it is because for other methods the same rule is applied. For example, getAllTaskPartiotions are the same in both classes.

helix-core/src/main/java/org/apache/helix/task/GenericTaskAssignmentCalculator.java
alirezazamani: I will sync up with you on this offline.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Yeah, I agree. removed the IN_PROGRESS condition.

helix-core/src/main/java/org/apache/helix/task/TaskDriver.java
alirezazamani: Added some java docs note. Thanks.

helix-core/src/main/java/org/apache/helix/task/GenericTaskAssignmentCalculator.java
alirezazamani: Synced offline. Resolved.

helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
alirezazamani: As we discussed offline, I will address this.

alirezazamani:

Final commit message:

Implement deleteTask API

In this commit, the changes needed to delete an existing
task from a general job have been added to TaskDriver.
Also, the changes that are needed on the controller side to
drop the task from the instances and remove the task from the
context have been implemented.

PR #1470, opened 2020-10-15
ExpiredJob Workaround for Selective Update Race Conditions

### LINKED ISSUE #1469: Task Garbage Collection deletes jobs due to selective update phantom read
### Describe the bug
Due to the way selective update is done, it's possible to read a WorkflowConfig containing more jobs in a DAG than the actual JobConfigs read. That causes Garbage Collection stage to remove the jobs from the DAG as well. 

### To Reproduce
1. Add job 1; 
2. finishes the "listing" operation during selective update; 
3. add job 2 and update WorkflowConfig DAG;
4. finishes the "updating" operation during selective update.
This results in a WorkflowConfig containing a DAG of both job 1 and job 2, but only job 1 is read into the cache. 

### Expected behavior
Task Garbage Collection stage should try not to delete jobs like those before a better solution is implemented.

NealSun96:
### Issues

Fixes #1469 

### Description

Cache selective update is a two step procedure. It first "lists", which gets all the ZNodes, and then it "updates" by comparing ZNode metadata to determine which ZNodes have been changed. Job enqueue is also a two step procedure, where it adds all JobConfigs one by one, and update WorkflowConfig's DAG in the end.  

If selective update "lists" before all JobConfigs are added, and "updates" after the WorkflowConfig is done updating, it's possible to read in a DAG containing more jobs than JobConfigs that are seen by selective update. This causes TaskGarbageCollection to remove the jobs that are "missing" JobConfigs, even though their JobConfigs actually exist. This PR implements a workaround for determining expired jobs while avoiding this case of cache "phantom read": if JobConfig doesn't exist in the cache, check ZK directly.

### Tests

```
[ERROR] Tests run: 1234, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 4,885.942 s <<< FAILURE! - in TestSuite
[ERROR] testPeriodicRefresh(org.apache.helix.integration.spectator.TestRoutingTableProviderPeriodicRefresh)  Time elapsed: 2.022 s  <<< FAILURE!

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[INFO] 
[ERROR] Tests run: 1234, Failures: 1, Errors: 0, Skipped: 1
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:21 h
[INFO] Finished at: 2020-10-21T12:14:46-07:00
```
### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
@@ -81,8 +81,8 @@ public class TaskGarbageCollectionStage extends AbstractAsyncBaseStage {
         if (nextPurgeTime <= currentTime) {
           nextPurgeTime = currentTime + purgeInterval;
           // Find jobs that are ready to be purged
-          Set<String> expiredJobs =
-              TaskUtil.getExpiredJobsFromCache(dataProvider, workflowConfig, workflowContext);
+          Set<String> expiredJobs = TaskUtil
+              .getExpiredJobsFromCache(dataProvider, workflowConfig, workflowContext, manager);
           if (!expiredJobs.isEmpty()) {
             expiredJobsMap.put(workflowConfig.getWorkflowId(), expiredJobs);
           }
--- a/helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
@@ -789,7 +789,7 @@ public class TaskUtil {
    */
   public static Set<String> getExpiredJobsFromCache(
       WorkflowControllerDataProvider workflowControllerDataProvider, WorkflowConfig workflowConfig,
-      WorkflowContext workflowContext) {
+      WorkflowContext workflowContext, HelixManager manager) {
     Set<String> expiredJobs = new HashSet<>();
     Map<String, TaskState> jobStates = workflowContext.getJobStates();
     for (String job : workflowConfig.getJobDag().getAllNodes()) {
@@ -797,6 +797,11 @@ public class TaskUtil {
         continue;
       }
       JobConfig jobConfig = workflowControllerDataProvider.getJobConfig(job);
+      // TODO: Temporary solution for cache selective update race conditions
+      if (jobConfig == null) {
+        jobConfig = TaskUtil.getJobConfig(manager, job);
+      }
+
       JobContext jobContext = workflowControllerDataProvider.getJobContext(job);
       TaskState jobState = jobStates.get(job);
       if (isJobExpired(job, jobConfig, jobContext, jobState)) {
--- a/helix-core/src/test/java/org/apache/helix/task/TestTaskUtil.java
+++ b/helix-core/src/test/java/org/apache/helix/task/TestTaskUtil.java
@@ -98,7 +98,7 @@ public class TestTaskUtil extends TaskTestBase {
     expectedJobs.add(workflowName + "_Job_3");
     Assert.assertEquals(TaskUtil
         .getExpiredJobsFromCache(workflowControllerDataProvider, jobQueue.getWorkflowConfig(),
-            workflowContext), expectedJobs);
+            workflowContext, _manager), expectedJobs);
   }
 
   @Test
@@ -188,7 +188,7 @@ public class TestTaskUtil extends TaskTestBase {
     expectedJobs.add(workflowName + "_Job_8");
     Assert.assertEquals(TaskUtil
         .getExpiredJobsFromCache(workflowControllerDataProvider, workflow.getWorkflowConfig(),
-            workflowContext), expectedJobs);
+            workflowContext, _manager), expectedJobs);
   }
 
   @Test


NealSun96:
Disclaimer: with help from @kaisun2000 , I used the debugger to create a situation outlined in the description, which causes certain job enqueuing tests (such as TestEnqueueJobs) to consistently fail. With this change, these tests consistently pass. The job enqueuing tests were also run for 20 times to verify consistent passing.

alirezazamani (approved):
LGTM

helix-core/src/main/java/org/apache/helix/task/TaskUtil.java
```diff
@@ -789,14 +789,19 @@ protected static boolean removeJobsFromWorkflow(final HelixDataAccessor dataAcce
    */
   public static Set<String> getExpiredJobsFromCache(
       WorkflowControllerDataProvider workflowControllerDataProvider, WorkflowConfig workflowConfig,
-      WorkflowContext workflowContext) {
+      WorkflowContext workflowContext, HelixManager manager) {
```
kaisun2000, line 792: This is a public API change. If there is other place using this API, the compilation may fail.

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
```diff
@@ -81,8 +81,8 @@ public void process(ClusterEvent event) throws Exception {
         if (nextPurgeTime <= currentTime) {
           nextPurgeTime = currentTime + purgeInterval;
           // Find jobs that are ready to be purged
-          Set<String> expiredJobs =
-              TaskUtil.getExpiredJobsFromCache(dataProvider, workflowConfig, workflowContext);
+          Set<String> expiredJobs = TaskUtil
+              .getExpiredJobsFromCache(dataProvider, workflowConfig, workflowContext, manager);
```
kaisun2000, line 85: This is public API change. I know it is added recently, but are we sure this won't break some other code using TaskUtil?

The other way can be simply add a check after line 85 to see if any expiredJobs are in the config and remove them from the expiredJob set.

kaisun2000:
The other thing is that can we run this fix in githut say 20 times to see if it fails any previous sensitive test? The 20 times test would give a signal as to whether the fix would work. Do we feel this is necessary?

helix-core/src/main/java/org/apache/helix/controller/stages/TaskGarbageCollectionStage.java
NealSun96: I see the concern here, but doing it in the way you described is not so clean. We need to check for the case of "JobConfig missing from cache but exist in ZK" for all the expiredJobs, and for each of those cases, we need to apply the "isJobExpired" logic again. There is a lot of logic redundancy here. 

Is that better than changing the signature? I think both ways have minor drawbacks, but I personally think changing the signature is fine.

kaisun2000: Good point. Maybe let us create something like `TaskUtil.getExpiredJobsFromCacheAndCheckZk` which uses `TaskUtil.getExpiredJobsFromCache()`

NealSun96: Is that going to be a public function? If so, can we remove that public function after we fix phantom write? 

I don't think this is a big problem; this function is too new to cause large impact imo.

kaisun2000: We can make something like "TaskUtil.getExpiredJobsFromCacheAndCheckZk" deprecated once we fixed root of the selective update issue. The fix may take several quarters though.

If you believe `TaskUtil.getExpiredJobsFromCache()` is not used by others. Then shall we run this diff against TMC build to see parterners in upstream can build against this one successfully? It would be much more costly if people later find this thing does not work in release time.

NealSun96: I think this may be an overkill. I did a code search and didn't find any code calling this function. 

In addition, I can allow `manager` to be null for maximum backward compatibility.

kaisun2000: Some of the partners code are actually also hosted in github. Code search won't really yield the current usage. In fact last time we  found some issue when integrating with Gobblin. Let me slack you the command to check build. 

If the build works, we can just use this diff without any change.

kaisun2000: The build seems working. let us just keep this change then.

kaisun2000 (approved):

NealSun96:
This PR is ready to be merged, approved by @alirezazamani @kaisun2000 
Final commit message:
## ExpiredJob Workaround for Selective Update Race Conditions  ##
This PR implements a workaround for determining expired jobs that avoids selective update race condition: if JobConfig doesn't exist in the cache, check ZK directly.

PR #1487, opened 2020-10-27
Add a configuration option to allow enable/disable writing error log to ZK

### LINKED ISSUE #1486: Improve Error log to Zookeeper server
Currently Helix (Participant) sends error logging to ZK and Helix controller would aggregate and remove these logs.

Some failing path of this process can result in repeated sending error messages to ZK, which brings down ZK.

For example, HelixTaskExecutor.OnMessage(), if failing to create messageHandler, would result in logging error to ZK and remove messages. Next Controller would send another messages and repeat this process. So on and so forth.

We should give an option to enable logging and by default disable these loggings, as Zookeeper is meta data store, should not be used as database or datastore.

kaisun2000:
### Issues

    fix #1486

### Description

    Improve statusUpdateUtil log error to ZK by adding an option to enabled
    it. By default, it would not log error to ZK. This is to avoid some
    error code path that keep flooding ZK sever which cause DoS to Zk.
    Such as HelixTaskExecutor onMessage creation messageHandler exception.

### Tests

  TestStatusUpdateUtil

github passed.
https://github.com/apache/helix/runs/1344486165?check_suite_focus=true

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
+++ b/helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
@@ -83,5 +83,5 @@ public class SystemPropertyKeys {
   public static final String MSDS_SERVER_ENDPOINT_KEY =
       MetadataStoreRoutingConstants.MSDS_SERVER_ENDPOINT_KEY;
 
-  public static final String STATEUPDATEUTIL_ERROR_LOG_ENABLED = "helix.StateUpdateUtil.errorLog.enabled";
+  public static final String STATEUPDATEUTIL_ERROR_LOG2ZK_ENABLED = "helix.StateUpdateUtil.errorLog.enabled";
 }
--- a/helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
+++ b/helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
@@ -82,4 +82,6 @@ public class SystemPropertyKeys {
   // System Property Metadata Store Directory Server endpoint key
   public static final String MSDS_SERVER_ENDPOINT_KEY =
       MetadataStoreRoutingConstants.MSDS_SERVER_ENDPOINT_KEY;
+
+  public static final String STATEUPDATEUTIL_ERROR_LOG_ENABLED = "helix.StateUpdateUtil.errorLog.enabled";
 }
--- a/helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
+++ b/helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
@@ -83,5 +83,5 @@ public class SystemPropertyKeys {
   public static final String MSDS_SERVER_ENDPOINT_KEY =
       MetadataStoreRoutingConstants.MSDS_SERVER_ENDPOINT_KEY;
 
-  public static final String STATEUPDATEUTIL_ERROR_LOG2ZK_ENABLED = "helix.StateUpdateUtil.errorLog.enabled";
+  public static final String STATEUPDATEUTIL_ERROR_PERSISTENCY_ENABLED = "helix.StateUpdateUtil.errorLog.enabled";
 }
--- a/helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
@@ -469,12 +469,12 @@ public final class HelixUtil {
     return propertyDefaultValue;
   }
 
-    /**
-     * Get the value of system property
-     * @param propertyKey
-     * @param propertyDefaultValue
-     * @return
-     */
+  /**
+   * Get the value of system property
+   * @param propertyKey
+   * @param propertyDefaultValue
+   * @return
+   */
   public static long getSystemPropertyAsLong(String propertyKey, long propertyDefaultValue) {
     String valueString = System.getProperty(propertyKey, "" + propertyDefaultValue);
 
--- a/helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
@@ -62,6 +62,7 @@ import org.apache.helix.model.Resource;
 import org.apache.helix.model.ResourceAssignment;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
+import org.apache.helix.tools.ClusterVerifiers.StrictMatchExternalViewVerifier;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -470,11 +471,26 @@ public final class HelixUtil {
   }
 
   /**
-   * Get the value of system property
+   * Get the boolean value of system property
    * @param propertyKey
    * @param propertyDefaultValue
    * @return
    */
+  public static boolean getSystemPropertyAsBoolean(String propertyKey, String propertyDefaultValue) {
+    String valueString = System.getProperty(propertyKey, "" + propertyDefaultValue);
+    if (valueString.toLowerCase().equals("true")) {
+      return true;
+    }
+    return false;
+  }
+
+
+    /**
+     * Get the value of system property
+     * @param propertyKey
+     * @param propertyDefaultValue
+     * @return
+     */
   public static long getSystemPropertyAsLong(String propertyKey, long propertyDefaultValue) {
     String valueString = System.getProperty(propertyKey, "" + propertyDefaultValue);
 
--- a/helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
@@ -62,7 +62,6 @@ import org.apache.helix.model.Resource;
 import org.apache.helix.model.ResourceAssignment;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
-import org.apache.helix.tools.ClusterVerifiers.StrictMatchExternalViewVerifier;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -469,21 +468,6 @@ public final class HelixUtil {
 
     return propertyDefaultValue;
   }
-
-  /**
-   * Get the boolean value of system property
-   * @param propertyKey
-   * @param propertyDefaultValue
-   * @return
-   */
-  public static boolean getSystemPropertyAsBoolean(String propertyKey, String propertyDefaultValue) {
-    String valueString = System.getProperty(propertyKey, "" + propertyDefaultValue);
-    if (valueString.toLowerCase().equals("true")) {
-      return true;
-    }
-    return false;
-  }
-
 
     /**
      * Get the value of system property
--- a/helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
@@ -57,7 +57,7 @@ public class StatusUpdateUtil {
   static Logger _logger = LoggerFactory.getLogger(StatusUpdateUtil.class);
 
   public static final boolean ERROR_LOG_TO_ZK_ENABLED =
-      Boolean.getBoolean(SystemPropertyKeys.STATEUPDATEUTIL_ERROR_LOG2ZK_ENABLED);
+      Boolean.getBoolean(SystemPropertyKeys.STATEUPDATEUTIL_ERROR_PERSISTENCY_ENABLED);
 
   public static class Transition implements Comparable<Transition> {
     private final String _msgID;
@@ -501,13 +501,12 @@ public class StatusUpdateUtil {
 
       if (isController) {
         propertyKey = keyBuilder.controllerTaskStatus(statusUpdateSubPath, statusUpdateKey);
-        accessor.updateProperty(propertyKey, new StatusUpdate(statusUpdateRecord));
       } else {
         propertyKey =
             keyBuilder.stateTransitionStatus(instanceName, sessionId, statusUpdateSubPath,
                 statusUpdateKey);
-        accessor.updateProperty(propertyKey, new StatusUpdate(statusUpdateRecord));
       }
+      accessor.updateProperty(propertyKey, new StatusUpdate(statusUpdateRecord));
 
       if (_logger.isTraceEnabled()) {
         _logger.trace("StatusUpdate path:" + propertyKey.getPath() + ", updates:"
@@ -519,13 +518,13 @@ public class StatusUpdateUtil {
     PropertyKey propertyKey;
     if (isController) {
       propertyKey = keyBuilder.controllerTaskStatus(statusUpdateSubPath, statusUpdateKey);
-      accessor.updateProperty(propertyKey, new StatusUpdate(record));
     } else {
       propertyKey =
           keyBuilder.stateTransitionStatus(instanceName, sessionId, statusUpdateSubPath,
               statusUpdateKey);
-      accessor.updateProperty(propertyKey, new StatusUpdate(record));
     }
+    accessor.updateProperty(propertyKey, new StatusUpdate(record));
+
     if (_logger.isTraceEnabled()) {
       _logger.trace("StatusUpdate path:" + propertyKey.getPath() + ", updates:" + record);
     }
--- a/helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
@@ -56,7 +56,7 @@ import org.slf4j.LoggerFactory;
 public class StatusUpdateUtil {
   static Logger _logger = LoggerFactory.getLogger(StatusUpdateUtil.class);
 
-  public static boolean ERROR_LOG_TO_ZK_ENABLED;
+  public static final boolean ERROR_LOG_TO_ZK_ENABLED;
   static {
     String valueString = System.getProperty(SystemPropertyKeys.STATEUPDATEUTIL_ERROR_LOG_ENABLED, "");
     ERROR_LOG_TO_ZK_ENABLED = valueString.equals("enabled") ? true : false;
--- a/helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
+++ b/helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
@@ -496,46 +496,38 @@ public class StatusUpdateUtil {
 
     Builder keyBuilder = accessor.keyBuilder();
     if (!_recordedMessages.containsKey(message.getMsgId())) {
+      ZNRecord statusUpdateRecord = createMessageLogRecord(message);
+      PropertyKey propertyKey;
+
       if (isController) {
-        accessor
-            .updateProperty(keyBuilder.controllerTaskStatus(statusUpdateSubPath, statusUpdateKey),
-                new StatusUpdate(createMessageLogRecord(message)));
-
+        propertyKey = keyBuilder.controllerTaskStatus(statusUpdateSubPath, statusUpdateKey);
+        accessor.updateProperty(propertyKey, new StatusUpdate(statusUpdateRecord));
       } else {
-
-        PropertyKey propertyKey =
+        propertyKey =
             keyBuilder.stateTransitionStatus(instanceName, sessionId, statusUpdateSubPath,
                 statusUpdateKey);
-
-        ZNRecord statusUpdateRecord = createMessageLogRecord(message);
-
-        // For now write participant StatusUpdates to log4j.
-        // we are using restlet as another data channel to report to controller.
-        if (_logger.isTraceEnabled()) {
-          _logger.trace("StatusUpdate path:" + propertyKey.getPath() + ", updates:"
-              + statusUpdateRecord);
-        }
         accessor.updateProperty(propertyKey, new StatusUpdate(statusUpdateRecord));
+      }
 
+      if (_logger.isTraceEnabled()) {
+        _logger.trace("StatusUpdate path:" + propertyKey.getPath() + ", updates:"
+            + statusUpdateRecord);
       }
       _recordedMessages.put(message.getMsgId(), message.getMsgId());
     }
 
+    PropertyKey propertyKey;
     if (isController) {
-      accessor.updateProperty(
-          keyBuilder.controllerTaskStatus(statusUpdateSubPath, statusUpdateKey), new StatusUpdate(
-              record));
+      propertyKey = keyBuilder.controllerTaskStatus(statusUpdateSubPath, statusUpdateKey);
+      accessor.updateProperty(propertyKey, new StatusUpdate(record));
     } else {
-
-      PropertyKey propertyKey =
+      propertyKey =
           keyBuilder.stateTransitionStatus(instanceName, sessionId, statusUpdateSubPath,
               statusUpdateKey);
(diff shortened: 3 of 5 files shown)


helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
```diff
@@ -82,4 +82,6 @@
   // System Property Metadata Store Directory Server endpoint key
   public static final String MSDS_SERVER_ENDPOINT_KEY =
       MetadataStoreRoutingConstants.MSDS_SERVER_ENDPOINT_KEY;
+
+  public static final String STATEUPDATEUTIL_ERROR_LOG_ENABLED = "helix.StateUpdateUtil.errorLog.enabled";
```
lei-xia, line 86: minor: "helix.StateUpdateUtil.errorLogToZK.enabled"?  We still log error to local logfile, right?

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
```diff
@@ -55,6 +56,12 @@
 public class StatusUpdateUtil {
   static Logger _logger = LoggerFactory.getLogger(StatusUpdateUtil.class);
 
+  public static final boolean ERROR_LOG_TO_ZK_ENABLED;
+  static {
+    String valueString = System.getProperty(SystemPropertyKeys.STATEUPDATEUTIL_ERROR_LOG_ENABLED, "");
+    ERROR_LOG_TO_ZK_ENABLED = valueString.equals("enabled") ? true : false;
```
lei-xia, line 62: Please use HelixUtil.getSystemPropertyAsBoolean(), it allows you to specify the default value there in the method.

helix-core/src/test/java/org/apache/helix/integration/manager/TestParticipantManager.java
```diff
@@ -66,6 +66,9 @@
 public class TestParticipantManager extends ZkTestBase {
   private MBeanServer _server = ManagementFactory.getPlatformMBeanServer();
   private String clusterName = TestHelper.getTestClassName();
+  static {
+    System.setProperty(SystemPropertyKeys.STATEUPDATEUTIL_ERROR_LOG_ENABLED, "enabled");
```
lei-xia, line 70: Same here, use HelixUtil.

helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
kaisun2000: This is not minor, this is a MAJOR, a very good point. Let me make sure we have error level log out.

helix-core/src/test/java/org/apache/helix/integration/manager/TestParticipantManager.java
kaisun2000: In this test, here we are setting this property so that later test which asserting existence of error record in Zk won't fail. HelixUtil does not apply here as it is used to retrieve record.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
kaisun2000: There is only HelixUtil.getSystemPropertyAsInt/Long, no asBoolean(). Let me add as Boolean then, make the changes a little bit wider.

helix-core/src/test/java/org/apache/helix/util/TestStatusUpdateUtil.java
```diff
@@ -0,0 +1,76 @@
...
+
+
+public class TestStatusUpdateUtil extends ZkTestBase {
+  private String clusterName = TestHelper.getTestClassName();
+  static {
+    System.clearProperty(SystemPropertyKeys.STATEUPDATEUTIL_ERROR_LOG_ENABLED);
+  }
+
+  @Test
+  public void testDisableErrorLogByDefault() throws Exception {
```
lei-xia, line 24: Can you test enable it by the new config?

kaisun2000: It is already tested in TestParticipantManager. testSessionExpiryInTransition assert the error log znode is not null. Do we need to add another explicit one here?

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
```diff
@@ -62,6 +62,7 @@
 import org.apache.helix.model.ResourceAssignment;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
+import org.apache.helix.tools.ClusterVerifiers.StrictMatchExternalViewVerifier;
```
pkuwm, line 65: This import is not used.

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
```diff
@@ -470,11 +471,26 @@ public static int getSystemPropertyAsInt(String propertyKey, int propertyDefault
   }
 
   /**
-   * Get the value of system property
+   * Get the boolean value of system property
    * @param propertyKey
    * @param propertyDefaultValue
    * @return
    */
+  public static boolean getSystemPropertyAsBoolean(String propertyKey, String propertyDefaultValue) {
```
pkuwm, line 479: We don't have to create this method to be redundant.
Simply, `Boolean.getBoolean(propertyKey, propertyDefaultValue)` is already what is needed, right?

lei-xia:
The PR title seems not grammar correct, maybe "Add a configuration option to allow enable/disable writing error log to ZK"?

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
```diff
@@ -55,6 +56,10 @@
 public class StatusUpdateUtil {
   static Logger _logger = LoggerFactory.getLogger(StatusUpdateUtil.class);
 
+  public static final boolean ERROR_LOG_TO_ZK_ENABLED =
+      HelixUtil.getSystemPropertyAsBoolean(SystemPropertyKeys.STATEUPDATEUTIL_ERROR_LOG_ENABLED, "false");
```
pkuwm, line 60: `Boolean.getBoolean(propertyKey, propertyDefaultValue)` is simple and good enough?

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
```diff
@@ -560,6 +565,10 @@ void publishStatusUpdateRecord(ZNRecord record, Message message, Level level,
    */
   void publishErrorRecord(ZNRecord record, String instanceName, String updateSubPath,
       String updateKey, String sessionId, HelixDataAccessor accessor, boolean isController) {
+    _logger.error("StatusUpdate Error record: {}", record.toString());
```
pkuwm, line 568: Nit, `record.toString()` is redundant, `record` is enough as `toString()` would be auto called.

helix-core/src/test/java/org/apache/helix/integration/manager/TestParticipantManager.java
```diff
@@ -66,6 +66,9 @@
 public class TestParticipantManager extends ZkTestBase {
   private MBeanServer _server = ManagementFactory.getPlatformMBeanServer();
   private String clusterName = TestHelper.getTestClassName();
+  static {
+    System.setProperty(SystemPropertyKeys.STATEUPDATEUTIL_ERROR_LOG_ENABLED, "true");
```
pkuwm, line 70: Can you help me understand why you set the property here, but clear it in `TestStatusUpdateUtil`? Why do we not put both in the test method? Otherwise, the system config would affect other tests, right?

helix-core/src/test/java/org/apache/helix/util/TestStatusUpdateUtil.java
```diff
@@ -0,0 +1,76 @@
+package org.apache.helix.util;
+
```
pkuwm, line 2: Missing apache license. Maybe you could setup apache license in your intellij?

kaisun2000:
> The PR title seems not grammar correct, maybe "Add a configuration option to allow enable/disable writing error log to ZK"?

changed.

helix-core/src/test/java/org/apache/helix/util/TestStatusUpdateUtil.java
```diff
@@ -0,0 +1,76 @@
...
+    // assert by default, not logged to Zookeeper
+    String errPath = PropertyPathBuilder.instanceError(clusterName, "localhost_12918",
+        participants[0].getSessionId(),
+        "TestDB", "TestDB_0");
+    try {
+      ZNRecord error = _gZkClient.readData(errPath);
+      Assert.fail("not expecting being able to send error logs to ZK by default.");
+    } catch (ZkException zke) {
+      Assert.assertTrue(true);
```
pkuwm, line 69: Why `assertTrue(True);` which is always true? Just put a comment it is expected. Or to be more accurate, we can assert the exception message to ensure the exception is the expected one not others(same exception type but different message thrown from other places).

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
```diff
@@ -560,6 +565,10 @@ void publishStatusUpdateRecord(ZNRecord record, Message message, Level level,
...
+    if (!ERROR_LOG_TO_ZK_ENABLED) {
```
jiajunwang, line 569: IMHO,

if (!ERROR_LOG_TO_ZK_ENABLED) {
  _logger.error("StatusUpdate Error record: {}", record.toString());
  return;
}
...

helix-core/src/test/java/org/apache/helix/util/TestStatusUpdateUtil.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    for (int i = 0; i < n; i++) {
+      participants[i].syncStop();
```
jiajunwang, line 73: This is a potential thread leak if test failure.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
lei-xia: Not actually, even we disable writting to ZK, we still need to log into local log file, right?

helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
lei-xia, line 86: Again, we are disabling log to ZK, not disabling error log (to log file) at all, right?

helix-core/src/test/java/org/apache/helix/util/TestStatusUpdateUtil.java
lei-xia: It is better to add both positive and negative tests in here. The main purpose of the other test is not testing error log though.

helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
dasahcc: We have a class ZkSystemPropertyKeys to hold this kind of system property. Please put this there.

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
dasahcc: +1. Unless you need some error handling in the function.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
jiajunwang: I don't have a strong preference here.
But I think it depends on how we treat this util, is it a more powerful logger? Or it is a status updater like the name suggested.

The current implementation is treating this class as a pure status updater. There is no extra log done unless the update process fails. And I think it is clean. If a log is needed, then the caller shall call logger method in parallel with the publishErrorRecord() method.

On the other hand, if we want to twist this class to be a more powerful logger, then we will have mixed usage. Only part of the code is using it. And whether to use it or not is not clear.

Overall, I treat the log here as a fallback logic when ERROR_LOG_TO_ZK_ENABLED == false. So it shall not log when the util still writes to ZK.

kaisun2000: So let us keep it this way. The reasoning is that source of truth logging is in log file.

helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
kaisun2000: @lei-xia, so let us change the name to " STATEUPDATEUTIL_ERROR_LOG2ZK_ENABLED?

kaisun2000: @dasahcc, ZkSystemPropertyKeys is for Zookeeper related. This is more to me like logging related.

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/util/HelixUtil.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
kaisun2000: changed.

helix-core/src/test/java/org/apache/helix/util/TestStatusUpdateUtil.java
kaisun2000: added.

helix-core/src/test/java/org/apache/helix/util/TestStatusUpdateUtil.java
kaisun2000: fixed in tearing down in afterClass()

helix-core/src/test/java/org/apache/helix/integration/manager/TestParticipantManager.java
kaisun2000:
```
  public void testSessionExpiryInTransition() throws Exception {

    String errPath = PropertyPathBuilder.instanceError(clusterName, "localhost_12918", oldSessionId,
        "TestDB0", "TestDB0_0");
    ZNRecord error = _gZkClient.readData(errPath);
    Assert.assertNotNull(error,
        "InterruptedException should happen in old session since task is being cancelled during handleNewSession");

```

To make sure the following test would work.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
jiajunwang: I'm fine with both designs. But as I mentioned, this change is half-way. Please either ensure that the source of truth is with the log for ALL the status util call (including log error, log status, etc.) or please don't log here otherwise the util is mixed up with some methods with log and the others don't.

helix-core/src/test/java/org/apache/helix/integration/manager/TestParticipantManager.java
kaisun2000: ERROR_LOG_TO_ZK_ENABLED in StatusUpdateUtil is static. I changed to use reflection to control the value now. So there is no need to set or clear it in `TestStatusUpdateUtil` anymore.

helix-core/src/test/java/org/apache/helix/util/TestStatusUpdateUtil.java
kaisun2000: both added now.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
kaisun2000: Good point. The log are in `publishStatusUpdateRecord' and `publishErrorRecord`.

Here publishStatusUpdateRecord would never log to Zk anymore. The participant side, it has trace level log to log4j, but not controller side. Let me revise it to the way that both controller and participant side has log to log4j then.

kaisun2000:
@dasahcc , @pkuwm , @jiajunwang , ping?

helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
```diff
@@ -82,4 +82,6 @@
   // System Property Metadata Store Directory Server endpoint key
   public static final String MSDS_SERVER_ENDPOINT_KEY =
       MetadataStoreRoutingConstants.MSDS_SERVER_ENDPOINT_KEY;
+
+  public static final String STATEUPDATEUTIL_ERROR_LOG2ZK_ENABLED = "helix.StateUpdateUtil.errorLog.enabled";
```
jiajunwang, line 86: I thought more about this configuration. Can we directly make it an enum now instead of changing it later? Otherwise, we will need to deprecate this item and go through the painful backward compatibility checklist in the future.

I suggest making it STATEUPDATE_ERROR_REPORTING_CHANNEL.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
```diff
@@ -492,47 +496,39 @@ void publishStatusUpdateRecord(ZNRecord record, Message message, Level level,
...
-      if (isController) {
-        accessor
-            .updateProperty(keyBuilder.controllerTaskStatus(statusUpdateSubPath, statusUpdateKey),
-                new StatusUpdate(createMessageLogRecord(message)));
+      ZNRecord statusUpdateRecord = createMessageLogRecord(message);
+      PropertyKey propertyKey;
 
+      if (isController) {
+        propertyKey = keyBuilder.controllerTaskStatus(statusUpdateSubPath, statusUpdateKey);
+        accessor.updateProperty(propertyKey, new StatusUpdate(statusUpdateRecord));
```
jiajunwang, line 504: nit, we can move this line out of the if-else block too.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
```diff
@@ -492,47 +496,39 @@ void publishStatusUpdateRecord(ZNRecord record, Message message, Level level,
...
       _recordedMessages.put(message.getMsgId(), message.getMsgId());
     }
 
+    PropertyKey propertyKey;
     if (isController) {
-      accessor.updateProperty(
-          keyBuilder.controllerTaskStatus(statusUpdateSubPath, statusUpdateKey), new StatusUpdate(
-              record));
+      propertyKey = keyBuilder.controllerTaskStatus(statusUpdateSubPath, statusUpdateKey);
+      accessor.updateProperty(propertyKey, new StatusUpdate(record));
```
jiajunwang, line 522: same here.

helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
kaisun2000: Can you illustrate the idea how to make it enum? Something like 

```
enum STATEUPDATE_ERROR_REPORTING_CHANNEL {
    LOG4J,
    LOG2ZK
};

```

and in the system property, 
`public static final String STATEUPDATEUTIL_ERROR_LOG2ZK_ENABLED = <enum value>`;
then we will convert the system property to enum value?

kaisun2000: STATEUPDATEUTIL_ERROR_PERSISTANCY_ENABLED

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
```diff
@@ -560,6 +556,10 @@ void publishStatusUpdateRecord(ZNRecord record, Message message, Level level,
    */
   void publishErrorRecord(ZNRecord record, String instanceName, String updateSubPath,
       String updateKey, String sessionId, HelixDataAccessor accessor, boolean isController) {
+    _logger.error("StatusUpdate Error record: {}", record);
```
pkuwm, line 559: Do we really need this error log for the whole record? It may eat up the disk, imagine if znrecord is close to 1 MB. I am not sure if we really need the record for troubleshooting. Maybe logging the instanceName, updateSubPath, sessionId, etc. is good enough?

kaisun2000: This is not a concern. Previously the issue is that N participant would write the log to one Zk quorum at the same time repeatedly. Here, we are writing to N participant at the same time. If this is a concern, we need to revisit all the participant side logging then.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
kaisun2000: fixed.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
kaisun2000: fixed.

helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
```diff
@@ -55,6 +56,9 @@
 public class StatusUpdateUtil {
   static Logger _logger = LoggerFactory.getLogger(StatusUpdateUtil.class);
 
+  public static final boolean ERROR_LOG_TO_ZK_ENABLED =
```
jiajunwang, line 59: nit, private?
Since we already doing reflection, I think making it private won't block the test, right?

jiajunwang (approved):
LGTM! One minor suggestion.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
kaisun2000: Actually it is not the case. Making it private won't work as I just test. I am not an expert in this area. If you know other trick, let me know.

kaisun2000:
The PR is approved, please help to merge

> Improve statusUpdateUtil log error to ZK by adding an option to enabled
it. By default, it would not log error to ZK. This is to avoid some
error code path that keep flooding ZK sever which cause DoS to Zk.
Such as HelixTaskExecutor onMessage creation messageHandler exception.

helix-core/src/main/java/org/apache/helix/util/StatusUpdateUtil.java
kaisun2000: This is not the case. Just tested, it does not work. I am not an expert in this area. Feel free to let me know if you know the trick here.

helix-core/src/test/java/org/apache/helix/integration/manager/TestParticipantManager.java
```diff
@@ -66,6 +66,9 @@
 public class TestParticipantManager extends ZkTestBase {
   private MBeanServer _server = ManagementFactory.getPlatformMBeanServer();
   private String clusterName = TestHelper.getTestClassName();
+  static {
+    System.setProperty(SystemPropertyKeys.STATEUPDATEUTIL_ERROR_PERSISTENCY_ENABLED, "true");
```
pkuwm, line 71: The system property is JVM level. If we don't unset the config, later tests will be using this config. Maybe there are no tests using the config yet. But I suggest, as a good practice, we also unset the config to make unit tests isolated, ensuring other tests won't be impacted by this test's configs.

pkuwm:
I'll merge this PR once the PR CI is successful.

Issue #1488, opened 2020-10-27
Participant side message handling logic may fail silently


jiajunwang:
### Describe the bug
There are more than one places in the participant code that the message processing may fail silently. Meaning no error state, no error message. The message is just discarded. When this kind of error happens, the controller or other components might keep sending the same messages repeatedly.
This may lead to a very serious result since controller / ZK resources may be exhausted.

### To Reproduce
Define a handler factory logic that will throw Exception on the create handler call.
The message will be removed but no ERROR state or other obvious failure can be observed from the controller side.

### Expected behavior
For state transition messages, we shall mark the corresponding partition to be ERROR.
For other messages, we shall leave the message that causes this error untouched.

PR #1489, opened 2020-10-27
Refine the HelixTaskExecutor to reduce duplicate code and clarify the code structure.

jiajunwang:
### Issues

This PR is related to #1488 given we need to clean up the related code so as to make the further fix. But it won't address the issue.

### Description

Refine the HelixTaskExecutor to reduce duplicate code and clarify the code structure.
There is a minor code logic change for optimization. But there is no business logic change in this PR.

### Tests

NA

[ERROR] Failures:
[INFO]
[ERROR] Tests run: 1234, Failures: 1, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time: 01:28 h
[INFO] Finished at: 2020-10-28T15:46:36-07:00

This test keeps failing on the master branch too. We are investigating this problem separately.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -799,7 +799,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
     Set<String> createCurStateNames = new HashSet<>();
 
     for (Message message : messages) {
-      if (checkForNoOpMessage(message, instanceName, changeContext, manager, sessionId,
+      if (checkAndProcessNoOpMessage(message, instanceName, changeContext, manager, sessionId,
           stateTransitionHandlers)) {
         // skip the following operations for the no-op messages.
         continue;
@@ -815,7 +815,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
         }
         if (message.getMsgType().equals(MessageType.STATE_TRANSITION.name()) || message.getMsgType()
             .equals(MessageType.STATE_TRANSITION_CANCELLATION.name())) {
-          if (validateStateTransitionMessage(message, instanceName, manager,
+          if (validateAndProcessStateTransitionMessage(message, instanceName, manager,
               stateTransitionHandlers, msgHandler)) {
             // Need future process by triggering state transition
             String msgTarget =
@@ -901,7 +901,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
    * @param stateTransitionHandlers
    * @return True if the message is no-op message and no other process step is required.
    */
-  private boolean checkForNoOpMessage(Message message, String instanceName,
+  private boolean checkAndProcessNoOpMessage(Message message, String instanceName,
       NotificationContext changeContext, HelixManager manager, String sessionId,
       Map<String, MessageHandler> stateTransitionHandlers) {
     HelixDataAccessor accessor = manager.getHelixDataAccessor();
@@ -997,10 +997,10 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
    * @param manager
    * @param stateTransitionHandlers
    * @param createHandler
-   * @return True if the requested state transition is valid.
+   * @return True if the requested state transition is valid, and need to schedule the transition.
    *         False if no more operation is required.
    */
-  private boolean validateStateTransitionMessage(Message message, String instanceName,
+  private boolean validateAndProcessStateTransitionMessage(Message message, String instanceName,
       HelixManager manager, Map<String, MessageHandler> stateTransitionHandlers,
       MessageHandler createHandler) {
     HelixDataAccessor accessor = manager.getHelixDataAccessor();
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -34,7 +34,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
@@ -75,6 +74,7 @@ import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.zkclient.DataUpdater;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 
 public class HelixTaskExecutor implements MessageListener, TaskExecutor {
   /**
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -192,43 +191,38 @@ public void registerMessageHandlerFactory(String type, MessageHandlerFactory fac
...
       if (!((MultiTypeMessageHandlerFactory) factory).getMessageTypes().contains(type)) {
         throw new HelixException("Message factory type mismatch. Type: " + type + ", factory: "
             + ((MultiTypeMessageHandlerFactory) factory).getMessageTypes());
       }
     } else {
       if (!factory.getMessageType().equals(type)) {
         throw new HelixException(
-            "Message factory type mismatch. Type: " + type + ", factory: " + factory.getMessageType());
+            "Message factory type mismatch. Type: " + type + ", factory: " + factory
+                .getMessageType());
```
kaisun2000, line 203: All the above changes seems to be non-essential change, right?

jiajunwang: This PR is all about non-essential changes that are the preparation for the real business logic changes.
Non-essential does not mean we are not going to do it.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -192,43 +191,38 @@ public void registerMessageHandlerFactory(String type, MessageHandlerFactory fac
...
-        }
-      });
-      ExecutorService prevExecutor = _executorMap.putIfAbsent(type, newPool);
-      if (prevExecutor != null) {
-        LOG.warn("Skip creating a new thread pool for type: " + type + ", already existing pool: "
-            + prevExecutor + ", isShutdown: " + prevExecutor.isShutdown());
-        newPool.shutdown();
-        newPool = null;
-      } else {
+      _executorMap.computeIfAbsent(type, msgType -> {
```
dasahcc, line 213: We can use putIfAbsent since we are not using the return value anymore here.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -997,30 +873,186 @@ public void onMessage(String instanceName, List<Message> messages,
...
+   * Inspect the message. Report and remove it if no operation needs to be done.
+   * @param message
+   * @param instanceName
+   * @param changeContext
+   * @param manager
+   * @param sessionId
+   * @param stateTransitionHandlers
+   * @return True if the message is no-op message and no other process step is required.
+   */
+  private boolean checkForNoOpMessage(Message message, String instanceName,
```
dasahcc, line 904: nit: better to name it as isNoOpMessage? The word check could be confusing here.

kaisun2000, line 904: This is not just check noOp, it also has other functionalities, check message expired or cancellation logic. Shall we change the name of this one?

More importantly, any exception thrown out here without statemodel marked as ERROR can cause controller sending another message repeatedly. Can you this invariant to the comment? So later if someone not familiar with this code can understand the intention here.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -810,164 +799,52 @@ public void onMessage(String instanceName, List<Message> messages,
...
           nonStateTransitionContexts.add(msgWorkingContext);
         }
       } catch (Exception e) {
         handleUnprocessableMessage(message, e, e.getMessage(), accessor, instanceName, manager);
         continue;
       }
 
-      markReadMessage(message, msgWorkingContext, manager);
-      readMsgs.add(message);
-
```
kaisun2000, line 964: why remove these two lines?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: putIfAbsent is not good since you will need to create the new pool, check, then close the new pool if there is already an existing one.
This is what I optimized here.

computeIfAbsent won't trigger the compute callback if the key already exists.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: This method is now also cleaning up the message, and it is not easy to move the logic out. So I prefer not calling it as an isXXXX method.

jiajunwang: Any suggestion regarding the name? I feel "checkAndProcessNoOpMessage" is too long. But maybe that's what we need.

Regarding your 2nd comment, could you please point out the potential code that throws Exception? I will take care of that in the following PR. We won't change any business logic in this one.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: Nothing being removed.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: Changed to checkAndProcessNoOpMessage

kaisun2000: Right now, the code is actually within a big try{} block. the catch block would remove message. I know in the next diff you will do more magic to make sure when this method exit, they must be a statemodel with some state. Currently, there is no logic error (probably the only thing is the first line of getHelixDataAcessor() is one is super picky). However, state this variant would help later developers to understand the reasoning here. So I think it would be a good idea to have a comment here.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -997,30 +873,186 @@ public void onMessage(String instanceName, List<Message> messages,
...
+   * @param changeContext
+   * @param manager
+   * @param sessionId
+   * @param stateTransitionHandlers
+   * @return True if the message is no-op message and no other process step is required.
+   */
+  private boolean checkAndProcessNoOpMessage(Message message, String instanceName,
+      NotificationContext changeContext, HelixManager manager, String sessionId,
+      Map<String, MessageHandler> stateTransitionHandlers) {
+    HelixDataAccessor accessor = manager.getHelixDataAccessor();
```
kaisun2000, line 907: nit, put this line also into try{} block.

kaisun2000 (approved):
Looks good to me. Please address the additional nit which is to put 1HelixDataAccessor accessor = manager.getHelixDataAccessor();` into try block. This is important to make sure you don't throw anything out from this method.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: If this fails, meaning we lost ZK connection and the wait is timed out. In this case, I think it would be better to throw the exception instead of proceeding. The handleUnprocessableMessage will fail too due to ZK connection breaks.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: Depends on the solution, we might need to comment on this logic from a different perspective. I'm not convinced by now that we shall return the statemodel. And the logic here is not only for state transition messages only. So comment here will introduce more confusion.
I prefer to make the existing logic clean enough first. I will add a comment if necessary when on the real fix.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: One more point why we should not change it, moving this line into the try-catch block changes the business logic.

kaisun2000:
```
  public HelixDataAccessor getHelixDataAccessor() {
    checkConnected(_waitForConnectedTimeout);
    return _dataAccessor;
  }
```

if somehow the _zkClient is closed, the above code would throw exception, which would cause the same issue of repeatedly controller sending messages.

Note, with new change to statusUpdateUtil, we won't really use _zkClient from ZKHelixManager to write to Zookeeper. 

In sum, I think the core idea (invariant) is that don't throw anything out from this method. Anyway, up to you.

jiajunwang: This specific case won't cause retry, since the message won't be removed when zkclient is down.
statusUpdateUtil is a different story. It should be considered in different PRs.

Based on what we discussed today with @dasahcc and @lei-xia , we agreed that it is not ideal or applicable for us to eliminate all potential Exceptions. And leave the message in any of the Exception cases is dangerous, IMO.
So we will take this case by case.

In the following PRs, we will ensure all the Exceptions of application code (behind the interfaces) will be handled gracefully. But our internal logic exceptions will be evaluated one by one in other PRs, if necessary.

jiajunwang:
Discussed with the team about the roadmap, the remaining comment does not fit the scope of this work. And the specific case mentioned in this PR won't cause trouble. So I will move forward with Kai's approval.

PR #1494, opened 2020-10-28
Add admin and Rest API to purge instances that have been offline more than specified period of time

### LINKED ISSUE #1493: Implement purging offline participants
Helix plan to have the function to allow users to purge offline participants that have been timed out. 

Provide both Java API and REST API for users to purge offline participants. Users may specify the timeout through API input or just fall back to the default value defined in cluster config. If either is defined as a timeout, Helix will not purge participants. 

**Additional context**

zhangmeng916:
### Issues

Fixed #1493 

### Description

Helix plan to have the function to allow users to purge offline participants that have been timed out. This PR provides both Java API and REST API for users to purge offline participants. Users may specify the timeout through API input or just fall back to the default value defined in cluster config. If either is defined as a timeout, Helix will not purge participants.

### Tests

TestZkHelixAdmin
TestClusterAccessor

helix-core:
[INFO] Tests run: 1237, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,878.091 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:21 h
[INFO] Finished at: 2020-11-01T17:40:42-08:00

helix-rest:
[INFO] Tests run: 171, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 181.257 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  03:05 min
[INFO] Finished at: 2020-11-01T16:49:19-08:00

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
@@ -229,6 +229,13 @@ public interface HelixAdmin {
   void dropInstance(String clusterName, InstanceConfig instanceConfig);
 
   /**
+   * Purge offline instances from a cluster
+   * @param clusterName
+   * @param customizedPurgeMap
+   */
+  void purgeOfflineInstances(String clusterName, Map<String, String> customizedPurgeMap);
+
+  /**
    * Get ideal state for a resource
    * @param clusterName
    * @param resourceName
--- a/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
@@ -229,11 +229,14 @@ public interface HelixAdmin {
   void dropInstance(String clusterName, InstanceConfig instanceConfig);
 
   /**
-   * Purge offline instances from a cluster
+   * Purge offline instances that have been offline for longer than the offline duration time
+   * from a cluster
    * @param clusterName
-   * @param timeout
+   * @param offlineDuration the threshold of offline time to decide whether an offline instance
+   *                        should be dropped when users call purge function. It is defined in
+   *                        milliseconds
    */
-  void purgeOfflineInstances(String clusterName, Long timeout);
+  void purgeOfflineInstances(String clusterName, long offlineDuration);
 
   /**
    * Get ideal state for a resource
--- a/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
@@ -231,9 +231,9 @@ public interface HelixAdmin {
   /**
    * Purge offline instances from a cluster
    * @param clusterName
-   * @param customizedPurgeMap
+   * @param timeout
    */
-  void purgeOfflineInstances(String clusterName, Map<String, String> customizedPurgeMap);
+  void purgeOfflineInstances(String clusterName, Long timeout);
 
   /**
    * Get ideal state for a resource
--- a/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
@@ -232,11 +232,11 @@ public interface HelixAdmin {
    * Purge offline instances that have been offline for longer than the offline duration time
    * from a cluster
    * @param clusterName
-   * @param offlineTimeout the threshold of offline time to decide whether an offline instance
-   *                        should be dropped when users call purge function. It is defined in
-   *                        milliseconds
+   * @param offlineDuration if an offline instance has been offline for longer than the set
+   *                        offlineDuration, the offline instance becomes eligible for being
+   *                        purged/deleted
    */
-  void purgeOfflineInstances(String clusterName, long offlineTimeout);
+  void purgeOfflineInstances(String clusterName, long offlineDuration);
 
   /**
    * Get ideal state for a resource
--- a/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
@@ -232,11 +232,12 @@ public interface HelixAdmin {
    * Purge offline instances that have been offline for longer than the offline duration time
    * from a cluster
    * @param clusterName
-   * @param offlineDuration the threshold of offline time to decide whether an offline instance
+   * @param offlineTimeout the threshold of offline time to decide whether an offline instance
    *                        should be dropped when users call purge function. It is defined in
    *                        milliseconds
+   * @return the list of instance names that failed to be purged.
    */
-  void purgeOfflineInstances(String clusterName, long offlineDuration);
+  List<String> purgeOfflineInstances(String clusterName, long offlineTimeout);
 
   /**
    * Get ideal state for a resource
--- a/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixAdmin.java
@@ -235,9 +235,8 @@ public interface HelixAdmin {
    * @param offlineTimeout the threshold of offline time to decide whether an offline instance
    *                        should be dropped when users call purge function. It is defined in
    *                        milliseconds
-   * @return the list of instance names that failed to be purged.
    */
-  List<String> purgeOfflineInstances(String clusterName, long offlineTimeout);
+  void purgeOfflineInstances(String clusterName, long offlineTimeout);
 
   /**
    * Get ideal state for a resource
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -76,6 +76,7 @@ import org.apache.helix.model.MaintenanceSignal;
 import org.apache.helix.model.Message;
 import org.apache.helix.model.Message.MessageState;
 import org.apache.helix.model.Message.MessageType;
+import org.apache.helix.model.ParticipantHistory;
 import org.apache.helix.model.PauseSignal;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
@@ -105,6 +106,7 @@ public class ZKHelixAdmin implements HelixAdmin {
   public static final String CONNECTION_TIMEOUT = "helixAdmin.timeOutInSec";
   private static final String MAINTENANCE_ZNODE_ID = "maintenance";
   private static final int DEFAULT_SUPERCLUSTER_REPLICA = 3;
+  public static final String OFFLINE_NODE_PURGE_TIMEOUT = "purge.timeout";
 
   private final RealmAwareZkClient _zkClient;
   private final ConfigAccessor _configAccessor;
@@ -248,6 +250,13 @@ public class ZKHelixAdmin implements HelixAdmin {
         }
       }
     }
+  }
+
+  @Override
+  public void purgeOfflineInstances(String clusterName, Map<String, String> customizedPurgeMap) {
+    Map<String, InstanceConfig> timeoutOfflineInstances = findTimeoutOfflineInstances(clusterName,
+        customizedPurgeMap);
+    timeoutOfflineInstances.values().forEach(value -> dropInstance(clusterName, value));
   }
 
   @Override
@@ -2060,5 +2069,46 @@ public class ZKHelixAdmin implements HelixAdmin {
           createZkClient(_realmMode, _realmAwareZkConnectionConfig, _realmAwareZkClientConfig,
               _zkAddress), false);
     }
+  }
+
+  private Map<String, InstanceConfig> findTimeoutOfflineInstances(String clusterName,
+      Map<String, String> customizedPurgeMap) {
+    String timeout = customizedPurgeMap.get(OFFLINE_NODE_PURGE_TIMEOUT);
+    Long timeoutValue;
+    // in case there is no customized timeout value, use the one defined in cluster config
+    if (timeout == null) {
+      timeoutValue = _configAccessor.getClusterConfig(clusterName).getOfflineNodeTimeOutForPurge();
+    } else {
+      timeoutValue = Long.valueOf(timeout);
+    }
+
+    Map<String, InstanceConfig> instanceConfigMap = new HashMap<>();
+    String path = PropertyPathBuilder.instanceConfig(clusterName);
+    BaseDataAccessor<ZNRecord> baseAccessor = new ZkBaseDataAccessor<>(_zkClient);
+    List<ZNRecord> znRecords = baseAccessor.getChildren(path, null, 0, 0, 0);
+    for (ZNRecord record : znRecords) {
+      if (record != null) {
+        InstanceConfig instanceConfig = new InstanceConfig(record);
+        instanceConfigMap.put(instanceConfig.getInstanceName(), instanceConfig);
+      }
+    }
+
+    path = PropertyPathBuilder.liveInstance(clusterName);
+    List<String> liveNodes = baseAccessor.getChildNames(path, 0);
+    liveNodes.forEach(liveNode -> instanceConfigMap.remove(liveNode));
+
+    Set<String> toRemoveInstances = new HashSet<>();
+    for (String instanceName : instanceConfigMap.keySet()) {
+      String historyPath = PropertyPathBuilder.instanceHistory(clusterName, instanceName);
+      ZNRecord znRecord = baseAccessor.get(historyPath, null, 0);
+      ParticipantHistory participantHistory = new ParticipantHistory(znRecord);
+      long lastOfflineTime = participantHistory.getLastOfflineTime();
+      if (lastOfflineTime == -1 || timeoutValue < 0
+          || System.currentTimeMillis() - lastOfflineTime < timeoutValue) {
+        toRemoveInstances.add(instanceName);
+      }
+    }
+    toRemoveInstances.forEach(instance -> instanceConfigMap.remove(instance));
+    return instanceConfigMap;
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -254,20 +254,21 @@ public class ZKHelixAdmin implements HelixAdmin {
   }
 
   @Override
-  public List<String> purgeOfflineInstances(String clusterName, long offlineTimeout) {
-    Map<String, InstanceConfig> timeoutOfflineInstances = findTimeoutOfflineInstances(clusterName
-        , offlineTimeout);
+  public void purgeOfflineInstances(String clusterName, long offlineTimeout) {
+    Map<String, InstanceConfig> timeoutOfflineInstances =
+        findTimeoutOfflineInstances(clusterName, offlineTimeout);
     List<String> failToPurgeInstances = new ArrayList<>();
     timeoutOfflineInstances.values().forEach(instance -> {
       try {
         dropInstance(clusterName, instance);
       } catch (HelixException e) {
-        logger.info("Failed to purge instance {} in cluster {}. Exception: {}", instance,
-            clusterName, e);
         failToPurgeInstances.add(instance.getInstanceName());
       }
     });
-    return failToPurgeInstances;
+    if (failToPurgeInstances.size() > 0) {
+      LOG.error("ZKHelixAdmin::purgeOfflineInstances(): failed to drop the following instances: "
+          + failToPurgeInstances);
+    }
   }
 
   @Override
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -99,7 +99,7 @@ import org.apache.zookeeper.KeeperException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import static org.apache.helix.model.ClusterConfig.OFFLINE_TIMEOUT_FOR_PURGE_NOT_SET;
+import static org.apache.helix.model.ClusterConfig.OFFLINE_DURATION_FOR_PURGE_NOT_SET;
 
 
 public class ZKHelixAdmin implements HelixAdmin {
@@ -254,9 +254,9 @@ public class ZKHelixAdmin implements HelixAdmin {
   }
 
   @Override
-  public void purgeOfflineInstances(String clusterName, long offlineTimeout) {
+  public void purgeOfflineInstances(String clusterName, long offlineDuration) {
     Map<String, InstanceConfig> timeoutOfflineInstances =
-        findTimeoutOfflineInstances(clusterName, offlineTimeout);
+        findTimeoutOfflineInstances(clusterName, offlineDuration);
     List<String> failToPurgeInstances = new ArrayList<>();
     timeoutOfflineInstances.values().forEach(instance -> {
       try {
@@ -2084,13 +2084,13 @@ public class ZKHelixAdmin implements HelixAdmin {
   }
 
   private Map<String, InstanceConfig> findTimeoutOfflineInstances(String clusterName,
-      long offlineTimeout) {
+      long offlineDuration) {
     Map<String, InstanceConfig> instanceConfigMap = new HashMap<>();
     // in case there is no customized timeout value, use the one defined in cluster config
-    if (offlineTimeout == OFFLINE_TIMEOUT_FOR_PURGE_NOT_SET) {
-      offlineTimeout =
-          _configAccessor.getClusterConfig(clusterName).getOfflineNodeTimeOutForPurge();
-      if (offlineTimeout == OFFLINE_TIMEOUT_FOR_PURGE_NOT_SET) {
+    if (offlineDuration == OFFLINE_DURATION_FOR_PURGE_NOT_SET) {
+      offlineDuration =
+          _configAccessor.getClusterConfig(clusterName).getOfflineDurationForPurge();
+      if (offlineDuration == OFFLINE_DURATION_FOR_PURGE_NOT_SET) {
         return instanceConfigMap;
       }
     }
@@ -2107,8 +2107,8 @@ public class ZKHelixAdmin implements HelixAdmin {
       ParticipantHistory participantHistory =
           accessor.getProperty(keyBuilder.participantHistory(instanceName));
       long lastOfflineTime = participantHistory.getLastOfflineTime();
-      if (lastOfflineTime == OFFLINE_TIMEOUT_FOR_PURGE_NOT_SET
-          || System.currentTimeMillis() - lastOfflineTime < offlineTimeout) {
+      if (lastOfflineTime == OFFLINE_DURATION_FOR_PURGE_NOT_SET
+          || System.currentTimeMillis() - lastOfflineTime < offlineDuration) {
         toRemoveInstances.add(instanceName);
       }
     }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -99,6 +99,8 @@ import org.apache.zookeeper.KeeperException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.helix.model.ClusterConfig.OFFLINE_TIMEOUT_FOR_PURGE_NOT_SET;
+
 
 public class ZKHelixAdmin implements HelixAdmin {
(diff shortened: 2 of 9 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2061,4 +2070,45 @@ public ZKHelixAdmin build() {
...
+
+  private Map<String, InstanceConfig> findTimeoutOfflineInstances(String clusterName,
+      Map<String, String> customizedPurgeMap) {
+    String timeout = customizedPurgeMap.get(OFFLINE_NODE_PURGE_TIMEOUT);
+    Long timeoutValue;
+    // in case there is no customized timeout value, use the one defined in cluster config
+    if (timeout == null) {
+      timeoutValue = _configAccessor.getClusterConfig(clusterName).getOfflineNodeTimeOutForPurge();
+    } else {
+      timeoutValue = Long.valueOf(timeout);
```
dasahcc, line 2082: Shall we do a try catch? If the timeout is invalid string by human error, shall we stop it or use default value?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2061,4 +2070,45 @@ public ZKHelixAdmin build() {
...
+    for (ZNRecord record : znRecords) {
+      if (record != null) {
+        InstanceConfig instanceConfig = new InstanceConfig(record);
+        instanceConfigMap.put(instanceConfig.getInstanceName(), instanceConfig);
+      }
+    }
+
+    path = PropertyPathBuilder.liveInstance(clusterName);
+    List<String> liveNodes = baseAccessor.getChildNames(path, 0);
+    liveNodes.forEach(liveNode -> instanceConfigMap.remove(liveNode));
```
dasahcc, line 2098: We can create a set and use instanceConfigMap.keySet().removeAll(set).

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2061,4 +2070,45 @@ public ZKHelixAdmin build() {
...
+      String historyPath = PropertyPathBuilder.instanceHistory(clusterName, instanceName);
+      ZNRecord znRecord = baseAccessor.get(historyPath, null, 0);
+      ParticipantHistory participantHistory = new ParticipantHistory(znRecord);
+      long lastOfflineTime = participantHistory.getLastOfflineTime();
+      if (lastOfflineTime == -1 || timeoutValue < 0
+          || System.currentTimeMillis() - lastOfflineTime < timeoutValue) {
+        toRemoveInstances.add(instanceName);
+      }
+    }
+    toRemoveInstances.forEach(instance -> instanceConfigMap.remove(instance));
```
dasahcc, line 2111: Same here, we can use instanceConfigMap.keySet().removeAll(toRemoveInstances); It has better performance.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2061,4 +2070,45 @@ public ZKHelixAdmin build() {
...
+    List<String> liveNodes = baseAccessor.getChildNames(path, 0);
+    liveNodes.forEach(liveNode -> instanceConfigMap.remove(liveNode));
+
+    Set<String> toRemoveInstances = new HashSet<>();
+    for (String instanceName : instanceConfigMap.keySet()) {
+      String historyPath = PropertyPathBuilder.instanceHistory(clusterName, instanceName);
+      ZNRecord znRecord = baseAccessor.get(historyPath, null, 0);
+      ParticipantHistory participantHistory = new ParticipantHistory(znRecord);
+      long lastOfflineTime = participantHistory.getLastOfflineTime();
+      if (lastOfflineTime == -1 || timeoutValue < 0
```
NealSun96, line 2106: If `timeoutValue < 0`, can we return earlier with an empty map?

NealSun96 (changes requested):

lei-xia:
To make the PR title more clear, maybe change it to "Add admin and Rest API to purge instances that have been offline more than specified period of time"?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -105,6 +106,7 @@
   public static final String CONNECTION_TIMEOUT = "helixAdmin.timeOutInSec";
   private static final String MAINTENANCE_ZNODE_ID = "maintenance";
   private static final int DEFAULT_SUPERCLUSTER_REPLICA = 3;
+  public static final String OFFLINE_NODE_PURGE_TIMEOUT = "purge.timeout";
```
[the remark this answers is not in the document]
lei-xia: If this is an API, should not the timeout be specified in the purgeOfflineInstances() call? Why still need to keep this in clusterConfig?

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -228,6 +228,13 @@ void addResource(String clusterName, String resourceName, int numPartitions, Str
    */
   void dropInstance(String clusterName, InstanceConfig instanceConfig);
 
+  /**
+   * Purge offline instances from a cluster
+   * @param clusterName
+   * @param customizedPurgeMap
+   */
+  void purgeOfflineInstances(String clusterName, Map<String, String> customizedPurgeMap);
```
lei-xia, line 236: Should not be the API looks like:  purgeOfflineInstances(clusterName, long timeout), which will delete all instances that have been offline longer than timeout?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -250,6 +252,13 @@ public void dropInstance(String clusterName, InstanceConfig instanceConfig) {
     }
   }
 
+  @Override
+  public void purgeOfflineInstances(String clusterName, Map<String, String> customizedPurgeMap) {
```
lei-xia, line 256: Why a map here, what could be specified other than timeout?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -282,6 +282,17 @@ public Response updateCluster(@PathParam("clusterId") String clusterId,
...
         }
         break;
+      case purgeOfflineParticipants:
+        Map<String, String> customizedPurgeMap = new HashMap<>();
+        try {
+          customizedPurgeMap =
+              OBJECT_MAPPER.readValue(content, new TypeReference<HashMap<String, String>>() {
+              });
+        } catch (Exception e) {
+          // NOP
```
pkuwm, line 292: Can we have a warning log for why json parsing fails which could help us debug.
Not sure if we have a unit test to test this kind of failure: content(eg. format) is incorret. Then we should not return OK, but a bad client request.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2061,4 +2070,45 @@ public ZKHelixAdmin build() {
               _zkAddress), false);
     }
   }
+
+  private Map<String, InstanceConfig> findTimeoutOfflineInstances(String clusterName,
+      Map<String, String> customizedPurgeMap) {
```
pkuwm, line 2075: From the API logic, it seems what is needed is only `timeout`? Then I think we don't have to pass along the map, but only the `timeout` value from the rest api layer: `timeout = customizedPurgeMap.get(OFFLINE_NODE_PURGE_TIMEOUT);`
Then the APIs only need the timeout param, not a map.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -282,6 +282,17 @@ public Response updateCluster(@PathParam("clusterId") String clusterId,
           return badRequest(e.getMessage());
         }
         break;
+      case purgeOfflineParticipants:
+        Map<String, String> customizedPurgeMap = new HashMap<>();
+        try {
+          customizedPurgeMap =
+              OBJECT_MAPPER.readValue(content, new TypeReference<HashMap<String, String>>() {
```
pkuwm, line 289: It seems only timeout is needed. Anther option is, we can add a QueryParam timeout, instead of putting it a body content? It'd be easier to use.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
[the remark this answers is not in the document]
zhangmeng916: We provide an option for users to define the default timeout in cluster config, so that they do not need to provide time in every call. If they do give a timeout in purge call, the value will override the default one in cluster config.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
zhangmeng916: Yeah, good catch!

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
zhangmeng916: With the input changed to Long, now it's not an issue.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -228,6 +228,13 @@ void addResource(String clusterName, String resourceName, int numPartitions, Str
    */
   void dropInstance(String clusterName, InstanceConfig instanceConfig);
 
+  /**
+   * Purge offline instances from a cluster
+   * @param clusterName
+   * @param timeout
+   */
+  void purgeOfflineInstances(String clusterName, Long timeout);
```
narendly, line 236: i think the javadoc here could use more work. we should try to add more context on why and when the interface should be used. 

also i'm not sure `timeout` is the right term for this. perhaps rename it to something like `offlineDuration`? also if you could please add a description as to how this `offlineDuration` is measured/calculated.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2061,4 +2069,43 @@ public ZKHelixAdmin build() {
               _zkAddress), false);
     }
   }
+
+  private Map<String, InstanceConfig> findTimeoutOfflineInstances(String clusterName,
```
narendly, line 2073: again, `timeout` might be a bit of a misnomer here.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
narendly: also is there a strong reason why we don't use a primitive `long`?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2061,4 +2069,43 @@ public ZKHelixAdmin build() {
...
+      Long timeout) {
```
narendly, line 2074: i'd prefer a primitive `long` here

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2061,4 +2069,43 @@ public ZKHelixAdmin build() {
...
+    Map<String, InstanceConfig> instanceConfigMap = new HashMap<>();
+    // in case there is no customized timeout value, use the one defined in cluster config
+    if (timeout == null) {
+      timeout = _configAccessor.getClusterConfig(clusterName).getOfflineNodeTimeOutForPurge();
+    }
```
narendly, line 2079: instead of doing the `null` check, maybe you could define a `NOT_SET` constant of some sort...

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2061,4 +2069,43 @@ public ZKHelixAdmin build() {
...
+      String historyPath = PropertyPathBuilder.instanceHistory(clusterName, instanceName);
+      ZNRecord znRecord = baseAccessor.get(historyPath, null, 0);
+      ParticipantHistory participantHistory = new ParticipantHistory(znRecord);
+      long lastOfflineTime = participantHistory.getLastOfflineTime();
+      if (lastOfflineTime == -1 || System.currentTimeMillis() - lastOfflineTime < timeout) {
+        toRemoveInstances.add(instanceName);
+      }
+    }
+    instanceConfigMap.keySet().removeAll(toRemoveInstances);
+    return instanceConfigMap;
```
narendly, line 2109: nit: is there a strong reason we're using a `BaseDataAccessor` instance here? for helix-related i/o, it is better taste to use `HelixDataAccessor` because you can avoid having to cast it to helix data model from znrecord. 

that way, code will be cleaner and easier to read, shorter, and more concise.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -250,6 +251,13 @@ public void dropInstance(String clusterName, InstanceConfig instanceConfig) {
     }
   }
 
+  @Override
+  public void purgeOfflineInstances(String clusterName, Long timeout) {
+    Map<String, InstanceConfig> timeoutOfflineInstances = findTimeoutOfflineInstances(clusterName
+        , timeout);
+    timeoutOfflineInstances.values().forEach(instance -> dropInstance(clusterName, instance));
```
narendly, line 258: two high-level comments:
1. are we not using the lock? as discussed offline, i think to complete this feature, it would be a good idea to use the distributed lock here.
2. what kind of failure handling are we doing in this call? we're making multiple dropInstance calls here and there's no guarantee that all would go through. plus, the method itself isn't returning anything so there's no way for the user to determine whether the purge operation was successful or not

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -113,6 +114,7 @@
   private final boolean _usesExternalZkClient;
 
   private static Logger logger = LoggerFactory.getLogger(ZKHelixAdmin.class);
+  private static final long VALUE_NOT_SET = -1;
```
narendly, line 117: sorry if i didnt make myself more clear. when i said `VALUE_NOT_SET`, we usually name these constants in a more descriptive manner - e.g.) `OFFLINE_DURATION_VALUE_NOT_SET` for example. this is ZKHelixAdmin, so it contains a lot of different operations, so it's better to make this crystal clear.

also nit: usually `long`s have an `L` appended to it. -> `-1L;`

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
narendly: 1. if this is going to be called by the user, i think it's fair to assume that there would not be race conditions (because we expect the user to do concurrency control). however, i'm still a little confused.

> The major concern from user side of having locks in purge function and in instance joining function is that they may have thousands of instances come online at the same time, if each of them tries to read and write the lock, zookeeper will have a burst, and this is dangerous.

what we ultimately want to achieve is to provide a function to do this automatically. for now, we shift the burden of calling this to the user, but we want to automate this eventually. i'm not quite getting the "dangerous" part - why is it dangerous to use zk to implement the lock? it's one of the things zk is meant for. having hundreds and thousands of zk clients with zk as the distributed lock is not unheard of? i'm not exactly understanding the situation you're describing.

2. i was initially thinking maybe it might be better to catch the exception and throw it from this function as a "purge failure", with the list of instances that were purged successfully and (optionally) the list of instances that were not due to a failure if there are any. Otherwise, there's no way to know for the user or ourselves what exactly happened. I don't think that would be too hard to implement on top of what you already have now - do you think that would help us debug this better if something goes wrong?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ClusterAccessor.java
```diff
@@ -282,6 +282,13 @@ public Response updateCluster(@PathParam("clusterId") String clusterId,
           return badRequest(e.getMessage());
         }
         break;
+      case purgeOfflineParticipants:
+        if (timeout == null || timeout < 0) {
+          helixAdmin.purgeOfflineInstances(clusterId, -1);
```
narendly, line 287: -1 is a magic number and we should generally avoid having magic numbers..

it would look nicer if you could make the not set constant a public constant and use it here - that way it's easy to track what's going on and code is much more readable.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -1008,4 +1008,58 @@ public void testUpdateCustomizedStateConfig() throws Exception {
...
+        "Instance should still be there");
+
+    // This purge will remove the instance as the customized purge timeout is met.
+    Map<String, String> purgeMap = new HashMap<>();
+    tool.purgeOfflineInstances(clusterName, 10000L);
+    Assert.assertFalse(_gZkClient.exists(keyBuilder.instanceConfig(instanceName).getPath()),
+        "Instance should already be dropped");
+
+    tool.dropCluster(clusterName);
+    System.out.println("END " + clusterName + " at " + new Date(System.currentTimeMillis()));
```
narendly, line 1063: important: should we clean up all zk resources after the test is done? e.g.) `tool.close()`

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2061,4 +2070,37 @@ public ZKHelixAdmin build() {
...
+    if (offlineDuration == VALUE_NOT_SET) {
+      offlineDuration =
+          _configAccessor.getClusterConfig(clusterName).getOfflineNodeTimeOutForPurge();
+      if (offlineDuration == VALUE_NOT_SET) {
+        return instanceConfigMap;
+      }
+    }
+
+    HelixDataAccessor accessor =
+        new ZKHelixDataAccessor(clusterName, new ZkBaseDataAccessor<ZNRecord>(_zkClient));
```
narendly, line 2087: could you double-check if this `zkclient` is a `RealmAwareZkClient`?

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
zhangmeng916: All these close parts are already handled in the aftersuite of base test class.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
zhangmeng916: Yeah, confirmed it's a RealmAwareZkClient.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
zhangmeng916: Interesting, in the other PR, there's a comment asking for generalize the NOT SET value, and that's why I modified it. I separated them into different definitions again. Let me know whether this is good.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
zhangmeng916: For question 1, Helix does not plan to integrate this purge logic into controller as it unnecessarily complicates the controller logic. Then it will be up to users to decide whether they would like take the risk for race condition, or afford the traffic in zookeeper. The concern from users was during deployment, zk is already undergoing a lot of traffic, and adding one more read/write per instance in a short time period (e.g. 10 or 20 seconds) for lock makes it worse without adding extra benefit for them. 
I think a possible future work could be Helix provides config to use lock or not during purge and instances come online, and users could choose to turn it on or off depending on their real use case.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -163,7 +165,7 @@
   private final static int MIN_REBALANCE_PREFERENCE = 0;
   public final static boolean DEFAULT_GLOBAL_REBALANCE_ASYNC_MODE_ENABLED = true;
   private static final int GLOBAL_TARGET_TASK_THREAD_POOL_SIZE_NOT_SET = -1;
-  private static final long VALUE_NOT_SET = -1;
+  private static final int  OFFLINE_NODE_TIME_OUT_FOR_MAINTENANCE_MODE_NOT_SET = -1;
```
narendly, line 168: Nit: space?

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -153,6 +153,8 @@
 
   public final static String TASK_QUOTA_RATIO_NOT_SET = "-1";
 
+  public static final long OFFLINE_TIMEOUT_FOR_PURGE_NOT_SET = -1L;
```
narendly, line 156: nit: offline "duration"

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkHelixAdmin.java
```diff
@@ -83,6 +83,8 @@
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
+import static org.apache.helix.model.ClusterConfig.OFFLINE_TIMEOUT_FOR_PURGE_NOT_SET;
```
narendly, line 86: generally we avoid static imports. you just import the class, and use it as CLASS.CONSTANT_NAME

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -250,6 +253,23 @@ public void dropInstance(String clusterName, InstanceConfig instanceConfig) {
...
+  @Override
+  public List<String> purgeOfflineInstances(String clusterName, long offlineTimeout) {
+    Map<String, InstanceConfig> timeoutOfflineInstances = findTimeoutOfflineInstances(clusterName
+        , offlineTimeout);
+    List<String> failToPurgeInstances = new ArrayList<>();
+    timeoutOfflineInstances.values().forEach(instance -> {
+      try {
+        dropInstance(clusterName, instance);
+      } catch (HelixException e) {
+        logger.info("Failed to purge instance {} in cluster {}. Exception: {}", instance,
```
narendly, line 265: info -> warn or error

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -250,6 +253,23 @@ public void dropInstance(String clusterName, InstanceConfig instanceConfig) {
...
+            clusterName, e);
+        failToPurgeInstances.add(instance.getInstanceName());
+      }
+    });
+    return failToPurgeInstances;
```
narendly, line 270: I think it's a bit awkward to return the list of instances whose deletion _failed_; usually these methods don't return anything, or they return the list of instances whose deletion _succeeded_.

how about we keep it as `void`, and only log one line at the end if there are any instances whose deletion failed? that way we persist the information we need while being concise.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
zhangmeng916: Actually I changed all other places to offline timeout. I feel `timeout` could better show the meaning for this duration.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -228,6 +228,16 @@ void addResource(String clusterName, String resourceName, int numPartitions, Str
    */
   void dropInstance(String clusterName, InstanceConfig instanceConfig);
 
+  /**
+   * Purge offline instances that have been offline for longer than the offline duration time
+   * from a cluster
+   * @param clusterName
+   * @param offlineTimeout the threshold of offline time to decide whether an offline instance
```
narendly, line 235: "the threshold of offline time to"... -> this is a little confusing. the fact that we are using timeout to describe this could be confusing to the reader. 

timeout is a term used when one actively counts down to see if an operation completes or not. in this case, as you clarified, we are not actively counting down or waiting on anything. so the use of timeout is still misleading. my suggestion is to use offlineDuration as that wouldn't cause any confusion or give readers a wrong idea.

for example, something like this would work:

`@param offlineDuration if an offline instance has been offline for longer than the set offlineDuration, the offline instance becomes eligible for being purged/deleted`

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
narendly: i add some more explanation here: https://github.com/apache/helix/pull/1494/files#r515667749

ultimately it's up to you to make the call since you're the owner, but i wanted to make it clear why i think using 'timeout' to describe this could be confusing.

narendly (approved):
please consider applying the suggested changes :)

zhangmeng916:
This PR is ready to be merged. Approved by @narendly 
Final commit message: 
Add admin and Rest API to purge instances that have been offline more than specified period of time

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -134,7 +134,7 @@
     // offline for more than this specified time period, and users call purge participant API,
     // then the node will be removed.
     // The unit is milliseconds.
-    OFFLINE_NODE_TIME_OUT_FOR_PURGE
+    OFFLINE_DURATION_FOR_PURGE
```
narendly, line 137: Optional: usually I see people do `OFFLINE_DURATION_FOR_PURGE_MS` but either way is fine, really.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -98,6 +99,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.helix.model.ClusterConfig.OFFLINE_DURATION_FOR_PURGE_NOT_SET;
```
narendly, line 102: Nit: would it be possible to avoid static imports? i think i've already commented on something similar...

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
narendly: @zhangmeng916 I actually agree with that suggestion that we should combine all NOT SET values into one. Sorry I wasn't aware that there were many other NOT SET bits floating around.

It's either or - you could combine all of them into one (just make sure they're all using -1) or leave them all separate and create a separate issue/PR for it.

zhangmeng916: I will create a new PR for standardize the NOT SET value. Seems there are a lot, and I'll modify together.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
narendly: @zhangmeng916 let's read that one a little more carefully. 

> as discussed, i think what we ultimately want to achieve is to provide a service or a sidecar controller to do this automatically

this doesn't mean that we will include the purge logic into the main controller. it clearly states that it will be a service or a sidecar controller. and i believe we all agreed that this would be the direction.

the reason we are shifting the responsibility to the user is for the ease of implementation. ultimately, we want to automate this in our holistic cluster management environment (HaaS).

> adding one more read/write per instance in a short time period (e.g. 10 or 20 seconds) for lock makes it worse without adding extra benefit for them.

I'm not sure if i completely agree with the statement. 10-20 sec is a very generous duration where zk server will be handle many read/writes. also I don't agree that there's no extra benefit. the benefit we're talking about here is concurrency control in a distributed setting? i'm not suggesting we add read/write operations for nothing. 

if you prefer to leave out locking for this API (which is totally fine with the right assumptions), i suggest we give some context in the JavaDoc and clearly state that this operation is not thread-safe and that it _should_ lock in the future when called automatically.

PR #1497, opened 2020-10-29
Fix TestRoutingTableProviderPeriodicRefresh.

### LINKED ISSUE #1250: Fix flaky test TestRoutingTableProviderPeriodicRefresh

jiajunwang:
### Issues

This is to mitigate #1250, but we still need a good complete fix.

### Description

The test was unstable and do not clean up the environment one the failure. This PR tolerates more test results to be accepted as a good result.
Also, enhance the cleanup logic to prevent leakages.

### Tests

TestRoutingTableProviderPeriodicRefresh passed

NA, pure test changes.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
@@ -21,7 +21,6 @@ package org.apache.helix.integration.spectator;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
@@ -29,7 +28,6 @@ import java.util.Map;
 import org.apache.helix.HelixManager;
 import org.apache.helix.HelixManagerFactory;
 import org.apache.helix.InstanceType;
-import org.apache.helix.NotificationContext;
 import org.apache.helix.PropertyType;
 import org.apache.helix.TestHelper;
 import org.apache.helix.common.ZkTestBase;
@@ -37,7 +35,6 @@ import org.apache.helix.integration.manager.ClusterControllerManager;
 import org.apache.helix.integration.manager.MockParticipantManager;
 import org.apache.helix.model.BuiltInStateModelDefinitions;
 import org.apache.helix.model.CurrentState;
-import org.apache.helix.model.CustomizedView;
 import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.LiveInstance;
@@ -49,6 +46,7 @@ import org.testng.Assert;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
+
 
 public class TestRoutingTableProviderPeriodicRefresh extends ZkTestBase {
   private static final org.slf4j.Logger logger =
@@ -63,6 +61,8 @@ public class TestRoutingTableProviderPeriodicRefresh extends ZkTestBase {
 
   private static final int PARTITION_NUMBER = 20;
   private static final int REPLICA_NUMBER = 3;
+
+  private static final long REFRESH_PERIOD_MS = 1000L;
 
   private HelixManager _spectator;
   private HelixManager _spectator_2;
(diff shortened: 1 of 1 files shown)


helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
```diff
@@ -203,34 +208,32 @@ synchronized int getRefreshCount() {
...
-    Thread.sleep(1000L);
+    // Wait to ensure that the initial refreshes finish (not triggered by periodic refresh timer)
+    Thread.sleep(REFRESH_PERIOD_MS * 2);
 
     // Test short refresh
     int prevRefreshCount = _routingTableProvider.getRefreshCount();
-    // Wait for one timer duration
-    Thread.sleep(1000L);
-    // The timer should have gone off, incrementing the refresh count
-    Assert.assertEquals(_routingTableProvider.getRefreshCount(), prevRefreshCount + 1);
```
NealSun96, line 214: Could you elaborate on the reason of failure before? Is it because periodic refresh sometimes take a bit more than 1000, so it's possible that `newRefreshCount == prevRefreshCount`; or, is it because periodic refresh may take less time, so it's possible that `newRefreshCount == prevRefreshCount + 2`?

helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
```diff
@@ -203,34 +208,32 @@ synchronized int getRefreshCount() {
...
+    // Wait for more than one timer duration
+    Thread.sleep((long) (REFRESH_PERIOD_MS * 1.5));
+    // The timer should have gone off, incrementing the refresh count by one or two depends on the
+    // timing.
+    int newRefreshCount = _routingTableProvider.getRefreshCount();
+    Assert.assertTrue(
+        newRefreshCount == prevRefreshCount + 1 || newRefreshCount == prevRefreshCount + 2);
```
NealSun96, line 222: While this eliminates flakiness, would it be too loose to ensure correctness?

NealSun96 (changes requested):

helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
jiajunwang: I guess it is just because of some timing issues. When you set up a time window that equals the interval, you might have one or two refreshes happen.

helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
jiajunwang: The java schedule task method that we are using is not accurate anyway. Moreover, the main test purpose here is to ensure the refresh happens. If it is faster a little bit, I guess we are still good.
On the other hand, if the refresh does respect the interval, so during 1.5 seconds, it happens more than 2 times, then the test will still fail.

helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
```diff
@@ -64,6 +62,8 @@
   private static final int PARTITION_NUMBER = 20;
   private static final int REPLICA_NUMBER = 3;
 
+  private static final long REFRESH_PERIOD_MS = 1000L;
```
lei-xia, line 65: why 1000ms, could 500ms, or even 100ms work? That will reduce the test time.

helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
NealSun96: This comment is tied with my other comment. If we already know it's flaky between 1 or 2 refreshes, why don't we just `assert(+1 || +2)`, while keeping `sleep(1000)`? 

If we change it to `sleep(1500)`, we are opening doors to "slow refresh" cases. If a refresh doesn't respect the refresh period, and instead took around 1500, the previous version of the test will catch it, but now it won't. Like you said, maybe even 1500 is fine since there is no hard requirement, we are just ensuring the refresh happens, but it feels to me that correctness check is relaxed.

helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
jiajunwang: Works in theory. But I prefer to keep the main test logic the same. I don't have a strong motivation to optimize this test since as Kai mentioned, we shall rewrite it eventually. Just want to make it work first.

helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
jiajunwang: As I replied to Lei, this PR is not to improve the test. This is mainly for unblocking the other devs since the test cannot pass. Alternatively, we can disable it. But I prefer the current version. At least it checks some of the logic.

We will need to address this by re-writing the test logic.

alirezazamani (approved):

jiajunwang:
Approved by @alirezazamani , I will merge this PR to unblock the test. But the issue is kept opening, since we still need a better solution.

helix-core/src/test/java/org/apache/helix/integration/spectator/TestRoutingTableProviderPeriodicRefresh.java
NealSun96: Talked offline: since periodic refresh can happen slightly less, equally, or slightly more frequent than the configured wait period, I agree with @jiajunwang 's change to use both `sleep(1500)` and `assert(+1 || +2)`.

PR #1499, opened 2020-10-30
Trim strings in znrecord fields for REST json content

### LINKED ISSUE #1498: Unexpected prefix space in value field on Helix UI
### Describe the bug
When we configure values on helix UI, we find that a prefix space is added to the expected value. It causes some issues because the values stored in ZK are unexpected with the prefix space.

Example: we expect to set a capacity key `[partition]`, but in ZK we find that `[ partition]` is set.
 
### To Reproduce
Configure the capacity value as below.
image

### Expected behavior
No space should be added as prefix.

pkuwm:
### Issues

Fixes #1498 

### Description

If users accidentally put a leading space in the json payloads and call helix rest to update cluster configs, the leading space is not correctly handled and causes configs not to be appropriately parsed. Then it leads to some issues in helix that relies on the configs.

This PR trims strings in the znrecords fields after the REST API's json payloads are deserialized to a znrecord.

### Tests

TestZNRecordUtil#testTrimFields

```
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestClusterAccessor.testUpdateRESTConfig:718->AbstractTestClass.post:529 expected:<200> but was:<500>
[INFO]
[ERROR] Tests run: 170, Failures: 1, Errors: 0, Skipped: 1
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time:  01:46 min
[INFO] Finished at: 2020-10-29T23:12:16-07:00
```
### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/AbstractHelixResource.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/AbstractHelixResource.java
@@ -33,7 +33,7 @@ import org.apache.helix.tools.ClusterSetup;
 import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.impl.client.ZkClient;
-
+import org.apache.helix.zookeeper.util.ZNRecordUtil;
 
 /**
  * This class provides methods to access Helix specific objects
@@ -83,7 +83,8 @@ public class AbstractHelixResource extends AbstractResource {
 
   protected static ZNRecord toZNRecord(String data)
       throws IOException {
-    return OBJECT_MAPPER.reader(ZNRecord.class).readValue(data);
+    ZNRecord record = OBJECT_MAPPER.readValue(data, ZNRecord.class);
+    return ZNRecordUtil.trimFields(record);
   }
 
   private ServerContext getServerContext() {
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/ZNRecordUtil.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/ZNRecordUtil.java
@@ -19,6 +19,10 @@ package org.apache.helix.zookeeper.util;
  * under the License.
  */
 
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
 import org.apache.helix.zookeeper.constant.ZkSystemPropertyKeys;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 
@@ -61,5 +65,32 @@ public class ZNRecordUtil {
     }
 
     return writeSizeLimit;
+  }
+
+  /**
+   * Trims leading/trailing spaces for all strings in Simple/List/Map fields of a {@link ZNRecord}.
+   *
+   * @param record the {@link ZNRecord} to be trimmed.
+   * @return a {@link ZNRecord} that has been trimmed: strings in its fields are trimmed.
+   */
+  public static ZNRecord trimFields(ZNRecord record) {
+    Map<String, String> simpleFields = record.getSimpleFields().entrySet().stream()
+        .collect(Collectors.toMap(e -> e.getKey().trim(), e -> e.getValue().trim()));
+
+    Map<String, List<String>> listFields = record.getListFields().entrySet().stream()
+        .collect(Collectors.toMap(e -> e.getKey().trim(),
+            e -> e.getValue().stream().map(String::trim).collect(Collectors.toList())));
+
+    Map<String, Map<String, String>> mapFields = record.getMapFields().entrySet().stream()
+        .collect(Collectors.toMap(e1 -> e1.getKey().trim(),
+            e1 -> e1.getValue().entrySet().stream()
+                .collect(Collectors.toMap(e2 -> e2.getKey().trim(), e2 -> e2.getValue().trim()))));
+
+    ZNRecord trimmedRecord = new ZNRecord(record);
+    trimmedRecord.setSimpleFields(simpleFields);
+    trimmedRecord.setListFields(listFields);
+    trimmedRecord.setMapFields(mapFields);
+
+    return trimmedRecord;
   }
 }
--- a/zookeeper-api/src/test/java/org/apache/helix/zookeeper/util/TestZNRecordUtil.java
+++ b/zookeeper-api/src/test/java/org/apache/helix/zookeeper/util/TestZNRecordUtil.java
@@ -1,0 +1,46 @@
+package org.apache.helix.zookeeper.util;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.testng.Assert;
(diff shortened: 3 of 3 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/AbstractHelixResource.java
```diff
@@ -83,7 +83,8 @@ public HelixDataAccessor getDataAccssor(String clusterName) {
 
   protected static ZNRecord toZNRecord(String data)
       throws IOException {
-    return OBJECT_MAPPER.reader(ZNRecord.class).readValue(data);
+    ZNRecord record = OBJECT_MAPPER.readValue(data, ZNRecord.class);
+    return ZNRecordUtil.trimFields(record);
```
jiajunwang, line 87: Here's an interesting problem.
Since the rest is not covering all of our APIs, so assuming there is a user use java API sets " test" = 123. Then in the UI, user got "test" = 123 (due to this trim). Then user modifies it to be "test" = 456. After this change, in ZK, we now have both keys.

" test" = 123
"test" = 456

To who reads the ZK directly, he or she does not know which one is the right value.
To who reads it from UI, he or she will either get a duplicate key exception or see a random value depends on which value overwrites the other one.

I think this problem is hard to address.
That's why I think we still should try to fix the UI : )

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/ZNRecordUtil.java
```diff
@@ -62,4 +66,31 @@ public static int getSerializerWriteSizeLimit() {
...
+            e1 -> e1.getValue().entrySet().stream()
+                .collect(Collectors.toMap(e2 -> e2.getKey().trim(), e2 -> e2.getValue().trim()))));
+
+    ZNRecord trimmedRecord = new ZNRecord(record);
+    trimmedRecord.setSimpleFields(simpleFields);
+    trimmedRecord.setListFields(listFields);
+    trimmedRecord.setMapFields(mapFields);
+
+    return trimmedRecord;
+  }
```
narendly, line 95: Two comments:
1. technically spaces are allowed in names I believe? sure it's ugly, but we haven't enforced the "no-leading/trailing spaces" rule, so these are valid strings as names.
2. this is probably more minor, but doing trimming this way right here might not be so ideal because this might increase gc pressure / heap usage.

all in all, I am also of the opinion that we enforce this at the UI level. this is users' responsibility to make sure their input is correct after all.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/AbstractHelixResource.java
pkuwm: @jiajunwang Thanks for the comment. There are many things to consider regarding this consistent problem

In the problem you described, two entries " test" = 123, "test" = 456 will show on UI. Then the user will see them and tries to fix it: removing one of them.

1. There are many ways to change the znode value: UI, rest API, java API, zooInspector, zk client, etc.. We cannot prevent users from setting the value with a leading space either by mistake or on purpose. 

2. Regarding leading/trailing spaces, I don't think they make sense, at least 99% are useless and human errors and cause problems. I wonder why don't we make an agreement: either we enforce to trim the strings at a layer(REST/Java API before writing to zk, or when reading the configs to tolerate spaces, because we cannot enforce it for all the ways that can change znodes), or just leave it as is, requiring the end users to be responsible for it (then when there is an issue, they would still come to us for help troubleshoot).

3. From the past experience, the mistakes are usually made by UI or REST API. We cannot prevent this from all layers, but considering the pain it brought to us, we can make tradeoffs and do better for 99% cases that use UI/REST and release us. Then if we really encounter the 1% case as you said, the duplicate configs will show on UI as " test" = 123, "test" = 456. At least the problem is exposed earlier, then let the user resolve the problem.

4. I still doubt UI has the bug: auto adding leading spaces, because we could not reproduce it and I could not find any code that would add the leading space. If it has the bug, I think this is what you meant by `fix the UI`, I completely agree that we fix it. But if it doesn't, trimming is just an improvement. Even we do the improvement on UI, we cannot prevent from other ways. There'll still be the problem as you said.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/ZNRecordUtil.java
pkuwm: 1. more discussion is in this thread: https://github.com/apache/helix/pull/1499/files#r515354253
2. At rest api layer, when this `toZNRecord()` is called, it is usually creating/updating configs. I don't think the QPS for updating configs is high at all. So I wouldn't worry about gc/heap pressure.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/AbstractHelixResource.java
jiajunwang: You mean that we only control the write path, but not the read path, right? In this case, if the user doesn't want to fix them, but decide to change a value first. Then due to the same logic, when they update the value of " test" key, their request will be really applied to "test" key. I think it might introduce bugs or at least confusion. Unless we **block any update when ambiguous keys are found**. It might be a good way.

It is true that we cannot prevent all random write.
My thought is that, eventually, we need to lock down ZK write and only allow Helix REST access. Then we can do whatever you described above.
Before that, instead of introducing a complicated logic that won't be 100% working in all conditions, **just high-light the suspicious keys in UI and it shall alert the user and the admin can fix it fast**.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/util/ZNRecordUtil.java
narendly: 1. I don't think that thread is answering my question. my original question is, who decided " abcname" or "abcname " were invalid? i'm just playing the devil's advocate here - it's perfectly fine for a user to create a partition named "xxx  ", for example?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/AbstractHelixResource.java
narendly: Here's an idea. The current implementation is fixing & modifying the user-provided input _silently_ (without informing the user) to make the input conform to what we think is the right thing. This is a hidden assumption that will inevitably become harder to track and will leave both the user and Helix maintainers confused at one point in the future.

Why don't we just fail it if the input is invalid? This way, the user will know, and as some reviewers pointed out above, we do not run the risk of writing to an existing value causing unwanted writes.

pkuwm: > You mean that we only control the write path, but not the read path, right? In this case, if the user doesn't want to fix them, but decide to change a value first. Then due to the same logic, when they update the value of " test" key, their request will be really applied to "test" key. I think it might introduce bugs or at least confusion. Unless we **block any update when ambiguous keys are found**. It might be a good way.
> 
> 
> 
> It is true that we cannot prevent all random write.
> 
> My thought is that, eventually, we need to lock down ZK write and only allow Helix REST access. Then we can do whatever you described above.
> 
> Before that, instead of introducing a complicated logic that won't be 100% working in all conditions, **just high-light the suspicious keys in UI and it shall alert the user and the admin can fix it fast**.

1. I actually was thinking we just warn users that the field has leading/trailing spaces. Then on UI, users know the info from the warning. But that only applies to UI, not to REST. And it is actually an improvement/feature, not a bug fix. It'll need some JavaScript work on ui. I think we can definitely do the improvement if we have a UI engineer. I was just concerned about that currently we don't have he capacity for the improvement. If someone can implement this feature, it'll help users find out the potential incorrect configs on UI.

2. At rest layer with this, at least we can avoid 99% of human errors that introduce the spaces. Like I said, I am not sure if a user finds " test" exists, why they still want to keep it and update its value, rather than use the appropriate one "test". In normal case, if they update " test" and then find that there are two entries:" test" and "test", they would just delete one of them, most possibly " test".

kaisun2000 (approved):
There are many good discussions. But unless we make some actions like this first or we can fix the UI issue quickly. There can be further production issues, which would take the team a lot of effort trouble shoot.

 @pkuwm, can you have a look to see if UI fix can be done quickly. I will give an approve for now.

jiajunwang:
@pkuwm I feel the PR worths a discussion. Could you please schedule a quick discussion so we can bring everyone onto the same page?

pkuwm:
> There are many good discussions. But unless we make some actions like this first or we can fix the UI issue quickly. There can be further production issues, which would take the team a lot of effort trouble shoot.
> 
> @pkuwm, can you have a look to see if UI fix can be done quickly. I will give an approve for now.

@kaisun2000 @jiajunwang Thanks for the review! I still doubt the UI auto adds a leading space. We'll discuss more to determine a fair solution to improve it, eg: validate the the input and give warning/bad request response, or trim the spaces like this PR, etc..

PR #1500, opened 2020-10-30
Refine the message handler error handling logic to avoid unnecessary retry.

### LINKED ISSUE #1488: Participant side message handling logic may fail silently

jiajunwang:
### Issues

Resolve #1488 

### Description

This PR enhances the message event processing logic to prevent silent failure and unnecessary retry.
1. If creating a message handler fails unexpectedly (meaning there is an Exception), then the message will be marked as UNPROCESSABLE unless the message is sent with a retry count larger than 0. When the retry count is configured, then before the message runs out of the retry count, the participant will keep retrying on any message callbacks.
2. The UNPROCESSABLE message, which is generated due to the previous point, will be left in the participant message folder and not automatically removed. This is to prevent unnecessary retry.
3. If the message handler fails due to the participant cannot schedule the task, then the message will be discarded. If the message is a state transition message, then the corresponding state model and the partition current state will be set to ERROR. This is also to prevent unnecessary retry.

### Tests

TestHelixTaskExecutor

[WARNING] Tests run: 1237, Failures: 0, Errors: 0, Skipped: 1
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  02:00 h
[INFO] Finished at: 2020-10-29T22:59:19-07:00

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -613,7 +613,12 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
 
       MsgHandlerFactoryRegistryItem item = _hdlrFtyRegistry.get(msgType);
       if (item.factory() != null) {
-        item.factory().reset();
+        try {
+          item.factory().reset();
+        } catch (Exception ex) {
+          LOG.error("Failed to reset the factory {} of message type {}.", item.factory().toString(),
+              msgType);
+        }
       }
     }
     // threads pool specific to STATE_TRANSITION.Key specific pool are not shut down.
@@ -788,8 +793,8 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
     List<MessageHandler> nonStateTransitionHandlers = new ArrayList<>();
     List<NotificationContext> nonStateTransitionContexts = new ArrayList<>();
 
-    // message read
-    List<Message> readMsgs = new ArrayList<>();
+    // message to be updated in ZK
+    List<Message> updatingMsgs = new ArrayList<>();
 
     String sessionId = manager.getSessionId();
     List<String> curResourceNames =
@@ -804,40 +809,66 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
         // skip the following operations for the no-op messages.
         continue;
       }
-      // create message handlers, if handlers not found, leave its state as NEW
+
       NotificationContext msgWorkingContext = changeContext.clone();
+      MessageHandler msgHandler = null;
       try {
-        MessageHandler msgHandler = createMessageHandler(message, msgWorkingContext);
-        if (msgHandler == null) {
-          // Failed to create message handler, skip processing this message in this callback.
-          // The same message process will be retried in the next round.
-          continue;
-        }
-        if (message.getMsgType().equals(MessageType.STATE_TRANSITION.name()) || message.getMsgType()
-            .equals(MessageType.STATE_TRANSITION_CANCELLATION.name())) {
-          if (validateAndProcessStateTransitionMessage(message, instanceName, manager,
-              stateTransitionHandlers, msgHandler)) {
-            // Need future process by triggering state transition
-            String msgTarget =
-                getMessageTarget(message.getResourceName(), message.getPartitionName());
-            stateTransitionHandlers.put(msgTarget, msgHandler);
-            stateTransitionContexts.put(msgTarget, msgWorkingContext);
-          } else {
-            // skip the following operations for the invalid/expired state transition messages.
-            continue;
-          }
+        // create message handlers, if handlers not found but no exception, leave its state as NEW
+        msgHandler = createMessageHandler(message, msgWorkingContext);
+      } catch (Exception ex) {
+        // Failed to create message handler and there is an Exception.
+        int remainingRetryCount = message.getRetryCount();
+        LOG.error(
+            "Exception happens when creating Message Handler for message {}. Current remaining retry count is {}.",
+            message.getMsgId(), remainingRetryCount);
+        // Set the message retry count to avoid infinite retrying.
+        message.setRetryCount(remainingRetryCount - 1);
+        message.setExecuteSessionId(sessionId);
+        // continue processing in the next section where handler object is double-checked.
+      }
+
+      if (msgHandler == null) {
+        if (message.getRetryCount() < 0) {
+          // If no more retry count remains, then mark the message to be UNPROCESSABLE.
+          String errorMsg = String
+              .format("Message %s has a negative remaining retry count %d. Stop processing it!",
+                  message.getMsgId(), message.getRetryCount());
+          updateUnprocessableMessage(message, null, errorMsg, manager);
+          updatingMsgs.add(message);
         } else {
-          // Need future process non state transition messages by triggering the handler
-          nonStateTransitionHandlers.add(msgHandler);
-          nonStateTransitionContexts.add(msgWorkingContext);
+          // Skip processing this message in this callback. The same message process will be retried
+          // in the next round.
+          LOG.warn("There is no existing handler for message {}."
+                  + " Skip processing it for now. Will retry on the next callback.",
+              message.getMsgId());
         }
-      } catch (Exception e) {
-        handleUnprocessableMessage(message, e, e.getMessage(), accessor, instanceName, manager);
         continue;
       }
 
-      // Update the processed message objects
-      readMsgs.add(markReadMessage(message, msgWorkingContext, manager));
+      if (message.getMsgType().equals(MessageType.STATE_TRANSITION.name()) || message.getMsgType()
+          .equals(MessageType.STATE_TRANSITION_CANCELLATION.name())) {
+        if (validateAndProcessStateTransitionMessage(message, manager, stateTransitionHandlers,
+            msgHandler)) {
+          // Need future process by triggering state transition
+          String msgTarget =
+              getMessageTarget(message.getResourceName(), message.getPartitionName());
+          stateTransitionHandlers.put(msgTarget, msgHandler);
+          stateTransitionContexts.put(msgTarget, msgWorkingContext);
+        } else {
+          // Skip the following operations for the invalid/expired state transition messages.
+          // Also remove the message since it might block the other state transition messages.
+          removeMessageFromZK(accessor, message, instanceName);
+          continue;
+        }
+      } else {
+        // Need future process non state transition messages by triggering the handler
+        nonStateTransitionHandlers.add(msgHandler);
+        nonStateTransitionContexts.add(msgWorkingContext);
+      }
+
+      // Update the normally processed messages
+      updatingMsgs.add(markReadMessage(message, msgWorkingContext, manager));
+
       // batch creation of all current state meta data
       // do it for non-controller and state transition messages only
       if (!message.isControlerMsg() && message.getMsgType()
@@ -869,25 +900,34 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
       try {
         accessor.createChildren(createCurStateKeys, metaCurStates);
       } catch (Exception e) {
-        LOG.error("fail to create cur-state znodes for messages: " + readMsgs, e);
+        LOG.error("fail to create cur-state znodes for messages: " + updatingMsgs, e);
       }
     }
 
-    // update message state to READ in batch and schedule tasks for all read messages
-    if (readMsgs.size() > 0) {
-      updateMessageState(readMsgs, accessor, instanceName);
+    // update message state in batch and schedule tasks for all read messages
+    updateMessageState(updatingMsgs, accessor, instanceName);
 
-      for (Map.Entry<String, MessageHandler> handlerEntry : stateTransitionHandlers.entrySet()) {
-        MessageHandler handler = handlerEntry.getValue();
-        NotificationContext context = stateTransitionContexts.get(handlerEntry.getKey());
-        scheduleTaskForMessage(instanceName, accessor, handler, context);
+    for (Map.Entry<String, MessageHandler> handlerEntry : stateTransitionHandlers.entrySet()) {
+      MessageHandler handler = handlerEntry.getValue();
+      NotificationContext context = stateTransitionContexts.get(handlerEntry.getKey());
+      if (!scheduleTaskForMessage(instanceName, accessor, handler, context)) {
+        try {
+          // Record error state to the message handler.
+          handler.onError(new HelixException(String
+                  .format("Failed to schedule the task for executing message handler for %s.",
+                      handler._message.getMsgId())), MessageHandler.ErrorCode.ERROR,
+              MessageHandler.ErrorType.FRAMEWORK);
+        } catch (Exception ex) {
+          LOG.error("Failed to trigger onError method of the message handler for {}",
+              handler._message.getMsgId(), ex);
+        }
       }
+    }
 
-      for (int i = 0; i < nonStateTransitionHandlers.size(); i++) {
-        MessageHandler handler = nonStateTransitionHandlers.get(i);
-        NotificationContext context = nonStateTransitionContexts.get(i);
-        scheduleTaskForMessage(instanceName, accessor, handler, context);
-      }
+    for (int i = 0; i < nonStateTransitionHandlers.size(); i++) {
+      MessageHandler handler = nonStateTransitionHandlers.get(i);
+      NotificationContext context = nonStateTransitionContexts.get(i);
+      scheduleTaskForMessage(instanceName, accessor, handler, context);
     }
   }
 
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -788,8 +793,8 @@ public void onMessage(String instanceName, List<Message> messages,
     List<MessageHandler> nonStateTransitionHandlers = new ArrayList<>();
     List<NotificationContext> nonStateTransitionContexts = new ArrayList<>();
 
-    // message read
-    List<Message> readMsgs = new ArrayList<>();
+    // message to be updated in ZK
+    List<Message> updatingMsgs = new ArrayList<>();
```
narendly, line 797: this is a little confusing. what's wrong with leaving it as `readMsgs`? `updatingMsgs` is a confusing name since messages don't update themselves. 

or `msgsToBeHandled` or `msgsToHandle` could also be as equally descriptive

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -804,40 +809,66 @@ public void onMessage(String instanceName, List<Message> messages,
...
+          updateUnprocessableMessage(message, null, errorMsg, manager);
+          updatingMsgs.add(message);
         } else {
-          // Need future process non state transition messages by triggering the handler
-          nonStateTransitionHandlers.add(msgHandler);
-          nonStateTransitionContexts.add(msgWorkingContext);
+          // Skip processing this message in this callback. The same message process will be retried
+          // in the next round.
+          LOG.warn("There is no existing handler for message {}."
+                  + " Skip processing it for now. Will retry on the next callback.",
```
narendly, line 842: is this log absolutely necessary? wouldn't it be sufficient to log when we've used up all the retry counts? i'm afraid this might just end up polluting the log

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -804,40 +809,66 @@ public void onMessage(String instanceName, List<Message> messages,
...
+        message.setRetryCount(remainingRetryCount - 1);
+        message.setExecuteSessionId(sessionId);
+        // continue processing in the next section where handler object is double-checked.
+      }
+
+      if (msgHandler == null) {
+        if (message.getRetryCount() < 0) {
+          // If no more retry count remains, then mark the message to be UNPROCESSABLE.
+          String errorMsg = String
+              .format("Message %s has a negative remaining retry count %d. Stop processing it!",
```
narendly, line 834: nit: instead of saying "negative retry count", what you really meant is "this message has been retried for "retryCount" times"

also i'd print the `msgHandler` here. although we could deduce this info from the code, we might want to indicate that msgHandler is null in the log message as well for faster debugging

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: readMsgs is no longer an accurate name with this change.
How about msgsToBeUpdated ?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: Not true for "this message has been retried for "retryCount" times". Since at this place, we don't know how many times this message has been retried. It is not possible to print this information out.

I can add the information that the handler is not available in the error message to make it self-containing.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -804,40 +809,66 @@ public void onMessage(String instanceName, List<Message> messages,
...
+        msgHandler = createMessageHandler(message, msgWorkingContext);
+      } catch (Exception ex) {
+        // Failed to create message handler and there is an Exception.
+        int remainingRetryCount = message.getRetryCount();
+        LOG.error(
+            "Exception happens when creating Message Handler for message {}. Current remaining retry count is {}.",
+            message.getMsgId(), remainingRetryCount);
+        // Set the message retry count to avoid infinite retrying.
+        message.setRetryCount(remainingRetryCount - 1);
+        message.setExecuteSessionId(sessionId);
```
narendly, line 826: are we writing this back to zk or is this for in-memory use only? if we're writing back, would that be an added amount of zk writes/updates that we should worry about? 

should be possible to track remaining retry counts purely in-memory per cached message using message id.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: When there is no exception, but the createHandler returns null, we don't have any other output except for this line. So I think it is necessary.
If we see many of this line, then the application logic in participants might have some serious issue, and we shall fix it.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: It is a good point.

There were some discussions earlier yesterday about whether we want this retry count. So the decision is that we want to allow several retries, but would be a very small number. Note that this PR won't really add retry to state transition messages, since the default value of retry count is 0. We will tackle this later in the other PR.

Now regarding if we want to use an in-memory object instead of ZK, it is totally doable. But I don't prefer doing so for 2 reasons.
1. ZK write won't make it worse. The retry is not infinite after this change. Compare with the current behavior that we keep retrying infinitely, the new retry logic is much safer. And we can set 0 to the retry count, so no retry at all. It is purely optional.
2. In-memory count won't work if the participant instance is reset. If the retry count is really large, then in some corner case (like frequent participant reset) the retry will never stop.

In addition, the in-memory requires us to track the messages, it does not exist on the participant side. So definitely will increase the complicity of code. There will be more corner cases that we need to consider. Like memory leakage, etc. Given the benefit of doing so is not critical, I prefer not to choose this option.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -804,40 +809,67 @@ public void onMessage(String instanceName, List<Message> messages,
...
+      }
+
+      if (msgHandler == null) {
+        if (message.getRetryCount() < 0) {
+          // If no more retry count remains, then mark the message to be UNPROCESSABLE.
+          String errorMsg = String
+              .format("No available message Handler found!"
+                      + " Stop processing message %s since it has a negative remaining retry count %d!",
+                  message.getMsgId(), message.getRetryCount());
+          updateUnprocessableMessage(message, null, errorMsg, manager);
```
zhangmeng916, line 837: Seems even there is no exception, you will still set the message as unprocessable after retry instead of leaving it as NEW?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -613,7 +613,12 @@ void reset() {
 
       MsgHandlerFactoryRegistryItem item = _hdlrFtyRegistry.get(msgType);
       if (item.factory() != null) {
-        item.factory().reset();
+        try {
+          item.factory().reset();
+        } catch (Exception ex) {
```
zhangmeng916, line 618: I don't quite understand why we swallow exception here. This exception means user's reset has an issue, and can we just let the error go?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -993,67 +1034,78 @@ private boolean checkAndProcessNoOpMessage(Message message, String instanceName,
...
+      }
+      if (stateTransitionHandlers.containsKey(messageTarget)) {
+        // If there are 2 messages in same batch about same partition's state transition,
+        // the later one is discarded
+        Message duplicatedMessage = stateTransitionHandlers.get(messageTarget)._message;
+        String errMsg = String.format(
+            "Duplicated state transition message: %s. Existing: %s->%s; New (Discarded): %s->%s",
+            message.getMsgId(), duplicatedMessage.getFromState(), duplicatedMessage.getToState(),
+            message.getFromState(), message.getToState());
+        updateUnprocessableMessage(message, null /* exception */, errMsg, manager);
```
zhangmeng916, line 1080: I think we changed the original behavior, and never remove messages in this case. There is a mismatch in the comment. But, is this what we would like to do? not removing duplicated message, but only mark it? Seems we leave many unnecessary messages on ZK.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
xyuanlu: Just wondering, is it better to log the backtrace as well for better debuggability?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: This is not true. The default retry count is 0. And we only reduce it if an exception happens. So if there is no exception, then it will keep retrying.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: This is only called when we clean up. All the paths lead to the termination of the executor. The only result of not swallowing the exception is that the later cleanup steps will not happen.

Backtrace is a good point. I think I missed it unintentionally. Thanks for pointing this out.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: I moved the removing message logic to the caller. So for this specific case, it will keep the original behavior.

This change is for reusing the updateUnprocessableMessage in the other conditions. In addition, I also want to avoid silently removing a message in a very deep private call. So the new logic puts all the message processing (update or remove) in a smaller scope.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -993,67 +1034,78 @@ private boolean checkAndProcessNoOpMessage(Message message, String instanceName,
...
+        String taskId = _messageTaskMap.get(messageTarget);
+        Message msg = _taskMap.get(taskId).getTask().getMessage();
+        // If there is another state transition for same partition is going on,
+        // discard the message. Controller will resend if this is a valid message
+        String errMsg = String.format(
+            "Another state transition for %s:%s is in progress with msg: %s, p2p: %s, read: %d, current:%d. Discarding %s->%s message",
+            message.getResourceName(), message.getPartitionName(), msg.getMsgId(),
+            msg.isRelayMessage(), msg.getReadTimeStamp(), System.currentTimeMillis(),
+            message.getFromState(), message.getToState());
+        updateUnprocessableMessage(message, null /* exception */, errMsg, manager);
```
dasahcc, line 1059: This changes the behavior. If the message is not exact state we are targeting, it should be a intermediate state, we should throw away this message instead of let it hanging there. Otherwise, we will face a lot of human involved operations.

For example, for the p2p arrived but timedout, the message will be discarded. But if you hang it as unprocessable, the partition will be stuck.

kaisun2000:
>If creating a message handler fails unexpectedly (meaning there is an Exception), then the message will be marked as UNPROCESSABLE unless the message is sent with a retry count larger than 0. When the retry count is configured, then before the message runs out of the retry count, the participant will keep retrying on any message callbacks

if the retry counts is large (say people put -1 instead 0), this is effectively retry forever. How can we make sure this is not going to happen?

kaisun2000:
>The UNPROCESSABLE message, which is generated due to the previous point, will be left in the participant message folder and not automatically removed. This is to prevent unnecessary retry.

Let say later the participant resolve issue and can create the messageHandler for this partition replica. Then does it mean when they recover, they need to manually (or via rest) to remove this UNPROCESSED message?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -804,40 +809,67 @@ public void onMessage(String instanceName, List<Message> messages,
...
+            "Exception happens when creating Message Handler for message {}. Current remaining retry count is {}.",
+            message.getMsgId(), remainingRetryCount);
+        // Set the message retry count to avoid infinite retrying.
+        message.setRetryCount(remainingRetryCount - 1);
+        message.setExecuteSessionId(sessionId);
+        // continue processing in the next section where handler object is double-checked.
+      }
+
+      if (msgHandler == null) {
+        if (message.getRetryCount() < 0) {
```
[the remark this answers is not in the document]
jiajunwang: Good catch! I guess I tried to simplify the code too much. Let me fix it.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: Please refer to what I reply to Meng earlier for the same code. I just move the remove logic a little bit for reusing the code. The messages will still be removed. Please check line 861.

Let me know if you still think something missed here. I agree this is very critical, and I also want to ensure nothing changed here. So please help to double-check for me : )

jiajunwang:
> if the retry counts is large (say people put -1 instead 0), this is effectively retry forever. How can we make sure this is not going to happen?

@kaisun2000 What do you mean? If it is -1, then the current logic will try once and stop retrying. And according to what Junkai mentioned, I would like to check for the count even before the first try. So that should not be the case, right?

> Let say later the participant resolve issue and can create the messageHandler for this partition replica. Then does it mean when they recover, they need to manually (or via rest) to remove this UNPROCESSED message? If so, do we have this REST API?

This is the plan. We have discussed this in the previous standup meeting. That task will be addressed separately. Before it is ready, we will require the application to reset the participant.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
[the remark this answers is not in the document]
jiajunwang: I tried to make the change. But then realize it might be a non-backward compatible change.
Overall, we only want to change the behavior IF the handler creating fails. But if we change like what @dasahcc mentioned, then the behavior is changed even if the handler creating logic will work fine. Given the send message interface is public, it is possible that the application configures -1 retry now.

In addition, the logic that uses this timeout before this PR is also executing the task before look at this value. So I think we can keep the same behavior here.

kaisun2000:
> > if the retry counts is large (say people put -1 instead 0), this is effectively retry forever. How can we make sure this is not going to happen?
> 
> @kaisun2000 What do you mean? If it is -1, then the current logic will try once and stop retrying. And according to what Junkai mentioned, I would like to check for the count even before the first try. So that should not be the case, right?
> 

Let us say if the message have a retry count of INT_MAX, this is not going to work. I think we can be a little bit more conservative that if the message have a retry count larger than a threshold, we just retry the threshold value. 


> > Let say later the participant resolve issue and can create the messageHandler for this partition replica. Then does it mean when they recover, they need to manually (or via rest) to remove this UNPROCESSED message? If so, do we have this REST API?
> 
> This is the plan. We have discussed this in the previous standup meeting. That task will be addressed separately. Before it is ready, we will require the application to reset the participant.

jiajunwang:
> Let us say if the message have a retry count of INT_MAX, this is not going to work. I think we can be a little bit more conservative that if the message have a retry count larger than a threshold, we just retry the threshold value.

@kaisun2000 , if the user explicitly sets the retry to be infinite, we shall not block it, right? Note that the retry won't cause any Helix controller issue, since it is only retried on the participant side. For ZK servers, yeah, it will lead to many ZK writes potentially. But I think that is a problem at a different level, ZK writes throttling for example.
In addition, compared with the current behavior, even a retry count == INT_MAX works better. Since every retry in the new logic only generates one write. The older logic will remove the message and creating a new one. So at least 2 write IOs for each retry.

kaisun2000:
> > Let us say if the message have a retry count of INT_MAX, this is not going to work. I think we can be a little bit more conservative that if the message have a retry count larger than a threshold, we just retry the threshold value.
> 
> @kaisun2000 , if the user explicitly sets the retry to be infinite, we shall not block it, right? Note that the retry won't cause any Helix controller issue, since it is only retried on the participant side. For ZK servers, yeah, it will lead to many ZK writes potentially. But I think that is a problem at a different level, ZK writes throttling for example.
> In addition, compared with the current behavior, even a retry count == INT_MAX works better. Since every retry in the new logic only generates one write. The older logic will remove the message and creating a new one. So at least 2 write IOs for each retry.

#1487 only makes it be default not logging to ZK. It can still log to ZK with some configuration. I think if you cap the retry count to some fixed value say 100 would be safer. Just my 2 cents. Up to you.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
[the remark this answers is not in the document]
dasahcc: I think the basic reason is we reuse the same entry that used for transition fail timeout. We can keep the logic here but let's have a TODO for additional field, which may have the logic more clear for the new definition of the retry only before execution.

dasahcc (approved):
Overall, LGTM. Good PR for cleaning up.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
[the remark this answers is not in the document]
jiajunwang: Sure, TODO might be overkill since I don't think we have a concrete plan yet. But let me comment more on this logic.

jiajunwang:
Approved by @dasahcc , I have added the comment as suggested. I will merge the PR shortly.

Issue #1507, opened 2020-11-03
Log Flooding for Missing Target Resource Jobs


NealSun96:
### Describe the bug
After IdealState removal for Task Framework, we observe log flooding regarding jobs that are missing target resources. Since certain JobConfigs exist for a long time on Zk, if their corresponding target resources get deleted, as long as the JobConfigs exist, each pipeline run will print out a log. 

### To Reproduce
Remove the target resource of certain jobs. 

### Expected behavior
Less log spamming.

PR #1508, opened 2020-11-03
Fix ondemand rebalance flooding and log flooding caused by dangling jobs

### LINKED ISSUE #1506: Ondemand Rebalance Flooding after TF IS Removal
### Describe the bug
After IdealState Removal for TaskFramework, we observe onDemand rebalance flooding. We traced back to a root cause: if a jobConfigs exists for a workflow that does not contain it in the DAG, the jobConfig will trigger runtimeJobDag refreshes every time a pipeline runs. Every time runtimeJobDag gets refreshed, it will re-process all the jobs that have been completed, and initiate a clean up process; such a process triggers an onDemand rebalance. 

### To Reproduce
Create a jobConfig for a workflow without adding it to the DAG. 

### Expected behavior
onDemand rebalances only happen when necessary.

NealSun96:
### Issues

Fixes #1506, #1507

### Description

This PR fixes 2 issues:
1. Ondemand rebalance flooding after TF IS removal. This is caused by an old problem. A jobConfig can exist for the previous iterations of a workflow, but the job doesn't exist in the job DAG. Because of runtime DAG refresh logic, such a jobConfig will cause the runtime DAG to be refreshed every time the pipeline runs. A new runtime DAG causes all the jobs to be reprocessed, and during processing, the cleanup logic is run for every job that has already completed. Before IS removal, the cleanup logic attempts to delete the IS again, which has no effect; in the new code, an onDemand rebalance is triggered instead. 
This "dangling job" will not be removed, so the job dag refresh keeps on happening, which causes the ondemand rebalances to keep on firing. 
2. Log flooding for jobs that miss target resources after TF IS removal. Similarly, "dangling jobs" can have missing target resources once their target resources are deleted. Before IS removal, this is not a problem because this log is only triggered when the job is first processed; now, the processing logic happens every pipeline (since it's config based, instead of IS based), so the log could keep on firing. 

### Tests

```
[INFO] Tests run: 1237, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,907.142 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:21 h
[INFO] Finished at: 2020-11-03T13:22:54-08:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
@@ -116,33 +116,6 @@ public class TaskDataCache extends AbstractDataCache {
       }
     }
 
-    // The following 3 blocks is for finding a list of workflows whose JobDAGs have been changed
-    // because their RuntimeJobDags would need to be re-built
-    // newly added jobs
-    for (String jobName : newJobConfigs.keySet()) {
-      if (!_jobConfigMap.containsKey(jobName) && newJobConfigs.get(jobName).getWorkflow() != null) {
-        workflowsUpdated.add(newJobConfigs.get(jobName).getWorkflow());
-      }
-
-      // Only for JobQueues when a new job is enqueued, there exists a race condition where only
-      // JobConfig is updated and the RuntimeJobDag does not get updated because when the client
-      // (TaskDriver) submits, it creates JobConfig ZNode first and modifies its parent JobDag next.
-      // To ensure that they are both properly updated, check that workflow's DAG and existing
-      // JobConfigs are consistent for JobQueues
-      JobConfig jobConfig = newJobConfigs.get(jobName);
-      if (_workflowConfigMap.containsKey(jobConfig.getWorkflow())) {
-        WorkflowConfig workflowConfig = _workflowConfigMap.get(jobConfig.getWorkflow());
-        // Check that the job's parent workflow's DAG contains this job
-        if ((workflowConfig.isJobQueue() || !workflowConfig.isTerminable()) && !_runtimeJobDagMap
-            .get(workflowConfig.getWorkflowId()).getAllNodes().contains(jobName)) {
-          // Inconsistency between JobConfigs and DAGs found. Add the workflow to workflowsUpdated
-          // to rebuild the RuntimeJobDag
-          workflowsUpdated.add(jobConfig.getWorkflow());
-        }
-      }
-    }
-
-    // Removed jobs
     // This block makes sure that the workflow config has been changed.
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
```diff
@@ -116,33 +116,6 @@ public synchronized boolean refresh(HelixDataAccessor accessor,
       }
     }
 
-    // The following 3 blocks is for finding a list of workflows whose JobDAGs have been changed
```
NealSun96, line 119: Why this section is removed: this section is refresh runtime dags based on JobConfigs. For every new job config, its workflow is added; for ever job config that doesn't exist in the jobqueue's dag (seems to be duplicate with the first part of logic), its workflow is added. 

With dangling jobs, their jobConfigs are never deleted, so this section is activated all the time and the workflow, if exists, will have its DAG refreshed all the time. The problem is detailed in PR description. 

I don't think the extra logic for new job config or for the race condition makes much sense, given that a new logic which compares WorkflowConfig version is in place - if a new job is added, eventually the DAG will be updated, which will cause a version change and a runtime dag refresh.

jiajunwang:
I think the logic change is fine. But it's better to add the test coverage.

helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
```diff
@@ -116,33 +116,6 @@ public synchronized boolean refresh(HelixDataAccessor accessor,
...
-          // Inconsistency between JobConfigs and DAGs found. Add the workflow to workflowsUpdated
-          // to rebuild the RuntimeJobDag
-          workflowsUpdated.add(jobConfig.getWorkflow());
-        }
-      }
-    }
-
-    // Removed jobs
     // This block makes sure that the workflow config has been changed.
     // This avoid the race condition where job config has been purged but job has not been deleted
```
alirezazamani, line 120: Maybe this comment needs to be changed. It was added for the purge race condition. Now it is serving as general-purpose workflow config modification.

alirezazamani (approved):
Overall LGTM. Please address Jiajun's comments. Please also run the test multiple times to make sure everything is working as expected. Thanks.

jiajunwang (approved):

NealSun96:
Final commit message:
## Fix ondemand rebalance flooding and log flooding caused by dangling jobs  ##
This PR changes runtime dag refresh logic to eliminate ondemand rebalance flooding caused by dangling jobs. This PR also changes log level to get rid of log flooding caused by missing target resources.

Issue #1512, opened 2020-11-06
A possible race condition causes ERROR task and block the job.


jiajunwang:
### Describe the bug
This issue is believed to be triggered by PR https://github.com/apache/helix/commit/f11396e5feebe20552d259e553342c17a8573a8e
The theory is that the PR logic is following the design and OK. But it triggers a potential problem.

1. We agreed that if the state transition task schedule fails, we should make the partition in ERROR state/
2. When we close a participant, we shut down the executor first. But since the callback handler is still alive, there is a race condition that the handler will try to execute a state transition. Since the thread pool already shutdown, it fails. And the partition is in ERROR state.
3. In most cases it is fine since when we shut down the thread pool the participant will be shut down immediately. And no thread pool so no side effect to the real application logic
4. Unfortunately, TF may get the error state during the race condition. And I find it will stop processing the job due to this ERROR task, even though the participant has been shutdown. No live instance.

I notice this because TestTaskRebalancerFailover becomes unstable due to this race condition.
Two potential ways to fix:
1. Change TF logic so it ignores the ERROR partition in an offline node.
2. Or we fix the participant shutdown process.

### To Reproduce
Run TestTaskRebalancerFailover several times and it usually stuck on 2nd or 3rd try.

### Expected behavior
The job shall finish.

PR #1514, opened 2020-11-06
Add default message handling retry count for state transition messages.

### LINKED ISSUE #1513: Add default retry for the State Transition messages to avoid ERROR partition when there is transient errors.
After PR https://github.com/apache/helix/commit/f11396e5feebe20552d259e553342c17a8573a8e, we enforce a more strict message handling logic to avoid repeatedly retry.
One of the results is that the message might be marked as unprocessable when there is a message handler creating error. As the best-effort to avoid the unrecoverable situation, we want to add several auto-retries if the issue is in the handler factory logic process.

Adding default retry 3 times (stretch goal is to make it configurable) for the State Transition messages. So they won't immediately fail if there is a transient error in the message handler logic.

**Additional context**

jiajunwang:
### Issues

Resolves #1513

### Description

The retry count will be counted when message handler creation fails or state transition times out.

In addition, this PR also improves several ambiguous behaviors that may block a normal retry.
1. The cache of known messages list in HelixTaskExecutor may not be updated correctly if a message is going to be retried.
2. The retry count is not strictly followed in some conditions.
3. The to-be-retried message is not automatically read when there are no other new message changes. In this PR we change this behavior and ensure the retry will be done even no other message created.

Finally, improve the tests to cover the new changes.

### Tests

TestStateTransitionAppFailureHandling

[ERROR] Failures:
[ERROR]   TestTaskRebalancerFailover.test » ThreadTimeout Method org.testng.internal.Tes...
[INFO]
[ERROR] Tests run: 1247, Failures: 1, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time: 01:36 h
[INFO] Finished at: 2020-11-06T00:46:51-08:00

Rerun the test and it will pass. But it is not stable due to issue #1512

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
@@ -70,6 +70,8 @@ public abstract class MessageGenerationPhase extends AbstractBaseStage {
       .getSystemPropertyAsLong(SystemPropertyKeys.CONTROLLER_MESSAGE_PURGE_DELAY, 60 * 1000);
   private final static String PENDING_MESSAGE = "pending message";
   private final static String STALE_MESSAGE = "stale message";
+  // TODO: Make the message retry count configurable through the Cluster Config or IdealStates.
+  public final static int DEFAULT_STATE_TRANSITION_MESSAGE_RETRY_COUNT = 3;
 
   private static Logger logger = LoggerFactory.getLogger(MessageGenerationPhase.class);
 
@@ -415,6 +417,9 @@ public abstract class MessageGenerationPhase extends AbstractBaseStage {
     message.setStateModelDef(stateModelDefName);
     message.setStateModelFactoryName(resource.getStateModelFactoryname());
     message.setBucketSize(resource.getBucketSize());
+    // Set the retry count for state transition messages.
+    // TODO: make the retry count configurable in ClusterConfig or IdealState
+    message.setRetryCount(DEFAULT_STATE_TRANSITION_MESSAGE_RETRY_COUNT);
 
     if (resource.getResourceGroupName() != null) {
       message.setResourceGroupName(resource.getResourceGroupName());
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -525,47 +525,51 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
 
   private void updateMessageState(Collection<Message> msgsToBeUpdated, HelixDataAccessor accessor,
       String instanceName) {
-    if (!msgsToBeUpdated.isEmpty()) {
-      Builder keyBuilder = accessor.keyBuilder();
-      List<String> updateMsgPaths = new ArrayList<>();
-      List<DataUpdater<ZNRecord>> updaters = new ArrayList<>();
-      for (Message msg : msgsToBeUpdated) {
-        updateMsgPaths.add(msg.getKey(keyBuilder, instanceName).getPath());
-        /**
-         * We use the updater to avoid race condition between writing message to zk as READ state and removing message after ST is done
-         * If there is no message at this path, meaning the message is removed so we do not write the message
-         */
-        updaters.add(currentData -> {
-          if (currentData == null) {
-            LOG.warn(
-                "Message {} targets at {} has already been removed before it is set as READ on instance {}",
-                msg.getId(), msg.getTgtName(), instanceName);
-            return null;
-          }
-          return msg.getRecord();
-        });
-      }
-      accessor.updateChildren(updateMsgPaths, updaters, AccessOption.PERSISTENT);
+    if (msgsToBeUpdated.isEmpty()) {
+      return;
     }
 
+    Builder keyBuilder = accessor.keyBuilder();
+    List<Message> updateMsgs = new ArrayList<>();
+    List<String> updateMsgPaths = new ArrayList<>();
+    List<DataUpdater<ZNRecord>> updaters = new ArrayList<>();
+    for (Message msg : msgsToBeUpdated) {
+      updateMsgs.add(msg);
+      updateMsgPaths.add(msg.getKey(keyBuilder, instanceName).getPath());
+      /**
+       * We use the updater to avoid race condition between writing message to zk as READ state and removing message after ST is done
+       * If there is no message at this path, meaning the message is removed so we do not write the message
+       */
+      updaters.add(currentData -> {
+        if (currentData == null) {
+          LOG.warn(
+              "Message {} targets at {} has already been removed before it is set as READ on instance {}",
+              msg.getId(), msg.getTgtName(), instanceName);
+          return null;
+        }
+        return msg.getRecord();
+      });
+    }
+    boolean[] updateResults =
+        accessor.updateChildren(updateMsgPaths, updaters, AccessOption.PERSISTENT);
+
+    boolean isMessageUpdatedAsNew = false;
     // Note that only cache the known message Ids after the update to ZK is successfully done.
     // This is to avoid inconsistent cache.
-
-    // if a message in "NEW" state is updated, then we might need to process it soon.
-    boolean isNewMessageUpdated = false;
-    for (Message msg : msgsToBeUpdated) {
+    for (int i = 0; i < updateMsgs.size(); i++) {
+      Message msg = updateMsgs.get(i);
       if (msg.getMsgState().equals(MessageState.NEW)) {
-        isNewMessageUpdated = true;
-        // If a message is still "NEW", it is not a known message. The message may not be able to
-        // processed now in an expected way.
-      } else {
-        // else, cache the known messages.
+        // If a message is updated as NEW state, then we might need to process it again soon.
+        isMessageUpdatedAsNew = true;
+        // And it shall not be treated as a known messages.
+      } else if (updateResults[i]) {
         _knownMessageIds.add(msg.getId());
-      }
+      } // else, if the message update fails, the message shall not be treated as a known message.
     }
-    if (isNewMessageUpdated) {
-      // Sending a NO-OP message to trigger another message callback to re-process the New and
-      // updated messsages.
+
+    if (isMessageUpdatedAsNew) {
+      // Sending a NO-OP message to trigger another message callback to re-process the messages
+      // that are updated as NEW state.
       sendNopMessage(accessor, instanceName);
     }
   }
@@ -853,7 +857,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
         if (message.getRetryCount() <= 0) {
           // If no more retry count remains, then mark the message to be UNPROCESSABLE.
           String errorMsg = String.format("No available message Handler found!"
-                  + " Stop processing message %s since it has negative remaining retry count %d!",
+                  + " Stop processing message %s since it has zero or negative remaining retry count %d!",
               message.getMsgId(), message.getRetryCount());
           updateUnprocessableMessage(message, null, errorMsg, manager);
         }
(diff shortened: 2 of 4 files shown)


helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -1288,6 +1308,19 @@ private void removeMessageFromZK(HelixDataAccessor accessor, Message message,
     }
   }
 
+  private void sendNopMessage(HelixDataAccessor accessor, String instanceName) {
```
zhangmeng916, line 1311: This function will trigger another event and retry the message, right? Why in the PR it said the message will not be automatically retried?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -1288,6 +1308,19 @@ private void removeMessageFromZK(HelixDataAccessor accessor, Message message,
...
+    try {
+      Message nopMsg = new Message(MessageType.NO_OP, UUID.randomUUID().toString());
+      nopMsg.setSrcName(instanceName);
+      nopMsg.setTgtName(instanceName);
+      accessor
+          .setProperty(accessor.keyBuilder().message(nopMsg.getTgtName(), nopMsg.getId()), nopMsg);
+      LOG.info("Send NO_OP message to " + nopMsg.getTgtName() + ", msgId: " + nopMsg.getId());
+    } catch (Exception e) {
+      LOG.error(e.toString());
```
zhangmeng916, line 1320: Can we have a more detailed error message here?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -521,29 +523,51 @@ public void finishTask(MessageTask task) {
...
+        });
+      }
+      accessor.updateChildren(updateMsgPaths, updaters, AccessOption.PERSISTENT);
+    }
+
+    // Note that only cache the known message Ids after the update to ZK is successfully done.
+    // This is to avoid inconsistent cache.
+
+    // if a message in "NEW" state is updated, then we might need to process it soon.
+    boolean isNewMessageUpdated = false;
```
zhangmeng916, line 555: The naming here makes me a bit confusing. This message is updated as NEW because it is cannot be processed right? Maybe call it MessageUpdatedAsNew or something.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -521,29 +523,51 @@ public void finishTask(MessageTask task) {
...
+        // If a message is still "NEW", it is not a known message. The message may not be able to
+        // processed now in an expected way.
+      } else {
+        // else, cache the known messages.
+        _knownMessageIds.add(msg.getId());
+      }
+    }
+    if (isNewMessageUpdated) {
+      // Sending a NO-OP message to trigger another message callback to re-process the New and
+      // updated messsages.
```
zhangmeng916, line 568: same here: what does "new and updated" mean?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -821,32 +843,29 @@ public void onMessage(String instanceName, List<Message> messages,
...
         // us to split the configuration item so as to avoid confusion.
-        if (message.getRetryCount() < 0) {
+        if (message.getRetryCount() <= 0) {
           // If no more retry count remains, then mark the message to be UNPROCESSABLE.
-          String errorMsg = String
-              .format("No available message Handler found!"
-                      + " Stop processing message %s since it has a negative remaining retry count %d!",
-                  message.getMsgId(), message.getRetryCount());
+          String errorMsg = String.format("No available message Handler found!"
+                  + " Stop processing message %s since it has negative remaining retry count %d!",
```
zhangmeng916, line 856: "negative" doesn't match if condition.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -70,6 +70,8 @@
       .getSystemPropertyAsLong(SystemPropertyKeys.CONTROLLER_MESSAGE_PURGE_DELAY, 60 * 1000);
   private final static String PENDING_MESSAGE = "pending message";
   private final static String STALE_MESSAGE = "stale message";
+  // TODO: Make the message retry count configurable through the Cluster Config or IdealStates.
+  public final static int DEFAULT_STATE_TRANSITION_MESSAGE_RETRY_COUNT = 3;
```
dasahcc, line 74: Is this the same thing as the message attribute of retrying? Otherwise, we need an accurate name since it is the retry for state model creation.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -521,29 +523,51 @@ public void finishTask(MessageTask task) {
...
+
+    // Note that only cache the known message Ids after the update to ZK is successfully done.
+    // This is to avoid inconsistent cache.
+
+    // if a message in "NEW" state is updated, then we might need to process it soon.
+    boolean isNewMessageUpdated = false;
+    for (Message msg : msgsToBeUpdated) {
+      if (msg.getMsgState().equals(MessageState.NEW)) {
+        isNewMessageUpdated = true;
+        // If a message is still "NEW", it is not a known message. The message may not be able to
```
dasahcc, line 559: nit: an known

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -521,29 +523,51 @@ public void finishTask(MessageTask task) {
...
+      }
+      accessor.updateChildren(updateMsgPaths, updaters, AccessOption.PERSISTENT);
+    }
+
+    // Note that only cache the known message Ids after the update to ZK is successfully done.
+    // This is to avoid inconsistent cache.
+
+    // if a message in "NEW" state is updated, then we might need to process it soon.
+    boolean isNewMessageUpdated = false;
+    for (Message msg : msgsToBeUpdated) {
```
dasahcc, line 556: Why we dont move this check in the previous block? Since it already does the looping of the msgsToBeUpdated.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: Oh yeah, good catch. I changed the design since during the test I find that would be a very bad way if we leave the message there and pray for some new messages comming.
I will modify the description.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: I agree, it was copy-pasted. Let me make it complete.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: Good name, let me change it.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: Good catch, I forget to change the comment after the code change.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
jiajunwang: It is the same. So in theory it will be applied to the timeout condition too. Although ST message does not have a timeout setup for now, we may have it in the future. So make this name too explicit may be confusing as well. I propose to make it a general retry count for ST messages. But it is debatable. Let me know what's your thought.

helix-core/src/test/java/org/apache/helix/integration/paticipant/TestStateTransitionAppFailureHandling.java
```diff
@@ -0,0 +1,203 @@
...
+        MessageGenerationPhase.DEFAULT_STATE_TRANSITION_MESSAGE_RETRY_COUNT - 1;
+    Map<String, RetryStateModelFactory> retryFactoryMap = resetParticipants(retryCountUntilSucceed);
+
+    _gSetupTool.addResourceToCluster(CLUSTER_NAME, TEST_DB, _PARTITIONS, STATE_MODEL);
+    _gSetupTool.rebalanceStorageCluster(CLUSTER_NAME, TEST_DB, REPLICAS);
+
+    HelixDataAccessor accessor = _controller.getHelixDataAccessor();
+    PropertyKey.Builder keyBuilder = accessor.keyBuilder();
+
+    // Verify and wait until all messages have been retried and failed.
```
zhangmeng916, line 158: You mean retried and succeeded finally, right?

helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java
```diff
@@ -563,23 +564,42 @@ public void testCreateHandlerException() throws InterruptedException {
...
       msg.setSrcName("127.101.1.23_2234");
       msg.setCorrelationId(UUID.randomUUID().toString());
       dataAccessor.setProperty(keyBuilder.message(manager.getInstanceName(), msg.getMsgId()), msg);
+      msgList.add(msg);
+    }
+
+    changeContext.setChangeType(HelixConstants.ChangeType.MESSAGE);
+    executor.onMessage(manager.getInstanceName(), Collections.emptyList(), changeContext);
+
+    Thread.sleep(1000);
```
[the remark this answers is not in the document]
jiajunwang: I think removing it will be fine. It was from some old code. Let me have a try.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
jiajunwang: Good point, I don't remember any limitation makes me do this. Let me have a try : )

jiajunwang: It turns out that I can move the check to the loop above, but I cannot remove the loop here after the update call. For the reason, please check the comment in code. Paste it here as well.

>     // Note that only cache the known message Ids after the update to ZK is successfully done.
>     // This is to avoid inconsistent cache.

helix-core/src/test/java/org/apache/helix/integration/paticipant/TestStateTransitionAppFailureHandling.java
jiajunwang: Fixed.

jiajunwang:
Rebased to the current master and retested.

[INFO] Results:
[INFO] 
[INFO] Tests run: 1247, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 01:29 h
[INFO] Finished at: 2020-11-10T14:54:05-08:00

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -525,47 +525,51 @@ public void finishTask(MessageTask task) {
...
       if (msg.getMsgState().equals(MessageState.NEW)) {
-        isNewMessageUpdated = true;
-        // If a message is still "NEW", it is not a known message. The message may not be able to
-        // processed now in an expected way.
-      } else {
-        // else, cache the known messages.
+        // If a message is updated as NEW state, then we might need to process it again soon.
+        isMessageUpdatedAsNew = true;
+        // And it shall not be treated as a known messages.
+      } else if (updateResults[i]) {
```
jiajunwang: On my second thought, I think this is dangerous since changed the behavior unnecessarily. Let me just log in this case instead of modifying the original logic.

alirezazamani: +1  I was thinking that an additional log for this case would be helpful.

dasahcc (approved):

jiajunwang:
Thank @dasahcc for the approval. I will merge the code soon.

PR #1523, opened 2020-11-10
Fix redundant context writes to ZK

### LINKED ISSUE #1522: Context gets written back to ZK even when it has not been modified
### Describe the bug
The controller writes the context information to ZK in each pipeline without checking of there is any change to the information. This causes redundant write requests to ZK.

### To Reproduce
Have a workflow and job running and check the version number of the znodes.

### Expected behavior
The should be selective update behavior that checks for any changes on the information being written to ZK.

alirezazamani:
### Issues

Fixes #1522  

### Description

In this PR, the redundant and excessive writes to ZK with regard to workflow and job context have been eliminated.

### Tests

TestContextUpdateIfNoChange

Helix-core:
```
[INFO] Tests run: 1246, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,172.557 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1246, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:26 h
[INFO] Finished at: 2020-11-10T10:47:08-08:00
```
Helix-rest:
```
[INFO] Tests run: 171, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 241.076 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  04:08 min
[INFO] Finished at: 2020-11-10T11:31:38-08:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
@@ -180,8 +180,7 @@ public class TaskDataCache extends AbstractDataCache {
 
     for (Map.Entry<String, ZNRecord> entry : _contextMap.entrySet()) {
       if (entry.getValue() != null) {
-        _initialContextMapHashcode.put(entry.getKey(),
-            Objects.hashCode(entry.getValue().toString()));
+        _initialContextMapHashcode.put(entry.getKey(), entry.getValue().toString().hashCode());
       }
     }
 
@@ -279,8 +278,7 @@ public class TaskDataCache extends AbstractDataCache {
    */
   private void updateContext(String resourceName, ZNRecord record) {
     if (record != null && _initialContextMapHashcode.containsKey(resourceName)
-        && _initialContextMapHashcode.get(resourceName)
-            .equals(Objects.hashCode(record.toString()))) {
+        && _initialContextMapHashcode.get(resourceName).equals((record.toString().hashCode()))) {
       return;
     }
     _contextMap.put(resourceName, record);
--- a/helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
+++ b/helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
@@ -25,6 +25,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.apache.commons.codec.digest.DigestUtils;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.helix.AccessOption;
@@ -44,6 +45,7 @@ import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
 /**
  * Cache for holding all task related cluster data, such as WorkflowConfig, JobConfig and Contexts.
  */
@@ -61,7 +63,7 @@ public class TaskDataCache extends AbstractDataCache {
   private Set<String> _contextToUpdate = new HashSet<>();
   private Set<String> _contextToRemove = new HashSet<>();
   // The following map will be used to enforce selective update behavior for contexts.
-  private Map<String, Integer> _initialContextMapHashcode = new HashMap<>();
+  private Map<String, String> _initialContextMapHashcode = new HashMap<>();
   // The following fields have been added for quota-based task scheduling
   private final AssignableInstanceManager _assignableInstanceManager =
       new AssignableInstanceManager();
@@ -178,7 +180,7 @@ public class TaskDataCache extends AbstractDataCache {
 
     for (Map.Entry<String, ZNRecord> entry : _contextMap.entrySet()) {
       if (entry.getValue() != null) {
-        _initialContextMapHashcode.put(entry.getKey(), entry.getValue().toString().hashCode());
+        _initialContextMapHashcode.put(entry.getKey(), DigestUtils.md5Hex(entry.getValue().toString()));
       }
     }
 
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
```diff
@@ -174,6 +178,13 @@ private void refreshContexts(HelixDataAccessor accessor) {
       }
     }
 
+    for (Map.Entry<String, ZNRecord> entry : _contextMap.entrySet()) {
+      if (entry.getValue() != null) {
+        _initialContextMapHashcode.put(entry.getKey(),
+            Objects.hashCode(entry.getValue().toString()));
```
dasahcc, line 184: Do we need to use Objects.hashCode? Since we already converted it to String, we can call the String's function hashCode();

helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
```diff
@@ -267,6 +278,11 @@ public void updateWorkflowContext(String resourceName, WorkflowContext workflowC
    * Update context of the Workflow or Job
    */
   private void updateContext(String resourceName, ZNRecord record) {
```
dasahcc, line 280: I think this may be an overkill by checking every update. We can let it keep updating but just one check for all the context before doing writes to ZK. Otherwise, say I am trying to update one context 10 times, I need to convert to String every time. 

Also we can leverage Java8 stream to do the checks before write to ZK, it could be efficient instead of every update check.

helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
alirezazamani: I changed it to String.hashcode()

helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
alirezazamani: Thanks for the suggestion. I will see if we can do the check at the update part

helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
```diff
@@ -267,6 +277,10 @@ public void updateWorkflowContext(String resourceName, WorkflowContext workflowC
...
+    if (record != null && _initialContextMapHashcode.containsKey(resourceName)
+        && _initialContextMapHashcode.get(resourceName).equals((record.toString().hashCode()))) {
+      return;
+    }
```
kaisun2000, line 283: The problem I see here is that hash code equals does not mean the content is equal, right? 

hashcode1 ! = hashcode2 ==> content1 != content2: this is fine

hashcode1 == hashcode2  ==> content1 == content2: this is not true.  Unless you use security hash, the probability is very high that this is true.

Or is it that I did not fully understand the intention here?

alirezazamani: Yeah, your point is true since the possible input of hashcode is bigger than the possible outputs. However, since we convert the input to a string, the chance is pretty low (note that it will not be random strings and input string is derived from the ZNRecord). We had other options like deep copy and etc. but that options could cause memory and GC issues.

kaisun2000: Can we do something like if hash equals, let to do a further check of content equality before skipping it?  otherwise, this code has the potential of losing update, which would be hard to debug in the field.

alirezazamani: That is not possible or costly. We do in place change the content. If we want to check the content we need to do a deep copy of the context caches at the refresh phase which increases the memory footprint a lot.

helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
alirezazamani: Addressed.

helix-core/src/main/java/org/apache/helix/common/caches/TaskDataCache.java
kaisun2000: Then what is the solution here? 

Is it possible to calculate security hash using say md5?

alirezazamani: @dasahcc @jiajunwang  what do you think about this discussion? I think the concern is valid in theory but not in practice. http://sigpwned.com/2018/08/10/string-hashcode-is-plenty-unique/

alirezazamani: We can also use different algorithms such as md5. I feel it is kind of overkill but if there is a strong preference for that we can use different algorithms.

kaisun2000: I think MD5 is probably ok while hashcode is not. The thing is that even in you doc you quote, it has around 'An aggregate collision rate of 8 per 10,000 still isn’t bad in an absolute sense.'  0.1% is still too high for our usage. 

MD5 etc secure hash is the foundation of digital signature. The have much much lower false rate. Basically people can claim for all practical purpose, they can replace equality test. 

We can talk about it tomorrow.

alirezazamani: I added the MD5 commit now. Both of them work for me and IMO both would work considering our input is not a random string. Anyway, we can go with either of them. If we decided to go with MD5 we can keep the last commit. If not we can revert it.

alirezazamani:
Based on our discussion, we decided to close this PR and come back to it later.

PR #1526, opened 2020-11-11
Leverage zk paginated getChildren API for ZkClient to fetch a large number of children

### LINKED ISSUE #1525: Support getting a large number of children using pagination
If a node has many children and the packet length exceeds the network buffer limit, the ZkClient's `getChildren` operation will fail. There are some cases that users will expect to fetch all the children successfully.

Pagination could be used to support this feature. It requires the native zk to have the paginated getChildren API.

**Additional context**

pkuwm:
### Issues

Resolves #1525 

### Description

If a node has many children and the packet length exceeds the network buffer limit, the ZkClient's getChildren operation will fail. There are some cases that users will expect to fetch all the children successfully. Pagination could be used to support this feature. It requires the native zk to have the paginated getChildren API.

This PR uses reflection to determine whether or not getChildren should use the native zk pagination api.

Changes:
1. Add a system property config: **zk.getChildren.pagination.disabled**:
  - if **false**, and when both zk client and server supports pagination, then the **paginated** api is called.
  - If **true**, helix zkClient will force to call **non-paginated** api.
2. Use reflection to get the method for getting children.
3. Handle exceptions thrown from reflection invoking so ZkClient can handle the exception accordingly and maintains backward compatibility.

### Tests

TestZkConnection.testGetChildren
TestZkConnection.testGetChildrenException

Since Helix does not use the LinkedIn ZK by default, unit test is not added for testing the pagination in LinkedIn zk. But it's manually tested by changing the zk lib to LinkedIn zk and running all helix tests.

helix-core
```
[INFO] Tests run: 1237, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,147.408 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:25 h
[INFO] Finished at: 2020-12-17T19:13:28-08:00
```

zookeeper-api

```
[INFO] Tests run: 52, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 65.804 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 52, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:10 min
[INFO] Finished at: 2020-12-17T17:45:46-08:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/constant/ZkSystemPropertyKeys.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/constant/ZkSystemPropertyKeys.java
@@ -63,4 +63,18 @@ public class ZkSystemPropertyKeys {
 
   /** System property key for jute.maxbuffer */
   public static final String JUTE_MAXBUFFER = "jute.maxbuffer";
+
+  /**
+   * Setting this property to {@code true} in system properties will force Helix ZkClient to use
+   * the <b>non-paginated</b> {@code getChildren} API, no matter if zookeeper supports pagination
+   * or not.
+   * <p>
+   * Given both the zookeeper client and server support <b>paginated</b> {@code getChildren} API as
+   * a prerequisite, if set to {@code false}, it will enable Helix ZkClient's {@code getChildren}
+   * API to call zookeeper's <b>paginated</b> {@code getChildren} API.
+   * <p>
+   * The default value is {@code false}.
+   */
+  public static final String ZK_GETCHILDREN_PAGINATION_DISABLED =
+      "zk.getChildren.pagination.disabled";
 }
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/constant/ZkSystemPropertyKeys.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/constant/ZkSystemPropertyKeys.java
@@ -74,7 +74,10 @@ public class ZkSystemPropertyKeys {
    * API to call zookeeper's <b>paginated</b> {@code getChildren} API.
    * <p>
    * The default value is {@code false}.
+   * <p>
+   * Note: be cautious to use this config as it can be deprecated soon.
    */
+  // TODO: deprecate this config after paginated API is deployed and stable
   public static final String ZK_GETCHILDREN_PAGINATION_DISABLED =
       "zk.getChildren.pagination.disabled";
 }
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.helix.zookeeper.constant.ZkSystemPropertyKeys;
 import org.apache.helix.zookeeper.zkclient.exception.ZkException;
 import org.apache.zookeeper.CreateMode;
@@ -52,7 +53,8 @@ public class ZkConnection implements IZkConnection {
   private static final boolean GETCHILDREN_PAGINATION_DISABLED =
       Boolean.getBoolean(ZkSystemPropertyKeys.ZK_GETCHILDREN_PAGINATION_DISABLED);
 
-  private ZooKeeper _zk = null;
+  @VisibleForTesting
+  protected ZooKeeper _zk = null;
   private Lock _zookeeperLock = new ReentrantLock();
   private Method _getChildrenMethod;
 
@@ -237,31 +239,29 @@ public class ZkConnection implements IZkConnection {
   }
 
   private void lookupGetChildrenMethod() {
-    doLookUpGetChildrenMethod();
+    _getChildrenMethod = doLookUpGetChildrenMethod();
 
     LOG.info("Pagination config {}={}, method to be invoked: {}",
         ZkSystemPropertyKeys.ZK_GETCHILDREN_PAGINATION_DISABLED, GETCHILDREN_PAGINATION_DISABLED,
         _getChildrenMethod.getName());
   }
 
-  private void doLookUpGetChildrenMethod() {
+  private Method doLookUpGetChildrenMethod() {
     if (!GETCHILDREN_PAGINATION_DISABLED) {
       try {
         // Lookup the paginated getChildren API
-        _getChildrenMethod =
-            ZooKeeper.class.getMethod("getAllChildrenPaginated", String.class, boolean.class);
-        return;
+        return ZooKeeper.class.getMethod("getAllChildrenPaginated", String.class, boolean.class);
       } catch (NoSuchMethodException e) {
         LOG.info("Paginated getChildren is not supported, fall back to non-paginated getChildren");
       }
     }
 
-    lookupNonPaginatedGetChildren();
+    return lookupNonPaginatedGetChildren();
   }
 
-  private void lookupNonPaginatedGetChildren() {
+  private Method lookupNonPaginatedGetChildren() {
     try {
-      _getChildrenMethod = ZooKeeper.class.getMethod("getChildren", String.class, boolean.class);
+      return ZooKeeper.class.getMethod("getChildren", String.class, boolean.class);
     } catch (NoSuchMethodException e) {
       // We should not expect this exception here.
       throw ExceptionUtil.convertToRuntimeException(e.getCause());
@@ -273,7 +273,7 @@ public class ZkConnection implements IZkConnection {
     if (cause instanceof KeeperException.UnimplementedException) {
       LOG.warn("Paginated getChildren is unimplemented in ZK server! "
           + "Falling back to non-paginated getChildren");
-      lookupNonPaginatedGetChildren();
+      _getChildrenMethod = lookupNonPaginatedGetChildren();
       // ZK server would disconnect this connection because of UnimplementedException.
       // Throw CONNECTIONLOSS so ZkClient can retry.
       // TODO: handle it in a better way without throwing CONNECTIONLOSS
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
@@ -47,6 +47,8 @@ public class ZkConnection implements IZkConnection {
   /** It is recommended to use quite large sessions timeouts for ZooKeeper. */
   private static final int DEFAULT_SESSION_TIMEOUT = 30000;
 
+  // A config to force disabling using ZK's paginated getChildren.
+  // By default the value is false.
   private static final boolean GETCHILDREN_PAGINATION_DISABLED =
       Boolean.getBoolean(ZkSystemPropertyKeys.ZK_GETCHILDREN_PAGINATION_DISABLED);
 
@@ -92,6 +94,7 @@ public class ZkConnection implements IZkConnection {
         LOG.debug("Closing ZooKeeper connected to " + _servers);
         _zk.close();
         _zk = null;
+        _getChildrenMethod = null;
       }
     } finally {
       _zookeeperLock.unlock();
@@ -109,6 +112,7 @@ public class ZkConnection implements IZkConnection {
         LOG.debug("Creating new ZookKeeper instance to reconnect to " + _servers + ".");
         _zk = new ZooKeeper(_servers, _sessionTimeOut, watcher);
         prevZk.close();
+        _getChildrenMethod = null;
       } catch (IOException e) {
         throw new ZkException("Unable to connect to " + _servers, e);
       }
@@ -236,29 +240,35 @@ public class ZkConnection implements IZkConnection {
       // Method is already cached.
       return;
     }
-    try {
-      if (GETCHILDREN_PAGINATION_DISABLED) {
-        lookupNonPaginatedGetChildren();
-      } else {
-        // Lookup the paginated getChildren API
-        _getChildrenMethod =
-            ZooKeeper.class.getMethod("getAllChildrenPaginated", String.class, boolean.class);
-      }
-    } catch (NoSuchMethodException e1) {
-      // Pagination API is not supported, fall back to non-paginated API
-      lookupNonPaginatedGetChildren();
-    }
+
+    doLookUpGetChildrenMethod();
+
     LOG.info("Pagination config {}={}, method to be invoked: {}",
         ZkSystemPropertyKeys.ZK_GETCHILDREN_PAGINATION_DISABLED, GETCHILDREN_PAGINATION_DISABLED,
         _getChildrenMethod.getName());
   }
 
+  private void doLookUpGetChildrenMethod() {
+    if (!GETCHILDREN_PAGINATION_DISABLED) {
+      try {
+        // Lookup the paginated getChildren API
+        _getChildrenMethod =
+            ZooKeeper.class.getMethod("getAllChildrenPaginated", String.class, boolean.class);
+        return;
+      } catch (NoSuchMethodException e) {
+        LOG.info("Paginated getChildren is not supported, fall back to non-paginated getChildren");
+      }
+    }
+
+    lookupNonPaginatedGetChildren();
+  }
+
   private void lookupNonPaginatedGetChildren() {
     try {
       _getChildrenMethod = ZooKeeper.class.getMethod("getChildren", String.class, boolean.class);
-    } catch (NoSuchMethodException e2) {
+    } catch (NoSuchMethodException e) {
       // We should not expect this exception here.
-      LOG.error("getChildren is not supported in this zookeeper version!");
+      throw ExceptionUtil.convertToRuntimeException(e.getCause());
     }
   }
 
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
@@ -20,12 +20,14 @@ package org.apache.helix.zookeeper.zkclient;
  */
 
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.List;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import org.apache.helix.zookeeper.constant.ZkSystemPropertyKeys;
 import org.apache.helix.zookeeper.zkclient.exception.ZkException;
-import org.apache.log4j.Logger;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.Op;
@@ -36,16 +38,21 @@ import org.apache.zookeeper.ZooKeeper;
 import org.apache.zookeeper.ZooKeeper.States;
 import org.apache.zookeeper.data.ACL;
 import org.apache.zookeeper.data.Stat;
-
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ZkConnection implements IZkConnection {
-  private static final Logger LOG = Logger.getLogger(ZkConnection.class);
+  private static final Logger LOG = LoggerFactory.getLogger(ZkConnection.class);
 
   /** It is recommended to use quite large sessions timeouts for ZooKeeper. */
   private static final int DEFAULT_SESSION_TIMEOUT = 30000;
 
+  private static final boolean GETCHILDREN_PAGINATION_DISABLED =
+      Boolean.getBoolean(ZkSystemPropertyKeys.ZK_GETCHILDREN_PAGINATION_DISABLED);
+
   private ZooKeeper _zk = null;
   private Lock _zookeeperLock = new ReentrantLock();
+  private Method _getChildrenMethod;
 
   private final String _servers;
   private final int _sessionTimeOut;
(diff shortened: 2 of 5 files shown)


lei-xia:
The PR title is not accurate I think.  It is more like "Leverage pagination ZK API to support fetching a large number of children node in getChilren", right?  We will not support pagination in our ZKClient's getChildren(), correct?

pkuwm:
> The PR title is not accurate I think. It is more like "Leverage pagination ZK API to support fetching a large number of children node in getChilren", right? We will not support pagination in our ZKClient's getChildren(), correct?

@lei-xia Thanks for the suggestion. You are right. I've corrected the title.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -132,10 +136,59 @@ public boolean exists(String path, boolean watch) throws KeeperException, Interr
...
       throws KeeperException, InterruptedException {
-    return _zk.getChildren(path, watch);
+    checkPaginatedGetChildren();
+    try {
+      // This cast is correct because the list we're creating
+      // is of the same type as the one passed in, which is List<String>.
+      @SuppressWarnings("unchecked")
+      List<String> children = (List<String>) (_isPaginationSupported
+          ? _getChildrenMethod.invoke(_zk, path, watch, -1, -1, -1)
+          : _getChildrenMethod.invoke(_zk, path, watch));
```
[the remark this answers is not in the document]
pkuwm: The code was updated. How about the current version? @dasahcc

jiajunwang:
I see "The following tests are written for this issue:" is checked but I don't see any test. Humm : (

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -132,10 +139,41 @@ public boolean exists(String path, boolean watch) throws KeeperException, Interr
...
+   * @param watch a boolean flag to indicate whether the watch should be added to the node
+   * @return a list of children of the given path
+   * @throws KeeperException if the server signals an error with a non-zero error code
+   * @throws InterruptedException if the server transaction is interrupted
+   */
   @Override
   public List<String> getChildren(final String path, final boolean watch)
       throws KeeperException, InterruptedException {
-    return _zk.getChildren(path, watch);
+    lookupGetChildrenMethod();
```
jiajunwang, line 162: Instead of calling this every time when client getChildren, can we just initialize the method when constructing the ZkConnection?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -192,4 +230,53 @@ public String getServers() {
...
+        ZkSystemPropertyKeys.ZK_GETCHILDREN_PAGINATION_DISABLED, GETCHILDREN_PAGINATION_DISABLED,
+        _getChildrenMethod.getName());
+  }
+
+  private void lookupNonPaginatedGetChildren() {
+    try {
+      _getChildrenMethod = ZooKeeper.class.getMethod("getChildren", String.class, boolean.class);
+    } catch (NoSuchMethodException e2) {
+      // We should not expect this exception here.
+      LOG.error("getChildren is not supported in this zookeeper version!");
```
jiajunwang, line 261: Just throw an exception. Fail earlier to block the whole logic from starting up. This is much safer than allowing a partially working client.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -192,4 +230,53 @@ public String getServers() {
...
+        lookupNonPaginatedGetChildren();
+      } else {
+        // Lookup the paginated getChildren API
+        _getChildrenMethod =
+            ZooKeeper.class.getMethod("getAllChildrenPaginated", String.class, boolean.class);
+      }
+    } catch (NoSuchMethodException e1) {
+      // Pagination API is not supported, fall back to non-paginated API
+      lookupNonPaginatedGetChildren();
+    }
```
jiajunwang, line 250: nit, how about the following logic?

if (! GETCHILDREN_PAGINATION_DISABLED) {
  try {
    _getChildrenMethod =
            ZooKeeper.class.getMethod("getAllChildrenPaginated", String.class, boolean.class);
    return;
  } catch (...) {
    log...
  }
}
lookupNonPaginatedGetChildren();
return;

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -192,4 +230,53 @@ public String getServers() {
...
+      LOG.error("getChildren is not supported in this zookeeper version!");
+    }
+  }
+
+  private void handleInvokedMethodException(Throwable cause)
+      throws KeeperException, InterruptedException {
+    if (cause instanceof KeeperException.UnimplementedException) {
+      LOG.warn("Paginated getChildren is unimplemented in ZK server! "
+          + "Falling back to non-paginated getChildren");
+      lookupNonPaginatedGetChildren();
```
jiajunwang, line 270: In this case, do you want to retry the operation?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -192,4 +230,53 @@ public String getServers() {
...
+  private void lookupNonPaginatedGetChildren() {
+    try {
+      _getChildrenMethod = ZooKeeper.class.getMethod("getChildren", String.class, boolean.class);
+    } catch (NoSuchMethodException e2) {
+      // We should not expect this exception here.
+      LOG.error("getChildren is not supported in this zookeeper version!");
+    }
+  }
+
+  private void handleInvokedMethodException(Throwable cause)
```
jiajunwang, line 265: Just to confirm, as we discussed, if the ZK server is upgraded in between, the client shall be able to leverage the new API immediately, right? In this case, I assume we expect the ZkConnection object to be recreated. And I believe this is the case when we handle new session creation.

But what if the server upgrade is fast and the session is not expired? In this case, since you update the method here (and fallback to the older one), then the client will never use pagination call unless we restart the client.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/constant/ZkSystemPropertyKeys.java
```diff
@@ -63,4 +63,18 @@
...
+   * the <b>non-paginated</b> {@code getChildren} API, no matter if zookeeper supports pagination
+   * or not.
+   * <p>
+   * Given both the zookeeper client and server support <b>paginated</b> {@code getChildren} API as
+   * a prerequisite, if set to {@code false}, it will enable Helix ZkClient's {@code getChildren}
+   * API to call zookeeper's <b>paginated</b> {@code getChildren} API.
+   * <p>
+   * The default value is {@code false}.
+   */
+  public static final String ZK_GETCHILDREN_PAGINATION_DISABLED =
```
lei-xia, line 78: If our code can automatically detect the new API, why do we need this config option?

pkuwm: It's offered to completely disable pagination even though pagination is supported in ZK. It is helpful for deployment: if something is wrong in application after pagination deployed, application can quickly disable pagination by using this system config, without changing the zk lib or roll back zk server.

And if an application wants to use LinkedIn zk but doesn't want to use paginated getChildren, this is an option to disable it as well.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
pkuwm: Right. That's the purpose of the next line `throw KeeperException.create(KeeperException.Code.CONNECTIONLOSS);`

When the `KeeperException.UnimplementedException` is thrown from the server, it is a connection loss, so just throw it to ZkClient and its retry logic will retry the `getChildren` operation appropriately.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
pkuwm: > Just to confirm, as we discussed, if the ZK server is upgraded in between, the client shall be able to leverage the new API immediately, right? In this case, I assume we expect the ZkConnection object to be recreated. And I believe this is the case when we handle new session creation.

If zk server is upgraded after the client has been created, we can not guarantee the client will always know server has enabled pagination. The example is what you said "what if the server upgrade is fast and the session is not expired". If session expires during the upgrade, client can reset the cache and re-detect the pagination. It cannot solving the problem of completely auto detecting.

> But what if the server upgrade is fast and the session is not expired? In this case, since you update the method here (and fallback to the older one), then the client will never use pagination call unless we restart the client.

ZkClient cannot automatically and dynamically handle this case. The agreement is, to enable pagination, ZK server should be supposed to upgrade to have pagination as the first step. If client has pagination first but server upgrades later, client should be expected to bounce.
To detect the server has enabled the pagination without a session change, every time the client's getChildren call must try pagination first. This is inefficient and not what we want. We have to trade off. And obviously we'd expect clients to restart if this case happens.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
pkuwm: If this is not supported, later then it is called, exception will also be thrown.
We should not expect this exception as all zk versions support this method.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
pkuwm: I thought about it. The reason for this is I'd like to include the logging in the end. If we return before logging, there is no logging.

Since you also prefer this, I have wrapped the code block into a private method so logging is after the method.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
jiajunwang: One possible way is that, once the call throws an unsupported exception, we fallback to the old getChildren method.

Then on every new session establishment, we reset it based on the configuration and zk client lib method availability.

pkuwm: This can only solve the case when session is re-established, not the case when server is upgrade but no session re-establishing. Yes, we've invalidated the cache each time session is changed.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
jiajunwang: I think the retry should be done here with some fallback logic (which changes the cached method). Otherwise, the caller's retry will fail too, right? So why bother?

pkuwm: The cached method is changed here, and later when zkClient retries calling `getChildren()`, it'll used the fallback API. 

We cannot retry calling zk server right now because the connection is loss. ZkClient has to wait for re-connected. So zkclient is the right place to retry the operation.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
pkuwm: 1. Because it will change back the method if we just executed a fallback how is it changed back the method? I mean we have if (method != null) return so if the method is already cached, we don’t go further to parse the zookeeper class and change the method cache.
2. The reason we need to reset the cache in reconnect is, when server upgrades to pagination and session expires, it could have a chance to detect server pagination is available. This is the benefit. If we don’t want it, reset in reconnect is unnecessary.

Now, there are multiple places to change the method:
1. reconnect / close, invalidate the cache: method = null
2. in getChildren:
       a. if method == null: lookup. else: continue to use the cached method.
       b. if method is unimplemented in server, fallback

With this strategy, we only have one place in getChildren to lookup the paginated method. Only the case is not able to dynamically detected: server upgrades to pagination from non pagination and there is no session re-establish. For this, we expect clients to restart to be able to use pagination.

@jiajunwang I am not sure if I fully get your point and answer it. Let me know if I don't. Thanks for the point.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
jiajunwang: This is hacky but I don't have any better way. Let's add a TODO here for the record. But the logic itself should be safe.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -192,4 +234,59 @@ public String getServers() {
   public void addAuthInfo(String scheme, byte[] auth) {
     _zk.addAuthInfo(scheme, auth);
   }
+
+  private void lookupGetChildrenMethod() {
```
kaisun2000, line 238: If people are using the original patch, is this going to be problem?

kaisun2000: for this reflection way to determine the code used. there are three possibilities of server.
1/ normal server 2/ server with your patch 3/ server with original path.

The point I thinking is that if 3/ is used by some team, make sure it can work with our client same as 1/?

pkuwm: Thanks, @kaisun2000 for the clarification!

I don’t think we have to do this. Firstly, we are solving the case for internal use cases. So we have our own ZK with pagination. Helix is by default still using apache ZK.
If the external users customize their ZK by directly using the original patch, I don’t think they would expect helix to be compatible with their custom ZK. So Helix will still call the non-paginated getChildren API that is supposed to be in their custom ZK. If they really want the pagination API in helix, it's recommended to use linkedin ZK or our patch, so helix will be compatible with that.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
pkuwm: @jiajunwang Here's the log (https://github.com/apache/helix/issues/1525#issuecomment-746900960) that proves the disconnection after `UnimplementedException` is returned from server.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -102,6 +112,7 @@ protected void reconnect(Watcher watcher) throws InterruptedException {
         LOG.debug("Creating new ZookKeeper instance to reconnect to " + _servers + ".");
         _zk = new ZooKeeper(_servers, _sessionTimeOut, watcher);
         prevZk.close();
+        _getChildrenMethod = null;
```
jiajunwang, line 115: Not sure if this will cause some race condition issues?

For example, the reconnect() cleans up the method here. And in the other application thread, the operation is about to use this object (with the same reference), then it will be an NPE.

jiajunwang (commented):
Overall logic looks good to me, but I think there is a possible race condition.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -192,4 +234,60 @@ public String getServers() {
   public void addAuthInfo(String scheme, byte[] auth) {
     _zk.addAuthInfo(scheme, auth);
   }
+
+  private void lookupGetChildrenMethod() {
```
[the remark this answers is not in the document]
pkuwm: I am not sure if I fully understand your point. I guess your point is to reduce the places where `_getChildrenMethod` is referred. I just tried and realized that the code is not as clean as current way, because `lookupNonPaginatedGetChildren` is called twice. It won't achieve "the _getChildrenMethod is only referred in this method".

Since it is an internal variable/cache, I think it's fine to refer it multiple places as current way.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
pkuwm: Theoretically it could happen, though the possibility is almost 0. It’s like
```
if (a == null) { a = xxx; }
a.call();
```
Between these two lines.
NPE does not cause correctness issue, the operation just fails. NPE could also happen now: like zkclient is closed, another thread is still trying the zk operation, it could also get NPE, as zk = null in close()

There are a few options to avoid it:

1. lock the method reference. Obviously we don’t want it as it’s not nice. Would ask why only getChildren is locked but not other APIs. Kind of weird..
2. Instead of setting method = null in reconnect, we do lookup() and assign a value to it, so null is avoided. This is something I considered before when comparing it with setting to null (I felt setting to null might be risky not a good practice at least..)
3. Just don’t reset method in reconnect. The purpose of resetting is for client has pagination zk version first, and then zk server upgrades, so client can reset and detect server supports it then client can start to use pagination. But it could not cover all cases, like if servers can upgrade really fast without expiring client session, client can still not detect server’s upgrade. So I think the right way is, if we want to onboard pagination, always expect server upgrades first then client. Vice versa, pagination is not leveraged in helix client.

2 is doable, but I prefer 3, because we always follow the rule: "if we want to onboard pagination, always expect server upgrades first then client".  2 is not transparent or consistent: sometimes it can detect and sometimes not.

What do you think?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -192,4 +234,60 @@ public String getServers() {
...
+    }
+
+    doLookUpGetChildrenMethod();
+
+    LOG.info("Pagination config {}={}, method to be invoked: {}",
+        ZkSystemPropertyKeys.ZK_GETCHILDREN_PAGINATION_DISABLED, GETCHILDREN_PAGINATION_DISABLED,
+        _getChildrenMethod.getName());
+  }
+
+  private void doLookUpGetChildrenMethod() {
```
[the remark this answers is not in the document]
jiajunwang: I don't get it. The comment was about the if the condition itself. I don't see any relationship between that suggestion and making it a private method.

jiajunwang (approved):
The overall logic looks good to me.
There are only some style-related comments from my side. I prefer that we fix those points or discuss more to ensure that we are on the same page. But it is not a must.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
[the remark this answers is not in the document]
jiajunwang: Synced with Huizhi, it helps to ensure the return won't prematurely end the method so we don't see the log. I think it is fine.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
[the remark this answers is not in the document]
pkuwm: Synced with @jiajunwang offline. The style change is not a must. Also updated the style to reduce the reference places.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/constant/ZkSystemPropertyKeys.java
pkuwm: Added a note in java doc to indicate the config is temporary - can be deprecated after pagination is deployed and stable.

pkuwm:

If a node has many children and the packet length exceeds the network buffer limit, the ZkClient's getChildren operation will fail. There are some cases that users will expect to fetch all the children successfully. Pagination could be used to support this feature. It requires the native zk to have the paginated getChildren API.

This uses reflection to determine whether or not getChildren should use the native zk pagination api.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
```diff
@@ -36,16 +38,23 @@
...
 public class ZkConnection implements IZkConnection {
-  private static final Logger LOG = Logger.getLogger(ZkConnection.class);
+  private static final Logger LOG = LoggerFactory.getLogger(ZkConnection.class);
 
   /** It is recommended to use quite large sessions timeouts for ZooKeeper. */
   private static final int DEFAULT_SESSION_TIMEOUT = 30000;
 
+  // A config to force disabling using ZK's paginated getChildren.
+  // By default the value is false.
+  private static final boolean GETCHILDREN_PAGINATION_DISABLED =
```
dasahcc, line 53: Would this cause problem if default is enabled? Let's say server version is not supporting that. Then our getChildren will keep failing in looking up the method?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
dasahcc, line 241: Is this one time look up? Why we dont do it at initial stage and make it final?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
pkuwm: It won't have a problem. If server does not support the feature, it will handle the `UnimplementedException` from server and fallback to the non-paginated API.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkConnection.java
pkuwm: 1. It is a two step lookup: lookup the client lib, and call server. If we do lookup at the construction stage, we have to call server to know if server supports pagination or not. This is an extra call. 
2.  We'd like to save this extra call by using lazy init.
a. only when the first getChildren call, we get the method from client zk, then send it to server
b. If server does not support, we fallback to non-paginated. And cache it.
c. Future calls, it’ll always use the cached method to getChildren

Synced with @dasahcc offline, he agreed.

Issue #343, opened 2019-07-18
Fine-grained state transition throttling and respect MIN_ACTIVE replica 


junkaixue:
MIN_ACTIVE replica is only applied in DelayAutoRebalancer and not respected by throttling logic in Helix. Thus if preference list for partition is: Master, Slave, Slave and current state is Master, Slave, Offline. It will generate a state transition (ST) Offline -> Slave. Even if the MIN_ACTIVE replica number is 2, Helix will treat this Offline -> Slave to be Recovery rebalance. We should fix this.

Also we need the logic to have more fine-grained state transition instead of marking all state transition of a partition to same state transition type. For example, the preference list is Master, Slave, Slave and current state is Slave, Slave, Offline. There will be two STs: Slave -> Master and Offline -> Slave. In current computation logic, they are all Recovery STs because this partition is marked as Recovery partition. With new logic, Slave -> Master is Recovery ST and Offline -> Slave will be Load rebalance when MIN_ACTIVE rebalance set to be 2.

i3wangyi:
I feel like we need to put more background knowledge and context on the topic. For example, why Helix will treat the previous ongoing offline -> slave as recovery balance given it was kicked off as load balance. One naive question, I understand the current limitation is **helix controller doesn't know the historical context about any ongoing state transitions in a participant** (is it true). Is it because of the **state transition message for the participant** get consumed? What if we preserve the pending message until completion and controller would be able to know the historical context? pros and cons? 

Let's put more background and context of the problem to have a meaningful discussion.

narendly:
The root cause for this problem is because we throttle things at partition granularity. For example, if Helix determines that a partition needs recovery balance, then it will subject all of its replicas to the recovery balance throttle config (although some replicas for the same partition might be undergoing load balance - these replicas will still be subject to recovery throttle, which is incorrect!).

Enabling Helix to throttle at replica granularity will solve this problem.

Hope this helps,
Hunter 

⁣Sent from BlueMail ​

dasahcc:
@i3wangyi We need dynamically type change of ST. For example, if you have one load balance ST Offline - Slave got blocked, but now all the replica of that partition are gone. As you memorized the type of this ST, it will be fully blocked and no active replica for that partition.

dasahcc:
We split the work into two phases:
1. Respect the MIN_ACTIVE_REPLICA in per partition level throttling.
[...]

PR #1532, opened 2020-11-14
Per replica throttle

kaisun2000:
### Issues

resolve #343 

### Description

  Per replica throttling replacing intermediate stage which is partition based. The finer granularity
  would skip boosting unnecessary replica in a recovery partition.

### Tests

TestPerReplicaThrottle
TestPerReplicaThrottleStage

(Before CI test pass, please copy & paste the result of "mvn test")

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -477,12 +477,12 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
       Pipeline rebalancePipeline = new Pipeline(pipelineName);
       rebalancePipeline.addStage(new BestPossibleStateCalcStage());
       rebalancePipeline.addStage(new IntermediateStateCalcStage());
-      // Need to add MaintenanceRecoveryStage here because MAX_PARTITIONS_PER_INSTANCE check could
-      // only occur after IntermediateStateCalcStage calculation
-      rebalancePipeline.addStage(new MaintenanceRecoveryStage());
       rebalancePipeline.addStage(new ResourceMessageGenerationPhase());
       rebalancePipeline.addStage(new MessageSelectionStage());
       rebalancePipeline.addStage(new PerReplicaThrottleStage());
+      // Need to add MaintenanceRecoveryStage here because MAX_PARTITIONS_PER_INSTANCE check could
+      // only occur after PerReplicaThrottleStage() calculation
+      rebalancePipeline.addStage(new MaintenanceRecoveryStage());
       rebalancePipeline.addStage(new MessageThrottleStage());
       rebalancePipeline.addStage(new ResourceMessageDispatchStage());
       rebalancePipeline.addStage(new PersistAssignmentStage());
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -81,6 +81,7 @@ import org.apache.helix.controller.stages.IntermediateStateCalcStage;
 import org.apache.helix.controller.stages.MaintenanceRecoveryStage;
 import org.apache.helix.controller.stages.MessageSelectionStage;
 import org.apache.helix.controller.stages.MessageThrottleStage;
+import org.apache.helix.controller.stages.PerReplicaThrottleStage;
 import org.apache.helix.controller.stages.PersistAssignmentStage;
 import org.apache.helix.controller.stages.ReadClusterDataStage;
 import org.apache.helix.controller.stages.ResourceComputationStage;
@@ -481,6 +482,7 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
       rebalancePipeline.addStage(new MaintenanceRecoveryStage());
       rebalancePipeline.addStage(new ResourceMessageGenerationPhase());
       rebalancePipeline.addStage(new MessageSelectionStage());
+      rebalancePipeline.addStage(new PerReplicaThrottleStage());
       rebalancePipeline.addStage(new MessageThrottleStage());
       rebalancePipeline.addStage(new ResourceMessageDispatchStage());
       rebalancePipeline.addStage(new PersistAssignmentStage());
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
@@ -48,7 +48,14 @@ public enum AttributeName {
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
   TO_BE_PURGED_WORKFLOWS,
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
+
   JOBS_WITHOUT_CONFIG,
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
-  TO_BE_PURGED_JOBS_MAP
+  TO_BE_PURGED_JOBS_MAP,
+
+  // This attribute denotes the messages output from Per Preplica Throttle stage
+  PER_REPLICA_THROTTLED_MESSAGES,
+
+  // This attribute deontes the targeted partition state mapping from Per Preplica Throttle stage
+  PER_REPLICA_RETRACED_STATES
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
@@ -56,6 +56,12 @@ public enum AttributeName {
   // This attribute denotes the messages output from Per Preplica Throttle stage
   PER_REPLICA_THROTTLED_MESSAGES,
 
-  // This attribute deontes the targeted partition state mapping from Per Preplica Throttle stage
-  PER_REPLICA_RETRACED_STATES
+  // This attribute denotes the targeted partition state mapping from Per Preplica Throttle stage
+  PER_REPLICA_RETRACED_STATES,
+
+  // This attribute denotes the filtered out messages deemed as recovery message
+  PER_REPLICA_THROTTLED_RECOVERY_MESSAGES,
+
+  // This attribute denotes the filtered out messages deemed as load message
+  PER_REPLICA_THOTTLED_LOAD_MESSAGES
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
@@ -63,5 +63,5 @@ public enum AttributeName {
   PER_REPLICA_THROTTLED_RECOVERY_MESSAGES,
 
   // This attribute denotes the filtered out messages deemed as load message
-  PER_REPLICA_THOTTLED_LOAD_MESSAGES
+  PER_REPLICA_THROTTLED_LOAD_MESSAGES
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
@@ -54,7 +54,7 @@ public enum AttributeName {
   TO_BE_PURGED_JOBS_MAP,
 
   // This attribute denotes the messages output from Per Preplica Throttle stage
-  PER_REPLICA_THROTTLED_MESSAGES,
+  PER_REPLICA_OUTPUT_MESSAGES,
 
   // This attribute denotes the targeted partition state mapping from Per Preplica Throttle stage
   PER_REPLICA_RETRACED_STATES,
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
@@ -57,6 +57,8 @@ import org.apache.helix.task.TaskConstants;
 import org.apache.helix.util.HelixUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import sun.rmi.runtime.Log;
+
 
 /**
  * For partition compute best possible (instance,state) pair based on
@@ -81,10 +83,14 @@ public class BestPossibleStateCalcStage extends AbstractBaseStage {
       throw new StageException(
           "Missing attributes in event:" + event + ". Requires CURRENT_STATE|RESOURCES|DataCache");
     }
+    
+    // LogUtil.logInfo(logger, _eventId, String.format("CurrentState for bestpossible is: %s", currentStateOutput));
 
     final BestPossibleStateOutput bestPossibleStateOutput =
         compute(event, resourceMap, currentStateOutput);
     event.addAttribute(AttributeName.BEST_POSSIBLE_STATE.name(), bestPossibleStateOutput);
+
+   // LogUtil.logInfo(logger, _eventId, String.format("Bestpossible output is : %s", bestPossibleStateOutput));
 
     final Map<String, InstanceConfig> instanceConfigMap = cache.getInstanceConfigMap();
     final Map<String, StateModelDefinition> stateModelDefMap = cache.getStateModelDefMap();
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
@@ -58,7 +58,6 @@ import org.apache.helix.util.HelixUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
 /**
  * For partition compute best possible (instance,state) pair based on
  * IdealState,StateModel,LiveInstance
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
@@ -83,13 +83,9 @@ public class BestPossibleStateCalcStage extends AbstractBaseStage {
           "Missing attributes in event:" + event + ". Requires CURRENT_STATE|RESOURCES|DataCache");
     }
 
-    // LogUtil.logInfo(logger, _eventId, String.format("CurrentState for bestpossible is: %s", currentStateOutput));
-
     final BestPossibleStateOutput bestPossibleStateOutput =
         compute(event, resourceMap, currentStateOutput);
     event.addAttribute(AttributeName.BEST_POSSIBLE_STATE.name(), bestPossibleStateOutput);
-
-   // LogUtil.logInfo(logger, _eventId, String.format("Bestpossible output is : %s", bestPossibleStateOutput));
 
     final Map<String, InstanceConfig> instanceConfigMap = cache.getInstanceConfigMap();
     final Map<String, StateModelDefinition> stateModelDefMap = cache.getStateModelDefMap();
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
@@ -57,7 +57,6 @@ import org.apache.helix.task.TaskConstants;
 import org.apache.helix.util.HelixUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import sun.rmi.runtime.Log;
 
 
 /**
@@ -83,7 +82,7 @@ public class BestPossibleStateCalcStage extends AbstractBaseStage {
       throw new StageException(
           "Missing attributes in event:" + event + ". Requires CURRENT_STATE|RESOURCES|DataCache");
     }
-    
+
     // LogUtil.logInfo(logger, _eventId, String.format("CurrentState for bestpossible is: %s", currentStateOutput));
 
     final BestPossibleStateOutput bestPossibleStateOutput =
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateOutput.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateOutput.java
@@ -109,11 +109,4 @@ public class BestPossibleStateOutput extends ResourcesStateMap {
     return _preferenceLists != null && _preferenceLists.containsKey(resource);
   }
 
-  @Override
-  public String toString() {
-    StringBuilder sb = new StringBuilder();
-    sb.append("preference list= ").append(_preferenceLists);
-    sb.append(", resourceStateMap= ").append(_resourceStateMap);
-    return sb.toString();
-  }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateOutput.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateOutput.java
@@ -108,5 +108,4 @@ public class BestPossibleStateOutput extends ResourcesStateMap {
   protected boolean containsResource(String resource) {
     return _preferenceLists != null && _preferenceLists.containsKey(resource);
   }
-
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateOutput.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateOutput.java
@@ -108,4 +108,12 @@ public class BestPossibleStateOutput extends ResourcesStateMap {
   protected boolean containsResource(String resource) {
     return _preferenceLists != null && _preferenceLists.containsKey(resource);
   }
+
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    sb.append("preference list= ").append(_preferenceLists);
+    sb.append(", resourceStateMap= ").append(_resourceStateMap);
+    return sb.toString();
+  }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
@@ -20,7 +20,6 @@ package org.apache.helix.controller.stages;
  */
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -33,9 +32,9 @@ import org.apache.helix.HelixDefinedState;
 import org.apache.helix.HelixException;
 import org.apache.helix.HelixManager;
 import org.apache.helix.api.config.StateTransitionThrottleConfig;
-import org.apache.helix.api.config.StateTransitionThrottleConfig.RebalanceType;
 import org.apache.helix.controller.LogUtil;
 import org.apache.helix.controller.common.PartitionStateMap;
+import org.apache.helix.controller.common.ResourcesStateMap;
 import org.apache.helix.controller.dataproviders.ResourceControllerDataProvider;
 import org.apache.helix.controller.pipeline.AbstractBaseStage;
 import org.apache.helix.controller.pipeline.StageException;
@@ -43,172 +42,97 @@ import org.apache.helix.model.BuiltInStateModelDefinitions;
 import org.apache.helix.model.ClusterConfig;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.MaintenanceSignal;
+import org.apache.helix.model.Message;
 import org.apache.helix.model.Partition;
 import org.apache.helix.model.Resource;
 import org.apache.helix.model.StateModelDefinition;
-import org.apache.helix.monitoring.mbeans.ClusterStatusMonitor;
-import org.apache.helix.monitoring.mbeans.ResourceMonitor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/**
- * For partition compute the Intermediate State (instance,state) pair based on the BestPossibleState
- * and CurrentState, with all constraints applied (such as state transition throttling).
- */
-public class IntermediateStateCalcStage extends AbstractBaseStage {
+
+public class PerReplicaThrottleStage extends AbstractBaseStage {
   private static final Logger logger =
-      LoggerFactory.getLogger(IntermediateStateCalcStage.class.getName());
+      LoggerFactory.getLogger(PerReplicaThrottleStage.class.getName());
+
+  private boolean isEmitThrottledMsg = false;
+
+  public PerReplicaThrottleStage() {
+    this(false);
+  }
+
+  public PerReplicaThrottleStage(boolean enableEmitThrottledMsg) {
+    isEmitThrottledMsg = enableEmitThrottledMsg;
+  }
 
   @Override
   public void process(ClusterEvent event) throws Exception {
     _eventId = event.getEventId();
+
     CurrentStateOutput currentStateOutput = event.getAttribute(AttributeName.CURRENT_STATE.name());
 
-    BestPossibleStateOutput bestPossibleStateOutput =
-        event.getAttribute(AttributeName.BEST_POSSIBLE_STATE.name());
+    MessageOutput selectedMessages = event.getAttribute(AttributeName.MESSAGES_SELECTED.name());
+    LogUtil.logDebug(logger, _eventId, String.format("selectedMessages is: %s", selectedMessages));
+
     Map<String, Resource> resourceToRebalance =
         event.getAttribute(AttributeName.RESOURCES_TO_REBALANCE.name());
     ResourceControllerDataProvider cache =
         event.getAttribute(AttributeName.ControllerDataProvider.name());
 
-    if (currentStateOutput == null || bestPossibleStateOutput == null || resourceToRebalance == null
+    if (currentStateOutput == null || selectedMessages == null || resourceToRebalance == null
         || cache == null) {
       throw new StageException(String.format("Missing attributes in event: %s. "
-          + "Requires CURRENT_STATE (%s) |BEST_POSSIBLE_STATE (%s) |RESOURCES (%s) |DataCache (%s)",
-          event, currentStateOutput, bestPossibleStateOutput, resourceToRebalance, cache));
+              + "Requires CURRENT_STATE (%s) |BEST_POSSIBLE_STATE (%s) |RESOURCES (%s) |DataCache (%s)",
+          event, currentStateOutput, selectedMessages, resourceToRebalance, cache));
     }
 
-    IntermediateStateOutput intermediateStateOutput =
-        compute(event, resourceToRebalance, currentStateOutput, bestPossibleStateOutput);
-    event.addAttribute(AttributeName.INTERMEDIATE_STATE.name(), intermediateStateOutput);
+    ResourcesStateMap retracedResourceStateMap = new ResourcesStateMap();
+    List<Message> throttledRecoveryMsg = new ArrayList<>();
+    List<Message> throttledLoadMsg = new ArrayList<>();
+    MessageOutput output =
+        compute(event, resourceToRebalance, currentStateOutput, selectedMessages, retracedResourceStateMap, throttledRecoveryMsg, throttledLoadMsg);
+
+    if (logger.isDebugEnabled()) {
+      LogUtil.logDebug(logger, _eventId, String.format("output is"));
+      for (String resource : resourceToRebalance.keySet()) {
+        if (output.getResourceMessages(resource) != null) {
+          LogUtil.logDebug(logger, _eventId, String.format("resource: %s", resource));
+          Map<Partition, List<Message>> partitionListMap = output.getResourceMessages(resource);
+          for (Partition partition : partitionListMap.keySet()) {
+            for (Message msg : partitionListMap.get(partition)) {
+              LogUtil.logDebug(logger, _eventId, String
+                  .format("\tresource: %s, partition: %s,  msg: %s", resource, partition, msg));
+            }
+          }
+        }
+      }
+    }
+    event.addAttribute(AttributeName.PER_REPLICA_THROTTLED_MESSAGES.name(), output);
+    LogUtil.logDebug(logger,_eventId, String.format("retraceResourceStateMap is: %s", retracedResourceStateMap));
+    event.addAttribute(AttributeName.PER_REPLICA_RETRACED_STATES.name(), retracedResourceStateMap);
+
+    if (isEmitThrottledMsg) {
+      event.addAttribute(AttributeName.PER_REPLICA_THROTTLED_RECOVERY_MESSAGES.name(), throttledRecoveryMsg);
+      event.addAttribute(AttributeName.PER_REPLICA_THOTTLED_LOAD_MESSAGES.name(), throttledLoadMsg);
+    }
 
     // Make sure no instance has more replicas/partitions assigned than maxPartitionPerInstance. If
     // it does, pause the rebalance and put the cluster on maintenance mode
     int maxPartitionPerInstance = cache.getClusterConfig().getMaxPartitionsPerInstance();
     if (maxPartitionPerInstance > 0) {
-      validateMaxPartitionsPerInstance(event, cache, intermediateStateOutput,
-          maxPartitionPerInstance);
+      validateMaxPartitionsPerInstance(retracedResourceStateMap, maxPartitionPerInstance, cache, event);
     }
-  }
-
-  /**
-   * Go through each resource, and based on BestPossibleState and CurrentState, compute
-   * IntermediateState as close to BestPossibleState while maintaining throttling constraints (for
-   * example, ensure that the number of possible pending state transitions does NOT go over the set
-   * threshold).
-   * @param event
-   * @param resourceMap
-   * @param currentStateOutput
-   * @param bestPossibleStateOutput
-   * @return
-   */
-  private IntermediateStateOutput compute(ClusterEvent event, Map<String, Resource> resourceMap,
-      CurrentStateOutput currentStateOutput, BestPossibleStateOutput bestPossibleStateOutput) {
-    IntermediateStateOutput output = new IntermediateStateOutput();
-    ResourceControllerDataProvider dataCache =
-        event.getAttribute(AttributeName.ControllerDataProvider.name());
-
-    StateTransitionThrottleController throttleController = new StateTransitionThrottleController(
-        resourceMap.keySet(), dataCache.getClusterConfig(), dataCache.getLiveInstances().keySet());
-
-    // Resource level prioritization based on the numerical (sortable) priority field.
-    // If the resource priority field is null/not set, the resource will be treated as lowest
-    // priority.
-    List<ResourcePriority> prioritizedResourceList = new ArrayList<>();
-    for (String resourceName : resourceMap.keySet()) {
-      prioritizedResourceList.add(new ResourcePriority(resourceName, Integer.MIN_VALUE));
-    }
-    // If resourcePriorityField is null at the cluster level, all resources will be considered equal
-    // in priority by keeping all priorities at MIN_VALUE
-    if (dataCache.getClusterConfig().getResourcePriorityField() != null) {
-      String priorityField = dataCache.getClusterConfig().getResourcePriorityField();
-      for (ResourcePriority resourcePriority : prioritizedResourceList) {
-        String resourceName = resourcePriority.getResourceName();
-
-        // Will take the priority from ResourceConfig first
-        // If ResourceConfig does not exist or does not have this field.
-        // Try to load it from the resource's IdealState. Otherwise, keep it at the lowest priority
-        if (dataCache.getResourceConfig(resourceName) != null
-            && dataCache.getResourceConfig(resourceName).getSimpleConfig(priorityField) != null) {
-          resourcePriority.setPriority(
-              dataCache.getResourceConfig(resourceName).getSimpleConfig(priorityField));
-        } else if (dataCache.getIdealState(resourceName) != null && dataCache
-            .getIdealState(resourceName).getRecord().getSimpleField(priorityField) != null) {
-          resourcePriority.setPriority(
-              dataCache.getIdealState(resourceName).getRecord().getSimpleField(priorityField));
-        }
-      }
-      prioritizedResourceList.sort(new ResourcePriorityComparator());
-    }
-
-    ClusterStatusMonitor clusterStatusMonitor =
-        event.getAttribute(AttributeName.clusterStatusMonitor.name());
-    List<String> failedResources = new ArrayList<>();
-
-    // Priority is applied in assignment computation because higher priority by looping in order of
-    // decreasing priority
-    for (ResourcePriority resourcePriority : prioritizedResourceList) {
-      String resourceName = resourcePriority.getResourceName();
-
-      if (!bestPossibleStateOutput.containsResource(resourceName)) {
-        LogUtil.logInfo(logger, _eventId, String.format(
-            "Skip calculating intermediate state for resource %s because the best possible state is not available.",
-            resourceName));
-        continue;
-      }
-
-      Resource resource = resourceMap.get(resourceName);
-      IdealState idealState = dataCache.getIdealState(resourceName);
-      if (idealState == null) {
-        // If IdealState is null, use an empty one
-        LogUtil.logInfo(logger, _eventId,
-            String.format(
-                "IdealState for resource %s does not exist; resource may not exist anymore",
-                resourceName));
-        idealState = new IdealState(resourceName);
-        idealState.setStateModelDefRef(resource.getStateModelDefRef());
-      }
-
-      try {
-        output.setState(resourceName,
-            computeIntermediatePartitionState(dataCache, clusterStatusMonitor, idealState,
-                resourceMap.get(resourceName), currentStateOutput,
-                bestPossibleStateOutput.getPartitionStateMap(resourceName),
-                bestPossibleStateOutput.getPreferenceLists(resourceName), throttleController));
-      } catch (HelixException ex) {
-        LogUtil.logInfo(logger, _eventId,
-            "Failed to calculate intermediate partition states for resource " + resourceName, ex);
-        failedResources.add(resourceName);
-      }
-    }
-
-    if (clusterStatusMonitor != null) {
-      clusterStatusMonitor.setResourceRebalanceStates(failedResources,
-          ResourceMonitor.RebalanceStatus.INTERMEDIATE_STATE_CAL_FAILED);
-      clusterStatusMonitor.setResourceRebalanceStates(output.resourceSet(),
-          ResourceMonitor.RebalanceStatus.NORMAL);
-    }
-
-    return output;
   }
 
   /**
    * Go through every instance in the assignment and check that each instance does NOT have more
    * replicas for partitions assigned to it than maxPartitionsPerInstance. If the assignment
    * violates this, put the cluster on maintenance mode.
-   * This logic could be integrated with compute() for IntermediateState calculation but appended
-   * separately for visibility and testing. Additionally, performing validation after compute()
-   * ensures that we have a full intermediate state mapping complete prior to validation.
-   * @param event
-   * @param cache
-   * @param intermediateStateOutput
+   * @param retracedResourceStateMap
    * @param maxPartitionPerInstance
    */
-  private void validateMaxPartitionsPerInstance(ClusterEvent event,
-      ResourceControllerDataProvider cache, IntermediateStateOutput intermediateStateOutput,
-      int maxPartitionPerInstance) {
-    Map<String, PartitionStateMap> resourceStatesMap =
-        intermediateStateOutput.getResourceStatesMap();
+  private void validateMaxPartitionsPerInstance(ResourcesStateMap retracedResourceStateMap,
+      int maxPartitionPerInstance,  ResourceControllerDataProvider cache, ClusterEvent event) {
+    Map<String, PartitionStateMap> resourceStatesMap = retracedResourceStateMap.getResourceStatesMap();
     Map<String, Integer> instancePartitionCounts = new HashMap<>();
 
     for (String resource : resourceStatesMap.keySet()) {
@@ -230,13 +154,11 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
           if (state.equals(HelixDefinedState.DROPPED.name())) {
             continue;
           }
-
           if (!instancePartitionCounts.containsKey(instance)) {
             instancePartitionCounts.put(instance, 0);
           }
-          int partitionCount = instancePartitionCounts.get(instance); // Number of replicas (from
-          // different partitions) held
-          // in this instance
+          int partitionCount = instancePartitionCounts.get(instance);
+          // Number of replicas (from different partitions) held in this instance
           partitionCount++;
           if (partitionCount > maxPartitionPerInstance) {
             HelixManager manager = event.getAttribute(AttributeName.helixmanager.name());
@@ -259,14 +181,7 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
                       + " mode due to an instance being assigned more replicas/partitions than "
                       + "the limit.");
             }
-
-            ClusterStatusMonitor clusterStatusMonitor =
-                event.getAttribute(AttributeName.clusterStatusMonitor.name());
-            if (clusterStatusMonitor != null) {
-              clusterStatusMonitor.setResourceRebalanceStates(Collections.singletonList(resource),
-                  ResourceMonitor.RebalanceStatus.INTERMEDIATE_STATE_CAL_FAILED);
-            }
-            // Throw an exception here so that messages won't be sent out based on this mapping
+            //TODO: add metrics
             throw new HelixException(errMsg);
           }
           instancePartitionCounts.put(instance, partitionCount);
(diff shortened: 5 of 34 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateOutput.java
```diff
@@ -108,4 +108,12 @@ public void setPreferenceLists(String resource,
...
     return _preferenceLists != null && _preferenceLists.containsKey(resource);
   }
+
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    sb.append("preference list= ").append(_preferenceLists);
+    sb.append(", resourceStateMap= ").append(_resourceStateMap);
+    return sb.toString();
+  }
```
narendly, line 118: Are we printing/logging this? If so, I don't think that's wise since that would be a lot of text being logged.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -81,11 +83,15 @@ public void process(ClusterEvent event) throws Exception {
       throw new StageException(
           "Missing attributes in event:" + event + ". Requires CURRENT_STATE|RESOURCES|DataCache");
     }
+    
+    // LogUtil.logInfo(logger, _eventId, String.format("CurrentState for bestpossible is: %s", currentStateOutput));
```
narendly, line 87: Please clean up these comments.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -57,6 +57,8 @@
 import org.apache.helix.util.HelixUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import sun.rmi.runtime.Log;
```
narendly, line 60: Please remove this.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
```diff
@@ -48,5 +48,8 @@
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
   TO_BE_PURGED_WORKFLOWS,
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
-  TO_BE_PURGED_JOBS_MAP
+  TO_BE_PURGED_JOBS_MAP,
+
+  PER_REPLICA_THROTTLED_MESSAGES,
+  PER_REPLICA_RETRACED_STATES
```
narendly, line 54: What does this mean?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,775 @@
...
+import org.apache.helix.model.ClusterConfig;
+import org.apache.helix.model.IdealState;
+import org.apache.helix.model.Message;
+import org.apache.helix.model.Partition;
+import org.apache.helix.model.Resource;
+import org.apache.helix.model.StateModelDefinition;
+import org.apache.helix.tools.ClusterVerifiers.StrictMatchExternalViewVerifier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import sun.rmi.runtime.Log;
```
narendly, line 48: Why are we using this library?

helix-core/src/test/java/org/apache/helix/integration/TestNoThrottleDisabledPartitions.java
```diff
@@ -37,9 +38,11 @@
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.LiveInstance;
+import org.apache.helix.model.Message;
 import org.apache.helix.model.builder.FullAutoModeISBuilder;
 import org.apache.helix.tools.ClusterVerifiers.BestPossibleExternalViewVerifier;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.jboss.netty.handler.codec.http.websocketx.WebSocketServerProtocolHandshakeHandler;
```
narendly, line 45: What's this?

helix-core/src/test/java/org/apache/helix/integration/TestNoThrottleDisabledPartitions.java
```diff
@@ -81,7 +84,7 @@ public void testDisablingTopStateReplicaByDisablingInstance() throws Exception {
     ClusterControllerManager controller =
         new ClusterControllerManager(ZK_ADDR, _clusterName, "controller_0");
     controller.syncStart();
-    Thread.sleep(500L);
+    Thread.sleep(10000L);
```
narendly, line 87: We probably shouldn't do this...

kaisun2000:
@narendly, thx for the early review. Still in early stage for this feature.Ssome major refactoring and adding more tests are pending.

helix-core/src/test/java/org/apache/helix/integration/TestNoThrottleDisabledPartitions.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
kaisun2000: comment added.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateOutput.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/integration/TestNoThrottleDisabledPartitions.java
kaisun2000: restored.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,951 @@
+package org.apache.helix.controller.stages;
```
dasahcc, line 1: This new file is hard to review... Can we do this? Since most of the code is copied from IntermediateStage. Let's use IntermediateStage file and make the change for the review first as a PR? Then after reviewed, we can rename the file to PerReplicaThrottleStage.

I know split it into two PRs could break tests for first one. But I think that could be more efficient for reviewing.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -483,7 +483,7 @@ private void chargePendingTransition(Resource resource, CurrentStateOutput curre
         rebalanceType = StateTransitionThrottleConfig.RebalanceType.LOAD_BALANCE;
       }
 
-      if (pendingMap.size() > 0) {
+      if (pendingMap.size() > 0 && rebalanceType != RebalanceType.NONE) {
```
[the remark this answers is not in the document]
kaisun2000: Agree. But this is a bug previously we have. It is also in production now.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Let me copied the code to Intermedidate stage for you to have a look at the difference. This won't compile though. 

That said, the code is not really copied from IntermediateStage. There is not one to one method. Be warned, there will be probably way too many differences.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -33,182 +32,107 @@
...
-            computeIntermediatePartitionState(dataCache, clusterStatusMonitor, idealState,
-                resourceMap.get(resourceName), currentStateOutput,
-                bestPossibleStateOutput.getPartitionStateMap(resourceName),
-                bestPossibleStateOutput.getPreferenceLists(resourceName), throttleController));
-      } catch (HelixException ex) {
-        LogUtil.logInfo(logger, _eventId,
-            "Failed to calculate intermediate partition states for resource " + resourceName, ex);
-        failedResources.add(resourceName);
-      }
+    if (isEmitThrottledMsg) {
```
dasahcc, line 113: Why we need this? Is that for backtrace the intermediate state map?

dasahcc (commented):
Since the code is too long.. I will sync up with you before next round of review...

helix-core/src/test/resources/log4j.properties
```diff
@@ -37,4 +39,8 @@ log4j.logger.org.apache=ERROR
 log4j.logger.com.noelios=ERROR
 log4j.logger.org.restlet=ERROR
 
+#log4j.logger.org.apache.helix.controller.stages.PerReplicaThrottleStage=INFO
+#log4j.logger.org.apache.helix.controller.stages.BestPossibleStateCalcStage=INFO
+
+#log4j.logger.org.apache.helix.integration.TestPerReplicaThrottle=INFO
```
jiajunwang, line 45: I assume these are for debugging? Please remove them for the official PR review.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,953 @@
...
+  private static final Logger logger =
+      LoggerFactory.getLogger(PerReplicaThrottleStage.class.getName());
+
+  private boolean isEmitThrottledMsg = false;
+
+  public PerReplicaThrottleStage() {
+    this(false);
+  }
+
+  public PerReplicaThrottleStage(boolean enableEmitThrottledMsg) {
```
jiajunwang, line 63: Is this for testing? Make it protected?

jiajunwang: Or another option is to provide a protected set method so only the test can use it.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
kaisun2000: The would expose throttled messages out if configured. (If not configured, no effect). This would facilitate debugging and testing. Especially stage testing.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
```diff
@@ -64,6 +64,18 @@ public void addMessages(String resourceName, Partition partition,
     return Collections.emptyList();
   }
 
+  public Map<Partition, List<Message>> getResourceMessages(String resourceName) {
+    Map<Partition, List<Message>> map = _messagesMap.get(resourceName);
+    if (map != null) {
+      return map;
```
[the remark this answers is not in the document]
kaisun2000: good point.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,953 @@
...
+                    MaintenanceSignal.AutoTriggerReason.MAX_PARTITION_PER_INSTANCE_EXCEEDED);
+              }
+              LogUtil.logWarn(logger, _eventId, errMsg);
+            } else {
+              LogUtil.logError(logger, _eventId,
+                  "HelixManager is not set/null! Failed to pause this cluster/enable maintenance"
+                      + " mode due to an instance being assigned more replicas/partitions than "
+                      + "the limit.");
+            }
+            //TODO: add metrics
```
jiajunwang, line 184: nit, please keep the original comment, adding TODO above it?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,953 @@
...
+    // If the resource priority field is null/not set, the resource will be treated as lowest
+    // priority.
+    List<ResourcePriority> prioritizedResourceList = new ArrayList<>();
+    for (String resourceName : resourceMap.keySet()) {
+      prioritizedResourceList.add(new ResourcePriority(resourceName, Integer.MIN_VALUE));
+    }
+    // If resourcePriorityField is null at the cluster level, all resources will be considered equal
+    // in priority by keeping all priorities at MIN_VALUE
+    if (dataCache.getClusterConfig().getResourcePriorityField() != null) {
+      String priorityField = dataCache.getClusterConfig().getResourcePriorityField();
```
jiajunwang, line 227: nit, dataCache.getClusterConfig().getResourcePriorityField() is called 2 times. We can use a local var to avoid one get.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,953 @@
...
+                throttleController, retracedPartitionsState, throttledRecoveryMsg, throttledLoadMsg);
+        output.addResourceMessages(resourceName, resourceMessages);
+        retracedResourceStateMap.setState(resourceName, retracedPartitionsState);
+      } catch (HelixException ex) {
+        LogUtil.logInfo(logger, _eventId,
+            "Failed to calculate per replica partition states for resource " + resourceName, ex);
+        failedResources.add(resourceName);
+      }
+    }
+
```
jiajunwang, line 290: Humm, where is the metrics part? At least we need a TODO here?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,953 @@
...
+  }
+
+  /*
+   * Apply per-replica throttling logic and filter out excessive recovery and load messages for a
+   * given resource.
+   * Reconstruct retrace partition states for a resource based on pending and targeted messages
+   * Return messages for partitions of a resource.
+   * Out param retracedPartitionsCurrentState
+   */
+  private Map<Partition, List<Message>> computePerReplicaPartitionState(IdealState idealState,
```
jiajunwang, line 301: Please give it a reasonable name.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,953 @@
...
+      CurrentStateOutput currentStateOutput, Map<Partition, List<Message>> selectedResourceMessages,
+      Resource resource, BestPossibleStateOutput bestPossibleStateOutput,
+      ResourceControllerDataProvider cache, StateTransitionThrottleController throttleController,
+      Map<Partition, Map<String, String>> retracedPartitionsStateMap,
```
jiajunwang, line 305: This method is crazily long. It is necessary but is it possible to split "retracedPartitionsStateMap" update out of this method? Basically, I assume it is possible that we update the state map based on the output messages outside, right?
I think this will help to simplify the code and make it reviewable.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,953 @@
...
+   * Classify the messages of each partition into recovery and load messages.
+   */
+  private void classifyMessages(
+      Resource resource,
+      CurrentStateOutput currentStateOutput,
+      BestPossibleStateOutput bestPossibleStateOutput,
+      IdealState idealState,
+      ResourceControllerDataProvider cache,
+      Map<Partition, List<Message>> selectedResourceMessages,
+
```
jiajunwang, line 594: format

jiajunwang (commented):
I tried the best to review files other than the gigantic PerReplicaThrottleStage.java. May I suggest 2 ways to make reviewing this file possible?
1. Could you please comment on all the modifications that you made based on the previous class?
2. [My preference] try to create the new class as a child class of the IntermediateStateCalcStage and then only modifying the necessary method. You can add a TODO here so we can do the cleanup work in another PR. In the later PR we can say that there is no logic change. So even the PR will still be large, there is no issue in reviewing it.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -481,6 +482,7 @@ private static PipelineRegistry createDefaultRegistry(String pipelineName) {
       rebalancePipeline.addStage(new MaintenanceRecoveryStage());
```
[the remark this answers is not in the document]
kaisun2000: Good point, let me check the details.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Thought about it before. This worth another diff. This one is already very large.

kaisun2000: will add a to-do.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Any proposal? I can adopt the name you proposed.

kaisun2000:
https://github.com/apache/helix/pull/1532#discussion_r541205021

> This method seems to be added for debug log only.
For simplifying the business logic, I suggest don't add this method but relying on the existing get method to retrieve information

This is not for debug info only. Note the selectedMessage from previous stage is of this type. Before, we don't need to access it this way. Now per replica stage is using MessageOutput as input, this why we need this method.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
```diff
@@ -64,6 +64,18 @@ public void addMessages(String resourceName, Partition partition,
...
+    }
+    return Collections.emptyMap();
+  }
+
+  public void addResourceMessages(String resourceName, Map<Partition, List<Message>> resourceMessages) {
```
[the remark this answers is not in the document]
kaisun2000: The usage is like this.
```
      MessageOutput output = new MessageOutput();
...
      try {
        Map<Partition, List<Message>> resourceMessages =
            computePerReplicaPartitionState(idealState, currentStateOutput,
                selectedMessage.getResourceMessages(resourceName), resourceMap.get(resourceName),
                bestPossibleStateOutput, dataCache,
                throttleController, retracedPartitionsState, throttledRecoveryMsg, throttledLoadMsg);
        output.addResourceMessages(resourceName, resourceMessages);
        retracedResourceStateMap.setState(resourceName, retracedPartitionsState);
```

if we remove this one, we have to expand this line `output.addResourceMessages(resourceName, resourceMessages);` to a loop, which would be less efficient. 

This is pipeline core logic, saving a few cpu cycle seems to me is worthy. Let me know what is your take?

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateCalcStage.java
```diff
@@ -58,6 +58,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
```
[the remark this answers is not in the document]
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/BestPossibleStateOutput.java
```diff
@@ -108,4 +108,5 @@ public void setPreferenceLists(String resource,
   protected boolean containsResource(String resource) {
     return _preferenceLists != null && _preferenceLists.containsKey(resource);
   }
+
```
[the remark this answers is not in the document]
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/resource/ResourceMessageGenerationPhase.java
```diff
@@ -30,8 +32,10 @@
 public class ResourceMessageGenerationPhase extends MessageGenerationPhase {
   @Override
   public void process(ClusterEvent event) throws Exception {
-    IntermediateStateOutput intermediateStateOutput =
-        event.getAttribute(AttributeName.INTERMEDIATE_STATE.name());
-    processEvent(event, intermediateStateOutput);
+    //IntermediateStateOutput intermediateStateOutput =
```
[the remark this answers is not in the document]
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
```diff
@@ -48,7 +48,20 @@
...
   TO_BE_PURGED_WORKFLOWS,
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
+
   JOBS_WITHOUT_CONFIG,
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
-  TO_BE_PURGED_JOBS_MAP
+  TO_BE_PURGED_JOBS_MAP,
+
+  // This attribute denotes the messages output from Per Preplica Throttle stage
+  PER_REPLICA_THROTTLED_MESSAGES,
```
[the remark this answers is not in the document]
kaisun2000: This is valid point. Later I also regret to use this name. Let me change to PER_REPLICA_OUTPUT_MESSAGE.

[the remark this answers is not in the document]
kaisun2000: Renamed.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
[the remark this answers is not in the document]
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: addressed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: done.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Sure. I will add another private method for retraced map handling. Basically step 6 would be in a separate method.

kaisun2000: refractored.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed to "computeReourcePartitionState()"

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: done

helix-core/src/test/resources/log4j.properties
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Good point. made this one protected.

helix-core/src/main/java/org/apache/helix/controller/stages/StateTransitionThrottleController.java
```diff
@@ -46,6 +46,20 @@
 
   private boolean _throttleEnabled = false;
 
+  @Override
+  public String toString() {
```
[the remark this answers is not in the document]
kaisun2000: for example `applyThrottling()` use it here as ` logger.trace("throttleControllerstate->{} before load", throttleController);`

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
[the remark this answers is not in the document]
jiajunwang: I understand your concern. One easy way to avoid this performance concern, while avoiding breaking the clean API design, is pass the output into the computePerReplicaPartitionState() methods. And inside the method, update the messages gradually use the existing API.

helix-core/src/main/java/org/apache/helix/controller/stages/StateTransitionThrottleController.java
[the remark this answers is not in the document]
jiajunwang: Got it. I think it would be helpful.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: throttlePerReplicaMessages?

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -396,447 +368,386 @@ private PartitionStateMap computeIntermediatePartitionState(ResourceControllerDa
...
+      IdealState idealState,
+      ResourceControllerDataProvider cache,
+      boolean onlyDownwardLoadBalance,
+      List<Message> messages,
+      Map<Message, Partition> messagePartitionMap,
+      Set<Message> throttledMessages,
+      StateTransitionThrottleConfig.RebalanceType rebalanceType
+  ) {
+    boolean isRecovery = rebalanceType == StateTransitionThrottleConfig.RebalanceType.RECOVERY_BALANCE;
+    if (isRecovery && onlyDownwardLoadBalance) {
```
zhangmeng916: This check seems redundant. This function is only used internally, and when you call it, you know it won't happen.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -396,447 +368,386 @@ private PartitionStateMap computeIntermediatePartitionState(ResourceControllerDa
...
+        continue;
+      }
+      for (Message message : partitionMessages) {
+        if (!Message.MessageType.STATE_TRANSITION.name().equals(message.getMsgType())) {
+          // todo: log?
+          // ignore cancellation message etc.
+          continue;
+        }
+        String toState = message.getToState();
+        // toIntance may not be in the retracedStateMap as so far it is current state based.
```
zhangmeng916: Can you explain a bit more what's in the retraced state map.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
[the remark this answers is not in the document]
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -33,182 +32,107 @@
...
+      LoggerFactory.getLogger(PerReplicaThrottleStage.class.getName());
+
+  private boolean isEmitThrottledMsg = false;
+
+  public PerReplicaThrottleStage() {
+    this(false);
+  }
+
+  public PerReplicaThrottleStage(boolean enableEmitThrottledMsg) {
+    isEmitThrottledMsg = enableEmitThrottledMsg;
```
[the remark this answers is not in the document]
kaisun2000: Feel free to skip the IntermediateStateCalcStage. Previously JK asked to compare IntermediateStateCalcStage with PerReplicaThrottleStage as it is deemed easier to look at the difference. But the new code can be done in a way to change just a few lines from IntermediateStateCalcStage as there new code is written from scratch with different input and output even the logic is similar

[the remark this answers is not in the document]
kaisun2000: The new logic is in PerReplicaThrottleStage.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -33,182 +32,107 @@
...
     ResourceControllerDataProvider cache =
         event.getAttribute(AttributeName.ControllerDataProvider.name());
 
-    if (currentStateOutput == null || bestPossibleStateOutput == null || resourceToRebalance == null
+    if (currentStateOutput == null || selectedMessages == null || resourceToRebalance == null
         || cache == null) {
       throw new StageException(String.format("Missing attributes in event: %s. "
-          + "Requires CURRENT_STATE (%s) |BEST_POSSIBLE_STATE (%s) |RESOURCES (%s) |DataCache (%s)",
-          event, currentStateOutput, bestPossibleStateOutput, resourceToRebalance, cache));
+              + "Requires CURRENT_STATE (%s) |BEST_POSSIBLE_STATE (%s) |RESOURCES (%s) |DataCache (%s)",
```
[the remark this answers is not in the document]
kaisun2000: Feel free to skip the IntermediateStateCalcStage, the new logic is in PerReplicaThrottleStage.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -276,116 +191,173 @@ private void validateMaxPartitionsPerInstance(ClusterEvent event,
...
-    Set<Partition> partitionsNeedRecovery = new HashSet<>();
-    Set<Partition> partitionsNeedLoadBalance = new HashSet<>();
-    Set<Partition> partitionsWithErrorStateReplica = new HashSet<>();
-    for (Partition partition : resource.getPartitions()) {
-      Map<String, String> currentStateMap =
-          currentStateOutput.getCurrentStateMap(resourceName, partition);
-      Map<String, String> bestPossibleMap =
-          bestPossiblePartitionStateMap.getPartitionMap(partition);
-      List<String> preferenceList = preferenceLists.get(partition.getPartitionName());
+    // Resource level prioritization based on the numerical (sortable) priority field.
```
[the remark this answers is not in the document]
kaisun2000: Feel free to skip the IntermediateStateCalcStage, the new logic is in PerReplicaThrottleStage.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -259,14 +181,7 @@ private void validateMaxPartitionsPerInstance(ClusterEvent event,
...
             }
-
-            ClusterStatusMonitor clusterStatusMonitor =
-                event.getAttribute(AttributeName.clusterStatusMonitor.name());
-            if (clusterStatusMonitor != null) {
-              clusterStatusMonitor.setResourceRebalanceStates(Collections.singletonList(resource),
-                  ResourceMonitor.RebalanceStatus.INTERMEDIATE_STATE_CAL_FAILED);
-            }
-            // Throw an exception here so that messages won't be sent out based on this mapping
+            //TODO: add metrics
```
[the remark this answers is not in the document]
kaisun2000: Feel free to skip the IntermediateStateCalcStage, the new logic is in PerReplicaThrottleStage.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -276,116 +191,173 @@ private void validateMaxPartitionsPerInstance(ClusterEvent event,
...
-              resourceName, partitionsWithErrorStateReplica));
+    return output;
+  }
+
+  /*
+   * Apply per-replica throttling logic and filter out excessive recovery and load messages for a
+   * given resource.
+   * Reconstruct retrace partition states for a resource based on pending and targeted messages
+   * Return messages for partitions of a resource.
+   * Out param retracedPartitionsCurrentState
```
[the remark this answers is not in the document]
kaisun2000: Feel free to skip the IntermediateStateCalcStage, the new logic is in PerReplicaThrottleStage.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
kaisun2000: Feel free to skip the IntermediateStateCalcStage, the new logic is in PerReplicaThrottleStage.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
kaisun2000: Feel free to skip the IntermediateStateCalcStage, the new logic is in PerReplicaThrottleStage.

kaisun2000:
@jiajunwang , thx for the review, address-ed current main logic feedback. Will add the comments about why test changed next.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
```diff
@@ -48,7 +48,20 @@
...
+  PER_REPLICA_OUTPUT_MESSAGES,
+
+  // This attribute denotes the targeted partition state mapping from Per Preplica Throttle stage
+  PER_REPLICA_RETRACED_STATES,
+
+  // This attribute denotes the filtered out messages deemed as recovery message
+  PER_REPLICA_THROTTLED_RECOVERY_MESSAGES,
+
+  // This attribute denotes the filtered out messages deemed as load message
+  PER_REPLICA_THOTTLED_LOAD_MESSAGES
```
pkuwm, line 66: Typo: `PER_REPLICA_THOTTLED_LOAD_MESSAGES` -> `PER_REPLICA_THROTTLED_LOAD_MESSAGES`

helix-core/src/main/java/org/apache/helix/controller/stages/resource/ResourceMessageGenerationPhase.java
```diff
@@ -19,7 +19,9 @@
  * under the License.
  */
 
+import org.apache.helix.controller.common.ResourcesStateMap;
 import org.apache.helix.controller.stages.AttributeName;
+import org.apache.helix.controller.stages.BestPossibleStateOutput;
```
pkuwm, line 24: This import is unused.

helix-core/src/test/java/org/apache/helix/mock/participant/MockOFModelFactory.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public MockOFModelFactory() {
+    this(null);
+  }
+
+  public MockOFModelFactory(MockTransition transition) {
+    _transition = transition;
+  }
+
+  public void setTrasition(MockTransition transition) {
```
pkuwm, line 35: Typo. `setTrasition` -> `setTransition`

helix-core/src/test/java/org/apache/helix/mock/MockHelixAdmin.java
```diff
@@ -45,6 +47,8 @@
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import sun.applet.Main;
```
pkuwm, line 50: Remove the unused imports  in this file/PR?

jiajunwang (commented):
Let me put some of my comments first. I'm still reviewing it.

My feeling is that the IntermediateStateCalcStage code has been patched too many times. There are some confusing logic and arbitrary designs. It might be easier to re-write it, or at least majorly refactor to simplify it. Otherwise, reviewing the code would cost weeks. And if the new change introduces some undesired behavior, it would be very hard to fix.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,956 @@
...
+          cache, currentStateMap, expectedStateCountMap, currentStateCounts);
+
+      Map<String, Message> pendingMessageMap =
+          currentStateOutput.getPendingMessageMap(resourceName, partition);
+      List<Message> pendingMessages = new ArrayList<>(pendingMessageMap.values());
+      String stateModelDefName = idealState.getStateModelDefRef();
+      StateModelDefinition stateModelDef = cache.getStateModelDef(stateModelDefName);
+
+      // sort pendingMessages based on transition priority then timeStamp for state transition message
+      pendingMessages.sort(new PartitionMessageComparator(stateModelDef));
```
dasahcc, line 516: This could potentially have problem. The sorting algorithm should be same as last round.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,956 @@
...
+
+        if (isUpward && ((currentCount < expectedCount) || (currentCount == expectedCount && toState
+            .equals(secondTopState) && currentTopCount < expectedTopCount))) {
+          recoveryMessages.add(msg);
+          partitionsNeedRecovery.add(partition);
+          // update
+          currentStateCounts.put(toState, currentCount + 1);
+        } else {
+          loadMessages.add(msg);
+        }
```
dasahcc, line 555: We should think about the general case instead of only single top state and secondary state case.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,956 @@
...
+        cache, false, recoveryMessages, messagePartitionMap,
+        throttledRecoveryMessages, StateTransitionThrottleConfig.RebalanceType.RECOVERY_BALANCE);
+
+    // Step 4: sorts load message list and applies throttling
+
+    // calculate error-on-recovery downward flag
+    // If the threshold (ErrorOrRecovery) is set, then use it, if not, then check if the old
+    // threshold (Error) is set. If the old threshold is set, use it. If not, use the default value
+    // for the new one. This is for backward-compatibility
+    int threshold = 1; // Default threshold for ErrorOrRecoveryPartitionThresholdForLoadBalance
```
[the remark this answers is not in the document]
kaisun2000: This logic and even comment are exactly the same as before. As you know, this is intentional to preserve them here, thus it would avoid any potential change of logic in production. 

Let me add a TODO here

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,956 @@
...
+import org.apache.helix.model.MaintenanceSignal;
+import org.apache.helix.model.Message;
+import org.apache.helix.model.Partition;
+import org.apache.helix.model.Resource;
+import org.apache.helix.model.StateModelDefinition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class PerReplicaThrottleStage extends AbstractBaseStage {
```
pkuwm, line 53: @kaisun2000 Maybe submit a diff between `PerReplicaThrottleStage` and `IntermediateStage` can help reviewers understand the logic change based on IntermediateStage.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
kaisun2000: This is for testing purpose. The constructor in the other PerReplicaStage is changed to protected as we usually do. See the comment with JJ.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Previously we have done this before, but it does not help. The main thing is that the input and output is different between these two stages. Thus the not 1:1 method correspondence.

helix-core/src/test/java/org/apache/helix/mock/MockHelixAdmin.java
kaisun2000: removed.

helix-core/src/test/java/org/apache/helix/mock/participant/MockOFModelFactory.java
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/resource/ResourceMessageGenerationPhase.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
kaisun2000: fixed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: added TODO

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Let me add in the implementation of metrics here too. There is not much change here.

kaisun2000: added.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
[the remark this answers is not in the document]
dasahcc: Then this should be reflect to the new file instead of this file. We need to only remove it from the pipeline but also the file itself.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+          Map<Partition, List<Message>> partitionListMap = output.getResourceMessages(resource);
+          for (Partition partition : partitionListMap.keySet()) {
+            for (Message msg : partitionListMap.get(partition)) {
+              LogUtil.logDebug(logger, _eventId, String
+                  .format("\tresource: %s, partition: %s,  msg: %s", resource, partition, msg));
+            }
+          }
+        }
+      }
+    }
```
dasahcc, line 111: We dont need this. Usually we print out the message has been throttled. Next stage should have the message send out. It will be duplicated logs.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+    event.addAttribute(AttributeName.PER_REPLICA_OUTPUT_MESSAGES.name(), output);
+    LogUtil.logDebug(logger,_eventId, String.format("retraceResourceStateMap is: %s", retracedResourceStateMap));
+    event.addAttribute(AttributeName.PER_REPLICA_RETRACED_STATES.name(), retracedResourceStateMap);
+
+    if (isEmitThrottledMsg) {
+      event.addAttribute(AttributeName.PER_REPLICA_THROTTLED_RECOVERY_MESSAGES.name(), throttledRecoveryMsg);
+      event.addAttribute(AttributeName.PER_REPLICA_THROTTLED_LOAD_MESSAGES.name(), throttledLoadMsg);
+    }
```
dasahcc, line 119: Let's minimize code structure change. We can let the print happening in the code instead of carrying them out at top level and print all.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+    if (currentStateOutput == null || selectedMessages == null || resourceToRebalance == null
+        || cache == null) {
+      throw new StageException(String.format("Missing attributes in event: %s. "
+              + "Requires CURRENT_STATE (%s) |BEST_POSSIBLE_STATE (%s) |RESOURCES (%s) |DataCache (%s)",
+          event, currentStateOutput, selectedMessages, resourceToRebalance, cache));
+    }
+
+    ResourcesStateMap retracedResourceStateMap = new ResourcesStateMap();
+    List<Message> throttledRecoveryMsg = new ArrayList<>();
+    List<Message> throttledLoadMsg = new ArrayList<>();
```
dasahcc, line 93: No need to have these inputs and carrying them out.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+  }
+
+  /**
+   * Go through every instance in the assignment and check that each instance does NOT have more
+   * replicas for partitions assigned to it than maxPartitionsPerInstance. If the assignment
+   * violates this, put the cluster on maintenance mode.
+   * @param retracedResourceStateMap
+   * @param maxPartitionPerInstance
+   */
+  private void validateMaxPartitionsPerInstance(ResourcesStateMap retracedResourceStateMap,
```
dasahcc, line 136: Let's not change the order of the function. That's the reason diff can not be generated successfully.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+          continue;
+        }
+        if (throttledLoadMessages.contains(message)) {
+          continue;
+        }
+        finalPartitionMessages.add(message);
+      }
+      out.put(partition, finalPartitionMessages);
+      output.addMessages(resourceName, partition, finalPartitionMessages);
+    }
```
dasahcc, line 429: We dont need to do some loops to construct. Why not we add all the message in new output. Then:
1. If it is not FULL_AUTO, we directly return.
2. If message has been throttled, we remove the message from output where it throttled.

So we will not see such a block to construct result.

dasahcc (commented):
Reviewed to getExpected Replica numbers. Still in reviewing for other parts.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Checked. This is not a concern. Indeed both chargingPendingMessage and later classifyMessages all use the same sorting comparator.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
xyuanlu: I think it could be simplified into 
if....else if ....else.
And we don't need he nested if here.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+          // todo: log?
+          // ignore cancellation message etc.
+          continue;
+        }
+        String toState = message.getToState();
+        // toIntance may not be in the retracedStateMap as so far it is current state based.
+        // new instance in best possible not in currentstate would not be in retracedStateMap yet.
+        String toInstance = message.getTgtName();
+        Map<String, String> retracedStateMap = retracedPartitionsStateMap.get(partition);
+        retracedStateMap.put(toInstance, toState);
```
xyuanlu, line 461: nit:
retracedStateMap.put(message.getTgtName(), message.getToState());
might be cleaner.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+    LogUtil.logInfo(logger, _eventId, String.format("Classify message for resource: %s", resourceName));
+
+    for (Partition partition : resource.getPartitions()) {
+      Map<String, String> currentStateMap =
+          currentStateOutput.getCurrentStateMap(resourceName, partition);
+      Map<String, Integer> expectedStateCountMap = new HashMap<>();
+      Map<String, Integer> currentStateCounts = new HashMap<>();
+
+      getPartitionExpectedAndCurrentStateCountMap(partition, preferenceLists, idealState,
+          cache, currentStateMap, expectedStateCountMap, currentStateCounts);
```
dasahcc, line 641: We should not count this again. Since when we charged the pending message, we need to use the retrace map. 

Also, expected state does not require to recompute again as I mentioned above.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+      return priority2.compareTo(priority1);
+    }
+  }
+
+  // compare message for throttling, note, all these message are of type state_transition: how about upward, downward?
+  // recovery are all upward
+  // 1) toState priority (toTop is higher than toSecond)
+  // 2) same toState, the message classification time, the less required toState meeting minActive requirement has higher priority
+  // 3) Higher priority for the partition of messages with fewer replicas with states matching with bestPossible ??? do we need this one
+  private static class MessageThrottleComparator implements Comparator<Message> {
```
dasahcc, line 822: As you said, if there is no diff from this comparator and another one. Please merge them. They are very complicated. It will be hard for reader to understand the both. I will review the logic in detail once you merged.

dasahcc (commented):
Reviewed to classifyMessage. Will continue for next round of review.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+            .equals(HelixDefinedState.ERROR.name())) {
+          continue;
+        }
+
+        Integer expectedCount = expectedStateCountMap.get(toState);
+        Integer currentCount = currentStateCounts.get(toState);
+        expectedCount = expectedCount == null ? 0 : expectedCount;
+        currentCount = currentCount == null ? 0 : currentCount;
+
+        boolean isUpward = !isDownwardTransition(idealState, cache, msg);
```
dasahcc, line 561: No need to check this. I know message non-upwards are load but anyway we need to check expected state vs current state. This check is not necessary.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
dasahcc: A general way is we have 2 things:
1. As you implemented the expected state map.
2. Totally matched replica numbers. This number is counting from top priority of the state and accumulatively down to the number of replicas. But there is another rule of "R".

I would suggest you first try to understand the state model rule of how we counting different states of different replicas first.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+
+        throttleController
+            .chargeInstance(StateTransitionThrottleConfig.RebalanceType.LOAD_BALANCE, toInstance);
+        throttleController.chargeCluster(StateTransitionThrottleConfig.RebalanceType.LOAD_BALANCE);
+        throttleController
+            .chargeResource(StateTransitionThrottleConfig.RebalanceType.LOAD_BALANCE, resourceName);
+        logger.trace("throttleControllerstate->{} after pending load charge msg:{}", throttleController, loadMsg);
+      }
+      retracedPartitionsStateMap.put(partition, retracedStateMap);
+    }
```
dasahcc, line 616: This code logic is very duplicated. We can refactor into previous message type decision piece. This would make the code much concise.

loop message:
   RebalanceType type = xxxx;
    throttleController
            .chargeInstance(type,
                toInstance);
        throttleController
            .chargeCluster(type);
        throttleController
            .chargeResource(type,
                resourceName);
        logger.trace("throttleControllerstate->{} after pending recovery charge msg:{}", throttleController, recoveryMsg);

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+        currentTopCount = currentTopCount == null ? 0 : currentTopCount;
+        expectedTopCount = expectedTopCount == null ? 0 : expectedTopCount;
+
+        if (isUpward && ((currentCount < expectedCount) || (currentCount == expectedCount && toState
+            .equals(secondTopState) && currentTopCount < expectedTopCount))) {
+          recoveryMessages.add(msg);
+          currentStateCounts.put(toState, currentCount + 1);
+        } else {
+          loadMessages.add(msg);
+        }
```
dasahcc, line 710: Same as I suggested above, let's loop the priority state with an accumulative number for it.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+    String resourceName = resource.getResourceName();
+
+    String stateModelDefName = idealState.getStateModelDefRef();
+    StateModelDefinition stateModelDef = cache.getStateModelDef(stateModelDefName);
+    messages.sort(new MessageThrottleComparator(bestPossibleMap, currentStateMap, messagePartitionMap, stateModelDef,isRecovery));
+    logger.trace("throttleControllerstate->{} before load", throttleController);
+    for (Message msg: messages) {
+      if (onlyDownwardLoadBalance) {
+        boolean isDownward = isDownwardTransition(idealState, cache, msg);
+        if (isDownward == false) {
```
dasahcc, line 742: NIT: no other usage of isDownward. Let's just have it as:

if (isDownwardTransition(idealState, cache, msg)) {
}

dasahcc (commented):
I went through all the logics already. Please address the comments. And let me know if you have uncertainty of the comments. The major concerns from my comments:

1. Understand how we compute the number of states and make the general counting algorithm as I suggested.
2. Merge the two comparators algorithms.

I will review it again after you addressing all the comments.  The comparator algorithm is complicated. I will review detailed logic after you finished merging.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
dasahcc: As I commented, the logic is complicated. Please merge them first. Otherwise, I cannot determine what is the difference. Will be very hard for reader understanding it.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: fixed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: @xyuanlu. This can be further simplified. Currently we keep the old code logic just to make sure there is not behavior change.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: This is for testing purpose. The test would assert the messages got throttled as load or recovery. It is not for printing purpose. See `TestPerReplicaThrottleStage`.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: This is for testing purpose. The test would assert the messages got throttled as load or recovery. It is not for printing purpose. See TestPerReplicaThrottleStage.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Maybe there is some confusion here. Let me first make sure we are on the same page about usage of two different comparator -- PartitionMessageComparator and MessageThrottleComparator.

`MessageThrottleComparator` is the counter part of `PartitionPriorityComparator` in IntermediateCalcStage. 

1/ `PartitionMessageComparator` are used in two places, namely a) pendingMessages load/recovery classification and b) input message from previous stage load/recovery classification.  -- note in the previous IntermediateCalcStage, the is not need for such a comparator as the load/recovery classification is done per partition (not per replica), thus there is no need to give a order.

2/ `MessageThrottleComparator` is basically the same as `PartitionPriorityComparator` in IntermediateCalcStage. The usage is to give an order when throttling is applied to load/recovery messages.
The logic is the following -- note this is basically the same as `PartitionPriorityComparator` in IntermediateCalcStage
- Higher priority for topState
- Higher priority for the partition with fewer active replicas
- Higher priority for the partition with fewer replicas with states matching with IdealState

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: accumulation logic is added in `getPartitionExpectedAndCurrentStateCountMap`. thus the testing logic is simplified a lot here.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: This is a good suggestion to make the code more generic. The accumulation logic is added in `getPartitionExpectedAndCurrentStateCountMap` and the testing of load/recovery logic is simplified a lot here.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Basically `PartitionMessageComparator` is only used to sort message (replica) with a partition. This logic is not need in previous IntermediateCalStage. 
`MessageThrottleComparator` is use to determine the message order across partitions and this is the same logic as PartitionPriorityComparator in IntermediateCalcStage.

What is you take?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed duplicated code.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+    if (clusterStatusMonitor != null) {
+      clusterStatusMonitor.updatePerReplicaRebalancerStats(resourceName, recoveryMessages.size(),
+          loadMessages.size(), throttledRecoveryMsgOut.size(),
+          throttledLoadMessageOut.size());
+    }
+  }
+
+  private void constructRetracedPartitionStateMap(Resource resource,
+      Map<Partition, Map<String, String>> retracedPartitionsStateMap,
+      Map<Partition, List<Message>> out) {
```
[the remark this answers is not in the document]
kaisun2000: changed to outMessagesByPartition

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+        // toIntance may not be in the retracedStateMap as so far it is current state based.
+        // new instance in best possible not in currentstate would not be in retracedStateMap yet.
+        String toInstance = message.getTgtName();
+        Map<String, String> retracedStateMap = retracedPartitionsStateMap.get(partition);
+        retracedStateMap.put(toInstance, toState);
+      }
+    }
+  }
+
+  private void getPartitionExpectedAndCurrentStateCountMap(
```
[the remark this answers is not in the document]
kaisun2000: I think the gist is that you are concerned that we will spend unnecessary CPU cycles to re-count. Instead, we can do the trick of store the data and re-use them later, right? typical space and time trade off.  

Changes done.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Changes done.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Checkout upward seems to be essential. Otherwise, for downward message if it happens `currentCount < expectedCount`, it may classify the message as recovery.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: In both places, this function are the last one in the file.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: The removing idea theoretically also works.  

`MessageOutput` is the class holding the message. However, this class does not have any removing message helper functions. It is debatable whether we want to add message removing apis to this class.

Considering the benefit of using alternative approach is not really essential, maybe let us just keep the existing approach?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: See my comment before. Let us chat about this one.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,982 @@
...
+      Partition partition,
+      Map<String, List<String>> preferenceLists,
+      IdealState idealState,
+      ResourceControllerDataProvider cache,
+      Map<String, String> currentStateMap,
+      Map<String, Integer> expectedStateCountMapOut,
+      Map<String, Integer> currentStateCountsOut
+      ) {
+    List<String> preferenceList = preferenceLists.get(partition.getPartitionName());
```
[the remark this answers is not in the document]
kaisun2000: PreferenceList is null seem to be in the case of dropping instances, let me test this a little bit more to make sure this part is addressed carefully. This can be potentially an issue. Will update a little bit later.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
dasahcc: I dont believe have real production code for testing is a good idea.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
dasahcc: Same here.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: Tested, this is a good point, we should have some check as original code.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
```diff
@@ -48,7 +48,20 @@
...
+  PER_REPLICA_OUTPUT_MESSAGES,
+
+  // This attribute denotes the targeted partition state mapping from Per Preplica Throttle stage
+  PER_REPLICA_RETRACED_STATES,
+
+  // This attribute denotes the filtered out messages deemed as recovery message
+  PER_REPLICA_THROTTLED_RECOVERY_MESSAGES,
+
+  // This attribute denotes the filtered out messages deemed as load message
+  PER_REPLICA_THROTTLED_LOAD_MESSAGES
```
narendly, line 66:
```suggestion
  TO_BE_PURGED_JOBS_MAP,

  // This attribute denotes the messages output from Per Preplica Throttle stage
  PER_REPLICA_OUTPUT_MESSAGES,

  // This attribute denotes the targeted partition state mapping from Per Preplica Throttle stage
  PER_REPLICA_RETRACED_STATES,

  // This attribute denotes the filtered out messages deemed as recovery message
  PER_REPLICA_THROTTLED_RECOVERY_MESSAGES,

  // This attribute denotes the filtered out messages deemed as load message
  PER_REPLICA_THROTTLED_LOAD_MESSAGES
```

helix-core/src/main/java/org/apache/helix/controller/stages/StateTransitionThrottleController.java
```diff
@@ -46,6 +46,20 @@
...
+    sb.append(_pendingTransitionAllowedInCluster.toString());
+    sb.append("\n");
+    sb.append("_pendingTransitionAllowedPerInstance:");
+    sb.append(_pendingTransitionAllowedPerInstance.toString());
+    sb.append("\n");
+    sb.append("_pendingTransitionAllowedPerResource:");
+    sb.append(_pendingTransitionAllowedPerResource);
+    sb.append("\n");
+    return sb.toString();
+  }
```
narendly, line 61:
```suggestion
    sb.append(_pendingTransitionAllowedInCluster.toString());
    sb.append("\n");
    sb.append("_pendingTransitionAllowedPerInstance:");
    sb.append(_pendingTransitionAllowedPerInstance.toString());
    sb.append("\n");
    sb.append("_pendingTransitionAllowedPerResource:");
    sb.append(_pendingTransitionAllowedPerResource);
    sb.append("\n");
    return sb.toString();
  }
```

asdfsdfasdfsdafasdf

jiajunwang:
Close due to inactive.

Issue #1539, opened 2020-11-17
TaskScheduling stage is slow because of lock contention


xyuanlu:
### Describe the bug
We found TaskScheduling stage is slow for task intensive workloads. From profiling result, we found the problem is:
TaskSchedulingStage.scheduleWorkflows -> getHelixPropertyStore() and CallBackHandler.invoke() all synchronized on zkHelixManager object, witch makes TaskSchedulingStage extremely slow.

PR #1540, opened 2020-11-17
Move subscribeForChange out of critical section

xyuanlu:
### Issues

#1539

### Description

We found TaskScheduling stage is slow for task intensive workloads. From profiling result, we found the problem is:
TaskSchedulingStage.scheduleWorkflows -> getHelixPropertyStore() and CallBackHandler.invoke() all synchronized on zkHelixManager object, witch makes TaskSchedulingStage extremely slow.

This PR moves CallBackHandler.subscribeForChange(), witch takes majority of the time in invoke() out of the critical section.

### Tests

NA

```
[INFO] Results:
[INFO]
[ERROR] Failures: 
[ERROR]   TestBatchMessage.testParticipantIncompatibleWithBatchMsg:355 expected:<true> but was:<false>                                                                            
[INFO]
[ERROR] Tests run: 1250, Failures: 1, Errors: 0, Skipped: 0
rerun:
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 13.508 s - in org.apache.helix.integration.messaging.TestBatchMessage
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -346,7 +346,6 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
     }
 
     synchronized (this) {
-
       if (!_expectTypes.contains(type)) {
         logger.warn("Callback handler {} received event in wrong order. Listener: {}, path: {}, "
             + "expected types: {}, but was {}", _uid, _listener, _path, _expectTypes, type);
@@ -359,9 +358,10 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
       }
     }
 
-    // This allows the listener to work with one change at a time
+    // This allows the Helix Manager to work with one change at a time
+    // TODO: Maybe we don't need to sync on _manager for all types of listener. PCould be a
+    // potential improvement candidate.
     synchronized (_manager) {
-
       if (_changeType == IDEAL_STATE) {
         IdealStateChangeListener idealStateChangeListener = (IdealStateChangeListener) _listener;
         List<IdealState> idealStates = preFetch(_propertyKey);
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -340,26 +340,28 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
   public void invoke(NotificationContext changeContext) throws Exception {
     Type type = changeContext.getType();
     long start = System.currentTimeMillis();
+    if (logger.isInfoEnabled()) {
+      logger.info("{} START: CallbackHandler {}, INVOKE {} listener: {} type: {}",
+          Thread.currentThread().getId(), _uid, _path, _listener, type);
+    }
 
-    // This allows the listener to work with one change at a time
-    synchronized (_manager) {
-      if (logger.isInfoEnabled()) {
-        logger
-            .info("{} START: CallbackHandler {}, INVOKE {} listener: {} type: {}", Thread.currentThread().getId(),
-                _uid, _path, _listener, type);
-      }
+    synchronized (this) {
 
       if (!_expectTypes.contains(type)) {
         logger.warn("Callback handler {} received event in wrong order. Listener: {}, path: {}, "
             + "expected types: {}, but was {}", _uid, _listener, _path, _expectTypes, type);
         return;
-
       }
       _expectTypes = nextNotificationType.get(type);
 
       if (type == Type.INIT || type == Type.FINALIZE || changeContext.getIsChildChange()) {
         subscribeForChanges(changeContext.getType(), _path, _watchChild);
       }
+    }
+
+    // This allows the listener to work with one change at a time
+    synchronized (_manager) {
+
       if (_changeType == IDEAL_STATE) {
         IdealStateChangeListener idealStateChangeListener = (IdealStateChangeListener) _listener;
         List<IdealState> idealStates = preFetch(_propertyKey);


helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -340,26 +340,28 @@ public void enqueueTask(NotificationContext changeContext) throws Exception {
...
-
       }
       _expectTypes = nextNotificationType.get(type);
 
       if (type == Type.INIT || type == Type.FINALIZE || changeContext.getIsChildChange()) {
         subscribeForChanges(changeContext.getType(), _path, _watchChild);
       }
+    }
+
+    // This allows the listener to work with one change at a time
```
jiajunwang, line 362: nit, "the listener" -> "the Helix Manager"?
Also, let's add a TODO here since it might be overkill as we discussed.

jiajunwang:
Should be good, but it would be great if you can rebase to the top of the master with more test fixes and see if it can pass without any error.
I can also help to trigger more github tests for you to ensure it is 100% safe.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -340,26 +340,28 @@ public void enqueueTask(NotificationContext changeContext) throws Exception {
...
 
-    // This allows the listener to work with one change at a time
-    synchronized (_manager) {
-      if (logger.isInfoEnabled()) {
-        logger
-            .info("{} START: CallbackHandler {}, INVOKE {} listener: {} type: {}", Thread.currentThread().getId(),
-                _uid, _path, _listener, type);
-      }
-
+    // TODO: Having subscribeForChanges here might be overkill. Maybe it could be moved out later.
```
kaisun2000, line 348: Why do we sync on this here?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -340,26 +340,29 @@ public void enqueueTask(NotificationContext changeContext) throws Exception {
...
-    synchronized (_manager) {
-      if (logger.isInfoEnabled()) {
-        logger
-            .info("{} START: CallbackHandler {}, INVOKE {} listener: {} type: {}", Thread.currentThread().getId(),
-                _uid, _path, _listener, type);
-      }
-
+    // TODO: Having subscribeForChanges in critical section might be overkill.
+    // Maybe it could be moved out later.
+    synchronized (this) {
```
kaisun2000, line 350: Why do we sync on this here? In fact, how do we reason it?

Assuming no batch, the calling path would be from enqueueTask(), which is called by handleXXX from ZkClient event thread. And also init() and reset()

Assuming batch, the calling path would be from _batchCallbackProcessor and also init() and reset().

init() is further called by either first time CallbackHandler construction or ZK session change.

It seems that ZK session change is the only places that there would be race condition, right?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -340,26 +340,28 @@ public void enqueueTask(NotificationContext changeContext) throws Exception {
...
 
       if (type == Type.INIT || type == Type.FINALIZE || changeContext.getIsChildChange()) {
         subscribeForChanges(changeContext.getType(), _path, _watchChild);
       }
+    }
+
+    // This allows the Helix Manager to work with one change at a time
+    // TODO: Maybe we don't need to sync on _manager for all types of listener. PCould be a
+    // potential improvement candidate.
+    synchronized (_manager) {
```
kaisun2000, line 364: why sync on _manager?

by the same argument as sync(this) above, it seems that there is no across _manager race condition, aside from ZK session change. Or am I missing something?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
xyuanlu: I guess it is possible for callBackHandler gets `Finalize` and a `callBack` at the same time. So the type checking and the subscribe should be in the same synced block..?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
xyuanlu: Yes In my understanding, I think it is possible that we don't need to make all different onXXXChange to be sequential. So this `TODO` is added..

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
xyuanlu: Please correct me if I am wrong. With batch enabled, it is possible that we get 'Finalize' and a 'callback' at the same time. If there is no lock in line 348 to 358, then it is possible that we do 
1. type check first for 'callback' and passed, 
2. type check for 'Finalize' and passed,
3. unsubscribe for 'Finalize'
4. Subscribe for 'callback' 

There won't be such issue with the lock here.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
jiajunwang: I think you already answered the question yourself.

jiajunwang (approved):

xyuanlu:
Thanks for all the reviews.
This change is ready to be merged.

Final commit message:
Move subscribeForChange out of critical section to solve lock contention.

PR #1550, opened 2020-11-24
Controller-side Task Current State Migration

### LINKED ISSUE #1549: Task Current State Migration
After adopting to CurrentState based RoutingTableProvider, customers start observing large quantities of ZooKeeper reads on routers. After investigation, it is concluded that the majority of ZooKeeper reads source from CurrentState reads. It is further concluded that Task Framework related CurrentStates contributed to the majority of CurrentState reads, since the read requests peak during workflow-execution-heavy hours. 

A solution is to move Task Framework CurrentStates out of the current path. Such migration allows routers to not listen on Task Framework CurrentStates change events, mitigating a large portion of ZooKeeper reads caused by Task Framework CurrentState changes. 

**Additional context**

NealSun96:
### Issues

Fixes #1549 (Partially)

### Description

To address excessive ZooKeeper reads caused by task framework CurrentState updates, it is proposed to move task framework CurrentStates to their separate path. 

The change is divided to multiple phases. This is the first phase, where the main focus of the changes are on the controller side. These changes should be backward compatible, and should not break anything even without further changes on participant sides

### Tests

```
[INFO] Tests run: 1250, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,108.977 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1250, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:25 h
[INFO] Finished at: 2020-12-01T12:14:17-08:00
```

### Documentation (Optional)

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -44,6 +44,7 @@
 import org.apache.helix.common.caches.CurrentStateCache;
 import org.apache.helix.common.caches.InstanceMessagesCache;
 import org.apache.helix.common.caches.PropertyCache;
+import org.apache.helix.common.caches.TaskCurrentStateCache;
```
alirezazamani, line 47: Unused import?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/WorkflowControllerDataProvider.java
```diff
@@ -78,14 +82,17 @@ private void refreshClusterStateChangeFlags(Set<HelixConstants.ChangeType> prope
...
+            .getAndSet(false) || _propertyDataChangedMap.get(HelixConstants.ChangeType.MESSAGE)
+            .getAndSet(false) || propertyRefreshed.contains(HelixConstants.ChangeType.CURRENT_STATE)
+            || propertyRefreshed.contains(HelixConstants.ChangeType.TASK_CURRENT_STATE)
             || propertyRefreshed.contains(HelixConstants.ChangeType.LIVE_INSTANCE);
   }
 
   public synchronized void refresh(HelixDataAccessor accessor) {
     long startTime = System.currentTimeMillis();
     Set<HelixConstants.ChangeType> propertyRefreshed = super.doRefresh(accessor);
+    _taskCurrentStateCache.refresh(accessor, getLiveInstanceCache().getPropertyMap());
```
alirezazamani, line 95: Why do we need to refreh here? Can TaskCurrentState get refreshed like regular current states?

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -89,6 +90,13 @@ public void process(ClusterEvent event) throws Exception {
       String instanceSessionId = instance.getEphemeralOwner();
 
       // update current states.
+      // Like ResourceComputationStage, we give priority to regular resources, so update task ones
+      // first and allow regular ones to overwrite if there's any name conflicts.
+      if (_isTaskFrameworkPipeline) {
```
alirezazamani, line 95: Assuming we only refresh the task current state for the task pipeline then I got my answer to the above question? 👍

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceComputationStage.java
```diff
@@ -79,7 +79,25 @@ public void process(ClusterEvent event) throws Exception {
...
+      String instanceName = instance.getInstanceName();
+      String clientSessionId = instance.getEphemeralOwner();
+
+      Map<String, CurrentState> currentStateMap =
+          cache.getCurrentState(instanceName, clientSessionId);
+      processCurrentStates(currentStateMap, resourceMap, resourceToRebalance, idealStates,
+          isTaskCache);
+      // Duplicate resource names between regular and task resources may happen, but most likely
+      // won't. If it does, let regular resources overwrite task resources. To avoid duplicate
+      // resource overwriting, it's better to split regular and task pipelines entirely.
```
alirezazamani, line 93: Good comment

alirezazamani (commented):
So the participant side change will come later on. Right? Then this PR will be a new controller just listening on both paths? Basically, the integration tests are passing and they are confirming the controller is able to listen on both paths. But the actual functionality of the new path will be done later when you do participant side change?

alirezazamani:
I did initial parsing for this PR. Did not see any specific issue. Will look more soon.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
NealSun96: Removed.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/WorkflowControllerDataProvider.java
NealSun96: It's to save performance on the regular pipeline. Other caches refresh in `super.doRefresh(accessor)`, and TF specific caches can refresh here.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
NealSun96: See above. I believe we're referring to the same thing.

NealSun96:
> So the participant side change will come later on. Right? Then this PR will be a new controller just listening on both paths? Basically, the integration tests are passing and they are confirming the controller is able to listen on both paths. But the actual functionality of the new path will be done later when you do participant side change?

The goal I'm trying to achieve is that this PR can be separately rolled out/back without breaking anything. Your understanding is right that there's no code that's making use of the new path, but the listener is ready to listen and respond with the new path.

helix-core/src/main/java/org/apache/helix/PropertyKey.java
```diff
@@ -485,6 +486,52 @@ public PropertyKey currentState(String instanceName, String sessionId, String re
       }
     }
 
+    /**
+     * Get a property key associated with {@link CurrentState} of an instance and session. This key
```
zhangmeng916, line 490: Both current state and task current state are using the same CurrentState. Logically it's fine, but it is really confusing. Is this the ultimate solution? Do we have plan to rename?

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -89,6 +89,13 @@ public void process(ClusterEvent event) throws Exception {
       String instanceSessionId = instance.getEphemeralOwner();
 
       // update current states.
+      // Like ResourceComputationStage, we give priority to regular resources, so update task ones
```
zhangmeng916, line 92: I have a high level question about this PR. It separates current states to different paths and update them separately. How do you ensure the backward compatibility? e.g., routing table provider listens to current state path change, but now task related current state changes are removed from that path, routing table provider does have a different behavior than before, right? This may be what you want? But still this is not backward compatible.

helix-core/src/main/java/org/apache/helix/PropertyKey.java
NealSun96: Down the line, task CurrentStates will have a different model. It's not within the scope of this change, though.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
NealSun96: That is what we want: routing table provider should not be triggered by task current states; you could see the issue description for more details.

The backward compatibility needs to be ensured on the controller side, since this change should not impact the controller.

zhangmeng916: So a related question is that when we roll out this controller version, are we sure no client will notice the difference. I mean are we sure no client relies on current state change for any task related notification?

NealSun96: The assumption I'm under is that no routing table provider should rely on task current states; preventing it from listening to task current states is the main goal of this change. This assumption is proposed before starting this project, so I'm inclined to believe that it's true. Conceptually, it also doesn't make sense for routing table provider to rely on task current states.

dasahcc (commented):
@NealSun96 I would like to give you a high level suggestion. This PR is a little large and hard to review. Let's break it into three:
1. Path building and Cache change.
2. Main controller logic change such as ComputationStageChange.
3. REST 1.0, 2.0 and tools change.

Then we can focus on the small PRs review.

NealSun96:
@dasahcc As requested, I have trimmed down this PR: I removed all the non-essential changes, including REST and tools (utils) changes. 

I kept both path building and controller logic change together. It doesn't make much sense for them to be separated. The controller change looks daunting, but it's actually just a refactor, so it's not very hard to review. Splitting it out would result too small of a PR. 

Please let me know if this looks good.

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceComputationStage.java
```diff
@@ -179,66 +197,53 @@ private void processJobConfigs(WorkflowControllerDataProvider taskDataCache, Map
   /*
    * Construct Resources based on CurrentStates and add them to resource maps
    */
-  private void processCurrentStates(BaseControllerDataProvider cache,
```
NealSun96, line 182: This change is just a refactor: same code before and after.

alirezazamani (approved):
LGTM. Since this will be a critical PR, please allow others to review this PR as well before finalizing it.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -193,6 +193,8 @@ public void addInstance(String clusterName, InstanceConfig instanceConfig) {
 
     _zkClient.createPersistent(PropertyPathBuilder.instanceMessage(clusterName, nodeId), true);
     _zkClient.createPersistent(PropertyPathBuilder.instanceCurrentState(clusterName, nodeId), true);
+    _zkClient
```
dasahcc, line 196: Please remember to handle the instances are created but not have this path.

helix-core/src/main/java/org/apache/helix/PropertyKey.java
```diff
@@ -485,6 +486,52 @@ public PropertyKey currentState(String instanceName, String sessionId, String re
...
+     * @param bucketName
+     * @return {@link PropertyKey}
+     */
+    public PropertyKey taskCurrentState(String instanceName, String sessionId, String resourceName,
+        String bucketName) {
+      if (bucketName == null) {
+        return new PropertyKey(TASKCURRENTSTATES, CurrentState.class, _clusterName, instanceName,
+            sessionId, resourceName);
+
+      } else {
```
dasahcc, line 529: This "else" keyword is not necessary. You can have it as:

if (xxx) {
return xxx
}
return xxxx;

helix-core/src/main/java/org/apache/helix/common/caches/TaskCurrentStateCache.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.helix.PropertyKey;
+import org.apache.helix.common.controllers.ControlContextProvider;
+import org.apache.helix.model.CurrentState;
+import org.apache.helix.model.LiveInstance;
+
+
+/**
+ * Cache to hold all task CurrentStates of a cluster.
+ */
+public class TaskCurrentStateCache extends ParticipantStateCache<CurrentState> {
```
dasahcc, line 37: This is same logic as regular current state, right? We can create two different object for CurrentStateCache instead of creating a new class

dasahcc (commented):
Reviewed part of the code. Will continue to the major code soon.

helix-core/src/main/java/org/apache/helix/HelixManager.java
```diff
@@ -229,6 +229,15 @@ void addCurrentStateChangeListener(CurrentStateChangeListener listener, String i
   void addCurrentStateChangeListener(org.apache.helix.CurrentStateChangeListener listener, String instanceName,
       String sessionId) throws Exception;
 
+  /**
+   * Uses CurrentStateChangeListener since TaskCurrentState shares the same CurrentState model
+   * @see CurrentStateChangeListener#onStateChange(String, List, NotificationContext)
+   * @param listener
+   * @param instanceName
+   */
+  void addTaskCurrentStateChangeListener(CurrentStateChangeListener listener, String instanceName,
```
pkuwm, line 238: I would suggest we add a default implementation for this newly added API for backwards compatibility. I've noticed that a helix user **ambry** implements this interface. So if we don't add a default implementation, it'll break ambry build if they upgrades helix.
And another benefit of default implementation is, we don't need to add an empty implementation in those classes like `MockClusterManager` in tests. Thus less changed files.

NealSun96: Thank you! Great tip!

helix-core/src/main/java/org/apache/helix/PropertyKey.java
NealSun96: Yep.

helix-core/src/main/java/org/apache/helix/common/caches/TaskCurrentStateCache.java
NealSun96: No, the `PopulateParticipantKeys` function is different: it adds `taskCurrentState` keys.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
```diff
@@ -89,6 +89,13 @@ public void process(ClusterEvent event) throws Exception {
...
+      // first and allow regular ones to overwrite if there's any name conflicts.
+      if (_isTaskFrameworkPipeline) {
+        Map<String, CurrentState> taskCurrentStateMap = ((WorkflowControllerDataProvider) cache)
```
dasahcc, line 95: So we already split them. Is this necessary to combine them into current state output? Can we just change the later logic get current state from cache?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -527,8 +540,11 @@ public void invoke(NotificationContext changeContext) throws Exception {
       _zkClient.unsubscribeChildChanges(path, this);
     }
 
-    // List of children could be empty, but won't be null.
-    return _zkClient.getChildren(path);
+    try {
+      return _zkClient.getChildren(path);
+    } catch (ZkNoNodeException e) {
+      return null;
+    }
```
dasahcc, line 547: This changes the behavior, right? Can you elaborate what kind of scenario you will see this?

NealSun96: This change was because you reminded me about the instances without the task path. :) 

For instance without TASKCURRENTSTATES path, the callback handler is still created, however, during reset(), subscribeChildChange() will be invoked and they will reach the above path. Since the path doesn't exist, this line causes a NoNodeException that's only caught on the outside by reset(). This causes some log pollution. 

I synced with @pkuwm and we determined that this is unexpected and should be fixed.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
NealSun96: It may be possible to keep them split for the 2 pipelines. Experimenting.

NealSun96: Looks like the pipelines are stuck, I'm checking again. 

I believe theoretically, it's possible to split the "CurrentStateOutput" for 2 pipelines. However, for this PR I would like to have minimal behavior changes. Combining the "CurrentStateOutput" is to maintain the task-side behavior; the task pipeline processes both regular and task framework resources, so there may be a caveat somewhere that's preventing this from working. If this isn't a big issue, I think it's wise for us to keep this behavior for now until a pipeline splitting project. What do you think? @dasahcc

helix-core/src/main/java/org/apache/helix/HelixConstants.java
```diff
@@ -34,6 +34,7 @@
     CLUSTER_CONFIG (PropertyType.CONFIGS),
     LIVE_INSTANCE (PropertyType.LIVEINSTANCES),
     CURRENT_STATE (PropertyType.CURRENTSTATES),
+    TASK_CURRENT_STATE (PropertyType.TASKCURRENTSTATES),
```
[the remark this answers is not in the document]
NealSun96: It's still CurrentState for now, so I prefer calling it TASK_CURRENT_STATE. It corresponds to the underlying model.

helix-core/src/main/java/org/apache/helix/PropertyKey.java
```diff
@@ -485,6 +486,50 @@ public PropertyKey currentState(String instanceName, String sessionId, String re
...
+
+    /**
+     * Get a property key associated with {@link CurrentState} of an instance, session, and
+     * resource. This key is for TaskCurrentState specifically.
+     * @param instanceName
+     * @param sessionId
+     * @param resourceName
+     * @return {@link PropertyKey}
+     */
+    public PropertyKey taskCurrentState(String instanceName, String sessionId, String resourceName) {
```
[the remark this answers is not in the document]
NealSun96: Sure

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
NealSun96: When the project was first started, the agreement was that we keep all current behavior, which means merging task and non-task current states; due to possible naming conflicts, we allow regular current states to overwrite task ones. In terms of safety, this is the safest option since it doesn't modify behavior. 

I think we definitely need to split it. For the future project, I am talking about code splitting: the current way is that the 2 pipelines are using one stage, which isn't truly split. I can do it here, but I have concerns over the timeline. This PR is only the first PR, and there are the "non-essential changes" and "participant-side changes" all requiring reviews. 

I'm still looking into how this can be split, but if the intention here is "it's a good to have", then I'm against doing it for this PR due to time concerns. It wasn't how we planned it either.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
dasahcc: My concern is if original behavior of other types will be changed as well. The exception we will see can be changed from NoNodeException to NullPointerException if you just return null for it.

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceComputationStage.java
```diff
@@ -79,7 +79,25 @@ public void process(ClusterEvent event) throws Exception {
...
     // idealState might be removed.
-    processCurrentStates(cache, resourceMap, resourceToRebalance, idealStates, isTaskCache);
+    Map<String, LiveInstance> availableInstances = cache.getLiveInstances();
+    for (LiveInstance instance : availableInstances.values()) {
+      String instanceName = instance.getInstanceName();
+      String clientSessionId = instance.getEphemeralOwner();
+
+      Map<String, CurrentState> currentStateMap =
+          cache.getCurrentState(instanceName, clientSessionId);
+      processCurrentStates(currentStateMap, resourceMap, resourceToRebalance, idealStates,
```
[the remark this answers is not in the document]
NealSun96: Since processing task current states and processing regular current states share the same logic, at one point along the code, the same logic will be factored out. It either happens in `process()` (processCurrentStates() is the "same logic"), or it happens in `processCurrentStates()` (requires another new function). I think it can happen in `process()`.

helix-core/src/main/java/org/apache/helix/HelixConstants.java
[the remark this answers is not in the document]
jiajunwang: I think it is just a name so it won't impact any logic.
But task current state sounds redundent. Task state is good enough.

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceComputationStage.java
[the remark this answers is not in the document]
jiajunwang: In this case, I will suggest extracting another private method to make process() code cleaner.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
NealSun96: All callers of this function (2 occasions) handle null cases. I believe this is the best solution when `path` doesn't exist, unless there's a better idea.

helix-core/src/main/java/org/apache/helix/HelixConstants.java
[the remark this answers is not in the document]
NealSun96: The name should reflect what it represents, and it represents a current state. 

Are you also suggesting that the zk path needs to be named as `TASKSTATES`? If that's the case, then `CURRENTSTATES` should be renamed to `RESOURCESTATES` to show the "counter part" relation. 

All in all, I really don't see a big problem with "task current state".

helix-core/src/main/java/org/apache/helix/controller/dataproviders/WorkflowControllerDataProvider.java
```diff
@@ -78,14 +82,17 @@ private void refreshClusterStateChangeFlags(Set<HelixConstants.ChangeType> prope
...
+            || _propertyDataChangedMap.get(HelixConstants.ChangeType.TASK_CURRENT_STATE)
+            .getAndSet(false) || _propertyDataChangedMap.get(HelixConstants.ChangeType.MESSAGE)
+            .getAndSet(false) || propertyRefreshed.contains(HelixConstants.ChangeType.CURRENT_STATE)
+            || propertyRefreshed.contains(HelixConstants.ChangeType.TASK_CURRENT_STATE)
             || propertyRefreshed.contains(HelixConstants.ChangeType.LIVE_INSTANCE);
   }
 
   public synchronized void refresh(HelixDataAccessor accessor) {
     long startTime = System.currentTimeMillis();
     Set<HelixConstants.ChangeType> propertyRefreshed = super.doRefresh(accessor);
```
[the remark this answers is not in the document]
NealSun96: Resolved offline.

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceComputationStage.java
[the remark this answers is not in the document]
NealSun96: Done.

helix-core/src/main/java/org/apache/helix/controller/stages/CurrentStateComputationStage.java
NealSun96: @dasahcc @jiajunwang I was very distracted by the argument itself that I lost sight to a very simple fact, my apology: splitting doesn't work here due to backward compatibility issues. If we completely split now, all task current states will fail to read for any participants with older helix versions. In another way, the splitting is already done in the sense that regular pipelines are not reading task current states anymore, which is an improvement done by this PR. :) 

This PR itself doesn't introduce participant-side changes, so all task current states are still in the old path. Therefore complete split causes test problems. What I observed as "pipelines stuck" are overflowing logs caused by message handling, preventing me from seeing test outputs. In short, initialization messages are continuously generated due to missing current states from CurrentStateOutput, and that causes a log flood. If you're interested to find out more details, feel free to let me know.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/WorkflowControllerDataProvider.java
```diff
@@ -78,14 +82,16 @@ private void refreshClusterStateChangeFlags(Set<HelixConstants.ChangeType> prope
         // This check (and set) is necessary for now since the current state flag in
         // _propertyDataChangedMap is not used by the BaseControllerDataProvider for now.
         _propertyDataChangedMap.get(HelixConstants.ChangeType.CURRENT_STATE).getAndSet(false)
-            || _propertyDataChangedMap.get(HelixConstants.ChangeType.MESSAGE).getAndSet(false)
-            || propertyRefreshed.contains(HelixConstants.ChangeType.CURRENT_STATE)
-            || propertyRefreshed.contains(HelixConstants.ChangeType.LIVE_INSTANCE);
+            || _propertyDataChangedMap.get(HelixConstants.ChangeType.TASK_CURRENT_STATE)
+            .getAndSet(false) || _propertyDataChangedMap.get(HelixConstants.ChangeType.MESSAGE)
+            .getAndSet(false) || propertyRefreshed
+            .contains(HelixConstants.ChangeType.LIVE_INSTANCE);
```
dasahcc, line 88: Let's make the consistent format

helix-core/src/main/java/org/apache/helix/HelixConstants.java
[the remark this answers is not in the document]
jiajunwang: Reduce the length of code is big deal to me. If this is not a big effort, I suggest we do it.
But this is my own opinion. I guess more ideas here will help. @dasahcc which name do you prefer?

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceComputationStage.java
```diff
@@ -189,56 +189,69 @@ private void processCurrentStates(BaseControllerDataProvider cache,
         String instanceName = instance.getInstanceName();
         String clientSessionId = instance.getEphemeralOwner();
 
-        Map<String, CurrentState> currentStateMap =
-            cache.getCurrentState(instanceName, clientSessionId);
-        if (currentStateMap == null || currentStateMap.size() == 0) {
-          continue;
+        processCurrentStateMap(cache.getCurrentState(instanceName, clientSessionId), resourceMap,
```
jiajunwang, line 192: This is for backward compatibility, right?
Can we create another private method to get the full task current state smartly? So you don't need 2 processCurrentStateMap calls here.

jiajunwang (commented):
I added a minor comment, but the other part looks good to me.
Please ensure tests are stable. Let's run several rounds.
And let's wait until another opinion about the naming : )

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceComputationStage.java
NealSun96: We always need to process current states from both paths. In the resource pipeline, we need the regular current states; in the task pipeline, we need task current states + regular for backward compatibility. 
Could you elaborate on the "getting smartly" part? Do you mean: combine the current state maps so there's only one call? That would incur additional memory and computation cost.

helix-core/src/main/java/org/apache/helix/HelixConstants.java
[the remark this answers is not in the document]
NealSun96: TASK_CURRENT_STATE is a completely fine name in all regard, in my opinion. It has to be at least a wrong name to warranty changes on all variables that expand for 3 PRs, that's my take. I'll also defer to @dasahcc .

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceComputationStage.java
jiajunwang: getCurrentState(boolean isTaskPipeline) {
  currentstates = cache.getCurrentState(...)
  if (isTaskPipeline) {
    filter currentStates for the tasks only
    currentstates.addAll(cache.getTaskCurrentState(...))
  } else {
    filter currentStates for the resources only
  }
  return currentStates
}

dasahcc (approved):
Please add clear todo description as we discussed. Overall, LGTM.

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceComputationStage.java
NealSun96: Right now, it can't be a part of `BaseControllerDataProvider` if that's what you're referring to. `TaskCurrentStateCache` exists in `WorkflowControllerDataProvider` only, and it's for a performance reason since it only needs to be refreshed there, not in the regular pipeline. Without first determining `isTaskPipeline`, we can't parse the base data provider to the workflow data provider.

Of course, we can change how that works. I'll experiment.

alirezazamani:
Do not forget to request to merge it to the Task_CurrentState branch

helix-core/src/main/java/org/apache/helix/HelixConstants.java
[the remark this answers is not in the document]
dasahcc: Personally, TASK_CURRENT_STATE sounds better in this case. That equivalents CURRENT_STATE we have right now. Also if we use TASK_STATE, that could be confused with the TaskState enum class. Actually that is pointing to the concept of Job level state. Although I know this is something not correct, unless we change everything, we need to keep our naming concept consistent.

dasahcc:
Yeah. I dont have major concern and already approve it. If there is no other concerns, feel free to check in.

jiajunwang (approved):

NealSun96:
This PR is ready to be merged, approved by @alirezazamani @dasahcc @jiajunwang 
Final commit message:
## Controller-side Task Current State Migration ##
First part of task current state migration. All changes made in this PR are on the controller side and are directly pipeline-impacting.

PR #1562, opened 2020-11-30
Fix ZkClient leakage by correctly setting usesExternalZkClient flag

### LINKED ISSUE #1561: ZkClient leak in select Helix API classes
### Describe the bug
There was a report saying ZkBaseDataAccessor, when created using its Builder, was leaking ZkClient threads. A similar leak was observed for Verifier classes.

### To Reproduce
The Builder does not properly use a private constructor that sets the usesExternalZkClient flag correctly, so it is reproducible.

### Expected behavior
Create a private constructor used by the Builder and make sure the flag is set correctly. See the example of ConfigAccessor and ZkHelixAdmin - both do this correctly.

narendly:
### Issues

Resolves #1561

### Description

ZkBaseDataAccessor and Helix API that uses it using its Builder were having zkClient thread leak issues because the usesExternalZkClient flag was not being set properly. Also, the family of Verifiers were having a similar issue. This change solves this by making sure the private/protected constructors used by the Builders of these classes set the flag correctly to false so that in their respective close() functions, the underlying zkClient (created by the Builder) would get closed.

### Tests

Manually monitored and tested

> [INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 1251, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:25 h

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -134,14 +134,13 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
    */
   @Deprecated
   public ZkBaseDataAccessor(RealmAwareZkClient zkClient) {
-    if (zkClient == null) {
-      throw new NullPointerException("zkclient is null");
-    }
-    _zkClient = zkClient;
-    _usesExternalZkClient = true;
+    this(zkClient, true);
   }
 
   private ZkBaseDataAccessor(RealmAwareZkClient zkClient, boolean usesExternalZkClient) {
+    if (zkClient == null) {
+      throw new NullPointerException("zkclient is null");
+    }
     _zkClient = zkClient;
     _usesExternalZkClient = usesExternalZkClient;
   }
@@ -1317,9 +1316,11 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
      */
     public ZkBaseDataAccessor<T> build() {
       validate();
+      // Initialize ZkBaseDataAccessor with usesExternalZkClient = false so that
+      // ZkBaseDataAccessor::close() would close ZkClient as well to prevent thread leakage
       return new ZkBaseDataAccessor<>(
           createZkClient(_realmMode, _realmAwareZkConnectionConfig, _realmAwareZkClientConfig,
-              _zkAddress));
+              _zkAddress), false);
     }
   }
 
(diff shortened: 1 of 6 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -134,14 +134,13 @@ public AccessResult() {
...
-      throw new NullPointerException("zkclient is null");
-    }
-    _zkClient = zkClient;
-    _usesExternalZkClient = true;
+    this(zkClient, true);
   }
 
   private ZkBaseDataAccessor(RealmAwareZkClient zkClient, boolean usesExternalZkClient) {
+    if (zkClient == null) {
+      throw new NullPointerException("zkclient is null");
```
lei-xia, line 142: minor: should we throw out InvalidParameterExeception instead?

dasahcc (commented):
A general question, I see some of verifiers use external client to be true, some are false.  Is that on purpose? Could you elaborate more why we do this?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
narendly: Yes. Good catch.

narendly:
> A general question, I see some of verifiers use external client to be true, some are false. Is that on purpose? Could you elaborate more why we do this?

Good question to ask here. 

Some constructors have `usesExternalZkClient = false` and others have `usesExternalZkClient=true`. Why?

The answer lies in where the ZkClient object is being passed in. You will see a pattern where the top-level, public constructors that accept an instance of `zkClient` (external zkclient) have the flag set to true, and the private constructors whose zkClient objects originate from the Builder (created from `createZkClient()` method) will have this flag set to false.

The difference here is that if the flag is set to true, the zkclient will NOT be closed because it is an external ZkClient (that might be used elsewhere - we don't control the fate of this ZkClient). If the flat is set to false, then the ZkClient will be closed because we control this ZkClient (we created it).

@dasahcc let me know if this clarifies it.

dasahcc (approved):
LGTM

narendly:

ZkBaseDataAccessor and Helix API that uses it using its Builder were having zkClient thread leak issues because the usesExternalZkClient flag was not being set properly. Also, the family of Verifiers were having a similar issue. This change solves this by making sure the private/protected constructors used by the Builders of these classes set the flag correctly to false so that in their respective close() functions, the underlying zkClient (created by the Builder) would get closed.

PR #1564, opened 2020-11-30
Implement Helix lock priority and notification

### LINKED ISSUE #1563: Helix lock priority and notification support
Current version of Helix lock does not support priority or preemption, meaning that if a client acquires a lock, no other client can grab it until it's timed out or released. This restricts the usage of Helix lock. In many occasions, priority and preemption is needed for lock, and preferably, when the lower priority client is preempted, it could be notified and given some time to bring the system back to stable state, e.g. do some rollback, etc.

We will define a few concepts of the lock, like priority, different timeout, and whether the lock is forceful. These parameters of a lock will determine how Helix would handle the acquire request. 
We will leverage Zookeeper to perform the notification and client would need to implement the cleanup callback if it is preempted.

**Additional context**

zhangmeng916:
### Issues

Fixed #1563 

### Description

Current version of Helix lock does not support priority or preemption, meaning that if a client acquires a lock, no other client can grab it until it's timed out or released. This restricts the usage of Helix lock. In many occasions, priority and preemption is needed for lock, and preferably, when the lower priority client is preempted, it could be notified and given some time to bring the system back to stable state, e.g. do some rollback, etc. 
In this PR, we define a few concepts of the lock, like priority, different timeout, and whether the lock is forceful. These parameters of a lock will determine how Helix would handle the acquire request.
We also leverage Zookeeper to perform the notification and client would need to implement the cleanup callback if it is preempted.

### Tests

TestZKHelixNonblockingLockWithPriority

helix-core:
[INFO] Tests run: 1253, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,070.423 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:24 h
[INFO] Finished at: 2020-12-14T13:37:09-08:00
                                                                                    
                                                                                   

helix-rest:
[INFO] Tests run: 171, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 239.077 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  04:03 min
[INFO] Finished at: 2020-12-14T12:19:14-08:00
[INFO] -------------------------------------------

### Commits

### Code Quality

### DIFF
--- a/helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
+++ b/helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
@@ -19,6 +19,7 @@ package org.apache.helix.lock;
  * under the License.
  */
 
+import org.apache.helix.lock.helix.LockConstants;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 
 
@@ -26,25 +27,14 @@ import org.apache.helix.zookeeper.datamodel.ZNRecord;
  * Structure represents a lock node information, implemented using ZNRecord
  */
 public class LockInfo {
-
-  // Default values for each attribute if there are no current values set by user
-  public static final String DEFAULT_OWNER_TEXT = "";
-  public static final String DEFAULT_MESSAGE_TEXT = "";
-  public static final long DEFAULT_TIMEOUT_LONG = -1;
-  public static final int DEFAULT_PRIORITY_INT = -1;
-  public static final long DEFAULT_WAITING_TIMEOUT_LONG = -1;
-  public static final long DEFAULT_CLEANUP_TIMEOUT_LONG = -1;
-  public static final String DEFAULT_REQUESTOR_ID = "";
-  public static final int DEFAULT_REQUESTOR_PRIORITY_INT = -1;
-  public static final long DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG = -1;
-  public static final long DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG = -1;
-
   // default lock info represents the status of a unlocked lock
   public static final LockInfo defaultLockInfo =
-      new LockInfo(DEFAULT_OWNER_TEXT, DEFAULT_MESSAGE_TEXT, DEFAULT_TIMEOUT_LONG,
-          DEFAULT_PRIORITY_INT, DEFAULT_WAITING_TIMEOUT_LONG, DEFAULT_CLEANUP_TIMEOUT_LONG,
-          DEFAULT_REQUESTOR_ID, DEFAULT_REQUESTOR_PRIORITY_INT,
-          DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG, DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG);
+      new LockInfo(LockConstants.DEFAULT_OWNER_TEXT, LockConstants.DEFAULT_MESSAGE_TEXT,
+          LockConstants.DEFAULT_TIMEOUT_LONG, LockConstants.DEFAULT_PRIORITY_INT,
+          LockConstants.DEFAULT_WAITING_TIMEOUT_LONG, LockConstants.DEFAULT_CLEANUP_TIMEOUT_LONG,
+          LockConstants.DEFAULT_REQUESTOR_ID, LockConstants.DEFAULT_REQUESTOR_PRIORITY_INT,
+          LockConstants.DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG,
+          LockConstants.DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG);
 
   public static final String ZNODE_ID = "LOCK";
   private ZNRecord _record;
@@ -70,10 +60,12 @@ public class LockInfo {
    */
   private LockInfo() {
     _record = new ZNRecord(ZNODE_ID);
-    setLockInfoFields(DEFAULT_OWNER_TEXT, DEFAULT_MESSAGE_TEXT, DEFAULT_TIMEOUT_LONG,
-        DEFAULT_PRIORITY_INT, DEFAULT_WAITING_TIMEOUT_LONG, DEFAULT_CLEANUP_TIMEOUT_LONG,
-        DEFAULT_REQUESTOR_ID, DEFAULT_REQUESTOR_PRIORITY_INT,
-        DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG, DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG);
+    setLockInfoFields(LockConstants.DEFAULT_OWNER_TEXT, LockConstants.DEFAULT_MESSAGE_TEXT,
+        LockConstants.DEFAULT_TIMEOUT_LONG, LockConstants.DEFAULT_PRIORITY_INT,
+        LockConstants.DEFAULT_WAITING_TIMEOUT_LONG, LockConstants.DEFAULT_CLEANUP_TIMEOUT_LONG,
+        LockConstants.DEFAULT_REQUESTOR_ID, LockConstants.DEFAULT_REQUESTOR_PRIORITY_INT,
+        LockConstants.DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG,
+        LockConstants.DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG);
   }
 
   /**
@@ -85,21 +77,23 @@ public class LockInfo {
     if (znRecord != null) {
       String ownerId = znRecord.getSimpleField(LockInfoAttribute.OWNER.name());
       String message = znRecord.getSimpleField(LockInfoAttribute.MESSAGE.name());
-      long timeout = znRecord.getLongField(LockInfoAttribute.TIMEOUT.name(), DEFAULT_TIMEOUT_LONG);
-      int priority = znRecord.getIntField(LockInfoAttribute.PRIORITY.name(), DEFAULT_PRIORITY_INT);
-      long waitingTimeout = znRecord
-          .getLongField(LockInfoAttribute.WAITING_TIMEOUT.name(), DEFAULT_WAITING_TIMEOUT_LONG);
-      long cleanupTimeout = znRecord
-          .getLongField(LockInfoAttribute.CLEANUP_TIMEOUT.name(), DEFAULT_CLEANUP_TIMEOUT_LONG);
+      long timeout = znRecord
+          .getLongField(LockInfoAttribute.TIMEOUT.name(), LockConstants.DEFAULT_TIMEOUT_LONG);
+      int priority = znRecord
+          .getIntField(LockInfoAttribute.PRIORITY.name(), LockConstants.DEFAULT_PRIORITY_INT);
+      long waitingTimeout = znRecord.getLongField(LockInfoAttribute.WAITING_TIMEOUT.name(),
+          LockConstants.DEFAULT_WAITING_TIMEOUT_LONG);
+      long cleanupTimeout = znRecord.getLongField(LockInfoAttribute.CLEANUP_TIMEOUT.name(),
+          LockConstants.DEFAULT_CLEANUP_TIMEOUT_LONG);
       String requestorId = znRecord.getSimpleField(LockInfoAttribute.REQUESTOR_ID.name());
-      int requestorPriority = znRecord
-          .getIntField(LockInfoAttribute.REQUESTOR_PRIORITY.name(), DEFAULT_REQUESTOR_PRIORITY_INT);
+      int requestorPriority = znRecord.getIntField(LockInfoAttribute.REQUESTOR_PRIORITY.name(),
+          LockConstants.DEFAULT_REQUESTOR_PRIORITY_INT);
       long requestorWaitingTimeout = znRecord
           .getLongField(LockInfoAttribute.REQUESTOR_WAITING_TIMEOUT.name(),
-              DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG);
+              LockConstants.DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG);
       long requestorRequestingTimestamp = znRecord
           .getLongField(LockInfoAttribute.REQUESTOR_REQUESTING_TIMESTAMP.name(),
-              DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG);
+              LockConstants.DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG);
       setLockInfoFields(ownerId, message, timeout, priority, waitingTimeout, cleanupTimeout,
           requestorId, requestorPriority, requestorWaitingTimeout, requestorRequestingTimestamp);
     }
@@ -134,9 +128,9 @@ public class LockInfo {
       long waitingTimeout, long cleanupTimeout, String requestorId, int requestorPriority,
       long requestorWaitingTimeout, long requestorRequestingTimestamp) {
     _record.setSimpleField(LockInfoAttribute.OWNER.name(),
-        ownerId == null ? DEFAULT_OWNER_TEXT : ownerId);
+        ownerId == null ? LockConstants.DEFAULT_OWNER_TEXT : ownerId);
     _record.setSimpleField(LockInfoAttribute.MESSAGE.name(),
-        message == null ? DEFAULT_MESSAGE_TEXT : message);
+        message == null ? LockConstants.DEFAULT_MESSAGE_TEXT : message);
     _record.setLongField(LockInfoAttribute.TIMEOUT.name(), timeout);
     _record.setIntField(LockInfoAttribute.PRIORITY.name(), priority);
     _record.setLongField(LockInfoAttribute.WAITING_TIMEOUT.name(), waitingTimeout);
@@ -155,7 +149,7 @@ public class LockInfo {
    */
   public String getOwner() {
     String owner = _record.getSimpleField(LockInfoAttribute.OWNER.name());
-    return owner == null ? DEFAULT_OWNER_TEXT : owner;
+    return owner == null ? LockConstants.DEFAULT_OWNER_TEXT : owner;
   }
 
   /**
@@ -164,7 +158,7 @@ public class LockInfo {
    */
   public String getMessage() {
     String message = _record.getSimpleField(LockInfoAttribute.MESSAGE.name());
-    return message == null ? DEFAULT_MESSAGE_TEXT : message;
+    return message == null ? LockConstants.DEFAULT_MESSAGE_TEXT : message;
   }
 
   /**
@@ -172,7 +166,8 @@ public class LockInfo {
    * @return the expiration timestamp of the lock, -1 if there is no timeout set
    */
   public Long getTimeout() {
-    return _record.getLongField(LockInfoAttribute.TIMEOUT.name(), DEFAULT_TIMEOUT_LONG);
+    return _record
+        .getLongField(LockInfoAttribute.TIMEOUT.name(), LockConstants.DEFAULT_TIMEOUT_LONG);
   }
 
   /**
@@ -180,7 +175,8 @@ public class LockInfo {
    * @return the priority of the lock, -1 if there is no priority set
    */
   public Integer getPriority() {
-    return _record.getIntField(LockInfoAttribute.PRIORITY.name(), DEFAULT_PRIORITY_INT);
+    return _record
+        .getIntField(LockInfoAttribute.PRIORITY.name(), LockConstants.DEFAULT_PRIORITY_INT);
   }
 
   /**
@@ -188,8 +184,8 @@ public class LockInfo {
    * @return the waiting timeout of the lock, -1 if there is no waiting timeout set
    */
   public Long getWaitingTimeout() {
-    return _record
-        .getLongField(LockInfoAttribute.WAITING_TIMEOUT.name(), DEFAULT_WAITING_TIMEOUT_LONG);
+    return _record.getLongField(LockInfoAttribute.WAITING_TIMEOUT.name(),
+        LockConstants.DEFAULT_WAITING_TIMEOUT_LONG);
   }
 
   /**
@@ -197,8 +193,8 @@ public class LockInfo {
    * @return the cleanup time of the lock, -1 if there is no cleanup timeout set
    */
   public Long getCleanupTimeout() {
-    return _record
-        .getLongField(LockInfoAttribute.CLEANUP_TIMEOUT.name(), DEFAULT_CLEANUP_TIMEOUT_LONG);
+    return _record.getLongField(LockInfoAttribute.CLEANUP_TIMEOUT.name(),
+        LockConstants.DEFAULT_CLEANUP_TIMEOUT_LONG);
   }
 
   /**
@@ -207,7 +203,7 @@ public class LockInfo {
    */
   public String getRequestorId() {
     String requestorId = _record.getSimpleField(LockInfoAttribute.REQUESTOR_ID.name());
-    return requestorId == null ? DEFAULT_REQUESTOR_ID : requestorId;
+    return requestorId == null ? LockConstants.DEFAULT_REQUESTOR_ID : requestorId;
   }
 
   /**
@@ -215,8 +211,8 @@ public class LockInfo {
    * @return the requestor priority of the lock, -1 if there is no requestor priority set
    */
   public int getRequestorPriority() {
-    return _record
-        .getIntField(LockInfoAttribute.REQUESTOR_PRIORITY.name(), DEFAULT_REQUESTOR_PRIORITY_INT);
+    return _record.getIntField(LockInfoAttribute.REQUESTOR_PRIORITY.name(),
+        LockConstants.DEFAULT_REQUESTOR_PRIORITY_INT);
   }
 
   /**
@@ -225,7 +221,7 @@ public class LockInfo {
    */
   public long getRequestorWaitingTimeout() {
     return _record.getLongField(LockInfoAttribute.REQUESTOR_WAITING_TIMEOUT.name(),
-        DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG);
+        LockConstants.DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG);
   }
 
   /**
@@ -235,7 +231,7 @@ public class LockInfo {
    */
   public long getRequestorRequestingTimestamp() {
     return _record.getLongField(LockInfoAttribute.REQUESTOR_REQUESTING_TIMESTAMP.name(),
-        DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG);
+        LockConstants.DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG);
   }
 
   /**
--- a/helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
+++ b/helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
@@ -30,13 +30,23 @@ public class LockInfo {
   // Default values for each attribute if there are no current values set by user
   public static final String DEFAULT_OWNER_TEXT = "";
   public static final String DEFAULT_MESSAGE_TEXT = "";
-  public static final long DEFAULT_TIMEOUT_LONG = -1L;
+  public static final long DEFAULT_TIMEOUT_LONG = -1;
+  public static final int DEFAULT_PRIORITY_INT = -1;
+  public static final long DEFAULT_WAITING_TIMEOUT_LONG = -1;
+  public static final long DEFAULT_CLEANUP_TIMEOUT_LONG = -1;
+  public static final String DEFAULT_REQUESTOR_ID = "";
+  public static final int DEFAULT_REQUESTOR_PRIORITY_INT = -1;
+  public static final long DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG = -1;
+  public static final long DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG = -1;
 
   // default lock info represents the status of a unlocked lock
   public static final LockInfo defaultLockInfo =
-      new LockInfo(DEFAULT_OWNER_TEXT, DEFAULT_MESSAGE_TEXT, DEFAULT_TIMEOUT_LONG);
+      new LockInfo(DEFAULT_OWNER_TEXT, DEFAULT_MESSAGE_TEXT, DEFAULT_TIMEOUT_LONG,
+          DEFAULT_PRIORITY_INT, DEFAULT_WAITING_TIMEOUT_LONG, DEFAULT_CLEANUP_TIMEOUT_LONG,
+          DEFAULT_REQUESTOR_ID, DEFAULT_REQUESTOR_PRIORITY_INT,
+          DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG, DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG);
 
-  private static final String ZNODE_ID = "LOCK";
+  public static final String ZNODE_ID = "LOCK";
   private ZNRecord _record;
 
   /**
@@ -45,7 +55,14 @@ public class LockInfo {
   public enum LockInfoAttribute {
     OWNER,
     MESSAGE,
-    TIMEOUT
+    TIMEOUT,
+    PRIORITY,
+    WAITING_TIMEOUT,
+    CLEANUP_TIMEOUT,
+    REQUESTOR_ID,
+    REQUESTOR_PRIORITY,
+    REQUESTOR_WAITING_TIMEOUT,
+    REQUESTOR_REQUESTING_TIMESTAMP
   }
 
   /**
@@ -53,7 +70,10 @@ public class LockInfo {
    */
   private LockInfo() {
     _record = new ZNRecord(ZNODE_ID);
-    setLockInfoFields(DEFAULT_OWNER_TEXT, DEFAULT_MESSAGE_TEXT, DEFAULT_TIMEOUT_LONG);
+    setLockInfoFields(DEFAULT_OWNER_TEXT, DEFAULT_MESSAGE_TEXT, DEFAULT_TIMEOUT_LONG,
+        DEFAULT_PRIORITY_INT, DEFAULT_WAITING_TIMEOUT_LONG, DEFAULT_CLEANUP_TIMEOUT_LONG,
+        DEFAULT_REQUESTOR_ID, DEFAULT_REQUESTOR_PRIORITY_INT,
+        DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG, DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG);
   }
 
   /**
@@ -66,33 +86,67 @@ public class LockInfo {
       String ownerId = znRecord.getSimpleField(LockInfoAttribute.OWNER.name());
       String message = znRecord.getSimpleField(LockInfoAttribute.MESSAGE.name());
       long timeout = znRecord.getLongField(LockInfoAttribute.TIMEOUT.name(), DEFAULT_TIMEOUT_LONG);
-      setLockInfoFields(ownerId, message, timeout);
+      int priority = znRecord.getIntField(LockInfoAttribute.PRIORITY.name(), DEFAULT_PRIORITY_INT);
+      long waitingTimeout = znRecord
+          .getLongField(LockInfoAttribute.WAITING_TIMEOUT.name(), DEFAULT_WAITING_TIMEOUT_LONG);
+      long cleanupTimeout = znRecord
+          .getLongField(LockInfoAttribute.CLEANUP_TIMEOUT.name(), DEFAULT_CLEANUP_TIMEOUT_LONG);
+      String requestorId = znRecord.getSimpleField(LockInfoAttribute.REQUESTOR_ID.name());
+      int requestorPriority = znRecord
+          .getIntField(LockInfoAttribute.REQUESTOR_PRIORITY.name(), DEFAULT_REQUESTOR_PRIORITY_INT);
+      long requestorWaitingTimeout = znRecord
+          .getLongField(LockInfoAttribute.REQUESTOR_WAITING_TIMEOUT.name(),
+              DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG);
+      long requestorRequestingTimestamp = znRecord
+          .getLongField(LockInfoAttribute.REQUESTOR_REQUESTING_TIMESTAMP.name(),
+              DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG);
+      setLockInfoFields(ownerId, message, timeout, priority, waitingTimeout, cleanupTimeout,
+          requestorId, requestorPriority, requestorWaitingTimeout, requestorRequestingTimestamp);
     }
   }
 
   /**
-   * Initialize a LockInfo with data for each field, set all null info fields to default data
+   * Initialize a LockInfo with data for each field, set all null info fields to default data.
    * @param ownerId value of OWNER attribute
    * @param message value of MESSAGE attribute
    * @param timeout value of TIMEOUT attribute
+   * @param priority value of PRIORITY attribute
+   * @param waitingTimout value of WAITING_TIMEOUT attribute
+   * @param cleanupTimeout value of CLEANUP_TIMEOUT attribute
+   * @param requestorId value of REQUESTOR_ID attribute
+   * @param requestorPriority value of REQUESTOR_PRIORITY attribute
+   * @param requestorWaitingTimeout value of REQUESTOR_WAITING_TIMEOUT attribute
+   * @param requestorRequestingTimestamp value of REQUESTOR_REQUESTING_TIMESTAMP attribute
    */
-  public LockInfo(String ownerId, String message, long timeout) {
+  public LockInfo(String ownerId, String message, long timeout, int priority, long waitingTimout,
+      long cleanupTimeout, String requestorId, int requestorPriority, long requestorWaitingTimeout,
+      long requestorRequestingTimestamp) {
     this();
-    setLockInfoFields(ownerId, message, timeout);
+    setLockInfoFields(ownerId, message, timeout, priority, waitingTimout, cleanupTimeout,
+        requestorId, requestorPriority, requestorWaitingTimeout, requestorRequestingTimestamp);
   }
 
   /**
-   * Set each field of lock info to user provided values if the values are not null, null values are set to default values
-   * @param ownerId value of OWNER attribute
-   * @param message value of MESSAGE attribute
-   * @param timeout value of TIMEOUT attribute
+   * Set each field of lock info to user provided values if the values are not null. Null values
+   * are set to default values.
    */
-  private void setLockInfoFields(String ownerId, String message, long timeout) {
+  private void setLockInfoFields(String ownerId, String message, long timeout, int priority,
+      long waitingTimeout, long cleanupTimeout, String requestorId, int requestorPriority,
+      long requestorWaitingTimeout, long requestorRequestingTimestamp) {
     _record.setSimpleField(LockInfoAttribute.OWNER.name(),
         ownerId == null ? DEFAULT_OWNER_TEXT : ownerId);
     _record.setSimpleField(LockInfoAttribute.MESSAGE.name(),
         message == null ? DEFAULT_MESSAGE_TEXT : message);
     _record.setLongField(LockInfoAttribute.TIMEOUT.name(), timeout);
+    _record.setIntField(LockInfoAttribute.PRIORITY.name(), priority);
+    _record.setLongField(LockInfoAttribute.WAITING_TIMEOUT.name(), waitingTimeout);
+    _record.setLongField(LockInfoAttribute.CLEANUP_TIMEOUT.name(), cleanupTimeout);
+    _record.setSimpleField(LockInfoAttribute.REQUESTOR_ID.name(), requestorId);
+    _record.setIntField(LockInfoAttribute.REQUESTOR_PRIORITY.name(), requestorPriority);
+    _record
+        .setLongField(LockInfoAttribute.REQUESTOR_WAITING_TIMEOUT.name(), requestorWaitingTimeout);
+    _record.setLongField(LockInfoAttribute.REQUESTOR_REQUESTING_TIMESTAMP.name(),
+        requestorRequestingTimestamp);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -30,13 +30,23 @@
...
   public static final String DEFAULT_MESSAGE_TEXT = "";
-  public static final long DEFAULT_TIMEOUT_LONG = -1L;
+  public static final long DEFAULT_TIMEOUT_LONG = -1;
+  public static final int DEFAULT_PRIORITY_INT = -1;
+  public static final long DEFAULT_WAITING_TIMEOUT_LONG = -1;
+  public static final long DEFAULT_CLEANUP_TIMEOUT_LONG = -1;
+  public static final String DEFAULT_REQUESTOR_ID = "";
+  public static final int DEFAULT_REQUESTOR_PRIORITY_INT = -1;
+  public static final long DEFAULT_REQUESTOR_WAITING_TIMEOUT_LONG = -1;
+  public static final long DEFAULT_REQUESTOR_REQUESTING_TIMESTAMP_LONG = -1;
```
dasahcc, line 40: Shall we have a LockConstant class to hold them? Seemed they are too many in same class with logic code.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -31,69 +31,162 @@
...
-      deadline = System.currentTimeMillis() + _timeout;
+    _baseDataAccessor.subscribeDataChanges(_lockPath, this);
+    LockUpdater updater = new LockUpdater(
+        new LockInfo(_userId, _lockMsg, getNonOverflowTimestamp(_leaseTimeout), _priority,
+            _waitingTimeout, _cleanupTimeout, null, 0, 0, 0));
+    boolean updateResult = _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
+
+    // Check whether the lock request is still pending. If yes, we will wait for the period
+    // recorded in _pendingTimeout.
+    if (_isPending) {
```
dasahcc, line 160: Usually, if we are waiting for something, we should use while to do the checks here.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -31,69 +31,162 @@
...
+      try {
+        wait(_pendingTimeout);
```
dasahcc, line 162: Are we just waiting until timeout and do a check? Shall we periodically wake up and check whether the lock has been released or not?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -125,6 +218,79 @@ public void close() {
     _baseDataAccessor.close();
   }
 
+  @Override
+  public void handleDataChange(String dataPath, Object data) throws Exception {
```
dasahcc, line 222: Can we leverage this handleDataChange to notify the waiting thread? Then we can use some Java internal lock.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: This wait() will be signaled by the following notify(), or if it never receives the notify, it'll wait until timeout.

  public void onAcquiredLockNotification() {
    synchronized (ZKDistributedNonblockingLock.this) {
      _isLocked = true;
      ZKDistributedNonblockingLock.this.notify();
    }
  }

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: This boolean variable is used as a flag to distinguish between different scenarios so that we can act differently. The actual functioning part is the following wait().

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
zhangmeng916: Good suggestion! Will do.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: Yeah, we are using this one to trigger a onAcquiredLockNotification callback, which will notify the waiting thread.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -31,69 +31,155 @@
...
-  public boolean tryLock() {
+  public synchronized boolean tryLock() {
     // Set lock information fields
-    long deadline;
-    // Prevent value overflow
-    if (_timeout > Long.MAX_VALUE - System.currentTimeMillis()) {
-      deadline = Long.MAX_VALUE;
-    } else {
-      deadline = System.currentTimeMillis() + _timeout;
+    _baseDataAccessor.subscribeDataChanges(_lockPath, this);
```
jiajunwang, line 145: I didn't see when this subscription is removed. Did I miss anything?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -31,69 +31,155 @@
...
+    LockUpdater updater = new LockUpdater(
+        new LockInfo(_userId, _lockMsg, getNonOverflowTimestamp(_leaseTimeout), _priority,
```
jiajunwang, line 147: Not related to this PR directly, but the isCurrentOwner() call inside the LockUpdater.update() triggers one more ZK call which is not necessary since you already have the current ZNRecord input. Please refine that logic.
The logic is not only suboptimal but also buggy since the update logic is not purely relying on the input current znode. This potentially fails when some race condition happens.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -31,69 +31,155 @@
...
+            _waitingTimeout, _cleanupTimeout, null, 0, 0, 0));
+    boolean updateResult = _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
```
jiajunwang, line 149: If update() fails, it throws Exception, right? Shall we try-catch it and return false instead?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
jiajunwang: The reason to put wait in the while loop is to prevent if anyone else updates the content after the notification and before the current thread regains the lock. I think what Junkai said is a valid comment.

jiajunwang (commented):
I haven't finished reviewing the detailed internal logic yet, but please check my initial comments first. I will keep reviewing the rest of the code.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -125,6 +211,79 @@ public void close() {
...
+  public void handleDataChange(String dataPath, Object data) throws Exception {
+    Stat stat = new Stat();
+    ZNRecord readData =
+        _baseDataAccessor.get(dataPath, stat, AccessOption.THROW_EXCEPTION_IFNOTEXIST);
+    LockInfo lockInfo = new LockInfo(readData);
+    // We are the current owner
+    if (lockInfo.getOwner().equals(_userId)) {
+      if (lockInfo.getRequestorId() == null || lockInfo.getPriority() > lockInfo
+          .getRequestorPriority()) {
+        LOG.info("We do not need to handle this data change");
```
jiajunwang, line 224: Verbose log. debug?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -125,6 +211,79 @@ public void close() {
...
 
+  @Override
+  public void handleDataChange(String dataPath, Object data) throws Exception {
+    Stat stat = new Stat();
+    ZNRecord readData =
+        _baseDataAccessor.get(dataPath, stat, AccessOption.THROW_EXCEPTION_IFNOTEXIST);
+    LockInfo lockInfo = new LockInfo(readData);
+    // We are the current owner
+    if (lockInfo.getOwner().equals(_userId)) {
+      if (lockInfo.getRequestorId() == null || lockInfo.getPriority() > lockInfo
```
jiajunwang, line 222: The Default ID is "" in this version, right? I guess we should check == null || .isEmpty() ?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -125,6 +211,79 @@ public void close() {
...
+          .getRequestorPriority()) {
+        LOG.info("We do not need to handle this data change");
+      } else {
+        _lockListener.onCleanupNotification();
```
jiajunwang, line 226: Not about code, but this means we don't have control on when the cleanup starts.
Would it be possible that the cleaner lock holder also updates the node so we are sure that cleaning up has been started?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -125,6 +211,79 @@ public void close() {
...
+        // read the lock information again to avoid stale data
```
jiajunwang, line 227: Could you please elaborate on this part? I didn't get the necessity of reading again.
You can relies on the updater to check the ownership for you, right? If the owner has been changed in between, then the update fails and that is OK.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -125,6 +211,79 @@ public void close() {
...
+          znRecord.setLongField(LockInfo.LockInfoAttribute.WAITING_TIMEOUT.name(), -1);
+          znRecord.setLongField(LockInfo.LockInfoAttribute.CLEANUP_TIMEOUT.name(), -1);
+          _baseDataAccessor
+              .update(_lockPath, new ZNRecordUpdater(znRecord), AccessOption.PERSISTENT);
+        } else {
+          LOG.info("We are not current lock owner");
+        }
+      }
+    } // We are the current requestor
+    else if (lockInfo.getRequestorId().equals(_userId)) {
```
jiajunwang, line 245: I cannot match it easily with the if condition above, but they look like the same : )

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -144,19 +303,88 @@ public ZNRecord update(ZNRecord current) {
       // If no one owns the lock, allow the update
       // If the user is the current lock owner, allow the update
       LockInfo curLockInfo = new LockInfo(current);
-      if (!(System.currentTimeMillis() < curLockInfo.getTimeout()) || isCurrentOwner()) {
+      if (System.currentTimeMillis() > curLockInfo.getTimeout() || isCurrentOwner()) {
```
jiajunwang, line 306: Same here, I'm concerned about the isCurrentOwner usage inside the update method.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: The reason I was doing another read is that I wasn't using lock updater, but use a normal updater. I feel for these callback cases, we should not use lock updater, which is used during first tryLock. The reason is that we usually know what we are doing in the callback updater, and put all these logic inside lock updater will make that logic very complicated and hard to read. 
For example, in the lock updater, we always assume there is a normal lock request, which has owner ID, etc. However, the cleanup callback updater actually removes the owner ID. Do you think we should combine all scenarios in one updater?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: The if condition is:
if (lockInfo.getOwner().equals(_userId))

Basically we need to distinguish between requestor and owner. Both of them may receive callback, but they don't need to handle unrelated callback.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: If there's another user (whose priority is even higher) acquired the lock, the onDeniedPendingLockNotification will be called, which will terminate the wait. We won't need to worry about this, right?

helix-lock/src/test/java/org/apache/helix/lock/helix/TestZKHelixNonblockingLockWithPriority.java
```diff
@@ -0,0 +1,334 @@
...
+  private String _lockPath;
+  private HelixLockScope _participantScope;
+  private AtomicBoolean _isCleanupNotified;
+
+  private final LockListener _lockListener = new LockListener() {
+    @Override
+    public void onCleanupNotification() {
+      _isCleanupNotified.set(true);
+      try {
+        Thread.sleep(20000);
```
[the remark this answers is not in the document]
alirezazamani: Ok. Makes sense

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -122,9 +210,91 @@ public void close() {
...
+      onDeniedPendingLockNotification();
+    }
+  }
+
+  /**
+   * call back called when the lock is acquired
+   */
+  public void onAcquiredLockNotification() {
+    synchronized (ZKDistributedNonblockingLock.this) {
+      _isLocked = true;
```
jiajunwang, line 277: I think _isPending would be better to be updated here.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -122,9 +210,91 @@ public void close() {
...
+      ZKDistributedNonblockingLock.this.notify();
+    }
+  }
+
+  /**
+   * call back called when the pending request is denied due to another higher priority request
+   */
+  public void onDeniedPendingLockNotification() {
+    synchronized (ZKDistributedNonblockingLock.this) {
+      _isLocked = false;
```
jiajunwang, line 287: Looks like a lock can only be in one state at a time. I suggest merging these 2 booleans into an enum for the cleanness of the code and the logic.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -144,19 +314,91 @@ public ZNRecord update(ZNRecord current) {
...
+      }
+
+      // For users who are not the lock owner and the priority is not higher than current lock
+      // owner, throw an exception. The exception will be caught by data accessor, and return
+      // false for the update
       LOG.error(
           "User " + _userId + " tried to update the lock at " + new Date(System.currentTimeMillis())
               + ". Lock path: " + _lockPath);
+
       throw new HelixException("User is not authorized to perform this operation.");
```
dasahcc, line 353: The message could be confusing here. We can say the user has failed to acquire the lock.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -144,19 +314,91 @@ public ZNRecord update(ZNRecord current) {
...
-      // For users who are not the lock owner and try to do an update on a lock that is held by
-      // someone else, exception thrown is to be caught by data accessor, and return false for
-      // the update
+
+      // higher priority lock request will preempt current lock owner that is with lower priority
+      if (!isCurrentOwner(curLockInfo) && _priority > curLockInfo.getPriority() && curLockInfo
+          .getRequestorId().equals("NONE")) {
+        // update lock Znode with requestor information
+        ZNRecord newRecord = composeNewRequestorRecord(curLockInfo, _record);
+        _isPending = true;
```
jiajunwang, line 326: As mentioned above, I think this would be better to set in the notification method.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -144,19 +314,91 @@ public ZNRecord update(ZNRecord current) {
...
+        _pendingTimeout = _waitingTimeout > curLockInfo.getCleanupTimeout() ? curLockInfo.getCleanupTimeout()
+                : _waitingTimeout;
+        return newRecord;
+      }
+
+      // If the requestor field is not empty, and the coming lock request has a even higher
+      // priority. The new request will replace current requestor field of the lock
+      if (!isCurrentOwner(curLockInfo) && _priority > curLockInfo.getPriority() && !curLockInfo
+          .getRequestorId().equals("NONE") && _priority > curLockInfo.getRequestorPriority()) {
```
dasahcc, line 335: This could be a sub branch of previous one since the previous two checks are same. Can we combine them?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -144,19 +314,91 @@ public ZNRecord update(ZNRecord current) {
       // If no one owns the lock, allow the update
       // If the user is the current lock owner, allow the update
       LockInfo curLockInfo = new LockInfo(current);
-      if (!(System.currentTimeMillis() < curLockInfo.getTimeout()) || isCurrentOwner()) {
+      if (System.currentTimeMillis() > curLockInfo.getTimeout() || isCurrentOwner(curLockInfo)) {
```
dasahcc, line 317: What if there is no owner at first beginning? Even no ZNode exist?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -122,9 +210,91 @@ public void close() {
...
+        LockUpdater updater = new LockUpdater(new LockInfo(znRecord));
+        _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
+        onAcquiredLockNotification();
+      } else {
+        LOG.info("We do not need to handle this data change");
+      }
+    } // If we are waiting for the lock, but find we are not the requestor any more, meaning we
+    // are preempted by an even higher priority request
+    else if (!lockInfo.getRequestorId().equals("NONE") && !lockInfo.getRequestorId().equals(_userId)
+        && _isPending) {
```
dasahcc, line 267: Do we need to check this _isPending? I think we can notify anyway if it is pending or not pending.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -31,69 +31,157 @@
...
+  private final long _leaseTimeout;
+  private final long _waitingTimeout;
+  private final long _cleanupTimeout;
+  private final int _priority;
+  private final boolean _isForceful;
+  private final LockListener _lockListener;
   private final BaseDataAccessor<ZNRecord> _baseDataAccessor;
+  private boolean _isLocked;
+  private boolean _isPending;
+  private boolean _isPreempted;
```
dasahcc, line 64: I would suggest use an enum to represent its current state:

LOCKED, PENDING, PREEMPTED. Otherwise, three flags can be confusing for mutual usage to present the lock state.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -31,69 +31,157 @@
...
+          } else {
+            _baseDataAccessor.unsubscribeDataChanges(_lockPath, this);
+            throw new HelixException("Cleanup has not been finished by lock owner");
+          }
+        }
+      }
     }
-    LockUpdater updater = new LockUpdater(new LockInfo(_userId, _lockMsg, deadline));
-    return _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
+    if (!updateResult) {
```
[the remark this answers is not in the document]
zhangmeng916: The isLocked is actually inside isPending condition, it checks whether the lock is acquired after the waiting timeout expired. And this updateResult is the immediate response from update operation. So it can be true or false depending on the request's priority and owner. E.g. if the request is not the owner and has a lower priority than the owner, the response will be false, and there is no pending.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: Yeah, there won't be Znode, and the input value is null.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -31,69 +31,157 @@
...
+    if (_isPending) {
+      try {
+        wait(_pendingTimeout);
+      } catch (InterruptedException e) {
+        throw new HelixException(
+            String.format("Interruption happened while %s is waiting for the " + "lock", _userId),
+            e);
+      }
+      // We have not acquired the lock yet.
+      if (!_isLocked) {
```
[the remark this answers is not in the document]
zhangmeng916: This _isLocked check is actually inside _isPending. It is only checked after wait timed out. Other immediately returned results are outside _isPending.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: There is no notification that would be received by the requestor before it gets the lock. So we cannot se the pending state for it. The notification it receives is that the previous owner finished the cleanup work and it acquires the lock.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: Please see the comments below. Here the status should be LOCKED.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: Clients need to implement the cleanup callback in their logic. We only trigger the registered listener.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -31,69 +31,157 @@
...
+   * @param userId a universal unique userId for lock owner identity
+   * @param priority the priority of the lock
+   * @param waitingTimeout the waiting timeout period of the lock when the tryLock request is issued
+   * @param cleanupTimeout the time period needed to finish the cleanup work by the lock when it
+   *                      is preempted
+   * @param isForceful whether the lock is a forceful one. This determines the behavior when the
+   *                   lock encountered an exception during preempting lower priority lock
+   * @param lockListener the listener associated to the lock
+   */
+  public ZKDistributedNonblockingLock(LockScope scope, String zkAddress, Long leaseTimeout,
```
pkuwm, line 96: It seems `leaseTimeout` does not need `Long` boxing. `long` is fine?

And, by seeing the list of params, just wow... So many params :)
Can we encapsulate some of the params into an object, eg. `ZkDistributedLockProperties` or `ZkDistributedLockConfig`.

zhangmeng916: long should be fine, but this is old API. I'm concerned about backward compatibility.
A wrapping class sounds good. I'll give it a try.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -122,9 +210,91 @@ public void close() {
...
+          znRecord.setLongField(LockInfo.LockInfoAttribute.TIMEOUT.name(),
+              LockConstants.DEFAULT_TIMEOUT_LONG);
+          znRecord.setIntField(LockInfo.LockInfoAttribute.PRIORITY.name(),
+              LockConstants.DEFAULT_PRIORITY_INT);
+          znRecord.setLongField(LockInfo.LockInfoAttribute.WAITING_TIMEOUT.name(),
+              LockConstants.DEFAULT_WAITING_TIMEOUT_LONG);
+          znRecord.setLongField(LockInfo.LockInfoAttribute.CLEANUP_TIMEOUT.name(),
+              LockConstants.DEFAULT_CLEANUP_TIMEOUT_LONG);
+          _baseDataAccessor
+              .update(_lockPath, new ZNRecordUpdater(znRecord), AccessOption.PERSISTENT);
```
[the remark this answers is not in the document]
zhangmeng916: I should use LockInfo instead of directly setting it in ZNRecord. Updated.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: Actually we need, because this callback can also be received by previous owner (who already cleaned up the ownership), and it should not act since it's not pending.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: I see your point. So here we cannot use while because pending is not the only checking condition. We also have timeout, which can terminate the wait. We cannot easily distinguish between spurious wake up and timeout. Using a timer is not accurate. I changed to use countdown latch, which is more convenient and doesn't have spurious wake up issue.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
dasahcc: Sorry. I am asking what if there is no ZNode, will this function breaking? Shall we create one? Because creating could also have race condition happening. How we handle the create fail.

zhangmeng916: Yeah, it should work. This is the design in MVP, and I think it has been tested by customers. 
The logic is, if there is no znode existing, we create LockInfo with default value, and take that as current value, update that value with user's requested lock information.

zhangmeng916: Also regarding the above question `Are we notifying them or we give them callback and process clean up here?`
Seems I cannot directly reply, so just answer here. We provide an interface of cleanup, and clients should implement that cleanup callback.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -65,34 +77,70 @@ public LockInfo(ZNRecord znRecord) {
     if (znRecord != null) {
       String ownerId = znRecord.getSimpleField(LockInfoAttribute.OWNER.name());
       String message = znRecord.getSimpleField(LockInfoAttribute.MESSAGE.name());
-      long timeout = znRecord.getLongField(LockInfoAttribute.TIMEOUT.name(), DEFAULT_TIMEOUT_LONG);
-      setLockInfoFields(ownerId, message, timeout);
+      long timeout = znRecord
```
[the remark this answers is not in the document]
jiajunwang: It would still be cleaner to just do a simple check instead of initializing the values 2 times.
For instance,
if (znRecord == null) {
  znRecord = new znRecord();
}
...
Then the default value will for sure be set in the same way since this record is empty.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -65,34 +77,70 @@ public LockInfo(ZNRecord znRecord) {
...
+   * @param priority value of PRIORITY attribute
+   * @param waitingTimout value of WAITING_TIMEOUT attribute
+   * @param cleanupTimeout value of CLEANUP_TIMEOUT attribute
+   * @param requestorId value of REQUESTOR_ID attribute
+   * @param requestorPriority value of REQUESTOR_PRIORITY attribute
+   * @param requestorWaitingTimeout value of REQUESTOR_WAITING_TIMEOUT attribute
+   * @param requestorRequestingTimestamp value of REQUESTOR_REQUESTING_TIMESTAMP attribute
    */
-  public LockInfo(String ownerId, String message, long timeout) {
+  public LockInfo(String ownerId, String message, long timeout, int priority, long waitingTimout,
```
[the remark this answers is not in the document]
jiajunwang: I understand logic-wise, the API will always work. But the question is that API is not easy for the users to use. It requires much knowledge and there are invisible dependencies here and there.
A clean API should be either one of the following cases,
1. No matter how the parameters look like, ALL of them works in an intuitive way.
2. If the parameters are not valid, then throw an exception to reject the call.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
jiajunwang: My proposal is that do the following steps in our code.

1. update znode about the cleanup start time.
2. trigger clean up callbacks.

So the cleanup start time is a more accurate sign than guessing.
I'm fine if we just add a TODO here about step 1.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
jiajunwang: Yeah, so _isPending should be false.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
jiajunwang: Isn't it receiving notification on preempting?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
jiajunwang: But anyway this should be resolved when you change to use enum instead of booleans, right?

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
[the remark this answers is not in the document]
zhangmeng916: All the new input parameters added this time are optional. I added a new validation function to throw exception if required field is missing or invalid value is given.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: Yeah, it's already changed to an enum. Please see the updated code.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: Notification would be received by all listeners. But for requestor, it won't handle its own preempting request. Only owner needs to start cleanup. Requestor only handle data change when owner finishes the cleanup work and remove itself from the owner field.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: I agree with you that current time calculation is not accurate enough. But even with recording cleanup start time in znode, we still have a problem letting the requestor know the start time and adjust its waiting. Right now, the waiting time is calculated when requestor updates the znode, and cleanup update znode happens after requestor updates znode. How can we adjust the waiting time for requestor in this case?

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
```diff
@@ -45,54 +40,105 @@
...
+      String requestorId = znRecord.getSimpleField(LockInfoAttribute.REQUESTOR_ID.name());
+      int requestorPriority = znRecord.getIntField(LockInfoAttribute.REQUESTOR_PRIORITY.name(),
+          LockConstants.DEFAULT_PRIORITY_INT);
+      long requestorWaitingTimeout = znRecord
+          .getLongField(LockInfoAttribute.REQUESTOR_WAITING_TIMEOUT.name(),
+              LockConstants.DEFAULT_WAITING_TIMEOUT_LONG);
+      long requestingTimestamp = znRecord
+          .getLongField(LockInfoAttribute.REQUESTING_TIMESTAMP.name(),
+              LockConstants.DEFAULT_REQUESTING_TIMESTAMP_LONG);
+      setLockInfoFields(ownerId, message, timeout, priority, waitingTimeout, cleanupTimeout,
```
jiajunwang, line 97: nit, combine this with line 71, then we can call the constructor directly instead of 2 lines.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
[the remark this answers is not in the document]
jiajunwang: Sounds good 👌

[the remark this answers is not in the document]
jiajunwang: I didn't see the validation method though.

helix-lock/src/main/java/org/apache/helix/lock/helix/LockConstants.java
```diff
@@ -0,0 +1,40 @@
...
+public class LockConstants {
+  public static final String DEFAULT_USER_ID = "NONE";
+  public static final String DEFAULT_MESSAGE_TEXT = "NONE";
+  public static final long DEFAULT_TIMEOUT_LONG = -1;
+  public static final int DEFAULT_PRIORITY_INT = -1;
+  public static final long DEFAULT_WAITING_TIMEOUT_LONG = -1;
+  public static final long DEFAULT_CLEANUP_TIMEOUT_LONG = -1;
+  public static final long DEFAULT_REQUESTING_TIMESTAMP_LONG = -1;
+
+  public enum LockStatus {
```
jiajunwang, line 35: nit, please comment on the meaning of these statuses, especially PREEMPTED : )

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -32,68 +33,167 @@
...
    * Initialize the lock with user provided information, e.g.,cluster, scope, etc.
    * @param scope the scope to lock
    * @param zkAddress the zk address the cluster connects to
-   * @param timeout the timeout period of the lock
+   * @param leaseTimeout the leasing timeout period of the lock
    * @param lockMsg the reason for having this lock
    * @param userId a universal unique userId for lock owner identity
    */
-  public ZKDistributedNonblockingLock(LockScope scope, String zkAddress, Long timeout,
+  public ZKDistributedNonblockingLock(LockScope scope, String zkAddress, Long leaseTimeout,
```
jiajunwang, line 72: I just noticed that there is zkAddress input. So this is not multi zk supported API, right? Will that be a concern?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -32,68 +33,167 @@
...
     }
-    LockUpdater updater = new LockUpdater(new LockInfo(_userId, _lockMsg, deadline));
-    return _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
+
+    // When the lock status is still pending, wait for the period recorded in _pendingTimeout.
+    try {
+      _countDownLatch.await(_pendingTimeout, TimeUnit.MILLISECONDS);
+    } catch (InterruptedException e) {
+      throw new HelixException(
+          String.format("Interruption happened while %s is waiting for the " + "lock", _userId), e);
```
jiajunwang, line 171: nit,

> the " + "lock"

I assume you want to add something in between?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
jiajunwang: The proposal is that the waiting time should start once cleanup starts. But that could be too complicated. So let's just add a TODO here. We shall only do this when it is proved to be required.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -122,11 +222,72 @@ public void close() {
...
+   * call back called when the pending request is denied due to another higher priority request
+   */
+  public void onDeniedPendingLockNotification() {
+    _lockStatus = LockConstants.LockStatus.PREEMPTED;
+    _countDownLatch.countDown();
+  }
+
+  @Override
+  public void handleDataDeleted(String dataPath) throws Exception {
+  }
```
jiajunwang, line 286: Once we support LOCK node cleanup, then we need to add some logic here, right?
Maybe some TODO comments would be helpful.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -144,19 +305,143 @@ public ZNRecord update(ZNRecord current) {
...
-      // someone else, exception thrown is to be caught by data accessor, and return false for
-      // the update
-      LOG.error(
-          "User " + _userId + " tried to update the lock at " + new Date(System.currentTimeMillis())
-              + ". Lock path: " + _lockPath);
-      throw new HelixException("User is not authorized to perform this operation.");
+
+      // higher priority lock request will try to  preempt current lock owner
+      if (!isCurrentOwner(curLockInfo) && _priority > curLockInfo.getPriority()) {
+        if (curLockInfo.getRequestorId().equals("NONE")) {
```
jiajunwang, line 314: Can we avoid hardcode NONE but refer to the default ID?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -144,19 +305,143 @@ public ZNRecord update(ZNRecord current) {
...
+        else if (_priority > curLockInfo.getRequestorPriority()) {
+          ZNRecord newRecord = composeNewRequestorRecord(curLockInfo, _record);
+          _lockStatus = LockConstants.LockStatus.PENDING;
+          long remainingCleanupTime =
+              curLockInfo.getCleanupTimeout() - (System.currentTimeMillis() - curLockInfo
+                  .getRequestingTimestamp());
+          _pendingTimeout = _waitingTimeout > remainingCleanupTime ? remainingCleanupTime :
+           _waitingTimeout;
+          return newRecord;
+        }
```
jiajunwang, line 333: There is some duplicate code, can we do this?

if (curLockInfo.getRequestorId().equals("NONE")) {
_pendingTimeout = .....
} else if (_priority > curLockInfo.getRequestorPriority()) {
_pendingTimeout = ......
}
ZNRecord newRecord = composeNewRequestorRecord(curLockInfo, _record);
      throw new HelixException("User is not authorized to perform this operation.");
 _lockStatus = LockConstants.LockStatus.PENDING;
return newRecord;

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -144,19 +305,143 @@ public ZNRecord update(ZNRecord current) {
...
+  private class CleanupUpdater implements DataUpdater<ZNRecord> {
+    public CleanupUpdater() {
+    }
+
+    @Override
+    public ZNRecord update(ZNRecord current) {
+      // If we are still the lock owner, clean owner field.
+      LockInfo curLockInfo = new LockInfo(current);
+      if (isCurrentOwner(curLockInfo)) {
+        ZNRecord record = current;
```
jiajunwang, line 359: If I just create a new LockInfo(null) object, then I get the same result, right?
I guess maybe you want to make the LockInfo(<no parameter>) constructor public.

jiajunwang: Or even better, just define a static empty LockInfo object for this usage.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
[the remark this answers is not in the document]
zhangmeng916: I actually added the validation logic in `ZKDistributedNonblockingLock` in the constructor. Did you suggest to add in this file? I feel it's more of a validation of customer's input.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -144,19 +305,143 @@ public ZNRecord update(ZNRecord current) {
...
+        return _record;
+      }
+      LOG.error("User {} is not current lock requestor, and cannot forcefully acquire the lock at "
+          + "{}", _userId, _lockPath);
+      throw new HelixException(String.format("User %s is not current lock requestor, and cannot "
+          + "forcefully acquire the lock at %s", _userId, _lockPath));
+    }
+  }
+
+  private ZNRecord composeNewRequestorRecord(LockInfo oldLockInfo, ZNRecord newLockZNRecord) {
```
jiajunwang, line 410: nit, 
oldLockInfo => existingLockInfo, and newLockZNRecord => requestorLockZNRecord ?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKLockConfig.java
```diff
@@ -0,0 +1,173 @@
...
+
+
+/**
+ * Hold configs used for a ZK distributed nonblocking lock.
+ */
+public class ZKLockConfig {
+  private static final Logger LOG =
+      LoggerFactory.getLogger(org.apache.helix.lock.helix.ZKLockConfig.class.getName());
+  private LockScope _lockScope;
+  private String _zkAddress;
```
jiajunwang, line 35: I do have some concerns about this direct ZkAddress parameter. This won't support multi-zk usage.
It might be OK, but the LOCK API will be treated as a special case, I guess.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: I believe it does support multi zk. Underlying it calls 
_zkClient = buildRealmAwareZkClientWithDefaultConfigs(
        new RealmAwareZkClient.RealmAwareZkClientConfig().setZkSerializer(zkSerializer), zkAddress,
        zkClientType);
If the zk address is null, it'll use default system property config to build a multi zk client. Right?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
zhangmeng916: Here we only clean up the owner related fields, but still contain requestor related fields, so that requestor can go ahead to acquire the lock. So we need to get the current lock and update based on that lock, instead of creating a default lock node.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKLockConfig.java
zhangmeng916: Please see the above comment. I feel multi ZK should be implicitly supported. Let me know if that's not the case.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
jiajunwang: I see. With the enum status instead of booleans, the logic has been clean enough. So I think it is good enough.

helix-lock/src/main/java/org/apache/helix/lock/LockInfo.java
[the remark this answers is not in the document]
jiajunwang: np, I saw it later.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -32,68 +33,167 @@
...
+   * @param userId a universal unique userId for lock owner identity
+   * @param priority the priority of the lock
+   * @param waitingTimeout the waiting timeout period of the lock when the tryLock request is issued
+   * @param cleanupTimeout the time period needed to finish the cleanup work by the lock when it
+   *                      is preempted
+   * @param isForceful whether the lock is a forceful one. This determines the behavior when the
+   *                   lock encountered an exception during preempting lower priority lock
+   * @param lockListener the listener associated to the lock
+   */
+  public ZKDistributedNonblockingLock(LockScope scope, String zkAddress, Long leaseTimeout,
```
jiajunwang, line 104: I think we want users to use zooscalibility features as far as they can. So shall we hide this constructor? The above one with zkAddress is left because we want to ensure backward compatibility. But this one is newly added. So make it private?

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
```diff
@@ -32,68 +33,167 @@
...
-    } else {
-      deadline = System.currentTimeMillis() + _timeout;
+    _baseDataAccessor.subscribeDataChanges(_lockPath, this);
+    LockUpdater updater = new LockUpdater(
+        new LockInfo(_userId, _lockMsg, getNonOverflowTimestamp(_leaseTimeout), _priority,
+            _waitingTimeout, _cleanupTimeout, null, 0, 0, 0));
+    boolean updateResult = _baseDataAccessor.update(_lockPath, updater, AccessOption.PERSISTENT);
+
+    // Immediately return if the lock statue is not PENDING.
+    if (_lockStatus != LockConstants.LockStatus.PENDING) {
```
dasahcc, line 159: Following lockStatus was protected by latch down. But this check could happening parallel to following lockStatus update. Do we need some synchronization?

zhangmeng916: If I understand correctly, is the concern regarding the race condition between here and the following part?
  public void onAcquiredLockNotification() {
    _lockStatus = LockConstants.LockStatus.LOCKED;
    _countDownLatch.countDown();
  }

It seems fine. Even the Lock status change and count down happen before await, the await will still exist when timeout, and the lock status is retrieved again after await.
I tried to leverage synchronized on both parts_lockstatus, but it doesn't work, cause it'll deadlock. Let me know if you still think it's not safe.

jiajunwang (approved):
LGTM! Please ensure the others' comments are addressed.

helix-lock/src/main/java/org/apache/helix/lock/helix/ZKDistributedNonblockingLock.java
dasahcc: Just make sure one thing, this is not shared status between two calls of same path. Then it should be safe for this case.

zhangmeng916: I see your concern. By current logic, it should be safe. If the previous lock was already acquired, subsequent tryLock from the same object will directly acquire the lock. If the previous lock was pending, subsequent tryLock won't check status as they're the same priority.

zhangmeng916:
This PR is ready to be merged. Approved by @jiajunwang 
Final commit message:
Implement priority for Helix distributed lock.
- Add priority definition and different kinds of timeout definition for Helix lock.
- Add notification support during high priority lock request preempting low priority lock.
- Add forceful lock concept when the cleanup work of low priority lock is not finished in the timeout.

PR #1568, opened 2020-12-01
Mark tasks as DROPPED if assigned participant does not exist

### LINKED ISSUE #1567: Job can stuck in STOPPING state when assigned instance goes offline
### Describe the bug
When the workflow target state is set to STOP, once the tasks are being canceled, if the assigned instance of the task goes offline, the job will be stuck in STOPPING state and will never go to the STOPPED state.

### To Reproduce
Have some delay in task cancellation. STOP the workflow and then stop the participants. In this case, the job will not reach to STOPPED state.

### Expected behavior
The controller should mark the tasks as DROPPED and mark the job as STOPPED.

alirezazamani:
### Issues

Fixes #1567 

### Description

In this PR, if the workflow or job is being stopped, but the assigned participant becomes offline for the task, the
task would be marked as DROPPED which allows the workflow and job to go to STOPPED state.

### Tests

TestStoppingWorkflowAndJob.testStopWorkflowInstanceOffline

Helix-core:
```
[INFO] Tests run: 1252, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,400.742 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 1252, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:30 h
[INFO] Finished at: 2020-11-30T16:09:06-08:00
```
The failed test is unrelated to this change and is failing even without this PR.

Helix-rest:
```
[INFO] Tests run: 171, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 263.332 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  04:27 min
[INFO] Finished at: 2020-11-30T16:14:32-08:00
```

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
```diff
@@ -317,6 +316,9 @@ private ResourceAssignment computeResourceMapping(String jobResource,
     // For delayed tasks, trigger a rebalance event for the closest upcoming ready time
     scheduleForNextTask(jobResource, jobCtx, currentTime);
 
+    // If the assigned instance is no longer live, so mark it as DROPPED in the context
+    markPartitionsWithoutLiveInstance(jobCtx, allPartitions, liveInstances);
```
dasahcc, line 320: This looks like an override of processing result... Can we have the logic in somewhere in scheduling task? This is more like a patching after scheduling computation. It waste the scheduling computation if instance is not live.

alirezazamani: We have this logic in the handleAdditionalTasks in filterTask method. Since that part of the code only hits when TargetState is equal to start, I moved the logic here so in all conditions we evaluate it. If the instance is not live, we still need to schedule the task if the task needs to be scheduled on other instances. If the instance is not live, then we do not see the task in current states which means we would no process the task in this case.

dasahcc: I understand the complexity of doing the check. So my question is when could this happening? We need to think about to put to the right place. Add here works, but it looks like a "saving action" without understand any assignment logic. I would prefer to understand the scenario and move the check logic for assignment there.

alirezazamani: It happens when workflow/job is being stopped and while the message is being sent and not processed yet (or the controller has not seen the effect yet) then the live instance is gone. In this case, the job will be stuck in the STOPPING state and the controller does not have a way to mark the task and job to be STOPPED. Note that all new assignment is happening when TargetState is START in handleAdditionTaskAssigment. If not, then assignment logic does not exist and will not be considered. We need to somehow manage this case before that. Note that when the live instance is gone, the controller does not see anything on the instance. I cannot find a better way to achieve this. If you know how it can be resolved, please let me know and I can change it accordingly. Also if you look at handleAdditionalTaskAssignment.filterTask method, it is the same code. I basically changed the place of this code and move it earlier in the pipeline.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -837,7 +835,7 @@ private static void addCompletedTasks(Set<Integer> set, JobContext ctx, Iterable
    * @param state
    * @return
    */
-  private boolean isTaskNotInTerminalState(TaskPartitionState state) {
+  protected static boolean isTaskNotInTerminalState(TaskPartitionState state) {
```
jiajunwang, line 838: You might want to move it to TaskPartitionState class.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
```diff
@@ -935,6 +933,20 @@ protected void markJobFailed(String jobName, JobContext jobContext, WorkflowConf
         currentTime);
   }
 
+  protected void markPartitionsWithoutLiveInstance(JobContext jobCtx, Set<Integer> allPartitions,
```
jiajunwang, line 936: It seems no one except JobDispatcher needs this method. Can we just move it to there?

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: TaskPartitionState is just an enum that lists all the states for the tasks. So I think it is better to keep it here like other similar methods.

helix-core/src/main/java/org/apache/helix/task/AbstractTaskDispatcher.java
alirezazamani: Sure. Moved to jobDispatcher.

helix-core/src/main/java/org/apache/helix/task/JobDispatcher.java
dasahcc: If this is the only scenario, my suggestion is moving the logic to following check in JobDispatcher. It checks whether we can determine a job is already stopped or not. It is also the right time that we check whether the instance assigned with tasks are gone or not. If it is gone, we can mark it as DROPPED.

  public static boolean checkJobStopped(JobContext jobContext) {
    for (int partition : jobContext.getPartitionSet()) {
      TaskPartitionState taskState = jobContext.getPartitionState(partition);
      if (taskState == TaskPartitionState.RUNNING) {
        return false;
      }
    }
    return true;
  }

alirezazamani: Moved to new place as suggessted.

dasahcc (approved):
LGTM

alirezazamani:

Final commit message:
Mark tasks as DROPPED if assigned participant is not live

In this commit, when the workflows or jobs are being stopped,
if the assigned participant becomes offline for the task, the
task would be marked as DROPPED which allows the workflows
and jobs to go to STOPPED state.

Issue #1578, opened 2020-12-08
The rebalance simulation tool may not generate accurate result


jiajunwang:
### Describe the bug
The rebalance tool relies on overriding the Data Provider caches to apply any user-specified simulation input. However, the current Data Provider set methods are not modifying any derived fields. This makes the override incomplete. So the simulated result is not accurate.

### To Reproduce
Run the simulation tool for a Helix cluster with disabled nodes. Meanwhile override the instance config so all the instances are enabled for the simulation.
However, the disabled nodes won't get any assignment due to the bug mentioned above.

### Expected behavior
All nodes are assigned with partitions since they are all enabled in the tool.

PR #1580, opened 2020-12-09
Update the internal fields of the Data Providr when overriding the data through set methods.

jiajunwang:
### Issues

#1578 

### Description

It was observed that the set methods in the Helix Data Provider classes, which are used but the Helix tools or tests, are not updating the derived fields.
This makes the override incomplete and it causes some potential issues when the tools are used.
This PR ensures that the set methods are modifying important derived fields.

### Tests

The main changed logic has been covered by the existing tests.
It is hard, and maybe not a good idea to cover all the possible util tool usages.

[ERROR] Failures:
[ERROR]   TestDisableCustomCodeRunner.test:236 expected:<false> but was:<true>
[INFO]
[ERROR] Tests run: 1252, Failures: 1, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD FAILURE
[INFO] Total time: 01:32 h
[INFO] Finished at: 2020-12-08T16:37:00-08:00

Rerun the failed test:

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 12.117 s - in org.apache.helix.integration.TestDisableCustomCodeRunner
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 18.469 s
[INFO] Finished at: 2020-12-08T17:36:03-08:00

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -338,10 +338,10 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     }
   }
 
-  private void updateIdealRuleMap() {
+  private void updateIdealRuleMap(ClusterConfig clusterConfig) {
     // Assumes cluster config is up-to-date
-    if (_clusterConfig != null) {
-      _idealStateRuleMap = _clusterConfig.getIdealStateRules();
+    if (clusterConfig != null) {
+      _idealStateRuleMap = clusterConfig.getIdealStateRules();
     } else {
       _idealStateRuleMap = new HashMap<>();
       LogUtil.logWarn(logger, getClusterEventId(), String
@@ -384,8 +384,8 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     _instanceMessagesCache.updateRelayMessages(_liveInstanceCache.getPropertyMap(),
         _currentStateCache.getParticipantStatesMap());
 
-    updateIdealRuleMap();
-    updateDisabledInstances();
+    updateIdealRuleMap(getClusterConfig());
+    updateDisabledInstances(getInstanceConfigMap().values(), getClusterConfig());
 
     return refreshedTypes;
   }
@@ -418,6 +418,8 @@ public class BaseControllerDataProvider implements ControlContextProvider {
   public void setClusterConfig(ClusterConfig clusterConfig) {
     _clusterConfig = clusterConfig;
     refreshAbnormalStateResolverMap(_clusterConfig);
+    updateIdealRuleMap(_clusterConfig);
+    updateDisabledInstances(getInstanceConfigMap().values(), _clusterConfig);
   }
 
   @Override
@@ -658,6 +660,7 @@ public class BaseControllerDataProvider implements ControlContextProvider {
    */
   public void setInstanceConfigMap(Map<String, InstanceConfig> instanceConfigMap) {
     _instanceConfigCache.setPropertyMap(instanceConfigMap);
+    updateDisabledInstances(instanceConfigMap.values(), getClusterConfig());
   }
 
   /**
@@ -744,30 +747,31 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     _updateInstanceOfflineTime = false;
   }
 
-  private void updateDisabledInstances() {
+  private void updateDisabledInstances(Collection<InstanceConfig> instanceConfigs,
+      ClusterConfig clusterConfig) {
     // Move the calculating disabled instances to refresh
     _disabledInstanceForPartitionMap.clear();
     _disabledInstanceSet.clear();
-    for (InstanceConfig config : _instanceConfigCache.getPropertyMap().values()) {
+    for (InstanceConfig config : instanceConfigs) {
       Map<String, List<String>> disabledPartitionMap = config.getDisabledPartitionsMap();
       if (!config.getInstanceEnabled()) {
         _disabledInstanceSet.add(config.getInstanceName());
       }
       for (String resource : disabledPartitionMap.keySet()) {
         if (!_disabledInstanceForPartitionMap.containsKey(resource)) {
-          _disabledInstanceForPartitionMap.put(resource, new HashMap<String, Set<String>>());
+          _disabledInstanceForPartitionMap.put(resource, new HashMap<>());
         }
         for (String partition : disabledPartitionMap.get(resource)) {
           if (!_disabledInstanceForPartitionMap.get(resource).containsKey(partition)) {
-            _disabledInstanceForPartitionMap.get(resource).put(partition, new HashSet<String>());
+            _disabledInstanceForPartitionMap.get(resource).put(partition, new HashSet<>());
           }
           _disabledInstanceForPartitionMap.get(resource).get(partition)
               .add(config.getInstanceName());
         }
       }
     }
-    if (_clusterConfig != null && _clusterConfig.getDisabledInstances() != null) {
-      _disabledInstanceSet.addAll(_clusterConfig.getDisabledInstances().keySet());
+    if (clusterConfig != null && clusterConfig.getDisabledInstances() != null) {
+      _disabledInstanceSet.addAll(clusterConfig.getDisabledInstances().keySet());
     }
   }
 


helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -384,8 +384,8 @@ public synchronized void refresh(HelixDataAccessor accessor) {
     _instanceMessagesCache.updateRelayMessages(_liveInstanceCache.getPropertyMap(),
         _currentStateCache.getParticipantStatesMap());
 
-    updateIdealRuleMap();
-    updateDisabledInstances();
+    updateIdealRuleMap(getClusterConfig());
```
kaisun2000, line 387: why here we use getClusterConfig() while later in `setClusterConfig` we use _clusterConfig?

In fact, 

```
  public ClusterConfig getClusterConfig() {
    return _clusterConfig;
  }
```

any specific reason?

kaisun2000: This seems to be a non-essential change?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -744,30 +747,31 @@ private void updateOfflineInstanceHistory(HelixDataAccessor accessor) {
...
+            _disabledInstanceForPartitionMap.get(resource).put(partition, new HashSet<>());
           }
           _disabledInstanceForPartitionMap.get(resource).get(partition)
               .add(config.getInstanceName());
         }
       }
     }
-    if (_clusterConfig != null && _clusterConfig.getDisabledInstances() != null) {
-      _disabledInstanceSet.addAll(_clusterConfig.getDisabledInstances().keySet());
+    if (clusterConfig != null && clusterConfig.getDisabledInstances() != null) {
```
kaisun2000, line 773: again, is `clusterConfig` passed in would always to the same as `_clusterConfig`?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -384,8 +384,8 @@ public synchronized void refresh(HelixDataAccessor accessor) {
     _instanceMessagesCache.updateRelayMessages(_liveInstanceCache.getPropertyMap(),
         _currentStateCache.getParticipantStatesMap());
 
-    updateIdealRuleMap();
-    updateDisabledInstances();
+    updateIdealRuleMap(getClusterConfig());
+    updateDisabledInstances(getInstanceConfigMap().values(), getClusterConfig());
```
kaisun2000, line 388: This seems to be non-essential too, right? The problem is in the` setClusterConfig` and `setInstanceConfig`

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
jiajunwang: Because setClusterConfig() is where you update _clusterConfig. And here, it is using it as an input.
It is a good practice that we don't refer to a private field directly in a private method. Note that the main reason no one calls this method to update for now is that we hide it too deep, so no one remembers. If the parameters are explicitly passed, meaning input and output are clearly divided, then we avoid potential bugs.
So, IMO this is tightly related to this change.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
jiajunwang: Please refer to the previous comment.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
jiajunwang: This it to improve the design so we don't make such a mistake again.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -418,6 +418,8 @@ public ClusterConfig getClusterConfig() {
   public void setClusterConfig(ClusterConfig clusterConfig) {
     _clusterConfig = clusterConfig;
     refreshAbnormalStateResolverMap(_clusterConfig);
+    updateIdealRuleMap(_clusterConfig);
+    updateDisabledInstances(getInstanceConfigMap().values(), _clusterConfig);
```
kaisun2000, line 422: Then why this place we don't use `getClusterConfig()`. 

Either way it is find to me. But please do use one approach.

kaisun2000:
One general comment, shall we add a simple test to make sure this is working as expected?

dasahcc (commented):
Change looks fine to me. Can you elaborate more we do this from external input? Why we cannot set cluster config and call these functions

kaisun2000:
One bigger picture question:

Think from controller caching layer (of zk)  point of view. The write of data is normally from tools or participant or even controller. But they do not use (should not use) these setClusterConfig etc methods of BaseControllerDataProvider.

Then who should use these set method? Note, these set method does not really persist the data to ZK. BaseControllerDataProvider is read only cache, right? 

In fact, we should mark these set method test only? Otherwise, when people use them, they would be surprised:

-  The data is not going to persist to next controller running cycle.

jiajunwang:
> Change looks fine to me. Can you elaborate more we do this from external input? Why we cannot set cluster config and call these functions

That's also a possible way. But the fact is that you cannot set and then trigger a partial refresh. So either you set to override one list, but the corresponding derived fields have old values. Or you call refresh and your set will be covered by the ZK data source.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
jiajunwang: I can change to "clusterConfig" instead of "_clusterConfig". But it makes very little difference.

getClusterConfig() is a different story. It is in set method, we should not call get method in a set method. It is just confusing.

Let's don't make it too complicated. My target is just to remove all the unnecessary private field references. And this specific method is a set method. It is necessary.

jiajunwang:
> One bigger picture question:
> 
> Think from controller caching layer (of zk) point of view. The write of data is normally from tools or participant or even controller. But they do not use (should not use) these setClusterConfig etc methods of BaseControllerDataProvider.
> 
> Then who should use these set method? Note, these set method does not really persist the data to ZK. BaseControllerDataProvider is read only cache, right?
> 
> In fact, we should mark these set method test only? Otherwise, when people use them, they would be surprised:
> 
> * The data is not going to persist to next controller running cycle.

I don't agree with you. The rebalancer tool is not test code. They are really used in production.
The best we can do is create a child class based on the basic cache class. But that is out of the scope of this PR. And it is not an essential change.

I suggest focusing on the issue fixing here. For the other topic, feel free to create other issues to track : )

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
jiajunwang: I think I might need to clarify it a little bit more. Since you seem to be confused by the style things...

I'm not sure what is the "approach" that you refer to... Calling the method or referring to the private fields are both OK. They are actually the same thing, right? Unless the method has some concurrent control or so.

I don't see any reason why we have to use getClusterConfig() or _clusterConfig or clusterConfig in this context. They make a minor difference. If I have to compare getClusterConfig() is the worst, since it triggers one more method call.

The difference is whether you pass the **INPUT** as the method parameters or directly refer to the private fields. This makes a difference. The latter introduces hidden dependencies between different private things. So this style is not good.
This is the small improvement that I'm trying to do here.

And obviously, this rule is not applicable in the set methods. Because they have to refer to the private fields, right?

kaisun2000:
> > One bigger picture question:
> > Think from controller caching layer (of zk) point of view. The write of data is normally from tools or participant or even controller. But they do not use (should not use) these setClusterConfig etc methods of BaseControllerDataProvider.
> > Then who should use these set method? Note, these set method does not really persist the data to ZK. BaseControllerDataProvider is read only cache, right?
> > In fact, we should mark these set method test only? Otherwise, when people use them, they would be surprised:
> > 
> > * The data is not going to persist to next controller running cycle.
> 
> I don't agree with you. The rebalancer tool is not test code. They are really used in production.
> The best we can do is create a child class based on the basic cache class. But that is out of the scope of this PR. And it is not an essential change.
> 
> I suggest focusing on the issue fixing here. For the other topic, feel free to create other issues to track : )

The the question is this:

Calling setClusterConfig method of BaseControllerDataProvider by tools would only change data in the cache. However, it is not going to change data in Zookeeper. Next controller run cycle, the data refreshed from BaseController would be old again.

 Is this something expected? I doubt.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
kaisun2000: Feel free to resolve this comment thread. The reason I ask is not for styling. It is that initially I not sure about the intention. I thought `getInstanceConfigMap()` had some tricks inside and that was why we did not use `_configMap`. If it is styling, no such intention, then either way is fine to me. Making them same style would also be good, but up to you.

jiajunwang:
> > > One bigger picture question:
> > > Think from controller caching layer (of zk) point of view. The write of data is normally from tools or participant or even controller. But they do not use (should not use) these setClusterConfig etc methods of BaseControllerDataProvider.
> > > Then who should use these set method? Note, these set method does not really persist the data to ZK. BaseControllerDataProvider is read only cache, right?
> > > In fact, we should mark these set method test only? Otherwise, when people use them, they would be surprised:
> > > 
> > > * The data is not going to persist to next controller running cycle.
> > 
> > 
> > I don't agree with you. The rebalancer tool is not test code. They are really used in production.
> > The best we can do is create a child class based on the basic cache class. But that is out of the scope of this PR. And it is not an essential change.
> > I suggest focusing on the issue fixing here. For the other topic, feel free to create other issues to track : )
> 
> The the question is this:
> 
> Calling setClusterConfig method of BaseControllerDataProvider by tools would only change data in the cache. However, it is not going to change data in Zookeeper. Next controller run cycle, the data refreshed from BaseController would be old again.
> 
> Is this something expected? I doubt.

As I mentioned this is out of the scope of this PR. And I already proposed a solution in the previous discussion. Do you have any other comments about the change itself? Or I suggest that we discuss future enhancement in another thread.
I think what you said is valid. But I'm not convinced that this is very important so we need to address in the near future. But if you don't agree, please feel free to create an issue for your concern.

dasahcc (approved):

jiajunwang:
Approved by @dasahcc, I will merge the change soon.

PR #1584, opened 2020-12-09
Participant-side Task Current State Migration

### LINKED ISSUE #1549: Task Current State Migration

NealSun96:
### Issues

Fixes #1549 (Partially)

### Description

To address excessive ZooKeeper reads caused by task framework CurrentState updates, it is proposed to move task framework CurrentStates to their separate path. 

The change is divided to multiple phases. This is the second phase, where the main focus of the changes are on the participant side. A system property flag is created to disable this feature when necessary. 

### Tests

TestTaskCurrentStatePathDisabled, testCurrentStatesRoutingTableIgnoreTaskCurrentStates

```
[INFO] Tests run: 1253, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,111.4 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:25 h
[INFO] Finished at: 2020-12-09T16:24:22-08:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
+++ b/helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
@@ -84,4 +84,6 @@ public class SystemPropertyKeys {
       MetadataStoreRoutingConstants.MSDS_SERVER_ENDPOINT_KEY;
 
   public static final String STATEUPDATEUTIL_ERROR_PERSISTENCY_ENABLED = "helix.StateUpdateUtil.errorLog.enabled";
+
+  public static final String TASK_CURRENT_PATH_DISABLED = "helix.taskCurrentPathDisabled";
 }
--- a/helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
+++ b/helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
@@ -85,5 +85,6 @@ public class SystemPropertyKeys {
 
   public static final String STATEUPDATEUTIL_ERROR_PERSISTENCY_ENABLED = "helix.StateUpdateUtil.errorLog.enabled";
 
-  public static final String TASK_CURRENT_PATH_DISABLED = "helix.taskCurrentPathDisabled";
+  public static final String TASK_CURRENT_STATE_PATH_DISABLED =
+      "helix.taskCurrentStatePathDisabled";
 }
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -45,6 +45,7 @@ import org.apache.helix.participant.statemachine.StateModel;
 import org.apache.helix.participant.statemachine.StateModelFactory;
 import org.apache.helix.participant.statemachine.StateModelParser;
 import org.apache.helix.participant.statemachine.StateTransitionError;
+import org.apache.helix.task.TaskStateModel;
 import org.apache.helix.util.StatusUpdateUtil;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.datamodel.ZNRecordBucketizer;
@@ -86,6 +87,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
   private final CurrentState _currentStateDelta;
   private final HelixManager _manager;
   private final StateModelFactory<? extends StateModel> _stateModelFactory;
+  private final boolean _isTaskMessage;
   volatile boolean _isTimeout = false;
 
   public HelixStateTransitionHandler(StateModelFactory<? extends StateModel> stateModelFactory,
@@ -98,6 +100,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
     _currentStateDelta = currentStateDelta;
     _manager = _notificationContext.getManager();
     _stateModelFactory = stateModelFactory;
+    _isTaskMessage = stateModel instanceof TaskStateModel;
   }
 
   void preHandleMessage() throws Exception {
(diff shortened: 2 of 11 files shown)


helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -264,8 +271,12 @@ void postHandleMessage() {
 
     try {
       // Update the ZK current state of the node
-      PropertyKey key = keyBuilder.currentState(instanceName, sessionId, resource,
-          bucketizer.getBucketName(partitionKey));
+      PropertyKey key =
+          _isTaskMessage && !Boolean.getBoolean(SystemPropertyKeys.TASK_CURRENT_STATE_PATH_DISABLED)
```
dasahcc, line 275: Let's have a variable to track TASK_CURRENT_STATE_PATH_DISABLED. We dont have to do the parse every time. Also, let's give a default value as true since participant will be the last component to deploy.

helix-core/src/main/java/org/apache/helix/task/TaskRunner.java
```diff
@@ -208,7 +209,10 @@ private static boolean setRequestedState(HelixDataAccessor accessor, String inst
         String.format("Requesting a state transition to %s for partition %s.", state, partition));
     try {
       PropertyKey.Builder keyBuilder = accessor.keyBuilder();
-      PropertyKey key = keyBuilder.currentState(instance, sessionId, resource);
+      PropertyKey key =
+          Boolean.getBoolean(SystemPropertyKeys.TASK_CURRENT_STATE_PATH_DISABLED) ? keyBuilder
```
dasahcc, line 213: Let's have the path build when TaskRunner instantiated.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
NealSun96: Default value should be false because it's "disabled" right? If participants get the new version of Helix, we should assume that they want to use the new path, so "disabled" should default to false. 

For variable tracking, there is a problem: if the boolean value is recorded at a central place, users cannot update this system property during run time, right?  If there is a variable that's tracking it, then it's as if it's cached.

helix-core/src/main/java/org/apache/helix/task/TaskRunner.java
NealSun96: For this I believe there's a similar problem: if users change this flag on the fly, the path has to be recreated. If we create the path when TaskRunner is instantiated, there's no way to recreate the path.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
dasahcc: Ah... Yeah, the "disable" should be false.

But for dynamic update, let's not overkill the design:
1. For Helix, we use ZK centralized config for dynamic update purpose. If we decide to use system property, it means it is not dynamic changeable.
2. The system property mostly comes from Java property, which require restart the application. So we cannot support dynamic update as well.

Make it a variable can also simplify the code. I would suggest create a variable to hold that when we create the object.

helix-core/src/main/java/org/apache/helix/task/TaskRunner.java
dasahcc: Same reason as the previous, dynamic update is overkill.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
NealSun96: Discussion ongoing.

NealSun96: Concluded to use one local variable when necessary.

helix-core/src/main/java/org/apache/helix/task/TaskRunner.java
NealSun96: Concluded to use one local variable when necessary.

helix-common/src/main/java/org/apache/helix/SystemPropertyKeys.java
```diff
@@ -84,4 +84,7 @@
       MetadataStoreRoutingConstants.MSDS_SERVER_ENDPOINT_KEY;
 
   public static final String STATEUPDATEUTIL_ERROR_PERSISTENCY_ENABLED = "helix.StateUpdateUtil.errorLog.enabled";
+
+  public static final String TASK_CURRENT_STATE_PATH_DISABLED =
```
zhangmeng916, line 88: With this config set up as "disabled", the default behavior would be to turn on task current state. Is this the expected behavior when customers roll out?

NealSun96: Yes. When customers roll out a new versions for *participants*, they should expect the behavior to change accordingly. This system property is for an emergency rollback.

dasahcc (approved):
Overall, LGTM. Please address others' comments.

alirezazamani (approved):
LGTM

NealSun96:
This PR is ready to be merged, approved by @dasahcc  @alirezazamani 
Final commit message:
## Participant-side Task Current State Migration ##
Second part of task current state migration. All changes made in this PR are on the participant side.

PR #1589, opened 2020-12-10
Reduce lock contention in CallbackHandler.enqueueTask

xyuanlu:
### Issues

#1590 - Lock contention in CallbackHandler.enqueueTask

### Description

During investigation and debugging, we found that a lot time is spent on waiting for the synchronized in CallbackHandler.enqueueTask.

```
public void enqueueTask(NotificationContext changeContext) throws Exception {
    // async mode only applicable to CALLBACK from ZK, During INIT and FINALIZE invoke the
    // callback's immediately.
    if (_batchModeEnabled && changeContext.getType() == NotificationContext.Type.CALLBACK) {
       .....
        synchronized (this) {                            <<<<<---------
          if (_batchCallbackProcessor != null) {
            _batchCallbackProcessor.queueEvent(changeContext.getType(), changeContext);
          } else {
      }
```

This PR removes this lock and uses AtomicReference instead.

### Tests

NA

```
[INFO] Tests run: 1253, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,189.07 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:26 h
[INFO] Finished at: 2020-12-09T19:10:21-08:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -650,6 +650,8 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
           callbackProcessorRef = new CallbackProcessor(this);
           if (_batchCallbackProcessor.compareAndSet(null, callbackProcessorRef)) {
             callbackProcessorRef.start();
+          } else {
+            callbackProcessorRef.shutdown();
           }
         }
     }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.helix.BaseDataAccessor;
 import org.apache.helix.HelixConstants.ChangeType;
@@ -112,7 +113,6 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
    * define the next possible notification types
    */
   private static Map<Type, List<Type>> nextNotificationType = new HashMap<>();
-
   static {
     nextNotificationType.put(Type.INIT, Arrays.asList(Type.CALLBACK, Type.FINALIZE));
     nextNotificationType.put(Type.CALLBACK, Arrays.asList(Type.CALLBACK, Type.FINALIZE));
@@ -134,8 +134,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
   private HelixCallbackMonitor _monitor;
 
   // TODO: make this be per _manager or per _listener instaed of per callbackHandler -- Lei
-  private CallbackProcessor _batchCallbackProcessor;
-  private Object _batchCallbackProcessorLock = new Object();
+  private AtomicReference<CallbackProcessor> _batchCallbackProcessor = new AtomicReference<>();;
   private boolean _watchChild = true; // Whether we should subscribe to the child znode's data
   // change.
 
@@ -321,10 +320,7 @@ public class CallbackHandler implements IZkChildListener, IZkDataListener {
         logger.info("CallbackHandler {} is not ready, ignore change callback from path: {}, for "
             + "listener: {}", _uid, _path, _listener);
       } else {
-        CallbackProcessor callbackProcessorRef;
-        synchronized (_batchCallbackProcessorLock) {
-          callbackProcessorRef = _batchCallbackProcessor;
-        }
+        CallbackProcessor callbackProcessorRef = _batchCallbackProcessor.get();
         if (callbackProcessorRef != null) {
           callbackProcessorRef.queueEvent(changeContext.getType(), changeContext);
         } else {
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -643,14 +643,17 @@ public void init() {
...
-          _batchCallbackProcessor.resetEventQueue();
+      CallbackProcessor callbackProcessorRef = _batchCallbackProcessor.get();
+        if (callbackProcessorRef != null) {
+          callbackProcessorRef.resetEventQueue();
         } else {
-          _batchCallbackProcessor = new CallbackProcessor(this);
-          _batchCallbackProcessor.start();
+          callbackProcessorRef = new CallbackProcessor(this);
+          if (_batchCallbackProcessor.compareAndSet(null, callbackProcessorRef)) {
+            callbackProcessorRef.start();
```
jiajunwang, line 652: Consider the following operation order:
1. init() and reset() is called at the same time.
2. In init(), the code reaches line 652, but not start yet.
3. At the same time, in reset(), code reaches line 776, and tries to shutdown the thread. But it cannot, since the thread has not been started yet. Note, in java doc, "Interrupting a thread that is not alive need not have any effect."
4. Then the code in init() try to start, the thread is started.

We have a leaked thread, eventually.

jiajunwang: I think you will need to start and then insert it into the atomic object.

But please revisit and see if there are any other ways.

xyuanlu: TFTR. I think you are correct. We should't touch the `callbackProcessorRef ` object after `compareAndSet` returns true.
Need to start and then insert

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -643,14 +643,17 @@ public void init() {
...
+          } else {
+            callbackProcessorRef.shutdown();
```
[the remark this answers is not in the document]
xyuanlu: I think `new CallbackProcessor(this)` creates a new thread and we need to close it if the new thread is not updated to `_batchCallbackProcessor`. However, since are doing start and then insert now, we need this `shutdown` anyway.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -766,16 +769,16 @@ void reset(boolean isShutdown) {
         isShutdown);
     try {
       _ready = false;
-      synchronized (this) {
-        if (_batchCallbackProcessor != null) {
+      CallbackProcessor callbackProcessorRef = _batchCallbackProcessor.get();
```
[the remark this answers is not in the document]
xyuanlu: Sorry I did not quite follow here. We still need to do nullptr check of the return value from getAndSet and also need to check isShutdown in the outer if. It does not seems to make this embedded if simpler.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -133,7 +134,7 @@
   private HelixCallbackMonitor _monitor;
 
   // TODO: make this be per _manager or per _listener instaed of per callbackHandler -- Lei
-  private CallbackProcessor _batchCallbackProcessor;
+  private AtomicReference<CallbackProcessor> _batchCallbackProcessorRef = new AtomicReference<>();;
```
jiajunwang, line 137: nit, extra ;

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -319,13 +320,12 @@ public void enqueueTask(NotificationContext changeContext) throws Exception {
...
             + "listener: {}", _uid, _path, _listener);
       } else {
-        synchronized (this) {
-          if (_batchCallbackProcessor != null) {
-            _batchCallbackProcessor.queueEvent(changeContext.getType(), changeContext);
-          } else {
-            throw new HelixException(
-                "Failed to process callback in batch mode. Batch Callback Processor does not exist.");
-          }
+        CallbackProcessor callbackProcessor = _batchCallbackProcessorRef.get();
```
jiajunwang, line 323: Not necessarily wrong behavior. But this may change the behavior.
The older logic will throw an exception if the event won't be processed. But with the new code, it is possible that we add the event to a closed processor without throwing exception?

xyuanlu: TFTR. Yes, it is a behavior change. However, I think in the old code when er get exception notice like this, we would treat it as a no-harm exception... So I think this change is ok.

jiajunwang (approved):
LGTM!

dasahcc:
Code looks good to me. But I have a general question. Fundamental, there is no difference for synchronize and atomic variable. Can this change solve the problem? Have we tested it for real use case?

xyuanlu:
> Code looks good to me. But I have a general question. Fundamental, there is no difference for synchronize and atomic variable. Can this change solve the problem? Have we tested it for real use case?

Yes, we have tested the change in our test env. The oendingCallback queue is almost zero and goes down as soon as number of callback request goes down. 
I think the major difference for this PR is it removed the lock on `this` keyword for _batchCallbackProcessor. The contention is not between different enqueue request, but between the enqueue and the `synchronize(this)` in function `invoke`, where `subscribeForChange` will take a lot time. 

I think the take away from this issue is that, we should always sync on as small object as possible.

xyuanlu:
This PR is ready to be merged in. Also verified the final change on test environment. 
approved by @jiajunwang 

Final commit message

Reduce lock contention in CallbackHandler.enqueueTask using atomic ref.

PR #1592, opened 2020-12-10
Cancel async callback when exception is thrown

### LINKED ISSUE #1591: ZkClient async callback waitForSuccess() is stuck if session mismatched
### Describe the bug
ZkClient `asyncCreate()` has a mechanism to retry the operation if the KeeperException code is retryable. But during the retry operation, if ZkSessionMismatchedException is thrown, the async callback is not being retried or cancelled, so the caller keeps waiting for the operation to complete. This issue blocks following pipeline running.

### To Reproduce
```
    try {
      zkClient.asyncCreate(path, record, CreateMode.PERSISTENT, createCallback);
      Assert.fail("Invalid session should not create znode");
    } catch (ZkSessionMismatchedException expected) {
      Assert.assertEquals(expected.getMessage(),
          "Failed to get expected zookeeper instance! There is a session id mismatch. Expected: "
              + "ExpectedSession. Actual: " + sessionId);
    }
    System.out.println("waiting");
    Assert.assertTrue(createCallback.waitForSuccess());
    System.out.println("after waiting");
[INFO] Running org.apache.helix.zookeeper.impl.client.TestRawZkClient
Starting ZK server at localhost:2127
waiting // stuck.........
```

### Expected behavior
Once an exception is thrown during the retry run, the callback context should be cancelled correctly.

pkuwm:
### Issues

Fixes #1591 

### Description

ZkClient asyncCreate() has a mechanism to retry the operation if the KeeperException code is retryable. The method `doRetry()` will push next retry to the retry queue if the error code is retryable such as connection loss or session expired.

But in the retry run, if `ZkSessionMismatchedException` is thrown from `getExpectedZookeeper()`, the actual async callback is neither being retried nor cancelled - the retry request is neither pushed to the retry queue nor canceled, then there is no chance the operation is finished. It is like a lock, the lock is not released, then the caller calls `wait()` will keep waiting forever.

Impact:
This issue blocks following pipeline running.

Expected behavior:
Once an exception is thrown during the retry run, the callback context should be cancelled correctly.

### Tests

- testAsyncWriteRetryThrowException
- testAsyncWriteByExpectedSession

In zookeeper-api

```
[INFO] Tests run: 51, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 64.415 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 51, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:08 min
[INFO] Finished at: 2020-12-10T15:34:19-08:00
```

Helix-core
```
[INFO] Tests run: 1253, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,328.498 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:28 h
[INFO] Finished at: 2020-12-11T17:53:11-08:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -1926,26 +1926,25 @@ public class ZkClient implements Watcher {
   private void doAsyncCreate(final String path, final byte[] data, final CreateMode mode,
       final long startT, final ZkAsyncCallbacks.CreateCallbackHandler cb,
       final String expectedSessionId) {
-    retryUntilConnected(() -> {
-      final ZooKeeper zk;
-      try {
-        zk = getExpectedZookeeper(expectedSessionId);
-      } catch (ZkSessionMismatchedException e) {
-        // Use BADARGUMENTS rc to represent session mismatch error (invalid expectedSession)
-        // and not retry.
-        cb.processResult(KeeperException.Code.BADARGUMENTS.intValue(), path,
-            new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
-        throw e;
-      }
-      zk.create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, mode, cb,
-          new ZkAsyncRetryCallContext(_asyncCallRetryThread, cb, _monitor, startT, 0, false) {
-            @Override
-            protected void doRetry() {
-              doAsyncCreate(path, data, mode, System.currentTimeMillis(), cb, expectedSessionId);
-            }
-          });
-      return null;
-    });
+    try {
+      retryUntilConnected(() -> {
+        getExpectedZookeeper(expectedSessionId)
+            .create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, mode, cb,
+                new ZkAsyncRetryCallContext(_asyncCallRetryThread, cb, _monitor, startT, 0, false) {
+                  @Override
+                  protected void doRetry() {
+                    doAsyncCreate(path, data, mode, System.currentTimeMillis(), cb,
+                        expectedSessionId);
+                  }
+                });
+        return null;
+      });
+    } catch (RuntimeException e) {
+      // Process callback to release caller from waiting
+      cb.processResult(ZkAsyncCallbacks.UNKNOWN_RET_CODE, path,
+          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+      throw e;
+    }
   }
 
   // Async Data Accessors
@@ -1965,28 +1964,25 @@ public class ZkClient implements Watcher {
 
   private void doAsyncSetData(final String path, byte[] data, final int version, final long startT,
       final ZkAsyncCallbacks.SetDataCallbackHandler cb, final String expectedSessionId) {
-    retryUntilConnected(() -> {
-      final ZooKeeper zk;
-      try {
-        zk = getExpectedZookeeper(expectedSessionId);
-      } catch (ZkSessionMismatchedException e) {
-        // Use BADARGUMENTS rc to represent session mismatch error (invalid expectedSession)
-        // and not retry.
-        cb.processResult(KeeperException.Code.BADARGUMENTS.intValue(), path,
-            new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
-        throw e;
-      }
-      zk.setData(path, data, version, cb,
-          new ZkAsyncRetryCallContext(_asyncCallRetryThread, cb, _monitor, startT,
-              data == null ? 0 : data.length, false) {
-            @Override
-            protected void doRetry() {
-              doAsyncSetData(path, data, version, System.currentTimeMillis(), cb,
-                  expectedSessionId);
-            }
-          });
-      return null;
-    });
+    try {
+      retryUntilConnected(() -> {
+        getExpectedZookeeper(expectedSessionId).setData(path, data, version, cb,
+            new ZkAsyncRetryCallContext(_asyncCallRetryThread, cb, _monitor, startT,
+                data == null ? 0 : data.length, false) {
+              @Override
+              protected void doRetry() {
+                doAsyncSetData(path, data, version, System.currentTimeMillis(), cb,
+                    expectedSessionId);
+              }
+            });
+        return null;
+      });
+    } catch (RuntimeException e) {
+      // Process callback to release caller from waiting
+      cb.processResult(ZkAsyncCallbacks.UNKNOWN_RET_CODE, path,
+          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+      throw e;
+    }
   }
 
   public void asyncGetData(final String path, final ZkAsyncCallbacks.GetDataCallbackHandler cb) {
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -1927,15 +1927,23 @@ public class ZkClient implements Watcher {
       final long startT, final ZkAsyncCallbacks.CreateCallbackHandler cb,
       final String expectedSessionId) {
     retryUntilConnected(() -> {
-      getExpectedZookeeper(expectedSessionId)
-          .create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, mode, cb,
-              new ZkAsyncRetryCallContext(_asyncCallRetryThread, cb, _monitor, startT, 0, false) {
-                @Override
-                protected void doRetry() {
-                  doAsyncCreate(path, data, mode, System.currentTimeMillis(), cb,
-                      expectedSessionId);
-                }
-              });
+      final ZooKeeper zk;
+      try {
+        zk = getExpectedZookeeper(expectedSessionId);
+      } catch (ZkSessionMismatchedException e) {
+        // Use BADARGUMENTS rc to represent session mismatch error (invalid expectedSession)
+        // and not retry.
+        cb.processResult(KeeperException.Code.BADARGUMENTS.intValue(), path,
+            new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+        throw e;
+      }
+      zk.create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, mode, cb,
+          new ZkAsyncRetryCallContext(_asyncCallRetryThread, cb, _monitor, startT, 0, false) {
+            @Override
+            protected void doRetry() {
+              doAsyncCreate(path, data, mode, System.currentTimeMillis(), cb, expectedSessionId);
+            }
+          });
       return null;
     });
   }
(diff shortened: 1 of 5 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncRetryThread.java
```diff
@@ -49,7 +49,9 @@ public void run() {
           context.cancel();
           interrupt();
         } catch (Throwable e) {
-          LOG.error("Error retrying callback " + context, e);
+          LOG.error("Error retrying callback {}, cancelling it", context, e);
+          // Cancel the context so the upstream caller can stop waiting
+          context.cancel();
```
dasahcc, line 54: Let's cancel in final statement?

dasahcc (commented):
Could you elaborate more in description why the retry is waiting? Is that because of the recursive call? This is the part  I am confusing, since the get ExpectedZookeeper will throw exception and not caught. So the recursive call will not continue.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncRetryThread.java
jiajunwang: No, if retry was done successfully, then cancel it prematurely will block the normal logic.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1927,15 +1927,23 @@ private void doAsyncCreate(final String path, final byte[] data, final CreateMod
...
+        // and not retry.
+        cb.processResult(KeeperException.Code.BADARGUMENTS.intValue(), path,
+            new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+        throw e;
+      }
+      zk.create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, mode, cb,
+          new ZkAsyncRetryCallContext(_asyncCallRetryThread, cb, _monitor, startT, 0, false) {
+            @Override
+            protected void doRetry() {
+              doAsyncCreate(path, data, mode, System.currentTimeMillis(), cb, expectedSessionId);
```
jiajunwang, line 1944: Why we need these given we have already cancel it in the doRetry()?

pkuwm: It is to handle the first call before it enters doRetry() and asyncThread. For the single thread calling, it is OK as the exception is thrown to the upstream caller and the caller won't be able to call waitForSuccess().
If it is multi threading: thread-1 calls asyncCreate(), and thread-2 waitForSuccess(), thread-2 will be waiting forever.
So I'd like to process it as when the sessions mismatch, we should treat the operation a failure.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncRetryThread.java
pkuwm: As @jiajunwang said, cancel it prematurely in the finally statement will impact the future retries: if this retry fails because of connection loss, it'll continue retrying. If we cancel it in the finally, the operation is marked as done so it won't retry the operation. We only want to cancel it if a non-retryable exception throws.

pkuwm:
> Could you elaborate more in description why the retry is waiting? Is that because of the recursive call? This is the part I am confusing, since the get ExpectedZookeeper will throw exception and not caught. So the recursive call will not continue.

I think you got it. If getExpectedZookeeper throws an exception, the recursive retry will not continue or push next retry request to queue. Then the final step of marking the operation done is not executed. In this case, the caller will `wait()` forever. To resolve it, we need to cancel it so `wait()` can be released.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
jiajunwang: It's a good catch!. But I think you should catch-all exceptions, not only session mismatch. There is no other place that we can cancel the callback.

pkuwm: @jiajunwang I seriously thought about catching a generic exception, but after digging into the code more, `getExpectedZookeeper()` is almost impossible to throw an exception other than `ZkSessionMismatchedException`. That's the reason I think catching `ZkSessionMismatchedException` is good enough and I just try...catch `getExpectedZookeeper` other than the whole block below. It's cleaner to understand we just want to handle exception for `getExpectedZookeeper`, excluding `create()`. The KeeperException error from create() is handled in the async retry thread.
```
try {
  getExpectedZookeeper(expectedSessionId).create();
} catch (Exception e) {
}
```
Maybe you could say catching a generic one is safer.

jiajunwang: I know the logic is absolutely correct. Just like when we first implemented this retry logic and there is no getExpectedZookeeper() call so nothing will go wrong.

The point is that we may have other exceptions if the new code is added here. Then this issue might happen if any user is using a different thread to wait.

Back to my original question, what is the problem of a generic try-catch? To be clear, what I mean is,

try {
  retryUntilConnect(....)
} catch (Exception ex) {
  handle callback
}

IMO, this looks cleaner : )

pkuwm: @jiajunwang OK got you. I don't think there is a function problem. I just could not find a good KeeperException code for handling callback. Previously I thought of a similar code to catch the whole private API, which might be a bit cleaner? - same thing, wrapping inside the API vs outside.
```
public void asyncCreate() {
  try {
    doAsyncCreate();
  } catch (RuntimeException e) {
   // I think we'll need a generic code here but I am not sure what error code is a good one
    cb.processResult(KeeperException.Code.BADARGUMENTS.intValue());
    throw e;
  }
```
I can't find a generic error code for the generic runtimeException: `KeeperException.Code.BADARGUMENTS` cannot represent a generic RuntimeException. Maybe a customized one like -255? 
The reason I put catch for getExpectedZookeeper is we may use BADARGUMENTS to represent session mismatch.

Let me know your opinions, including what error code number is preferred.

jiajunwang: In this case, I think UNKNOWN_RET_CODE is fine. Because you really don't have it, right?

pkuwm: Refactored. @jiajunwang Could you take another look? Thanks.

dasahcc (commented):
Are we sure this is the reason that cause pipeline stop? Did you see the error log that cause the pipeline hanging? Because after a certain time, we do not see any of these exceptions.

pkuwm:
> Are we sure this is the reason that cause pipeline stop? Did you see the error log that cause the pipeline hanging? Because after a certain time, we do not see any of these exceptions.

@dasahcc We do believe it is the root cause. Below is the stack trace right before the controller was restarted. By combining the previous session mismatch stack trace with this, it makes sense the message async creation operation is kept waiting by `waitForSuccess()`.  Session mismatched exception is thrown and interrupt the retrying. The retry run catches the exception but does not cancel the callback or send it for next retry - the async callback operation is never marked as done but not ever tried again after the exception. 

Since the pipeline handling is stuck at `waitForSuccess()`, it cannot process future events or send any messages. That's why you also did not see any task messages written.

```
org.apache.helix.zookeeper.zkclient.exception.ZkSessionMismatchedException: <LOG MESSAGE REDACTED DUE TO GDPR REQUIREMENTS>
	at org.apache.helix.zookeeper.zkclient.ZkClient.getExpectedZookeeper(ZkClient.java:2373)
	at org.apache.helix.zookeeper.zkclient.ZkClient.lambda$doAsyncCreate$3(ZkClient.java:1930)
	at org.apache.helix.zookeeper.zkclient.ZkClient.retryUntilConnected(ZkClient.java:1655)
	at org.apache.helix.zookeeper.zkclient.ZkClient.doAsyncCreate(ZkClient.java:1929)
	at org.apache.helix.zookeeper.zkclient.ZkClient.access$900(ZkClient.java:85)
	at org.apache.helix.zookeeper.zkclient.ZkClient$13.doRetry(ZkClient.java:1935)
	at org.apache.helix.zookeeper.zkclient.callback.ZkAsyncRetryThread.run(ZkAsyncRetryThread.java:46)
```

```
2020/12/10 01:54:30.867 INFO [GenericHelixController] [HelixController-pipeline-default-cluster-(79b55ff9_DEFAULT)] [helix] [] END ClusterEventProcessor thread default-cluster
2020/12/10 01:54:30.867 INFO [GenericHelixController] [HelixTaskExecutor-message_handle_thread_5] [helix] [] Shutting down TASK pipeline
2020/12/10 01:54:30.867 ERROR [ZkAsyncCallbacks] [HelixController-pipeline-task-cluster-(99dc3f03_TASK)] [helix] [] Interrupted waiting for success
java.lang.InterruptedException: null
    at java.lang.Object.wait(Native Method) ~[?:1.8.0_172]
    at java.lang.Object.wait(Object.java:502) ~[?:1.8.0_172]
    at org.apache.helix.zookeeper.zkclient.callback.ZkAsyncCallbacks$DefaultCallback.waitForSuccess(ZkAsyncCallbacks.java:220) [helix-core-1.0.1.2200.jar:1.0.1.2200]
    at org.apache.helix.manager.zk.ZkBaseDataAccessor.create(ZkBaseDataAccessor.java:761) [helix-core-1.0.1.2200.jar:1.0.1.2200]
    at org.apache.helix.manager.zk.ZkBaseDataAccessor.createChildren(ZkBaseDataAccessor.java:831) [helix-core-1.0.1.2200.jar:1.0.1.2200]
    at org.apache.helix.manager.zk.ZKHelixDataAccessor.createChildren(ZKHelixDataAccessor.java:516) [helix-core-1.0.1.2200.jar:1.0.1.2200]
    at org.apache.helix.controller.stages.MessageDispatchStage.sendMessages(MessageDispatchStage.java:187) [helix-core-1.0.1.2200.jar:1.0.1.2200]
    at org.apache.helix.controller.stages.MessageDispatchStage.processEvent(MessageDispatchStage.java:96) [helix-core-1.0.1.2200.jar:1.0.1.2200]
    at org.apache.helix.controller.stages.task.TaskMessageDispatchStage.process(TaskMessageDispatchStage.java:33) [helix-core-1.0.1.2200.jar:1.0.1.2200]
    at org.apache.helix.controller.pipeline.Pipeline.handle(Pipeline.java:68) [helix-core-1.0.1.2200.jar:1.0.1.2200]
    at org.apache.helix.controller.GenericHelixController.handleEvent(GenericHelixController.java:802) [helix-core-1.0.1.2200.jar:1.0.1.2200]
    at org.apache.helix.controller.GenericHelixController.access$500(GenericHelixController.java:129) [helix-core-1.0.1.2200.jar:1.0.1.2200]
    at org.apache.helix.controller.GenericHelixController$ClusterEventProcessor.run(GenericHelixController.java:1442) [helix-core-1.0.1.2200.jar:1.0.1.2200]
```

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1926,18 +1926,25 @@ public void asyncCreate(final String path, Object datat, final CreateMode mode,
...
+                new ZkAsyncRetryCallContext(_asyncCallRetryThread, cb, _monitor, startT, 0, false) {
+                  @Override
+                  protected void doRetry() {
+                    doAsyncCreate(path, data, mode, System.currentTimeMillis(), cb,
+                        expectedSessionId);
+                  }
+                });
+        return null;
+      });
+    } catch (RuntimeException e) {
```
jiajunwang, line 1942: nit, RuntimeException  -> Throwable?

jiajunwang (approved):
LGTM! only one minor comment.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
pkuwm: I thought about it but I think RuntimeException is good enough. If the RetryUntilConnected is changed to throw a checked exception, all the APIs would have to change and throw the checked exception. I just don't think the behavior would change. Let's keep the RuntimeException as it indicates we know what we are doing.

alirezazamani (approved):

pkuwm:
This PR is ready to be merged, approved by @jiajunwang @alirezazamani 

Problem:
In the ZkClient async retry run, if ZkSessionMismatchedException is thrown from getExpectedZookeeper(), the actual async callback is neither being retried nor cancelled, causing the callback waitForSuccess() to hang.

Solution:
Once an exception is thrown during the retry run, the callback context should be processed or cancelled correctly.

PR #1619, opened 2021-01-21
Use WAGED rebalancer as the default rebalancer for the controller resources in the super cluster.

### LINKED ISSUE #1618: Support WAGED rebalancer as the default rebalance configuration for the super cluster controller resources
We are using a hardcoded rebalance setup for the supercluster controller resources. This means whenever a new cluster is activated to join the supercluster, it will be FULL_AUTO, Delayed Rebalanced, and using CrushEd.
However, the fact is that we might want to use WAGED or another strategy. Since it would be sub-optimal to adjust the IdealState after the cluster is already activated, it would be more efficient and safer to enable creating the resource directly as it is supposed to be.

I proposed to add a supercluster cluster-level configuration that setups the default rebalance config. When admins activate the cluster using the activateCluster API in HelixAdmin, the API implementation will read this config and apply the setup.

[update] With recent discussion, we think enforcing WAGED rebalancer as the default rebalancer would be easier and more direct.

**Additional context**

jiajunwang:
### Issues

Resolves #1618 

### Description

This PR changes the default rebalancer for the controller resources in the supercluster to be the WAGED rebalancer.

### Tests

TestAddClusterV2 has been modified accordingly.

[INFO] Tests run: 1259, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,544.234 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 1259, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -1266,11 +1266,10 @@ public class ZKHelixAdmin implements HelixAdmin {
     idealState.setNumPartitions(1);
     idealState.setStateModelDefRef("LeaderStandby");
     idealState.setRebalanceMode(RebalanceMode.FULL_AUTO);
-    idealState.setRebalancerClassName(DelayedAutoRebalancer.class.getName());
-    idealState.setRebalanceStrategy(CrushEdRebalanceStrategy.class.getName());
+    idealState.setRebalancerClassName(WagedRebalancer.class.getName());
     // TODO: Give user an option, say from RestAPI to config the number of replicas.
     idealState.setReplicas(Integer.toString(DEFAULT_SUPERCLUSTER_REPLICA));
-    idealState.getRecord().setListField(clusterName, new ArrayList<String>());
+    idealState.getRecord().setListField(clusterName, new ArrayList<>());
 
     List<String> controllers = getInstancesInCluster(grandCluster);
     if (controllers.size() == 0) {
@@ -1278,10 +1277,12 @@ public class ZKHelixAdmin implements HelixAdmin {
     }
 
     ZKHelixDataAccessor accessor =
-        new ZKHelixDataAccessor(grandCluster, new ZkBaseDataAccessor<ZNRecord>(_zkClient));
+        new ZKHelixDataAccessor(grandCluster, new ZkBaseDataAccessor<>(_zkClient));
     PropertyKey.Builder keyBuilder = accessor.keyBuilder();
 
     accessor.setProperty(keyBuilder.idealStates(idealState.getResourceName()), idealState);
+    LOG.info("Cluster {} has been added to grand cluster {} with rebalance configuration {}.",
+        clusterName, grandCluster, idealState.getRecord().getSimpleFields().toString());
   }
 
   @Override
--- a/helix-core/src/test/java/org/apache/helix/integration/TestAddClusterV2.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/TestAddClusterV2.java
@@ -21,10 +21,13 @@ package org.apache.helix.integration;
 
 import java.util.Date;
 
+import org.apache.helix.HelixAdmin;
 import org.apache.helix.TestHelper;
 import org.apache.helix.common.ZkTestBase;
+import org.apache.helix.controller.rebalancer.waged.WagedRebalancer;
 import org.apache.helix.integration.manager.ClusterDistributedController;
 import org.apache.helix.integration.manager.MockParticipantManager;
+import org.apache.helix.model.IdealState;
 import org.apache.helix.tools.ClusterSetup;
 import org.apache.helix.tools.ClusterVerifiers.BestPossibleExternalViewVerifier;
 import org.apache.helix.tools.ClusterVerifiers.ZkHelixClusterVerifier;
@@ -90,7 +93,13 @@ public class TestAddClusterV2 extends ZkTestBase {
 
   @Test
   public void Test() {
-
+    // Verify the super cluster resources are all rebalanced by the WAGED rebalancer.
+    HelixAdmin admin = _gSetupTool.getClusterManagementTool();
+    for (String clusterName : admin.getResourcesInCluster(CONTROLLER_CLUSTER)) {
+      IdealState is = _gSetupTool.getClusterManagementTool()
+          .getResourceIdealState(CONTROLLER_CLUSTER, clusterName);
+      Assert.assertEquals(is.getRebalancerClassName(), WagedRebalancer.class.getName());
+    }
   }
 
   @AfterClass
--- a/helix-core/src/test/java/org/apache/helix/tools/TestHelixAdminCli.java
+++ b/helix-core/src/test/java/org/apache/helix/tools/TestHelixAdminCli.java
@@ -27,10 +27,9 @@ import java.util.HashSet;
 import java.util.Set;
 
 import org.apache.helix.BaseDataAccessor;
+import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.TestHelper;
-import org.apache.helix.tools.ClusterVerifiers.BestPossibleExternalViewVerifier;
-import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.common.ZkTestBase;
 import org.apache.helix.integration.manager.ClusterDistributedController;
 import org.apache.helix.integration.manager.MockParticipantManager;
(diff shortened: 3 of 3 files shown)


helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -1018,6 +1032,22 @@ public void setAbnormalStateResolverMap(Map<String, String> resolverMap) {
...
+    _record.setMapField(ClusterConfigProperty.SUPER_CLUSTER_REBALANCE_CONFIG.name(),
+        rebalanceConfig.getConfigsMap());
+  }
+
+  public RebalanceConfig getSuperClusterRebalanceConfig() {
+    Map<String, String> rebalanceConfigMap =
+        _record.getMapField(ClusterConfigProperty.SUPER_CLUSTER_REBALANCE_CONFIG.name());
+    if (rebalanceConfigMap != null) {
+      return new RebalanceConfig(rebalanceConfigMap);
+    } else {
```
NealSun96, line 1045: Unnecessary else block

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -167,6 +173,14 @@
   private static final int GLOBAL_TARGET_TASK_THREAD_POOL_SIZE_NOT_SET = -1;
   private static final int OFFLINE_NODE_TIME_OUT_FOR_MAINTENANCE_MODE_NOT_SET = -1;
 
+  private static final Map<String, String> DEFAULT_SUPER_CLUSTER_REBALANCE_CONFIG = ImmutableMap
+      .of(RebalanceConfig.RebalanceConfigProperty.REBALANCE_MODE.name(),
+          RebalanceConfig.RebalanceMode.FULL_AUTO.name(),
+          RebalanceConfig.RebalanceConfigProperty.REBALANCER_CLASS_NAME.name(),
+          DelayedAutoRebalancer.class.getName(),
```
NealSun96, line 180: I thought it's WAGED by default?

NealSun96 (changes requested):

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: We cannot change the existing default behavior. So this PR is to add the configuration item to make WAGED default.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: It expresses the code logic better, IMO.

I can find many arguments online about this style, but I don't see much difference in this condition. If the previous return is for the exception or some early termination logic, then I think we should not use "else". But here, I don't see much difference. And my mind follows the current code better : )

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -1261,27 +1260,46 @@ public void addClusterToGrandCluster(String clusterName, String grandCluster) {
...
-    IdealState idealState = new IdealState(clusterName);
+    ZKHelixDataAccessor accessor =
+        new ZKHelixDataAccessor(grandCluster, new ZkBaseDataAccessor<>(_zkClient));
+    PropertyKey.Builder keyBuilder = accessor.keyBuilder();
+
+    ClusterConfig grandClusterConfig = accessor.getProperty(keyBuilder.clusterConfig());
+    RebalanceConfig superClusterRebalanceConfig =
+        grandClusterConfig.getSuperClusterRebalanceConfig();
+
+    // Note that the other rebalance configurations that are not verified below will be ignored.
```
zhangmeng916, line 1271: I'm confused about this comment here. Does the following check validate all possible scenarios or not? Do you mean "other configuration" as valid or invalid?

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
```diff
@@ -393,4 +398,32 @@ private void trySetInvalidAbnormalStatesResolverMap(ClusterConfig testConfig,
...
+    Assert.assertEquals(testConfig.getSuperClusterRebalanceConfig().getRebalanceStrategy(),
+        CrushEdRebalanceStrategy.class.getName());
+    Assert.assertEquals(testConfig.getSuperClusterRebalanceConfig().getRebalanceClassName(),
+        DelayedAutoRebalancer.class.getName());
+    Assert.assertEquals(testConfig.getSuperClusterRebalanceConfig().getRebalanceMode(),
+        RebalanceConfig.RebalanceMode.FULL_AUTO);
+
+    // Check set and read
+    RebalanceConfig rebalanceConfig = testConfig.getSuperClusterRebalanceConfig();
+    rebalanceConfig.setRebalanceClassName(WagedRebalancer.class.getName());
```
zhangmeng916, line 416: Do we plan to also support REST API for this change? Feel that'll be very useful.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
jiajunwang: It is not. Since we reused the existing structure RebalanceConfig, there are several fields that are not applicable.
That is also why I put the check in the API instead of the config class.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
jiajunwang: Rest API already supports modify cluster config. There is no need to add a specific method just for supercluster configuration, IMO. It is debatable for sure. But either way, we can do this PR first.

zhangmeng916: I think REST API through cluster config is good enough. Maybe you can add one test case in `TestClusterAccessor`, similar as `testPurgeOfflineParticipants` so that your PR is complete.

jiajunwang: I doubt that it will increase any meaningful test coverage. 1. the REST API support is not the focus of this change. 2. There is no change in the REST, I think the existing tests cover the logic well.

To clarify, my point is that I don't have a good case that will pass the existing HelixAdmin tests but fail a REST test (because of a relevant change in this PR). So I think another REST test is just adding a boring layer to the already covered logic.
But if you do have such a case, please let me know and I will be glad to add the test.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
dasahcc: +1 to @NealSun96 . I would suggest to follow the coding convention.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -134,7 +137,10 @@
     // offline for more than this specified time period, and users call purge participant API,
     // then the node will be removed.
     // The unit is milliseconds.
-    OFFLINE_DURATION_FOR_PURGE_MS
+    OFFLINE_DURATION_FOR_PURGE_MS,
+
+    // Resource rebalance configuration for the super cluster which manages Helix controller resources.
+    SUPER_CLUSTER_REBALANCE_CONFIG
```
dasahcc, line 143: This could be confusing. Because other sub clusters can have this entry. What does it mean to the sub cluster? If this is only designed for super cluster. Would this be meaningful in ClusterConfig?

I have concerns to add it as specific for super cluster IS rebalance strategy. We should make it generic. Either we add it as default rebalance strategy for existing cluster. Or let's make the the configurable passing down from the admin API every time.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: Which code convention are we talking about? I think there is no such rule... But please share a link so I can read it.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: My argument is that a sub-cluster can be a supercluster at the same time. It is possible to do a cascading setup. So for a sub-cluster, the additional setup still means a supercluster configuration, which will be used when it acts as a supercluster.

About the first suggestion, are you talking about use WAGED as the default rebalancer for the sub-cluster resource directly? That was the original design. I just have a concern about backward compatibility. But if you think it is fine, I'm totally fine with this simpler solution.

Regarding the other option, I thought about modifying the activeCluster API to accept the optional parameters. But this means we will need to modify helix-rest and helix-front as well. And there is a backward compatibility issue. I would rather deprecate this API and restrict it to only use the general resource creation APIs for activating a sub-cluster.

jiajunwang: As discussed with @dasahcc offline, we will just modify the default rebalance option for the sub-clusters. This will simplify this PR quite a lot. 👍

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: @NealSun96 as discussed with Junkai, we decided to make WAGED the default option : )

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: This line is not necessary with the latest code change.

dasahcc (approved):
LGTM

dasahcc (approved):

jiajunwang:
Approved by @dasahcc , I will merge the change.

PR #1620, opened 2021-01-22
Per Replica Throttle -- 1st: Skeleton implementation with output message same as input

kaisun2000:
### Issues

enhance #343; first part skeleton.

### Description

First part of Per Replica throttling, skeleton.

Per replica throttling replacing intermediate stage which is partition based. The finer granularity
would skip boosting unnecessary replica in a recovery partition.

### Tests

### Documentation (Optional)

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+import org.apache.helix.monitoring.mbeans.ResourceMonitor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class PerReplicaThrottleStage extends AbstractBaseStage {
+  private static final Logger logger =
+      LoggerFactory.getLogger(PerReplicaThrottleStage.class.getName());
+
+  public PerReplicaThrottleStage() {
```
jiajunwang, line 29: What is this? I think the default constructor is not necessary in a normal case.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    ResourcesStateMap retracedResourceStateMap = new ResourcesStateMap();
+    List<Message> throttledRecoveryMsg = new ArrayList<>();
+    List<Message> throttledLoadMsg = new ArrayList<>();
+    MessageOutput output =
+        compute(event, resourceToRebalance, currentStateOutput, selectedMessages, retracedResourceStateMap, throttledRecoveryMsg, throttledLoadMsg);
+
+    event.addAttribute(AttributeName.PER_REPLICA_OUTPUT_MESSAGES.name(), output);
+    LogUtil.logDebug(logger,_eventId, String.format("retraceResourceStateMap is: %s", retracedResourceStateMap));
```
jiajunwang, line 60: If it is going to be output as the "intermediate states", then we do have a way to configure persisting this value, right? In this case, we can debug with the IdealState map field. So the debug log is not necessary here.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+    MessageOutput output = new MessageOutput();
+
+    ResourceControllerDataProvider dataCache =
+        event.getAttribute(AttributeName.ControllerDataProvider.name());
+
+    StateTransitionThrottleController throttleController =
+        new StateTransitionThrottleController(resourceMap.keySet(), dataCache.getClusterConfig(),
+            dataCache.getLiveInstances().keySet());
+
+    // Resource level prioritization based on the numerical (sortable) priority field.
```
jiajunwang, line 90: Please abstract these re-ordering logic to a private method. The output of the method would be prioritizedResourceList.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+            && dataCache.getIdealState(resourceName).getRecord().getSimpleField(priorityField)
+            != null) {
+          resourcePriority.setPriority(
+              dataCache.getIdealState(resourceName).getRecord().getSimpleField(priorityField));
+        }
+      }
+      prioritizedResourceList.sort(new ResourcePriorityComparator());
+    }
+
+    ClusterStatusMonitor clusterStatusMonitor =
```
jiajunwang, line 121: I suggest splitting the monitor functions to another PR. It should be added together with the corresponding unit tests.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+        LogUtil.logInfo(logger, _eventId,
+            "Failed to calculate per replica partition states for resource " + resourceName, ex);
+        failedResources.add(resourceName);
+      }
+    }
+
+    if (clusterStatusMonitor != null) {
+      clusterStatusMonitor.setResourceRebalanceStates(failedResources,
+          ResourceMonitor.RebalanceStatus.PER_REPLICA_STATE_CAL_FAILED);
+      clusterStatusMonitor.setResourceRebalanceStates(output.resourceSet(),
```
jiajunwang, line 167: Shouldn't it be prioritizedResourceList, instead of output.resourceSet()?

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
```diff
@@ -36,6 +37,10 @@ public MessageOutput() {
 
   }
 
+  public Set<String> resourceSet() {
```
jiajunwang, line 40: As commented below, I think this method is not necessary.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
```diff
@@ -64,6 +69,14 @@ public void addMessages(String resourceName, Partition partition,
     return Collections.emptyList();
   }
 
+  public Map<Partition, List<Message>> getResourceMessages(String resourceName) {
```
jiajunwang, line 72: This method could be costly since it needs to read and create an unmodifiable view for all the messages for a resource. I think getMessages with resource name and partition shall be working fine. Can we use that instead?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+      Resource resource, BestPossibleStateOutput bestPossibleStateOutput,
+      ResourceControllerDataProvider cache, StateTransitionThrottleController throttleController,
+      Map<Partition, Map<String, String>> retracedPartitionsStateMap,
+      List<Message> throttledRecoveryMsgOut, List<Message> throttledLoadMessageOut,
+      MessageOutput output) {
+    String resourceName = resource.getResourceName();
+    LogUtil.logInfo(logger, _eventId, String.format("Processing resource: %s", resourceName));
+
+    if (!throttleController.isThrottleEnabled() || !IdealState.RebalanceMode.FULL_AUTO
+        .equals(idealState.getRebalanceMode())) {
```
jiajunwang, line 194: Let's add a TODO here since we discussed enabling throttling for the other rebalance modes as well. For this PR, I agree that we keep the current behavior. But since this will be treated as a new class technically, the condition here is not quite self-explaining. So some comments like TODO would make sense.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+    return output;
+  }
+
+  /*
+   * Apply per-replica throttling logic and filter out excessive recovery and load messages for a
+   * given resource.
+   * Reconstruct retrace partition states for a resource based on pending and targeted messages
+   * Return messages for partitions of a resource.
+   * Out param retracedPartitionsCurrentState
+   * Out param output
```
jiajunwang, line 180: Can you use the formal java doc format?
@param output .....

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+
+    if (!throttleController.isThrottleEnabled() || !IdealState.RebalanceMode.FULL_AUTO
+        .equals(idealState.getRebalanceMode())) {
+      retracedPartitionsStateMap.putAll(bestPossibleStateOutput.getPartitionStateMap(resourceName).getStateMap());
+      for (Partition partition : selectedResourceMessages.keySet()) {
+        output.addMessages(resourceName, partition, selectedResourceMessages.get(partition));
+      }
+      return;
+    }
+    // TODO:
```
jiajunwang, line 201: So this PR will throttle all the messages, right? How are we going to unit test it?
I suggest that we let all messages pass first. Then the test can simply check if the output contains all the messages and if the retraced state is expected.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+      try {
+        _priority = Integer.parseInt(priority);
+      } catch (Exception e) {
+        logger.warn(
+            String.format("Invalid priority field %s for resource %s", priority, _resourceName));
+      }
+    }
+  }
+
+  private static class ResourcePriorityComparator implements Comparator<ResourcePriority> {
```
jiajunwang, line 242: If it is just calling compareTo(), do we need a special class as the comparator?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+              dataCache.getResourceConfig(resourceName).getSimpleConfig(priorityField));
+        } else if (dataCache.getIdealState(resourceName) != null
+            && dataCache.getIdealState(resourceName).getRecord().getSimpleField(priorityField)
+            != null) {
+          resourcePriority.setPriority(
+              dataCache.getIdealState(resourceName).getRecord().getSimpleField(priorityField));
+        }
+      }
+      prioritizedResourceList.sort(new ResourcePriorityComparator());
+    }
```
jiajunwang, line 119: If we just input the clusterConfig and resouceConfig as 2 parameters to the ResourcePriority constructor, then we can put all the above logic into the ResourcePriority class, right?
It would be a cleaner OO design.

jiajunwang (commented):
Thanks for the effort on splitting the PR. It definitely helps the reviewer to focus on the important logics.
Besides the inline comments, I have 2 more high-level comments.
1. Please don't use the key ward resolve in the PR description. Otherwise Github will automatically close the issue when this PR is merged. But you do have more.
2. As we discussed, each PR shall have some unit test. Although it would be simple at the beginning, it helps to ensure the quality.

lei-xia:
Can we make the PR title more meaningful, what is PR1 here mean?  I know it is your first PR on this feature, but that does not really make much sense as the PR title.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: see line 118 `prioritizedResourceList.sort(new ResourcePriorityComparator());`

The other thing is that these code (ResourcePriorityComparator etc)  is copied from intermediate stage. I think it is a good idea to keep it the same way. Otherwise, it may confuse people that we are implementing some new logic.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: these code (ResourcePriorityComparator etc) is copied from intermediate stage. I think it is a good idea to keep it the same way. Otherwise, it may confuse people that we are implementing some new logic. 

What is your take here?

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
kaisun2000: We talked about this before. 
The usage is `selectedMessage.getResourceMessages(resourceName)`. Without adding this new API, how can you construct `Map<Partition, List<Message>>`? 

One have to do something like this:

```
Map<Partition, List<Message>> partitonMsgMap = new HashMap();
for (Partition partition : resource.getPartitions()) {
   List<Message> msgList = selectedMessage.getMessage( resource, parttiton);
   partitionMsgMap.put(partition, messages);
}

```
This would loop against all partition for resource. From CPU cycle, it won't save anything. 
Also the code seems to be bloated to me.

This is not essential. If you think this way is better, we can also do it this way. Or there is another way in your mind, please draw the code out. If it works, we can also use it that way.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
kaisun2000: usage is like this:

 `clusterStatusMonitor.setResourceRebalanceStates(output.resourceSet(),
          ResourceMonitor.RebalanceStatus.NORMAL);`

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: This really boils down to where we should cut the line. Currently I am thinking let us first have skeleton in. Next PR, let us add these TODO steps and add the stage test in. 

Do do you want to have these steps and stage test in one PR? This could be a lot larger then. Several hundreds more lines.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: I think this should be `output`.

Note input resources are basically `prioritizedResourceList`.  But some resource may be failed and no messages throttled or for whatever reason.  All the resources ends up in the  `output` are resources going out. 

Or you think we should use just `prioritizedResourceList`?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: This is good idea. However, this code is carried from intermedidate stage.  is just to keep it the same as IntermediateStage to help people to understand this is not real change here. Let me know what is your take? Do we want to change?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Debug message is way better than persisting the value in ZK. The persist value only reflect the current snapshot of pipeline running. Debug messages give historically time series value. Also this is "debug" level. Normally without enabling it, it will not go out anyway. In time we really need to trouble shoot something, it would be much easier to enable it than examining persisted value or make a new build. 

If you really think this is not that necessary, we can also remove this line.

jiajunwang: Your point is valid. I think it would help in general.
But in reality, I guess the raw data would be very hard to check if the cluster is very large. It would be best if we have some abstraction.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: What is intermedidate stage? Just kidding.

This is a new code. Please assume that the intermedidate stage does not exist.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: I see. I agree with the logic. But I suggest making a new local resource name set to track the non-failure resources.
The resourceSet() method added to the output is not necessary, IMHO.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
jiajunwang: We don't need to get the whole map. I think you can just pass the output to the user method and then call the existing method to fetch the list for each resource/partition.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: Either way works. 1. Only add the skeleton (which is not the only thing in this PR), 2. add the simple logic with some tests.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: Intermediate stage has been patched too much now. Please don't refer to it. By splitting, we do expect every line to be reviewed as new code.

Also about the code style, I think Lambda is simpler. But no strong preference. If you can have a try and see which one do you prefer : )

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: Please assume intermediate stage does not exist. I will review with this assumption at least.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,248 @@
...
+    if (currentStateOutput == null || selectedMessages == null || resourceToRebalance == null
+        || cache == null) {
+      throw new StageException(String.format("Missing attributes in event: %s. "
+              + "Requires CURRENT_STATE (%s) |BEST_POSSIBLE_STATE (%s) |RESOURCES (%s) |DataCache (%s)",
+          event, currentStateOutput, selectedMessages, resourceToRebalance, cache));
+    }
+
+    ResourcesStateMap retracedResourceStateMap = new ResourcesStateMap();
+    List<Message> throttledRecoveryMsg = new ArrayList<>();
+    List<Message> throttledLoadMsg = new ArrayList<>();
```
dasahcc, line 55: Still as discussed before, we should avoid this kind of code just for testing purpose. If you need the info, we can log it where throttling happens.

kaisun2000: I understand you would like to see these lists hold messages only for testing purpose to be separated from the production logic. I will try to address this concern.

For now, I would first make simple (only 1) test pass. Will using the test as an example, I will propose some ways to achieve the goal next in this same discussion thread, with some psudo-codes. If you agree with the proposal, I will make the change subsequently. Or you can draw some psudo-codes to illustrate a better idea.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: any proposal?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Abstract this logic to new private method `getResourcePriorityList`.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Will add monitoring in a separate PR.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed for now.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
kaisun2000: removed for now as we will have another PR for monitoring.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: TODO added. 
This one we need to be careful, as current intermediateCalcStage based throttling does not extend beyond full_auto.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: fixed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Added one recovery stage test and minimal code to make it pass for now. We can start from here.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Let us Collection.sort() to get rid of `ResourcePriorityComparator` class.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Moved these logic to constructor of resource priority.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
kaisun2000: removed getResourceMessages.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Capture the test related messages into `MsgRecordingPerReplicaThrottleStage` now. Production code does not `event` to store the messages anymore.

helix-core/src/test/java/org/apache/helix/controller/stages/MsgRecordingPerReplicaThrottleStage.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.helix.controller.dataproviders.ResourceControllerDataProvider;
+import org.apache.helix.model.IdealState;
+import org.apache.helix.model.Message;
+import org.apache.helix.model.Partition;
+import org.apache.helix.model.Resource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class MsgRecordingPerReplicaThrottleStage extends PerReplicaThrottleStage {
```
jiajunwang, line 19: Any comment about what is this class for? Why we need a child class of PerReplicaThrottleStage instead of just adding the logic there?

jiajunwang (commented):
I see the new classes are not licensed. Please do so.

Moreover, I guess you didn't get my point of adding the test. Sorry that I should have made it clearer.
Overall, we want the PR to be small for reviewing. And it against our original goal that you add more logic back just to make the test complete. The test should cover whatever the code logic has, but the reverse way. For example, as I mentioned that the original logic only generates the empty result. This is hard to test. But you can just copy-paste whatever in the MessageOutput into the new throttling stage's result. In this case, I don't think you will need to add all the logic in 4d452e7.
And the test will verify 2 things only: 1. the stage can run without exception. 2. the output contains all the messages.

helix-core/src/test/java/org/apache/helix/controller/stages/MsgRecordingPerReplicaThrottleStage.java
kaisun2000: See JK's comment https://github.com/apache/helix/pull/1620#discussion_r564007432. This is to address his concern over there.

kaisun2000: JK's point in sum is this: Basically the test need to know the throttled msgs and their type. This logic is not needed in PerReplicaThrottleStage.java

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
```diff
@@ -50,5 +50,11 @@
...
   JOBS_WITHOUT_CONFIG,
   // This attribute should only be used in TaskGarbageCollectionStage, misuse could cause race conditions.
-  TO_BE_PURGED_JOBS_MAP
+  TO_BE_PURGED_JOBS_MAP,
+
+  // This attribute denotes the messages output from Per Preplica Throttle stage
+  PER_REPLICA_OUTPUT_MESSAGES,
+
+  // This attribute denotes the targeted partition state mapping from Per Preplica Throttle stage
+  PER_REPLICA_RETRACED_STATES
```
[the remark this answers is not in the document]
kaisun2000: Let use option 1. This is more flexible.

Option 2 can also be very confusing.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageOutput.java
```diff
@@ -24,6 +24,7 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
```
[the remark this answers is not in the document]
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
```diff
@@ -47,7 +47,8 @@
     UNKNOWN,
     NORMAL,
     BEST_POSSIBLE_STATE_CAL_FAILED,
-    INTERMEDIATE_STATE_CAL_FAILED
+    INTERMEDIATE_STATE_CAL_FAILED,
+    PER_REPLICA_STATE_CAL_FAILED
```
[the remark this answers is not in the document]
kaisun2000: changed.

helix-core/src/test/java/org/apache/helix/controller/stages/MsgRecordingPerReplicaThrottleStage.java
kaisun2000: Anyway, for this diff, I already removed this file. We will add them back later.

kaisun2000:
> I see the new classes are not licensed. Please do so.
> 
> Moreover, I guess you didn't get my point of adding the test. Sorry that I should have made it clearer.
> Overall, we want the PR to be small for reviewing. And it against our original goal that you add more logic back just to make the test complete. The test should cover whatever the code logic has, but the reverse way. For example, as I mentioned that the original logic only generates the empty result. This is hard to test. But you can just copy-paste whatever in the MessageOutput into the new throttling stage's result. In this case, I don't think you will need to add all the logic in [4d452e7](https://github.com/apache/helix/commit/4d452e76497d713e86acccc1d05e99d48c6ddac0).
> And the test will verify 2 things only: 1. the stage can run without exception. 2. the output contains all the messages.

Changed.

helix-core/src/test/java/org/apache/helix/controller/stages/MsgRecordingPerReplicaThrottleStage.java
jiajunwang: I thought we need these 2 lists for the metric reporting, no? Anyway, we can defer the discussion.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,270 @@
+package org.apache.helix.controller.stages;
+
+import java.util.ArrayList;
```
jiajunwang, line 3: @kaisun2000 As I commented in the previous review, please add the license here.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
jiajunwang, line 59: I thought it has been commented on, so I did not comment in the previous review.
The names of these 2 fields are very misleading. Please try to make it understandable even without reading the comment.

Also please check the other namings in the PR. Although I don't like to be strict on the naming, we shall avoid confusing names. Please double check.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,270 @@
...
+      outMessagesByPartition.put(partition, finalPartitionMessages);
+      output.addMessages(resourceName, partition, finalPartitionMessages);
+    }
+    // Step 6: constructs all retraced partition state map for the resource
+    // TODO: next PR
+    // Step 7: emit metrics
+    // TODO: next PR
+  }
+
+
```
jiajunwang, line 212: nit, please remove the extra line.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,270 @@
...
+    public void setPriority(String priority) {
+      try {
+        _priority = Integer.parseInt(priority);
+      } catch (Exception e) {
+        logger.warn(
+            String.format("Invalid priority field %s for resource %s", priority, _resourceName));
+      }
+    }
+  }
+
```
jiajunwang, line 269: nit, please remove the extra line.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,270 @@
...
+      String priorityField = dataCache.getClusterConfig().getResourcePriorityField();
+      if (priorityField != null) {
+        // Will take the priority from ResourceConfig first
+        // If ResourceConfig does not exist or does not have this field.
+        // Try to load it from the resource's IdealState. Otherwise, keep it at the lowest priority
+        if (dataCache.getResourceConfig(resourceName) != null
+            && dataCache.getResourceConfig(resourceName).getSimpleConfig(priorityField) != null) {
+          this.setPriority(
+              dataCache.getResourceConfig(resourceName).getSimpleConfig(priorityField));
+        } else if (dataCache.getIdealState(resourceName) != null
```
jiajunwang, line 236: nit, these codes could be improved a little bit, IMHO. Note the "dataCache.getResourceConfig(resourceName)" call is made 3 times.

I suggest changing to the following,
ResourceConfig config = dataCache.getResourceConfig(resourceName)
if (config != null) {
  String priority = dataCache.getResourceConfig(resourceName).getSimpleConfig(priorityField);
  this.setPriority(priority == null ? DEFAULT_PRIORITY : priority);
}

The other condition branch can be simplified in the same way.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,270 @@
...
+      List<Message> partitionMessages = selectedResourceMessages.get(partition);
+      if (partitionMessages == null) {
+        continue;
+      }
+      List<Message> finalPartitionMessages = new ArrayList<>();
+      for (Message message: partitionMessages) {
+        // TODO: next PR messages exclusion
+        finalPartitionMessages.add(message);
+      }
+      outMessagesByPartition.put(partition, finalPartitionMessages);
```
jiajunwang, line 203: This is not used in this PR, right? Can we defer adding it and re-evaluate in the future PR when it is consumed?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,270 @@
...
+   * @param bestPossibleStateOutput
+   * @param cache
+   * @param throttleController
+   * @param retracedPartitionsStateMap
+   * @param throttledRecoveryMsgOut
+   * @param throttledLoadMessageOut
+   * @param output
+   * @return
+   */
+  private void throttlePerReplicaMessages(IdealState idealState,
```
jiajunwang, line 165: Instead of adding an "output" parameter, please return the message output.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,270 @@
...
+      CurrentStateOutput currentStateOutput, Map<Partition, List<Message>> selectedResourceMessages,
+      Resource resource, BestPossibleStateOutput bestPossibleStateOutput,
+      ResourceControllerDataProvider cache, StateTransitionThrottleController throttleController,
+      Map<Partition, Map<String, String>> retracedPartitionsStateMap,
+      MessageOutput output) {
```
jiajunwang, line 170: There are some inputs not used in the logic, currentStateOutput for example. Please remove them in this PR. We will re-evaluate them whenever they are really used.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: added. Got confused with another file that was removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/AttributeName.java
kaisun2000: You mean "PER_REPLICA_THROTTLED_OUTPUT_MESSAGES"? The previous review comment about another enum. thanks for pointing this out. Just checked, these two are the only left two.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: 1. We can always add the string.
2. The conversion is happening anyway. The value you get from the Map is also a String. This is a different topic.
3. Why not make this small change to simplify code without any side-effect?

kaisun2000: Changed to 

```
      _priority = Integer.valueOf(DEFAULT_PRIORITY);
      _resourceName = resourceName;
      String priorityField = dataCache.getClusterConfig().getResourcePriorityField();

      if (priorityField != null) {
        // Will take the priority from ResourceConfig first
        // If ResourceConfig does not exist or does not have this field.
        // Try to load it from the resource's IdealState. Otherwise, keep it at the lowest priority
        ResourceConfig config = dataCache.getResourceConfig(resourceName);
        IdealState idealState = dataCache.getIdealState(resourceName);
        if (config != null) {
          String priority = dataCache.getResourceConfig(resourceName).getSimpleConfig(priorityField);
          this.setPriority(priority == null ? DEFAULT_PRIORITY : priority);
        } else if (idealState != null) {
          String priority = idealState.getRecord().getSimpleField(priorityField);
          this.setPriority(priority == null ? DEFAULT_PRIORITY : priority);
        }
      }
```

kaisun2000: On a second thought, @jiajunwang , I think there is a bug in your new proposed code. They are not equivalent to the original code. 

Assuming the above is what you suggested. The case that if resourceConfig having priorityField, but retrieved value priority is null, we have priority as min_int (DEFAULT_PRIORITY). However, in the original code, in the original code, we should check idealstate priorityField value. 

This is kind of tricky, as simplifying the original code has change to introduce new bug.  since the above code is pushed to  this PR now. Shall we use the original code or shall we use following?

```
 private final String DEFAULT_PRIORITY = Integer.valueOf(Integer.MIN_VALUE).toString();

      _priority = Integer.valueOf(DEFAULT_PRIORITY);
      _resourceName = resourceName;
      String priorityField = dataCache.getClusterConfig().getResourcePriorityField();

      if (priorityField != null) {
        // Will take the priority from ResourceConfig first
        // If ResourceConfig does not exist or does not have this field.
        // Try to load it from the resource's IdealState. Otherwise, keep it at the lowest priority
        ResourceConfig config = dataCache.getResourceConfig(resourceName);
        IdealState idealState = dataCache.getIdealState(resourceName);
        if (config != null) {
          String priority = dataCache.getResourceConfig(resourceName).getSimpleConfig(priorityField);
          if (priority != null) {
              this.setPriority priority);
          }
        } else if (idealState != null) {
          String priority = idealState.getRecord().getSimpleField(priorityField);
          this.setPriority(priority == null ? DEFAULT_PRIORITY : priority);
        }
      }

```

kaisun2000: Sync-ed with JJ offline, the main concern is `getResourceConfig` is used three time, better to use once. Same for `getIdealState`. 
Reduce to using once would work, still keeping the original code logic.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Chatted offline. Feel free to reopen.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,285 @@
...
+    // Step 1: charge existing pending messages and update retraced state map.
+    // Step 2: classify all the messages into recovery message list and load message list
+    // Step 3: sorts recovery message list and applies throttling
+    // Step 4: sorts load message list and applies throttling
+    // Step 5: construct output
+    for (Partition partition : resource.getPartitions()) {
+      List<Message> partitionMessages = selectedResourceMessages.get(partition);
+      if (partitionMessages == null) {
+        continue;
+      }
```
jiajunwang, line 216: I think the resource parameter is not necessary. It is used for 2 purposes. 1. get resource name (which is available in idealState), 2. get the partition list (which can be got from selectedResourceMessages). Let's remove it to simplify the code, and to avoid potential bugs that can be introduced by the possible conflicting parameters.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,285 @@
...
+        if (config != null && config.getSimpleConfig(priorityField) != null) {
+          this.setPriority(config.getSimpleConfig(priorityField));
+        } else if (idealState != null
+            && idealState.getRecord().getSimpleField(priorityField) != null) {
+          this.setPriority(idealState.getRecord().getSimpleField(priorityField));
+        }
+      }
+    }
+
+    ResourcePriority(String resourceName, int priority) {
```
jiajunwang, line 261: Is this constructor used anywhere?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  private List<ResourcePriority> getResourcePriorityList(
+      Map<String, Resource> resourceMap,
+      ResourceControllerDataProvider dataCache) {
+    List<ResourcePriority> prioritizedResourceList = new ArrayList<>();
+    for (String resourceName : resourceMap.keySet()) {
+      prioritizedResourceList.add(new ResourcePriority(resourceName, dataCache));
+    }
+    Collections.sort(prioritizedResourceList);
```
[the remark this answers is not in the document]
kaisun2000: changed to 
```
resourceMap.keySet().stream().forEach(
        resourceName -> prioritizedResourceList.add(new ResourcePriority(resourceName, dataCache)));
```

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,285 @@
...
+   * of possible pending state transitions does NOT go over the set threshold).
+   * @param event
+   * @param resourceMap
+   * @param currentStateOutput
+   * @param selectedMessage
+   * @param retracedResourceStateMap out
+   * @return
+   */
+  private MessageOutput compute(ClusterEvent event, Map<String, Resource> resourceMap,
+      CurrentStateOutput currentStateOutput, MessageOutput selectedMessage,
```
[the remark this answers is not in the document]
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed to
```
for (Partition partition : selectedResourceMessages.keySet()) {
}
```

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,275 @@
...
+   * Apply per-replica throttling logic and filter out excessive recovery and load messages for a
+   * given resource.
+   * Reconstruct retrace partition states for a resource based on pending and targeted messages
+   * Return messages for partitions of a resource.
+   * @param idealState
+   * @param selectedResourceMessages
+   * @param bestPossibleStateOutput
+   * @param throttleController
+   * @param retracedPartitionsStateMap
+   * @return output
```
jiajunwang, line 178: nit, but please fix it before merging. Here, return what output???

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,275 @@
...
+
+      Map<Partition, Map<String, String>> retracedPartitionsState = new HashMap<>();
+      try {
+        Map<Partition, List<Message>> partitonMsgMap = new HashMap<>();
+        for (Partition partition : resource.getPartitions()) {
+          List<Message> msgList = selectedMessage.getMessages(resourceName, partition);
+          partitonMsgMap.put(partition, msgList);
+        }
+        MessageOutput resourceMsgOut =
+            throttlePerReplicaMessages(idealState, partitonMsgMap, bestPossibleStateOutput,
```
jiajunwang, line 150: nit, not sure if you can just return a Map<Partition, List<Message>> instead of the MessageOutput? It seems to be overkill. But the logic is fine, obviously.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -0,0 +1,275 @@
...
+
+  /**
+   * Go through each resource, and based on messageSelected and currentState, compute
+   * messageOutput while maintaining throttling constraints (for example, ensure that the number
+   * of possible pending state transitions does NOT go over the set threshold).
+   * @param event
+   * @param resourceMap
+   * @param selectedMessage
+   * @param retracedResourceStateMap out
+   * @return
```
jiajunwang, line 99: Please add something here about what is returned. Or just remove @return.

jiajunwang (approved):
LGTM!

There are several minor comments. Please address before we merge the code.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: As you said, logic is obviously right, let us keep it this way.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Let me remove it, the same way as the other one.

kaisun2000:
This PR is approved, please help to merge in
> First part of Per Replica throttling, skeleton only output the input messages.

Per replica throttling replacing intermediate stage which is partition based. The finer granularity
would skip boosting unnecessary replica in a recovery partition.

PR #1628, opened 2021-02-04
Per Replica Throttle -- 2nd: messages classification and basic throttle application

kaisun2000:
### Issues

enhance #343; second part: message classification, basic throttle logic application. Also basic test cases.

### Description

### Tests

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
@@ -1,0 +1,273 @@
+package org.apache.helix.controller.stages;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.helix.HelixException;
+import org.apache.helix.controller.LogUtil;
+import org.apache.helix.controller.common.ResourcesStateMap;
+import org.apache.helix.controller.dataproviders.ResourceControllerDataProvider;
+import org.apache.helix.controller.pipeline.AbstractBaseStage;
+import org.apache.helix.controller.pipeline.StageException;
+import org.apache.helix.model.IdealState;
+import org.apache.helix.model.Message;
+import org.apache.helix.model.Partition;
+import org.apache.helix.model.Resource;
+import org.apache.helix.model.ResourceConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class PerReplicaThrottleStage extends AbstractBaseStage {
+  private static final Logger logger =
+      LoggerFactory.getLogger(PerReplicaThrottleStage.class.getName());
+
+  @Override
+  public void process(ClusterEvent event) throws Exception {
+    _eventId = event.getEventId();
+
+    CurrentStateOutput currentStateOutput = event.getAttribute(AttributeName.CURRENT_STATE.name());
+
+    MessageOutput selectedMessages = event.getAttribute(AttributeName.MESSAGES_SELECTED.name());
+    LogUtil.logDebug(logger, _eventId, String.format("selectedMessages is: %s", selectedMessages));
+
+    Map<String, Resource> resourceToRebalance =
+        event.getAttribute(AttributeName.RESOURCES_TO_REBALANCE.name());
+    ResourceControllerDataProvider cache =
+        event.getAttribute(AttributeName.ControllerDataProvider.name());
+
+    if (currentStateOutput == null || selectedMessages == null || resourceToRebalance == null
+        || cache == null) {
+      throw new StageException(String.format("Missing attributes in event: %s. "
+              + "Requires CURRENT_STATE (%s) |BEST_POSSIBLE_STATE (%s) |RESOURCES (%s) |DataCache (%s)",
+          event, currentStateOutput, selectedMessages, resourceToRebalance, cache));
+    }
+
+    ResourcesStateMap retracedResourceStateMap = new ResourcesStateMap();
+    MessageOutput output =
+        compute(event, resourceToRebalance, selectedMessages, retracedResourceStateMap);
+
+    event.addAttribute(AttributeName.PER_REPLICA_THROTTLE_OUTPUT_MESSAGES.name(), output);
+    LogUtil.logDebug(logger, _eventId,
+        String.format("retraceResourceStateMap is: %s", retracedResourceStateMap));
+    event.addAttribute(AttributeName.PER_REPLICA_THROTTLE_RETRACED_STATES.name(),
+        retracedResourceStateMap);
+
+    //TODO: enter maintenance mode logic in next PR
+  }
+
+  private List<ResourcePriority> getResourcePriorityList(Map<String, Resource> resourceMap,
+      ResourceControllerDataProvider dataCache) {
+    List<ResourcePriority> prioritizedResourceList = new ArrayList<>();
+    resourceMap.keySet().stream().forEach(
+        resourceName -> prioritizedResourceList.add(new ResourcePriority(resourceName, dataCache)));
+    Collections.sort(prioritizedResourceList);
+
+    return prioritizedResourceList;
+  }
+
+  /**
+   * Go through each resource, and based on messageSelected and currentState, compute
+   * messageOutput while maintaining throttling constraints (for example, ensure that the number
+   * of possible pending state transitions does NOT go over the set threshold).
+   * @param event
+   * @param resourceMap
+   * @param selectedMessage
+   * @param retracedResourceStateMap out
+   */
+  private MessageOutput compute(ClusterEvent event, Map<String, Resource> resourceMap,
+      MessageOutput selectedMessage, ResourcesStateMap retracedResourceStateMap) {
+    MessageOutput output = new MessageOutput();
+
+    ResourceControllerDataProvider dataCache =
+        event.getAttribute(AttributeName.ControllerDataProvider.name());
+
+    StateTransitionThrottleController throttleController =
+        new StateTransitionThrottleController(resourceMap.keySet(), dataCache.getClusterConfig(),
+            dataCache.getLiveInstances().keySet());
+
+    List<ResourcePriority> prioritizedResourceList =
+        getResourcePriorityList(resourceMap, dataCache);
+
+    List<String> failedResources = new ArrayList<>();
+
+    // Priority is applied in assignment computation because higher priority by looping in order of
+    // decreasing priority
+    for (ResourcePriority resourcePriority : prioritizedResourceList) {
+      String resourceName = resourcePriority.getResourceName();
+
+      BestPossibleStateOutput bestPossibleStateOutput =
+          event.getAttribute(AttributeName.BEST_POSSIBLE_STATE.name());
+      if (!bestPossibleStateOutput.containsResource(resourceName)) {
+        LogUtil.logInfo(logger, _eventId, String.format(
+            "Skip calculating per replica state for resource %s because the best possible state is not available.",
+            resourceName));
+        continue;
+      }
+
+      Resource resource = resourceMap.get(resourceName);
+      IdealState idealState = dataCache.getIdealState(resourceName);
+      if (idealState == null) {
+        // If IdealState is null, use an empty one
+        LogUtil.logInfo(logger, _eventId, String
+            .format("IdealState for resource %s does not exist; resource may not exist anymore",
+                resourceName));
+        idealState = new IdealState(resourceName);
+        idealState.setStateModelDefRef(resource.getStateModelDefRef());
+      }
+
+      Map<Partition, Map<String, String>> retracedPartitionsState = new HashMap<>();
+      try {
+        Map<Partition, List<Message>> partitonMsgMap = new HashMap<>();
+        for (Partition partition : resource.getPartitions()) {
+          List<Message> msgList = selectedMessage.getMessages(resourceName, partition);
+          partitonMsgMap.put(partition, msgList);
+        }
+        MessageOutput resourceMsgOut =
+            throttlePerReplicaMessages(idealState, partitonMsgMap, bestPossibleStateOutput,
+                throttleController, retracedPartitionsState);
+        for (Partition partition : resource.getPartitions()) {
+          List<Message> msgList = resourceMsgOut.getMessages(resourceName, partition);
+          output.addMessages(resourceName, partition, msgList);
+        }
+        retracedResourceStateMap.setState(resourceName, retracedPartitionsState);
+      } catch (HelixException ex) {
+        LogUtil.logInfo(logger, _eventId,
+            "Failed to calculate per replica partition states for resource " + resourceName, ex);
+        failedResources.add(resourceName);
+      }
+    }
+
+    // TODO: add monitoring in next PR.
+    return output;
+  }
+
+  /*
+   * Apply per-replica throttling logic and filter out excessive recovery and load messages for a
+   * given resource.
+   * Reconstruct retrace partition states for a resource based on pending and targeted messages
+   * Return messages for partitions of a resource.
+   * @param idealState
+   * @param selectedResourceMessages
+   * @param bestPossibleStateOutput
+   * @param throttleController
+   * @param retracedPartitionsStateMap
+   */
+  private MessageOutput throttlePerReplicaMessages(IdealState idealState,
+      Map<Partition, List<Message>> selectedResourceMessages,
+      BestPossibleStateOutput bestPossibleStateOutput,
+      StateTransitionThrottleController throttleController,
+      Map<Partition, Map<String, String>> retracedPartitionsStateMap) {
+    MessageOutput output = new MessageOutput();
+    String resourceName = idealState.getResourceName();
+    LogUtil.logInfo(logger, _eventId, String.format("Processing resource: %s", resourceName));
+
+    // TODO: expand per-replica-throttling beyond FULL_AUTO
+    if (!throttleController.isThrottleEnabled() || !IdealState.RebalanceMode.FULL_AUTO
+        .equals(idealState.getRebalanceMode())) {
+      retracedPartitionsStateMap
+          .putAll(bestPossibleStateOutput.getPartitionStateMap(resourceName).getStateMap());
+      for (Partition partition : selectedResourceMessages.keySet()) {
+        output.addMessages(resourceName, partition, selectedResourceMessages.get(partition));
+      }
+      return output;
+    }
+
+    // TODO: later PRs
+    // Step 1: charge existing pending messages and update retraced state map.
+    // Step 2: classify all the messages into recovery message list and load message list
+    // Step 3: sorts recovery message list and applies throttling
+    // Step 4: sorts load message list and applies throttling
+    // Step 5: construct output
+    for (Partition partition : selectedResourceMessages.keySet()) {
+      List<Message> partitionMessages = selectedResourceMessages.get(partition);
+      if (partitionMessages == null) {
+        continue;
+      }
+      List<Message> finalPartitionMessages = new ArrayList<>();
+      for (Message message : partitionMessages) {
+        // TODO: next PR messages exclusion
+        finalPartitionMessages.add(message);
+      }
+      output.addMessages(resourceName, partition, finalPartitionMessages);
+    }
+    // Step 6: constructs all retraced partition state map for the resource
+    // TODO: next PR
+    // Step 7: emit metrics
+    // TODO: next PR
+    return output;
+  }
+
+  // ------------------ utilities ---------------------------
+
+  /**
+   * POJO that maps resource name to its priority represented by an integer.
+   */
+  private static class ResourcePriority implements Comparable<ResourcePriority> {
+    private String _resourceName;
+    private int _priority;
+
+    ResourcePriority(String resourceName, ResourceControllerDataProvider dataCache) {
+      // Resource level prioritization based on the numerical (sortable) priority field.
+      // If the resource priority field is null/not set, the resource will be treated as lowest
+      // priority.
+      _priority = Integer.MIN_VALUE;
+      _resourceName = resourceName;
+      String priorityField = dataCache.getClusterConfig().getResourcePriorityField();
+      if (priorityField != null) {
+        // Will take the priority from ResourceConfig first
+        // If ResourceConfig does not exist or does not have this field.
+        // Try to load it from the resource's IdealState. Otherwise, keep it at the lowest priority
+        ResourceConfig config = dataCache.getResourceConfig(resourceName);
+        IdealState idealState = dataCache.getIdealState(resourceName);
+        if (config != null && config.getSimpleConfig(priorityField) != null) {
+          this.setPriority(config.getSimpleConfig(priorityField));
+        } else if (idealState != null
+            && idealState.getRecord().getSimpleField(priorityField) != null) {
+          this.setPriority(idealState.getRecord().getSimpleField(priorityField));
+        }
+      }
+    }
+
+    @Override
+    public int compareTo(ResourcePriority resourcePriority) {
+      // make sure larger _priority is in front of small _priority at sort time
+      return Integer.compare(resourcePriority._priority, _priority);
+    }
+
+    public String getResourceName() {
+      return _resourceName;
+    }
+
+    public void setPriority(String priority) {
+      try {
+        _priority = Integer.parseInt(priority);
+      } catch (Exception e) {
+        logger.warn(
+            String.format("Invalid priority field %s for resource %s", priority, _resourceName));
+      }
+    }
+  }
+}
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
@@ -22,10 +22,15 @@ package org.apache.helix.controller.stages;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
+import org.apache.helix.HelixDefinedState;
 import org.apache.helix.HelixException;
+import org.apache.helix.api.config.StateTransitionThrottleConfig;
 import org.apache.helix.controller.LogUtil;
 import org.apache.helix.controller.common.ResourcesStateMap;
 import org.apache.helix.controller.dataproviders.ResourceControllerDataProvider;
@@ -36,6 +41,7 @@ import org.apache.helix.model.Message;
 import org.apache.helix.model.Partition;
 import org.apache.helix.model.Resource;
 import org.apache.helix.model.ResourceConfig;
+import org.apache.helix.model.StateModelDefinition;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -66,8 +72,8 @@ public class PerReplicaThrottleStage extends AbstractBaseStage {
     }
 
     ResourcesStateMap retracedResourceStateMap = new ResourcesStateMap();
-    MessageOutput output =
-        compute(event, resourceToRebalance, selectedMessages, retracedResourceStateMap);
+    MessageOutput output = compute(event, resourceToRebalance, currentStateOutput, selectedMessages,
+        retracedResourceStateMap);
 
     event.addAttribute(AttributeName.PER_REPLICA_THROTTLE_OUTPUT_MESSAGES.name(), output);
     LogUtil.logDebug(logger, _eventId,
@@ -94,11 +100,13 @@ public class PerReplicaThrottleStage extends AbstractBaseStage {
    * of possible pending state transitions does NOT go over the set threshold).
    * @param event
    * @param resourceMap
+   * @param currentStateOutput
    * @param selectedMessage
    * @param retracedResourceStateMap out
    */
   private MessageOutput compute(ClusterEvent event, Map<String, Resource> resourceMap,
-      MessageOutput selectedMessage, ResourcesStateMap retracedResourceStateMap) {
+      CurrentStateOutput currentStateOutput, MessageOutput selectedMessage,
+      ResourcesStateMap retracedResourceStateMap) {
     MessageOutput output = new MessageOutput();
 
     ResourceControllerDataProvider dataCache =
@@ -113,8 +121,7 @@ public class PerReplicaThrottleStage extends AbstractBaseStage {
 
     List<String> failedResources = new ArrayList<>();
 
-    // Priority is applied in assignment computation because higher priority by looping in order of
-    // decreasing priority
+    // Priority is applied in assignment computation
     for (ResourcePriority resourcePriority : prioritizedResourceList) {
       String resourceName = resourcePriority.getResourceName();
 
@@ -146,8 +153,8 @@ public class PerReplicaThrottleStage extends AbstractBaseStage {
           partitonMsgMap.put(partition, msgList);
         }
         MessageOutput resourceMsgOut =
-            throttlePerReplicaMessages(idealState, partitonMsgMap, bestPossibleStateOutput,
-                throttleController, retracedPartitionsState);
+            throttlePerReplicaMessages(idealState, currentStateOutput, partitonMsgMap,
+                bestPossibleStateOutput, dataCache, throttleController, retracedPartitionsState);
         for (Partition partition : resource.getPartitions()) {
           List<Message> msgList = resourceMsgOut.getMessages(resourceName, partition);
           output.addMessages(resourceName, partition, msgList);
@@ -170,14 +177,16 @@ public class PerReplicaThrottleStage extends AbstractBaseStage {
    * Reconstruct retrace partition states for a resource based on pending and targeted messages
    * Return messages for partitions of a resource.
    * @param idealState
+   * @param currentStateOutput
    * @param selectedResourceMessages
    * @param bestPossibleStateOutput
+   * @param cache
    * @param throttleController
    * @param retracedPartitionsStateMap
    */
   private MessageOutput throttlePerReplicaMessages(IdealState idealState,
-      Map<Partition, List<Message>> selectedResourceMessages,
-      BestPossibleStateOutput bestPossibleStateOutput,
+      CurrentStateOutput currentStateOutput, Map<Partition, List<Message>> selectedResourceMessages,
+      BestPossibleStateOutput bestPossibleStateOutput, ResourceControllerDataProvider cache,
       StateTransitionThrottleController throttleController,
       Map<Partition, Map<String, String>> retracedPartitionsStateMap) {
     MessageOutput output = new MessageOutput();
@@ -195,11 +204,49 @@ public class PerReplicaThrottleStage extends AbstractBaseStage {
       return output;
     }
 
-    // TODO: later PRs
+    Set<Partition> partitionsWithErrorStateReplica = new HashSet<>();
+
+    Map<Partition, Map<String, Integer>> expectedStateCountByPartition = new HashMap<>();
+    Map<Partition, Map<String, Integer>> currentStateCountsByPartition = new HashMap<>();
+
+    calculateExistingAndCurrentStateCount(selectedResourceMessages, currentStateOutput,
+        bestPossibleStateOutput, idealState, cache, expectedStateCountByPartition,
+        currentStateCountsByPartition);
+
     // Step 1: charge existing pending messages and update retraced state map.
+    // TODO: later PRs
     // Step 2: classify all the messages into recovery message list and load message list
+    List<Message> recoveryMessages = new ArrayList<>();
+    List<Message> loadMessages = new ArrayList<>();
+    classifyMessages(resourceName, idealState, cache, selectedResourceMessages, recoveryMessages,
+        loadMessages, expectedStateCountByPartition, currentStateCountsByPartition);
+
     // Step 3: sorts recovery message list and applies throttling
+    Set<Message> throttledRecoveryMessages = new HashSet<>();
+    LogUtil.logDebug(logger, _eventId,
+        String.format("applying recovery rebalance with resource %s", resourceName));
+    applyThrottling(resourceName, throttleController, idealState, cache, false, recoveryMessages,
+        throttledRecoveryMessages, StateTransitionThrottleConfig.RebalanceType.RECOVERY_BALANCE);
+
     // Step 4: sorts load message list and applies throttling
+    // Perform regular load balance only if the number of partitions in recovery and in error is
+    // less than the threshold. Otherwise, only allow downward-transition load balance
+    // TODO: calculate error-on-recovery downward threshold
+    boolean onlyDownwardLoadBalance = partitionsWithErrorStateReplica.size() > 1;
+    Set<Message> throttledLoadMessages = new HashSet<>();
+    LogUtil.logDebug(logger, _eventId, String
+        .format("applying load rebalance with resource %s, onlyDownwardLoadBalance %s",
+            resourceName, onlyDownwardLoadBalance));
+    applyThrottling(resourceName, throttleController, idealState, cache, onlyDownwardLoadBalance,
+        loadMessages, throttledLoadMessages,
+        StateTransitionThrottleConfig.RebalanceType.LOAD_BALANCE);
+
+    LogUtil.logDebug(logger, _eventId, String
+        .format("resource %s, throttled recovery message: %s", resourceName,
+            throttledRecoveryMessages));
+    LogUtil.logDebug(logger, _eventId, String
+        .format("resource %s, throttled load messages: %s", resourceName, throttledLoadMessages));
+
     // Step 5: construct output
     for (Partition partition : selectedResourceMessages.keySet()) {
       List<Message> partitionMessages = selectedResourceMessages.get(partition);
@@ -208,7 +255,12 @@ public class PerReplicaThrottleStage extends AbstractBaseStage {
       }
       List<Message> finalPartitionMessages = new ArrayList<>();
       for (Message message : partitionMessages) {
-        // TODO: next PR messages exclusion
+        if (throttledRecoveryMessages.contains(message)) {
+          continue;
+        }
+        if (throttledLoadMessages.contains(message)) {
+          continue;
+        }
         finalPartitionMessages.add(message);
       }
       output.addMessages(resourceName, partition, finalPartitionMessages);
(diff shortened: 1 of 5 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+      expectedStateCountMap
+          .put(curState, prevCnt + expectedStateCountMap.getOrDefault(curState, 0));
+      prevState = curState;
+    }
+  }
+
+  private void getPartitionExpectedAndCurrentStateCountMap(Partition partition,
+      Map<String, List<String>> preferenceLists, IdealState idealState,
+      ResourceControllerDataProvider cache, Map<String, String> currentStateMap,
+      Map<String, Integer> expectedStateCountMapOut, Map<String, Integer> currentStateCountsOut) {
```
pkuwm, line 306: A function having 2 output params just looks unclean to me :) It'd be better if we can figure out a cleaner method.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+        break;
+      }
+      index++;
+      String curState = stateList.get(index);
+      String num = stateModelDef.getNumInstancesPerState(curState);
+      if ("-1".equals(num)) {
+        break;
+      }
+      Integer prevCnt = expectedStateCountMap.get(prevState);
+      expectedStateCountMap
```
pkuwm, line 297: Just a question: for index = 0, prevState is the same as curState. Is it correct?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+  }
+
+  private void getPartitionExpectedAndCurrentStateCountMap(Partition partition,
+      Map<String, List<String>> preferenceLists, IdealState idealState,
+      ResourceControllerDataProvider cache, Map<String, String> currentStateMap,
+      Map<String, Integer> expectedStateCountMapOut, Map<String, Integer> currentStateCountsOut) {
+    List<String> preferenceList = preferenceLists.get(partition.getPartitionName());
+    if (preferenceList == null) {
+      preferenceList = Collections.emptyList();
+    }
```
pkuwm, line 310: `getOrDefault()`?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -195,11 +204,48 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+            resourceName, onlyDownwardLoadBalance));
+    applyThrottling(resourceName, throttleController, idealState, cache, onlyDownwardLoadBalance,
+        loadMessages, throttledLoadMessages,
+        StateTransitionThrottleConfig.RebalanceType.LOAD_BALANCE);
+
+    LogUtil.logDebug(logger, _eventId, String
+        .format("resource %s, throttled recovery message: %s", resourceName,
+            throttledRecoveryMessages));
+    LogUtil.logDebug(logger, _eventId, String
+        .format("resource %s, throttled load messages: %s", resourceName, throttledLoadMessages));
```
pkuwm, line 247: logDebug is not protected by `if debugEnabled`, which would cause unexpected string creation overhead.
Please add `if debugEnabled`, or use `logger.debug();`, so strings won't be created/formatted at info/error logging level. 

Same for other `logDebug` in this PR.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+                .format("Message: %s not subject to throttle in resource: %s with toState %s", msg,
+                    resourceName, toState));
+          }
+          continue;
+        }
+
+        Integer expectedCount = expectedStateCountMap.get(toState);
+        Integer currentCount = currentStateCounts.get(toState);
+        expectedCount = expectedCount == null ? 0 : expectedCount;
+        currentCount = currentCount == null ? 0 : currentCount;
```
pkuwm, line 428: `map.getOrDefault()` will help.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+      return;
+    }
+
+    // TODO: add message sorting in next PR
+    logger.trace("throttleControllerstate->{} before load", throttleController);
+    for (Message msg : messages) {
+      if (onlyDownwardLoadBalance) {
+        if (!isDownwardTransition(idealState, cache, msg)) {
+          throttledMessages.add(msg);
+          if (logger.isDebugEnabled()) {
```
pkuwm, line 457: The debug code is kind of distracting. If they are really necessary, I would suggest changing them to `logger.debug()` which could reduce lines of distracting code.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+          recoveryMessages.add(msg);
+          currentStateCounts.put(toState, currentCount + 1);
+        } else {
+          loadMessages.add(msg);
+        }
+      }
+    }
+  }
+
+  protected void applyThrottling(String resourceName,
```
pkuwm, line 440: 8 params look too many to me. I would simplify it if possible. There are too many logic branches inside the method, based on  `onlyDownwardLoadBalance` and `rebalanceType`, which could also conflict when `isRecovery && onlyDownwardLoadBalance`. Might not be a good design. Not sure if we could split it into two methods for load rebalance and recover rebalance, or upward and downward..

Roughly looking, `throttledMessages` is the output, why not put it as a return?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
     return output;
   }
 
+  private void propagateCountsTopDown(StateModelDefinition stateModelDef,
+      Map<String, Integer> expectedStateCountMap) {
+    // attribute state in higher priority to lower priority
+    List<String> stateList = stateModelDef.getStatesPriorityList();
+    if (stateList.size() <= 0) {
```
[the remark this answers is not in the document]
kaisun2000: changed to `if ( stateList == null || stateList.size() <= 0) {`. 

Properly constructed stateModelDef should not have null StatePriorityList.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+      return;
+    }
+    int index = 0;
+    String prevState = stateList.get(index);
+    if (!expectedStateCountMap.containsKey(prevState)) {
+      expectedStateCountMap.put(prevState, 0);
+    }
```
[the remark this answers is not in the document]
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: right. The idea is to add the index 0 count to index 1 count and then to index2 count ...

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: let me sync with you offline. break would have a lot of duplicated code.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed to 

`List<String> preferenceList =
        preferenceLists.getOrDefault(partition.getPartitionName(), Collections.emptyList());`

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed to
 `        Integer expectedCount = expectedStateCountMap.getOrDefault(toState, 0);
        Integer currentCount = currentStateCounts.getOrDefault(toState, 0);

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+        if (!Message.MessageType.STATE_TRANSITION.name().equals(msg.getMsgType())) {
+          if (logger.isDebugEnabled()) {
+            LogUtil.logDebug(logger, _eventId, String
+                .format("Message: %s not subject to throttle in resource: %s with type %s", msg,
+                    resourceName, msg.getMsgType()));
+          }
+          continue;
+        }
+
+        boolean isUpward = !isDownwardTransition(idealState, cache, msg);
```
jiajunwang, line 398: Hasn't we decided to put only "basic throttle" logic in this PR? The DownwardTransition shall be put in a later PR, right?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+
+        // for disabled disabled instance, the downward transition is not subjected to load throttling
```
jiajunwang, line 400: nit, but this kind of comment really prevents the reviewers to understand your intention. Please revisit and fix here and if more comments like this.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+          }
+          continue;
+        }
+
+        Integer expectedCount = expectedStateCountMap.get(toState);
+        Integer currentCount = currentStateCounts.get(toState);
+        expectedCount = expectedCount == null ? 0 : expectedCount;
+        currentCount = currentCount == null ? 0 : currentCount;
+
+        if (isUpward && (currentCount < expectedCount)) {
```
jiajunwang, line 430: The first sentence of the related issue is, "MIN_ACTIVE replica is only applied in DelayAutoRebalancer and not respected by throttling logic in Helix." So I think is logic is not what we want eventually. Are you planning to add it in a later PR? If so, then let's add a TODO here.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+          recoveryMessages.add(msg);
+          currentStateCounts.put(toState, currentCount + 1);
```
jiajunwang, line 432: Why do we need to change currentStateCounts because of the classifying?

I mean I know the reason. But please add a comment here.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+      Map<Partition, Map<String, Integer>> currentStateCountsByPartition) {
+    String resourceName = idealState.getResourceName();
+    Map<String, List<String>> preferenceLists =
+        bestPossibleStateOutput.getPreferenceLists(resourceName);
+    for (Partition partition : selectedResourceMessages.keySet()) {
+      Map<String, String> currentStateMap =
+          currentStateOutput.getCurrentStateMap(resourceName, partition);
+
+      Map<String, Integer> expectedStateCounts = new HashMap<>();
+      Map<String, Integer> currentStateCounts = new HashMap<>();
```
jiajunwang, line 352: I guess you want to use an ordered map here? Or what's the point of referring to the state model and read state priority in the later logic?

jiajunwang (commented):
I have reviewed the main logic. And I have some difficulties understanding the logic. I can barely read through since I am familiar with the previous logic. But for the others, I guess they may not be able to follow. I suggest adding more comments, and carefully refine the parameters' names (they are really confusing).

In addition, I left some questions about the logic as well. However, due to the fact that I'm not sure if I get what you want to achieve 100%, some of them might be inaccurate. Please try to clarify the logic more.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: This is critical. Let me add a comment here.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
     return output;
   }
 
+  private void propagateCountsTopDown(StateModelDefinition stateModelDef,
+      Map<String, Integer> expectedStateCountMap) {
```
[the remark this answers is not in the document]
kaisun2000: right.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+    currentStateMapWithoutDisabled.keySet().removeAll(cache
+        .getDisabledInstancesForPartition(idealState.getResourceName(),
+            partition.getPartitionName()));
+    Map<String, Integer> currentStateCounts =
+        StateModelDefinition.getStateCounts(currentStateMapWithoutDisabled);
+
+    expectedStateCountMapOut.putAll(expectedStateCountMap);
+    currentStateCountsOut.putAll(currentStateCounts);
+    propagateCountsTopDown(stateModelDef, expectedStateCountMapOut);
+    propagateCountsTopDown(stateModelDef, currentStateCountsOut);
```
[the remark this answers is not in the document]
kaisun2000: `List<String> stateList = stateModelDef.getStatesPriorityList();` 
inside, we need to loop from top state to lower state.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: See the following, this logic is working for FULL_AUTO only.  Is this going to be a problem? Let me sync with you offline.

```
private MessageOutput throttlePerReplicaMessages

    // TODO: expand per-replica-throttling beyond FULL_AUTO
    if (!throttleController.isThrottleEnabled() || !IdealState.RebalanceMode.FULL_AUTO
        .equals(idealState.getRebalanceMode())) {
      retracedPartitionsStateMap
          .putAll(bestPossibleStateOutput.getPartitionStateMap(resourceName).getStateMap());
      for (Partition partition : selectedResourceMessages.keySet()) {
        output.addMessages(resourceName, partition, selectedResourceMessages.get(partition));
      }
      return output;
    }

```

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: This is why we have the populateTopDownLogic. Let me add some comment, see if it make sense to you. Or we can sync offline.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
     return output;
   }
 
+  private void propagateCountsTopDown(StateModelDefinition stateModelDef,
```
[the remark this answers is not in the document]
kaisun2000: Let me give an example to illustrate

N1(O), N2(S), N3(O), message N3(O->S) should be classified as recovery.

Assuming without the propagateCounts logic,

Here, the expectedStateCountMap is {M->1, S->1, O->0}, given MIN_ACTIVE=2. 
The currentStateCountMap is {M->0, S->1, O->0}

See the message N3(O->S), the logic to test load/recovery is following:

```
if (isUpward && (currentCount < expectedCount)) {
          recoveryMessages.add(msg);
          currentStateCounts.put(toState, currentCount + 1);
        } else {
          loadMessages.add(msg);
        }
```
Here the currentCount (slave) is 1. The expectedCount is also 1.  Since current state no less than expected count for this state (SLAVE), it would classify the message as load. This is wrong of course.

With population logic, we have expectedStateCountMap is {M->1, S->2 O->3}. currentStateCountMap is {M->0, S->1, O->1}. This is going to work.

In theory,

When determining a message as LOAD or RECOVERY,  we look at the toState of this message. If the toState current count meet the required expected count of the toState, we will treat it as Load, otherwise, it is Recovery.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+      ResourceControllerDataProvider cache, Map<Partition, List<Message>> selectedResourceMessages,
+      List<Message> recoveryMessages, List<Message> loadMessages,
+      Map<Partition, Map<String, Integer>> expectedStateCountByPartition,
+      Map<Partition, Map<String, Integer>> currentStateCountsByPartition) {
+    LogUtil.logInfo(logger, _eventId,
+        String.format("Classify message for resource: %s", resourceName));
+
+    for (Partition partition : selectedResourceMessages.keySet()) {
+      Map<String, Integer> expectedStateCountMap = expectedStateCountByPartition.get(partition);
+      Map<String, Integer> currentStateCounts = currentStateCountsByPartition.get(partition);
```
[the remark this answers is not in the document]
kaisun2000: >The names of these vars seem to be set reversely.
The current expectedStateCountMap is really expectedStateCountByPartition, the current expectedStateCountByPartition is really the whole count map.

expectedStateCountMap is for one partition.
expectedStateCountByPartition means keyed by partition. Thus a map keyed by partition.

Let me rename to the following:

```
Map<String, Integer> partitionExpectedStateCountsOne = expectedStateCountByPartition.get(partition);
Map<String, Integer> partitionCurrentStateCounts = currentStateCountByPartition.get(partition);

```

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+    Map<String, String> currentStateMapWithoutDisabled = new HashMap<>(currentStateMap);
+    currentStateMapWithoutDisabled.keySet().removeAll(cache
+        .getDisabledInstancesForPartition(idealState.getResourceName(),
+            partition.getPartitionName()));
+    Map<String, Integer> currentStateCounts =
+        StateModelDefinition.getStateCounts(currentStateMapWithoutDisabled);
+
+    expectedStateCountMapOut.putAll(expectedStateCountMap);
+    currentStateCountsOut.putAll(currentStateCounts);
+    propagateCountsTopDown(stateModelDef, expectedStateCountMapOut);
```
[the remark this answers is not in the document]
kaisun2000: No. See my comments as why we need propagateCountsTopDown.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -146,8 +153,8 @@ private MessageOutput compute(ClusterEvent event, Map<String, Resource> resource
           partitonMsgMap.put(partition, msgList);
         }
         MessageOutput resourceMsgOut =
-            throttlePerReplicaMessages(idealState, partitonMsgMap, bestPossibleStateOutput,
-                throttleController, retracedPartitionsState);
+            throttlePerReplicaMessages(idealState, currentStateOutput, partitonMsgMap,
+                bestPossibleStateOutput, dataCache, throttleController, retracedPartitionsState);
```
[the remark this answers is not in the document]
kaisun2000: let me see how long the parameter list would be.

cache.getStateModelDef(stateModelDefName);
cache.getEnabledLiveInstances()

This one is deadly, as we need to get per partition. We have to have cache in throttlePerReplicaMessages.

cache
        .getDisabledInstancesForPartition(idealState.getResourceName(),
            partition.getPartitionName())

Let reduce passing cache to some private method called by `throttlePerReplicaMessages` if the don't call getDisabledInstancesForPartition . But I can't get rid of cache here. Some private method do need it.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+      if (index == stateList.size() - 1) {
+        break;
+      }
+      index++;
+      String curState = stateList.get(index);
+      String num = stateModelDef.getNumInstancesPerState(curState);
+      if ("-1".equals(num)) {
+        break;
+      }
+      Integer prevCnt = expectedStateCountMap.get(prevState);
```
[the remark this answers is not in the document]
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Will change to use two different private methods.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+                .format("Message: %s throttled in resource as not downward: %s with type: %s", msg,
+                    resourceName, rebalanceType));
+          }
+          continue;
+        }
+      }
+
+      if (throttleController.shouldThrottleForResource(rebalanceType, resourceName)) {
+        throttledMessages.add(msg);
+        if (logger.isDebugEnabled()) {
```
zhangmeng916, line 468: We now have logs for the total number of partitions that get throttled for different rebalance types before. Do you have that kind of general log information besides in debug mode?

helix-core/src/test/java/org/apache/helix/controller/stages/TestPerReplicaThrottleStage.java
```diff
@@ -100,6 +115,246 @@ public void testNoThrottleMessagePass() {
     Assert.assertTrue(msg.getId().equals("001"));
   }
 
+  // case 0. N1(O), N2(S), N3(O), message N3(O->S) is treated as recovery
```
zhangmeng916, line 118: I was thinking a test case that can validate the following scenario:
Current states: N1(S), N2(O), N3(O)
BestPossible: N1(M), N2(S), N3(S)
Only one message is considered as recovery rebalance based on your defined priority rule. 
Do you think this can be testified in this PR? You current test cases do not seem to be testing this.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+            : idealState.getMinActiveReplicas();
+    Set<String> activeList = new HashSet<>(preferenceList);
+    activeList.retainAll(cache.getEnabledLiveInstances());
+
+    // For each state, check that this partition currently has the required number of that state as
+    // required by StateModelDefinition.
+    String stateModelDefName = idealState.getStateModelDefRef();
+    StateModelDefinition stateModelDef = cache.getStateModelDef(stateModelDefName);
+    LinkedHashMap<String, Integer> expectedStateCountMap =
+        stateModelDef.getStateCountMap(activeList.size(), replica); // StateModelDefinition's counts
```
zhangmeng916, line 323: Please reformat all the code in the editor and clean up comments.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -94,11 +100,13 @@ public void process(ClusterEvent event) throws Exception {
...
    * @param selectedMessage
    * @param retracedResourceStateMap out
    */
   private MessageOutput compute(ClusterEvent event, Map<String, Resource> resourceMap,
-      MessageOutput selectedMessage, ResourcesStateMap retracedResourceStateMap) {
+      CurrentStateOutput currentStateOutput, MessageOutput selectedMessage,
+      ResourcesStateMap retracedResourceStateMap) {
     MessageOutput output = new MessageOutput();
 
     ResourceControllerDataProvider dataCache =
```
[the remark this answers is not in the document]
kaisun2000: See my comment bellow, there are two places using `cache.getDisabledInstancesForPartition(resourceName, partition.getPartitionName());`. 

We can't really get rid of it here. But I did cleaned all the path inside `throttlePerReplicaMessages` that won't pass cache around unless we need call the above call.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: We can't really get rid of it here. But I did cleaned all the path inside throttlePerReplicaMessages so that it won't pass cache around unless we need call the above api

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: see my comment in `propagateCountsTopDown`

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed. special treatment of like this would be in next PRs

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -270,4 +539,24 @@ public void setPriority(String priority) {
...
+  private boolean isDownwardTransition(IdealState idealState, ResourceControllerDataProvider cache,
+      Message message) {
+    boolean isDownward = false;
+
+    String stateModelDefName = idealState.getStateModelDefRef();
+    StateModelDefinition stateModelDef = cache.getStateModelDef(stateModelDefName);
+    Map<String, Integer> statePriorityMap = stateModelDef.getStatePriorityMap();
+    String fromState = message.getFromState();
+    String toState = message.getToState();
+    if (statePriorityMap.containsKey(fromState) && statePriorityMap.containsKey(toState)) {
```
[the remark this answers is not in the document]
kaisun2000: change to following comment:

// Only when both fromState and toState can be found in the statePriorityMap, comparision of
    // state priority is used to determine if the transition is downward. Otherwise, we can't
    // really determine the order and by default consider the transition not downard.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -195,11 +204,48 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+
     // Step 3: sorts recovery message list and applies throttling
+    Set<Message> throttledRecoveryMessages = new HashSet<>();
+    LogUtil.logDebug(logger, _eventId,
+        String.format("applying recovery rebalance with resource %s", resourceName));
+    applyThrottling(resourceName, throttleController, idealState, cache, false, recoveryMessages,
+        throttledRecoveryMessages, StateTransitionThrottleConfig.RebalanceType.RECOVERY_BALANCE);
+
     // Step 4: sorts load message list and applies throttling
+    // TODO: calculate error-on-recovery downward threshold with complex backward compatibility next
```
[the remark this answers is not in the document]
kaisun2000: Basic throttle is what we have now, namely classifying the messages to load/recovery, throttle accordingly.
More features are 1/ pending message accounting 2/ Error or recovery with relationship to downward only 3/ disabled partition handling 4/ dropped and error message handling. 

None of them are nice to have. They all must to have. We will add them in the subsequent PRs.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: fixed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -208,7 +254,12 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
       }
       List<Message> finalPartitionMessages = new ArrayList<>();
       for (Message message : partitionMessages) {
-        // TODO: next PR messages exclusion
+        if (throttledRecoveryMessages.contains(message)) {
```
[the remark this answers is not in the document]
kaisun2000: changed

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: I guess this was intentional, from intermedidateStage. Let me remove "// StateModelDefinition's counts". No value of this comment as it is obvious anyway.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: A little bit confused. let me sync with you offline.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,224 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+    Map<String, Integer> currentStateCounts =
+        StateModelDefinition.getStateCounts(currentStateMapWithoutDisabled);
+
+    expectedStateCountMapOut.putAll(expectedStateCountMap);
+    currentStateCountsOut.putAll(currentStateCounts);
+    propagateCountsTopDown(stateModelDef, expectedStateCountMapOut);
+    propagateCountsTopDown(stateModelDef, currentStateCountsOut);
+  }
+
+  void calculateExistingAndCurrentStateCount(Map<Partition, List<Message>> selectedResourceMessages,
```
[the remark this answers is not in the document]
kaisun2000: You are right. Now it is broken to two functions The typo does not exists anymore.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: Think about if there is specially defined statemodel which has no statelist. That does mean it is necessarily wrong. It is just that there is not state count  top down calculation needed, right?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: changed description.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: comment added. We can discuss in that thread. Resolve this one.

helix-core/src/test/java/org/apache/helix/controller/stages/TestPerReplicaThrottleStage.java
kaisun2000: added as `testOneRecoveryOneLoad`.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
jiajunwang: This is a nice design, but does it cover this case, current state {M>0, S>2, 0>1}, and the message is O->S. The expectation is that the message will be done and then the S->M is the next one. However, it will be treated as loadbalance, right?

It seems to be a very difficult problem. But I think we need to address it because the original algorithm will generate the right result.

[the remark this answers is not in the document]
jiajunwang: Another potential flaw of this design is that we do not consider if a state transition is possible between 2 states. The current implementation implies that a lower priority state can always be transited to a higher priority state. This is not always true.

jiajunwang:
I guess the core of this PR is how we calculate the expected replica count and how we use it to categorize the message. I have some questions regarding the design, already comment on the PR. But I think it would be more efficient if you can help to host a meeting for the stakeholders to discuss it.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,239 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+      IdealState idealState, Set<String> enabledLiveInstance) {
+    Map<String, Integer> expectedStateCountsOut = new HashMap<>();
+
+    List<String> preferenceList = preferenceLists.get(partition.getPartitionName());
+    if (preferenceList == null) {
+      preferenceList = Collections.emptyList();
+    }
+
+    int replica =
+        idealState.getMinActiveReplicas() == -1 ? idealState.getReplicaCount(preferenceList.size())
```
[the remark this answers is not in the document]
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: As discussed offline, here is reasoning about first case as agreed.

The first issue raised can be illustrated use this example:
```
Preference list N0, N1, N2
Current N0 - O, N1 - S, N2 - S
current count  M(0), S(2), O (2)
Message
      N0 O -> S (Load)

```
The N0 (0->S) is classified as load in this case. Since eventually N0 will be master based on PreferenceList. So should we give it preference as recovery? 

For now, we decide not do this kind of optimization for the reason that classifying it as load won't delay a master being promoted to N1. Thus, eventually N0 would be promoted to Master once N0 arrive at slave state. The gist is best possible would step-by-step lead to preference list. For now, we can rely on best possible stage to make the choice. We will note this potential optimization in design doc.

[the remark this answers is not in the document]
kaisun2000: > Another potential flaw of this design is that we do not consider if a state transition is possible between 2 states. The current implementation implies that a lower priority state can always be transited to a higher priority state. This is not always true.

This is addressed with updated `propagateCountsTopDown` implementation as discussed. The higher priority state count would be added to all the lower states that can transit to the higher priority state.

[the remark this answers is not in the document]
kaisun2000: ping

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,274 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+      // for all states S that can transition to curState, add curState count back to S in stateCountMap
+      for (String fromState : fromStatesMap.getOrDefault(curState, Collections.emptySet())) {
+        Integer fromStateCount = stateCountMap.getOrDefault(fromState, 0);
+        stateCountMap.put(fromState, Integer.sum(fromStateCount, curCount));
+      }
+      index++;
+    }
+  }
+
+  Map<String, Integer> getPartitionExpectedStateCounts(Partition partition,
```
jiajunwang, line 356: It seems that the only difference between getPartitionExpectedStateCounts and getPartitionCurrentStateCounts is that you need to calculate the expected state count map from the preference list. Is it possible to move this logic outside to calculateExpectedStateCounts so that we can avoid some duplicate code?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,274 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+        // TODO: add disabled disabled instance special treatment
+
+        String toState = msg.getToState();
+
+        // TODO: dropped and Error state special treatment
+
+        Integer minimumRequiredCount = partitionExpectedStateCounts.getOrDefault(toState, 0);
+        Integer currentCount = partitionCurrentStateCounts.getOrDefault(toState, 0);
+
+        //
```
jiajunwang, line 480: ?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,274 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+        if (isUpward && (currentCount < minimumRequiredCount)) {
+          recoveryMessages.add(msg);
+          // It is critical to increase toState value by one here. For example, current state
+          // of 3 replica in a partition is (M, O, O). Two messages here bringing up the two O to S.
+          // In this case, the first O->S would be classified as recovery. Then this line would
+          // increase the S state in partitionCurrentStateCounts value by one. Next O->S message
+          // would be correctly marked as load message.
+          partitionCurrentStateCounts.put(toState, currentCount + 1);
```
jiajunwang, line 488: Do we need to propagate the count here too? Since the original input count was calculated with propagation, right?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,274 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+    if (isRecovery && onlyDownwardLoadBalance) {
+      logger.error("onlyDownwardLoadBalance can't be used together with recovery_rebalance");
+      return;
+    }
+
+    // TODO: add message sorting in next PR
+    logger.trace("throttleControllerstate->{} before load", throttleController);
+    for (Message msg : messages) {
+      if (onlyDownwardLoadBalance) {
+        if (!isDownwardTransition(stateModelDef, msg)) {
```
jiajunwang, line 511: nit, if (onlyDownwardLoadBalance && !isDownwardTransition(stateModelDef, msg))... for simplicity.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -270,4 +589,23 @@ public void setPriority(String priority) {
       }
     }
   }
+
+  private boolean isDownwardTransition(StateModelDefinition stateModelDef, Message message) {
```
jiajunwang, line 593: I feel you can use some variance of this method to calculate the propagate state count map. If this is the case, we can move this method to the state model definition (or a util class) to cleanup the code here. More importantly, the class will provide better leverage.

jiajunwang (commented):
It seems the PR is getting large again.
I would suggest splitting the message classification from the other logic. Maybe we can organize the code better, like some util classes, to make the code cleaner.
Of course, we can keep reviewing the current PR. But with its current scope, I have to find a long period of time to review. Otherwise, I get lost easily.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
jiajunwang: I think the new implementation reflects the overall design as we discussed. So I only have several comments about the code itself. But the overall direction looks fine to me.

But still, please try to schedule a meeting with Junkai, who might have a better idea.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: I was looking back to this previous comment because I remember that I suggest that we put only "basic throttle" logic in this PR. But it seems not the case : )

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
jiajunwang: Where is the change, please?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,274 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+    if (stateList == null || stateList.size() <= 0) {
+      return;
+    }
+
+    Map<String, Integer> statePriorityMap = new HashMap<>();
+
+    // calculate rank of each state. Next use the rank to compare if a transition is upward or not
+    int rank = 0;
+    for (String state : stateList) {
+      statePriorityMap.put(state, Integer.valueOf(rank));
```
dasahcc, line 312: So the rank is int, why we need Integer.valueOf(rank)?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,274 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+      rank++;
```
dasahcc, line 313: I believe state model should have the function to get priority of states. Even if it does not have it, this computation logic should be there.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,274 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+        continue;
+      }
+      fromStatesMap.putIfAbsent(toState, new HashSet<>());
+      fromStatesMap.get(toState).add(fromState);
+    }
+
+    // propagation by adding state counts of current state to all lower priority state that can
+    // transit to this current state
+    int index = 0;
+    while (true) {
```
dasahcc, line 336: It is not safe to use while (true). The only condition is index < stateList.size(), then let's put it here.

Or we can just use for loop to lopp state list.

pkuwm: +1. A for loop is preferred: index var will be a local/smaller scope variable in the for loop.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -51,7 +55,7 @@ public void process(ClusterEvent event) throws Exception {
     CurrentStateOutput currentStateOutput = event.getAttribute(AttributeName.CURRENT_STATE.name());
 
     MessageOutput selectedMessages = event.getAttribute(AttributeName.MESSAGES_SELECTED.name());
-    LogUtil.logDebug(logger, _eventId, String.format("selectedMessages is: %s", selectedMessages));
+    logger.debug("event info: {}, selectedMessages is: {}", _eventId, selectedMessages);
```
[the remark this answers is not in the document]
kaisun2000: LogUtil was used before. @pkuwm pointed out that there is runtime inefficiency of LogUtil. He prefers to use raw logger which does not have this inefficiency. That is why all of them changed to logger.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: I did removed dropped partition and error partition handling.

 Removed one more part
 ```
      Set<String> disabledInstances =
          cache.getDisabledInstancesForPartition(resourceName, partition.getPartitionName());
```
So the code is now like this:
```
      for (Message msg : partitionMessages) {
        if (!Message.MessageType.STATE_TRANSITION.name().equals(msg.getMsgType())) {
          logger.debug(
              "Event info: {} Message: {} not subject to throttle in resource: {} with type {}",
              _eventId, msg, resourceName, msg.getMsgType());
          continue;
        }

        boolean isUpward = !isDownwardTransition(stateModelDef, msg);

        // TODO: add disabled disabled instance special treatment

        String toState = msg.getToState();

        // TODO: dropped and Error state special treatment

        Integer minimumRequiredCount = partitionExpectedStateCounts.getOrDefault(toState, 0);
        Integer currentCount = partitionCurrentStateCounts.getOrDefault(toState, 0);

        //
        if (isUpward && (currentCount < minimumRequiredCount)) {
          recoveryMessages.add(msg);
          // It is critical to increase toState value by one here. For example, current state
          // of 3 replica in a partition is (M, O, O). Two messages here bringing up the two O to S.
          // In this case, the first O->S would be classified as recovery. Then this line would
          // increase the S state in partitionCurrentStateCounts value by one. Next O->S message
          // would be correctly marked as load message.
          partitionCurrentStateCounts.put(toState, currentCount + 1);
        } else {
          loadMessages.add(msg);
        }
      }

```

Let me know if there is anything else? Or we can comment on the new code.

File: helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java [hunk 40]
```diff
@@ -195,11 +202,52 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+    Set<Message> throttledRecoveryMessages = new HashSet<>();
+    logger.debug("Event info {}, applying recovery rebalance with resource {}", _eventId,
+        resourceName);
+    applyThrottling(resourceName, throttleController, stateModelDef, false, recoveryMessages,
+        throttledRecoveryMessages, StateTransitionThrottleConfig.RebalanceType.RECOVERY_BALANCE);
+
     // Step 4: sorts load message list and applies throttling
+    // TODO: calculate error-on-recovery downward threshold with complex backward compatibility next
+    // TODO: this can be done together with chargePendingMessage() where partitionsNeedRecovery is from
+    boolean onlyDownwardLoadBalance = partitionsWithErrorStateReplica.size() > 1;
```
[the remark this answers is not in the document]
kaisun2000: default to false.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,274 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+    LinkedHashMap<String, Integer> expectedStateCountMap =
+        stateModelDef.getStateCountMap(activeList.size(), replica);
+
+    expectedStateCountsOut.putAll(expectedStateCountMap);
+    propagateCountsTopDown(stateModelDef, expectedStateCountsOut);
+
+    return expectedStateCountsOut;
+  }
+
+  Map<Partition, Map<String, Integer>> calculateExpectedStateCounts(
```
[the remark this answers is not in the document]
kaisun2000: good point. Added private.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: value in the statePriorityMap is Integer, not int.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed to for loop.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: The original code has calculatedExpectedStateCounts and calculateCurrentStateCount in one function. Later, based on the review feedback from you and @pkuwm, we decided to separate this into two paths. That is why I would say let stick to current way and not moving back.

The second point is that the code getPartitionExpectedStateCounts and getPartitionCurrentStateCounts does not have much duplication, seem to me? Do you have a proposal of how to de-dupe the duplicate logic?

To be on the same page, let me list the code here. Maybe you can draw some code to illustrate the idea.

```
  private Map<String, Integer> getPartitionExpectedStateCounts(Partition partition,
      Map<String, List<String>> preferenceLists, StateModelDefinition stateModelDef,
      IdealState idealState, Set<String> enabledLiveInstance) {
    Map<String, Integer> expectedStateCountsOut = new HashMap<>();

    List<String> preferenceList = preferenceLists.get(partition.getPartitionName());
    if (preferenceList == null) {
      preferenceList = Collections.emptyList();
    }

    int replica =
        idealState.getMinActiveReplicas() <= 0 ? idealState.getReplicaCount(preferenceList.size())
            : idealState.getMinActiveReplicas();
    Set<String> activeList = new HashSet<>(preferenceList);
    activeList.retainAll(enabledLiveInstance);

    // For each state, check that this partition currently has the required number of that state as
    // required by StateModelDefinition.
    LinkedHashMap<String, Integer> expectedStateCountMap =
        stateModelDef.getStateCountMap(activeList.size(), replica);

    expectedStateCountsOut.putAll(expectedStateCountMap);
    propagateCountsTopDown(stateModelDef, expectedStateCountsOut);

    return expectedStateCountsOut;
  }
```

and 

```
private Map<String, Integer> getPartitionCurrentStateCounts(Map<String, String> currentStateMap,
      String resourceName, String partitionName, StateModelDefinition stateModelDef,
      ResourceControllerDataProvider cache) {
    // Current counts without disabled partitions or disabled instances
    Map<String, String> currentStateMapWithoutDisabled = new HashMap<>(currentStateMap);
    currentStateMapWithoutDisabled.keySet()
        .removeAll(cache.getDisabledInstancesForPartition(resourceName, partitionName));
    Map<String, Integer> currentStateCountsOut =
        StateModelDefinition.getStateCounts(currentStateMapWithoutDisabled);

    propagateCountsTopDown(stateModelDef, currentStateCountsOut);

    return currentStateCountsOut;
  }

```

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
dasahcc: But it is not necessary to explicitly do this.

kaisun2000: Now I see your point. Will change. Nowadays, with auto boxing, we don't need to do it this way explicitly.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: removed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -220,6 +271,274 @@ private MessageOutput throttlePerReplicaMessages(IdealState idealState,
...
+    Map<String, Integer> statePriorityMap = new HashMap<>();
+
+    // calculate rank of each state. Next use the rank to compare if a transition is upward or not
+    int rank = 0;
+    for (String state : stateList) {
+      statePriorityMap.put(state, Integer.valueOf(rank));
+      rank++;
+    }
+
+    // given a key state, find the set of states that can transit to this key state upwards.
```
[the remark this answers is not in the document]
kaisun2000: Let me move this logic to StateModelDefinition.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Let me move this logic to StateModelDefinition as utility function.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Yes, I think so. 

We can reason it this way:
The propagated current state counts only reflected what current state is. This line is more like saying, if this new message delivered and the new state change happens, what it will be. 

Let me what is your take here?

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: Let me make sure I understand.

1/ " feel you can use some variance of this method to calculate the propagate state count map"? This is to test if a message is downward. Seems to me, totally irrelevant to calculate the propagate state count map. Can you elaborate? Maybe some code. I totally don't get it.

2/ For sure I can move this one to stateModel definition. 

Let me sync with your over slack.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -270,4 +554,23 @@ public void setPriority(String priority) {
       }
     }
   }
+
+  private boolean isDownwardTransition(StateModelDefinition stateModelDef, Message message) {
```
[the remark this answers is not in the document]
kaisun2000: Based on previous discussion, will move this one to stateModelDefinition.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: Ok, for now let me resolve this issue.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed and moved this part to StateModelDefinition.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: changed and moved this part to StateModelDefinition.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
```diff
@@ -66,12 +70,11 @@ public void process(ClusterEvent event) throws Exception {
...
     ResourcesStateMap retracedResourceStateMap = new ResourcesStateMap();
-    MessageOutput output =
-        compute(event, resourceToRebalance, selectedMessages, retracedResourceStateMap);
+    MessageOutput output = compute(event, resourceToRebalance, currentStateOutput, selectedMessages,
+        retracedResourceStateMap);
 
     event.addAttribute(AttributeName.PER_REPLICA_THROTTLE_OUTPUT_MESSAGES.name(), output);
-    LogUtil.logDebug(logger, _eventId,
-        String.format("retraceResourceStateMap is: %s", retracedResourceStateMap));
+    logger.debug("Event info: {}, retraceResourceStateMap is: {}", retracedResourceStateMap);
```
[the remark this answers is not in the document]
kaisun2000: All changed to this way.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: moved to SateModelDefinition, adopt this pattern.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
kaisun2000: moved to StateModelDefinition.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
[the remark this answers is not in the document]
kaisun2000: changed to pattern 

```
if (logger.isDebugEnabled()) {
            LogUtil.logDebug(logger, _eventId, String.format())

```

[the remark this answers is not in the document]
jiajunwang: I remember that conversation, and I think Huizhi is right. But to solve the concern, why not modifying the LogUtil to fix the inefficiency? I checked the code, 8 lines of change. However, if we discard this util, then there would be much more code to change. Anyway, we don't need to include the util change in this PR. But please keep using the util and we can fix it in a separate PR.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: Let me take a look and reply with a proposal. Please don't resolve the conversion with an open discussion.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: I mean simply add one to the immediate toState is not propagation, right? Unless I take something wrong, but I think you should add all the further to states down the line here.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: StateModelDefinition.isDownwardTransition(message), just like that.

jiajunwang: There is a logic in the PR to find all the related state for propagation, right? I was just thinking if we can let these 2 methods share some common components to reduce duplication. To be more specific, we can add a method to StateModelDefinition which return all the down stream or upper stream states given a certain state. Then the propagation count calculation can use this method to find all the propagate target, and this downward transition method can use the same method to find if the to state is one of the down stream state of the from state.

helix-core/src/main/java/org/apache/helix/controller/stages/PerReplicaThrottleStage.java
jiajunwang: If you were referring to the very first version, then the original method is called "getPartitionExpectedAndCurrentStateCountMap". That method is definitely NOT what I'm trying to suggest you to do. That was one single method, but you were just putting all the logic together.

For this new code, what I thought is actually avoid duplicating the loops in the calculateXXXXCount() methods. However, as you also mentioned, you will have to pre-calculate the input with a loop for expected state map and current state map separately anyway, so it is not possible to save here.

The only possible way is to abstract the pre-calculate logic into other classes and then make it uniform calls. This requires a larger scope refactoring and may still not make the cut. So obviously, the cost is much higher than the gain. Let's just keep the current implemenation. I can help to have a try to see if it can be refined, later once this PR is in.

jiajunwang:
Close due to inactive.

PR #1178, opened 2020-07-26
Feature: ZooScalability Improvements

narendly:
### Issues

- [x] My PR addresses the following Helix issues and references them in the PR description:

Fixes #1162
Fixes #1175
Fixes #1176
Fixes #1177
Fixes #1188

### Description

- [x] Here are some details about my PR, including screenshots of any UI changes:

This PR contains a host of ZooScalability critical feature additions for usability. For details, please see the individual commit titles and messages.

Implement RoutingDataManager to replace HttpRoutingDataReader
Change interface for RoutingDataReader
Add HttpZkFallbackRoutingDataReader
Implement ZkRoutingDataReader
Modify realm-aware ZkClient and Helix API for configurable routing source
Make RoutingDataManager a pure Singleton with double-checked locking
Implement routing data update upon cache miss for FederatedZkClient 
Implement throttling for routing data update on cache miss 

### Tests

- [x] The following tests are written for this issue:

Various tests are added to ensure the features are stable and correct. Please see the individual commits.

- [x] The following is the result of the "mvn test" command on the appropriate module:

zookeeper-api:
```
[INFO] Tests run: 41, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 51.96 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 41, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
```

helix-core:
```
[ERROR] Failures: 
[ERROR]   TestCustomizedViewAggregation.testCustomizedViewAggregation:413->validateAggregationSnapshot:257 expected:<true> but was:<false>
[ERROR]   TestEnableCompression.testEnableCompressionResource:117 expected:<true> but was:<false>
[ERROR]   TestClusterStatusMonitorLifecycle.testClusterStatusMonitorLifecycle:290 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1172, Failures: 3, Errors: 0, Skipped: 0
```
Failures run individually:
```
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 45.75 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
```

[...]

PR #1627, opened 2021-02-02
Fix ZkAddress null case for TaskStateModelFactory

NealSun96:
### Issues

- [x] My PR addresses the following Helix issues and references them in the PR description:

Fix #1626 

### Description

- [x] Here are some details about my PR, including screenshots of any UI changes:

If zkAddress is null for the HelixManager passed to TaskStateModelFactory and multizk is not enabled, a NullPointerException will be raised. This PR fixes that by adding a null check. 

When https://github.com/apache/helix/pull/1183 was introduced and null checks were added, this section of code was in a feature branch, and was therefore not covered by the PR 1183. 

### Tests

- [x] The following tests are written for this issue:

(List the names of added unit/integration tests)

- The following is the result of the "mvn test" command on the appropriate module:

(If CI test fails due to known issue, please specify the issue and test PR locally. Then copy & paste the result of "mvn test" to here.)

### Documentation (Optional)

- In case of new functionality, my PR adds documentation in the following wiki page:

(Link the GitHub wiki you added)

### Commits

- My commits all reference appropriate Apache Helix GitHub issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Code Quality

- My diff has been formatted using helix-style.xml 
(helix-style-intellij.xml if IntelliJ IDE is used)

PR #1644, opened 2021-02-08
Fix TaskStateModelFactory no msds endpoint multizk situation

### LINKED ISSUE #1643: TaskStateModelFactory no msds endpoint multizk bug
### Describe the bug
When not specifying a zkaddress or enabling multizk, it's possible for customers to not specify an msds endpoint in the system properties. When that's the case, the federated zkclient created in TaskStateModelFactory will fail. 

### To Reproduce
Enble multizk or not provide a zkaddress to the manager, without specifying an msds endpoint in the system properties, create a TaskStateModelFactory. 

### Expected behavior
The msds endpoint contained in the manager's connection config should be respected.

NealSun96:
### Issues

Fixes #1643 

### Description

When not specifying a zkaddress or enabling multizk, it's possible for customers to not specify an msds endpoint in the system properties. When that's the case, the federated zkclient created in TaskStateModelFactory will fail. The msds endpoint contained in the manager's connection config should be respected.

When https://github.com/apache/helix/pull/1178 was introduced and the possibility of not including msds endpoint in system property was added, this section of code was in a feature branch, and was therefore not covered. This section of code was under old assumptions that became outdated during its development. 

### Tests

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/HelixManager.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManager.java
@@ -364,13 +364,6 @@ public interface HelixManager {
   String getMetadataStoreConnectionString();
 
   /**
-   * @return the RealmAwareZkConnectionConfig usd to create a realm aware zkClient
-   */
-  default RealmAwareZkClient.RealmAwareZkConnectionConfig getRealmAwareZkConnectionConfig() {
-    return null;
-  }
-
-  /**
    * Returns the instanceName used to connect to the cluster
    * @return the associated instance name
    */
--- a/helix-core/src/main/java/org/apache/helix/HelixManager.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManager.java
@@ -50,7 +50,6 @@ import org.apache.helix.participant.HelixStateMachineEngine;
 import org.apache.helix.participant.StateMachineEngine;
 import org.apache.helix.spectator.RoutingTableProvider;
 import org.apache.helix.store.zk.ZkHelixPropertyStore;
-import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 
 
--- a/helix-core/src/main/java/org/apache/helix/HelixManager.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManager.java
@@ -50,6 +50,7 @@ import org.apache.helix.participant.HelixStateMachineEngine;
 import org.apache.helix.participant.StateMachineEngine;
 import org.apache.helix.spectator.RoutingTableProvider;
 import org.apache.helix.store.zk.ZkHelixPropertyStore;
+import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 
 
@@ -361,6 +362,13 @@ public interface HelixManager {
    * @return a string used to connect to metadata store
    */
   String getMetadataStoreConnectionString();
+
+  /**
+   * @return the RealmAwareZkConnectionConfig usd to create a realm aware zkClient
+   */
+  default RealmAwareZkClient.RealmAwareZkConnectionConfig getRealmAwareZkConnectionConfig() {
+    return null;
+  }
 
   /**
    * Returns the instanceName used to connect to the cluster
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
@@ -692,6 +692,9 @@ public class ZKHelixManager implements HelixManager, IZkStateListener {
     return _zkAddress;
   }
 
+  /**
+   * @return the RealmAwareZkConnectionConfig usd to create a realm aware zkClient
+   */
   public RealmAwareZkClient.RealmAwareZkConnectionConfig getRealmAwareZkConnectionConfig() {
     return _realmAwareZkConnectionConfig;
   }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
@@ -692,6 +692,10 @@ public class ZKHelixManager implements HelixManager, IZkStateListener {
     return _zkAddress;
   }
 
+  public RealmAwareZkClient.RealmAwareZkConnectionConfig getRealmAwareZkConnectionConfig() {
+    return _realmAwareZkConnectionConfig;
+  }
+
   @Override
   public String getInstanceName() {
     return _instanceName;
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
@@ -693,7 +693,7 @@ public class ZKHelixManager implements HelixManager, IZkStateListener {
   }
 
   /**
-   * @return the RealmAwareZkConnectionConfig usd to create a realm aware zkClient
+   * @return the RealmAwareZkConnectionConfig used to create a realm aware ZkClient
    */
   public RealmAwareZkClient.RealmAwareZkConnectionConfig getRealmAwareZkConnectionConfig() {
     return _realmAwareZkConnectionConfig;
--- a/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
@@ -147,10 +147,16 @@ public class TaskStateModelFactory extends StateModelFactory<TaskStateModel> {
     if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED) || zkAddress == null) {
       String clusterName = manager.getClusterName();
       String shardingKey = HelixUtil.clusterNameToShardingKey(clusterName);
+      RealmAwareZkClient.RealmAwareZkConnectionConfig zkConnectionConfig =
+          manager.getRealmAwareZkConnectionConfig();
       RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig =
           new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder()
               .setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM)
-              .setZkRealmShardingKey(shardingKey).build();
+              .setZkRealmShardingKey(shardingKey).setRoutingDataSourceEndpoint(
+              (zkConnectionConfig != null) ? zkConnectionConfig.getRoutingDataSourceEndpoint()
+                  : null).setRoutingDataSourceType(
+              (zkConnectionConfig != null) ? zkConnectionConfig.getRoutingDataSourceType() : null)
+              .build();
       try {
         return new FederatedZkClient(connectionConfig, clientConfig);
       } catch (InvalidRoutingDataException | IllegalArgumentException e) {
--- a/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
@@ -140,8 +140,10 @@ public class TaskStateModelFactory extends StateModelFactory<TaskStateModel> {
     // can use manager's connection). We need to think about the right order and determine if we
     // want to enforce it, which may cause backward incompatibility.
     if (!(manager instanceof ZKHelixManager)) {
-      throw new IllegalArgumentException(
-          "Provided manager must be a ZKHelixManager for configurable thread pool.");
+      // TODO: None-ZKHelixManager cannot initialize this class. After interface rework of
+      // HelixManager, the initialization should be allowed.
+      throw new UnsupportedOperationException(
+          "Only ZKHelixManager is supported for configurable thread pool.");
     }
     RealmAwareZkClient.RealmAwareZkClientConfig clientConfig =
         new RealmAwareZkClient.RealmAwareZkClientConfig().setZkSerializer(new ZNRecordSerializer());
--- a/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
@@ -153,6 +153,11 @@ public class TaskStateModelFactory extends StateModelFactory<TaskStateModel> {
     if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED) || zkAddress == null) {
       RealmAwareZkClient.RealmAwareZkConnectionConfig zkConnectionConfig =
           ((ZKHelixManager) manager).getRealmAwareZkConnectionConfig();
+      // TODO: a fallback logic is created because it's possible for the ZKHelixManager to not
+      // have a connection config, since a connection config may be created during
+      // ZKHelixManager.connect(). This is the same problem as described earlier because connect()
+      // may happen before or after TaskStateModelFactory initialization. Clean this up after that
+      // problem is fixed.
       if (zkConnectionConfig == null) {
         String clusterName = manager.getClusterName();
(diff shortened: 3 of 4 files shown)


helix-core/src/main/java/org/apache/helix/HelixManager.java
```diff
@@ -362,6 +363,13 @@ void addExternalViewChangeListener(org.apache.helix.ExternalViewChangeListener l
    */
   String getMetadataStoreConnectionString();
 
+  /**
+   * @return the RealmAwareZkConnectionConfig usd to create a realm aware zkClient
+   */
+  default RealmAwareZkClient.RealmAwareZkConnectionConfig getRealmAwareZkConnectionConfig() {
```
narendly, line 369: A high-level comment:

- The overarching principle here is that the underlying ZkClients in the given HelixManager and the ZkClient we are using in TaskStateModelFactory have to have the same connection config.
- Instead of passing ZkConnectionConfig that returns a null as the default implementation, can we just have them share the underlying ZkClient?
- This way, we can guarantee that whatever HelixManager sees is exactly the same as what we see in TaskStateModelFactory, and we do not have to worry about managing the lifecycles of two separate ZkClients.

NealSun96: This implementation is a result of our (you and I mainly) discussion when Configurable Task Pool was first created, and I don't think the situation has changed. 

See line 137 in TaskStateModelFactory.java where the TODO is. The main problem is that some customers register TaskStateModel before connecting the manager, some do it after. For those that register before connecting, we have to create a separate zkClient. If we enforce the ordering, customers have to change participant code, which we decided against.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -147,10 +147,16 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
     if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED) || zkAddress == null) {
```
kaisun2000, line 147: This probably should be pointed out in https://github.com/apache/helix/pull/1627. 

if zkAddress is null, we use federatedZkClient, which is not exactly the same as dedicated ZkClient.

My question is more like this: 
Why do we think that MULTI_ZK_ENABLED and zkAddress == null belongs tot the same handling logic? 

Or put it another way, this condition, "if zkAddress is null for the HelixManager passed to TaskStateModelFactory and multizk is not enabled"? what should be the expected behavior?

kaisun2000 (changes requested):

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Hi Kai, please see https://github.com/apache/helix/pull/1183 for details. In short, there are customer requirements for this type of usage.

kaisun2000: Let me rephrase it this way, under this situation, which ZooKeeper should we use?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -147,10 +147,16 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
...
+          manager.getRealmAwareZkConnectionConfig();
       RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig =
           new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder()
               .setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM)
-              .setZkRealmShardingKey(shardingKey).build();
+              .setZkRealmShardingKey(shardingKey).setRoutingDataSourceEndpoint(
+              (zkConnectionConfig != null) ? zkConnectionConfig.getRoutingDataSourceEndpoint()
+                  : null).setRoutingDataSourceType(
+              (zkConnectionConfig != null) ? zkConnectionConfig.getRoutingDataSourceType() : null)
+              .build();
```
kaisun2000, line 159: How do we reason `setZkRealmShardingKey(shardingKey).setRoutingDataSourceEndpoint(...).setRoutingDataSourceType(...)`?

Note, this pattern is different from #1183 `configAccessor`.

kaisun2000 (changes requested):

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Under this condition, like all other conditions modified in #1183, simply means we're in multiZk mode. The ZooKeeper used is the one corresponding to the cluster (which is made into a sharding key).

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Please see PR description for why this is done. We need the routing data source endpoint from manager's connection config, because there're use cases where it's not provided among system properties.

kaisun2000: They why #1183 configAccessor does not have `setZkRealmShardingKey(shardingKey).setRoutingDataSourceEndpoint(...).setRoutingDataSourceType(...)`?

helix-core/src/main/java/org/apache/helix/HelixManager.java
narendly: Thanks for explaining the context. Yes, there are implementations that may not connect the manager prior to registering their task. Can we add a TODO or a separate issue tracking this? Now that we have different modes of underlying ZkClient instantiation, we need to be more careful about not creating multiple ZkClients, which could only add to the confusion for future developers.

For the extent of this PR, I'd be fine creating a one-off ZkClient using the connection config.

With that said, why does this need to be a default implementation? Nothing wrong with using the default interface, but in this case, the default behavior is not well-defined. I prefer not making this default.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -147,10 +147,16 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
     if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED) || zkAddress == null) {
       String clusterName = manager.getClusterName();
       String shardingKey = HelixUtil.clusterNameToShardingKey(clusterName);
+      RealmAwareZkClient.RealmAwareZkConnectionConfig zkConnectionConfig =
```
narendly, line 150: Could you explain why we don't just use zkConnectionConfig directly?

helix-core/src/main/java/org/apache/helix/HelixManager.java
NealSun96: A TODO has been added at the place where the zkClient is created, so I think we have that covered. 

About the default behavior, none of the test-only HelixManagers have the `_helixManagerProperty` field, so even if I add implementations, they will be `return null`.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Some of them (like ConfigAccessor) seem to be deprecated and for testing only, but some of them are not. I think this is a good question for @narendly : is it necessary to guard the other places with similar patterns, such as `ZkBucketDataAccessor`?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: You're right, we don't have to create a sharding key and rebuild config anymore. Changing.

jiajunwang:
An alternative way that I think would be cleaner and future proof, we make the create ZkClient method in the HelixManager public. Of course, we may need to refine this method since it is for now a private method only. But as long as it is done, the HelixManager users no longer need to care about the internal details. The manager object just generates a new client according to the initial config.

This is a proposal I thought for a while. There are several benefits,
1. Hide details, as mentioned above.
2. Potentially simplifying code. We can even enforce all the Helix user to create client through a HelixManager in the future. So this class becomes the only one who understands the necessary mess.
3. Potentially a better place (or another handy layer) to manage connections compared with the other static pool object.

helix-core/src/main/java/org/apache/helix/HelixManager.java
jiajunwang: I think we should not even add the method to the HelixManager interface. 1. it is ZK config, conceptually, HelixManager does not have to be implemented by a ZkHelixManager. 2. if we expose more internal implementation details from the public interface, we may be stuck with this design and cannot change anything in the future. This is not a good OO design.

Please see my proposal that I put here earlier.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
jiajunwang: I think the question is that what is the possible failure if we don't guard. But if you are considering if people will use it in the "wrong" way, then my comment is that when there is a way, someone will use it. And the bug will eventually be returned to us.
So as long as there is a possible hard failure case, please guard it. If it turns to be too complicated, then let's reconsider the overall design instead of patching here and there.

helix-core/src/main/java/org/apache/helix/HelixManager.java
NealSun96: Resolved with @jiajunwang offline: not a good OO design, so better solution: moving this to `ZkHelixManager` only.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Resolved offline with @jiajunwang: the design concern over ZooScalability is unrelated to this particular bug fix.

NealSun96:
> An alternative way that I think would be cleaner and future proof, we make the create ZkClient method in the HelixManager public. Of course, we may need to refine this method since it is for now a private method only. But as long as it is done, the HelixManager users no longer need to care about the internal details. The manager object just generates a new client according to the initial config.
> 
> This is a proposal I thought for a while. There are several benefits,
> 
> 1. Hide details, as mentioned above.
> 2. Potentially simplifying code. We can even enforce all the Helix user to create client through a HelixManager in the future. So this class becomes the only one who understands the necessary mess.
> 3. Potentially a better place (or another handy layer) to manage connections compared with the other static pool object.
> 
> I suggest we discuss this, and maybe other ideas, before push forward this PR. Since if we choose a different direction, then we shall not add more methods to the public interface.

Resolved with @jiajunwang offline, the design concern over ZooScalability is unrelated to this particular bug fix. We established that the difficulty of this code lies within the fact that TaskStateModelFactory may be initialized before ZkHelixManager connection, which is why we need to create a new ZkClient. @jiajunwang 's main concern is over the OO design, to which we have a solution now.

helix-core/src/main/java/org/apache/helix/HelixManager.java
```diff
@@ -50,6 +50,7 @@
 import org.apache.helix.participant.StateMachineEngine;
 import org.apache.helix.spectator.RoutingTableProvider;
 import org.apache.helix.store.zk.ZkHelixPropertyStore;
+import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
```
jiajunwang, line 53: Remove

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -140,19 +139,18 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
     // and some connect the manager before registering the state model factory (in which case we
     // can use manager's connection). We need to think about the right order and determine if we
     // want to enforce it, which may cause backward incompatibility.
+    if (!(manager instanceof ZKHelixManager)) {
```
jiajunwang, line 142: Do we want to add a TODO here to indicate that we might need to extend the support in the future?

jiajunwang (approved):

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: I actually want to get an opinion on this: is it really better to reject non-ZKHelixManagers, comparing to returning a null here, and let the util function to use the default pool size? This means open source customers who extend HelixManager literally cannot create a TaskStateModelFactory.

narendly: You can throw an [UnsupportedOperationException](https://docs.oracle.com/javase/6/docs/api/java/lang/UnsupportedOperationException.html) here and let the user know that only ZkHelixManager is currently supported. If we have different implementations of HelixManager down the road, you could add the support for them.

If there are custom implementations that extend HelixManager, they still can't create a ZkClient whether you have this change or not, so it's not making things any worse. Also, HelixManager interface already assumes a ZK-based implementation, so trying to accommodate some non-ZK HelixManager implementation out there that may or may not exist in this PR is a moot point. This is a good point to consider if we want to introduce a set of new interfaces that will replace existing data access interfaces though.

NealSun96: I'll add a TODO here. The fact that non-ZK HelixManager simply cannot create this factory is pretty absurd, especially given that we have a default value that can be used. 

UnsupportedOperationException is a good suggestion, changing.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -692,6 +692,13 @@ public String getMetadataStoreConnectionString() {
     return _zkAddress;
   }
 
+  /**
+   * @return the RealmAwareZkConnectionConfig usd to create a realm aware zkClient
```
narendly, line 696: Nit: Please fix the Javadoc

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -140,19 +140,28 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
...
-              .setZkRealmShardingKey(shardingKey).build();
+      RealmAwareZkClient.RealmAwareZkConnectionConfig zkConnectionConfig =
+          ((ZKHelixManager) manager).getRealmAwareZkConnectionConfig();
+      if (zkConnectionConfig == null) {
+        String clusterName = manager.getClusterName();
+        String shardingKey = HelixUtil.clusterNameToShardingKey(clusterName);
+        zkConnectionConfig = new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder()
+            .setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM)
+            .setZkRealmShardingKey(shardingKey).build();
+      }
```
narendly, line 162: How are we able to infer this from `zkConnectionConfig` being `null`? If the client creates a HelixManager to use it in a multi-zk environment, they should have provided a non-null `zkConnectionConfig` so we shouldn't expect it to be null, correct? If it is null, shouldn't we throw an exception instead?

NealSun96: Resolved offline with @narendly : this is necessary because, once again, no ordering is guaranteed between HelixManager.connect() and TaskStateModelFactory creation. It is entirely valid for a HelixManager to exist without zkConnectionConfig.

NealSun96:
Final commit message:
## Fix TaskStateModelFactory no msds endpoint multizk situation ##
Fixes a case where msds endpoint isn't provided in system property. This bug is caused by a timing issue where code in a feature branch isn't changed when another feature was introduced.

NealSun96:
Final commit message:
## Fix TaskStateModelFactory no msds endpoint multizk situation ##
Fix TaskStateModelFactory creation when msds endpoint is not included in system property. This bug was introduced because when the feature was introduced to allow no msds endpoint in system property, the code in this PR was still in a feature branch and didn't get updated properly.

Issue #1642, opened 2021-02-08
Too many thread causing memory issue when enable batch mode in CallbackHandler


xyuanlu:
### Describe the bug
After enabling batch mode for prod, we notice the thread count is high. 
The reason is that when enable batch mode, each callBackHandler has a dedicated thread (CallBackProcessor) handles all callbacks. Originally then batch mode disabled, the _eventThread in zlClient handles all callbacks. 
Proposed solution : A thread pool for each zkHelixManager to limit the thread number

PR #1645, opened 2021-02-09
Use thread pool for batched call back handling events

xyuanlu:
### Issues

#1642 Too many thread causing memory issue when enable batch mode in CallbackHandler 

### Description

After enabling batch mode for prod, we notice the thread count is high.
The reason is that when enable batch mode, each callBackHandler has a dedicated thread (CallBackProcessor) handles all callbacks. Originally then batch mode disabled, the _eventThread in zlClient handles all callbacks.

### Tests

NA

```
[INFO] Tests run: 1262, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 6,064.174 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 1262, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:41 h
[INFO] Finished at: 2021-02-22T22:20:42-08:00
```
### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
@@ -46,6 +46,7 @@ public class CallbackEventExecutor {
   private final HelixManager _manager;
   private Future _futureCallBackProcessEvent = null;
   private ThreadPoolExecutor _threadPoolExecutor;
+  private boolean _isShutdown = false;
 
   public CallbackEventExecutor(HelixManager manager) {
     _callBackEventQueue = new DedupEventBlockingQueue<>();
@@ -80,9 +81,12 @@ public class CallbackEventExecutor {
     }
   }
 
-  public void submitEventToEvecutor(NotificationContext.Type eventType, NotificationContext event,
+  public void submitEventToExecutor(NotificationContext.Type eventType, NotificationContext event,
       CallbackHandler handler) {
     synchronized (_callBackEventQueue) {
+      if (_isShutdown) {
+        logger.error("Failed to process callback. CallbackEventExecutor is already shut down.");
+      }
       if (_futureCallBackProcessEvent == null || _futureCallBackProcessEvent.isDone()) {
         _futureCallBackProcessEvent =
             _threadPoolExecutor.submit(new CallbackProcessor(handler, event));
@@ -116,6 +120,7 @@ public class CallbackEventExecutor {
   }
 
   public void unregisterFromFactory() {
+    _isShutdown = true;
     reset();
     CallbackEventThreadPoolFactory.unregisterEventProcessor(_manager.hashCode());
     _threadPoolExecutor = null;
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
@@ -19,7 +19,6 @@ package org.apache.helix.manager.zk;
  * under the License.
  */
 
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
 
@@ -46,7 +45,7 @@ public class CallbackEventExecutor {
       _callBackEventQueue;
   private final HelixManager _manager;
   private Future _futureCallBackProcessEvent = null;
-  private ExecutorService _threadPoolExecutor;
+  private ThreadPoolExecutor _threadPoolExecutor;
   private boolean _isShutdown = false;
 
   public CallbackEventExecutor(HelixManager manager) {
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
@@ -116,6 +116,7 @@ public class CallbackEventExecutor {
   }
 
   public void unregisterFromFactory() {
+    reset();
     CallbackEventTPFactory.unregisterEventProcessor(_manager.hashCode());
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
@@ -60,7 +60,7 @@ public class CallbackEventExecutor {
     private final NotificationContext _event;
 
     public CallbackProcessor(CallbackHandler handler, NotificationContext event) {
-      _processorName = _manager.getClusterName() + "CallbackProcessor@" + Integer
+      _processorName = _manager.getClusterName() + "-CallbackProcessor@" + Integer
           .toHexString(handler.hashCode());
       _handler = handler;
       _event = event;
@@ -104,7 +104,8 @@ public class CallbackEventExecutor {
           _futureCallBackProcessEvent =
               _threadPoolExecutor.submit(new CallbackProcessor(handler, event));
         } catch (InterruptedException e) {
-          e.printStackTrace();
+          logger
+              .error("Error when submitting pending HandleCallBackEvent to manager thread pool", e);
         }
       }
     }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
@@ -38,7 +38,7 @@ import org.slf4j.LoggerFactory;
  * 2. The previous call back event handling process is finished in thread pool.
  */
 
-public class CallbackEventTPExecutor {
+public class CallbackEventExecutor {
   private static Logger logger = LoggerFactory.getLogger(CallbackHandler.class);
 
   private DedupEventBlockingQueue<NotificationContext.Type, NotificationContext>
@@ -47,10 +47,10 @@ public class CallbackEventTPExecutor {
   private Future _futureCallBackProcessEvent = null;
   private final ThreadPoolExecutor _threadPoolExecutor;
 
-  public CallbackEventTPExecutor(HelixManager manager) {
+  public CallbackEventExecutor(HelixManager manager) {
     _callBackEventQueue = new DedupEventBlockingQueue<>();
     _manager = manager;
-    _threadPoolExecutor = CallbackEventTPFactory.getThreadPool(manager.hashCode());
+    _threadPoolExecutor = CallbackEventTPFactory.getOrCreateThreadPool(manager.hashCode());
   }
 
   class CallbackProcessor implements Runnable {
@@ -113,5 +113,9 @@ public class CallbackEventTPExecutor {
         _futureCallBackProcessEvent.cancel(false);
       }
     }
+  }
+
+  public void unregisterFromFactory() {
+    CallbackEventTPFactory.unregisterEventProcessor(_manager.hashCode());
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
@@ -45,18 +45,18 @@ public class CallbackEventExecutor {
       _callBackEventQueue;
   private final HelixManager _manager;
   private Future _futureCallBackProcessEvent = null;
-  private final ThreadPoolExecutor _threadPoolExecutor;
+  private ThreadPoolExecutor _threadPoolExecutor;
 
   public CallbackEventExecutor(HelixManager manager) {
     _callBackEventQueue = new DedupEventBlockingQueue<>();
     _manager = manager;
-    _threadPoolExecutor = CallbackEventTPFactory.getOrCreateThreadPool(manager.hashCode());
+    _threadPoolExecutor = CallbackEventThreadPoolFactory.getOrCreateThreadPool(manager.hashCode());
   }
 
   class CallbackProcessor implements Runnable {
-    private CallbackHandler _handler;
-    protected String _processorName;
-    NotificationContext _event;
+    private final CallbackHandler _handler;
+    protected final String _processorName;
+    private final NotificationContext _event;
 
     public CallbackProcessor(CallbackHandler handler, NotificationContext event) {
       _processorName = _manager.getClusterName() + "CallbackProcessor@" + Integer
@@ -72,7 +72,7 @@ public class CallbackEventExecutor {
       } catch (ZkInterruptedException e) {
         logger.warn(_processorName + " thread caught a ZK connection interrupt", e);
       } catch (ThreadDeath death) {
-        throw death;
+        logger.error(_processorName + " thread dead " + _processorName, death);
       } catch (Throwable t) {
         logger.error(_processorName + " thread failed while running " + _processorName, t);
       }
@@ -117,6 +117,7 @@ public class CallbackEventExecutor {
 
   public void unregisterFromFactory() {
     reset();
-    CallbackEventTPFactory.unregisterEventProcessor(_manager.hashCode());
+    CallbackEventThreadPoolFactory.unregisterEventProcessor(_manager.hashCode());
+    _threadPoolExecutor = null;
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
@@ -19,6 +19,7 @@ package org.apache.helix.manager.zk;
  * under the License.
  */
 
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
 
@@ -45,7 +46,7 @@ public class CallbackEventExecutor {
       _callBackEventQueue;
   private final HelixManager _manager;
   private Future _futureCallBackProcessEvent = null;
-  private ThreadPoolExecutor _threadPoolExecutor;
+  private ExecutorService _threadPoolExecutor;
   private boolean _isShutdown = false;
 
   public CallbackEventExecutor(HelixManager manager) {
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPExecutor.java
@@ -1,35 +1,34 @@
 package org.apache.helix.manager.zk;
 
+    /*
+     * Licensed to the Apache Software Foundation (ASF) under one
+     * or more contributor license agreements.  See the NOTICE file
+     * distributed with this work for additional information
+     * regarding copyright ownership.  The ASF licenses this file
+     * to you under the Apache License, Version 2.0 (the
+     * "License"); you may not use this file except in compliance
+     * with the License.  You may obtain a copy of the License at
+     *
+     *     http://www.apache.org/licenses/LICENSE-2.0
+     *
+     * Unless required by applicable law or agreed to in writing,
+     * software distributed under the License is distributed on an
+     * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+     * KIND, either express or implied.  See the License for the
+     * specific language governing permissions and limitations
+     * under the License.
+     */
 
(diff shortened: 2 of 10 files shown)


lei-xia:
I think we have a pending discussion on how to utilize the threadpool, right?  Have we reached any agree on the design?

xyuanlu:
> I think we have a pending discussion on how to utilize the threadpool, right? Have we reached any agree on the design?

Thanks for the reply. Yea I posted the PR so is easier to get comments and advices. I will schedule an offline discuss as well.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -136,32 +135,13 @@
   private HelixCallbackMonitor _monitor;
 
   // TODO: make this be per _manager or per _listener instaed of per callbackHandler -- Lei
-  private AtomicReference<CallbackProcessor> _batchCallbackProcessorRef = new AtomicReference<>();
+  private AtomicReference<CallbackEventTPExecutor> _batchCallbackExecutorrRef = new AtomicReference<>();
```
jiajunwang, line 138: typo? double 'r'

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -136,32 +135,13 @@
   private HelixCallbackMonitor _monitor;
 
   // TODO: make this be per _manager or per _listener instaed of per callbackHandler -- Lei
```
jiajunwang, line 137: I think this TODO is no longer needed, but please correct me if I am wrong.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackHandler.java
```diff
@@ -660,15 +641,12 @@ public void init() {
...
-      if (callbackProcessor != null) {
-        callbackProcessor.resetEventQueue();
+      CallbackEventTPExecutor callbackTFExecutor = _batchCallbackExecutorrRef.get();
+      if (callbackTFExecutor != null) {
+        callbackTFExecutor.reset();
       } else {
-        callbackProcessor = new CallbackProcessor(this);
-        callbackProcessor.start();
-        if (!_batchCallbackProcessorRef.compareAndSet(null, callbackProcessor)) {
-          callbackProcessor.shutdown();
```
jiajunwang, line 670: We don't need to shutdown the thread pool anymore?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
```diff
@@ -0,0 +1,76 @@
...
+    // first check if the key is already in the map
+    if (_managerToCallBackThreadPoolMap.containsKey(mapHash)) {
+      // downgrade to read lock
+      _lock.readLock().lock();
+      result = _managerToCallBackThreadPoolMap.get(mapHash);
+      _lock.readLock().unlock();
+    } else {
+      ThreadFactory namedThreadFactory = new ThreadFactoryBuilder()
+          .setNameFormat(String.format("CallbackHandlerExecutorService - %s ", mapHash)).build();
+      result = new ThreadPoolExecutor(5, 5, 1, TimeUnit.SECONDS, new LinkedBlockingQueue<>(),
```
jiajunwang, line 68: When is the newly created executor shutdown?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// returns a thread pool object given a HelixManager identifier
```
jiajunwang, line 24: Nit, the class comment should be put below ahead of the class name.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
```diff
@@ -0,0 +1,76 @@
...
+      result = _managerToCallBackThreadPoolMap.get(mapHash);
+      _lock.readLock().unlock();
+    } else {
+      _lock.readLock().unlock();
+      result = getAndCreateThreadPool(mapHash);
+    }
+    return result;
+  }
+
+  public static ThreadPoolExecutor getAndCreateThreadPool(int mapHash) {
```
jiajunwang, line 56: Why we have 2 public methods that are doing. almost the same thing?
I would suggest having one for getOrCreate(), another one for create() with writelock.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
+
+public class CallbackEventTPFactory {
```
jiajunwang, line 38: Just curious, what is the benefit of having this additional layer of the factory? Considering the threadpool is one-one mapping to the executor, the factory does not help with any additional sharing or reusing logic. However, it introduces considerable cleanup overhead. Note that I didn't see any logic in your current version to cleanup the threadpools.

xyuanlu: TFTR. The threadpool is one-one mapping to HelixManager, not executor. I think having a factory will provide a layer of abstraction for thread pool creation, so each CallbackEventExecutor don't need to know if the thread pool for the helixManager is already created. CallbackEventExecutor could just call `getOrCreateThreadPool ` in its constructor.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPExecutor.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Each batch mode is enabled, each CallbackHandler has a CallbackEventTPExecutor.
+ * It has a dedupe queue for pending call back event. Pending call back event will
+ * be submitted to a thread pool one at a time when
+ * 1. This is the first ever call back event for the callBackHandler, or
+ * 2. The previous call back event handling process is finished in thread pool.
+ */
+
+public class CallbackEventTPExecutor {
```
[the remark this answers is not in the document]
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
jiajunwang: I see. Makes sense. Thanks for the explanation.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
```diff
@@ -0,0 +1,96 @@
...
+    if (_managerToCallBackThreadPoolMap.containsKey(managerHash)) {
+      // downgrade to read lock since we dont need to modify the map
+      _lock.readLock().lock();
+      result = _managerToCallBackThreadPoolMap.get(managerHash);
+      _lock.readLock().unlock();
+    } else {
+      ThreadFactory namedThreadFactory = new ThreadFactoryBuilder()
+          .setNameFormat(String.format("CallbackHandlerExecutorService - %s ", managerHash))
+          .build();
+      result = new ThreadPoolExecutor(5, 5, 1, TimeUnit.SECONDS, new LinkedBlockingQueue<>(),
```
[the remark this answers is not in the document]
jiajunwang: On my second thought, I think we should not keep so many threads. Otherwise, if the manager has only one callback for a path, we will still create so many threads. This does not make sense.
One reasonable way is to set the core thread count same as the required callback count. This can be derived from the HelixMananger listener fields. However, there seems to be no easy way to get the number directly even we have the manager object.

Let's think more about it.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Override
+    public void run() {
+      try {
+        _handler.invoke(_event);
+      } catch (ZkInterruptedException e) {
+        logger.warn(_processorName + " thread caught a ZK connection interrupt", e);
+      } catch (ThreadDeath death) {
+        throw death;
```
jiajunwang, line 75: I think we need to "submitPendingHandleCallBackEventToManagerThreadPool" no matter what is the exception in the previous execution. So it should either be put in the finally block or this throw should be an error log too.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
```diff
@@ -0,0 +1,122 @@
...
+      _callBackEventQueue.clear();
+      if (_futureCallBackProcessEvent != null) {
+        _futureCallBackProcessEvent.cancel(false);
+      }
+    }
+  }
+
+  public void unregisterFromFactory() {
+    reset();
+    CallbackEventTPFactory.unregisterEventProcessor(_manager.hashCode());
```
jiajunwang, line 120: You will need to set _threadPoolExecutor = null too.

Moreover, to ensure it is safe after unregister, the executor shall reject any new task submission after this call.

jiajunwang (commented):
Highlevelly it looks good. However, due to the multiple layers design, I think some details need more time to polish given the complicity. Please check if my comments make sense : )

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
```diff
@@ -0,0 +1,96 @@
...
+    return result;
+  }
+
+  private static ThreadPoolExecutor getOrCreateThreadPoolHelper(int managerHash) {
+    ThreadPoolExecutor result;
+    _lock.writeLock().lock();
+    // first check if the key is already in the map
+    if (_managerToCallBackThreadPoolMap.containsKey(managerHash)) {
+      // downgrade to read lock since we dont need to modify the map
+      _lock.readLock().lock();
```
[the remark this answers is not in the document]
xyuanlu: This will downgrade the lock to a read lock so that other read request won't be blocked.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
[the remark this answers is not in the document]
xyuanlu: The reason why I am using ThreadPoolExecutor(5, 5, 1, TimeUnit.SECONDS,...) is exactly the same. Since we set `allowCoreThreadTimeOut ` to true, thread will be terminated if not used after 1 sec.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
```diff
@@ -0,0 +1,122 @@
...
+  public CallbackEventExecutor(HelixManager manager) {
+    _callBackEventQueue = new DedupEventBlockingQueue<>();
+    _manager = manager;
+    _threadPoolExecutor = CallbackEventTPFactory.getOrCreateThreadPool(manager.hashCode());
+  }
+
+  class CallbackProcessor implements Runnable {
+    private CallbackHandler _handler;
+    protected String _processorName;
+    NotificationContext _event;
```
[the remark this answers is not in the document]
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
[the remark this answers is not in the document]
jiajunwang: I thought that you need to unlock the write lock to achieve what you said. It is not doing it automatically. Please correct me if I am wrong.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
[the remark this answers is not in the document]
jiajunwang: " when the number of threads is greater than the core, this is the maximum time that excess idle threads will wait for new tasks before terminating."

Since the core thread is also 5, it will keep 5 live threads in the pool.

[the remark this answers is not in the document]
jiajunwang: Seems to be a very strange design, why not just set the core thread to be 0? I'm just curious. The logic seems to be working.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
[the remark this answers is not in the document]
xyuanlu: Yea you are correct. Updated.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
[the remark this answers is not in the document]
xyuanlu: From java Doc:
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html#allowsCoreThreadTimeOut()
```If there are more than corePoolSize but less than maximumPoolSize threads running, a new thread will be created only if the queue is full.
```
If have an unbounded queue when core = 1, and max = 5, all new tasks (except the first one) will be pushed to the queue and thread count is always 1. I think the only way to have a dynamically increasing/decreasing thread pool with unbounded queue is having a coreThread==maxThread==X and timeToLive = Y.
However, the thread pool will have 0 idle threads after the timeout has expired. So I decided to change back to a fixSize thread pool.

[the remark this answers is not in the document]
jiajunwang: I cannot believe that I never noticed this before. Thanks for raising this up. I think the original solution will work. But some details need another thought to refine.

Alternatively, we can also do as this post said, https://stackoverflow.com/questions/15485840/threadpoolexecutor-with-unbounded-queue-not-creating-new-threads
I have tested the solution and it works great. Admittedly, it is more complex. So up to you : )

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
```diff
@@ -0,0 +1,128 @@
...
+    _threadPoolExecutor = CallbackEventThreadPoolFactory.getOrCreateThreadPool(manager.hashCode());
+  }
+
+  class CallbackProcessor implements Runnable {
+    private final CallbackHandler _handler;
+    protected final String _processorName;
+    private final NotificationContext _event;
+
+    public CallbackProcessor(CallbackHandler handler, NotificationContext event) {
+      _processorName = _manager.getClusterName() + "CallbackProcessor@" + Integer
```
jiajunwang, line 63: Add a "-" before CallbackProcessor will make the output looks better, I guess.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventExecutor.java
```diff
@@ -0,0 +1,128 @@
...
+
+  private void submitPendingHandleCallBackEventToManagerThreadPool(CallbackHandler handler) {
+    synchronized (_callBackEventQueue) {
+      if (_callBackEventQueue.size() != 0) {
+        try {
+          NotificationContext event = _callBackEventQueue.take();
+          _futureCallBackProcessEvent =
+              _threadPoolExecutor.submit(new CallbackProcessor(handler, event));
+        } catch (InterruptedException e) {
+          e.printStackTrace();
```
jiajunwang, line 107: Log the error?

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventThreadPoolFactory.java
```diff
@@ -0,0 +1,98 @@
...
+    _lock.readLock().unlock();
+    if (result == null) {
+      result = getOrCreateThreadPoolHelper(hash);
+    }
+    return result;
+  }
+
+  private static ThreadPoolExecutor getOrCreateThreadPoolHelper(int hash) {
+    ThreadPoolExecutor result;
+    _lock.writeLock().lock();
```
jiajunwang, line 64: Better to put the unlock call in a final block to ensure safety.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventThreadPoolFactory.java
```diff
@@ -0,0 +1,98 @@
...
+    if (_managerToCallBackThreadPoolMap.containsKey(hash)) {
+      // downgrade to read lock since we dont need to modify the map
+      _lock.readLock().lock();
+      _lock.writeLock().unlock();
+      result = _managerToCallBackThreadPoolMap.get(hash);
+      _lock.readLock().unlock();
+    } else {
+      ThreadFactory namedThreadFactory = new ThreadFactoryBuilder()
+          .setNameFormat(String.format("CallbackHandlerExecutorService - %s ", hash)).build();
+      result = new ThreadPoolExecutor(CALLBACK_EVENT_THREAD_POOL_SIZE,
```
jiajunwang, line 75: Did you remove the allow core thread to close option? In this case, it is just a fixed size threadpool, right?
I'm fine with both ways. But the current one can be simplified if I understand the intention correctly.

jiajunwang (commented):
Looks good in general. I left some minor comments. Please address.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventThreadPoolFactory.java
xyuanlu: Yea I forgot to add it back..

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventThreadPoolFactory.java
xyuanlu: TFTR. 
Please correct if I am wrong. I think putting this unlock call to final block wont down grade the lock to a read lock. This need to happen before the 
`result = _managerToCallBackThreadPoolMap.get(hash);`

jiajunwang: Not the expected unlock, that one is fine.
What I meant is that, in general, if an exception is thrown unexpectedly, then try-catch-finally will help to ensure the lock is released instead of being held forever. An alternative way is to use some autoclosable lock design. In your current code, one example is that if the new threadpoolexecutor fails, then the writelock won't be released.

xyuanlu: TFTR. Good point. Updated.

helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventTPFactory.java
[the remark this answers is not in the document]
xyuanlu: From our offline discussion. I will keep the same but update the TTL and thread count.

jiajunwang (approved):
LGTM!

xyuanlu:
The change is ready to be merged in. Approved by @jiajunwang 

Final commit message:
Use thread pool for batched call back handling events

Originally when enable batch mode, each callBackHandler has a dedicated thread (CallBackProcessor) handles all callbacks.  This change adds a fixed sized thread pool per each Helix manager handling all callback events. Thus the total events count when enabling batch mode would be low.

PR #1658, opened 2021-02-26
New PartitionMovementConstraint and BaselineInfluenceConstraint for Waged

### LINKED ISSUE #1614: Baseline Movement Constraint
The current Waged implementation combines baseline and bestPossible movement controls in the same constraint: `PartitionMovementConstraint`. The constraint's weight (and `LESS_MOVEMENT` preference) contributes to both "how much baseline moves per rebalancer iteration" and "how much bestPossible moves per rebalancer iteration". These should ideally be separated for flexible control. 

Implement 2 constraints that control baseline and bestPossible moves separately. 

**Additional context**

NealSun96:
### Issues

Fixes #1614 

### Description

The current implementation offers a unified control for both baseline movements and bestPossible movements, but it's growing more apparent that we need more flexible controls. For example, allowing more aggressive convergence to baseline movements (higher weight) is now combined with limiting movements based on best possible states (higher weight), but these two effects are contradicting. Baseline convergence should be a "evenness" focused constraint, while limiting movements based on best possible should be a "less movement" constraint.

This PR splits `PartitionMovementConstraint` into separate constraints that control baseline convergence and best possible movements respectively. Those factors allow an option to, for example, let baseline completely override best possible and force a baseline convergence, or to let baseline only influence the decision similar to the usual Waged configurations. 

### Tests

```
Changed TestPartitionMovementConstraint; Added testGetRebalancePreferenceMissingKey, testSetRebalancePreferenceMissingKey
```

```
[INFO] Tests run: 1265, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,209.189 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:26 h
[INFO] Finished at: 2021-03-05T18:14:05-08:00
```

### Documentation (Optional)

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/AbstractPartitionMovementConstraint.java
```diff
@@ -0,0 +1,86 @@
...
+   */
+  @Override
+  protected abstract double getAssignmentScore(AssignableNode node, AssignableReplica replica,
+      ClusterContext clusterContext);
+
+  /**
+   * @return The scale factor to adjust score when the proposed allocation partially matches the
+   * assignment plan but will require a state transition (with partition movement).
+   */
+  protected abstract double getStateTransitionCostFactor();
```
jiajunwang, line 59: It seems that both of the child classes have the same implementation. We don't need this abstract method now.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/BaselineInfluenceConstraint.java
```diff
@@ -0,0 +1,54 @@
...
+ * This constraint promotes movements for evenness. If best possible doesn't exist, baseline will be
+ * used to restrict movements, so this constraint should give no score in that case.
+ */
+public class BaselineInfluenceConstraint extends AbstractPartitionMovementConstraint {
+  @Override
+  protected double getAssignmentScore(AssignableNode node, AssignableReplica replica,
+      ClusterContext clusterContext) {
+    Map<String, String> bestPossibleAssignment =
+        getStateMap(replica, clusterContext.getBestPossibleAssignment());
+    Map<String, String> baselineAssignment =
```
jiajunwang, line 41: nit, put this section after the bestPossibleAssignment check to avoid unnecessary calculation.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
```diff
@@ -71,48 +43,13 @@ protected double getAssignmentScore(AssignableNode node, AssignableReplica repli
     String state = replica.getReplicaState();
 
     if (bestPossibleAssignment.isEmpty()) {
-      // If bestPossibleAssignment of the replica is empty, indicating this is a new replica.
```
jiajunwang, line 74: I think we should keep this comment. Or it would be hard to explain why we are using baselineAssignment as the input here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
```diff
@@ -19,48 +19,20 @@
...
-  // plan but will require a state transition (with partition movement).
-  // TODO: these factors will be tuned based on user's preference
-  private static final double STATE_TRANSITION_COST_FACTOR = 0.5;
-  private static final double MOVEMENT_COST_FACTOR = 0.25;
-
-  PartitionMovementConstraint() {
-    super(MAX_SCORE, MIN_SCORE);
-  }
-
-  /**
```
jiajunwang, line 53: Please refine the comment instead of removing it.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
```diff
@@ -19,48 +19,20 @@
...
-   * @return 1 if the proposed assignment completely matches the previous best possible assignment
-   *         (or baseline assignment if the replica is newly added).
-   *         STATE_TRANSITION_COST_FACTOR if the proposed assignment's allocation matches the
-   *         previous Best Possible assignment (or baseline assignment if the replica is newly
-   *         added) but state does not match.
-   *         MOVEMENT_COST_FACTOR if the proposed assignment's allocation matches the baseline
-   *         assignment only, but not matches the previous best possible assignment.
-   *         0 if the proposed assignment is a pure random movement.
-   */
-  @Override
```
jiajunwang, line 63: Why remove?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/BaselineInfluenceConstraint.java
```diff
@@ -0,0 +1,54 @@
...
+    if (bestPossibleAssignment.isEmpty()) {
+      return getMinScore();
+    }
+    return calculateAssignmentScore(node.getInstanceName(), replica.getReplicaState(),
+        baselineAssignment);
+  }
+
+  @Override
+  protected double getStateTransitionCostFactor() {
+    return MAX_SCORE / 2;
```
jiajunwang, line 52: We shall return STATE_TRANSITION_COST_FACTOR here.
1. it should return factor not the actual score.
2. it shall not be hardcoded.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/AbstractPartitionMovementConstraint.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return assignment.get(resourceName).getReplicaMap(new Partition(partitionName));
+  }
+
+  protected double calculateAssignmentScore(String nodeName, String state,
+      Map<String, String> instanceToStateMap) {
+    if (instanceToStateMap.containsKey(nodeName)) {
+      return state.equals(instanceToStateMap.get(nodeName)) ?
+          MAX_SCORE : // if state matches, no state transition required for the proposed assignment
+          getStateTransitionCostFactor(); // if state does not match, then the proposed assignment
```
jiajunwang, line 76: Should be (MAX_SCORE - MIN_SCORE) * getStateTransitionCostFactor() + MIN_SCORE?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
```diff
@@ -19,48 +19,20 @@
...
- * The cost is evaluated based on the difference between the old assignment and the new assignment.
- * In detail, we consider the following two previous assignments as the base.
- * - Baseline assignment that is calculated regardless of the node state (online/offline).
- * - Previous Best Possible assignment.
- * Any change to these two assignments will increase the partition movements cost, so that the
- * evaluated score will become lower.
+ * Evaluate the proposed assignment according to the potential partition movements cost based on
+ * the previous best possible assignment.
+ * The previous best possible assignment is the sole reference, and if it's missing, use baseline
+ * assignment instead.
```
jiajunwang, line 33: "if it's missing, use baseline assignment instead.", this part deserves more description. At least mention that we are doing so because it is assumed to be a new resource that has never been allocated before.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
```diff
@@ -122,13 +122,6 @@ public void testGetRebalancePreferenceDefault() {
     ClusterConfig testConfig = new ClusterConfig("testId");
     Assert.assertEquals(testConfig.getGlobalRebalancePreference(),
         ClusterConfig.DEFAULT_GLOBAL_REBALANCE_PREFERENCE);
-
-    Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preference = new HashMap<>();
```
jiajunwang, line 126: What would be the expected preference in this condition then?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
```diff
@@ -39,6 +38,7 @@
     {
       // The default setting
       put(PartitionMovementConstraint.class.getSimpleName(), 2f);
+      put(BaselineInfluenceConstraint.class.getSimpleName(), 0.5f);
       put(InstancePartitionsCountConstraint.class.getSimpleName(), 1f);
       put(ResourcePartitionAntiAffinityConstraint.class.getSimpleName(), 1f);
       put(ResourceTopStateAntiAffinityConstraint.class.getSimpleName(), 3f);
```
jiajunwang, line 44: nit, could you please merge the most recent code about top state usage constraint then we can review the updated code.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -160,7 +161,8 @@
       DEFAULT_GLOBAL_REBALANCE_PREFERENCE =
```
jiajunwang, line 161: I think the change does make sense. But is this value still being used anymore?

Maybe we can change the logic to, if nothing is configured, then apply this default value.
If partially missing, then put whatever user set in the ClusterConfig but leave the processing logic to the rebalancer like what you are doing right now.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
```diff
@@ -61,21 +61,34 @@ public static RebalanceAlgorithm getInstance(
             new ReplicaActivateConstraint(), new NodeMaxPartitionLimitConstraint(),
             new ValidGroupTagConstraint(), new SamePartitionOnInstanceConstraint());
 
-    int evennessPreference =
-        preferences.getOrDefault(ClusterConfig.GlobalRebalancePreferenceKey.EVENNESS, 1);
-    int movementPreference =
-        preferences.getOrDefault(ClusterConfig.GlobalRebalancePreferenceKey.LESS_MOVEMENT, 1);
+    int evennessPreference = 1;
```
jiajunwang, line 64: Refer to the value in DEFAULT_GLOBAL_REBALANCE_PREFERENCE?
Please avoid hardcode.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
```diff
@@ -61,21 +61,34 @@ public static RebalanceAlgorithm getInstance(
...
+      evennessPreference = preferences.get(ClusterConfig.GlobalRebalancePreferenceKey.EVENNESS);
+      movementPreference =
+          preferences.get(ClusterConfig.GlobalRebalancePreferenceKey.LESS_MOVEMENT);
+    }
+
+    boolean forceBaselineConverge = preferences
+        .getOrDefault(ClusterConfig.GlobalRebalancePreferenceKey.FORCE_BASELINE_CONVERGE, 0)
+        > 0;
+    if (forceBaselineConverge) {
+      MODEL.put(PartitionMovementConstraint.class.getSimpleName(), 0.001f);
```
jiajunwang, line 77: I think it would be better not to touch this weight. Unlike BaselineInfluenceConstraint, PartitionMovementConstraint is used in the baseline calculation too. So if you modify the weight here, then the baseline will be over evened. This may cause a large partition shuffling.

What we can do should be modifying BaselineInfluenceConstraint **after** the preference has been applied to the the constraint weight.

jiajunwang: And I guess set its weight to (10000 * PartitionMovementConstraint weight * LESS_MOVEMENT preference factor) should be enough and 0.001f maybe not required.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
```diff
@@ -61,21 +61,34 @@ public static RebalanceAlgorithm getInstance(
...
-    Map<SoftConstraint, Float> softConstraintsWithWeight = Maps.toMap(softConstraints, key -> {
-      String name = key.getClass().getSimpleName();
-      float weight = MODEL.get(name);
-      return name.equals(PartitionMovementConstraint.class.getSimpleName()) ?
-          movementPreference * weight : evennessPreference * weight;
-    });
+    Map<SoftConstraint, Float> softConstraintsWithWeight = new HashMap<>();
+    for (SoftConstraint softConstraint : softConstraints) {
+      float constraintWeight = MODEL.get(softConstraint.getClass().getSimpleName());
+      softConstraintsWithWeight.put(softConstraint,
```
jiajunwang, line 88: Comment on why we treat BaselineInfluenceConstraint as an evennessPreference instead of movementPreference please.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
jiajunwang: Moreover, if we move the auto-fill logic to ConstraintBasedAlgorithmFactory.java, then we need to add a test case there to cover the change.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/AbstractPartitionMovementConstraint.java
NealSun96: I'm reusing the old name, but it's really just the score (0.5), not a scaling factor. I'm renaming the method to clearly indicate that it's returning a score, not a scaling factor.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/BaselineInfluenceConstraint.java
NealSun96: Mentioned this in an earlier comment, I will rename the factor to a score, since that was the old behavior. 

I honestly think this way is fine, because it's clearer. The abstract function's methods act as the default implementation: it returns MAX_SCORE if totally match, MAX_SCORE/2 if requires state transtion, MIN_SCORE for anything else. If the extended function wants to be different, it can just override. 

Maybe I can even remove getStateTransitionCostFactor() and just put the default value in calculateAssignmentScore(). What do you think?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
NealSun96: For the cluster that motivates this change, an extremely small number like 0.001 is required to achieve better baseline evenness than CrushED. The quality of the baseline is only controlled by this factor. 

Note that we cannot set the 0.001 with preferences, since they must be integers. This part is where I find this approach having a hard time creating something elegant, unless we use system properties to overwrite.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
NealSun96: It has been moved to `AbstractPartitionMovementConstraint`

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
NealSun96: Which part are you referring to? If you mean line 43, it has been moved to `AbstractPartitionMovementConstraint` above the `getStateTransitionCostFactor` function (that is getting renamed). If you're referring to line 54, it has also been moved to `AbstractPartitionMovementConstraint` above `getAssignmentScore`.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
NealSun96: I have already added that to the javadoc of this constraint. I can reiterate it, sure.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
NealSun96: This default map is still being used, and the current behavior is exactly as you have described. Looks to me there's nothing to change here :)

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
NealSun96: They would be able to successfully set the preference, which is already tested by the above test case. 

I can add a test case in ConstraintBasedAlgorithmFactory for this.

NealSun96: On a second thought, do you think it's actually better to move the validation logic in ConstraintBasedAlgorithmFactory? My reasoning for the current approach is that we shouldn't make a special case out of certain keys because I want to keep the solution generic, but doing the check in ConstraintBasedAlgorithmFactory is just delaying the same thing. 

It might be better to keep the validation in setter since it's not silent.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
```diff
@@ -73,19 +97,13 @@ protected double getAssignmentScore(AssignableNode node, AssignableReplica repli
     if (bestPossibleAssignment.isEmpty()) {
       // If bestPossibleAssignment of the replica is empty, indicating this is a new replica.
       // Then the baseline is the only reference.
-      return calculateAssignmentScore(nodeName, state, baselineAssignment);
+      return calculateAssignmentScore(nodeName, state, baselineAssignment)
```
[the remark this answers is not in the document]
NealSun96: Outdated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
jiajunwang: Yeah, but it is OK just leave it here since it is still valid.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
jiajunwang: I meant the line 54 part. Then it is OK. Thanks for confiming.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
jiajunwang: Yeah, with your revised version, it looks good now. Before, since you hardcoded the value there, it is not really using the default numbers.

There is one more thing I suggest to change. It is not used anymore in this file. So we shall move it to ConstraintBasedAlgorithmFactory.java. Since the default preference is no longer part of the config class logic anymore.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
jiajunwang: Either way works. But we need to make it clean. The validation should be done in one place. So ConstraintBasedAlgorithmFactory might be a better option given the logic becomes more complex. But I don't have a strong preference.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
jiajunwang: Let's discuss offline, what you want to do is 2 features instead of one. This could be concerning. Since this forceBaselineConverge option is not exactly what it claims to be.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/BaselineInfluenceConstraint.java
jiajunwang: 1. removing getStateTransitionCostFactor() is the same thing as I commented before that we can make this method non-abstract. So no objection on my side.
2. MAX_SCORE / 2 is hardcode. Not good. What if we find that the cost is higher or lower than we thought? Or we want to make it configurable for different clusters? Then we need to fix it anyway.
3. I think the factor design is easier to maintain and cleaner, considering that we are not 100% sure that 0.5 is a good rate for everyone.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/AbstractPartitionMovementConstraint.java
jiajunwang: It is OK, but I don't like that design. First of all, I think we all agree that the logic would be the same.
My concern is that factor design is bug free. By returning a number between 0-1, the eventual score would be sure to be between MIN_SCORE and MAX_SCORE. However, if you directly return the getStateTransitionCost(), then we rely on the coder's discipline.

Actually, even you change the method to be getStateTransitionCost(), I think it shall still be implemented with the above formula,
(MAX_SCORE - MIN_SCORE) * getStateTransitionCostFactor() + MIN_SCORE

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/PartitionMovementConstraint.java
NealSun96: Ah, fixed.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
NealSun96: That's exactly my concern. Let's discuss.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
NealSun96: It's still used by `getGlobalRebalancePreference()` in the config class. I think it's fine where it is, because it has the format of a map, and logically belongs to ClusterConfig. Technically its default values should belong in the factory class, but that may be an overkill.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/AbstractPartitionMovementConstraint.java
NealSun96: That's a valid point. Ok.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
NealSun96: Conclusion: for this PR, I'm removing this line. Another solution will be introduced later if necessary.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
NealSun96: Moving back to ClusterConfig to validate during `set()`.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
```diff
@@ -61,19 +66,33 @@ public static RebalanceAlgorithm getInstance(
...
+    int movementPreference = preferences
+        .getOrDefault(ClusterConfig.GlobalRebalancePreferenceKey.LESS_MOVEMENT,
+            ClusterConfig.DEFAULT_GLOBAL_REBALANCE_PREFERENCE
+                .get(ClusterConfig.GlobalRebalancePreferenceKey.LESS_MOVEMENT));
+
+    boolean forceBaselineConverge = preferences
+        .getOrDefault(ClusterConfig.GlobalRebalancePreferenceKey.FORCE_BASELINE_CONVERGE, 0)
+        > 0;
+    if (forceBaselineConverge) {
+      MODEL.put(BaselineInfluenceConstraint.class.getSimpleName(), FORCE_BASELINE_CONVERGE_WEIGHT);
```
jiajunwang, line 82: I still suggest moving this after the soft constraint weight adjustment, and directly set to softConstraintsWithWeight instead of modifying the MODEL.

1. The 3 preference configurations shall be clearly defined. In theory, FORCE_BASELINE_CONVERGE
 is not related to EVENNESS or LESS_MOVEMENT. So we shall not let its static weight being polluted by the EVENNESS preference.
2. MODEL is a static map, but FORCE_BASELINE_CONVERGE can be changed dynamically. You cannot just enable it, but you will need to check and remove it if FORCE_BASELINE_CONVERGE is not setup. This is low efficiency. Better to directly put it into the dynamic map softConstraintsWithWeight. This is computed from scratch every time anyway.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
```diff
@@ -171,6 +164,15 @@ public void testSetRebalancePreferenceInvalidNumber() {
     testConfig.setGlobalRebalancePreference(preference);
   }
 
+  @Test(expectedExceptions = IllegalArgumentException.class)
+  public void testSetRebalancePreferenceMissingKey() {
+    Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> preference = new HashMap<>();
+    preference.put(EVENNESS, 1);
+
+    ClusterConfig testConfig = new ClusterConfig("testId");
+    testConfig.setGlobalRebalancePreference(preference);
```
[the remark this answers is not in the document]
NealSun96: Before, if they set a mapping with only one side defined, it's actually failing silently, in the sense that it never worked - the default weights were used. I think this is a step-up: we are now alerting users of this behavior that was previously silent.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithmFactory.java
NealSun96: Makes sense, changing.

helix-core/src/test/java/org/apache/helix/model/TestClusterConfig.java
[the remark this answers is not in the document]
jiajunwang: The original logic will still work though. Just the setup is kind of ignored. I agree that we can make it clearer. But please document this cleanly in the ClusterConfig class.

The concern for me is that if an existing cluster has been configured with a single number, this new version will fail the rebalance in that cluster. There is no alerting to the user except for the failure rebalancer in that case.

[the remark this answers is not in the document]
NealSun96: The old check is added again for backward compatibility. Documents will be updated.

jiajunwang (approved):
Looks great! Thanks for the hard work.

NealSun96:
Final commit message:
## Add new PartitonMovementConstraint and BaselineInfluenceConstraint for Waged ##
This PR splits PartitionMovementConstraint into separate constraints that control baseline convergence and best possible movements respectively.

PR #1663, opened 2021-03-05
Track failures for async read/write in ZkClient

### LINKED ISSUE #1662: Add ZkClient failure count metrics for batch read/write
Currently, batch (async) read/write failure count in ZkClient is not recorded, which is due to the different path for these operations compared with writeData/readData, where the read/write failure count is recorded.

Track the failure count for batch (async) read/write in ZkAsyncCallbacks.

rabashizade:
### Issues

Fixes #1662 

### Description

This PR addresses the issue of tracking failure counts for batch (async) read/write by checking for operation return code in `ZkAsyncCallbacks`, and if failure is detected, uses the `recordFailure()` function which is added to `ZkAsyncCallMonitorContext` to record these metrics in a monitor.

### Tests

`TestZkClientAsyncFailure`

```
[[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestExpandCluster.testClusterExpansion:61 expected:<false> but was:<true>
[ERROR]   TestZeroReplicaAvoidance.testWagedRebalancer:182 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1264, Failures: 3, Errors: 0, Skipped: 2
[INFO] 
[INFO] Reactor Summary for Apache Helix 1.0.2-SNAPSHOT:
[INFO] 
[INFO] Apache Helix ....................................... SUCCESS [  0.357 s]
[INFO] Apache Helix :: Metrics Common ..................... SUCCESS [  1.602 s]
[INFO] Apache Helix :: Metadata Store Directory Common .... SUCCESS [  1.688 s]
[INFO] Apache Helix :: ZooKeeper API ...................... SUCCESS [09:12 min]
[INFO] Apache Helix :: Helix Common ....................... SUCCESS [  1.003 s]
[INFO] Apache Helix :: Core ............................... FAILURE [  02:03 h]
```
The three failed tests in helix-core passed in later tries.

### Documentation (Optional)

### Commits

### Code Quality


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallMonitorContext.java
```diff
@@ -62,4 +62,18 @@ void recordAccess(String path) {
...
   }
+
+  /**
+   * Record the operation failure into the specified ZkClient monitor.
+   * @param path The monitored path
+   */
+  void recordFailure(String path) {
+    if (_monitor != null) {
+      if (_isRead) {
+        _monitor.recordFailure(path, ZkClientMonitor.AccessType.READ);
```
NealSun96, line 73: We could condense the lines to `_isRead ? ZkClientMonitor.AccessType.READ : ZkClientMonitor.AccessType.WRITE`

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -50,6 +50,10 @@ public void processResult(int rc, String path, Object ctx, byte[] data, Stat sta
         if (_data != null && ctx != null && ctx instanceof ZkAsyncCallMonitorContext) {
           ((ZkAsyncCallMonitorContext) ctx).setBytes(_data.length);
         }
+      } else if(rc != Code.NONODE.intValue()) {
+        if (ctx instanceof ZkAsyncCallMonitorContext) {
```
NealSun96, line 54: We could condense the two if lines into one && condition.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -67,6 +71,10 @@ public void handle() {
     public void processResult(int rc, String path, Object ctx, Stat stat) {
       if (rc == 0) {
         _stat = stat;
+      } else {
+        if (ctx instanceof ZkAsyncCallMonitorContext) {
```
NealSun96, line 75: `else if` would work here

NealSun96 (changes requested):

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestZkClientAsyncFailure.java
```diff
@@ -0,0 +1,207 @@
...
+
+    _monitor = new ZkClientMonitor(TEST_TAG, TEST_KEY, TEST_INSTANCE, false, null);
+    _monitor.register();
+  }
+
+  @AfterClass
+  public void afterClass() {
+    _monitor.unregister();
+    _zkClient.deleteRecursively(TEST_ROOT);
+    _zkClient.close();
```
pkuwm, line 71: This zkClient should not be closed, because it is attached to a zk server and could be used in other tests. If it is closed in this test, and following tests try to use this zkClient, the tests will fail because the zkClient is closed.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestZkClientAsyncFailure.java
```diff
@@ -0,0 +1,207 @@
...
+import org.apache.helix.zookeeper.zkclient.metric.ZkClientPathMonitor;
+import org.apache.zookeeper.KeeperException;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import static org.apache.helix.zookeeper.zkclient.callback.ZkAsyncCallbacks.UNKNOWN_RET_CODE;
+
+public class TestZkClientAsyncFailure extends ZkTestBase {
```
zhangmeng916, line 43: This test should probably be named as TestZkClientAsyncFailureMetric since your new code is only adding metrics. The aync function should be already covered in existing tests. 
With that said, have you checked whether it's possible to only add metrics testing to existing tests by mocking the results? That would save us one additional test file. If there is no available code path or too many efforts involved, I'm fine with having a separate test too. But please check the zkclient and monitor test.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestZkClientAsyncFailure.java
```diff
@@ -0,0 +1,207 @@
...
+      // asyncSet should fail because the return code is APIERROR
+      testZkClient.setAsyncCallRC(KeeperException.Code.APIERROR.intValue());
+      testZkClient.asyncSetData(NODE_PATH, tmpRecord, -1, setCallback);
+      setCallback.waitForSuccess();
+      Assert.assertEquals(setCallback.getRc(), KeeperException.Code.APIERROR.intValue());
+      Assert.assertEquals((long) _beanServer.getAttribute(instancesName,
+          ZkClientPathMonitor.PredefinedMetricDomains.WriteFailureCounter.toString()), 1);
+    } finally {
+      testZkClient.close();
+      _zkClient.delete(NODE_PATH);
```
pkuwm, line 122: This `_zkClient` might not be the one you want to use, because the test suite could start 3 zk servers. If testZkClient connects address1, and _zkClient connects to address2 (because you get it from the map), this cleanup won't work.

I suggest you may want to create a zkclient connecting to the `ZK_ADDR` and use this client for the whole test class. Then close it afterClass

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestZkClientAsyncFailure.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  private ObjectName buildPathMonitorObjectName(String tag, String key, String instance,
+      String path) throws MalformedObjectNameException {
+    return new ObjectName(String.format("%s,%s=%s", buildObjectName(tag, key, instance).toString(),
+        ZkClientPathMonitor.MONITOR_PATH, path));
+  }
+
+  @Test
+  public void testAsyncWrite() throws JMException {
```
zhangmeng916, line 86: Same for the test here. Do you think we miss async write and read testing in existing tests? If not, we only need to test metrics.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallMonitorContext.java
rabashizade: Thanks, @NealSun96 for your suggestion. The reason I didn't write it that way is, as @zhangmeng916 suggested, the function above this one uses the same pattern so I kept it that way for consistency.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
rabashizade: Thanks, I'll do it in the next commit.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestZkClientAsyncFailure.java
rabashizade: Thank you, @zhangmeng916 for your comments. They make sense and I will address them in the next commit. Just to be clear, are you suggesting adding these test methods to an existing test class, or incorporating these testing paths into an existing test method?

zhangmeng916: Yeah, I feel it should be possible to test metric in existing test files. It'd better if we could combine them.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -50,6 +50,10 @@ public void processResult(int rc, String path, Object ctx, byte[] data, Stat sta
         if (_data != null && ctx != null && ctx instanceof ZkAsyncCallMonitorContext) {
           ((ZkAsyncCallMonitorContext) ctx).setBytes(_data.length);
         }
+      } else if(rc != Code.NONODE.intValue()) {
```
pkuwm, line 53: This logic means, for each retry, we'll record one metric. Is it the expected requirement?
My thought was, only when the whole method fails, we record the failure.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestZkClientAsyncFailure.java
rabashizade: I'll probably remove this test file and include the methods in another existing test file, but are you sure about this? I saw a similar pattern (_zkClient.close() called in afterClass()) in TestZkClientAsyncRetry, that's why I called _zkClient.close() here.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
rabashizade: This is a good point, thanks. I should perform more tests on this and fix it.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestZkClientAsyncFailure.java
rabashizade: This makes sense, thanks. But is this a different scenario than for instance `TestZkClientAsyncRetry.testAsyncRetryCategories()`?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -67,6 +71,10 @@ public void handle() {
     public void processResult(int rc, String path, Object ctx, Stat stat) {
       if (rc == 0) {
         _stat = stat;
+      } else {
+        if (ctx instanceof ZkAsyncCallMonitorContext) {
+          ((ZkAsyncCallMonitorContext) ctx).recordFailure(path);
```
zhangmeng916, line 76: This is only for the get operation, right? What about write metrics?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
zhangmeng916: This is actually `ZkAsyncCallMonitorContext`, retry won't really go to this path. But I think @rabashizade has other findings regarding retry. He'll update later.

rabashizade: Thanks, @zhangmeng916 . Yes, actually due to the catch statement added to asyncSetData by @pkuwm in #1592, the retry and failure paths are different and their monitor contexts are also different. In the case of retry, no failure metric is increased. But a similar check doesn't exist for asyncGetData, and adding that should solve this problem unless there's a reason not to add this logic to asyncGetData.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestZkClientAsyncFailure.java
pkuwm: If you doubt it, feel free to dig it :) 
I'd say that one also needs fixing.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -50,6 +50,10 @@ public void processResult(int rc, String path, Object ctx, byte[] data, Stat sta
         if (_data != null && ctx != null && ctx instanceof ZkAsyncCallMonitorContext) {
           ((ZkAsyncCallMonitorContext) ctx).setBytes(_data.length);
         }
+      } else if(rc != Code.NONODE.intValue()) {
+        if (ctx instanceof ZkAsyncCallMonitorContext) {
+          ((ZkAsyncCallMonitorContext) ctx).recordFailure(path);
+        }
       }
       callback(rc, path, ctx);
```
jiajunwang, line 58: Why not process the result in the callback method? So that you process all the results in one place. The current version only records failures of 2 out of 5 zk calls.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestZkClientAsyncFailure.java
jiajunwang, line 43: I would suggest combining this test with TestZkClientAsyncRetry. There is some common logic.
Maybe you can just add several test methods there or modifying the existing logic a little bit and then check the metrics to verify that it works fine. I bet that would be easier than writing a whole new test.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
rabashizade: Thank you, @jiajunwang, your suggestion makes more sense. I will use it in the next commit.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestZkClientAsyncFailure.java
rabashizade: Thanks, yes it would be easier that way. I'll combine the new tests with existing ones in TestZkClientAsyncRetry.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -170,7 +170,17 @@ public void callback(int rc, String path, Object ctx) {
       }
 
       if (ctx != null && ctx instanceof ZkAsyncCallMonitorContext) {
-        ((ZkAsyncCallMonitorContext) ctx).recordAccess(path);
+        ZkAsyncCallMonitorContext monitor = (ZkAsyncCallMonitorContext) ctx;
+        monitor.recordAccess(path);
+        // Record failure if the return code isn't 0 (i.e., OK), and this object
+        // is one of the five callback types derived from DefaultCallback
```
jiajunwang, line 176: This check looks verbose. How about defining a reportFailure boolean in the DefaultCallback class. By default it is true. Then you overwrite it for SyncCallbackHandler (it seems to be the only one that you don't want to record failure, Why is that BTW?)
Then you can read this boolean field here and execute accordingly.

rabashizade: Thanks, I will use a boolean variable to simplify the check. And thanks for catching SyncCallbackHandler, not recording it wasn't intentional but since there wasn't an equivalent recording in ZkClient I missed it. I will add this to the check too. Only one question, for GetDataCallbackHandler and ExistsCallbackHandler I don't record a failure if the return code is NONODE (given the logic in their sync equivalent methods in ZkClient), should I again make this exception for SyncCallbackHandler? Or this doesn't make sense because SyncCallbackHandler checks the root path?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -170,7 +170,17 @@ public void callback(int rc, String path, Object ctx) {
       }
 
       if (ctx != null && ctx instanceof ZkAsyncCallMonitorContext) {
```
zhangmeng916, line 172: You still need to change the asynGet method to catch exception to make sure the ctx is an instance of `ZkAsyncCallMonitor` instead of `ZkAsyncRetryCallContext` right?

rabashizade: Yes, I think so. I'm discussing that with @pkuwm.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
jiajunwang: Please refer to the ZK document to see if Sync ever throws NO NODE error. I believe it does not. But I might be wrong : )

One more thing, given this logic, I think you can make the check more flexible instead of just a boolean field. Something like this,

boolean shouldRecordFailure(int rc) {
  // default return true if rc != 0;
  // special handle get/exists calls in the child class implementation.
}

Not a must. If it becomes too complicated, then you shall make an alternative design. My intention is mainly to reduce duplicate code when we have a clean OO design.

rabashizade: This is very helpful, thank you. I'll use it in the next commit.

zhangmeng916:
I see why you're doing this. If this can help simply a lot of redundant code, then it is fine.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2001,44 +2008,65 @@ protected void doRetry() {
...
-            @Override
-            protected void doRetry() {
-              asyncDelete(path, cb);
-            }
-          });
-      return null;
-    });
+    try {
+      retryUntilConnected(() -> {
+        ((ZkConnection) getConnection()).getZookeeper().delete(path, -1, cb,
```
jiajunwang, line 2055: I believe the only known case that this could fail on the first try is that if we use "getExpectedZookeeper(expectedSessionId)" to avoid the unexpected updates after a session change. Given that says,
1. @rabashizade did we see any real concern that it may fail on the first try? Or you are just trying to add more protection? I think the change is good, just want to understand if any real case you foresee.
2. @pkuwm any idea why we are not using getExpectedZookeeper(expectedSessionId) for the asyncDelete call?

rabashizade: I didn't encounter any real issues, I mostly put the check there due to the types of exceptions that could be thrown by `retryUntilConnected()`, and so I wanted to be able to record the failure in the event that one happens.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientPathMonitor.java
```diff
@@ -204,6 +224,22 @@ protected synchronized void record(int bytes, long latencyMilliSec, boolean isFa
...
+  /**
+   * Records metrics for async operations
+   */
+  protected synchronized void recordAsync(int bytes, long latencyMilliSec, boolean isFailure,
+      boolean isRead) {
+    if (isFailure) {
+      increaseAsyncFailureCounter(isRead);
+    } else {
+      increaseAsyncCounter(isRead);
+      increaseTotalLatency(isRead, latencyMilliSec);
```
zhangmeng916, line 236: I suggest not to record the async latency into the total latency. It may skew the current latency data and cause confusion. We can choose not to record its latency or put into a different metric.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/metric/ZkClientPathMonitor.java
```diff
@@ -204,6 +224,22 @@ protected synchronized void record(int bytes, long latencyMilliSec, boolean isFa
...
+      if (bytes > 0) {
+        increaseBytesCounter(isRead, bytes);
```
zhangmeng916, line 238: Same here. Better to keep current sync metrics intact.

zhangmeng916 (approved):

rabashizade:
This PR is ready to be merged, approved by @zhangmeng916.

Commit message:

Track failures for async read/write in ZkClient

Defines separate metrics for async (batch) operations and tracks failure count by checking for operation return code in ZkAsyncCallbacks. Also, cancels async get/exists/delete callbacks when an exception is thrown to be able to detect failure and record the metric. Finally, in TestRawZkClient sync read metrics were used to check async reads, which is fixed to reflect the separation of sync and async metrics.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallMonitorContext.java
```diff
@@ -56,9 +56,23 @@ void setBytes(int bytes) {
   void recordAccess(String path) {
     if (_monitor != null) {
       if (_isRead) {
-        _monitor.record(path, _bytes, _startTimeMilliSec, ZkClientMonitor.AccessType.READ);
+        _monitor.recordAsync(path, _bytes, _startTimeMilliSec, ZkClientMonitor.AccessType.READ);
```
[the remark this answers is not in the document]
rabashizade: Thanks for your comment. I agree in general, and changed the code in other locations accordingly which I will push soon, but in this particular location changing it requires refactoring the class (i.e., the `bool isRead` was defined in the class and used elsewhere in the code before this PR). Is that preferable to do? Or is a separate PR more appropriate for that?

PR #1676, opened 2021-03-23
Improve the GenericHelixController global tracking record to support multiple controller objects for the same cluster in one JVM.

### LINKED ISSUE #1617: The static map of GenericHelixController could be overwritten if create multiple controllers for the same cluster in one JVM
### Describe the bug
A static map of GenericHelixController is used to track all the controllers that are created in the JVM. This record is used to find the required controller object using the cluster name. However, if we run multiple controllers for a single cluster in one JVM, then these records will overwrite each other.

The problem usually happens in test when we need to simulate controller leadership election. But it may also happen in prod environment if we use a single JVM to run multiple controller instances.
One of the direct failures is that the scheduleOnDemandPipeline() method will fail since it cannot find the right controller instance from the map.
Note that this impacts WAGED rebalancer (and maybe TF logics) since they rely on the schedule method to trigger later rebalance.

### To Reproduce
In one JVM, create multiple controller instances for the same cluster. Then try to schedule an on-demand rebalance by using WAGED rebalance or Task framework features. In detail, just try to make some changes in the cluster which has WAGED usage, or TF usage. Then the scheduling of on-demand rebalance shall fail since the code cannot find the right controller to send schedule requests.

### Expected behavior
The rebalance schedule request shall be sent to the current leader controller. And the request shall be accepted and triggered normally.

### Additional context
The following link points to the problematic code.
https://github.com/apache/helix/blob/361e18da4152c0146daa9d9dc7929f1f2bdcd9dc/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java#L228

jiajunwang:
### Issues

Resolves #1617

### Description

This PR refines the GenericHelixController global tracking record so it acts as expected even we have more than one instance in the JVM. This impacts test logic only in general.

### Tests

The previously added workaround code has been removed. So the new logic can be tested by the original code.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -38,7 +38,9 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixManager;
@@ -204,7 +206,7 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
    */
   private long _lastPipelineEndTimestamp;
 
-  private String _clusterName;
+  private final String _clusterName;
   private final Set<Pipeline.Type> _enabledPipelineTypes;
 
   private HelixManager _helixManager;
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -225,10 +227,52 @@ protected StatefulRebalancer createRebalancer(HelixManager helixManager) {
...
    */
-  /* Map of cluster->GenrichelixController */
-  private static Map<String, GenericHelixController> HelixControllerFactory = new ConcurrentHashMap<>();
-  public static GenericHelixController getController(String clusterName) {
-    return HelixControllerFactory.get(clusterName);
+  /* Map of cluster->Set of GenericHelixController*/
+  private static Map<String, ImmutableSet<GenericHelixController>> _helixControllerFactory =
+      new ConcurrentHashMap<>();
+
+  public static GenericHelixController getLeaderController(String clusterName) {
```
dasahcc, line 234: Question: Is this necessary? This is usually in memory triggered rebalance. It only triggers when the node is leader for the cluster. Also you should not see one generic helix controller perform as LEADER and STANDBY at same time for a cluster. Do we need this set?

jiajunwang: This problem fails more than 2 tests. In production, and if the application dev is reasonable, then it won't appear. 
My take is that 1. the new change is harmless for the production logic. 2. if not doing this, we have to hack our test logic. So it may hide some real problem. Since the test is not doing what production code does.
So it is really necessary.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -225,10 +227,52 @@ protected StatefulRebalancer createRebalancer(HelixManager helixManager) {
    *  1) ZK callback should go to ClusterDataCache and trigger data cache refresh only
    *  2) then ClusterDataCache.refresh triggers rebalance pipeline.
    */
-  /* Map of cluster->GenrichelixController */
-  private static Map<String, GenericHelixController> HelixControllerFactory = new ConcurrentHashMap<>();
-  public static GenericHelixController getController(String clusterName) {
-    return HelixControllerFactory.get(clusterName);
+  /* Map of cluster->Set of GenericHelixController*/
+  private static Map<String, ImmutableSet<GenericHelixController>> _helixControllerFactory =
```
lei-xia, line 231: minor: for static variables, what is our name convention?  I think _wordAnotherWord format is for instance variable?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
lei-xia: Same question, why we could have two controllers for same cluster? Is this just happening in tests because the controller from other tests were not properly closed?

jiajunwang: There are more than one test cases need to start multiple controllers. If not doing so, then all these tests have to trigger the rebalance manually, since the auto scheduled ondemand rebalance won't work. This is a potential problem since the workaround hide problems.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
jiajunwang: The constants (static final) should be all upper case + underscore. But this var is not constants. Then, all the other related naming conventions include,
1. Class and instance variables: _fooBar (this is the one that makes sense)
2. Local variables: fooBar

On the other hand, HelixControllerFactory is not a good option. Since it conflicts with the Classes/Interfaces naming convention.

dasahcc (approved):

jiajunwang:
Approved by @dasahcc , I will merge the code. Thanks!

PR #1678, opened 2021-03-24
Close zkClients created by TaskStateModelFactory

### LINKED ISSUE #1677: Unclosed ZkClient in TaskStateModelFactory
### Describe the bug
TaskStateModelFactory creates a zkClient to read the user-customized thread pool size, however the zkClient is not closed. 

### To Reproduce
Initialize a TaskStateModelFactory. 

### Expected behavior
The zkClient should be closed to avoid leakage.

NealSun96:
### Issues

Fixes #1677 

### Description

TaskStateModelFactory creates either a shared ZkClient or a FederatedZkClient to read the user-customized number for the configurable thread pool. Neither of the ZkClients are closed in the end, resulting in thread leakage. 

### Tests

```
[INFO] Tests run: 1264, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,046.067 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1264, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  01:24 h
[INFO] Finished at: 2021-03-26T12:09:01-07:00
```
### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
@@ -125,12 +125,6 @@ public class TaskStateModelFactory extends StateModelFactory<TaskStateModel> {
    * Create a RealmAwareZkClient to get thread pool sizes
    */
   protected static RealmAwareZkClient createZkClient(HelixManager manager) {
-    // TODO: revisit the logic here - we are creating a connection although we already have a
-    // manager. We cannot use the connection within manager because some users connect the manager
-    // after registering the state model factory (in which case we cannot use manager's connection),
-    // and some connect the manager before registering the state model factory (in which case we
-    // can use manager's connection). We need to think about the right order and determine if we
-    // want to enforce it, which may cause backward incompatibility.
     if (!(manager instanceof ZKHelixManager)) {
       // TODO: None-ZKHelixManager cannot initialize this class. After interface rework of
       // HelixManager, the initialization should be allowed.
@@ -163,25 +157,39 @@ public class TaskStateModelFactory extends StateModelFactory<TaskStateModel> {
       }
     }
 
-    return SharedZkClientFactory.getInstance().buildZkClient(
-        new HelixZkClient.ZkConnectionConfig(zkAddress),
-        clientConfig.createHelixZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
+    // Fail early instead of retrying and blocking TaskStateModelFactory creation
+    HelixZkClient.ZkClientConfig zkClientConfig =
+        clientConfig.createHelixZkClientConfig().setZkSerializer(new ZNRecordSerializer());
+    zkClientConfig.setOperationRetryTimeout((long) 60000);
+    return SharedZkClientFactory.getInstance()
+        .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress), zkClientConfig);
   }
 
   private static ScheduledExecutorService createThreadPoolExecutor(HelixManager manager) {
+    // TODO: revisit the logic here - we are creating a connection although we already have a
+    // manager. We cannot use the connection within manager because some users connect the manager
+    // after registering the state model factory (in which case we cannot use manager's connection),
+    // and some connect the manager before registering the state model factory (in which case we
+    // can use manager's connection). We need to think about the right order and determine if we
+    // want to enforce it, which may cause backward incompatibility.
     RealmAwareZkClient zkClient = createZkClient(manager);
-    ScheduledExecutorService executorService = Executors.newScheduledThreadPool(TaskUtil
-            .getTargetThreadPoolSize(zkClient, manager.getClusterName(), manager.getInstanceName()),
-        new ThreadFactory() {
-          private AtomicInteger threadId = new AtomicInteger(0);
+    int targetThreadPoolSize;
 
-          @Override
-          public Thread newThread(Runnable r) {
-            return new Thread(r, "TaskStateModelFactory-task_thread-" + threadId.getAndIncrement());
-          }
-        });
-    zkClient.close();
+    // Ensure the zkClient is closed after reading the pool size;
+    try {
+      targetThreadPoolSize = TaskUtil
+          .getTargetThreadPoolSize(zkClient, manager.getClusterName(), manager.getInstanceName());
+    } finally {
+      zkClient.close();
+    }
 
-    return executorService;
+    return Executors.newScheduledThreadPool(targetThreadPoolSize, new ThreadFactory() {
+      private AtomicInteger threadId = new AtomicInteger(0);
+
+      @Override
+      public Thread newThread(Runnable r) {
+        return new Thread(r, "TaskStateModelFactory-task_thread-" + threadId.getAndIncrement());
+      }
+    });
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
@@ -52,6 +52,9 @@ import org.slf4j.LoggerFactory;
 public class TaskStateModelFactory extends StateModelFactory<TaskStateModel> {
   private static Logger LOG = LoggerFactory.getLogger(TaskStateModelFactory.class);
 
+  // Unit in minutes. Need a retry timeout to prevent zkClient from hanging infinitely.
+  private static int ZKCLIENT_OPERATION_RETRY_TIMEOUT = 5;
+
   private final HelixManager _manager;
   private final Map<String, TaskFactory> _taskFactoryRegistry;
   private final ScheduledExecutorService _taskExecutor;
@@ -135,7 +138,8 @@ public class TaskStateModelFactory extends StateModelFactory<TaskStateModel> {
     RealmAwareZkClient.RealmAwareZkClientConfig clientConfig =
         new RealmAwareZkClient.RealmAwareZkClientConfig().setZkSerializer(new ZNRecordSerializer());
     // Set operation retry timeout to prevent hanging infinitely
-    clientConfig.setOperationRetryTimeout(Duration.ofMinutes(5).toMillis());
+    clientConfig
+        .setOperationRetryTimeout(Duration.ofMinutes(ZKCLIENT_OPERATION_RETRY_TIMEOUT).toMillis());
     String zkAddress = manager.getMetadataStoreConnectionString();
 
     if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED) || zkAddress == null) {
--- a/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
@@ -19,6 +19,7 @@ package org.apache.helix.task;
  * under the License.
  */
 
+import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -134,7 +135,7 @@ public class TaskStateModelFactory extends StateModelFactory<TaskStateModel> {
     RealmAwareZkClient.RealmAwareZkClientConfig clientConfig =
         new RealmAwareZkClient.RealmAwareZkClientConfig().setZkSerializer(new ZNRecordSerializer());
     // Set operation retry timeout to prevent hanging infinitely
-    clientConfig.setOperationRetryTimeout((long) 60000);
+    clientConfig.setOperationRetryTimeout(Duration.ofMinutes(5).toMillis());
     String zkAddress = manager.getMetadataStoreConnectionString();
 
     if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED) || zkAddress == null) {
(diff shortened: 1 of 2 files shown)


helix-core/src/test/java/org/apache/helix/task/TestTaskStateModelFactory.java
```diff
@@ -95,6 +95,7 @@ public void testZkClientCreationMultiZk() throws Exception {
         .getTargetThreadPoolSize(zkClient, anyParticipantManager.getClusterName(),
             anyParticipantManager.getInstanceName()), TEST_TARGET_TASK_THREAD_POOL_SIZE);
     Assert.assertTrue(zkClient instanceof FederatedZkClient);
+    zkClient.close();
```
xyuanlu, line 98: Do you think add thins `zkClient.close()` to a final block may help? If assert fail then this Utest fails, we will miss the close.

NealSun96: Very true, let me fix it up.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -176,4 +167,21 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
...
+          private AtomicInteger threadId = new AtomicInteger(0);
+
+          @Override
+          public Thread newThread(Runnable r) {
+            return new Thread(r, "TaskStateModelFactory-task_thread-" + threadId.getAndIncrement());
+          }
+        });
+    zkClient.close();
+
+    return executorService;
```
narendly, line 185: I think it might be helpful to break this function down to:

1. Do I/O read from ZK (`createZkClient(manager)`, `getTargetThreadPoolSize()`, cache the read value into a local variable), and close the zkClient.
2. Initialize an instance of `executorService` using the read value

You could refactor Step 1 into the constructor (this should be 2-3 lines at most) - this way, `createThreadPoolExecutor` only does what its name says it does - just creating creating an in-memory thread pool, nothing else. This way, we keep private functions modular and the intent is clearer to readers.

NealSun96: @narendly Unfortunately, to keep things DRY, the constructor has to use `this()`, and `Call to 'this()' must be first statement in constructor body`. 

The current approach is fine in my opinion: `createThreadPoolExecutor()` does what it needs to do to create the thread pool and cleans up after itself.

narendly: Can you refactor ZK operations out first? What I am saying is if `Executors` fails for any reason would zkclient be closed properly? I don't see any issues with closing zkclient before we move straight on to thread pool initialization.

NealSun96: That's true. I wouldn't be able to refactor it into the constructor, but I can place that I/O first in this helper. Thanks for the tips.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -176,4 +167,21 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
...
+            .getTargetThreadPoolSize(zkClient, manager.getClusterName(), manager.getInstanceName()),
+        new ThreadFactory() {
+          private AtomicInteger threadId = new AtomicInteger(0);
+
+          @Override
+          public Thread newThread(Runnable r) {
+            return new Thread(r, "TaskStateModelFactory-task_thread-" + threadId.getAndIncrement());
+          }
+        });
+    zkClient.close();
```
pkuwm, line 183: Use try..finally to make sure the client is close, even any exception is thrown before the close.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -176,4 +167,21 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
         new HelixZkClient.ZkConnectionConfig(zkAddress),
         clientConfig.createHelixZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
   }
+
+  private static ScheduledExecutorService createThreadPoolExecutor(HelixManager manager) {
+    RealmAwareZkClient zkClient = createZkClient(manager);
```
narendly, line 172: Could we also set the client config so we don't have infinite operation timeout for this particular zk client?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -163,25 +157,39 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
...
 
-    return SharedZkClientFactory.getInstance().buildZkClient(
-        new HelixZkClient.ZkConnectionConfig(zkAddress),
-        clientConfig.createHelixZkClientConfig().setZkSerializer(new ZNRecordSerializer()));
+    // Fail early instead of retrying and blocking TaskStateModelFactory creation
+    HelixZkClient.ZkClientConfig zkClientConfig =
+        clientConfig.createHelixZkClientConfig().setZkSerializer(new ZNRecordSerializer());
+    zkClientConfig.setOperationRetryTimeout((long) 60000);
+    return SharedZkClientFactory.getInstance()
+        .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddress), zkClientConfig);
```
narendly, line 165: This might be problematic (only applies to SharedZkClient):

```
  /**
   * Construct and init a ZkConnection Manager.
   *
   * @param zkConnection
   * @param connectionTimeout
   */
  protected ZkConnectionManager(IZkConnection zkConnection, long connectionTimeout,
      String monitorKey) {
    super(zkConnection, (int) connectionTimeout, HelixZkClient.DEFAULT_OPERATION_TIMEOUT,
        new BasicZkSerializer(new SerializableSerializer()), MONITOR_TYPE, monitorKey, null, true);
    _monitorKey = monitorKey;
    LOG.info("ZkConnection {} was created for sharing.", _monitorKey);
  }
```

Even if we set and pass in a zkClientConfig, it just gets ignored and will take on a default value, which is infinite. This is a limitation for shared zk client - I'm okay with not addressing this in this PR, but we should document this erroneous behavior.

Also, is there a reason we're not giving a client config with timeout to FederatedZkClient?

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -172,8 +157,39 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
...
+    }
+
+    return Executors.newScheduledThreadPool(targetThreadPoolSize, new ThreadFactory() {
+      private AtomicInteger threadId = new AtomicInteger(0);
+
+      @Override
+      public Thread newThread(Runnable r) {
+        return new Thread(r, "TaskStateModelFactory-task_thread-" + threadId.getAndIncrement());
+      }
+    });
```
narendly, line 193: It might be beneficial to log it here somewhere that includes various pieces of information (e.g. `targetThreadPoolSize`)

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: Strange. I will revert these lines then, since setting the retry timeout is obsolete. Issue created: https://github.com/apache/helix/issues/1682.

I'm not adding a timeout to FederatedZkClient because it creates new connections, instead of possibly reuses old connections like SharedZkClient. I don't see the possibility of newly created connections having mismatching zxid problems and hang infinitely; if that's a concern, then `ZkBaseDataAccessor` or `ZkBucketDataAccessor` should also set a retry timeout. In short, since it creates new connections, I don't see it as necessary to go beyond the default behavior and set retry timeouts.

narendly: My suggestion is to give it a valid zk client config with a non-infinite operation retry timeout regardless of what type of ZkClient we're using here.

Operation retry timeout and the server's refusal due to lagging zxid seen are different modes of failure. Other Helix APIs like you mentioned can accept client config at creation - those are managed by users.

NealSun96: Discussed offline: there are values in preventing infinite hanging on client side. Added retry timeout to both.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -148,6 +133,8 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
     }
     RealmAwareZkClient.RealmAwareZkClientConfig clientConfig =
         new RealmAwareZkClient.RealmAwareZkClientConfig().setZkSerializer(new ZNRecordSerializer());
+    // Set operation retry timeout to prevent hanging infinitely
+    clientConfig.setOperationRetryTimeout((long) 60000);
```
pkuwm, line 137: Question: Are 60 secs good enough? There could transient zk connection issues. My suggestion is make it a bit longer, eg. 5, 10, or 30 mins, to have a higher rate of success.
And a code style suggestion, try to use java `Duration` instead of this hard coded number, which will make it much easier to read/maintain.
```
Duration.ofMinutes(10).toMillis()
```

The lazy initialization way mentioned by @jiajunwang was also what I was thinking of when I debugged the code here. Too many things are done for constructing an object was my first impression: creating a ZkClient, connecting to zk and reading the znode. If I did, I'd prefer to lazily construct the threadpool/reading znode after this factory object is constructed. Can be done later.

helix-core/src/test/java/org/apache/helix/task/TestTaskStateModelFactory.java
```diff
@@ -91,28 +93,22 @@ public void testZkClientCreationMultiZk() throws Exception {
 
     RoutingDataManager.getInstance().reset();
     RealmAwareZkClient zkClient = TaskStateModelFactory.createZkClient(anyParticipantManager);
-    Assert.assertEquals(TaskUtil
-        .getTargetThreadPoolSize(zkClient, anyParticipantManager.getClusterName(),
-            anyParticipantManager.getInstanceName()), TEST_TARGET_TASK_THREAD_POOL_SIZE);
-    Assert.assertTrue(zkClient instanceof FederatedZkClient);
+    verifyThreadPoolSizeAndZkClientClass(zkClient, anyParticipantManager,
```
pkuwm, line 96: Why not also include the zkclient creation `RealmAwareZkClient zkClient = TaskStateModelFactory.createZkClient(anyParticipantManager);` in the verify method? It seems a better practice to create and close the zkclient in the same method.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
NealSun96: That's a good suggestion on code style, thanks. I will go with 5 minutes.

Lazy construction was considered when this feature was first developed. It requires a larger scale of code change, so it won't be done in this bugfix PR. It can be done later, though, like you said.

helix-core/src/test/java/org/apache/helix/task/TestTaskStateModelFactory.java
NealSun96: The test case is for zkClient, so style-wise I didn't want to wrap the main actor in a helper; but since it's a test case, sure, why not.

dasahcc (approved):

NealSun96:
Final commit message:
## Close zkClients created by TaskStateModelFactory ##
This PR closes the previously unclosed zkClients in TaskStateModelFactory, also added retry timeout and log to the logic.

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -148,6 +134,8 @@ protected static RealmAwareZkClient createZkClient(HelixManager manager) {
     }
     RealmAwareZkClient.RealmAwareZkClientConfig clientConfig =
         new RealmAwareZkClient.RealmAwareZkClientConfig().setZkSerializer(new ZNRecordSerializer());
+    // Set operation retry timeout to prevent hanging infinitely
+    clientConfig.setOperationRetryTimeout(Duration.ofMinutes(5).toMillis());
```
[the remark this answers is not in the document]
NealSun96: 1. Done.
2. It works for FederatedClient, so we still need to add it.

narendly (approved):
Thanks for working on this!

helix-core/src/main/java/org/apache/helix/task/TaskStateModelFactory.java
```diff
@@ -51,23 +52,17 @@
 public class TaskStateModelFactory extends StateModelFactory<TaskStateModel> {
   private static Logger LOG = LoggerFactory.getLogger(TaskStateModelFactory.class);
 
+  // Unit in minutes. Need a retry timeout to prevent zkClient from hanging infinitely.
+  private static int ZKCLIENT_OPERATION_RETRY_TIMEOUT = 5;
```
narendly, line 56: Nit: constant declarations also include `final`. So it would be `private static final int`

NealSun96: Wow, my 10pm brain is really sloppy... thanks for catching it.

NealSun96:
Final commit message:
## Close zkClients created by TaskStateModelFactory ##
This PR closes the previously unclosed zkClients in TaskStateModelFactory, also added retry timeout and log to the logic.

NealSun96:
Final commit message:
## Close zkClients created by TaskStateModelFactory ##
This PR closes the previously unclosed zkClients in TaskStateModelFactory, also added retry timeout and log to the logic.

Issue #1675, opened 2021-03-23
Remove requested state update in task framework


xyuanlu:
**Is your feature request related to a problem? Please describe.**
In the current design, although Task's CurrentState update requests are sent be participant, they are all instructed by controller's message. In TaskFramework, all task's current state update is instructed by messages sent by controller. When task finish execution or failed with error, participant will set requested_state field in CurrentState ZNode requesting a state transition message. The following graph illustrates how state transition is done when task finish executing. It shows all ZK writes including messages and CurrentState ZNode update. Since the task thread is actually running in participant side, participant should update task's CurrentState directly. There are numbers of redundant ZNode updates to be optimized.

![Remove requested state (3)](https://user-images.githubusercontent.com/6392021/112074133-cc786500-8b32-11eb-8789-1978e93142cf.jpg)

Apart from the redundant IO, when task is actually finished running on participant's task runner thread, one more round of pipeline run → sending message → state transition → ZK update is needed before the actual CurrentState is Completed or TaskError. It will introduce some ambiguous for users when task is actually not running on task running thread but CurrentState is Running.

**Describe the solution you'd like**
Participant will update CurrentState directly in taskRunner thread.

<img width="1230" alt="Screen Shot 2021-03-10 at 11 57 15 AM" src="https://user-images.githubusercontent.com/6392021/112074178-f3cf3200-8b32-11eb-9414-e2e91cd9784e.png">

In the origin design, all CurrentState updates are scheduled by controller. When task finishes executing or ends with error state, task runner thread uses Requested thread to communicate with controller and controller will only send one state transition message at a time.

If Task runner thread update CurrentState when task finishes or ends with error state, it is possible that the state transition handler is also trying to update CurrentState. It could be an INIT→ RUNNING, RUNNING→CANCEL, or any other state transition message message.  We may run into the following write conflict. 

<img width="1148" alt="Screen Shot 2021-03-17 at 6 28 37 PM" src="https://user-images.githubusercontent.com/6392021/112361768-fa73bb80-8c90-11eb-86cd-3fac751330e1.png">

We can prevent over write by checking the in memory state _stateModel and update ZK conditionally in message handling. Also the in memory state _stateModel update, comparing  and ZK update need to be protected by a lock. 

Originally in `StateTransitionHandler`, the whole message processing is also synchronized on the _state object. StateTransitionHandler.invoke will create a task runner thread executing the task synchronically for 'RUNNING' -> 'STOPPED' and asynchronically for all other state transitions. 
[...]

PR #1681, opened 2021-03-24
Remove lock in HelixStateTransitionHandler

xyuanlu:
### Issues

#1675 Remove requested state update in task framwork.
This is the first PR of the issue. 
This PR is for lock scope deduction only, because lock scope change expects more review and has independent logic itself.

### Description

*High level design of the lock scope: 

We can prevent over write by checking the in memory state _stateModel and update ZK conditionally in message handling. Also the in memory state _stateModel update (in both task Runner thread and state transition handling thread), comparing and ZK update need to be protected by a lock.

*Is it safe to do so? 
In current message handling design, `HelixTaskExecutor` (a message lister that creates and schedules all message handler) guarantees that only one on going state transition per each partition/task at a time, removing this synchronization shouldn't cause any problem. 
When `onMessage` receives a STATE_TRANSITION_CANCELLATION messages and a ST message for the same entity, a `HelixStateTransitionCancellationHanlder` is created can call cancel(). Changing the lock scope does not have any influence on `HelixStateTransitionCancellationHanlder`.

*Why need to Remove lock in HelixStateTransitionHandle
In the design to remove requested state, task runner thread update CurrentState when task finishes or ends with error state. 
However, it is possible that the state transition handler is also trying to update CurrentState. It could be an INIT→ RUNNING, RUNNING→CANCEL, or any other state transition message message. We will run into the write conflict.
This PR removes this redundant synchronization. Will add a smaller scoped lock in the PR that removes requested state.

### Tests

Our current Task framework tests covers all state transition types, should be able to cover this change. 

```
[ERROR] Failures:                                                                                                                                                                      
[ERROR] org.apache.helix.integration.rebalancer.DelayedAutoRebalancer.TestDelayedAutoRebalanceWithDisabledInstance.beforeTest(org.apache.helix.integration.rebalancer.DelayedAutoRebalan
cer.TestDelayedAutoRebalanceWithDisabledInstance)                                                                                                                                      
[ERROR]   Run 1: TestDelayedAutoRebalanceWithDisabledInstance.beforeTest:304->enableInstance:315 expected:<true> but was:<false>                                                       
[INFO]   Run 2: PASS                                                                                                                                                                   
[INFO]                                                                                                                                                                                 
[INFO]                                                                        
[ERROR] Tests run: 9, Failures: 1, Errors: 0, Skipped: 6                       
[INFO]                                                        
[INFO] BUILD FAILURE                                                                                                                                                                   
[INFO] Total time:  40.449 s                                                                                                                                                           
[INFO] Finished at: 2021-04-19T16:45:27-07:00  
rerun:
[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 78.981 s - in org.apache.helix.integration.rebalancer.DelayedAutoRebalancer.TestDelayedAutoRebalanceWithDisabledI
nstance                                                            
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -369,9 +369,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
     taskResult.setCompleteTime(System.currentTimeMillis());
     // add task result to context for postHandling
     context.add(MapKey.HELIX_TASK_RESULT.toString(), taskResult);
-    synchronized (_stateModel) {
-      postHandleMessage();
-    }
+    postHandleMessage();
 
     return taskResult;
   }
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -318,60 +318,62 @@ public class HelixStateTransitionHandler extends MessageHandler {
     NotificationContext context = _notificationContext;
     Message message = _message;
 
-    synchronized (_stateModel) {
-      HelixTaskResult taskResult = new HelixTaskResult();
-      HelixManager manager = context.getManager();
+    HelixTaskResult taskResult = new HelixTaskResult();
+    HelixManager manager = context.getManager();
 
-      _statusUpdateUtil.logInfo(message, HelixStateTransitionHandler.class,
-          "Message handling task begin execute", manager);
-      message.setExecuteStartTimeStamp(System.currentTimeMillis());
+    _statusUpdateUtil
+        .logInfo(message, HelixStateTransitionHandler.class, "Message handling task begin execute",
+            manager);
+    message.setExecuteStartTimeStamp(System.currentTimeMillis());
 
-      try {
-        preHandleMessage();
-        invoke(manager, context, taskResult, message);
-      } catch (HelixDuplicatedStateTransitionException e) {
-        // Duplicated state transition problem is fine
-        taskResult.setSuccess(true);
-        taskResult.setMessage(e.toString());
-        taskResult.setInfo(e.getMessage());
-      } catch (HelixStateMismatchException e) {
-        // Simply log error and return from here if State mismatch.
-        // The current state of the state model is intact.
+    try {
+      preHandleMessage();
+      invoke(manager, context, taskResult, message);
+    } catch (HelixDuplicatedStateTransitionException e) {
+      // Duplicated state transition problem is fine
+      taskResult.setSuccess(true);
+      taskResult.setMessage(e.toString());
+      taskResult.setInfo(e.getMessage());
+    } catch (HelixStateMismatchException e) {
+      // Simply log error and return from here if State mismatch.
+      // The current state of the state model is intact.
+      taskResult.setSuccess(false);
+      taskResult.setMessage(e.toString());
+      taskResult.setException(e);
+    } catch (Exception e) {
+      String errorMessage =
+          "Exception while executing a state transition task " + message.getPartitionName();
+      logger.error(errorMessage, e);
+      if (e.getCause() != null && e.getCause() instanceof InterruptedException) {
+        e = (InterruptedException) e.getCause();
+      }
+
+      if (e instanceof HelixRollbackException || (e.getCause() != null && e
+          .getCause() instanceof HelixRollbackException)) {
+        // TODO : Support cancel to any state
+        logger.info(
+            "Rollback happened of state transition on resource \"" + _message.getResourceName()
+                + "\" partition \"" + _message.getPartitionName() + "\" from \"" + _message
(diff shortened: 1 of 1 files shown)


dasahcc (commented):
Question: will it be save to exclude the invoke() for the locking? There could be two messages at the same time invoked for the same state mode: 1 regular ST and 1 cancel message. I am not quite sure whether that could cause problem.

xyuanlu:
> Question: will it be save to exclude the invoke() for the locking? There could be two messages at the same time invoked for the same state mode: 1 regular ST and 1 cancel message. I am not quite sure whether that could cause problem.

Thanks for the review. 
Please correct me if I am wrong. I don't think there could be two messages at the same time invoked for the same state mode. 

When `onMessage` receives a `STATE_TRANSITION_CANCELLATION` messages and a ST message for the came entity, it will try to
1. Delete the previous created `stateTransitionHandler` if the to be canceled state transition is not started. In this step, there will be no `stateTransitionHandler` created for this cancelation message.
2. If the the to be canceled state transition is not started, we will try to create a `stateTransitionHandler` for this cancelation message only when there is no on going ST for the same entity. 


```
public void onMessage(){
    for (Message message : messages) {
        if (checkAndProcessNoOpMessage(message, instanceName, changeContext, manager, sessionId,
          stateTransitionHandlers)) {  /*  Call cancelNotStartedStateTransition */                           ------(1)
   .......
        msgHandler = createMessageHandler(message, msgWorkingContext);

        if (message.getMsgType().equals(MessageType.STATE_TRANSITION.name()) || message.getMsgType()
          .equals(MessageType.STATE_TRANSITION_CANCELLATION.name())) {
          if (validateAndProcessStateTransitionMessage(message, manager, stateTransitionHandlers,              ------(2)
              msgHandler)) {
    }
}
......
}
 private boolean validateAndProcessStateTransitionMessage{
     if (stateTransitionHandlers.containsKey(messageTarget)) {
        // If there are 2 messages in same batch about same partition's state transition,
        // the later one is discarded
        Message duplicatedMessage = stateTransitionHandlers.get(messageTarget)._message;
        String errMsg = String.format(
            "Duplicated state transition message: %s. Existing: %s->%s; New (Discarded): %s->%s",
            message.getMsgId(), duplicatedMessage.getFromState(), duplicatedMessage.getToState(),
            message.getFromState(), message.getToState());
        updateUnprocessableMessage(message, null /* exception */, errMsg, manager);
        return false;
      }
}

...
// Try to cancel this state transition that has not been started yet.
// Three Types of Cancellation: 1. Message arrived with previous state transition
//                              2. Message handled but task not started
//                              3. Message handled and task already started
// This method tries to handle the first two cases, it returns true if no further cancellation is needed,
// false if not been able to cancel the state transition (i.e, further cancellation is needed).
private boolean cancelNotStartedStateTransition(Message message,
      Map<String, MessageHandler> stateTransitionHandlers, HelixDataAccessor accessor,
      String instanceName) {
...
}

```

jiajunwang:
Question, Could you explain more about the necessity of the change? In detail, the main thread creates taskRunner thread which updates the state at the end. And the main thread continues to run after taskRunner thread is created. In this case, even the main thread holds the lock, the taskRunner can still execute. And the runner thread can wait until the main thread finishes the state update. I remember you mentioned this when we talked offline, but please document the detailed reason in this ticket too.

xyuanlu:
> Question, Could you explain more about the necessity of the change? In detail, the main thread creates taskRunner thread which updates the state at the end. And the main thread continues to run after taskRunner thread is created. In this case, even the main thread holds the lock, the taskRunner can still execute. And the runner thread can wait until the main thread finishes the state update. I remember you mentioned this when we talked offline, but please document the detailed reason in this ticket too.

Thanks for the question.
For most task state transition messages, it is true that the main state transition handler thread creates taskRunner thread and the main thread continues to run after taskRunner thread is created. 

However, when handling 'RUNNING' -> 'STOPPED' , stateTransition handler creates a taskRunner, wait for state transition finishes and then continue the rest part.  So if we include the `invoke()` in the lock scope, we will have a deadlock here.


```

  @Transition(to = "STOPPED", from = "RUNNING")
  public String onBecomeStoppedFromRunning(Message msg, NotificationContext context) {
    String taskPartition = msg.getPartitionName();
    if (_taskRunner == null) {
      throw new IllegalStateException(String.format(
          "Invalid state transition. There is no running task for partition %s.", taskPartition));
    }

    _taskRunner.cancel();
    TaskResult r = _taskRunner.waitTillDone();
    LOG.info("Task {} completed with result {}.", msg.getPartitionName(), r);

    timeout_task.cancel(false);

    return r.getInfo();
  ```

xyuanlu:
Changing this PR to removing the redundant lock since it is not logically complete.
Will add a smaller scoped lock in the PR that removes requested state.

dasahcc (approved):

xyuanlu:
This PR is ready to be merged, approved by @dasahcc.

Final commit message:

Remove synchronization on _staleModel in HelixStateTransitionHandler 

In current message handling design, `HelixTaskExecutor` (a message lister that creates and schedules all message handler) guarantees that only one on-going state transition is created at a time.  This change remove this duplicated synchronization in HelixStateTransitionHandler.

PR #1689, opened 2021-03-26
Add option to continue checks on failures for stoppable api

### LINKED ISSUE #1688: Add an option to always perform all checks in Helix stoppable check API
Helix stoppable check currently does the check in 3 sequential stages: Helix standard checks, custom instance health check and custom partition health check. When one stage had any failure, the subsequent stages are not executed.

This poses an issue that if participant happens to ignore all the failed checks in stage 1 but stage 2/3 would have reported some other failures, then it would think it's safe to stop the instance when it's in fact not safe. For example, if an instance would fail INSTANCE_NOT_STABLE in stage 1 and custom partition health check in stage 3, Helix will only return INSTANCE_NOT_STABLE and if participant ignores this failure reason, it will consider the instance safe to stop when it's not.

Provide an option in Helix stoppable check API, such that when the option is used, Helix will always perform all checks and return all failed checks.

pkuwm:
### Issues

Resolves #1688 

### Description

Helix stoppable check currently does the check in 3 sequential stages: Helix standard checks, custom instance health check and custom partition health check. When one stage had any failure, the subsequent stages are not executed.

This poses an issue that if participant happens to ignore all the failed checks in stage 1 but stage 2/3 would have reported some other failures, then it would think it's safe to stop the instance when it's in fact not safe. For example, if an instance would fail INSTANCE_NOT_STABLE in stage 1 and custom partition health check in stage 3, Helix will only return INSTANCE_NOT_STABLE and if participant ignores this failure reason, it will consider the instance safe to stop when it's not.

Provide an option in Helix stoppable check API, such that when the option is used, Helix will always perform all checks and return all failed checks. Query param: **continueOnFailures**

### Tests

- testGetStoppableWithAllChecks

Helix rest
```
17:14:54,948 [INFO] Tests run: 172, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 92.112 s - in TestSuite
17:14:55,296 [INFO]
17:14:55,296 [INFO] Results:
17:14:55,296 [INFO]
17:14:55,296 [INFO] Tests run: 172, Failures: 0, Errors: 0, Skipped: 0
17:14:55,297 [INFO]
17:14:55,305 [INFO] ------------------------------------------------------------------------
17:14:55,305 [INFO] BUILD SUCCESS
17:14:55,306 [INFO] ------------------------------------------------------------------------
17:14:55,308 [INFO] Total time:  01:36 min
17:14:55,309 [INFO] Finished at: 2021-03-25T17:14:55-07:00
17:14:55,309 [INFO] ------------------------------------------------------------------------
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/json/instance/StoppableCheck.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/json/instance/StoppableCheck.java
@@ -78,4 +78,9 @@ public class StoppableCheck {
   public List<String> getFailedChecks() {
     return failedChecks;
   }
+
+  public void add(StoppableCheck other) {
+    failedChecks.addAll(other.getFailedChecks());
+    isStoppable = failedChecks.isEmpty();
+  }
 }
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/InstancesAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/InstancesAccessor.java
@@ -151,7 +151,7 @@ public class InstancesAccessor extends AbstractHelixResource {
   public Response instancesOperations(
       @PathParam("clusterId") String clusterId,
       @QueryParam("command") String command,
-      @QueryParam("allChecks") boolean allChecks,
+      @QueryParam("continueOnFailures") boolean continueOnFailures,
       @QueryParam("skipZKRead") boolean skipZKRead,
       String content) {
     Command cmd;
@@ -181,7 +181,7 @@ public class InstancesAccessor extends AbstractHelixResource {
         admin.enableInstance(clusterId, enableInstances, false);
         break;
       case stoppable:
-        return batchGetStoppableInstances(clusterId, node, skipZKRead, allChecks);
+        return batchGetStoppableInstances(clusterId, node, skipZKRead, continueOnFailures);
       default:
         _logger.error("Unsupported command :" + command);
         return badRequest("Unsupported command :" + command);
@@ -198,7 +198,7 @@ public class InstancesAccessor extends AbstractHelixResource {
   }
 
   private Response batchGetStoppableInstances(String clusterId, JsonNode node, boolean skipZKRead,
-      boolean allChecks) throws IOException {
+      boolean continueOnFailures) throws IOException {
     try {
       // TODO: Process input data from the content
       InstancesAccessor.InstanceHealthSelectionBase selectionBase =
@@ -228,7 +228,7 @@ public class InstancesAccessor extends AbstractHelixResource {
           InstancesAccessor.InstancesProperties.instance_not_stoppable_with_reasons.name());
       InstanceService instanceService =
           new InstanceServiceImpl((ZKHelixDataAccessor) getDataAccssor(clusterId),
-              getConfigAccessor(), skipZKRead, allChecks, getNamespace());
+              getConfigAccessor(), skipZKRead, continueOnFailures, getNamespace());
       ClusterService clusterService = new ClusterServiceImpl(getDataAccssor(clusterId), getConfigAccessor());
       ClusterTopology clusterTopology = clusterService.getClusterTopology(clusterId);
       switch (selectionBase) {
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/InstancesAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/InstancesAccessor.java
@@ -148,8 +148,12 @@ public class InstancesAccessor extends AbstractHelixResource {
   @ResponseMetered(name = HttpConstants.WRITE_REQUEST)
   @Timed(name = HttpConstants.WRITE_REQUEST)
   @POST
-  public Response instancesOperations(@PathParam("clusterId") String clusterId,
-      @QueryParam("skipZKRead") String skipZKRead, @QueryParam("command") String command, String content) {
+  public Response instancesOperations(
+      @PathParam("clusterId") String clusterId,
+      @QueryParam("command") String command,
+      @QueryParam("allChecks") boolean allChecks,
+      @QueryParam("skipZKRead") boolean skipZKRead,
+      String content) {
     Command cmd;
     try {
       cmd = Command.valueOf(command);
@@ -177,7 +181,7 @@ public class InstancesAccessor extends AbstractHelixResource {
         admin.enableInstance(clusterId, enableInstances, false);
         break;
       case stoppable:
-        return batchGetStoppableInstances(clusterId, node, Boolean.valueOf(skipZKRead));
+        return batchGetStoppableInstances(clusterId, node, skipZKRead, allChecks);
       default:
         _logger.error("Unsupported command :" + command);
         return badRequest("Unsupported command :" + command);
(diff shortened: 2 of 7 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
```diff
@@ -225,9 +248,11 @@ public StoppableCheck getInstanceStoppableCheck(String clusterId, String instanc
...
-          finalStoppableCheckByInstance.put(instance, stoppableCheck);
-        } else {
-          // instance passed this around of check will be checked in the next round
+          addStoppableCheck(finalStoppableCheckByInstance, instance, stoppableCheck);
+        }
+        if (stoppableCheck.isStoppable() || _allChecks){
+          // instance passed this around of check or mandatory all checks
+          // will be checked in the next round
           instancesForNextCheck.add(instance);
         }
```
dasahcc, line 257: They are correlated logic. Two "if" makes it independent checks.
So the logic should be: 

if (!stop && !_allChecks) {
   addStoppableCheck(finalStoppableCheckByInstance, instance, stoppableCheck);
} else {
   instancesForNextCheck.add(instance);
}

pkuwm: @dasahcc I've thought about it and this logic won't work. Because when allChecks enabled, customers would like to have all the failed checks in the response. In the logic you presented, if allChecks == true, it won't add the failed check.

I do think the current 2 "if"s are clearer to understand.

dasahcc: I see you concerns. Just last suggestion, we can do it in a separate RB. The logic for adding stoppable checks is not very good. We can keep a map of failed checks for instances as a global view instead of keeping creating StoppableCheck object. It was designed for old paths. Now it's time we can change it.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
```diff
@@ -206,14 +216,27 @@ public StoppableCheck getInstanceStoppableCheck(String clusterId, String instanc
...
+  private void addStoppableCheck(Map<String, StoppableCheck> stoppableChecks, String instance,
+      StoppableCheck stoppableCheck) {
+    if (!stoppableChecks.containsKey(instance)) {
+      stoppableChecks.put(instance, stoppableCheck);
+    } else {
+      // Merge two checks
+      StoppableCheck existedCheck = stoppableChecks.get(instance);
+      existedCheck.add(stoppableCheck);
+      stoppableChecks.put(instance, existedCheck);
+    }
```
NealSun96, line 237: NIT: Can be replaced with `computeIfAbsent`.

NealSun96 (changes requested):

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
pkuwm: I thought about both `putIfAbsent` and `computeIfAbsent` but neither of them fit in the logic.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
pkuwm: Agreed. This was also something I would do to make it cleaner when working on the code. But it should be fine for now as one stoppable call does not have too many instances, so StoppableCheck objects won't be too many for the life cycle of a call. It'll need some refactoring work. An issue is created track it: #1692 .

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/InstancesAccessor.java
```diff
@@ -148,8 +148,12 @@ public Response getAllInstances(@PathParam("clusterId") String clusterId,
   @ResponseMetered(name = HttpConstants.WRITE_REQUEST)
   @Timed(name = HttpConstants.WRITE_REQUEST)
   @POST
-  public Response instancesOperations(@PathParam("clusterId") String clusterId,
-      @QueryParam("skipZKRead") String skipZKRead, @QueryParam("command") String command, String content) {
```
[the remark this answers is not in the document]
pkuwm: Believe me, the Jar rs jersey framework handles it pretty well. For this case, the value will be false. So it won't fail.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/InstancesAccessor.java
```diff
@@ -148,8 +148,12 @@ public Response getAllInstances(@PathParam("clusterId") String clusterId,
...
+  public Response instancesOperations(
+      @PathParam("clusterId") String clusterId,
+      @QueryParam("command") String command,
+      @QueryParam("allChecks") boolean allChecks,
```
jiajunwang, line 154: As we discussed, please change to use a different param name.

helix-rest/src/main/java/org/apache/helix/rest/server/json/instance/StoppableCheck.java
```diff
@@ -78,4 +78,9 @@ public boolean isStoppable() {
   public List<String> getFailedChecks() {
     return failedChecks;
   }
+
+  public void add(StoppableCheck other) {
```
jiajunwang, line 82: 2 comments,
1. I think we should call it "mergeFailures", or something similar, like appendFailures.
2. Can we merge to check results with the different categories? If not, shall we assert when trying to merge?

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
NealSun96: Sorry, not `computeIfAbsent`, but `compute`. Do you think this would work?
`stoppableChecks.compute(instance, (k, v) -> v == null ? stoppableCheck : v.add(stoppableCheck)`

NealSun96 (changes requested):

helix-rest/src/main/java/org/apache/helix/rest/server/json/instance/StoppableCheck.java
pkuwm: 1. I think `add(Stoppable)` is better. What I consider is like `list.addAll();`. 
`mergeFailures` should be `void mergeFailures(failures1, failures2);` which does not fit in this design. Similar to appendFailures.
2. Each stoppable is a different categories. They are strings. So the failures list is actually a list of strings of different categories: `[HELIX:abc, CUSTOM_PARTITION:efg, CUSTOM_INSTANCE: xyz]`

jiajunwang: I see. I misunderstood the content of the string. Maybe you can help to add some comments in the code to help the readers? That would be very helpful.

Regarding the name, "add" makes sense under this context, which I didn't understand before. Please keep it.

helix-rest/src/main/java/org/apache/helix/rest/server/service/InstanceServiceImpl.java
pkuwm: I thought about it. But `v.add(stoppableCheck)` would require the `add()` method to return `this` object which I don't prefer to.

jiajunwang (approved):
LGTM!

pkuwm:

Issue #1653, opened 2021-02-22
Gracefully handle very large partitions in WAGED


jiajunwang:
**Is your feature request related to a problem? Please describe.**
It has been detected that a very large partition may impact WAGED rebalance performance. For example, a partition that is weighted more than 50% of the instance capacity might not be able to re-allocated without lifting the other smaller partitions. For now, WAGED won't do this type of cascading re-allocating. As a result, the rebalance calculation will fail and the large partition won't be re-allocated even it is necessary.

**Describe the solution you'd like**
My proposal to handle this situation is that,
1. WAGED should read and respect resource priority. If it is configured, WAGED should always rebalance the higher priority resource partitions first.
2. If there is no enough room for the high priority resource partition to be re-allocation (but overall cluster capacity allows), then the rebalancer shall lift lower priority resources to accommodate the high prioritized one.

**Additional context**

jiajunwang:
In another related but different case, if the partitions in one resource have very different weight, then even the resource priority does not help.
This indicates that the 2nd logic we mentioned above should also be applied to a wider range. Not only for the lower priority resources.

PR #1691, opened 2021-03-30
Improve the WAGED.ConstraintBasedAlgorithm sorting logic to prioritize replica with larger impact

xyuanlu:
### Issues

 #1653 Gracefully handle very large partitions in WAGED 

### Description

Every time we chose the replica that has highest impact on the whole cluster when state priority is the same.
Total impact = SUM(weight * Resource_capacity) where weight = 1/(1-resource_util)
where weight of each capacity dimension = 1/(1-usagePct)

### Tests

TestConstraintBasedAlgorithm.testSortingByResourceCapacity()
This newly added test is failing on master, pass with this PR.

```
[INFO] Results:
[INFO]
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:52 h
[INFO] Finished at: 2021-04-02T17:58:58-07:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
@@ -42,6 +42,7 @@ import org.apache.helix.model.ResourceAssignment;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
 /**
  * The algorithm is based on a given set of constraints
  * - HardConstraint: Approve or deny the assignment given its condition, any assignment cannot
@@ -80,9 +81,9 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
             .map(replica -> new AssignableReplicaWithScore(replica, clusterModel))
             .collect(Collectors.toSet());
 
-    AssignableReplicaWithScore replicaWithScore =
-        getNextAssignableReplica(toBeAssignedReplicas,  overallClusterRemainingCapacityMap);
-    while (replicaWithScore != null) {
+    while (!toBeAssignedReplicas.isEmpty()) {
+      AssignableReplicaWithScore replicaWithScore =
+          getNextAssignableReplica(toBeAssignedReplicas, overallClusterRemainingCapacityMap);
       toBeAssignedReplicas.remove(replicaWithScore);
       AssignableReplica replica = replicaWithScore.getAssignableReplica();
       Optional<AssignableNode> maybeBestNode =
@@ -98,10 +99,10 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
       }
       AssignableNode bestNode = maybeBestNode.get();
       // Assign the replica and update the cluster model.
-      clusterModel.assign(replica.getResourceName(), replica.getPartitionName(),
-          replica.getReplicaState(), bestNode.getInstanceName());
+      clusterModel
+          .assign(replica.getResourceName(), replica.getPartitionName(), replica.getReplicaState(),
+              bestNode.getInstanceName());
       updateOverallClusterRemainingCapacity(overallClusterRemainingCapacityMap, replica);
-      replicaWithScore = getNextAssignableReplica(toBeAssignedReplicas,  overallClusterRemainingCapacityMap);
     }
     optimalAssignment.updateAssignments(clusterModel);
     return optimalAssignment;
@@ -214,9 +215,13 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
       // score = SUM(weight * (resource_capacity/cluster_capacity) where weight = 1/(1-total_util%)
       // it could be be simplified to "resource_capacity/cluster_remainingCapacity".
       for (Map.Entry<String, Integer> resourceCapacity : _replica.getCapacity().entrySet()) {
-        score += resourceCapacity.getValue() * 100 / (
-            overallClusterRemainingCapMap.get(resourceCapacity.getKey())
-                + 0.01f /*avoid divide by 0*/);
+        score = resourceCapacity.getValue() == 0 ? score
+            : (overallClusterRemainingCapMap.get(resourceCapacity.getKey()) == 0 ? Integer.MAX_VALUE
+                : score + Math.min(100, resourceCapacity.getValue() * 100 / (overallClusterRemainingCapMap
+                    .get(resourceCapacity.getKey()))));
+        if (score == Integer.MAX_VALUE) {
+          break;
+        }
       }
       _score = score;
     }
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
@@ -42,6 +42,8 @@ import org.apache.helix.model.ResourceAssignment;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static java.lang.Float.compare;
+
 
 /**
  * The algorithm is based on a given set of constraints
@@ -196,7 +198,7 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
 
   private class AssignableReplicaWithScore implements Comparable<AssignableReplicaWithScore> {
     private final AssignableReplica _replica;
-    private int _score = 0;
+    private float _score = 0;
     private boolean _isInBestPossibleAssignment;
     private boolean _isInBaselineAssignment;
     private final Integer  _replicaHash;
@@ -211,15 +213,17 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
     }
 
     public void computeScore(Map<String, Integer> overallClusterRemainingCapMap) {
-      int score = 0;
+      float score = 0;
       // score = SUM(weight * (resource_capacity/cluster_capacity) where weight = 1/(1-total_util%)
       // it could be be simplified to "resource_capacity/cluster_remainingCapacity".
       for (Map.Entry<String, Integer> resourceCapacity : _replica.getCapacity().entrySet()) {
         score = resourceCapacity.getValue() == 0 ? score
-            : (overallClusterRemainingCapMap.get(resourceCapacity.getKey()) == 0 ? Integer.MAX_VALUE
-                : score + Math.min(100, resourceCapacity.getValue() * 100 / (overallClusterRemainingCapMap
-                    .get(resourceCapacity.getKey()))));
-        if (score == Integer.MAX_VALUE) {
+            : ((overallClusterRemainingCapMap.get(resourceCapacity.getKey()) == 0
+                || resourceCapacity.getValue() > (overallClusterRemainingCapMap
+                .get(resourceCapacity.getKey()))) ? Float.MAX_VALUE
+                : score + (float) resourceCapacity.getValue() / (overallClusterRemainingCapMap
+                    .get(resourceCapacity.getKey())));
+        if (compare(score, Float.MAX_VALUE) == 0) {
           break;
         }
       }
@@ -262,8 +266,9 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
 
       // 3. Sort according to the replica impact based on the weight.
       // So the greedy algorithm will place the replicas with larger impact first.
-      if (replica2._score != _score) {
-        return replica2._score - _score;
+      int result = compare(replica2._score, _score);
+      if (result != 0) {
+        return result;
       }
 
       // 4. Sort according to the resource/partition name.
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
@@ -31,6 +31,8 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 import com.google.common.collect.Maps;
+import org.apache.commons.lang3.tuple.MutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.helix.HelixRebalanceException;
 import org.apache.helix.controller.rebalancer.waged.RebalanceAlgorithm;
 import org.apache.helix.controller.rebalancer.waged.model.AssignableNode;
@@ -56,6 +58,7 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
   private final List<HardConstraint> _hardConstraints;
   private final Map<SoftConstraint, Float> _softConstraints;
 
+
   ConstraintBasedAlgorithm(List<HardConstraint> hardConstraints,
       Map<SoftConstraint, Float> softConstraints) {
     _hardConstraints = hardConstraints;
@@ -68,23 +71,47 @@ class ConstraintBasedAlgorithm implements RebalanceAlgorithm {
     List<AssignableNode> nodes = new ArrayList<>(clusterModel.getAssignableNodes().values());
     Set<String> busyInstances =
         getBusyInstances(clusterModel.getContext().getBestPossibleAssignment().values());
-    // Sort the replicas so the input is stable for the greedy algorithm.
-    // For the other algorithm implementation, this sorting could be unnecessary.
-    for (AssignableReplica replica : getOrderedAssignableReplica(clusterModel)) {
+    List<AssignableReplica> toBeAssignedReplicas = clusterModel.getAssignableReplicaMap().values().stream()
+        .flatMap(replicas -> replicas.stream()).collect(Collectors.toList());
+    // Compute overall utilization of the cluster. Capacity dimension -> <total usage, total capacity>
+    Map<String, MutablePair<Integer, Integer>> overallClusterUtilMap =
+        computeOverallClusterUtil(nodes);
+    Map<String, Integer> replicaHashCodeMap = toBeAssignedReplicas.parallelStream().collect(Collectors
+        .toMap(AssignableReplica::toString,
+            replica -> Objects.hash(replica.toString(), clusterModel.getAssignableNodes().keySet()),
+            (hash1, hash2) -> hash2));
+    Pair<AssignableReplica, List<AssignableReplica>> replicaRes =
+        getNextAssignableReplica(toBeAssignedReplicas, overallClusterUtilMap, clusterModel,
+            replicaHashCodeMap);
+    AssignableReplica replica = replicaRes.getLeft();
+    toBeAssignedReplicas = replicaRes.getRight();
+
+    while (replica != null) {
       Optional<AssignableNode> maybeBestNode =
           getNodeWithHighestPoints(replica, nodes, clusterModel.getContext(), busyInstances,
               optimalAssignment);
       // stop immediately if any replica cannot find best assignable node
-      if (optimalAssignment.hasAnyFailure()) {
-        String errorMessage = String
-            .format("Unable to find any available candidate node for partition %s; Fail reasons: %s",
+      if (!maybeBestNode.isPresent() || optimalAssignment.hasAnyFailure()) {
+        String errorMessage = String.format(
+            "Unable to find any available candidate node for partition %s; Fail reasons: %s",
             replica.getPartitionName(), optimalAssignment.getFailures());
+        System.out.println(String.format(
+            "Unable to find any available candidate node for partition %s; Fail reasons: %s",
+            replica.getPartitionName(), optimalAssignment.getFailures()));
         throw new HelixRebalanceException(errorMessage,
             HelixRebalanceException.Type.FAILED_TO_CALCULATE);
       }
-      maybeBestNode.ifPresent(node -> clusterModel
+      AssignableNode bestNode = maybeBestNode.get();
+      // Assign the replica and update the cluster model.
+      clusterModel
           .assign(replica.getResourceName(), replica.getPartitionName(), replica.getReplicaState(),
-              node.getInstanceName()));
+              bestNode.getInstanceName());
+      updateOverallClusterUtil(overallClusterUtilMap, replica);
+      // Compute next best replica and update remaining list.
+      replicaRes = getNextAssignableReplica(toBeAssignedReplicas, overallClusterUtilMap, clusterModel,
+          replicaHashCodeMap);
+      replica = replicaRes.getLeft();
+      toBeAssignedReplicas = replicaRes.getRight();
     }
     optimalAssignment.updateAssignments(clusterModel);
     return optimalAssignment;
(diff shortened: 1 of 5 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +179,147 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+      List<AssignableNode> nodes) {
+    Map<String, MutablePair<Integer, Integer>> utilizationMap = new HashMap<>();
+    for (AssignableNode node : nodes) {
+      for (String resourceKey : node.getMaxCapacity().keySet()) {
 
-    Map<String, ResourceAssignment> bestPossibleAssignment =
-        clusterModel.getContext().getBestPossibleAssignment();
-    Map<String, ResourceAssignment> baselineAssignment =
-        clusterModel.getContext().getBaselineAssignment();
+        if (utilizationMap.containsKey(resourceKey)) {
```
NealSun96, line 188: `resourceKey` is not the best name, because Resource has a meaning in WAGED. Maybe `capacityKey` is better.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +179,147 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+    // Compare every replica with current candidate, update candidate if needed
+    for (AssignableReplica replica : allReplica) {
+      // score = SUM(weight * Resource_capacity) where weight = 1/(1-resource_util)
+      float score = 0.01f;
+      for (Map.Entry<String, Integer> resourceCapacity : replica.getCapacity().entrySet()) {
+        // use int to store % * 100
+        int replicaUsageInClusterPoolPct =
+            (resourceCapacity.getValue() * 100) / overallClusterUtilMap
+                .get(resourceCapacity.getKey()).right;
+        score += replicaUsageInClusterPoolPct / (weightMap.get(resourceCapacity.getKey()) + 0.01);
```
NealSun96, line 244: Is the 0.01 to avoid divide by 0? May be a good idea to comment that.

Also, this division cancels out the denominator if I'm not mistaken? This score can be simplified to "replica usage / remaining capacity in the cluster".

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +179,147 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+        replicaForNextRound.add(replica);
       }
-    });
-    return orderedAssignableReplicas;
+    }
+    return new MutablePair<>(nextAssinableReplica, replicaForNextRound);
+  }
+
+  // return true is replica1 is better
+  private boolean keepOriginalPreferredReplica(AssignableReplica replica1,
```
NealSun96, line 264: NIT: function naming should be closer to something like `compareReplicas`, because the logic is not about "keeping a replica" per se.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +179,147 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+        nextAssinableReplica = replica;
       } else {
-        // If the best possible assignment contains the assignment, prioritize the replica.
-        return bestPossibleAssignment.containsKey(resourceName1) ? -1 : 1;
+        replicaForNextRound.add(replica);
       }
-    });
-    return orderedAssignableReplicas;
+    }
+    return new MutablePair<>(nextAssinableReplica, replicaForNextRound);
```
NealSun96, line 260: Why does `replicaForNextRound` need to be computed and returned by `getNextAssignableReplica`? It looks like it has no purpose in `getNextAssignableReplica`, because the logic is to find the replica with the highest score. 

Would it be more efficient to just remove the "best replica" from the `toBeAssignedReplicas` that's already computed in the outer scope?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -68,23 +71,47 @@ public OptimalAssignment calculate(ClusterModel clusterModel) throws HelixRebala
...
               optimalAssignment);
       // stop immediately if any replica cannot find best assignable node
-      if (optimalAssignment.hasAnyFailure()) {
-        String errorMessage = String
-            .format("Unable to find any available candidate node for partition %s; Fail reasons: %s",
+      if (!maybeBestNode.isPresent() || optimalAssignment.hasAnyFailure()) {
+        String errorMessage = String.format(
+            "Unable to find any available candidate node for partition %s; Fail reasons: %s",
             replica.getPartitionName(), optimalAssignment.getFailures());
+        System.out.println(String.format(
```
NealSun96, line 98: NIT: might want to remove this.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +179,147 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+    // Random functions here. Use hashcode based on the cluster topology information to get
+    // a controlled randomized order is good enough.
+    Integer replicaHash1 = replicaHashCodeMap.get(replica1.toString());
+    Integer replicaHash2 = replicaHashCodeMap.get(replica2.toString());
+    if (!replicaHash1.equals(replicaHash2)) {
+      return replicaHash1.compareTo(replicaHash2) < 0;
+    } else {
+      // In case of hash collision, return order according to the name.
+      return replica1.toString().compareTo(replica2.toString()) < 0;
+    }
```
NealSun96, line 322: NIT: I suppose you could also have pre-sorted the list of replicas with the other 3 conditions, store the sorted list outside, then just compare the scores during `getNextAssignableReplica`. It doesn't provide any performance boost but it looks cleaner IMO.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +179,147 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+        .entrySet()) {
+      float freePercentage =
+          (float) (resourceUsage.getValue().right - resourceUsage.getValue().left) / resourceUsage
+              .getValue().right;
+      weightMap.put(resourceUsage.getKey(), freePercentage);
+    }
+    // Compare every replica with current candidate, update candidate if needed
+    for (AssignableReplica replica : allReplica) {
+      // score = SUM(weight * Resource_capacity) where weight = 1/(1-resource_util)
+      float score = 0.01f;
```
NealSun96, line 238: Question: why does it start with 0.01?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -68,23 +71,47 @@ public OptimalAssignment calculate(ClusterModel clusterModel) throws HelixRebala
     List<AssignableNode> nodes = new ArrayList<>(clusterModel.getAssignableNodes().values());
     Set<String> busyInstances =
         getBusyInstances(clusterModel.getContext().getBestPossibleAssignment().values());
-    // Sort the replicas so the input is stable for the greedy algorithm.
-    // For the other algorithm implementation, this sorting could be unnecessary.
-    for (AssignableReplica replica : getOrderedAssignableReplica(clusterModel)) {
+    List<AssignableReplica> toBeAssignedReplicas = clusterModel.getAssignableReplicaMap().values().stream()
+        .flatMap(replicas -> replicas.stream()).collect(Collectors.toList());
+    // Compute overall utilization of the cluster. Capacity dimension -> <total usage, total capacity>
+    Map<String, MutablePair<Integer, Integer>> overallClusterUtilMap =
```
NealSun96, line 77: The Pair usage makes it very unclear what each number represents, especially when this map is passed into later functions. The references of `left` and `right` has no obvious meaning later on. 
I suggest:
1. getting rid of the `total capacity`. As I commented later, seems like the calculation cancels out `total capacity` and it may not be needed at all;
2. creating helper functions in ClusterModel with meaningful names, and leave all calculations/updates in the helper functions;
3. using 2 maps with meaningful names.

NealSun96 (changes requested):
Overall, I suggest another look at several variable names as they could be more informative. I also suggest adding more comments as I imagine other readers will have difficulties understanding the logic.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +179,147 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+    Map<String, MutablePair<Integer, Integer>> utilizationMap = new HashMap<>();
+    for (AssignableNode node : nodes) {
+      for (String resourceKey : node.getMaxCapacity().keySet()) {
 
-    Map<String, ResourceAssignment> bestPossibleAssignment =
-        clusterModel.getContext().getBestPossibleAssignment();
-    Map<String, ResourceAssignment> baselineAssignment =
-        clusterModel.getContext().getBaselineAssignment();
+        if (utilizationMap.containsKey(resourceKey)) {
+          Integer newUtil = utilizationMap.get(resourceKey).getLeft() + node.getRemainingCapacity()
```
jiajunwang, line 189: getRemainingCapacity() gives you idle capacity, it seems not what you want. Please double-check.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -68,23 +71,47 @@ public OptimalAssignment calculate(ClusterModel clusterModel) throws HelixRebala
...
         getBusyInstances(clusterModel.getContext().getBestPossibleAssignment().values());
-    // Sort the replicas so the input is stable for the greedy algorithm.
-    // For the other algorithm implementation, this sorting could be unnecessary.
-    for (AssignableReplica replica : getOrderedAssignableReplica(clusterModel)) {
+    List<AssignableReplica> toBeAssignedReplicas = clusterModel.getAssignableReplicaMap().values().stream()
+        .flatMap(replicas -> replicas.stream()).collect(Collectors.toList());
+    // Compute overall utilization of the cluster. Capacity dimension -> <total usage, total capacity>
+    Map<String, MutablePair<Integer, Integer>> overallClusterUtilMap =
+        computeOverallClusterUtil(nodes);
+    Map<String, Integer> replicaHashCodeMap = toBeAssignedReplicas.parallelStream().collect(Collectors
```
jiajunwang, line 79: Please add some comments here since the logic has been moved away from where it is used. Might be hard to understand.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +179,147 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
-          return baselineAssignment.containsKey(resourceName1) ? -1 : 1;
+  private Pair<AssignableReplica, List<AssignableReplica>> getNextAssignableReplica(
+      List<AssignableReplica> allReplica,
+      Map<String, MutablePair<Integer, Integer>> overallClusterUtilMap, ClusterModel clusterModel,
+      Map<String, Integer> _replicaHashCodeMap) {
+    List<AssignableReplica> replicaForNextRound = new ArrayList<>();
+    float highestScore = -1;
+    AssignableReplica nextAssinableReplica = null;
+    Map<String, Float> weightMap = new HashMap<>();
+    // compute weight for each capacity dimension
```
jiajunwang, line 227: Should be a separate method. And in this way, you can just pass the weight map to this select replica method.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +179,147 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+    for (Map.Entry<String, MutablePair<Integer, Integer>> resourceUsage : overallClusterUtilMap
+        .entrySet()) {
+      float freePercentage =
+          (float) (resourceUsage.getValue().right - resourceUsage.getValue().left) / resourceUsage
```
jiajunwang, line 231: For protecting any possible future errors (for example, we allow overprovisioning in the future), please bound the subtraction result it should be 0 at least, I guess. Or if the later calculation can handle a negative percentage number, then it would be fine here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +179,147 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+        // use int to store % * 100
+        int replicaUsageInClusterPoolPct =
+            (resourceCapacity.getValue() * 100) / overallClusterUtilMap
+                .get(resourceCapacity.getKey()).right;
+        score += replicaUsageInClusterPoolPct / (weightMap.get(resourceCapacity.getKey()) + 0.01);
+      }
+      if (nextAssinableReplica == null) {
+        nextAssinableReplica = replica;
+        highestScore = score;
+      } else if (!keepOriginalPreferredReplica(nextAssinableReplica, replica, clusterModel,
```
jiajunwang, line 249: I prefer to keep the sorting algorithm simple. Can we do the following steps,
1. Calculate the score, and set it to replica (with some wrapper class)
2. Compare the replicas to find the "maximum" using the customized compare method (which has been put into keepOriginalPreferredReplica in this PR).

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -68,23 +71,47 @@ public OptimalAssignment calculate(ClusterModel clusterModel) throws HelixRebala
...
+        .toMap(AssignableReplica::toString,
+            replica -> Objects.hash(replica.toString(), clusterModel.getAssignableNodes().keySet()),
+            (hash1, hash2) -> hash2));
+    Pair<AssignableReplica, List<AssignableReplica>> replicaRes =
+        getNextAssignableReplica(toBeAssignedReplicas, overallClusterUtilMap, clusterModel,
+            replicaHashCodeMap);
+    AssignableReplica replica = replicaRes.getLeft();
+    toBeAssignedReplicas = replicaRes.getRight();
+
+    while (replica != null) {
```
jiajunwang, line 89: Could do-while fit better here?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
NealSun96: As discussed with @xyuanlu , pre-sorting will have additional n(log n) complexity, since the usage here is only "find max", not sorting. Feel free to discard this comment.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
NealSun96: As discussed, the concern is that removing the "best replica" from the list is O(n). Please consider using a set for the outer scope data structure since the list-typing serves no purpose.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: Thanks for the review. I updated the PR based on our offline discussion.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: Good point, Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: Thanks for the review. I updated the PR based on our offline discussion.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: I am not sure... I think we still need to call `getNextAssignableReplica` twice when using do-while .

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +169,135 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+    }
+
+    public void reComputeScore(Map<String, Integer> overallClusterRemainingCapMap) {
+      int score = 0;
+      // score = SUM(weight * (resource_capacity/cluster_capacity) where weight = 1/(1-total_util%)
+      // it could be be simplified to "resource_capacity/cluster_remainingCapacity".
+      for (Map.Entry<String, Integer> resourceCapacity : _replica.getCapacity().entrySet()) {
+        score += resourceCapacity.getValue() * 100 / (
+            overallClusterRemainingCapMap.get(resourceCapacity.getKey())
+                + 0.01f /*avoid divide by 0*/);
```
NealSun96, line 225: Does 0.01 really make sense, in case when a certain capacity is measured in decimals, and 0.01 becomes a significant portion (total cluster capacity = 0.01 TB, as an extreme example, then adding 0.01 is a 100% increase to the remaining cap). 

On the other hand, why are we handling the case for "remaining cap = 0"? When that happens, this assignment is already doomed, right? I don't think it's possible to place this replica any more.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +169,135 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+  private void updateOverallClusterRemainingCapacity(
+      Map<String, Integer> overallClusterRemainingCapacityMap, AssignableReplica replica) {
+    for (Map.Entry<String, Integer> resourceUsage : replica.getCapacity().entrySet()) {
+      Integer newRemainingCapacity =
+          overallClusterRemainingCapacityMap.get(resourceUsage.getKey()) - resourceUsage.getValue();
+      overallClusterRemainingCapacityMap.put(resourceUsage.getKey(), newRemainingCapacity);
+    }
+  }
+
+  private class AssignableReplicaWithScore implements Comparable<AssignableReplicaWithScore> {
```
NealSun96, line 202: Two comments about function naming here: 

1. We could just say "computeScore" instead of "reComputeScore". We're re-computing by calling the compute function again. 
2. We could just say "getAssignableReplica" instead of "getRawAssignableReplica". 

Happy to discuss.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -68,23 +68,40 @@ public OptimalAssignment calculate(ClusterModel clusterModel) throws HelixRebala
...
+        clusterModel.getAssignableReplicaMap().values().stream()
+            .flatMap(replicas -> replicas.stream())
+            .map(replica -> new AssignableReplicaWithScore(replica, clusterModel))
+            .collect(Collectors.toSet());
+
+    AssignableReplicaWithScore replica =
+        getNextAssignableReplica(toBeAssignedReplicas,  overallClusterRemainingCapacityMap);
+    while (replica != null) {
+      toBeAssignedReplicas.remove(replica);
+      AssignableReplica rawReplica = replica.getRawAssignableReplica();
```
NealSun96, line 87: (This is related to my comment on function naming for `AssignableReplicaWithScore`. )
`rawReplica` here is really just `replica`, and the `replica` before should be `replicaWithScore`, right? I believe that's more clear as the names reflect their classes, and `getRawAssignableReplica()` could just become `getAssignableReplica()`.

NealSun96 (changes requested):

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: TFTR. Good suggestion. Will update.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: Thanks for the comment. I think it reveals a valid point -- should we continue the process when remaining capacity is 0? I think we could safe several compute cycle by checking hard constrain first and quit early. That might be more elegant comparing to this hacky way of avoiding divide by 0. 
Regarding the example in your comment, I am not sure if this will happen since the remaining capacity is an Integer. I think adding 0.01 to 1 might not change the result a lot since we only care about the relative relationship between each replicas. 
Considering the effort in development, I think maybe adding 0.01 here is the easiest way to avoid divide by 0. 

All comments are welcome.  :D

NealSun96: My original comment is more focused on the fact that 0.01 is a magical number that could have an impact to the remaining capacities, since the remaining capacities aren't normalized and their units can vary. You're right that it's an integer so my extreme example wouldn't hold, but I hope my point has come across. :) 

I firmly believe that using a magical number to avoid "divide by 0" is hacky no matter what value is chosen. I suggest throwing an exception when the remaining capacity is 0 and terminate the assignment right then and there. For code cleanness, you may want to consider using do while as suggested by @jiajunwang . I will also comment there to see if that solution makes sense.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
NealSun96: How about 
```
do {
...
}
while (!toBeAssignedReplicas.isEmpty())
```
?

xyuanlu: Thanks for the offline discussion. Changed to use `toBeAssignedReplicas.isEmpty()` in while evaluate expression.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +169,135 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
-          } else {
-            // Note we shall prioritize the replica with a higher state priority,
-            // the smaller priority number means higher priority.
-            return statePriority1 - statePriority2;
-          }
+  private Map<String, Integer> computeOverallClusterRemainingCapacity(List<AssignableNode> nodes) {
+    Map<String, Integer> utilizationMap = new HashMap<>();
+    for (AssignableNode node : nodes) {
+      for (String capacityKey : node.getMaxCapacity().keySet()) {
+        if (utilizationMap.containsKey(capacityKey)) {
```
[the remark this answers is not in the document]
xyuanlu: TFTR. Updated

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: We had an offline discussion with jiajunwang. Our final decision is:
1. function `computeScore` should add 0 to final score when replica's capacity on a dimension is 0.
2. function `computeScore` will return Integer.MAX_VALUE when cluster's remaining capacity is 0.
3. `computeScore` won't throw exception or do admission check (won't throw exception when cluster's remaining capacity is 0). Constrain check will fail this placement in the next step.
4. Also, to avoid overflow, the score is caped at 100 for each dimension.

dasahcc:
Question: a general question, seemed this sorting to be happened in every pipeline. Can we add some checks there? If no condition needed for such sorting, we skip it?

xyuanlu:
> Question: a general question, seemed this sorting to be happened in every pipeline. Can we add some checks there? If no condition needed for such sorting, we skip it?

Thanks for the question. 
In my understanding, we sort replicas for two reasons:
1. For every rebalance, the assignment of each replicas should be in the same sequence. This is already garnered in the original design.
2. We want to place replicas with larger impact first to solve the failure by placing very large partition after small partitions.  We only know the placement will fail afterward. So it is hard to know is we need to sort beforehand. 
Hope my explain is clear.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -42,6 +42,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static java.lang.Float.compare;
```
jiajunwang, line 45: nit, better directly call Float.compare in the code. Static import here is not quite desired.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +172,136 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+      Integer newRemainingCapacity =
+          overallClusterRemainingCapacityMap.get(resourceUsage.getKey()) - resourceUsage.getValue();
+      overallClusterRemainingCapacityMap.put(resourceUsage.getKey(), newRemainingCapacity);
+    }
+  }
+
+  private class AssignableReplicaWithScore implements Comparable<AssignableReplicaWithScore> {
+    private final AssignableReplica _replica;
+    private float _score = 0;
+    private boolean _isInBestPossibleAssignment;
```
jiajunwang, line 202: nit, final?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +172,136 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+          clusterModel.getContext().getBaselineAssignment().containsKey(replica.getResourceName());
+      _replicaHash = Objects.hash(replica.toString(), clusterModel.getAssignableNodes().keySet());
+    }
+
+    public void computeScore(Map<String, Integer> overallClusterRemainingCapMap) {
+      float score = 0;
+      // score = SUM(weight * (resource_capacity/cluster_capacity) where weight = 1/(1-total_util%)
+      // it could be be simplified to "resource_capacity/cluster_remainingCapacity".
+      for (Map.Entry<String, Integer> resourceCapacity : _replica.getCapacity().entrySet()) {
+        score = resourceCapacity.getValue() == 0 ? score
```
jiajunwang, line 220: nit, IMO, checking resourceCapacity.getValue() == 0 and skip the loop makes the code cleaner. And it helps to avoid the unnecessary score value update if the compiler does not optimize it automatically.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +172,136 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+  }
+
+  private class AssignableReplicaWithScore implements Comparable<AssignableReplicaWithScore> {
+    private final AssignableReplica _replica;
+    private float _score = 0;
+    private boolean _isInBestPossibleAssignment;
+    private boolean _isInBaselineAssignment;
+    private final Integer  _replicaHash;
+
+    AssignableReplicaWithScore(AssignableReplica replica, ClusterModel clusterModel) {
```
jiajunwang, line 206: Just extend AssignableReplica?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -68,23 +71,40 @@ public OptimalAssignment calculate(ClusterModel clusterModel) throws HelixRebala
...
+    // Create a wrapper for each AssignableReplica.
+    Set<AssignableReplicaWithScore> toBeAssignedReplicas =
+        clusterModel.getAssignableReplicaMap().values().stream()
+            .flatMap(replicas -> replicas.stream())
+            .map(replica -> new AssignableReplicaWithScore(replica, clusterModel))
+            .collect(Collectors.toSet());
+
+    while (!toBeAssignedReplicas.isEmpty()) {
+      AssignableReplicaWithScore replicaWithScore =
+          getNextAssignableReplica(toBeAssignedReplicas, overallClusterRemainingCapacityMap);
```
jiajunwang, line 88: Can we remove the selected replica in the getNextAssignableReplica() method? Maybe call it pollNextAssignableReplica()? Then I guess we can just use a linked list or some other more efficient data structure to make the loop faster.

jiajunwang (commented):
I left some minor comments. But overall, it looks good to me.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: Please correct me if I am wrong. We already have a list of AssignableReplica objects in clusterModel. Why creating many duplicated objects? I think wrapper is a better choice here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: Moving the remove part to getNextAssignableReplica is more cleaner.  However, based on our offline discussion, it is not easy to do remove the selected replica from linkedList/listArray with the same time complexity O(1). Will keep using set here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
xyuanlu: Updated.

NealSun96 (changes requested):
Bar the few comments, overall the changes look good. Great work @xyuanlu !

As a NIT: consider cleaning up some extra parentheses as they appear confusing.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -68,23 +69,39 @@ public OptimalAssignment calculate(ClusterModel clusterModel) throws HelixRebala
...
```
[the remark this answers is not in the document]
xyuanlu: Good suggestion. will update.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
```diff
@@ -152,69 +169,139 @@ private double getAssignmentNormalizedScore(AssignableNode node, AssignableRepli
...
+          clusterModel.getContext().getBaselineAssignment().containsKey(replica.getResourceName());
+      _replicaHash = Objects.hash(replica.toString(), clusterModel.getAssignableNodes().keySet());
+    }
+
+    public void computeScore(Map<String, Integer> overallClusterRemainingCapMap) {
+      float score = 0;
+      // score = SUM(weight * (resource_capacity/cluster_capacity) where weight = 1/(1-total_util%)
+      // it could be be simplified to "resource_capacity/cluster_remainingCapacity".
+      for (Map.Entry<String, Integer> resourceCapacity : _replica.getCapacity().entrySet()) {
+        if (resourceCapacity.getValue() == 0) {
```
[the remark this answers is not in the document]
xyuanlu: JJ mentioned a good point here and I agree with that.  "checking resourceCapacity.getValue() == 0 and skip the loop makes the code cleaner. And it helps to avoid the unnecessary score value update if the compiler does not optimize it automatically"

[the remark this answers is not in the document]
NealSun96: I think the next line's calculation encapsulates the mathematical logic completely (including when there's no resource_capacity), and doesn't need this extra check; imo this doesn't look as clean. But since there are 2 votes already, I'll oblige.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/constraints/ConstraintBasedAlgorithm.java
jiajunwang: This is just another way to wrap the class. My intention is just help you to reduce some of the methods like, toString(), getAssignableReplica() - since it is already an assignableReplia.

NealSun96 (approved):

xyuanlu:
This PR is ready to be merged. Approved by @NealSun96 

Final commit message:
Improve the WAGED.ConstraintBasedAlgorithm sorting logic to prioritize replica with larger impact.

PR #1697, opened 2021-04-13
Directly use Best Possible State to calculate DifferenceWithIdealStateGauge metrics instead of relying on the persisted IdealState.

### LINKED ISSUE #1670: IdealState does not appear to be written
### Describe the bug
There appears to be a bug with either the calculation of the "DifferenceWithIdealStateGauge" metric, or the writing of ideal state.  We recently went through an upgrade from Helix 0.6.9 to 1.0.1.  When viewing the current state of Helix `--listResourceInfo`, IdealState in both 0.6.9 and 1.0.1 appears to always be empty.  However, the metric calculation seems to have changed between the two versions.  In 0.6.9, "DifferenceWithIdealStateGauge" would be emitted as 0.  In 1.0.1, it will correctly emit an accurate metric as an empty IdealState with all partitions accurately assigned will always emit the total partitions as the "DifferenceWithIdealStateGauge" metric.

// Metric calculation in 1.0.1
[(1.0.1) ResourceMonitor.java](https://github.com/apache/helix/blob/helix-1.0.1/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java#L243)

// Metric calculation in 0.6.9
[(0.6.9) ResourceMonitor.java](https://github.com/apache/helix/blob/helix-0.6.9/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java#L98)

This could potentially be related to https://github.com/apache/helix/issues/609.

### To Reproduce
https://github.com/jeff-blaisdell/helix-issue contains a full working example.  Greater details on environment / cluster setup can be found in project README.

### Expected behavior
DifferenceWithIdealStateGauge should correctly report 0.

jiajunwang:
### Issues

Resolve #1670 

### Description

This PR aims to resolve the undesired dependencies between DifferenceWithIdealStateGauge metric and the PERSIST_XXXXXX_ASSIGNMENT configuration. Before the change, if the assignment is not persisted, then the metric will report incorrect data.
To fix the issue, the calculation must be moved from the ExternalViewComputeStage to the BestPossibleStateCalcStage. Because the Best Possible State is only available in the BestPossibleStateCalcStage assuming the PERSIST_XXXXXX_ASSIGNMENT option is not turned on. In addition to the main logic migration, other changes listed following are required to ensure multi-thread safety.
1. Concurrent control of the EV cache since it will now be read in the BestPossibleStateCalcStage in addition to the ExternalViewComputeStage.
2. Minor changes in the diff computing logic for the corner cases since the Best Possible State state mapping is not exactly the same as IdealState persist assignment.
3. Cleanup PersistAssignmentStage logic so it won't modify the IdealState cache anymore. It is supposed to be read-only.
4. Test cases are modified to cover the changes. Also, the metric is now reporting correct results in some corner cases such as all the nodes are disabled. The test cases which are testing based on the wrong behavior have been fixed.

### Tests

Modify TestClusterAggregateMetrics.java according to the new behavior.
Adding new checks in TestClusterStatusMonitor.java to validate the DifferenceWithIdealStateGauge result.

[INFO] Tests run: 1264, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,080.412 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 1264, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:24 h
[INFO] Finished at: 2021-04-13T20:15:05-07:00

Note that TestCleanupExternalView is not stable currently. It frequently fails in the master branch too. We will take care of that separately.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
@@ -205,7 +205,9 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
     // As we are not listening on external view change, external view will be
     // refreshed once during the cache's first refresh() call, or when full refresh is required
     if (_propertyDataChangedMap.get(HelixConstants.ChangeType.EXTERNAL_VIEW).getAndSet(false)) {
-      _externalViewCache.refresh(accessor);
+      synchronized (_externalViewCache) {
+        _externalViewCache.refresh(accessor);
+      }
     }
   }
 
@@ -269,7 +271,9 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
    * @return
    */
   public Map<String, ExternalView> getExternalViews() {
-    return _externalViewCache.getPropertyMap();
+    synchronized (_externalViewCache) {
+      return _externalViewCache.getPropertyMap();
+    }
   }
 
   /**
@@ -277,8 +281,10 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
    * @param externalViews
    */
   public void updateExternalViews(List<ExternalView> externalViews) {
-    for (ExternalView ev : externalViews) {
-      _externalViewCache.setProperty(ev);
+    synchronized (_externalViewCache) {
+      for (ExternalView ev : externalViews) {
+        _externalViewCache.setProperty(ev);
+      }
     }
   }
 
@@ -312,8 +318,10 @@ public class ResourceControllerDataProvider extends BaseControllerDataProvider {
    */
 
   public void removeExternalViews(List<String> resourceNames) {
-    for (String resourceName : resourceNames) {
-      _externalViewCache.deletePropertyByName(resourceName);
+    synchronized (_externalViewCache) {
+      for (String resourceName : resourceNames) {
+        _externalViewCache.deletePropertyByName(resourceName);
+      }
     }
   }
 
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/PersistAssignmentStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/PersistAssignmentStage.java
@@ -30,7 +30,6 @@ import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixException;
 import org.apache.helix.HelixManager;
 import org.apache.helix.PropertyKey;
-import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.controller.LogUtil;
 import org.apache.helix.controller.common.PartitionStateMap;
 import org.apache.helix.controller.dataproviders.ResourceControllerDataProvider;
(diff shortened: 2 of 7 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/PersistAssignmentStage.java
```diff
@@ -128,25 +130,30 @@ private void persistAssignment(final Resource resource, final ResourceController
...
+            ZNRecord deltaRecord = delta.getRecord();
+            // Overwrite MapFields and ListFields items with the same key.
+            if (!deltaRecord.getMapFields().isEmpty()) {
+              // Note that default merge will keep old values in the maps unchanged, which is not desired.
               current.getMapFields().clear();
-              current.getMapFields().putAll(idealState.getRecord().getMapFields());
-              current.getListFields().putAll(idealState.getRecord().getListFields());
+              current.getMapFields().putAll(deltaRecord.getMapFields());
+            }
+            if (!deltaRecord.getListFields().isEmpty()) {
```
dasahcc, line 149: Question: why we need this check? We can add lists field anyway just not clear it up, right?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/ResourceControllerDataProvider.java
```diff
@@ -205,7 +205,9 @@ private void refreshExternalViews(final HelixDataAccessor accessor) {
     // As we are not listening on external view change, external view will be
     // refreshed once during the cache's first refresh() call, or when full refresh is required
     if (_propertyDataChangedMap.get(HelixConstants.ChangeType.EXTERNAL_VIEW).getAndSet(false)) {
-      _externalViewCache.refresh(accessor);
+      synchronized (_externalViewCache) {
+        _externalViewCache.refresh(accessor);
```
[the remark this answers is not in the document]
jiajunwang: That would be too heavy. And I'm almost 100% sure it will impact performance since the cache is used everywhere in the pipeline.

helix-core/src/main/java/org/apache/helix/controller/stages/PersistAssignmentStage.java
jiajunwang: This is for backward compatibility. Not adding it will be fine, logic-wise. But it will change the behavior, so some tests fail. This is because the change in this PR changes the input.

Since adding this check also makes sense to me, I decide to keep the same behavior.
What's your opinion?

dasahcc: I agree better to keep the behavior. But the original code does not perform the check, right? The check changed the behavior of our code that only list field contains something then we add. 

Either way is fine. I dont have strong preference because the only thing I can think about is custom mode / non rebalanced resource does not have the list field.  For non rebalanced resource, it will be eventually caught up if the it get rebalanced with list field fulfilled.

dasahcc (approved):
overall, lgtm

helix-core/src/main/java/org/apache/helix/controller/stages/PersistAssignmentStage.java
jiajunwang: Since I changed the logic, if I pass an empty map through the "deltaRecord", it means don't update. Originally it means the IdealState should have the empty map. So the meaning of the input has been changed. So this check will ensure the behavior keeps the same.

jiajunwang:
Approved by @dasahcc, I will merge later.

PR #1701, opened 2021-04-16
Improve WAGED simulation API by throwing exception on FAILED_TO_CALCULATE

### LINKED ISSUE #1700: Improve WAGED simulation API to reflect calculation failure
WAGED simulation API has been used to predict rebalance outcome. However, one critical feature it lacks is that it cannot reflect "FAILED_TO_CALCULATE" except for an error log. This is because the simulation API calls Helix pipelines directly to calculate the result, but the pipelines don't propagate "FAILED_TO_CALCULATE" exception; instead, the pipelines process it and return the last known good result when it happens. 

The simulation API should be able to propagate the "FAILED_TO_CALCULATE" exception so that the caller can be notified when this happens. That way, the simulation API may be used to alert future calculation failure. 

**Additional context**

NealSun96:
### Issues

Resolves #1700 

### Description

When WAGED encounters a "FAILED_TO_CALCULATE" exception, the exception is processed by the pipeline without propagating. This is by design because WAGED doesn't consider this exception as pipeline breaking - WAGED simply returns the last known good result instead and continue the pipeline. 

WAGED simulation API calls the pipelines to get a simulated rebalance result, and as a result, it also cannot reflect "FAILED_TO_CALCULATE" except for an error log. 

Down the line, we would like the simulation API to be used to forecast potential calculation failure: when a customer makes partition weight changes, for example, they can first use the API to see the aftermath of such change, and don't do the changes if they can result calculation failures. 

Therefore, we need the simulation API to have the ability to propagate "FAILED_TO_CALCULATE" exception. 

### Tests

```
TestWagedRebalancer.testAlgorithmException is modified. 
```

```
[INFO] Tests run: 1265, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,170.788 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] 
[INFO] Loading execution data file /home/nesun/helix/helix-core/target/jacoco.exec
[INFO] Analyzed bundle 'Apache Helix :: Core' with 890 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  01:26 h
[INFO] Finished at: 2021-05-04T15:07:24-07:00

```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
@@ -19,10 +19,12 @@ package org.apache.helix.controller.rebalancer.waged;
  * under the License.
  */
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.helix.HelixRebalanceException;
 import org.apache.helix.controller.dataproviders.ResourceControllerDataProvider;
@@ -52,7 +54,8 @@ public class ReadOnlyWagedRebalancer extends WagedRebalancer {
   @Override
   protected List<HelixRebalanceException.Type> failureTypesToPropagate() {
     // Also propagate FAILED_TO_CALCULATE for ReadOnlyWagedRebalancer
-    List<HelixRebalanceException.Type> failureTypes = super.failureTypesToPropagate();
+    List<HelixRebalanceException.Type> failureTypes =
+        new ArrayList<>(super.failureTypesToPropagate());
     failureTypes.add(HelixRebalanceException.Type.FAILED_TO_CALCULATE);
     return failureTypes;
   }
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
@@ -19,6 +19,8 @@ package org.apache.helix.controller.rebalancer.waged;
  * under the License.
  */
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -49,8 +51,11 @@ public class ReadOnlyWagedRebalancer extends WagedRebalancer {
   }
 
   @Override
-  protected boolean fallbackOnFailedCalculate() {
-    return false;
+  protected List<HelixRebalanceException.Type> failureTypesToPropagate() {
+    // Also propagate FAILED_TO_CALCULATE for ReadOnlyWagedRebalancer
+    return Arrays.asList(HelixRebalanceException.Type.INVALID_REBALANCER_STATUS,
+        HelixRebalanceException.Type.UNKNOWN_FAILURE,
+        HelixRebalanceException.Type.FAILED_TO_CALCULATE);
   }
 
   private static class ReadOnlyAssignmentMetadataStore extends AssignmentMetadataStore {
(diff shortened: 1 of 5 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
```diff
@@ -48,6 +50,14 @@ public ReadOnlyWagedRebalancer(String metadataStoreAddress, String clusterName,
         ConstraintBasedAlgorithmFactory.getInstance(preferences), Optional.empty());
   }
 
+  @Override
+  protected List<HelixRebalanceException.Type> failureTypesToPropagate() {
+    // Also propagate FAILED_TO_CALCULATE for ReadOnlyWagedRebalancer
+    return Arrays.asList(HelixRebalanceException.Type.INVALID_REBALANCER_STATUS,
```
jiajunwang, line 56: nit, something like returning super.failureTypesToPropagate().add(HelixRebalanceException.Type.FAILED_TO_CALCULATE) ?

jiajunwang:
One more thing, please add a test for the tool method to cover the fail to rebalance case. This is the target usage, right?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java
NealSun96: I don't think that's better because ReadOnlyWagedRebalancer doesn't always propagate a superset of failure types; that might change in the future. Separate lists are clearer.

jiajunwang: I think ReadOnlyWagedRebalancer should cover all the failure types of the regular rebalancer. Otherwise, what to return if the basic rebalancer cannot finish the calculation? Based on this assumption, adding an additional item to the list would be easier to be maintained.

NealSun96: I disagree with this: for example, some extension of `WagedRebalancer (if not `ReadOnlyWagedRebalancer`) could think UNKNOWN_FAILURE is okay and just take the last-known good result instead of an exception. 

I understand where you come from: if we add a new HelixRebalanceException.Type, it seems easier if we do super sets so we just modify the base class. However, even if we introduce a new type, each implementation should be reviewed to see if that type should be propagated. Imo this coupling is unnecessary here. If you don't feel strongly about this, I'd like to stick to separate lists.

jiajunwang: 1. Let's worry about another extension of the WAGED rebalancer when they are created. And no matter what you do for the ReadOnlyWagedRebalancer, you can do it differently according to the new requirement later. Not relevant.
2. For ReadOnlyWagedRebalancer, I don't think we need to revisit. We can just carry over whatever WagedRebalancer has plus the FAILED_TO_CALCULATE failure type. If this is not guaranteed, then ReadOnlyWagedRebalancer cannot be used in the simulate tool, since it may cause different behavior. That's exactly why I strongly prefer we just call the method of super class here. It prevents potential bugs in the future.

NealSun96: I can see your point about ReadOnlyWagedRebalancer now. Making changes accordingly.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -393,6 +394,12 @@ public void close() {
     return finalIdealStateMap;
   }
 
+  protected List<HelixRebalanceException.Type> failureTypesToPropagate() {
+    // By default, only propagate the following types of HelixRebalanceException
+    return Arrays.asList(HelixRebalanceException.Type.INVALID_REBALANCER_STATUS,
```
jiajunwang, line 399: nit, let's define a static unmodifiable list for the WagedRebalancer class to avoid creating a list on each call. Then we can always create a new list in the ReadOnlyWagedRebalancer.java. Considering ReadOnlyWagedRebalancer.java is less used and does not impact the critical pipeline time, it is better to create new list there.

jiajunwang (approved):

NealSun96:
Final commit message:
## Improve WAGED simulation API by throwing exception on FAILED_TO_CALCULATE ##
WAGED simulation API now throws HelixException upon encountering FAILED_TO_CALCULATE

PR #1703, opened 2021-04-19
[Replica Level Throttle] Add per replica rebalance type compute logic

dasahcc:
### Issues

resolve #343 

### Description

The first step for per replica level throttling is supporting logic to determine what message is recovery/rebalance or load rebalance.
Three functions added: 1) rebalance type computation required state.
                                        2) rebalance type per message
                                        3) message sorting rules and comparators to determine which message to apply first.

For global change contexts please refer: https://github.com/apache/helix/pull/1713
### Tests

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
@@ -29,6 +29,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import org.apache.helix.HelixDefinedState;
 import org.apache.helix.HelixException;
 import org.apache.helix.HelixManager;
@@ -43,6 +45,7 @@ import org.apache.helix.model.BuiltInStateModelDefinitions;
 import org.apache.helix.model.ClusterConfig;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.MaintenanceSignal;
+import org.apache.helix.model.Message;
 import org.apache.helix.model.Partition;
 import org.apache.helix.model.Resource;
 import org.apache.helix.model.StateModelDefinition;
@@ -795,6 +798,55 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
     }
   }
 
+  private RebalanceType getRebalanceTypePerMessage(Map<String, Integer> requiredStates, Message message,
+      Map<String, String> derivedCurrentStates) {
+    Map<String, Integer> requiredStatesSnapshot = new HashMap<>(requiredStates);
+    // Looping existing current states to see whether current states fulfilled all the required states.
+    for (String state : derivedCurrentStates.values()) {
+      if (requiredStatesSnapshot.containsKey(state)) {
+        requiredStatesSnapshot.put(state, requiredStatesSnapshot.get(state) - 1);
+        if (requiredStatesSnapshot.get(state) == 0) {
+          requiredStatesSnapshot.remove(state);
+        }
+      }
+    }
+
+    // If the message is trying to bring the required state remaining in the map, it is recovery rebalance.
+    // Otherwise it is load rebalance.
+    return requiredStatesSnapshot.containsKey(message.getToState()) ? RebalanceType.RECOVERY_BALANCE
+        : RebalanceType.LOAD_BALANCE;
+  }
+
+  private Map<String, Integer> getRequiredStates(String resourceName,
+      ResourceControllerDataProvider resourceControllerDataProvider, List<String> preferenceList) {
+
+    // Prepare required inputs: 1) Priority State List 2) required number of replica
+    IdealState idealState = resourceControllerDataProvider.getIdealState(resourceName);
+    StateModelDefinition stateModelDefinition =
+        resourceControllerDataProvider.getStateModelDef(idealState.getStateModelDefRef());
+    int requiredNumReplica = idealState.getMinActiveReplicas() == -1
+        ? idealState.getReplicaCount(preferenceList.size())
+        : idealState.getMinActiveReplicas();
+    Set<String> activeList = new HashSet<>(preferenceList);
+    activeList.retainAll(resourceControllerDataProvider.getEnabledLiveInstances());
+
+    // For each state, check that this partition currently has the required number of that state as
+    // required by StateModelDefinition.
+    LinkedHashMap<String, Integer> expectedStateCountMap =
+        stateModelDefinition.getStateCountMap(activeList.size(), requiredNumReplica); // StateModelDefinition's counts
+
+    Map<String, Integer> requiredStates = new HashMap<>();
+    for (String state : stateModelDefinition.getStatesPriorityList()) {
+      if (requiredNumReplica <= 0) {
+        break;
+      }
+
+      requiredStates.put(state, Math.min(requiredNumReplica, expectedStateCountMap.get(state)));
+      requiredNumReplica -= requiredStates.get(state);
+    }
+    return requiredStates;
+  }
+
   /**
    * Log rebalancer metadata for debugging purposes.
    * @param resource
@@ -874,7 +926,31 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
     }
   }
 
-  // Compare partitions according following standard:
+  private class MessagePriorityComparator implements Comparator<Message> {
+    private Map<String, Integer> _preferenceInstanceMap;
+    private Map<String, Integer> _statePriorityMap;
+
+    MessagePriorityComparator(List<String> preferenceList, Map<String, Integer> statePriorityMap) {
+      // Get instance -> priority map.
+      _preferenceInstanceMap = IntStream.range(0, preferenceList.size())
+          .boxed()
+          .collect(Collectors.toMap(preferenceList::get, index -> index));
+      _statePriorityMap = statePriorityMap;
+    }
+
+    @Override
+    public int compare(Message m1, Message m2) {
+      //Compare rules:
+      //     1. Higher target state has higher priority.
+      //     2. If target state is same, range it as preference list order.
+      if (m1.getToState().equals(m2.getToState())) {
+        return _preferenceInstanceMap.get(m1.getTgtName()).compareTo(_preferenceInstanceMap.get(m2.getTgtName()));
+      }
+      return _statePriorityMap.get(m1.getToState()).compareTo(_statePriorityMap.get(m2.getToState()));
+    }
+  }
+
+    // Compare partitions according following standard:
   // 1) Partition without top state always is the highest priority.
   // 2) For partition with top-state, the more number of active replica it has, the less priority.
   private class PartitionPriorityComparator implements Comparator<Partition> {
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
@@ -798,21 +798,29 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
     }
   }
 
-  private RebalanceType getRebalanceTypePerMessage(Map<String, Integer> requiredStates, Message message,
+  /**
+   * Determine the message rebalance type with message and current states.
+   * @param desiredStates         Ideally how may states we needed for guarantee the health of replica
+   * @param message               The message to be determined what is the rebalance type
+   * @param derivedCurrentStates  Derived from current states with previous messages not be throttled.
+   * @return                      Rebalance type. Recovery or load.
+   */
+  private RebalanceType getRebalanceTypePerMessage(Map<String, Integer> desiredStates, Message message,
       Map<String, String> derivedCurrentStates) {
-    Map<String, Integer> requiredStatesSnapshot = new HashMap<>(requiredStates);
+    Map<String, Integer> requiredStatesSnapshot = new HashMap<>(desiredStates);
     // Looping existing current states to see whether current states fulfilled all the required states.
     for (String state : derivedCurrentStates.values()) {
       if (requiredStatesSnapshot.containsKey(state)) {
-        requiredStatesSnapshot.put(state, requiredStatesSnapshot.get(state) - 1);
-        if (requiredStatesSnapshot.get(state) == 0) {
+        if (requiredStatesSnapshot.get(state) == 1) {
           requiredStatesSnapshot.remove(state);
+        } else {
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -795,6 +798,54 @@ private RebalanceType getRebalanceType(ResourceControllerDataProvider cache,
     }
   }
 
+  private RebalanceType getRebalanceTypePerMessage(Map<String, Integer> requiredStates, Message message,
```
alirezazamani, line 801: This method has not been used. Right? Is it because it is skeleton PR?

dasahcc: Not used. I was trying to make each PR independent and easy to review. Then we can focus on the logic of the content itself.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -795,6 +798,55 @@ private RebalanceType getRebalanceType(ResourceControllerDataProvider cache,
...
+    Map<String, Integer> requiredStatesSnapshot = new HashMap<>(requiredStates);
+    // Looping existing current states to see whether current states fulfilled all the required states.
+    for (String state : derivedCurrentStates.values()) {
+      if (requiredStatesSnapshot.containsKey(state)) {
+        requiredStatesSnapshot.put(state, requiredStatesSnapshot.get(state) - 1);
+        if (requiredStatesSnapshot.get(state) == 0) {
+          requiredStatesSnapshot.remove(state);
+        }
+      }
+    }
```
jiajunwang, line 812: This calculation would be done multiple times if a partition has more than one message. Should we just pre-calculate the requiredStatesSnapshot (which is basically the gap between the derived current state and required states) and then use the snapshot? Obviously, we need to update the snapshot once a message is confirmed. But that calculation would be much faster than loop these 2 types of maps every time.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -874,7 +926,31 @@ public int compare(ResourcePriority priority1, ResourcePriority priority2) {
...
+      _statePriorityMap = statePriorityMap;
+    }
+
+    @Override
+    public int compare(Message m1, Message m2) {
+      //Compare rules:
+      //     1. Higher target state has higher priority.
+      //     2. If target state is same, range it as preference list order.
+      if (m1.getToState().equals(m2.getToState())) {
+        return _preferenceInstanceMap.get(m1.getTgtName()).compareTo(_preferenceInstanceMap.get(m2.getTgtName()));
```
jiajunwang, line 947: A message might target an instance that does not exist in the preference list.

jiajunwang: For safety and future extension, maybe it is a good idea to sort the messages by name as the last resort.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -795,6 +798,55 @@ private RebalanceType getRebalanceType(ResourceControllerDataProvider cache,
...
 
+  private RebalanceType getRebalanceTypePerMessage(Map<String, Integer> requiredStates, Message message,
+      Map<String, String> derivedCurrentStates) {
+    Map<String, Integer> requiredStatesSnapshot = new HashMap<>(requiredStates);
+    // Looping existing current states to see whether current states fulfilled all the required states.
+    for (String state : derivedCurrentStates.values()) {
+      if (requiredStatesSnapshot.containsKey(state)) {
+        requiredStatesSnapshot.put(state, requiredStatesSnapshot.get(state) - 1);
+        if (requiredStatesSnapshot.get(state) == 0) {
+          requiredStatesSnapshot.remove(state);
```
narendly, line 809: You can potentially save a `put()` call by
```
int requiredStateCount = requiredStatesSnapshot.get(state);
if (--requiredStateCount == 0) {
  requiredStatesSnapshot.remove(state);
} else {
  requiredStatesSnapshot.put(state, requiredStateCount);
}
```

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -795,6 +798,55 @@ private RebalanceType getRebalanceType(ResourceControllerDataProvider cache,
     }
   }
 
+  private RebalanceType getRebalanceTypePerMessage(Map<String, Integer> requiredStates, Message message,
+      Map<String, String> derivedCurrentStates) {
+    Map<String, Integer> requiredStatesSnapshot = new HashMap<>(requiredStates);
+    // Looping existing current states to see whether current states fulfilled all the required states.
+    for (String state : derivedCurrentStates.values()) {
```
narendly, line 805: Did we agree that something like `desiredStateCounts` is a better description?

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -795,6 +798,55 @@ private RebalanceType getRebalanceType(ResourceControllerDataProvider cache,
...
+      if (requiredStatesSnapshot.containsKey(state)) {
+        requiredStatesSnapshot.put(state, requiredStatesSnapshot.get(state) - 1);
+        if (requiredStatesSnapshot.get(state) == 0) {
+          requiredStatesSnapshot.remove(state);
+        }
+      }
+    }
+
+    // If the message is trying to bring the required state remaining in the map, it is recovery rebalance.
+    // Otherwise it is load rebalance.
```
narendly, line 815: We could maybe reword this: 

"If the message contains any "required" state changes, then it is considered recovery rebalance. Otherwise, it is load balance."

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -795,6 +798,55 @@ private RebalanceType getRebalanceType(ResourceControllerDataProvider cache,
...
+
+    // Prepare required inputs: 1) Priority State List 2) required number of replica
+    IdealState idealState = resourceControllerDataProvider.getIdealState(resourceName);
+    StateModelDefinition stateModelDefinition =
+        resourceControllerDataProvider.getStateModelDef(idealState.getStateModelDefRef());
+    int requiredNumReplica = idealState.getMinActiveReplicas() == -1
+        ? idealState.getReplicaCount(preferenceList.size())
+        : idealState.getMinActiveReplicas();
+    Set<String> activeList = new HashSet<>(preferenceList);
+    activeList.retainAll(resourceControllerDataProvider.getEnabledLiveInstances());
```
narendly, line 831: I think we could avoid doing a deep copy since we only seem to need the size of live instances in the preference list?

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -795,6 +798,55 @@ private RebalanceType getRebalanceType(ResourceControllerDataProvider cache,
...
+    // If the message is trying to bring the required state remaining in the map, it is recovery rebalance.
+    // Otherwise it is load rebalance.
+    return requiredStatesSnapshot.containsKey(message.getToState()) ? RebalanceType.RECOVERY_BALANCE
+        : RebalanceType.LOAD_BALANCE;
+  }
+
+  private Map<String, Integer> getRequiredStates(String resourceName,
+      ResourceControllerDataProvider resourceControllerDataProvider, List<String> preferenceList) {
+
+    // Prepare required inputs: 1) Priority State List 2) required number of replica
```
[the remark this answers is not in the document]
dasahcc: I split the PR for easy review. Finally, I will clear up the unused the method.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -795,6 +798,55 @@ private RebalanceType getRebalanceType(ResourceControllerDataProvider cache,
...
+        resourceControllerDataProvider.getStateModelDef(idealState.getStateModelDefRef());
+    int requiredNumReplica = idealState.getMinActiveReplicas() == -1
+        ? idealState.getReplicaCount(preferenceList.size())
+        : idealState.getMinActiveReplicas();
+    Set<String> activeList = new HashSet<>(preferenceList);
+    activeList.retainAll(resourceControllerDataProvider.getEnabledLiveInstances());
+
+    // For each state, check that this partition currently has the required number of that state as
+    // required by StateModelDefinition.
+    LinkedHashMap<String, Integer> expectedStateCountMap =
```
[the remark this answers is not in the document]
dasahcc: Good catch!

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
dasahcc: No. We cannot, because the required state can be update and will be more complicated just update the required state. For example, in the state of: required state 1 LEADER, 1 FOLLOWER
1: FOLLOWER
2: OFFLINE

required state : 1 LEADER
with first message FOLLOWER -> LEADER applied.
required state : 1 FOLLOWER

So with this message apply, you need to give back the FOLLOWER. It will be complicated to determine when we should give the state back to required and when is not. I am not saying we cannot do that but just make the logic more complicated.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
dasahcc: I can make the sort check for that. But we dont need sort it by name. Because if the instance does not exist in preference list, it means the message for this instance should be downward or something. Then sort by state of target state will make it last priority.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
dasahcc: No. We need the enabled ones. If we dont do deep copy, the original data will be changed. This is something we learned from previous building IntermediateStage.

narendly: I think what you don't want is a concurrent modification exception here.

Let's see... `getEnabledLiveInstances()` already gives you a deep-copied set. And we only need an activeList size. Where are we getting this `preferenceList` from? Isn't that a deep-copy in and of itself? I can't seem to find its usage in this PR. I am suggesting doing something like `stream().filter().count()` without having to do a deep copy for each resource, which could be potentially expensive.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
jiajunwang: The reason for sorting by naming is more deterministic. I agree that there is no solid reason that we have to sort by the names. But it doesn't hurt and can help to avoid any randomness which might be annoying.

BTW, the first problem that I mentioned is not related to sorting. I believe an NPE will be thrown if you don't do a null check.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -795,6 +798,54 @@ private RebalanceType getRebalanceType(ResourceControllerDataProvider cache,
...
+  /**
+   * Determine the message rebalance type with message and current states.
+   * @param desiredStates         Ideally how may states we needed for guarantee the health of replica
+   * @param message               The message to be determined what is the rebalance type
+   * @param derivedCurrentStates  Derived from current states with previous messages not be throttled.
+   * @return                      Rebalance type. Recovery or load.
+   */
+  private RebalanceType getRebalanceTypePerMessage(Map<String, Integer> desiredStates, Message message,
+      Map<String, String> derivedCurrentStates) {
+    Map<String, Integer> requiredStatesSnapshot = new HashMap<>(desiredStates);
```
alirezazamani, line 810: Since this Map is created based on desiredState, how about we call it desiredStatesSnapshot?

alirezazamani (approved):
Overall looks good to me.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
jiajunwang: I think the time complexities of the 2 methods are very different. O(n^2): (n + n-1 + n-2...) vs, O(n). I get the point that logic would be a little bit complicated, but I would still prefer updating the state map on the fly instead repeat the same calculation.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -795,6 +798,54 @@ private RebalanceType getRebalanceType(ResourceControllerDataProvider cache,
...
+    StateModelDefinition stateModelDefinition =
+        resourceControllerDataProvider.getStateModelDef(idealState.getStateModelDefRef());
+    int requiredNumReplica = idealState.getMinActiveReplicas() == -1
+        ? idealState.getReplicaCount(preferenceList.size())
+        : idealState.getMinActiveReplicas();
+    Set<String> activeList = new HashSet<>(preferenceList);
+    activeList.retainAll(resourceControllerDataProvider.getEnabledLiveInstances());
+
+    // For each state, check that this partition currently has the required number of that state as
+    // required by StateModelDefinition.
```
alirezazamani, line 842: I couldn't quite understand the comment here. Can you please simplify it a little bit?

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
dasahcc: Synced offline. This could involve more complicated logic such as return state back over the size of the what it required. Then the following computation will not be accurate. Will keep this as it and see whether it could cause more delay in computation.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
dasahcc: Yes. It will not be modified by other stages. ConcurrentModification should not happen.

dasahcc:

PR #1714, opened 2021-04-23
[Replica Level Throttle] Change throttling logic to per message

dasahcc:
### Issues

resolves #343

### Description

Apply the logic for throttling with per message quota charge. This method will be leveraged for load rebalance / recover rebalance throttling in main logic.

Original function is used for charge the throttling based on partition.

### Tests

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
@@ -631,104 +631,55 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
   }
 
   /**
-   * Check the status on throttling at every level (cluster, resource, instance) and set
-   * intermediatePartitionStateMap accordingly per partition.
-   * @param throttleController
-   * @param resourceName
-   * @param partition
-   * @param currentStateOutput
-   * @param bestPossiblePartitionStateMap
-   * @param partitionsThrottled
-   * @param intermediatePartitionStateMap
-   * @param rebalanceType
-   * @param cache
+   * Check the status for a single message on throttling at every level (cluster, resource, replica) and set
+   * intermediatePartitionStateMap accordingly for that replica.
+   * @param throttleController                throttle controller object for throttling quota
+   * @param resourceName                      the resource for throttling check
+   * @param partition                         the partition for throttling check
+   * @param messageToThrottle                 the message to be throttled
+   * @param messagesThrottled                 the cumulative set of messages that have been throttled already. These
+   *                                          messages represent the replicas of this partition that have been throttled.
+   * @param intermediatePartitionStateMap     the cumulative partition-state mapping as a result of the throttling step
+   *                                          of IntermediateStateCalcStage
+   * @param rebalanceType                     the rebalance type to charge quota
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -631,104 +631,53 @@ private void chargePendingTransition(Resource resource, CurrentStateOutput curre
...
+          hasReachedThrottlingLimit = true;
+          if (logger.isDebugEnabled()) {
+            LogUtil.logDebug(logger, _eventId, String.format(
+                "Throttled because of instance level quota is full on instance {%s} for message {%s} of partition {%s} in resource {%s}",
+                messageToThrottle.getId(), messageToThrottle.getTgtName(), partition.getPartitionName(), resourceName));
           }
         }
       }
     }
+    // If there is a room to hold existing messages, charge this message.
```
narendly, line 671: Nit: did you mean: "If there is still room for this replica, proceed to charge at the cluster and resource level and set the intermediate partition-state mapping so that the state transition will be generated."

Question: please confirm with me if my understanding is correct: we do not have `chargeReplica()`; instead we incorporate the replica-level throttling by having each message represent a replica.

So the number of calls for `throttleStateTransitionsForPartition()` would be the total number of partitions in the cluster whereas the number of calls for `throttleStateTransitionsForReplica()` would be the total number of replicas in the cluster, correct?

On an average case, we will have 3x replicas as partitions (because applications usually use 3 as the replication factor). I am curious, are you able to do a quick local benchmark of the `IntermediateStateCalcStage` for before & after for a production-grade cluster? It might help us see if we are unwittingly creating a bottleneck here somewhere.

dasahcc: For your question:

1. Yes, even this function is called more times than single partition, there is no cost operations, such as new object creation / deletion, data collection looping. The only operations are read/update the map in throttle controllers and intermediate state map.
2. Usually you will have 3x as the one partition for throttling. Because a lot of messages have dependencies, which are removed by MessageSelectionStage. From historical data in production log, you can see that the number of messages is far less than what we have now.

Finally, as I said, this is non avoidable operation. Because each of the throttling charge operation success or fail can lead different result of next message rebalance type. The only thing we can do is trying to make sure no cost operations for this simple function call and that is also the reason I am trying to let this function to be reviewed independently instead of combing it into a big logic code.

narendly (approved):

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
narendly: @dasahcc 

I agree that if we want to do replica-based throttling, this is something that would be needed considering the way we throttle. 

Have we, though, considered making this optional/configurable? For example, many users of Helix do not require throttling capability at this level of granularity?

dasahcc: That's a good suggestion! If user does not enable the throttling, we can skip this stage even. I will add this to following PR.

PR #1718, opened 2021-04-28
Clarify error for ZkSessionMismatchedException in ZkAsyncCallbacks

### LINKED ISSUE #1717: Improve the error message for ZkSessionMismatchedException in ZkAsyncCallbacks
In ZkAsyncCallbacks, when an operation fails due to a ZkSessionMismatchedException, a NullPointerException is thrown which can cause confusion in the logs.

Improve the error message related to this issue in ZkAsyncCallbacks.

rabashizade:
### Issues

Fixes #1717 

### Description

When there is a ZK session mismatch for async operations in ZkClient, passing UNKNOWN_RET_CODE to the appropriate ZkAsyncCallback and not checking for this condition in the needRetry() method in the ZkAsyncCallback results in throwing a NullPointerException which can be confusing. This PR defines an additional return code for the ZK session mismatch situation and by checking for this condition in the needRetry() method, provides a more meaningful message in the log.

### Tests

helix-core:
```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestExpandCluster.testClusterExpansion:61 expected:<false> but was:<true>
[INFO] 
[ERROR] Tests run: 1265, Failures: 1, Errors: 0, Skipped: 2
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  01:56 h
[INFO] Finished at: 2021-04-28T09:57:01-07:00
```

TestExpandCluster:
```
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 94.638 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 
[INFO] 
[INFO] Loading execution data file /Users/rabashiz/IdeaProjects/helix/helix-core/target/jacoco.exec
[INFO] Analyzed bundle 'Apache Helix :: Core' with 890 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  01:41 min
[INFO] Finished at: 2021-04-28T10:01:06-07:00
```

zookeeper-api:
```
[INFO] Results:
[INFO] 
[INFO] Tests run: 54, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] --- jacoco-maven-plugin:0.8.6:report (generate-code-coverage-report) @ zookeeper-api ---
[INFO] Loading execution data file /Users/rabashiz/IdeaProjects/helix/zookeeper-api/target/jacoco.exec
[INFO] Analyzed bundle 'Apache Helix :: ZooKeeper API' with 115 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  09:18 min
[INFO] Finished at: 2021-04-28T10:16:51-07:00

```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -1962,8 +1962,10 @@ public class ZkClient implements Watcher {
       });
     } catch (RuntimeException e) {
       // Process callback to release caller from waiting
-      cb.processResult(ZkAsyncCallbacks.UNKNOWN_RET_CODE, path,
-          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+      int rc =
+          (e instanceof ZkSessionMismatchedException) ? ZkAsyncCallbacks.ZK_SESSION_MISMATCHED_CODE
+              : ZkAsyncCallbacks.UNKNOWN_RET_CODE;
+      cb.processResult(rc, path, new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
       throw e;
     }
   }
@@ -2000,8 +2002,10 @@ public class ZkClient implements Watcher {
       });
     } catch (RuntimeException e) {
       // Process callback to release caller from waiting
-      cb.processResult(ZkAsyncCallbacks.UNKNOWN_RET_CODE, path,
-          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+      int rc =
+          (e instanceof ZkSessionMismatchedException) ? ZkAsyncCallbacks.ZK_SESSION_MISMATCHED_CODE
+              : ZkAsyncCallbacks.UNKNOWN_RET_CODE;
+      cb.processResult(rc, path, new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
       throw e;
     }
   }
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -1960,12 +1960,15 @@ public class ZkClient implements Watcher {
                 });
         return null;
       });
+    } catch (ZkSessionMismatchedException e) {
+      // Process callback to release caller from waiting
+      cb.processResult(ZkAsyncCallbacks.ZK_SESSION_MISMATCHED_CODE, path,
+          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+      throw e;
     } catch (RuntimeException e) {
       // Process callback to release caller from waiting
-      int rc =
-          (e instanceof ZkSessionMismatchedException) ? ZkAsyncCallbacks.ZK_SESSION_MISMATCHED_CODE
-              : ZkAsyncCallbacks.UNKNOWN_RET_CODE;
-      cb.processResult(rc, path, new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+      cb.processResult(ZkAsyncCallbacks.UNKNOWN_RET_CODE, path,
+          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
       throw e;
     }
   }
@@ -2000,12 +2003,15 @@ public class ZkClient implements Watcher {
             });
         return null;
       });
+    } catch (ZkSessionMismatchedException e) {
+      // Process callback to release caller from waiting
+      cb.processResult(ZkAsyncCallbacks.ZK_SESSION_MISMATCHED_CODE, path,
+          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+      throw e;
     } catch (RuntimeException e) {
       // Process callback to release caller from waiting
-      int rc =
-          (e instanceof ZkSessionMismatchedException) ? ZkAsyncCallbacks.ZK_SESSION_MISMATCHED_CODE
-              : ZkAsyncCallbacks.UNKNOWN_RET_CODE;
-      cb.processResult(rc, path, new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+      cb.processResult(ZkAsyncCallbacks.UNKNOWN_RET_CODE, path,
+          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
       throw e;
     }
   }
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
@@ -36,6 +36,11 @@ import org.slf4j.LoggerFactory;
 public class ZkAsyncCallbacks {
   private static Logger LOG = LoggerFactory.getLogger(ZkAsyncCallbacks.class);
   public static final int UNKNOWN_RET_CODE = 255;
+
+  // Whenever there's a mismatch between the expected Zookeeper session ID and the actual
+  // Zookeeper session ID for the async operation, the ZkClient which performs the async
+  // operation passes this return code to the async callback indicate that it has caught
+  // a ZkSessionMismatchedException.
   public static final int ZK_SESSION_MISMATCHED_CODE = 127;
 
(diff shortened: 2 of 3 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -289,7 +290,11 @@ protected boolean needRetry(int rc) {
           return false;
         }
       } catch (ClassCastException | NullPointerException ex) {
-        LOG.error("Failed to handle unknown return code {}. Skip retrying.", rc, ex);
+        if(rc == ZK_SESSION_MISMATCHED_CODE) {
```
narendly, line 293: Nit: format

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -36,6 +36,7 @@
 public class ZkAsyncCallbacks {
   private static Logger LOG = LoggerFactory.getLogger(ZkAsyncCallbacks.class);
   public static final int UNKNOWN_RET_CODE = 255;
+  public static final int ZK_SESSION_MISMATCHED_CODE = 127;
```
narendly, line 39: Is 127 an arbitrary number?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
rabashizade: Thanks, I'll fix it in the next commit.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
rabashizade: Yes, sort of. I noticed that KeeperException.Code int values are all negative, so I chose a positive number (inspired by the choice of 255 for UNKNOWN_RET_CODE) to avoid any unwanted collision with those. I don't what the best practice is for choosing return codes, please let me know if I should change it to a more meaningful value.

narendly: `ZkAsyncCallbacks` is a `zookeeper-api` module's code, so essentially it doesn't really matter and we don't have a set convention for what value we need to use.

I'd suggest that we add a block of comment here describing the situation where we should throw this exception code? For whoever gets to debug this using the code or for future developers so they know whether using this code would be appropriate.

narendly: Also, it might be a good idea to check https://zookeeper.apache.org/doc/r3.6.2/apidocs/zookeeper-server/constant-values.html to avoid collision.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -289,7 +290,11 @@ protected boolean needRetry(int rc) {
           return false;
         }
       } catch (ClassCastException | NullPointerException ex) {
-        LOG.error("Failed to handle unknown return code {}. Skip retrying.", rc, ex);
+        if(rc == ZK_SESSION_MISMATCHED_CODE) {
+          LOG.error("Actual session ID doesn't match with expected session ID. Skip retrying.");
+        } else {
+          LOG.error("Failed to handle unknown return code {}. Skip retrying.", rc, ex);
+        }
```
narendly, line 297: This code seems just one-off and doesn't look to extendable. Do you think this would be a good opportunity to move all Helix Project's `zookeeper-api` error codes into one class so it's easier to track?

For example, see the example in `KeeperException` (this is part of the ZooKeeper project):

```
  @Public
  public static enum Code implements KeeperException.CodeDeprecated {
    OK(0),
    SYSTEMERROR(-1),
    RUNTIMEINCONSISTENCY(-2),
    DATAINCONSISTENCY(-3),
    CONNECTIONLOSS(-4),
    MARSHALLINGERROR(-5),
    UNIMPLEMENTED(-6),
    OPERATIONTIMEOUT(-7),
    BADARGUMENTS(-8),
    APIERROR(-100),
    NONODE(-101),
    NOAUTH(-102),
    BADVERSION(-103),
    NOCHILDRENFOREPHEMERALS(-108),
    NODEEXISTS(-110),
    NOTEMPTY(-111),
    SESSIONEXPIRED(-112),
    INVALIDCALLBACK(-113),
    INVALIDACL(-114),
    AUTHFAILED(-115),
    SESSIONMOVED(-118),
    NOTREADONLY(-119);
```

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
rabashizade: Thanks for the resources. I'll make sure there's no collision with these constant values, and will also add appropriate comments to the code.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
rabashizade: This makes sense, I'll move all of the error codes into one class.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
jiajunwang, line 293: This should not fall into Exception handling logic, it is the expected RC number.
So if we are moving all the return code into a new class, then the switch case should check using that new class, and this is no longer a special case.
If, otherwise, we are not using a new class then, we can check this condition before the switch block.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1962,8 +1962,10 @@ protected void doRetry() {
       });
     } catch (RuntimeException e) {
```
jiajunwang, line 1963: nit, I slightly prefer the following alternative.

} catch (ZkSessionMismatchedException e) {
  ...
} catch (RuntimeException e) {
  ...
}

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
rabashizade: Based on the different exception classes that extend `ZkException` in the package `org.apache.helix.zookeeper.zkclient.exception`, only `ZkSessionMismatchedException` and `ZkInterruptedException` don't already have corresponding codes in `KeeperException.Code`, and `ZkInterruptedException` seems to be only adding thread interruption capability to `ZkException`. So it seems to me that `ZkSessionMismatchedException` is the only exception type in Helix project without a corresponding `KeeperException.Code`. Do we expect more error codes to be added in the future? Because otherwise, it seems to me that adding a new class for this one error code is a bit of an overkill. But if you think this is the right approach, I will add the class.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
rabashizade: Thanks for the comment, I moved it before the switch block. For the moment I haven't defined a new class, but if I do so I will use that class in the switch block.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
rabashizade: Thanks, it is easier to read the way you suggested and I fixed it.

jiajunwang (approved):
LGTM!

rabashizade:

**Commit message:**

Clarify error for ZkSessionMismatchedException in ZkAsyncCallbacks

Change the error message for ZkSessionMismatchedException in ZkAsyncCallbacks to be clearer instead of throwing a NullPointerException.

PR #1719, opened 2021-04-28
[Replica Level Throttle] Apply per replica level throttling logic to Recovery/Load Rebalance and PendingMessage

dasahcc:
### Issues

resolves #343 

### Description

This commit contains change of:
1. Recovery/Load Rebalance and pending messages charge adapting per replica throttling.
2. Change partition level comparator based on 1) missed top states, 2) missed active replicas and 3) ideal state matching or not.

### Tests

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
@@ -365,9 +365,7 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
               resourceName, partitionsWithErrorStateReplica));
     }
 
-    chargePendingTransition(resource, currentStateOutput, throttleController,
-        partitionsNeedRecovery, partitionsNeedLoadBalance, cache,
-        bestPossiblePartitionStateMap, intermediatePartitionStateMap);
+    chargePendingTransition(resource, currentStateOutput, throttleController, cache, preferenceLists, stateModelDef, intermediatePartitionStateMap);
 
     // Perform recovery balance
     Set<Partition> recoveryThrottledPartitions =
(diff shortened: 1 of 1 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -423,211 +421,112 @@ private PartitionStateMap computeIntermediatePartitionState(ResourceControllerDa
...
+      StateTransitionThrottleController throttleController, ResourceControllerDataProvider cache,
+      Map<String, List<String>> preferenceLists, StateModelDefinition stateModelDefinition,
       PartitionStateMap intermediatePartitionStateMap) {
     String resourceName = resource.getResourceName();
-
     // check and charge pending transitions
     for (Partition partition : resource.getPartitions()) {
+      // Custom mode does not apply recovery/load reblance since they are independent on each partition's replica number.
+      Map<String, Integer> requiredStates =
+          getRequiredStates(resourceName, cache, preferenceLists.get(partition.getPartitionName()));
```
narendly, line 459: Nit: typo - "reblance"

Also, this comment is a little confusing - where in the code do we check if the current rebalancer is a custom rebalancer here? And what do you mean by they are independent on each partition's replica number?

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -423,211 +421,112 @@ private PartitionStateMap computeIntermediatePartitionState(ResourceControllerDa
...
-   * @param partitionsNeedRecovery
-   * @param currentStateOutput
-   * @param topState
-   * @param cache
-   * @return a set of partitions that need recovery but did not get recovered due to throttling
+   * Thin wrapper for per message throttling with recovery rebalance type. Also populate
+   * intermediatePartitionStateMap with generated messages from {@link MessageGenerationPhase}.
+   * @param resource                      the resource to throttle
+   * @param throttleController            throttle controller object
+   * @param messageToThrottle             the message to be throttle
```
narendly, line 489: "throttled"?

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -423,211 +421,112 @@ private PartitionStateMap computeIntermediatePartitionState(ResourceControllerDa
...
+   * @param intermediatePartitionStateMap output result for this stage that intemediate state map
+   * @param cache                         cache object for computational metadata from external storage
+   * @param messagesThrottled             messages are already throttled
```
narendly, line 492: "messages that have already been throttled"

narendly:
A general question - why is it necessary to remove throttled messages one by one from `resourceStateMap`? I think if we could explain that a little further in the code as a comment, that would be helpful.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
dasahcc: Yes. Let me fix this.

dasahcc:
> A general question - why is it necessary to remove throttled messages one by one from `resourceStateMap`? I think if we could explain that a little further in the code as a comment, that would be helpful.

Do you mean remove messages from resourceMessageMap? They are the output from message generation and will be dispatched by the following stage. If the message has been throttled, we need to remove it. I will comment it.

dasahcc:
@narendly  Could you please take another round of look?

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -423,211 +421,113 @@ private PartitionStateMap computeIntermediatePartitionState(ResourceControllerDa
...
-      PartitionStateMap bestPossiblePartitionStateMap,
+      StateTransitionThrottleController throttleController, ResourceControllerDataProvider cache,
+      Map<String, List<String>> preferenceLists, StateModelDefinition stateModelDefinition,
       PartitionStateMap intermediatePartitionStateMap) {
     String resourceName = resource.getResourceName();
-
     // check and charge pending transitions
     for (Partition partition : resource.getPartitions()) {
+      // Custom mode does not apply recovery/load rebalance since user can define different number of replicas for
+      // different partitions.
```
narendly, line 458: There's still part of my previous comment that's not addressed: 
To the reader it's a little confusing what Custom Rebalance mode has anything to do with this particular line of code - we are checking and charging pending transitions, but why are we discussing Custom mode here? I don't see here that we are specifically checking whether it's custom mode. Like:

```
if (customMode) {
    return;
}
```

dasahcc: Thanks for clarifying! This comment I put is to explain the assumption of how we derive the states we required. But for real situation, the custom mode even not go so far.

In resource level at top level function in IntermediateStage in #1713 line 300, it will return :

 if (!throttleController.isThrottleEnabled() || !IdealState.RebalanceMode.FULL_AUTO.equals(
        idealState.getRebalanceMode()) || resourceMessageMap.isEmpty()) {
      return bestPossiblePartitionStateMap;
    }

narendly: Could you add this to the comment above? That would reduce confusion.

narendly (approved):

PR #1721, opened 2021-04-30
Remove the logic that disables the partitions which fails to be state-transited from the ERROR state.

### LINKED ISSUE #1720: Avoid the behavior that disables partition automatically when it's ST from ERROR state fails

Currently, Helix automatically disables the partition when the ST from ERROR state fails.
Given the partition will remain in ERROR state, disabling it seems to be unnecessary and cause many admin difficulties.

Just remove this behavior. 

**Additional context**

jiajunwang:
### Issues

close #1720 

### Description

This logic is causing confusion when the partitions are disabled unexpected.
In reality, if the partition fails to finish a state-transition from the ERROR state, it will remain in the ERROR state. And disabling it, in addition, is not desired.

### Tests

There is no additional logic added. I also removed the test case which checks if the partition is disabled on such a failure.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -263,11 +263,6 @@ public class HelixStateTransitionHandler extends MessageHandler {
         _stateModel.rollbackOnError(_message, _notificationContext, error);
         _currentStateDelta.setState(partitionKey, HelixDefinedState.ERROR.toString());
         _stateModel.updateState(HelixDefinedState.ERROR.toString());
-
-        // if we have errors transit from ERROR state, disable the partition
-        if (_message.getFromState().equalsIgnoreCase(HelixDefinedState.ERROR.toString())) {
-          disablePartition();
-        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -263,11 +263,6 @@ void postHandleMessage() {
         _stateModel.rollbackOnError(_message, _notificationContext, error);
         _currentStateDelta.setState(partitionKey, HelixDefinedState.ERROR.toString());
         _stateModel.updateState(HelixDefinedState.ERROR.toString());
-
-        // if we have errors transit from ERROR state, disable the partition
-        if (_message.getFromState().equalsIgnoreCase(HelixDefinedState.ERROR.toString())) {
-          disablePartition();
```
narendly, line 269: This is a general question related to the original intent/design of this feature. My understanding was that we disable partitions/replicas going from `ERROR` to other states because we want some control in how the cluster will be rebalanced/shuffled after that partition exits out of the ERROR state; in other words, we don't want it to "rejoin" the cluster as an enabled partition right away.

This is because for partitions transitioning out of ERROR are usually being monitored by the application admin (they have to explicitly "reset" (or fix) the partition, and usually they want to wait until the application has properly finished resetting on their end. So to take extra caution, we transition it out of ERROR in a disabled state. Once the application admin deems it okay for this particular partition to be rebalanced into the cluster, they can manually enable it.

What do you think? Isn't this the current usage pattern with Helix applications?

jiajunwang: An ERROR partition that is reset without error won't be disabled. Only when the resetting fails, it is disabled.

And I also think it is designed for extra cautious. However, if you think about this story, the only reason a partition becomes OFFLINE and disabled is that the admin tries to reset multiple times, and a latter attempt is successfully done. Meaning the Helix lib behaves differently according to how many times the reset is done to a partition. This is very confusing.

Assuming Helix users are reasonable and they do the necessary checks in the ERROR -> OFFLINE ST to confirm the partition data is in good shape, then disabling the partition does not really help anyone.
On the other hand, if we think ERROR -> OFFLINE ST is not safe and the application logic is not complete, why shall we trust the human validate logic?

Overall, I think the validation should be the responsibility of ERROR -> OFFLINE ST. And if the check fails, the partition will remain in an ERROR state. So disabling it or not does not make difference.

narendly (approved):

jiajunwang:
Approved by @narendly , ready to be merged.

PR #1723, opened 2021-05-04
Remove task requested state

xyuanlu:
### Issues

#1675 Remove requested state update in task framework.

### Description

We can prevent over write by checking the in memory state _stateModel and update ZK conditionally in message handling. Also the in memory state _stateModel update (in both task Runner thread and state transition handling thread), comparing and ZK update need to be protected by a lock.

### Tests

NA

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -176,9 +176,9 @@ public class HelixStateTransitionHandler extends MessageHandler {
         deltaList.add(delta);
         _currentStateDelta.setDeltaList(deltaList);
         _stateModelFactory.removeStateModel(_message.getResourceName(), partitionKey);
-      }
-      // if the partition is not to be dropped, update _stateModel to the TO_STATE
-      if (_stateModel.getCurrentState().equals(_message.getFromState())) {
+      } else if (_stateModel.getCurrentState().equals(_message.getFromState())) {
+        // if the partition is not to be dropped, update _stateModel to the TO_STATE
+        // need this check because TaskRunner may change _stateModel before reach here.
         _stateModel.updateState(toState);
       }
     } else if (!taskResult.isCancelled()) {
@@ -225,6 +225,7 @@ public class HelixStateTransitionHandler extends MessageHandler {
       if (!_stateModel.getCurrentState().equals(_currentStateDelta.getState(partitionKey))
           && !_currentStateDelta.getState(partitionKey)
           .equalsIgnoreCase(HelixDefinedState.DROPPED.toString())) {
+        logger.warn("_stateModel is already updated by TaskRunner. Skip ZK update in StateTransitionHandler");
         return;
       }
       int bucketSize = _message.getBucketSize();
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
@@ -26,7 +26,6 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
-import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixDefinedState;
 import org.apache.helix.HelixException;
@@ -121,9 +120,6 @@ public class HelixStateTransitionHandler extends MessageHandler {
         + _message.getPartitionNames() + " from:" + _message.getFromState() + " to:"
         + _message.getToState() + ", relayedFrom: " + _message.getRelaySrcHost());
 
-    HelixDataAccessor accessor = _manager.getHelixDataAccessor();
-    String partitionName = _message.getPartitionName();
-
     // Set start time right before invoke client logic
     _currentStateDelta.setStartTime(_message.getPartitionName(), System.currentTimeMillis());
 
@@ -142,25 +138,15 @@ public class HelixStateTransitionHandler extends MessageHandler {
     HelixTaskResult taskResult =
         (HelixTaskResult) _notificationContext.get(MapKey.HELIX_TASK_RESULT.toString());
     Exception exception = taskResult.getException();
-
     String partitionKey = _message.getPartitionName();
-    String resource = _message.getResourceName();
-    String sessionId = _message.getTgtSessionId();
-    String instanceName = _manager.getInstanceName();
-
-    HelixDataAccessor accessor = _manager.getHelixDataAccessor();
-    Builder keyBuilder = accessor.keyBuilder();
-
-    int bucketSize = _message.getBucketSize();
-    ZNRecordBucketizer bucketizer = new ZNRecordBucketizer(bucketSize);
 
     // No need to sync on manager, we are cancel executor in expiry session before start executor in
     // new session
     // sessionId might change when we update the state model state.
     // for zk current state it is OK as we have the per-session current state node
     if (!_message.getTgtSessionId().equals(_manager.getSessionId())) {
-      logger.warn("Session id has changed. Skip postExecutionMessage. Old session "
-          + _message.getExecutionSessionId() + " , new session : " + _manager.getSessionId());
+      logger.warn("Session id has changed. Skip postExecutionMessage. Old session " + _message
+          .getExecutionSessionId() + " , new session : " + _manager.getSessionId());
       return;
     }
 
@@ -170,17 +156,14 @@ public class HelixStateTransitionHandler extends MessageHandler {
     _currentStateDelta.setPreviousState(partitionKey, _message.getFromState());
 
     // add host name this state transition is triggered by.
-    if (Message.MessageType.RELAYED_MESSAGE.name().equals(_message.getMsgSubType())) {
-      _currentStateDelta.setTriggerHost(partitionKey, _message.getRelaySrcHost());
-    } else {
-      _currentStateDelta.setTriggerHost(partitionKey, _message.getMsgSrc());
-    }
+    _currentStateDelta.setTriggerHost(partitionKey,
+        Message.MessageType.RELAYED_MESSAGE.name().equals(_message.getMsgSubType()) ?
+            _message.getRelaySrcHost() : _message.getMsgSrc());
 
     if (taskResult.isSuccess()) {
       // String fromState = message.getFromState();
       String toState = _message.getToState();
       _currentStateDelta.setState(partitionKey, toState);
-
       if (toState.equalsIgnoreCase(HelixDefinedState.DROPPED.toString())) {
         // for "OnOfflineToDROPPED" message, we need to remove the resource key record
         // from the current state of the instance because the resource key is dropped.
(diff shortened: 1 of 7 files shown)


helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -138,42 +136,6 @@ void preHandleMessage() throws Exception {
       throw err.exception;
     }
 
-    // Reset the REQUESTED_STATE property if it exists.
```
alirezazamani, line 141: Question: Please confirm this scenario: Let's say we have requested state and this code is just rolled out. In this case since requested state is only related to TASK statemodel and we do not carry the current state of the TASK to next session, then requested state will be gone and there won't be roll out issue?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -233,7 +195,9 @@ void postHandleMessage() {
         _stateModelFactory.removeStateModel(resource, partitionKey);
       } else {
         // if the partition is not to be dropped, update _stateModel to the TO_STATE
-        _stateModel.updateState(toState);
```
alirezazamani, line 236: Question: Why removing the requested state needs this check? Don't we already need such a check?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -272,24 +236,31 @@ void postHandleMessage() {
...
-          .currentState(instanceName, sessionId, resource, bucketizer.getBucketName(partitionKey));
-      if (_message.getAttribute(Attributes.PARENT_MSG_ID) == null) {
-        // normal message
-        if (!accessor.updateProperty(key, _currentStateDelta)) {
-          throw new HelixException("Fails to persist current state back to ZK for resource "
-              + resource + " partition: " + _message.getPartitionName());
+        // Update the ZK current state of the node
+        if (!_stateModel.getCurrentState().equals(_currentStateDelta.getState(partitionKey))
+            && !_currentStateDelta.getState(partitionKey)
+            .equalsIgnoreCase(HelixDefinedState.DROPPED.toString())) {
```
alirezazamani, line 242: Question: Why DROPPED case needs to be handled differently?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: Originally `_stateModel` is only updated here at `postHandleMessage`. Now `TaskRunner` also updates `TaskRunner` on a different thread, so we need to check if `_stateModel` is already updated here before writing the new thread.  
Considering the following case. When controller sends a INIT->RUNNING for a certain task. `TaskRunner` is create in a separate thread and finishes executing the task super fast and update `_stateModel` to COMPLETED.  Now `postHandleMessage()` wants to change `_stateModel` to RUNNING, we should reject such change. If current `_stateModel` is different from message's from state, it means `_stateModel` is updated by `TaskRunner`.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: This check is to make sure that when updating `CurrentState` ZNode, `TaskRunner` haven't touch `CurrentState` ZNode yet. 
In line 185 in the changedHelixStateTransitionHandler (or line 222 in original version), `_stateModel` is not updated in DROPPED case. So the first condition is always fail for DROPPED case.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
dasahcc: +1. Then what is the behavior for the rest of the process? Let's say we dont update it. Do we need to throw an exception for this message processing or we just drop the message and treated it as nothing happened? This could be a confusing behavior.

helix-core/src/main/java/org/apache/helix/task/TaskRunner.java
```diff
@@ -203,23 +206,29 @@ private void requestStateTransition(TaskPartitionState state) {
...
               .currentState(instance, sessionId, resource)
               : keyBuilder.taskCurrentState(instance, sessionId, resource);
-      CurrentState currStateDelta = new CurrentState(resource);
-      currStateDelta.setRequestedState(partition, state.name());
 
-      return accessor.updateProperty(key, currStateDelta);
+      String prevState = _stateModel.getCurrentState();
+      CurrentState currentStateDelta = new CurrentState(resource);
+      currentStateDelta.setSessionId(sessionId);
+      currentStateDelta.setStateModelDefRef("Task");
```
dasahcc, line 223: NIT: we have this "Task" constant. Let's use that instead of hard code it here.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: I think if this is the case we  just don't update `_stateModel` or `CurrentState`. No exception will be thrown.

helix-core/src/main/java/org/apache/helix/task/TaskRunner.java
xyuanlu: THX. Will update.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
junkaixue: @xyuanlu It will be confusing for the logic:
 
You have a message and finished the process ready to update current state. But you did not update the current state. Then you ignore it without anything.

My questions: 1) How you define this message status, success? or failed?
                         2) With this message processing, the application status could be changed internally. The real current state does not reflect the application real current state. Shall we still let it to be the state?

xyuanlu: 1. The message status will still be success. Task did finished the state transition indicated by the message. For example, when participant receives a message INIT->RUNNING, task did started to run and completed. The code high lighted here only skips ZK update. The message handling part is not skipped.
2. We skip ZK update here to make sure the current state on ZK is the same as real task's current state. When task finished running and ZK current state is COMPLETED (Updated in TaskRunner). It does not make sense to change it to RUNNING (When postMessageHandler try to update ZK current state to the message's to state here).

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -138,42 +136,6 @@ void preHandleMessage() throws Exception {
...
-          new StateTransitionError(ErrorType.FRAMEWORK, ErrorCode.ERROR, e);
-      _stateModel.rollbackOnError(_message, _notificationContext, error);
-      _statusUpdateUtil.logError(
-          _message, HelixStateTransitionHandler.class, e, "Error when removing "
-              + CurrentState.CurrentStateProperty.REQUESTED_STATE.name() + " from current state.",
-          _manager);
-    }
   }
 
   void postHandleMessage() {
```
jiajunwang, line 141: The changes in this PR make sense. However, it introduces even more special cases in addition to the already complicated code. For example,
1. Drop ST is handled differently from the other ST message. If from-state mismatch, we don't touch the state model. But as long as it is a drop message, then we can delete it.
2. If ST fails, no matter if the states match or not, the ERROR state will be updated.

I suggest refactoring postHandleMessage to clean up (or group) the special cases so it remains readable. I wish the logic could be,
1. Basic read.
2. Check corner cases, special handling.
3. Normal case processing.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -227,56 +172,73 @@ void postHandleMessage() {
...
+    HelixDataAccessor accessor = _manager.getHelixDataAccessor();
+    String partitionKey = _message.getPartitionName();
+    String resource = _message.getResourceName();
+    String sessionId = _message.getTgtSessionId();
+    String instanceName = _manager.getInstanceName();
 
     try {
-      // Update the ZK current state of the node
+      // We did not update _stateModel for DROPPED state, so it won't match _stateModel.
+      if (!_stateModel.getCurrentState().equals(_currentStateDelta.getState(partitionKey))
```
jiajunwang, line 225: nit, this happens when race condition happens, right? Shall we log a warning just for debugging?

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
```diff
@@ -227,56 +172,73 @@ void postHandleMessage() {
...
-        List<ZNRecordDelta> deltaList = new ArrayList<ZNRecordDelta>();
+        List<ZNRecordDelta> deltaList = new ArrayList<>();
         deltaList.add(delta);
         _currentStateDelta.setDeltaList(deltaList);
-        _stateModelFactory.removeStateModel(resource, partitionKey);
-      } else {
-        // if the partition is not to be dropped, update _stateModel to the TO_STATE
+        _stateModelFactory.removeStateModel(_message.getResourceName(), partitionKey);
+      }
+      // if the partition is not to be dropped, update _stateModel to the TO_STATE
```
jiajunwang, line 180: The comment seems not to match the code logic here.

jiajunwang (commented):
I see some comments have not been addressed yet. Also, I left 2 minor comments. Please take a look.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: TFTR. Updated.

jiajunwang: DROPPED does not count...

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixStateTransitionHandler.java
xyuanlu: TFTR. Updated.

helix-core/src/main/java/org/apache/helix/task/TaskRunner.java
```diff
@@ -180,15 +182,16 @@ private void signalStarted() {
...
-        setRequestedState(_manager.getHelixDataAccessor(), _instance, _sessionId, _taskName,
-            _taskPartition, state);
-    if (!success) {
-      LOG.error(String
-          .format(
-              "Failed to set the requested state to %s for instance %s, session id %s, task partition %s.",
-              state, _instance, _sessionId, _taskPartition));
+  private void updateCurrentState(TaskPartitionState state) {
+    synchronized (_stateModel) {
+      if (setZKCurrentState(_manager.getHelixDataAccessor(), _instance, _sessionId, _taskName,
```
jiajunwang, line 187: In postHandleMessage, we set statemodel before update current state. Shall we do the same here? Or you have some other thought?

jiajunwang (commented):
Looks good in general. Only one question remaining.

jiajunwang (approved):
LGTM!

alirezazamani (approved):
LGTM

xyuanlu:
This PR is ready to be merged. Approved by @jiajunwang @alirezazamani 

Final merge message:
Remove task requested state update.

PR #1724, opened 2021-05-06
Applying per replica logic for entire stage

junkaixue:
### Issues

resolves #343 

### Description

This commit contains:
1. Per resource looping with dynamic rebalance type computation and partition level ordering.
2. Let entire stage works against per replica throttling logic.

### Tests

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
@@ -73,16 +73,17 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
         event.getAttribute(AttributeName.RESOURCES_TO_REBALANCE.name());
     ResourceControllerDataProvider cache =
         event.getAttribute(AttributeName.ControllerDataProvider.name());
+    MessageOutput messageOutput =event.getAttribute(AttributeName.MESSAGES_SELECTED.name());
 
     if (currentStateOutput == null || bestPossibleStateOutput == null || resourceToRebalance == null
-        || cache == null) {
+        || cache == null || messageOutput == null) {
       throw new StageException(String.format("Missing attributes in event: %s. "
-          + "Requires CURRENT_STATE (%s) |BEST_POSSIBLE_STATE (%s) |RESOURCES (%s) |DataCache (%s)",
-          event, currentStateOutput, bestPossibleStateOutput, resourceToRebalance, cache));
+              + "Requires CURRENT_STATE (%s) |BEST_POSSIBLE_STATE (%s) |RESOURCES (%s) |MESSAGE_SELECT (%s) |DataCache (%s)",
+          event, currentStateOutput, bestPossibleStateOutput, resourceToRebalance, messageOutput, cache));
     }
 
     IntermediateStateOutput intermediateStateOutput =
-        compute(event, resourceToRebalance, currentStateOutput, bestPossibleStateOutput);
+        compute(event, resourceToRebalance, currentStateOutput, bestPossibleStateOutput, messageOutput);
     event.addAttribute(AttributeName.INTERMEDIATE_STATE.name(), intermediateStateOutput);
 
     // Make sure no instance has more replicas/partitions assigned than maxPartitionPerInstance. If
@@ -106,7 +107,7 @@ public class IntermediateStateCalcStage extends AbstractBaseStage {
    * @return
    */
   private IntermediateStateOutput compute(ClusterEvent event, Map<String, Resource> resourceMap,
-      CurrentStateOutput currentStateOutput, BestPossibleStateOutput bestPossibleStateOutput) {
+      CurrentStateOutput currentStateOutput, BestPossibleStateOutput bestPossibleStateOutput, MessageOutput messageOutput) {
     IntermediateStateOutput output = new IntermediateStateOutput();
     ResourceControllerDataProvider dataCache =
         event.getAttribute(AttributeName.ControllerDataProvider.name());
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -294,99 +294,43 @@ private PartitionStateMap computeIntermediatePartitionState(ResourceControllerDa
...
+    Set<String> messagesThrottledForLoad = new HashSet<>();
     ClusterConfig clusterConfig = cache.getClusterConfig();
 
     // If the threshold (ErrorOrRecovery) is set, then use it, if not, then check if the old
     // threshold (Error) is set. If the old threshold is set, use it. If not, use the default value
     // for the new one. This is for backward-compatibility
     int threshold = 1; // Default threshold for ErrorOrRecoveryPartitionThresholdForLoadBalance
-    int partitionCount = partitionsWithErrorStateReplica.size();
+    // Keep the error count as partition level. This logic only applies to downward state transition determination
+    int partitionCount = (int) currentStateOutput.getCurrentStateMap(resourceName)
```
narendly, line 326: Nit: consider renaming this to `numPartitionsWithErrorReplica`?

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -398,21 +342,71 @@ private PartitionStateMap computeIntermediatePartitionState(ResourceControllerDa
...
+      List<Message> messagesToThrottle = new ArrayList<>(resourceMessageMap.get(partition));
+      if (messagesToThrottle == null || messagesToThrottle.isEmpty()) {
+        continue;
+      }
+      Collections.sort(messagesToThrottle,
+          new MessagePriorityComparator(preferenceList, stateModelDef.getStatePriorityMap()));
+      for (Message message : messagesToThrottle) {
+        RebalanceType rebalanceType = getRebalanceTypePerMessage(requiredState, message, derivedCurrentStateMap);
+
+        // Number of states required by StateModelDefinition are not satisfied, need recovery
```
narendly, line 367: This comment needs to be right above line 365 where the rebalanceType is determined.

narendly (approved):
Approved with minor comments. Well done!

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
junkaixue: Sounds good.

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -396,23 +340,74 @@ private PartitionStateMap computeIntermediatePartitionState(ResourceControllerDa
...
+        currentStateOutput.getCurrentStateMap(resourceName), stateModelDef.getTopState()));
+    for (Partition partition : partitions) {
+      Map<String, String> derivedCurrentStateMap = currentStateOutput.getCurrentStateMap(resourceName, partition)
+          .entrySet()
+          .stream()
+          .collect(Collectors.toMap(entry -> entry.getKey(), entry -> entry.getValue()));
+      List<String> preferenceList = preferenceLists.get(partition.getPartitionName());
+      Map<String, Integer> requiredState = getRequiredStates(resourceName, cache, preferenceList);
+      List<Message> messagesToThrottle = new ArrayList<>(resourceMessageMap.get(partition));
+      if (messagesToThrottle == null || messagesToThrottle.isEmpty()) {
```
jiajunwang, line 360: nit move to the beginning to avoid unnecessary current state converting?

helix-core/src/main/java/org/apache/helix/controller/stages/IntermediateStateCalcStage.java
```diff
@@ -294,99 +294,43 @@ private PartitionStateMap computeIntermediatePartitionState(ResourceControllerDa
...
+        idealState.getRebalanceMode()) || resourceMessageMap.isEmpty()) {
       return bestPossiblePartitionStateMap;
     }
 
     String stateModelDefName = idealState.getStateModelDefRef();
     StateModelDefinition stateModelDef = cache.getStateModelDef(stateModelDefName);
-    PartitionStateMap intermediatePartitionStateMap = new PartitionStateMap(resourceName);
+    // This require a deep copy of current state map because some of the states will be overwritten by applying
+    // messages to it.
+    PartitionStateMap intermediatePartitionStateMap =
```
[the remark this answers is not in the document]
junkaixue: Let me put a TODO here. If we do see this is a bottleneck, let's optimize it. This may require a complete design since in persistent stage, we need to output this result to ZK. And if you make it async, you cannot guarantee the correctness when you need to output in that stage.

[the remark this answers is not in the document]
jiajunwang: Sure, it can be future improvements. But I would suggest moving the intermediate state map calculation out of the compute() method. So it would be easier for use to make the change. Basically, compute() returns the message list. And we have another method to combine the messages to the current state and generate the intermediate state map in the execute() method.

Regarding my proposal, I searched the related code:
There are 3 stages need the intermediate state map, all of them are async. So I think it would be safe to do so. At least it is the current assumption that the check or persistence is done asynchronously.

I'm not 100% sure about the monitoring part. My quick search shows that all the related monitoring items do not require you to compute the full intermediate state map. The messages would be enough.

Moreover, regarding the 3 stages,
1. PersistAssignmentStage needs the whole map, so I suggest we do the intermediate state calculation there.
2. TargetExteralViewCalcStage also needs the whole map, I'm not sure if we should do the calculation 2 times to avoid latency or adding a fancier design so that only one of them finishes the calculation, and then we are good.
3. MaintenanceRecoveryStage needs the map to check MAX_PARTITION_PER_INSTANCE_EXCEEDED violation, but I'm pretty sure it can also be done by checking the current state + the message list. In which case, you don't need to copy-paste the state map and process it.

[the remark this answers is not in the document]
junkaixue: Moved the derive logic to another function.

PR #1732, opened 2021-05-11
Change Workflow monitor and PerInstanceResource monitor from static m…

### LINKED ISSUE #1683: Convert legacy static monitors to use dynamic metrics
### Describe the bug
There are some legacy code in metrics component that use static monitor to report interested metrics. As we know, those static metrics are hard to maintain and cause a lot of boilerplate code. We should convert them to use our dynamic metric framework so that our code base is clean and concise.

### To Reproduce
Check Helix code base for static monitors. 

### Expected behavior
After conversion, all metrics should work the same as before.

zhangmeng916:
…etrics to dynamic metrics

### Issues

Fixed #1683 
### Description

These PR continues to change some legacy metrics that use static metric to the standard ones that use dynamic metric. The changed monitor includes workflow monitor and perinstanceresource monitor. There are also some corresponding changes made in the test and caller functions.

### Tests

Helix-core:
[INFO] Tests run: 1268, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,062.131 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 1268, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO]
[INFO] Loading execution data file /home/mnzhang/helix/helix-core/target/jacoco.exec
[INFO] Analyzed bundle 'Apache Helix :: Core' with 894 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  01:24 h
[INFO] Finished at: 2021-05-11T11:27:32-07:00

### Documentation (Optional)

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/monitoring/mbeans/WorkflowMonitor.java
```diff
@@ -112,27 +81,28 @@ public void updateWorkflowCounters(TaskState to) {
...
+      incrementSimpleDynamicMetric(_failedWorkflowCount, 1);
     } else if (to.equals(TaskState.COMPLETED)) {
-      _successfulWorkflowCount++;
+      incrementSimpleDynamicMetric(_successfulWorkflowCount, 1);
 
       // Only record latency larger than 0 and succeeded workflows
-      _maximumWorkflowLatencyGauge = Math.max(_maximumWorkflowLatencyGauge, latency);
-      _totalWorkflowLatencyCount += latency > 0 ? latency : 0;
+      _maximumWorkflowLatencyGauge
+          .updateValue(Math.max(_maximumWorkflowLatencyGauge.getValue(), latency));
```
dasahcc, line 90: NIT: we can use the same function of incrementSimpleDynamicMetric

dasahcc (approved):
Overall, LGTM

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -710,8 +709,8 @@ private String preProcessWorkflow(String workflowType) {
       if (!_perTypeWorkflowMonitorMap.containsKey(workflowType)) {
         WorkflowMonitor monitor = new WorkflowMonitor(_clusterName, workflowType);
         try {
-          registerWorkflow(monitor);
-        } catch (MalformedObjectNameException e) {
+          monitor.register();
+        } catch (Exception e) {
```
jiajunwang, line 713: Should only throw JMException? And the same question to the other catches.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -898,26 +896,19 @@ private void unregisterPerInstanceResources(
       throws MalformedObjectNameException {
     synchronized (_perInstanceResourceMonitorMap) {
       for (PerInstanceResourceMonitor.BeanName beanName : beanNames) {
-        unregister(getObjectName(
-            getPerInstanceResourceBeanName(beanName.instanceName(), beanName.resourceName())));
+        _perInstanceResourceMonitorMap.get(beanName).unregister();
```
jiajunwang, line 899: Better to check null to avoid NPE.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/PerInstanceResourceMonitor.java
```diff
@@ -113,6 +118,10 @@ public String getResourceName() {
     return _resourceName;
   }
 
+  public String getBeanName() {
```
jiajunwang, line 121: Is this method used anywhere? And the return string seems not to be a valid MBean name, which should look like, "%s,%s=%s,%s=%s"

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/PerInstanceResourceMonitor.java
```diff
@@ -131,13 +140,41 @@ public synchronized void update(Map<Partition, String> stateMap, Set<String> tag
...
+  @Override
+  public DynamicMBeanProvider register() throws JMException {
+    List<DynamicMetric<?, ?>> attributeList = new ArrayList<>();
+    attributeList.add(_partitions);
+    String beanName = getPerInstanceResourceBeanName(_participantName, _resourceName);
+    doRegister(attributeList, MBEAN_DESCRIPTION, getObjectName(beanName));
+    return this;
+  }
+
+  /**
```
jiajunwang, line 161: These are just for generating the Object name, since BeanName won't be used in the ClusterStatusMonitor anymore, can we just make it private and change the logic to return a complete ObjectName?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/WorkflowMonitor.java
```diff
@@ -142,12 +113,72 @@ public void resetGauges() {
    */
   public void updateWorkflowGauges(TaskState current) {
     if (current == null || current.equals(TaskState.NOT_STARTED)) {
-      _queuedWorkflowGauge++;
+      incrementSimpleDynamicMetric(_queuedWorkflowGauge, 1);
```
jiajunwang, line 116: Not for this PR, but this method seems to be a better fit in the SimpleDynamicMetric class

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
zhangmeng916: Unregister already has a null check like this. Should be enough, right?
public static void unregister(ObjectName objectName) {
    if (objectName != null && _beanServer.isRegistered(objectName)) {
      try {
..................

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -874,13 +873,12 @@ private void unregisterAllEventMonitors() {
...
 
   private void registerPerInstanceResources(Collection<PerInstanceResourceMonitor> monitors)
-      throws MalformedObjectNameException {
+      throws JMException {
     synchronized (_perInstanceResourceMonitorMap) {
       for (PerInstanceResourceMonitor monitor : monitors) {
         String instanceName = monitor.getInstanceName();
         String resourceName = monitor.getResourceName();
-        String beanName = getPerInstanceResourceBeanName(instanceName, resourceName);
-        register(monitor, getObjectName(beanName));
```
rabashizade, line 883: Is it okay that this method (`private void register(Object bean, ObjectName name)`) is still used in other places in the code?

zhangmeng916: Yeah, as long as we changed the register of workflow monitor and perinstanceresource monitor to use the dynamic register, we shouldn't be concerned about other places in this monitor. Later we may still need to change them though if we would like to change corresponding monitors.

rabashizade: I see, thanks for clarification, but what about in other classes? My main concern is its usage in `public void active()` method, which in turn is used in `private void enableClusterStatusMonitor(boolean enable)` method in `GenericHelixController`. Although since the tests are passing it should be fine, but I was just curious.

zhangmeng916: I see. So this register is only registering cluster status monitor, which we haven't changed it yet. And inside cluster status monitor, other monitors got registered individually with different methods. Once we start to change cluster status monitor, we may need to handle its register from its calling class.

rabashizade: Got it, thanks.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
jiajunwang: _perInstanceResourceMonitorMap.get(beanName) could be null?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/PerInstanceResourceMonitor.java
jiajunwang: I don't think this comment has been addressed.
Originally, the PerInstanceResourceMonitor does not know cluster info. The BeanName object only contains the instance and resource info. So the registration is done in the ClusterStatusMonitor.
In this new version, since we encapsulate the register method inside PerInstanceResourceMonitor (since now it is a DynamicMBeanProvider). Based on this, I think there are 2 options,

1. Remove BeanName class and always let the PerInstanceResourceMonitor initialized with the fully generated ObjectName (which includes the cluster name). For the existing BeanName usages in the ClusterStatusMonitor, we can change to use the ObjectName. The downside of this option is the PerInstanceResourceMonitor needs to understand MonitorDomainNames.ClusterStatus to generate full ObjectName.
2. Since every PerInstanceResourceMonitor is generated in the ClusterStatusMonitor, we add a private method in the ClusterStatusMonitor to generate full ObjectName of the PerInstanceResourceMonitor when constructing it. And PerInstanceResourceMonitor does not need to know about the domain names etc. For example, we are now doing this design for InstanceMonitor. In this case, BeanName can also be removed since it is duplicated with the ObjectName parameter.

Please consider picking up either one of these 2 designs. The current implementation is a mix-up and the MBean name generating logic has duplicated code. It might introduce bugs if we change in one side (BeanName), but forget to change the other side (getObjectName).

jiajunwang (approved):
LGTM!

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/PerInstanceResourceMonitor.java
zhangmeng916: Discussed offline. We will have everything inside BeanName structure, and deprecate other related get bean name functions to concentrate all logic.

zhangmeng916:
This PR is ready to merge. Approve by @junkaixue and @jiajunwang 
Final commit message:
Change Workflow monitor and PerInstanceResource monitor from static metric to dynamic metric.

PR #1736, opened 2021-05-14
Properly clean up TaskCurrentState sessions

### LINKED ISSUE #1735: TaskCurrentState Session Cleanup
### Describe the bug
Currently, old TaskCurrentState session cleanup is dependent on regular CurrentState sessions: for each session in the CurrentState path, the same session will be removed in the TaskCurrentState path. If there is a TaskCurrentState session that isn't also in the CurrentState path, then the session will not be cleaned up correctly. 

### To Reproduce
Create a mock TaskCurrentState session. It will never be cleaned up. 

### Expected behavior
All old sessions should be cleaned up automatically.

NealSun96:
### Issues

Fixes #1735 

### Description

Currently, old TaskCurrentState session cleanup is dependent on regular CurrentState sessions: for each session in the CurrentState path, the same session will be removed in the TaskCurrentState path. 

This is normally acceptable because both CurrentStates are created by the same session, however, under certain unusual circumstances, this may not behave correctly. For example, after creating a TaskCurrentState, if a user changes Helix versions to a version before TaskCurrentState is introduced, then the clean up logic will only clean up regular CurrentState and leave the TaskCurrentState - this leaves a TaskCurrentState session that doesn't exist on CurrentState side, which will never be cleaned up using the current logic. 

### Tests

TestHandleSession.testRemoveOldSession

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -418,7 +418,16 @@ public class ParticipantManager {
       String path = _keyBuilder.currentStates(_instanceName, session).getPath();
       LOG.info("Removing current states from previous sessions. path: " + path);
       _zkclient.deleteRecursively(path);
-      path = _keyBuilder.taskCurrentStates(_instanceName, session).getPath();
+    }
+
+    // Remove all previous task current state sessions
+    for (String session : _dataAccessor
+        .getChildNames(_keyBuilder.taskCurrentStateSessions(_instanceName))) {
+      if (session.equals(_sessionId)) {
+        continue;
+      }
+
+      String path = _keyBuilder.taskCurrentStates(_instanceName, session).getPath();
       LOG.info("Removing task current states from previous sessions. path: " + path);
       _zkclient.deleteRecursively(path);
     }
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -154,6 +154,7 @@ public class ParticipantManager {
     // should not be created by an expired zk session.
     createLiveInstance();
     carryOverPreviousCurrentState();
+    removePreviousTaskCurrentStates();
 
     /**
      * setup message listener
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -418,7 +418,16 @@ private void carryOverPreviousCurrentState() {
       String path = _keyBuilder.currentStates(_instanceName, session).getPath();
       LOG.info("Removing current states from previous sessions. path: " + path);
       _zkclient.deleteRecursively(path);
-      path = _keyBuilder.taskCurrentStates(_instanceName, session).getPath();
+    }
+
+    // Remove all previous task current state sessions
```
jiajunwang, line 423: I suggest you move the task CS sessions removal logic to a separate method. Just call it carryOverPreviousTaskCurrentState or cleanPreviousCurrentState.

We should avoid huge methods whenever possible.

jiajunwang (commented):
A minor comment. In general looks good to me.

jiajunwang (approved):
LGTM!

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -362,6 +363,7 @@ private void carryOverPreviousCurrentState() {
         }
 
         // If the the current state is related to tasks, there is no need to carry it over to new session.
+        // Note: this check is not necessary due to TaskCurrentStates, but keep it for backwards compatibility
```
alirezazamani, line 366: I am not sure if I understand this comment. We have this check to not carry over task related current states. Why it is not necessary?

NealSun96: Because this logic is working with `_keyBuilder.currentStates`. TaskCurrentStates have their own path now (`keyBuilder.taskCurrentStates`), meaning this check will not be used unless for backward compatibility purposes.

NealSun96:
Final commit message:
## Properly clean up TaskCurrentState sessions ##
This PR fixes TaskCurrentState session clean up to be independent from regular sessions.

alirezazamani (approved):

PR #1740, opened 2021-05-16
Add java api for enable/disable cluster pause mode

### LINKED ISSUE #1739: Java API for enable/disable cluster pause mode
The existing controller pause mode does not apply to participants. We need a new feature (cluster pause mode) to pause both controller and participant. APIs are needed to enable/disable cluster pause mode.

Add APIs to HelixAdmin:
`enableClusterPauseMode()`
`disableClusterPauseMode()`

**Additional context**
There will be more PRs following this issue.

huizhilu:
### Issues

Resolves #1739 

### Description

The existing controller pause mode does not apply to participants. We need a new feature (cluster pause mode) to pause both controller and participant. APIs are needed to enable/disable cluster pause mode.

This PR adds 2 java APIs to HelixAdmin
```
void setClusterManagementMode(ClusterManagementModeRequest request);
```

And a new data structure to represent the cluster management mode and status:
`ClusterManagementMode`

### Tests

`testEnableDisableClusterPauseMode`
`testLiveInstanceStatus`

```
20:24:36,171 [INFO] Tests run: 1269, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,849.579 s - in TestSuite
20:24:36,687 [INFO]
20:24:36,688 [INFO] Results:
20:24:36,688 [INFO]
20:24:36,688 [INFO] Tests run: 1269, Failures: 0, Errors: 0, Skipped: 0
20:24:36,689 [INFO]
20:24:36,693 [INFO]
20:24:36,693 [INFO] --- jacoco-maven-plugin:0.8.6:report (generate-code-coverage-report) @ helix-core ---
20:24:36,725 [INFO] Loading execution data file /home/hulu/Projects/helix/helix-core/target/jacoco.exec
20:24:37,202 [INFO] Analyzed bundle 'Apache Helix :: Core' with 901 classes
20:24:38,665 [INFO] ------------------------------------------------------------------------
20:24:38,665 [INFO] BUILD SUCCESS
20:24:38,665 [INFO] ------------------------------------------------------------------------
20:24:38,667 [INFO] Total time:  01:20 h
20:24:38,667 [INFO] Finished at: 2021-05-18T20:24:38-07:00
20:24:38,667 [INFO] ------------------------------------------------------------------------
```

### Documentation (Optional)

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/model/PauseSignal.java
```diff
@@ -28,7 +28,11 @@
 public class PauseSignal extends HelixProperty {
 
   public enum PauseSignalProperty {
-    REASON
+    REASON,
+    PAUSE_CLUSTER,
```
junkaixue, line 32: Question: I am not quite sure what do you mean the PAUSE_CLUSTER? Can you elaborate more on each of the entry. FROM_HOST is something I am not clear as well.

Also do we need to explicitly put CANCEL_PENDING_ST or not. We can just make cluster status as PAUSING/EXITING something.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -497,6 +499,67 @@ public boolean isInMaintenanceMode(String clusterName) {
         .exists(keyBuilder.maintenance().getPath(), AccessOption.PERSISTENT);
   }
 
+  @Override
+  public void enableClusterPauseMode(String clusterName, boolean cancelPendingST, String reason) {
+    logger.info("Enable cluster pause mode for cluster: {}. Cancel pending ST: {}. Reason: {}.",
+        clusterName, cancelPendingST, reason);
+
+    PauseSignal pauseSignal = new PauseSignal("pause");
```
junkaixue, line 507: NIT: we should not hard code it.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -497,6 +499,67 @@ public boolean isInMaintenanceMode(String clusterName) {
...
+    pauseSignal.setPauseCluster(Boolean.toString(true));
+    pauseSignal.setCancelPendingST(cancelPendingST);
+    pauseSignal.setFromHost(NetworkUtil.getLocalhostName());
+    pauseSignal.setTriggerTime(Instant.now().toString());
```
junkaixue, line 511: Question: why not use System.currenttimemillis

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -367,6 +368,41 @@ void manuallyEnableMaintenanceMode(String clusterName, boolean enabled, String r
...
+   * Gets a cluster out of pause mode. When this method returns, it means
+   * the resume signal has been successfully sent, but it does not mean the cluster has
+   * fully exited pause mode.
+   * <p>
+   * To check whether the cluster has fully exited pause mode, call
+   * {@link #getClusterPauseStatus(String)}.
+   *
+   * @param clusterName Cluster name.
+   */
+  void disableClusterPauseMode(String clusterName);
```
junkaixue, line 404: Why we need two APIs? enableClusterPauseMode can add an argument with enabled to be true or false. It follows our convention.

You can have private function to support enable/disable separately. But they can share the same API.

helix-core/src/main/java/org/apache/helix/model/PauseSignal.java
```diff
@@ -28,7 +28,11 @@
 public class PauseSignal extends HelixProperty {
 
   public enum PauseSignalProperty {
```
junkaixue, line 30: I would suggest not mess up with Cluster Status and Action you would like to perform. Better to have two enums to hold them separately. One for user command one for cluster status.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
huizhilu: It's the same as the other API: `disableCluster()`...
Will put "pause" as a default pause id in `PauseSignal`

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
huizhilu: I understand there are many similar APIs in Helix. However, to me, passing boolean parameters to differentiate the behaviors is really a code smell. Ref: https://softwareengineering.stackexchange.com/questions/208224/is-it-better-to-have-one-method-that-takes-a-bool-as-a-parameter-or-two-methods

I would try to avoid that. So I prefer two methods.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
huizhilu: `Instant.now()` gives a human readable format at UTC zone. The epoch timestamp really does not help being written to znode, because we usually will directly read the znode. So a readable format will really help.
The trade-off is also added to the design doc.

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
```diff
@@ -40,7 +40,20 @@
...
+
+    /** Represents {@link LiveInstanceStatus} */
+    STATUS
+  }
+
+  /**
+   * Represents status of a live instance, eg. PAUSED.
+   */
+  public enum LiveInstanceStatus {
+    NORMAL,
```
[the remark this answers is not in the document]
huizhilu: Yeah that's correct. I've thought about it. It is put there to help in the return of java/rest api. If the cluster has fully exited pause, what will the query result look like? "NORMAL" is helping with it. I just wanted to make it as a constant instead of hardcoding in the API.

helix-core/src/main/java/org/apache/helix/model/PauseSignal.java
huizhilu: We don't want to put cluster status into the PAUSE znode. The PAUSE znode is only created/updated by users. Controller does not write any cluster status to it.
The PAUSE znode will be deleted by controller after cluster fully exits.

helix-core/src/main/java/org/apache/helix/model/PauseSignal.java
huizhilu: These fields are representing the user's action. It'd be a record to help tracking and debugging. 
PAUSE_CLUSTER: whether enable/disable pause mode
FROM_HOST: where the pause signal is sent from, eg. Helix REST or a user's laptop that directly calls java api.
CANCEL_PENDING_ST: whether to cancel pending st. Controller will read this field to determine to send cancel messages.

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
```diff
@@ -40,7 +40,20 @@
...
+    PAUSED,
+    ENTERING_PAUSE,
+    EXITING_PAUSE
```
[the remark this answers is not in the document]
huizhilu: Yes it could be derived. But do we want to hard code those status in the API? I thought about it. I think it's cleaner to define the status here in live instance, which could be used for controller and participant.
When exiting pause, once controller receives the exit notification, the controller status will marked as "EXITING_PAUSE" in LEADER.
So I think this enum will help.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -367,6 +368,41 @@ void manuallyEnableMaintenanceMode(String clusterName, boolean enabled, String r
...
+   */
+  void enableClusterPauseMode(String clusterName, boolean cancelPendingST, String reason);
+
+  /**
+   * Gets cluster pause status {@link ClusterPauseStatus}.
+   *
+   * @param clusterName cluster name.
+   * @return {@link ClusterPauseStatus}
+   */
+  ClusterPauseStatus getClusterPauseStatus(String clusterName);
```
jiajunwang, line 392: I suggest to make some meaningful refactoring instead of introducing more patches.
1. ClusterPauseStatus -> ClusterManagementStatus {NORMAL, CONTROLLER_PAUSED,  CLUSTER_PAUSING, CLUSTER_PAUSED, CLUSTER_UNPAUSING, CONTROLLER_MAINTENANCE}
2. No more enable or disable, just call it set status.
3. Deprecate all the other existing method to pause or maintain the cluster.

helix-core/src/main/java/org/apache/helix/api/exceptions/HelixPauseExitsException.java
```diff
@@ -0,0 +1,28 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import org.apache.helix.HelixException;
+
+public class HelixPauseExitsException extends HelixException {
+  public HelixPauseExitsException(String message) {
```
jiajunwang, line 25: Question, why we need an additional exception type here? Could you please share more thoughts?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -497,6 +499,67 @@ public boolean isInMaintenanceMode(String clusterName) {
...
+    if (reason != null && !reason.isEmpty()) {
+      pauseSignal.setReason(reason);
+    }
+
+    BaseDataAccessor<ZNRecord> baseDataAccessor = new ZkBaseDataAccessor<>(_zkClient);
+    HelixDataAccessor accessor = new ZKHelixDataAccessor(clusterName, baseDataAccessor);
+
+    if (baseDataAccessor.exists(accessor.keyBuilder().pause().getPath(), AccessOption.PERSISTENT)) {
+      throw new HelixPauseExitsException(clusterName + " pause signal already exits");
```
jiajunwang, line 520: 1. The description does not necessarily match the Exception class name. Do you mean "exits" or "exists"?
2. If pause mode has existed, shall we consider the operation is done? But I agree that we should not just return nothing, since it might be paused due to a different reason.

This makes me think about whether or not we shall enter pause mode if the cluster is in controller pause or maintenance mode. My preference is don't mix things up. Reject the call under any confusing situations. So you have to check for the other signals here.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/NetworkUtil.java
```diff
@@ -120,7 +120,7 @@ public static String getLocalhostName() {
         try {
             return InetAddress.getLocalHost().getHostName();
         } catch (final UnknownHostException e) {
-            throw new RuntimeException("unable to retrieve localhost name");
+            return "UNKNOWN";
```
jiajunwang, line 123: Why? If you return "UNKNOWN" when getLocalhostName() gets an error, the application will continue executing and cause bigger trouble.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
huizhilu: I total agree with this direction, which I have thought about, as I designed the REST endpoint `PUT /cluster/status`.
However, I think this is out of scope of this PR as it'd need quite many changes including the code in REST. It'd be cleaner to refactor the APIs in another PR.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
huizhilu: Ah typo... Should be Exists.
Right, if it already exists, because we have different modes for the PAUSE znode, we don't want to overwrite it. So the PauseExistsException tells the znode already exists and rejects the request as a clear way.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/NetworkUtil.java
huizhilu: This static API is not used anywhere else. I don't think it would cause rebalance errors.

helix-core/src/main/java/org/apache/helix/api/exceptions/HelixPauseExitsException.java
huizhilu: To specifically represents the pause signal has already existed. Could be used for REST api when determining the response code. We can reject the request upon existence or still continue updating the pause znode.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/NetworkUtil.java
jiajunwang: Then I think we should not change it.
1. Throw exception is the right way when the tool hits any problem.
2. If you are convinced that no one is using this method, just remove or deprecate it.

This change does not make sense to me.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
jiajunwang: So do you agree that we should fail the operation even if the cluster is only in the Maintenance mode?

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
jiajunwang: Not true. If you make this change, then we are diverged from what we want to implement. It would be impossible to remove this public API once we have it in a short period.

This is a new API, we don't add APIs that known to be refactored soon.

helix-core/src/main/java/org/apache/helix/api/exceptions/HelixPauseExitsException.java
jiajunwang: I suggest calling it HelixStatusConflictException, or other more general names. Since my suggestion is that we check Maintenance mode too. And I think having 2 or more classes for a similar reason makes sense.
You can always carry the current status in the Exception object for the rest to process.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -497,6 +500,77 @@ public boolean isInMaintenanceMode(String clusterName) {
...
+
+    PauseSignal pauseSignal = new PauseSignal();
+    pauseSignal.setPauseCluster(Boolean.toString(true));
+    pauseSignal.setCancelPendingST(cancelPendingST);
+    pauseSignal.setFromHost(hostname);
+    pauseSignal.setTriggerTime(Instant.now().toString());
+    if (reason != null && !reason.isEmpty()) {
+      pauseSignal.setReason(reason);
+    }
+    if (!accessor.createPause(pauseSignal)) {
```
jiajunwang, line 548: There could be a race condition between maintenance node creating and pause node creating. But it cannot be resolved easily since they are different paths. Let's add a TODO here that, once we merge the status signal nodes into one path, then there won't be any more race conditions.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -497,6 +500,77 @@ public boolean isInMaintenanceMode(String clusterName) {
...
+
+    if (baseDataAccessor.exists(accessor.keyBuilder().pause().getPath(), AccessOption.PERSISTENT)) {
+      throw new HelixConflictException(clusterName + " pause signal already exits");
+    }
+    if (baseDataAccessor.exists(accessor.keyBuilder().maintenance().getPath(), AccessOption.PERSISTENT)) {
+      throw new HelixConflictException(clusterName + " maintenance signal already exits");
+    }
+
+    PauseSignal pauseSignal = new PauseSignal();
+    pauseSignal.setPauseCluster(Boolean.toString(true));
```
jiajunwang, line 541: nit, Can we make it in the constructor parameter to avoid possible misusage?

Ultimately, it would be great if we have a super class for all signals. And the 3 modes are child classes.
Then on set, directly create child class objects.
On read, the super class can have a static method to convert into the child class according to the internal mode field.
This would help to avoid possible misusage and duplicate code.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -497,6 +500,77 @@ public boolean isInMaintenanceMode(String clusterName) {
...
+    }
+  }
+
+  private void disableClusterPauseMode(String clusterName) {
+    logger.info("Disable cluster pause mode for cluster: {}", clusterName);
+    HelixDataAccessor accessor =
+        new ZKHelixDataAccessor(clusterName, new ZkBaseDataAccessor<>(_zkClient));
+    PropertyKey pausePropertyKey = accessor.keyBuilder().pause();
+    PauseSignal pauseSignal = accessor.getProperty(pausePropertyKey);
+    if (pauseSignal == null || pauseSignal.getPauseCluster() == null) {
```
jiajunwang, line 559: getPauseCluster == false? or !isPauseCluster() ?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -497,6 +500,77 @@ public boolean isInMaintenanceMode(String clusterName) {
...
+
+    if (!accessor.removeProperty(pausePropertyKey)) {
+      throw new HelixException("Failed to disable cluster pause mode for cluster: " + clusterName);
+    }
+  }
+
+  @Override
+  public ClusterManagementMode getClusterManagementMode(String clusterName) {
+    // TODO: implement logic
+    return null;
```
jiajunwang, line 571: If you are not adding this implementation, please delay the API into the next PR too.

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
```diff
@@ -52,6 +55,8 @@
 
   private static final Logger _logger = LoggerFactory.getLogger(LiveInstance.class.getName());
 
+  public static final String PAUSED_STATUS = "PAUSED";
```
jiajunwang, line 58: enum?

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
```diff
@@ -130,6 +135,22 @@ public void setHelixVersion(String helixVersion) {
...
+  public String getStatus() {
+    return _record.getSimpleField(LiveInstanceProperty.STATUS.name());
+  }
+
+  /**
+   * Sets the status in simple field.
+   *
+   * @param status status value
+   */
+  public void setStatus(String status) {
```
jiajunwang, line 150: Accepts enum?

helix-core/src/main/java/org/apache/helix/model/PauseSignal.java
```diff
@@ -63,4 +72,36 @@ public String getReason() {
   public boolean isValid() {
     return true;
   }
+
+  public void setPauseCluster(String pause) {
```
jiajunwang, line 76: What is this (╯°Д°)╯ ┻━┻

Why not directly passing boolean in? Do you support other pause String? Like setPauseCluster("Foobar") ???

helix-core/src/main/java/org/apache/helix/model/PauseSignal.java
```diff
@@ -63,4 +72,36 @@ public String getReason() {
...
+
+  public void setCancelPendingST(boolean cancel) {
+    _record.setBooleanField(PauseSignalProperty.CANCEL_PENDING_ST.name(), cancel);
+  }
+
+  public boolean getCancelPendingST() {
+    return _record.getBooleanField(PauseSignalProperty.CANCEL_PENDING_ST.name(), false);
+  }
+
+  public void setTriggerTime(String time) {
```
jiajunwang, line 100: I suggest taking a long type as the input. This will help to avoid misusage.

helix-core/src/main/java/org/apache/helix/model/PauseSignal.java
```diff
@@ -63,4 +72,36 @@ public String getReason() {
   public boolean isValid() {
     return true;
   }
+
+  public void setPauseCluster(String pause) {
+    _record.setSimpleField(PauseSignalProperty.CLUSTER_PAUSE.name(), pause);
+  }
+
+  public String getPauseCluster() {
```
jiajunwang, line 80: Same here, return boolean, please.

helix-core/src/main/java/org/apache/helix/model/management/ClusterManagementMode.java
```diff
@@ -0,0 +1,70 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Represents the management mode of the cluster:
+ * 1. what type of mode it targets to be;
+ * 2. what progress status it is now.
+ */
+public class ClusterManagementMode {
```
jiajunwang, line 27: You can just make this class enum and list all the possible mode,
NORMAL()
CLUSTER_PAUSE(Status status) // inprogress or completed
CONTROLLER_PAUSE()
MAINTENANCE()

Then you don't need to face a Maintenance mode object with inprogress status. Which does not make sense.

jiajunwang: As discussed, this is not doable with enum. Let's keep the current design.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -497,6 +500,72 @@ public boolean isInMaintenanceMode(String clusterName) {
...
+      case CLUSTER_PAUSE:
+        enableClusterPauseMode(clusterName, request.isCancelPendingST(), reason);
+        break;
+      case NORMAL:
+        // If from other modes, should check what mode it is in and call the api accordingly.
+        // If we put all mode config in one znode, one generic method is good enough.
+        disableClusterPauseMode(clusterName);
+        break;
+      default:
+        throw new IllegalArgumentException("ClusterManagementMode " + mode + " not supported");
```
jiajunwang, line 520: nit "is not supported"?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -497,6 +500,72 @@ public boolean isInMaintenanceMode(String clusterName) {
...
+    String hostname = NetworkUtil.getLocalhostName();
+    logger.info(
+        "Enable cluster pause mode for cluster: {}. CancelPendingST: {}. Reason: {}. From Host: {}",
+        clusterName, cancelPendingST, reason, hostname);
+
+    BaseDataAccessor<ZNRecord> baseDataAccessor = new ZkBaseDataAccessor<>(_zkClient);
+    HelixDataAccessor accessor = new ZKHelixDataAccessor(clusterName, baseDataAccessor);
+
+    if (baseDataAccessor.exists(accessor.keyBuilder().pause().getPath(), AccessOption.PERSISTENT)) {
+      throw new HelixConflictException(clusterName + " pause signal already exits");
```
jiajunwang, line 534: nit, but please double-check all the spelling. exits -> exists.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -497,6 +500,72 @@ public boolean isInMaintenanceMode(String clusterName) {
...
+    }
+    if (baseDataAccessor.exists(accessor.keyBuilder().maintenance().getPath(), AccessOption.PERSISTENT)) {
+      throw new HelixConflictException(clusterName + " maintenance signal already exits");
```
jiajunwang, line 537: exits -> exists

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
```diff
@@ -40,7 +40,17 @@
...
+
+    /** Represents the status of live instance, eg. PAUSED */
+    STATUS
+  }
+
+  /**
+   * Saved values for the {@link LiveInstanceProperty#STATUS} field
+   */
+  public enum LiveInstanceStatus {
+    PAUSED
```
jiajunwang, line 53: Question, do we need NORMAL here too?

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
```diff
@@ -130,6 +140,22 @@ public void setHelixVersion(String helixVersion) {
     _record.setSimpleField(LiveInstanceProperty.HELIX_VERSION.toString(), helixVersion);
   }
 
+  /**
+   * Gets the live instance's status.
+   */
+  public String getStatus() {
```
jiajunwang, line 146: Return LiveInstanceStatus object?

helix-core/src/main/java/org/apache/helix/model/PauseSignal.java
```diff
@@ -63,4 +74,33 @@ public String getReason() {
...
+  public String getFromHost() {
+    return _record.getSimpleField(PauseSignalProperty.FROM_HOST.name());
+  }
+
+  public void setCancelPendingST(boolean cancel) {
+    _record.setBooleanField(PauseSignalProperty.CANCEL_PENDING_ST.name(), cancel);
+  }
+
+  public boolean getCancelPendingST() {
+    return _record.getBooleanField(PauseSignalProperty.CANCEL_PENDING_ST.name(), false);
```
jiajunwang, line 99: Question, I am thinking that shall we default cancel the pending ST?

jiajunwang (commented):
Looks good in general. Several minor comments.

helix-core/src/main/java/org/apache/helix/HelixAdmin.java
```diff
@@ -37,6 +37,8 @@
 import org.apache.helix.model.MaintenanceSignal;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
+import org.apache.helix.api.status.ClusterManagementMode;
+import org.apache.helix.model.management.ClusterManagementModeRequest;
```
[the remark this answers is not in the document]
junkaixue: +1 Better to keep both of them there. They are all user faced objects.

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
```diff
@@ -130,6 +140,22 @@ public void setHelixVersion(String helixVersion) {
     _record.setSimpleField(LiveInstanceProperty.HELIX_VERSION.toString(), helixVersion);
   }
 
+  /**
+   * Gets the live instance's status.
+   */
+  public LiveInstanceStatus getStatus() {
+    return LiveInstanceStatus.valueOf(_record.getSimpleField(LiveInstanceProperty.STATUS.name()));
```
jiajunwang, line 147: As you said, "For backward compatibility, the status field in live instance will be removed once it goes back to NORMAL.", this method will throw NPE under normal status. I guess your current test does not cover this case.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
jiajunwang: ？？？

helix-core/src/main/java/org/apache/helix/model/LiveInstance.java
```diff
@@ -130,6 +140,22 @@ public void setHelixVersion(String helixVersion) {
     _record.setSimpleField(LiveInstanceProperty.HELIX_VERSION.toString(), helixVersion);
   }
 
+  /**
+   * Gets the live instance's status.
```
jiajunwang, line 144: nit, please comment about the return value, it would be null if there is no status specified.

jiajunwang (approved):

huizhilu:
This PR is ready to be merged, approved @jiajunwang 

Cluster pause mode feature is going to be added. This commit adds java api to set cluster pause mode.

Issue #1746, opened 2021-05-19
New REST api partitionAssignment  -- return potential assignment given cluster change


xyuanlu:
**Describe the solution you'd like**
This API integrates our different rebalance tools and serves to provide a single access point that returns optimal replica assignments of a current cluster or given future potential change. Users won't need to query different rebalancer tools for different resources. It also has a more user friendly input format, it only need changed part in REST payload.

PR #1747, opened 2021-05-19
New REST api partitionAssignment -- return potential assignment given cluster change

xyuanlu:
### Issues

#1746

### Description

This API integrates our different rebalance tools and serves to provide a single access point that returns optimal replica assignments of a current cluster or given future potential change. 

### Tests

TestResourceAssignmentOptimizer.testComputePartitionAssignment
TestResourceAssignmentOptimizer.testComputePartitionAssignmentWaged
TestResourceAssignmentOptimizer.testComputePartitionAssignmentNegativeInput

```
[INFO] Tests run: 177, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  03:34 min
[INFO] Finished at: 2021-05-19T16:11:11-07:00

```

### Documentation (Optional)

### Commits

### Code Quality


helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
```diff
@@ -0,0 +1,347 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.helix.util.HelixUtil.computeIdealMapping;
+import static org.apache.helix.util.HelixUtil.getIdealAssignmentForFullAuto;
+import static org.apache.helix.util.HelixUtil.getTargetAssignmentForWagedFullAuto;
+
+
+@Path("/clusters/{clusterId}/partitionAssignment")
+public class ResourceAssignmentOptimizer extends AbstractHelixResource {
```
zhangmeng916, line 63: For the class naming, I think we can follow the convention for rest class, i.e. "***accessor" so that we can easily identify it is rest.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
```diff
@@ -0,0 +1,347 @@
...
+  private static Logger LOG = LoggerFactory
+      .getLogger(org.apache.helix.rest.server.resources.helix.ClusterAccessor.class.getName());
```
zhangmeng916, line 65: We should log this class's name here.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
```diff
@@ -0,0 +1,347 @@
...
+    InputFields inputMap = new InputFields();
+    customFieldsMap =
+        OBJECT_MAPPER.readValue(content, new TypeReference<HashMap<String, Map<String, Object>>>() {
+        });
+
+    // Content is given as a KV mapping.
+    for (Map.Entry<String, Map<String, Object>> entry : customFieldsMap.entrySet()) {
+      String key = entry.getKey();
+      switch (key) {
+        case "InstanceChange":
```
zhangmeng916, line 136: I'd suggest to put the acceptable commands in an enum so that it is easy to track and validate. You can check the `Command` enum in `AbstractResource`

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
```diff
@@ -0,0 +1,347 @@
...
+          .equals(WagedRebalancer.class.getName())) {
+        wagedResourceIdealState.add(idealState);
+        continue;
+      }
+      // For non Waged resources, we don't compute resources not in white list.
+      if (!inputFields.resourceFilter.isEmpty() && !inputFields.resourceFilter.contains(resource)) {
+        continue;
+      }
+      // Use getIdealAssignmentForFullAuto for FULL_AUTO resource.
+      if (idealState.getRebalanceMode() == IdealState.RebalanceMode.FULL_AUTO) {
```
zhangmeng916, line 242: It's worth to have private functions to implement FULL_AUTO and SEMI_AUTO scenarios. Then this compute function would be much clear and organized.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
```diff
@@ -0,0 +1,347 @@
...
+        String rebalanceStrategy = idealState.getRebalanceStrategy();
+        if (rebalanceStrategy == null || rebalanceStrategy
+            .equalsIgnoreCase(RebalanceStrategy.DEFAULT_REBALANCE_STRATEGY)) {
+          rebalanceStrategy = AutoRebalanceStrategy.class.getName();
+        }
+        Map<String, Map<String, String>> partitionAssignments = new TreeMap<>(
+            getIdealAssignmentForFullAuto(clusterState.clusterConfig, clusterState.instanceConfigs,
+                clusterState.instances, idealState, new ArrayList<>(idealState.getPartitionSet()),
+                rebalanceStrategy));
+        instanceSwapAndFilter(inputFields, partitionAssignments, resource, result);
```
zhangmeng916, line 252: Also this function can be called once instead of three times as it's used by each case.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
```diff
@@ -0,0 +1,347 @@
...
+public class ResourceAssignmentOptimizer extends AbstractHelixResource {
+  private static Logger LOG = LoggerFactory
+      .getLogger(org.apache.helix.rest.server.resources.helix.ClusterAccessor.class.getName());
+
+  private static class InputFields {
+    Set<String> resourceFilter = new HashSet<>();
+    Map<String, String> nodeSwap = new HashMap<>(); // old instance -> new instance.
+    List<String> newInstances = new ArrayList<>();
+    List<String> instancesToRemove = new ArrayList<>();
+    Set<String> instanceFilter = new HashSet<>();
```
zhangmeng916, line 72: Is there a reason that some fields are list, while others are set? Keep them consistent is better.

helix-rest/src/test/java/org/apache/helix/rest/server/TestResourceAssignmentOptimizer.java
```diff
@@ -0,0 +1,201 @@
...
+  public void beforeClass() {
+    helixDataAccessor = new ZKHelixDataAccessor(cluster, _baseAccessor);
+    _gSetupTool.addInstanceToCluster(cluster, instance1);
+    resources = _gSetupTool.getClusterManagementTool().getResourcesInCluster(cluster);
+    liveInstances =  helixDataAccessor.getChildNames(helixDataAccessor.keyBuilder().liveInstances());
+    Assert.assertFalse(resources.isEmpty() || liveInstances.isEmpty());
+
+    toRemoveInstance = liveInstances.get(0);
+    swapOldInstance = liveInstances.get(1);
+
```
zhangmeng916, line 63: NIT: There're some blank lines in some files, and some typos in comments. Maybe you can double check and reformat.

helix-rest/src/test/java/org/apache/helix/rest/server/TestResourceAssignmentOptimizer.java
```diff
@@ -0,0 +1,201 @@
...
+  @Test
+  public void testComputePartitionAssignment() throws IOException {
+    System.out.println("Start test :" + TestHelper.getTestMethodName());
+
+    // set all resource to FULL_AUTO except one
+    for (int i = 0; i < resources.size() - 1; ++i) {
+      String resource = resources.get(i);
+      IdealState idealState =
+          _gSetupTool.getClusterManagementTool().getResourceIdealState(cluster, resource);
+      idealState.setRebalanceMode(IdealState.RebalanceMode.FULL_AUTO);
```
zhangmeng916, line 75: SEMI_AUTO case is not tested. Shall we also test that case?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
jiajunwang, line 136: All the keys, commands should be enum or static Strings whenever possible. Hard code String is not good.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
```diff
@@ -0,0 +1,347 @@
...
+import org.apache.helix.model.ResourceAssignment;
+import org.apache.helix.model.ResourceConfig;
+import org.apache.helix.model.StateModelDefinition;
+import org.apache.helix.rest.common.HttpConstants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.helix.util.HelixUtil.computeIdealMapping;
+import static org.apache.helix.util.HelixUtil.getIdealAssignmentForFullAuto;
+import static org.apache.helix.util.HelixUtil.getTargetAssignmentForWagedFullAuto;
```
jiajunwang, line 59: Please avoid static imports like this. HelixUtil would be a good reference for the code readers to understand where the methods are coming from.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -201,6 +201,10 @@ public ZkClient getZkClient() {
     return (ZkClient) getRealmAwareZkClient();
   }
 
+  public String getZkAddr() {
```
jiajunwang, line 204: I think we don't want to expose the ZkAddr to avoid unnecessary dependencies. In a multi ZK scenario, the address might be a mock address or a fake one. Or we may eventually get rid of this field.

This method is the only safe option,   public RealmAwareZkClient getRealmAwareZkClient(); However, there is no way to get "address" from a RealmAwareZkClient interface.
Although this is a problem that we have noticed for a long time, it seems that we are still facing the same problem.

IMO, the right solution is to keep refactoring the WAGED tool even WAGED rebalancer so it accepts RealmAwareZkClient instead of the raw address. This means we ensure our assumption, the physical ZK address is hidden in the RealmAwareZkClient interface, is guaranteed through all of our code.

However, I'm not sure how much will this change cost. It seems to be a must.

helix-rest/src/test/java/org/apache/helix/rest/server/TestResourceAssignmentOptimizer.java
xyuanlu: By default, all resources are created SEMI_AUTO. We kept one resource SEMI_AUTO and set the rest to FULL_AUTO.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
xyuanlu: TFTR. Updated with newly added API in rebalancer tool.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
xyuanlu: TFTR. Updated.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
xyuanlu: Filter fields will need to return if any given participant/resource is in this allowed list. So `Set` data structure is used here.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
```diff
@@ -278,6 +281,18 @@ public HelixDataAccessor getDataAccessor(String clusterName) {
     return _byteArrayZkBaseDataAccessor;
   }
 
+  public ZkBucketDataAccessor getZkBucketDataAccessor() {
+    if (_zkBucketDataAccessor == null) {
+      synchronized (this) {
+        if (_zkBucketDataAccessor == null) {
+          _zkBucketDataAccessor = new ZkBucketDataAccessor(_zkAddr);
```
jiajunwang, line 288: This still looks not right. Any client that is created without using createRealmAwareZkClient() is not compatible with the multi-ZK configuration, right?

helix-rest/src/test/java/org/apache/helix/rest/server/AbstractTestClass.java
```diff
@@ -316,7 +316,7 @@ private void setupZooKeepers() {
   }
 
   protected void setupHelixResources() {
-    _clusters = createClusters(3);
+    _clusters = createClusters(4);
```
jiajunwang, line 319: Just curious, why we need to modify the cluster count when adding new tests? Can't we create the cluster inside the new test cases? Or are we supposed to modify more than one test file every time when we need to add a test case?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -0,0 +1,354 @@
...
+    }
+  }
+
+  private InputFields readInput(String content)
+      throws InvalidParameterException, JsonProcessingException {
+
+    Map<String, Map<String, Object>> customFieldsMap;
+    InputFields inputMap = new InputFields();
+    customFieldsMap =
+        OBJECT_MAPPER.readValue(content, new TypeReference<HashMap<String, Map<String, Object>>>() {
```
jiajunwang, line 136: If you define a class with the field names like INSTANCE_CHANGE, then ObjectMapper can handle it automatically for you. It would help to simplify code, and more importantly, reduce bug and maintenance costs.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -0,0 +1,354 @@
...
+          break;
+        default:
+          throw new InvalidParameterException(
+              "Unsupported command for partitionAssignment : " + key);
+      }
+    }
+    return inputMap;
+  }
+
+  private ClusterState readClusterStateAndValidateInput(String clusterId, InputFields inputFields)
```
jiajunwang, line 186: Have you explored the possibility of keeping a cluster data cache here? So we avoid read latency and we don't need to manually reconstruct a simplified rebalance logic here.

Let's add a TODO here at least.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
junkaixue: +1. I was trying to make this comment, for realm aware client, get a dedicated zk address client is not a good idea.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -0,0 +1,354 @@
...
+  private static Logger LOG = LoggerFactory.getLogger(
+      org.apache.helix.rest.server.resources.helix.ResourceAssignmentOptimizerAccessor.class
+          .getName());
+
+  private static class InputFields {
+    List<String> newInstances = new ArrayList<>();
+    List<String> instancesToRemove = new ArrayList<>();
+    Map<String, String> nodeSwap = new HashMap<>(); // old instance -> new instance.
+    Set<String> instanceFilter = new HashSet<>();
+    Set<String> resourceFilter = new HashSet<>();
```
junkaixue, line 71: Are these cached? If yes, better cached with some objected instead of static fields. Because the data cached is not synchronized between requests. Shared simple data structure could be problematic.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -0,0 +1,354 @@
...
+        });
+
+    // Content is given as a KV mapping.
+    for (Map.Entry<String, Map<String, Object>> entry : customFieldsMap.entrySet()) {
+      String key = entry.getKey();
+      switch (key) {
+        case INSTANCE_CHANGE:
+          for (Map.Entry<String, Object> instanceChange : entry.getValue().entrySet()) {
+            String instanceChangeKey = instanceChange.getKey();
+            if (instanceChangeKey.equals(INSTANCE_CHANGE_ADD_INSTANCES) && (instanceChange
```
junkaixue, line 146: Can we have nested switch case for here? That could be easy for us to read.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -0,0 +1,354 @@
...
+    // It will throw exception if there is no instanceConfig for newly added instance.
+    for (String instance : clusterState.instances) {
+      InstanceConfig config = cfgAccessor.getInstanceConfig(clusterId, instance);
+      clusterState.instanceConfigs.add(config);
+    }
+    clusterState.clusterConfig = cfgAccessor.getClusterConfig(clusterId);
+    return clusterState;
+  }
+
+  private AssignmentResult computeOptimalAssignmentForResources(InputFields inputFields,
```
junkaixue, line 227: Shall we move the logic into HelixUtil or RebalanceUtil? Because this could be something leveraged by Java API as well.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
xyuanlu: TFTR. Actually they are not. I will add a TODO for future.

helix-rest/src/main/java/org/apache/helix/rest/server/ServerContext.java
xyuanlu: ZkBucketDataAccessor will handle the multi-ZK configuration. This is a workaround since we can not create a zkClient from a zkClient.

helix-rest/src/test/java/org/apache/helix/rest/server/AbstractTestClass.java
xyuanlu: I agree that the optimal way is for each test to create cluster inside the test itself. However, this need a larger code refactor for all our helix-rest integration test.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
xyuanlu: HelixUtil already have multiple different APIs serving this feature. Also `InputFields` and `ClusterState` are private class and only have meaning in this context.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
xyuanlu: Thanks for suggestion. I have tried and actually I am not sure if adding a separate function for several lines with many parameters would be more organized...

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizer.java
xyuanlu: `instanceSwapAndFilter` is called for each resources. For non Waged resources, we compute assignment one by one and filter result one by one. Waged resources need to be computed in a batch and filter result one by one. 
I am not sure if these three can be merged.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
xyuanlu: TFTR. Updated.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
jiajunwang: @junkaixue This logic is needed only when we don't have the cluster data cache maintained for helix-rest. I believe adding cluster data cache in helix-rest is our next step. And after that, we can just run a rebalance pipeline instead of duplicating these logics here.
As I discussed with Xiaoyuan, the next step will take time. So we proposed to have this private method temporarily. But we should keep it private since it will be removed eventually.

jiajunwang (approved):
LGTM! Please address the remaining comments.

xyuanlu:
Thanks all for reviewing this PR! Rebased to TOT master and waiting for CI to complete.

This PR is ready to be merged.  Approved by @jiajunwang 
Final commit message:
New REST api partitionAssignment -- return potential assignment given cluster change

PR #1751, opened 2021-05-21
Remove a potential deadlock when shutting down a RoutingTableProvider.

### LINKED ISSUE #1748: Deadlock in RoutingTableProvider shutdown process
### Describe the bug
In a race condition, the RoutingTableProvider shutdown process will be deadlocked. In detail, considering the following 2 threads,
Thread A: Application shutdown the provider.
A.1. shutting down all the provider threads.
A.2. lock  _lastSeenSessions in updateCurrentStatesListeners method
A.3. manager.removeListener, lock the HelixManager.
A.4. unlock HelixManager.
A.5. unlock _lastSeenSessions
A.6. Continue other shutting down logic

Thread B (Event processing thread): process Current State change event, and trigger onLiveInstanceChange() of the RoutingTableProvider.
B.1. Lock HelixManager for invoking the listner (RoutingTableProvider.onLiveInstanceChange)
B.2. lock  _lastSeenSessions in updateCurrentStatesListeners method
B.3. manager.removeListener, lock the HelixManager.
B.4. unlock HelixManager.
B.5. unlock _lastSeenSessions
B.6. continue other event processing logic.

Note that on A.3. and B.2., Thread A holds _lastSeenSessions lock, and requests locking HelixManager. While Thread B holds HelixManager lock and requests locking _lastSeenSessions. This is a perfect example of deadlock.

### To Reproduce
I find this bug when investigating testCurrentStatePathLeakingByAsycRemoval() test unstable. Usually, this deadlock only happens in a rare race condition. That's also why the test is not failing definitely.
To guarantee reproducing the issue, we need to temporary suspend the RoutingTableProvider.onLiveInstanceChange methods after the HelixManager is locked. Then we call RoutingTableProvider.shutdown(). After the shutdown() thread is blocked, we can resume the onLiveInstanceChange event handling threads. And the test case or the 2 threads will be locked forever.

### Expected behavior
We shall remove the deadlock.

### Additional context
This bug directly causes https://github.com/apache/helix/issues/1284. I will close issue #1284 and use this issue to track the fix.

jiajunwang:
### Issues

Fixes #1748 

### Description

This PR aims to avoid a potential deadlock that happens in a race condition when shutting down RoutingTableProvider.

The race condition happens when the RoutingTableProvider is shut down and the corresponding HelixManager is processing an onLiveInstanceChange event. Both threads require 2 locks, HelixManager lock, and RoutingTableProvider._lastSeenSessions, in different orders.
To resolve this problem, this PR changes 2 logic.
1. Remove the _lastSeenSessions lock usage. So no more deadlock. The logic correctness is now ensured by the AtomicReference.getAndSet method.
2. Remove LiveInstanceChangeListener in the shutdown call before processing the other logic. This is a missing logic and should be done regardless of the deadlock problem. Moreover, this change also avoids a more complicated race condition introduced by the previous change that may lead to CurrentStates callback handler leakage.

Test cases in TestZkCallbackHandlerLeak have been updated to validate the logic changes.

### Tests

TestZkCallbackHandlerLeak

I run the test over the weekend, passed continuously 3533 times already.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
@@ -33,6 +33,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 import javax.management.JMException;
 
 import com.google.common.collect.ImmutableMap;
@@ -325,13 +326,15 @@ public class RoutingTableProvider
     }
     _routerUpdater.shutdown();
 
-
     for (PropertyType propertyType : _monitorMap.keySet()) {
       _monitorMap.get(propertyType).unregister();
     }
 
     if (_helixManager != null) {
       PropertyKey.Builder keyBuilder = _helixManager.getHelixDataAccessor().keyBuilder();
+      // Clean up all the listeners.
+      _helixManager.removeListener(keyBuilder.liveInstances(), this);
+      _helixManager.removeListener(keyBuilder.instanceConfigs(), this);
       for (PropertyType propertyType : _sourceDataTypeMap.keySet()) {
         switch (propertyType) {
         case EXTERNALVIEW:
@@ -350,7 +353,7 @@ public class RoutingTableProvider
         case CURRENTSTATES:
           NotificationContext context = new NotificationContext(_helixManager);
           context.setType(NotificationContext.Type.FINALIZE);
-          updateCurrentStatesListeners(Collections.<LiveInstance> emptyList(), context);
+          updateCurrentStatesListeners(Collections.emptyList(), context);
           break;
         default:
           break;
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
```diff
@@ -751,51 +754,49 @@ private void updateCurrentStatesListeners(List<LiveInstance> liveInstances,
...
 
-    Map<String, LiveInstance> curSessions = new HashMap<>();
-    for (LiveInstance liveInstance : liveInstances) {
-      curSessions.put(liveInstance.getEphemeralOwner(), liveInstance);
-    }
+    Map<String, LiveInstance> curSessions = liveInstances.stream().collect(
+        Collectors.toConcurrentMap(LiveInstance::getEphemeralOwner, liveInstance -> liveInstance));
 
-    // Go though the live instance list and update CurrentState listeners
-    synchronized (_lastSeenSessions) {
```
junkaixue, line 764: Question, if we remove lock, could be there any in memory leakage of watchers?

junkaixue (commented):
Overall, looks good

helix-core/src/main/java/org/apache/helix/spectator/RoutingTableProvider.java
jiajunwang: Here's my proof: Every _lastSeenSessions update is followed by cleanup. If clean-up does not fail, then we won't have leakage. Otherwise, there is no recovery anyway.
One potential exception is that the shutdown happens earlier than the final onLiveInstance event processing. In this case, the final event processing logic actually sees a clean session record (updated by the shutdown process). However, this is not possible. Because in this PR, I remove the onLiveInstance listener before initializing the shutdown process. It is guaranteed that there is no onLiveInstance event being processed after shutdown.

junkaixue (approved):

jiajunwang:
Approved by @junkaixue , I will merge the PR.

PR #1753, opened 2021-05-25
Add concurrent control to DistClusterControllerStateModel._controller access to avoid NPE.

### LINKED ISSUE #1752: NPE in the DistClusterControllerStateModel state transition method
### Describe the bug
One NPE is observed in the DistClusterControllerStateModel when cleanup is triggered in parallel with other state transitions.

29066 [HelixTaskExecutor-message_handle_STATE_TRANSITION] ERROR org.apache.helix.messaging.handling.HelixStateTransitionHandler  - Exception while executing a state transition task TestClusterStatusMonitorLifecycle0_2
java.lang.reflect.InvocationTargetException
	at sun.reflect.GeneratedMethodAccessor11.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.helix.messaging.handling.HelixStateTransitionHandler.invoke(HelixStateTransitionHandler.java:386)
	at org.apache.helix.messaging.handling.HelixStateTransitionHandler.handleMessage(HelixStateTransitionHandler.java:313)
	at org.apache.helix.messaging.handling.HelixTask.call(HelixTask.java:97)
	at org.apache.helix.messaging.handling.HelixTask.call(HelixTask.java:49)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException
	at org.apache.helix.participant.DistClusterControllerStateModel.onBecomeLeaderFromStandby(DistClusterControllerStateModel.java:72)
	... 11 more

I believe this exception makes the test TestClusterStatusMonitorLifecycle unstable.

### To Reproduce
Trigger leader to standby transition and close the controller-participant at the same time.

### Expected behavior
No NPE.

jiajunwang:
### Issues

Resolves #1752

### Description

This PR fixes a potential NPE Exception that may be thrown in the DistClusterControllerStateModel state transition methods.
When this error happens, a follower to leader state transition might be interrupted. The controller instance partition will be set with the ERROR state but the controller instance may have connected to the Zookeeper already. This causes inconsistency and leakage since the controller instance won't be properly cleaned up when it is dropped directly from ERROR state.

### Tests

The original issue is hard to test. Especially with the fix, it is not possible to reproduce the NPE problem anymore.
I suggest we mainly rely on the TestClusterStatusMonitorLifecycle to validate the correctness. It should become very stable after this change.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/participant/DistClusterControllerStateModel.java
+++ b/helix-core/src/main/java/org/apache/helix/participant/DistClusterControllerStateModel.java
@@ -19,6 +19,7 @@ package org.apache.helix.participant;
  * under the License.
  */
 
+import java.util.Optional;
 import java.util.Set;
 
 import com.google.common.collect.Sets;
@@ -32,20 +33,18 @@ import org.apache.helix.participant.statemachine.StateModelInfo;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@StateModelInfo(initialState = "OFFLINE", states = {
-    "LEADER", "STANDBY"
-})
+
+@StateModelInfo(initialState = "OFFLINE", states = {"LEADER", "STANDBY"})
 public class DistClusterControllerStateModel extends AbstractHelixLeaderStandbyStateModel {
   private static Logger logger = LoggerFactory.getLogger(DistClusterControllerStateModel.class);
-  protected HelixManager _controller = null;
+  protected Optional<HelixManager> _controllerOpt = Optional.empty();
   private final Set<Pipeline.Type> _enabledPipelineTypes;
 
   public DistClusterControllerStateModel(String zkAddr) {
     this(zkAddr, Sets.newHashSet(Pipeline.Type.DEFAULT, Pipeline.Type.TASK));
   }
 
-  public DistClusterControllerStateModel(String zkAddr,
-      Set<Pipeline.Type> enabledPipelineTypes) {
+  public DistClusterControllerStateModel(String zkAddr, Set<Pipeline.Type> enabledPipelineTypes) {
     super(zkAddr);
     _enabledPipelineTypes = enabledPipelineTypes;
   }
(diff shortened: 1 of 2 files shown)


helix-core/src/test/java/org/apache/helix/monitoring/TestClusterStatusMonitorLifecycle.java
```diff
@@ -288,16 +288,13 @@ public void testClusterStatusMonitorLifecycle() throws Exception {
...
+            exp1);
 
-    // Note, the _asyncTasksThreadPool shutting down logic in GenericHelixController is best effort
-    // there is not guarantee that all threads in the pool is gone. Mossstly they will, but not always.
-    // see https://github.com/apache/helix/issues/1280
     boolean result = TestHelper.verify(() -> ManagementFactory.getPlatformMBeanServer()
         .queryMBeans(new ObjectName("ClusterStatus:*"), exp2).isEmpty(), TestHelper.WAIT_DURATION);
-    Assert.assertTrue(result,
-        "A small chance this may fail due to _asyncThread pool in controller may not shutdown in time. Please check issue 1280 to verify if this is the case.");
+    Assert.assertTrue(result, "Remaining MBeans: " + ManagementFactory.getPlatformMBeanServer()
```
zhangmeng916, line 297: There are a few other test failures besides this test. Will they be fixed by this PR as well?

helix-core/src/main/java/org/apache/helix/participant/DistClusterControllerStateModel.java
```diff
@@ -85,7 +85,7 @@ public void onBecomeStandbyFromLeader(Message message, NotificationContext conte
 
     logger.info(controllerName + " becoming standby from leader for " + clusterName);
 
-    if (_controller != null) {
+    if (_controllerOpt.isPresent()) {
```
junkaixue, line 88: If this is not in the scope of lock, could we leave leakage there when a race condition happen?

helix-core/src/test/java/org/apache/helix/monitoring/TestClusterStatusMonitorLifecycle.java
jiajunwang: That is not clear to me. I prefer to validate by keeping tracking unstable tests.

Moreover, the reason I remove issue 1280 is that it is an out-of-date comment. There is no clear relationship between that issue and the problem here.

helix-core/src/main/java/org/apache/helix/participant/DistClusterControllerStateModel.java
jiajunwang: 1. State transitions are all happening in one thread. So there is no race condition between the reset here and the creation logic. This PR is to fix the NPE when the reset is triggered by shutdown. So this line won't cause the issue even without the fix.
2. This fix is only to fix NPE, it won't fix all potential leakage. Note that if the partition is in ERROR state, then when we drop the partition, there is no clean up. This is the most serious leakage, actually. If the test is still not stable, I plan to fix this problem in another PR.

junkaixue: I remember Xiaoyuan's change will make the state model can call reset and ST at same time, right?

jiajunwang: That would be fine. Your concern is basically "onBecomeLeaderFromStandby" and "onBecomeStandbyFromLeader" are called at the same time. But even with Xiaoyuan's change, this is not possible, right?
If the reset is triggered by the other logic, like shutdown, then we have synchronized (_controllerOpt) in reset() to ensure the calls are mutually exclusive.

junkaixue (approved):

jiajunwang:
Approved by @junkaixue , I will merge the code soon.

PR #1769, opened 2021-05-28
Add management mode pipeline registry and switch logic

### LINKED ISSUE #1766: Add Management Mode Pipeline Registry
### Description
Management Mode Pipeline Registry will help check the cluster status and determine whether the default pipelines can be run.
Use case is, it will help controller to decide when it can exit the cluster freeze mode.

huizhilu:
### Issues

Resolves #1766 

### Description

Management Mode Pipeline Registry will help check the cluster status and determine whether the default pipelines can be run.
The use case is, it will help controller to decide when it can exit the cluster freeze mode.

The overall logic is, for each onControllerChange event, it will enter the management mode pipeline. The management mode pipeline will check if it is entering or existing cluster freeze/maintenance/controller pause mode. And the pipeline will check the required status. After the cluster exits controller pause or maintenance mode, the management mode pipeline will trigger a resume event to switch to the default pipeline.

Additionally, in the default pipeline, if any abnormal instances are found in the ResourceValidationStage, it will exit the pipeline and force switching to the management mode pipeline.

### Tests

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -168,7 +168,7 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
   private final ClusterEventProcessor _taskEventThread;
 
   // Controller will switch to run management mode pipeline when set to true.
-  private boolean _runManagementModePipeline;
+  private boolean _inManagementMode;
   private final ClusterEventBlockingQueue _managementModeEventQueue;
   private final ClusterEventProcessor _managementModeEventThread;
 
@@ -290,28 +290,26 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
    */
   public GenericHelixController() {
     this(createDefaultRegistry(Pipeline.Type.DEFAULT.name()),
-        createTaskRegistry(Pipeline.Type.TASK.name()),
-        createManagementModeRegistry(Pipeline.Type.MANAGEMENT_MODE.name()));
+        createTaskRegistry(Pipeline.Type.TASK.name()));
   }
 
   public GenericHelixController(String clusterName) {
     this(createDefaultRegistry(Pipeline.Type.DEFAULT.name()),
         createTaskRegistry(Pipeline.Type.TASK.name()),
-        createManagementModeRegistry(Pipeline.Type.MANAGEMENT_MODE.name()),
-        clusterName,
+        createManagementModeRegistry(Pipeline.Type.MANAGEMENT_MODE.name()), clusterName,
         Sets.newHashSet(Pipeline.Type.TASK, Pipeline.Type.DEFAULT));
   }
 
-  public GenericHelixController(String clusterName, Set<Pipeline.Type> enabledPipelines) {
+  public GenericHelixController(String clusterName, Set<Pipeline.Type> enabledPipelins) {
     this(createDefaultRegistry(Pipeline.Type.DEFAULT.name()),
         createTaskRegistry(Pipeline.Type.TASK.name()),
         createManagementModeRegistry(Pipeline.Type.MANAGEMENT_MODE.name()),
         clusterName,
-        enabledPipelines);
+        enabledPipelins);
   }
 
-  public void setRunManagementModePipeline(boolean turnOn) {
-    _runManagementModePipeline = turnOn;
+  public void setInManagementMode(boolean enabled) {
+    _inManagementMode = enabled;
   }
 
   class RebalanceTask extends TimerTask {
@@ -583,6 +581,8 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
       registry
           .register(ClusterEventType.OnDemandRebalance, dataRefresh, autoExitMaintenancePipeline,
               dataPreprocess, externalViewPipeline, rebalancePipeline);
+      registry.register(ClusterEventType.ControllerChange, dataRefresh, autoExitMaintenancePipeline,
+          dataPreprocess, externalViewPipeline, rebalancePipeline);
       // TODO: We now include rebalance pipeline in customized state change for correctness.
       // However, it is not efficient, and we should improve this by splitting the pipeline or
       // controller roles to multiple hosts.
@@ -642,6 +642,8 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
           rebalancePipeline);
       registry.register(ClusterEventType.OnDemandRebalance, dataRefresh, dataPreprocess,
           rebalancePipeline);
+      registry.register(ClusterEventType.ControllerChange, dataRefresh, dataPreprocess,
+          rebalancePipeline);
       return registry;
     }
   }
@@ -659,6 +661,7 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
 
       PipelineRegistry registry = new PipelineRegistry();
       Arrays.asList(
+          ClusterEventType.ControllerChange,
           ClusterEventType.LiveInstanceChange,
           ClusterEventType.MessageChange,
           ClusterEventType.OnDemandRebalance,
@@ -670,10 +673,9 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
   }
 
   // TODO: refactor the constructor as providing both registry but only enabling one looks confusing
-  public GenericHelixController(PipelineRegistry registry, PipelineRegistry taskRegistry,
-      PipelineRegistry managementModeRegistry) {
-    this(registry, taskRegistry, managementModeRegistry, null, Sets.newHashSet(
-        Pipeline.Type.TASK, Pipeline.Type.DEFAULT, Pipeline.Type.MANAGEMENT_MODE));
+  public GenericHelixController(PipelineRegistry registry, PipelineRegistry taskRegistry) {
+    this(registry, taskRegistry, createManagementModeRegistry(Pipeline.Type.MANAGEMENT_MODE.name()),
+        null, Sets.newHashSet(Pipeline.Type.TASK, Pipeline.Type.DEFAULT));
   }
 
   private GenericHelixController(PipelineRegistry registry, PipelineRegistry taskRegistry,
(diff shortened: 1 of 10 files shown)


helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -628,18 +652,44 @@ private static PipelineRegistry createTaskRegistry(String pipelineName) {
...
+      // cluster data cache refresh
+      Pipeline dataRefresh = new Pipeline(pipelineName);
+      dataRefresh.addStage(new ReadClusterDataStage());
+
+      // cluster management mode process
+      Pipeline managementMode = new Pipeline(pipelineName);
+      managementMode.addStage(new ManagementModeStage());
+
+      PipelineRegistry registry = new PipelineRegistry();
+      Arrays.asList(
```
junkaixue, line 667: where is onController change?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -688,6 +738,21 @@ private GenericHelixController(PipelineRegistry registry, PipelineRegistry taskR
       _taskEventThread = null;
     }
 
+    if (_enabledPipelineTypes.contains(Pipeline.Type.MANAGEMENT_MODE)) {
```
junkaixue, line 741: The code her is exact same logic with previous two, just variable different. Can we have a private function to abstract the common logic and make the code more clear?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -775,12 +840,35 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
...
+    } else {
+      logger.warn(String
+          .format("No %s pipeline to run for event: %s::%s", dataProvider.getPipelineName(),
+              event.getEventType(), event.getEventId()));
+      return;
+    }
+
+    // Default/task pipelines should not run while management mode pipeline is enabled to run
+    if (!Pipeline.Type.MANAGEMENT_MODE.equals(pipelineType)
+        && _managementControllerDataProvider.shouldRunManagementPipeline()) {
```
junkaixue, line 869: This is not a good idea to take the cache for management pipelines. It can cause race conditions.

I would suggest to use the cache from the event since this function is general function for all three types of caches.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1227,6 +1299,19 @@ private void pushToEventQueues(ClusterEventType eventType, NotificationContext c
     for (Map.Entry<String, Object> attr : eventAttributes.entrySet()) {
       event.addAttribute(attr.getKey(), attr.getValue());
     }
+
+    // Management mode event will force management mode pipeline.
+    if (ClusterEventType.ManagementModePipeline.equals(eventType)
+        || _managementControllerDataProvider.shouldRunManagementPipeline()) {
```
junkaixue, line 1305: Same here, better not use the object reference. We have the cache dedicated in event. Let's bound the operation in the function scope.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
huizhilu: This is a very tricky place. The method `shouldRunManagementPipeline()` should act differently for the default and management pipelines (on the contrary)
 
I tried to use the dedicated cache. It's really difficult to switch and update the cache correctly: when management pipeline is running, it has the new data. But for the default pipeline cache, it still has the old data. The default pipeline's old data won't work. It'd be really difficult to update the data or signal the refresh from management pipeline.

 Thus, it'd be better and easier to have a single source of truth, which is the management cache. I'll sync with you offline for more details.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
alirezazamani: +1 Don't we need onController change event?

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceValidationStage.java
```diff
@@ -41,6 +42,14 @@ public void process(ClusterEvent event) throws Exception {
     if (cache == null) {
       throw new StageException("Missing attributes in event:" + event + ". Requires DataCache");
     }
+
+    // Check if cluster is still in management mode. Eg. there exists any frozen live instance.
+    if (!cache.shouldRunManagementPipeline()) {
+      // Trigger an immediate management mode pipeline.
+      RebalanceUtil.scheduleManagementModePipeline(event.getClusterName());
```
alirezazamani, line 49: Question: What would happen if you schedule onDemand pipeline here? Sine it is paused, the pipeline would go to Management?

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
huizhilu: We'll set `_runManagementModePipeline=true` and `OnDemandRebalance` for the onController change event. No need to add an additional event type.

helix-core/src/main/java/org/apache/helix/util/RebalanceUtil.java
```diff
@@ -145,6 +145,26 @@
...
+    GenericHelixController leaderController =
+        GenericHelixController.getLeaderController(clusterName);
+    if (leaderController != null) {
+      LOG.info("Switching management mode pipeline for cluster={}, turnOn={}", clusterName,
+          turnOn);
+      leaderController.setRunManagementModePipeline(turnOn);
+    } else {
+      LOG.error(
+          "Failed to switch management mode pipeline. Controller for cluster {} does not exist",
+          clusterName);
```
zhangmeng916, line 164: Can you double check the java doc and comment and make it more precise? I think you mean switch to management mode.

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceValidationStage.java
```diff
@@ -41,6 +42,15 @@ public void process(ClusterEvent event) throws Exception {
     if (cache == null) {
       throw new StageException("Missing attributes in event:" + event + ". Requires DataCache");
     }
+
+    // Check if cluster is still in management mode. Eg. there exists any frozen live instance.
+    if (!cache.shouldRunManagementPipeline()) {
+      // Trigger an immediate management mode pipeline.
+      RebalanceUtil.setRunManagementModePipeline(event.getClusterName(), true);
+      RebalanceUtil.scheduleOnDemandPipeline(event.getClusterName(), 0L);
+      throw new StageException("Pipeline should not be run because cluster is in management mode");
```
zhangmeng916, line 51: I don't quite get the logic here. Where we always throw an exception?

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
```diff
@@ -0,0 +1,44 @@
...
+ * under the License.
+ */
+
+import org.apache.helix.controller.dataproviders.ManagementControllerDataProvider;
+import org.apache.helix.controller.pipeline.AbstractBaseStage;
+import org.apache.helix.util.RebalanceUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ManagementModeStage extends AbstractBaseStage {
```
zhangmeng916, line 28: Please remember to add a test case for this stage later.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -775,12 +834,34 @@ private void handleEvent(ClusterEvent event, BaseControllerDataProvider dataProv
...
+      pipelineType = Pipeline.Type.MANAGEMENT_MODE;
+    } else {
+      logger.warn(String
+          .format("No %s pipeline to run for event: %s::%s", dataProvider.getPipelineName(),
+              event.getEventType(), event.getEventId()));
+      return;
+    }
+
+    // Default/task pipelines should not run while management mode pipeline is enabled to run
+    if (_runManagementModePipeline && !Pipeline.Type.MANAGEMENT_MODE.equals(pipelineType)) {
```
zhangmeng916, line 862: Do we need to check the opposite? I mean how to ensure management mode pipeline doesn't run during the other two pipelines?

helix-core/src/main/java/org/apache/helix/controller/stages/ResourceValidationStage.java
alirezazamani: Basically, my question was why did you introduce a new event which is similar to ondemand pipeline? Since you updated this PR and you are now using Ondemand event type, this can be resolved.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
junkaixue: Wait. If the cluster in pause mode, it will run the management pipline right? But if there is a controller change, you will miss that event.

alirezazamani:
Overall considering the codes in this PR looks good to me. The question that I have is about the controller leadership change. If in management mode, the new controller gets the leadership, then how management pipeline would run?

jiajunwang (commented):
I see 2 high-level design questions. Let's try to confirm these 2 problems before reviewing the detail code.
1. I remember we agreed that the Task pipeline should not be impacted. Or are we going to pause Task processing too?
2. I don't think we need an additional queue. Since the default pipeline and the management mode pipeline is mutually exclusive, using a single queue is cleaner and will help to simplify the concurrent control implementation. Regarding this point, my suggestion is to read and consume _runManagementModePipeline in the handleEvent() method only.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
huizhilu: Added TODO: refactor to simplify below similar code of the 3 pipelines.
This should be wrapped into an object and use a HashMap to map each pipeline to its objects(thread, queue, etc).

alirezazamani (approved):
LGTM. Thanks for addressing the comments

helix-core/src/main/java/org/apache/helix/controller/pipeline/Pipeline.java
```diff
@@ -35,7 +35,8 @@
 
   public enum Type {
     DEFAULT,
-    TASK
+    TASK,
+    MANAGEMENT_MODE
```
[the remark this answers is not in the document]
huizhilu: Discussed. We do need this MANAGEMENT_MODE type because we use the type name for the pipeline in our code. It makes the code consistent.
Meanwhile, a java doc comment is added for this type, making it clear that this type is meant to be used by Helix internally, not by external users.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -1495,7 +1575,9 @@ private void triggerResumeEvent(NotificationContext changeContext, boolean prevP
     if (!_paused && (prevPaused || (prevInMaintenanceMode && !_inMaintenanceMode))) {
       pushToEventQueues(ClusterEventType.Resume, changeContext, Collections.EMPTY_MAP);
       logger.info("controller is now resumed from paused/maintenance state");
+      return true;
     }
+    return false;
```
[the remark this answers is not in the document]
huizhilu: This method will be deleted and the according logic will be moved to the management pipeline in the next PR. I changed it to return boolean to make it self explained in this PR. Not necessarily keeping it with our management pipeline.

junkaixue (approved):

helix-core/src/main/java/org/apache/helix/controller/stages/ClusterEventType.java
```diff
@@ -37,6 +37,7 @@
   Resume,
   PeriodicalRebalance,
   OnDemandRebalance,
+  ControllerChange,
```
jiajunwang, line 40: I think using OnDemandRebalance is good enough. ControllerChange could be misleading since not all controller change notifications are generating an event. It is actually mainly processed by the onControllerChange(). The event is a derivation from the real processing logic.

jiajunwang: Synced with Huizhi offline. So this type will be a general event type to trigger entering special mode or exit special mode. It is potentially more handy than Resume and has a different meaning of OnDemandRebalance. He convinced me that it would be a good idea to add it.

jiajunwang (approved):
Thanks for the hard work!

huizhilu:
This PR is ready to be merged, approved by @junkaixue @alirezazamani @jiajunwang 
I will merge it later.

Management Mode Pipeline will help check the cluster status and determine whether the default pipelines can be run.
One use case is, it will help controller to decide when it can exit the cluster freeze mode.

This commit adds management mode pipeline and logic to switch from/to the default resource/task pipelines.

PR #1778, opened 2021-06-01
Get rid of non-official ZK error code to avoid NPEs.

jiajunwang:
### Issues

This PR will partially fix unstable tests like #1763 

### Description

This PR removes the non-official ZK error codes that are returned in Helix ZkClient. The original design was made for differentiating the failure cases. However, that design causes unexpected NPE since most of Helix's logic relies on the Zookeeper KeeperException class to identify the error types.

FYI,
```
2021-05-27T13:44:01.7121680Z 1810401 [ExternalViewComputeWorker-TestClusterStatusMonitorLifecycle0_2] ERROR org.apache.helix.controller.pipeline.AbstractAsyncBaseStage  - Failed to process DEFAULT::ExternalViewComputeStage asynchronously
2021-05-27T13:44:01.7127825Z java.lang.NullPointerException
2021-05-27T13:44:01.7130637Z 	at org.apache.helix.manager.zk.ZkBaseDataAccessor.set(ZkBaseDataAccessor.java:912)
2021-05-27T13:44:01.7135009Z 	at org.apache.helix.manager.zk.ZkBaseDataAccessor.setChildren(ZkBaseDataAccessor.java:856)
2021-05-27T13:44:01.7140459Z 	at org.apache.helix.manager.zk.ZKHelixDataAccessor.setChildren(ZKHelixDataAccessor.java:578)
2021-05-27T13:44:01.7146249Z 	at org.apache.helix.controller.stages.ExternalViewComputeStage.execute(ExternalViewComputeStage.java:134)
2021-05-27T13:44:01.7149517Z 	at org.apache.helix.controller.pipeline.AbstractAsyncBaseStage$1.run(AbstractAsyncBaseStage.java:47)
2021-05-27T13:44:01.7152294Z 	at org.apache.helix.controller.GenericHelixController$3.handleEvent(GenericHelixController.java:701)
2021-05-27T13:44:01.7154919Z 	at org.apache.helix.controller.GenericHelixController$3.handleEvent(GenericHelixController.java:697)
2021-05-27T13:44:01.7157110Z 	at org.apache.helix.common.DedupEventProcessor.run(DedupEventProcessor.java:61)
```

One option is to change all the logic that reads KeeperException. It would cause a huge change and potentially diverge Helix from the official Zookeeper release. So it would eventually make the whole ZK error processing logics harder to maintain.

Here, I proposed to convert back to the official ZK error code. Given the current processing logic does not need a different customized code to branch, I don't see any necessity of having these additional error codes.

### Tests

Related tests have been modified.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -1276,7 +1276,7 @@ public class ZkClient implements Watcher {
           });
     } catch (RuntimeException e) {
       // Process callback to release caller from waiting
-      cb.processResult(ZkAsyncCallbacks.UNKNOWN_RET_CODE, path,
+      cb.processResult(KeeperException.Code.APIERROR.intValue(), path,
           new ZkAsyncCallMonitorContext(_monitor, startT, 0, true));
       throw e;
     }
@@ -1960,14 +1960,9 @@ public class ZkClient implements Watcher {
                 });
         return null;
       });
-    } catch (ZkSessionMismatchedException e) {
-      // Process callback to release caller from waiting
-      cb.processResult(ZkAsyncCallbacks.ZK_SESSION_MISMATCHED_CODE, path,
-          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
-      throw e;
     } catch (RuntimeException e) {
       // Process callback to release caller from waiting
-      cb.processResult(ZkAsyncCallbacks.UNKNOWN_RET_CODE, path,
+      cb.processResult(KeeperException.Code.APIERROR.intValue(), path,
           new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
       throw e;
     }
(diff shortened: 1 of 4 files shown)


rabashizade (approved):
Handling the current use cases by swapping the custom error codes with `KeeperException.Code.APIERROR` seems a practical solution. My only concern is that perhaps with some unforeseen errors, we will get an APIERROR which is unrelated to the main cause, but for current usages that doesn't seem to be the case so this looks good to me.

jiajunwang:
> Handling the current use cases by swapping the custom error codes with `KeeperException.Code.APIERROR` seems a practical solution. My only concern is that perhaps with some unforeseen errors, we will get an APIERROR which is unrelated to the main cause, but for current usages that doesn't seem to be the case so this looks good to me.

Thanks for the review. For the uncertain future, my thought is that we should avoid invading into the KeeperException.Code space. Note that in the newer ZK version, they are extending the error codes. If there is a conflict, then we do not have a solution.
The right way, IMO, is add another layer of information as a supplement to the official ZK error code. I can imagine that we extend the KeeperException.Code class and add our field to it so as to indicate our detailed APIERROR. This is safer and can avoid conflict.

jiajunwang:
Approved by @rabashizade , I will merge the PR.

PR #1787, opened 2021-06-08
Support currentState format for partitionAssignment

xyuanlu:
### Issues

#1746

### Description

Rest API `partitionAssignment`returns optimal replica assignments of a current cluster or given future potential change. This change adds an option for user to be able to choose return format. 

### Tests

testComputePartitionAssignment

```
[INFO] Results:
[INFO] 
[INFO] Tests run: 178, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] Analyzed bundle 'Apache Helix :: Restful Interface' with 78 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  03:31 min
[INFO] Finished at: 2021-06-08T16:14:59-07:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
@@ -71,7 +71,7 @@ public class ResourceAssignmentOptimizerAccessor extends AbstractHelixResource {
     Map<String, String> nodeSwap = new HashMap<>(); // old instance -> new instance.
     Set<String> instanceFilter = new HashSet<>();
     Set<String> resourceFilter = new HashSet<>();
-    String returnFormat = IDEAL_STATE_FORMAT;
+    String returnFormat = AssignmentFormat.IdealStateFormat.name();
   }
 
   // TODO: We could add a data cache here to avoid read latency.
@@ -107,8 +107,10 @@ public class ResourceAssignmentOptimizerAccessor extends AbstractHelixResource {
     Map<String, String> swapInstances;
   }
 
-  private static final String IDEAL_STATE_FORMAT = "IdealStateFormat";
-  private static final String CURRENT_STATE_FORMAT = "CurrentStateFormat";
+  private enum AssignmentFormat{
+    IdealStateFormat,
+    CurrentStateFormat
+  }
 
   private static class OptionsMap {
     @JsonProperty("InstanceFilter")
@@ -173,8 +175,8 @@ public class ResourceAssignmentOptimizerAccessor extends AbstractHelixResource {
           .ifPresent(inputFields.resourceFilter::addAll);
       Optional.ofNullable(inputJsonContent.optionsMap.instanceFilter)
           .ifPresent(inputFields.instanceFilter::addAll);
-      inputFields.returnFormat =
-          Optional.ofNullable(inputJsonContent.optionsMap.returnFormat).orElse(IDEAL_STATE_FORMAT);
+      inputFields.returnFormat = Optional.ofNullable(inputJsonContent.optionsMap.returnFormat)
+          .orElse(AssignmentFormat.IdealStateFormat.name());
     }
 
     return inputFields;
@@ -219,8 +221,8 @@ public class ResourceAssignmentOptimizerAccessor extends AbstractHelixResource {
     }
     clusterState.clusterConfig = cfgAccessor.getClusterConfig(clusterId);
 
-    if (!inputFields.returnFormat.equals(CURRENT_STATE_FORMAT) &&
-        !inputFields.returnFormat.equals(IDEAL_STATE_FORMAT)) {
+    if (!inputFields.returnFormat.equals(AssignmentFormat.CurrentStateFormat.name()) &&
+        !inputFields.returnFormat.equals(AssignmentFormat.IdealStateFormat.name())) {
       throw new InvalidParameterException(
           "Invalid input: Options.ReturnFormat [" + inputFields.returnFormat + "] is invalid.");
     }
@@ -294,7 +296,7 @@ public class ResourceAssignmentOptimizerAccessor extends AbstractHelixResource {
           result);
     }
 
-    if (inputFields.returnFormat.equals(CURRENT_STATE_FORMAT)) {
+    if (inputFields.returnFormat.equals(AssignmentFormat.CurrentStateFormat.name())) {
       return changeToCurrentStateFormat(result);
     } else {
       return result;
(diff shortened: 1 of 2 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -104,11 +107,16 @@ public AssignmentResult() {
     Map<String, String> swapInstances;
   }
 
+  private static final String IDEAL_STATE_FORMAT = "IdealStateFormat";
+  private static final String CURRENT_STATE_FORMAT = "CurrentStateFormat";
```
jiajunwang, line 111: Define the options by enum?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -277,7 +296,26 @@ private AssignmentResult computeOptimalAssignmentForResources(InputFields inputF
           result);
     }
 
-    return result;
+    if (inputFields.returnFormat.equals(AssignmentFormat.CurrentStateFormat.name())) {
+      return changeToCurrentStateFormat(result);
+    } else {
+      return result;
```
jiajunwang, line 302: Same reason as the comment above, "else" here is too wide considering if more enum items are going to be added. Switch with the default condition would help to avoid potential bugs in the future.

jiajunwang: And I would suggest letting the changeXXXXXFormat function handle this check. It will just return whatever format requested.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -277,7 +296,26 @@ private AssignmentResult computeOptimalAssignmentForResources(InputFields inputF
...
+    }
+  }
+
+  // IdealState format   : Map of resource -> partition -> instance -> state.  (default)
+  // CurrentState format : Map of instance -> resource -> partition -> state.
+  private AssignmentResult changeToCurrentStateFormat(AssignmentResult idealStateFormatResult) {
```
jiajunwang, line 308: AssignmentResult is now more complicated, it may have different information. It would be best if we can return this information in the response. Considering if users are going to async call and the handling logic does not know what the original requested format is.

But either way, in the server code here, we should keep tracking what format the AssignmentResult object is using. Either adding a field (indicating format) or child classes (cleaner separate the logic but more code) is fine.

xyuanlu: TFTR. 
I am not sure if we need to include the return format in the response since the response itself is self explanatory. It is clear that this is a IdealState format or a CurrentState format. However, if user is doing async call with some filter, we might want to return that information.

jiajunwang: I don't think it is really "self-explanatory". If the controller is managing hosts as the "Helix Resource", then the resource name looks exactly the same as a host name. User will need to compare with the participant name or check with the resource name to ensure. And remember we do not require a resource must be named differently from the participant. So even the logic that I described is a guess : )

It is a very corner case, but it worths the effort to ensure our public API is crystal clear.

xyuanlu: As or offline discussion, it is better to add metadata (response format, user defined filter etc) in response header. Will have another PR for this since it is a separate logic.  I will add a TODO for now.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -208,6 +220,13 @@ private ClusterState readClusterStateAndValidateInput(String clusterId, InputFie
       clusterState.instanceConfigs.add(config);
     }
     clusterState.clusterConfig = cfgAccessor.getClusterConfig(clusterId);
+
+    if (!inputFields.returnFormat.equals(AssignmentFormat.CurrentStateFormat.name()) &&
```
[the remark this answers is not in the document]
xyuanlu: TFTR. Updated. JSON parsing will return IllegalArgumentException for invalid input.

jiajunwang (approved):
LGTM. Please resolve the remaining comment.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -104,11 +107,18 @@ public AssignmentResult() {
     Map<String, String> swapInstances;
   }
 
+  private enum AssignmentFormat{
+    IdealStateFormat,
+    CurrentStateFormat
```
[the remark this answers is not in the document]
xyuanlu: TFTR. Normally it is advised to have enum with this convention. However, when we want to access enum name, isn't this format better than having enum with value? (Or at lease keep all enums in REST in the same format? Like how ClusterAccessor.java and other REST resources do it.)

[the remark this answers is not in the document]
jiajunwang: I vote to ensure all enums in the project keep the same style. I know rest has some legacy namings that not necessarily follow the rule. But we should not add more.

Regarding which name is better, you can add a string field for the Enum object if really needed. But they make not much difference from my perspective. It is up to you.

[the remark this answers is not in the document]
xyuanlu: We had an offline discussion and decide to keep this way for the following reason:
1. Change enum name to uppercase like `IDEAL_STATE_FORMAT` and keep the same format in input JSON. This is not optimal because it will impact user readability in input.
2. Because we are doing auto converting from JSON input, adding a string field for the Enum will require more code change like comparing the string many times and have dedicated sanity check code.

xyuanlu:
This PR is ready to be merged. Approved by @jiajunwang 

Final commit message:
Support currentState format for partitionAssignment.

PR #1798, opened 2021-06-16
Check cluster management mode status

### LINKED ISSUE #1776: Check participant freeze status
### Description
Controller needs to know the participant freeze status so it can send freeze/unfreeze messages for entering/exiting freeze mode. The status check is done in management mode stage. And also update the status to metadata store.

huizhilu:
### Issues

Resolves #1776 

### Description

Controller needs to know the participant freeze status so it can send freeze/unfreeze messages for entering/exiting freeze mode. The status check is done in management mode stage.

This PR adds methods to check cluster management mode status, and update the status and history accordingly. 

### Tests

 `TestManagementModeStage`

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
+++ b/helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
@@ -105,9 +105,7 @@ public class PropertyPathBuilder {
     addEntry(PropertyType.CUSTOMIZEDVIEW, 2, "/{clusterName}/CUSTOMIZEDVIEW/{customizedStateType}");
     addEntry(PropertyType.CUSTOMIZEDVIEW, 3, "/{clusterName}/CUSTOMIZEDVIEW/{customizedStateType}/{resourceName}");
     addEntry(PropertyType.STATUS, 1, "/{clusterName}/STATUS");
-    addEntry(PropertyType.STATUS, 2, "/{clusterName}/STATUS/{scope}");
-    addEntry(PropertyType.STATUS, 3, "/{clusterName}/STATUS/{scope}/{scopeKey}");
-
+    addEntry(PropertyType.STATUS, 2, "/{clusterName}/STATUS/{clusterName}");
 
     addEntry(PropertyType.TARGETEXTERNALVIEW, 1, "/{clusterName}/TARGETEXTERNALVIEW");
     addEntry(PropertyType.TARGETEXTERNALVIEW, 2,
--- a/helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
+++ b/helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
@@ -105,6 +105,9 @@ public class PropertyPathBuilder {
     addEntry(PropertyType.CUSTOMIZEDVIEW, 2, "/{clusterName}/CUSTOMIZEDVIEW/{customizedStateType}");
     addEntry(PropertyType.CUSTOMIZEDVIEW, 3, "/{clusterName}/CUSTOMIZEDVIEW/{customizedStateType}/{resourceName}");
     addEntry(PropertyType.STATUS, 1, "/{clusterName}/STATUS");
+    addEntry(PropertyType.STATUS, 2, "/{clusterName}/STATUS/{scope}");
+    addEntry(PropertyType.STATUS, 3, "/{clusterName}/STATUS/{scope}/{scopeKey}");
+
 
     addEntry(PropertyType.TARGETEXTERNALVIEW, 1, "/{clusterName}/TARGETEXTERNALVIEW");
     addEntry(PropertyType.TARGETEXTERNALVIEW, 2,
--- a/helix-core/src/main/java/org/apache/helix/controller/pipeline/PipelineSwitchException.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/pipeline/PipelineSwitchException.java
@@ -1,0 +1,29 @@
+package org.apache.helix.controller.pipeline;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Used to exit the current controller pipeline and switch to another pipeline.
+ */
+public class PipelineSwitchException extends StageException {
+  public PipelineSwitchException(String message) {
+    super(message);
+  }
+}
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
@@ -21,7 +21,6 @@ package org.apache.helix.controller.stages;
 
 import java.time.Instant;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 
@@ -72,7 +71,7 @@ public class ManagementModeStage extends AbstractBaseStage {
         checkClusterFreezeStatus(cache.getEnabledLiveInstances(), cache.getLiveInstances(),
             cache.getAllInstancesMessages(), cache.getPauseSignal());
 
-    recordManagementModeStatus(managementMode, accessor);
+    recordClusterStatus(managementMode, accessor);
     recordManagementModeHistory(managementMode, cache.getPauseSignal(), manager.getInstanceName(),
         accessor);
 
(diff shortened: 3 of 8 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
```diff
@@ -53,4 +88,87 @@ private void checkInManagementMode(String clusterName, ManagementControllerDataP
...
+
+    // Record a management mode history in controller history
+    String path = accessor.keyBuilder().controllerLeaderHistory().getPath();
+    if (!accessor.getBaseDataAccessor().update(path, oldRecord -> {
+      if (oldRecord == null) {
+        oldRecord = new ZNRecord(PropertyType.HISTORY.toString());
+      }
+      String fromHost = (pauseSignal == null ? null : pauseSignal.getFromHost());
+      String reason = (pauseSignal == null ? null : pauseSignal.getReason());
+      return new ControllerHistory(oldRecord).updateManagementModeHistory(controllerName, mode,
```
jiajunwang, line 168: Question, why not update using the oldRecord object? It is a local var. So just update it won't cause any issue, right?

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
```diff
@@ -53,4 +88,87 @@ private void checkInManagementMode(String clusterName, ManagementControllerDataP
...
+          fromHost, Instant.now().toEpochMilli(), reason);
```
jiajunwang, line 169: nit, Instant.now().toEpochMilli() should be fetched before the update for a more accurate status change time and avoid repeatedly system calls, since the update method might be retried.

helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
```diff
@@ -105,6 +105,9 @@
     addEntry(PropertyType.CUSTOMIZEDVIEW, 2, "/{clusterName}/CUSTOMIZEDVIEW/{customizedStateType}");
     addEntry(PropertyType.CUSTOMIZEDVIEW, 3, "/{clusterName}/CUSTOMIZEDVIEW/{customizedStateType}/{resourceName}");
     addEntry(PropertyType.STATUS, 1, "/{clusterName}/STATUS");
+    addEntry(PropertyType.STATUS, 2, "/{clusterName}/STATUS/{scope}");
+    addEntry(PropertyType.STATUS, 3, "/{clusterName}/STATUS/{scope}/{scopeKey}");
```
jiajunwang, line 109: What are the other possible values regarding scope? I thought it would be a single node for the whole cluster status. But please share the detailed design here.

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
```diff
@@ -53,4 +88,87 @@ private void checkInManagementMode(String clusterName, ManagementControllerDataP
...
+  }
+
+  private boolean hasPendingMessage(Collection<Message> messages, MessageType type) {
+    return messages.stream().anyMatch(message -> type.name().equals(message.getMsgType()));
+  }
+
+  private void recordManagementModeStatus(ClusterManagementMode mode, HelixDataAccessor accessor) {
+    // update cluster status
+    PropertyKey statusPropertyKey = accessor.keyBuilder().clusterStatus();
+    ClusterStatus clusterStatus = accessor.getProperty(statusPropertyKey);
```
junkaixue, line 138: Question, this status is similar to the history and other properties. I remember we read in cache every time and flush by cache to ZK. Can we do the same thing?

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
```diff
@@ -53,4 +88,87 @@ private void checkInManagementMode(String clusterName, ManagementControllerDataP
...
+    if (clusterStatus == null) {
+      clusterStatus = new ClusterStatus();
+    }
+    if (!mode.getMode().equals(clusterStatus.getManagementMode())
+        || !mode.getStatus().equals(clusterStatus.getManagementModeStatus())) {
+      // status is different, need to write to metastore
```
junkaixue, line 144: NIT: "metadata store"

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
```diff
@@ -53,4 +88,87 @@ private void checkInManagementMode(String clusterName, ManagementControllerDataP
...
+  private void recordManagementModeHistory(ClusterManagementMode mode, PauseSignal pauseSignal,
+      String controllerName, HelixDataAccessor accessor) {
+    // Only record completed status
+    if (!ClusterManagementMode.Status.COMPLETED.equals(mode.getStatus())) {
+      return;
+    }
+
+    // Record a management mode history in controller history
+    String path = accessor.keyBuilder().controllerLeaderHistory().getPath();
+    if (!accessor.getBaseDataAccessor().update(path, oldRecord -> {
```
junkaixue, line 162: Question: are we seeing any potential update conflict between different threads that requires our own updater logic? I think only controller and management mode pipeline does it right? And it's sequential.

helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
alirezazamani: I agree. Do we need more than one node?

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
```diff
@@ -53,4 +88,87 @@ private void checkInManagementMode(String clusterName, ManagementControllerDataP
       RebalanceUtil.enableManagementMode(clusterName, false);
     }
   }
+
+  // Checks cluster freeze, controller pause mode and status.
+  private ClusterManagementMode checkClusterFreezeStatus(
```
alirezazamani, line 93: Question: Can we have not enabled a live instance that is paused? Can we pause an instance if it is not enabled? What about if we are in frozen mode and one of the disabled instances becomes enabled?

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
huizhilu: For HISTORY, I don't see it's updated like that - flushing cache to ZK. The history is always updated using a DataAccessor.

helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
jiajunwang: Do you have a concrete example? It seems to be over-engineering.
I think the entities in Helix world include,
Resource - we have current state nodes and external view nodes,
Instance - we have liveInstance nodes
Controller - we have controller nodes

And these STATUS nodes is to provide some level of aggregation. There is no other entity that we can or need to add here.

helix-core/src/main/java/org/apache/helix/controller/pipeline/PipelineSwitchException.java
```diff
@@ -0,0 +1,29 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Used to exit the current controller pipeline and switch to another pipeline.
+ */
+public class PipelineSwitchException extends StageException {
```
jiajunwang, line 25: Switch is quite confusing and not apply to all cases (for example, entering Maintenance mode), how about PipelineEarlyTerminateException?

Moreover, I think this should be a temporary solution. Using Exception for regular pipeline branches (or alternative logic switches) is very expensive. It would be better to let the pipeline support gracefully switch by returning a execute result, or even support DAG in pipeline.
Of course, not in this PS. TODO here would be enough.

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
```diff
@@ -53,4 +87,90 @@ private void checkInManagementMode(String clusterName, ManagementControllerDataP
...
+  }
+
+  private boolean instancesFullyFrozen(Set<String> enabledLiveInstances,
+      Map<String, LiveInstance> liveInstanceMap,
+      Map<String, Collection<Message>> allInstanceMessages) {
+    // 1. All live instances are frozen
+    // 2. No pending participant status change message.
+    return enabledLiveInstances.stream().noneMatch(
+        instance -> !LiveInstanceStatus.PAUSED.equals(liveInstanceMap.get(instance).getStatus())
+            || hasPendingMessage(allInstanceMessages.get(instance),
```
jiajunwang, line 125: Question, do we need the messages check here? Can we just rely on the paused status that is reported by the participant?

The thing is that the participant might get some new messages after it entering frozen mode, but as long as it is in paused status, it won't execute any of them. In this case, it is still frozen.

helix-core/src/main/java/org/apache/helix/controller/pipeline/PipelineSwitchException.java
huizhilu: Agreed, we should have some logic to more peacefully switch, instead of throwing an exception. Will add a TODO.

helix-core/src/main/java/org/apache/helix/PropertyPathBuilder.java
huizhilu: Synced offline with @jiajunwang We decided to use `/{clusterName}/STATUS/{clusterName}`.  Currently there is no need to add the `{scope}` layer.

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
huizhilu: Syned offline. Yes, we need the messages check to make sure participants are fully frozen with no pending state transitions. After it's frozen, if user sends ST messages, the cluster status should not be changed to in progress as it's in frozen status.

alirezazamani (approved):
LGTM

huizhilu:
I will merge it a bit later.

Controller needs to know the participant freeze status so it can send freeze/unfreeze messages for entering/exiting freeze mode. The status check is done in management mode stage.

This commit adds methods to check cluster management mode status, and update the status and history accordingly.

helix-core/src/main/java/org/apache/helix/controller/pipeline/PipelineSwitchException.java
jiajunwang: How about the naming suggestion that I commented on?

jiajunwang (approved):
LGTM!

I don't have a strong preference for the Exception name. But please put your ideas to respond the comment : )

helix-core/src/main/java/org/apache/helix/controller/pipeline/PipelineSwitchException.java
huizhilu: Discussed with Jiajunwang. I thought PipelineEarlyTerminateException is kind of duplicate with StageException , because StageException is also thrown to early terminate the pipeline. So I wanted a more specific PipelineSwitchException for our case here, which is smaller set of stageException.
We'll keep the current name. If there is a need to change in the future, we can easily refactor it without breaking backward compatibility.

PR #1803, opened 2021-06-24
Add message generation logic for management pipeline

### LINKED ISSUE #1774: Create freeze messages in management mode stage
### Description
In cluster freeze mode, controller sends freeze/unfreeze messages to participants. The management mode stage will be responsible for that. Methods to create/send messages are required.

huizhilu:
### Issues

Resolves #1774 

### Description

In cluster freeze mode, controller sends freeze/unfreeze and pending ST cancellation messages to participants.

In this PR, the existing message generation stage is leveraged to create the ST cancellation messages. Best possible out is built by copying the state map from current state for generating cancellation messages. For freeze/unfreeze messages, the logic is added in `ManagementMessageGenerationPhase`. The existing `MessageDispatchStage` is also used for dispatching all the messages.

### Tests

TestManagementMessageGeneration

```
22:24:55,835 [INFO] Tests run: 1271, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4,946.388 s - in TestSuite
22:24:56,378 [INFO]
22:24:56,378 [INFO] Results:
22:24:56,378 [INFO]
22:24:56,378 [INFO] Tests run: 1271, Failures: 0, Errors: 0, Skipped: 0
22:24:56,378 [INFO]
22:24:56,382 [INFO]
22:24:56,382 [INFO] --- jacoco-maven-plugin:0.8.6:report (generate-code-coverage-report) @ helix-core ---
22:24:56,417 [INFO] Loading execution data file /home/hulu/Projects/helix/helix-core/target/jacoco.exec
22:24:56,883 [INFO] Analyzed bundle 'Apache Helix :: Core' with 908 classes
22:24:58,418 [INFO] ------------------------------------------------------------------------
22:24:58,418 [INFO] BUILD SUCCESS
22:24:58,419 [INFO] ------------------------------------------------------------------------
22:24:58,420 [INFO] Total time:  01:22 h
22:24:58,420 [INFO] Finished at: 2021-06-28T22:24:58-07:00
22:24:58,420 [INFO] ------------------------------------------------------------------------
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/api/status/ClusterManagementMode.java
+++ b/helix-core/src/main/java/org/apache/helix/api/status/ClusterManagementMode.java
@@ -67,4 +67,8 @@ public class ClusterManagementMode {
     public Type getMode() {
         return mode;
     }
+
+    public boolean isFullyInNormalMode() {
+        return Type.NORMAL.equals(mode) && Status.COMPLETED.equals(status);
+    }
 }
--- a/helix-core/src/main/java/org/apache/helix/api/status/ClusterManagementMode.java
+++ b/helix-core/src/main/java/org/apache/helix/api/status/ClusterManagementMode.java
@@ -19,6 +19,8 @@ package org.apache.helix.api.status;
  * under the License.
  */
 
+import org.apache.helix.model.LiveInstance;
+
 /**
  * Represents the management mode of the cluster:
  * 1. what type of mode it targets to be;
@@ -70,5 +72,23 @@ public class ClusterManagementMode {
 
     public boolean isFullyInNormalMode() {
         return Type.NORMAL.equals(mode) && Status.COMPLETED.equals(status);
+    }
+
+    /**
+     * Gets the desired live instance status for this management mode.
+     *
+     * @return The desired {@link org.apache.helix.model.LiveInstance.LiveInstanceStatus}.
+     * If participants status is not expected to change for the management mode, null is returned.
+     */
+    public LiveInstance.LiveInstanceStatus getDesiredParticipantStatus() {
+        switch (mode) {
+            case CLUSTER_PAUSE:
+                return LiveInstance.LiveInstanceStatus.PAUSED;
+            case NORMAL:
+                return LiveInstance.LiveInstanceStatus.NORMAL;
+            default:
+                // Other modes don't need to change participant status
+                return null;
+        }
     }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -82,7 +82,10 @@ import org.apache.helix.controller.stages.CustomizedViewAggregationStage;
 import org.apache.helix.controller.stages.ExternalViewComputeStage;
 import org.apache.helix.controller.stages.IntermediateStateCalcStage;
 import org.apache.helix.controller.stages.MaintenanceRecoveryStage;
+import org.apache.helix.controller.stages.ManagementMessageDispatchStage;
+import org.apache.helix.controller.stages.ManagementMessageGenerationPhase;
 import org.apache.helix.controller.stages.ManagementModeStage;
+import org.apache.helix.controller.stages.MessageGenerationPhase;
 import org.apache.helix.controller.stages.MessageSelectionStage;
 import org.apache.helix.controller.stages.MessageThrottleStage;
 import org.apache.helix.controller.stages.PersistAssignmentStage;
@@ -647,9 +650,16 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
       Pipeline dataRefresh = new Pipeline(pipelineName);
       dataRefresh.addStage(new ReadClusterDataStage());
 
+      // data pre-process pipeline
+      Pipeline dataPreprocess = new Pipeline(pipelineName);
+      dataPreprocess.addStage(new ResourceComputationStage());
+      dataPreprocess.addStage(new CurrentStateComputationStage());
+
       // cluster management mode process
       Pipeline managementMode = new Pipeline(pipelineName);
       managementMode.addStage(new ManagementModeStage());
+      managementMode.addStage(new ManagementMessageGenerationPhase());
+      managementMode.addStage(new ManagementMessageDispatchStage());
 
       PipelineRegistry registry = new PipelineRegistry();
       Arrays.asList(
@@ -658,7 +668,7 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
           ClusterEventType.MessageChange,
           ClusterEventType.OnDemandRebalance,
           ClusterEventType.PeriodicalRebalance
-      ).forEach(type -> registry.register(type, dataRefresh, managementMode));
+      ).forEach(type -> registry.register(type, dataRefresh, dataPreprocess, managementMode));
 
       return registry;
     }
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/ManagementMessageDispatchStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/ManagementMessageDispatchStage.java
@@ -1,0 +1,62 @@
+package org.apache.helix.controller.stages;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.List;
+
+import org.apache.helix.HelixManager;
+import org.apache.helix.api.status.ClusterManagementMode;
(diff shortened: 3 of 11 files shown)


helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -647,9 +650,16 @@ private static PipelineRegistry createManagementModeRegistry(String pipelineName
       Pipeline dataRefresh = new Pipeline(pipelineName);
       dataRefresh.addStage(new ReadClusterDataStage());
 
+      // data pre-process pipeline
+      Pipeline dataPreprocess = new Pipeline(pipelineName);
+      dataPreprocess.addStage(new ResourceComputationStage());
```
junkaixue, line 655: NIT: Do we need ResourceValidationStage? It seemed to bundle with ResourceComputationStage.

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementMessageDispatchStage.java
```diff
@@ -0,0 +1,62 @@
...
+public class ManagementMessageDispatchStage extends MessageDispatchStage {
+  private static final Logger LOG = LoggerFactory.getLogger(ManagementMessageDispatchStage.class);
+
+  @Override
+  public void process(ClusterEvent event) throws Exception {
+    MessageOutput messageOutput =
+        event.getAttribute(AttributeName.MESSAGES_ALL.name());
+    processEvent(event, messageOutput);
+
+    // Send participant status change messages.
```
junkaixue, line 46: Question: was this for participant pause message? I did not see the code to do that? If you are trying to cache the messages sent out, I think the original phase already did that.

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementMessageGenerationPhase.java
```diff
@@ -0,0 +1,129 @@
...
+              cache.getLiveInstances(), cache.getAllInstancesMessages(), manager.getInstanceName(),
+              manager.getSessionId());
+      MessageOutput messageOutput =
+          event.getAttributeWithDefault(AttributeName.MESSAGES_ALL.name(), new MessageOutput());
+      messageOutput.addStatusChangeMessages(messages);
+      event.addAttribute(AttributeName.MESSAGES_ALL.name(), messageOutput);
+    }
+  }
+
+  private List<Message> generateStatusChangeMessages(ClusterManagementMode.Type managementMode,
```
junkaixue, line 85: When you generate it, you should add them to cache. Not let dispatch phase to cache it.

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementMessageGenerationPhase.java
```diff
@@ -0,0 +1,129 @@
...
+      Set<String> enabledLiveInstances, Map<String, LiveInstance> liveInstanceMap,
+      Map<String, Collection<Message>> allInstanceMessages, String managerInstance,
+      String managerSessionId) {
+    List<Message> messagesToSend = new ArrayList<>();
+
+    LiveInstanceStatus fromStatus = null;
```
junkaixue, line 91: I would suggest you do the similar logic as message generation.  Use the current status of instance as "FROMSTATE" and expected state as "TOSTATE". Then check whether instance already in that state.

This logic could be more generic.

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
huizhilu: No. ResourceValidation is to match the ideal state rules, which is not required management pipeline. We can assume in management pipeline, the ideal states are validated since management pipeline is switched from default pipeline.
The pipeline switch logic from default to management is ResourceValidation. To avoid handling it in management pipeline and make the logic simple, I don't want to add ResourceValidation.

helix-core/src/main/java/org/apache/helix/api/status/ClusterManagementMode.java
```diff
@@ -67,4 +69,26 @@ public Status getStatus() {
     public Type getMode() {
         return mode;
     }
+
+    public boolean isFullyInNormalMode() {
```
jiajunwang, line 73: Question, do we have a term for partially normal status, or what does that stand for?

I assume it should be,
1. NORMAL, COMPLETE,
2. CLUSTER_PAUSE, IN_PROGRESS
3. CLUSTER_PAUSE, COMPLETE
4. CLUSTER_PAUSE, IN_PROGRESS ---> it is recovering from pause, but since it is still partially paused, I think to mark it CLUSTER_PAUSE is more appropriate.
5. NORMAL, COMPLETE

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementMessageDispatchStage.java
```diff
@@ -0,0 +1,62 @@
...
+    HelixManager manager = event.getAttribute(AttributeName.helixmanager.name());
+    List<Message> messagesSent =
+        super.sendMessages(manager.getHelixDataAccessor(), messageOutput.getStatusChangeMessages());
+    ManagementControllerDataProvider cache =
+        event.getAttribute(AttributeName.ControllerDataProvider.name());
+    cache.cacheMessages(messagesSent);
+
+    // Can exit management mode pipeline after fully in normal mode
+    ClusterManagementMode managementMode = event.getAttribute(AttributeName.CLUSTER_STATUS.name());
+    if (managementMode.isFullyInNormalMode()) {
```
jiajunwang, line 56: Why it is determined here instead of ManagementModeStage?

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementMessageGenerationPhase.java
```diff
@@ -0,0 +1,118 @@
...
+          + ". Requires HelixManager|ClusterStatus|DataCache");
+    }
+
+    PauseSignal pauseSignal = cache.getPauseSignal();
+    if (cache.getClusterConfig().isStateTransitionCancelEnabled()
+        && pauseSignal != null && pauseSignal.getCancelPendingST()) {
+      // Generate ST cancellation messages.
+      LogUtil.logInfo(LOG, _eventId,
+          "Generating ST cancellation messages for cluster " + clusterName);
+      super.process(event);
```
jiajunwang, line 67: One suggestion, put the current state to best possible state transformation here so it won't be done unless necessary.

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
```diff
@@ -66,6 +68,14 @@ public void process(ClusterEvent event) throws Exception {
     HelixDataAccessor accessor = manager.getHelixDataAccessor();
     ManagementControllerDataProvider cache =
         event.getAttribute(AttributeName.ControllerDataProvider.name());
+    CurrentStateOutput currentStateOutput =
+        event.getAttribute(AttributeName.CURRENT_STATE.name());
+    final Map<String, Resource> resourceMap =
+        event.getAttribute(AttributeName.RESOURCES_TO_REBALANCE.name());
+
+    final BestPossibleStateOutput bestPossibleStateOutput =
```
jiajunwang, line 76: I assume this is only used to cancel pending ST. Calculate it here might be confusing because it is very far away from where it is used.

jiajunwang (commented):
Overall, I think we are reusing awkwardly here. It would be cleaner if we just define a new message generation stage and dispatch stage (also the related output classes) just for the management mode. From this PR, there is not much super.XXXXXX method called.

In order to send the cancellation message, obviously we still need to call the current message generation stage and dispatch stage, but they can be 2 different stage in the pipeline. So you don't need to add your logic here and there which adds many alternative use case into the existing classes. This is not very good design from OO perspective.

helix-core/src/main/java/org/apache/helix/api/status/ClusterManagementMode.java
huizhilu: What I'd like to have is:
1. What is the desired/target mode?
2. What is the status, IN PROGRESS or COMPLETED?

`isFullyInNormalMode` means the cluster is stable in the normal mode. For the situation: recovering from pause, it should be IN_PROGRESS to NORMAL. It's already set to NORMAL mode, so we want to mark it NORMAL, IN PROGRESS.

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementMessageDispatchStage.java
huizhilu: We want to finish running the pipeline. And at the final step, decide whether if it can exit. 
The pros: 
* we don't need to throw an exception to terminate the pipeline. 
* It makes sense to determine if it should switch at the final step after finishing all the previous checks.

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementMessageGenerationPhase.java
huizhilu: I'd like to make the stages clearly do what it is supposed to do:
* ManagementModeStage: check status, prepare for message generation.
* ManagementMessageGenerationPhase: only create messages, based on the results of the previous stages.

I don't want to mix the responsibilities. So I think putting the logic: CS -> best possible in management mode stage is appropriate as preparation. It also aligns with the logic as the default pipeline: messages are generated purely based on the previous results.

helix-core/src/main/java/org/apache/helix/controller/stages/ManagementModeStage.java
huizhilu: Replied in the above thread. I don't think it's that far. It's used in the next stage. This stage prepares required output for the next stage. It makes sense, also aligned with the default/task pipelines.

junkaixue:
Overall, it looks good to me. Please resolve JJ's comments.

junkaixue (approved):

huizhilu:

PR #1834, opened 2021-08-05
Add TF Available Threads Metrics

### LINKED ISSUE #1833: Add TF available threads metrics
There's no current metrics for the available thread counts in the entire cluster. Some TF issues may be caused by unavailable threads, and it's not easy to detect this cause. Since unavailable threads is a simple case, we would like to avoid spending too much time debugging. 

Add metrics for available thread counts in the entire cluster. 

**Additional context**

NealSun96:
### Issues

Fixes #1833

### Description

There is no existing metrics about Task Framework available threads. Certain production issues are caused by simple reasons such as out of threads, and in order to spot those reasons, developers have to spend time reading through logs. 

With metrics for available threads, grouped by job types and aggregated on the cluster level, it's now easy for developers to find the thread availability of any cluster. 

### Tests

TestClusterStatusMonitor.testRecordAvailableThreadsPerType

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/task/TaskSchedulingStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/task/TaskSchedulingStage.java
@@ -81,13 +81,16 @@ public class TaskSchedulingStage extends AbstractBaseStage {
     // Reset current INIT/RUNNING tasks on participants for throttling
     cache.resetActiveTaskCount(currentStateOutput);
 
+    ClusterStatusMonitor clusterStatusMonitor =
+        event.getAttribute(AttributeName.clusterStatusMonitor.name());
     buildQuotaBasedWorkflowPQsAndInitDispatchers(cache,
-        (HelixManager) event.getAttribute(AttributeName.helixmanager.name()),
-        (ClusterStatusMonitor) event.getAttribute(AttributeName.clusterStatusMonitor.name()));
+        (HelixManager) event.getAttribute(AttributeName.helixmanager.name()), clusterStatusMonitor);
 
     final BestPossibleStateOutput bestPossibleStateOutput =
         compute(event, resourceMap, currentStateOutput);
     event.addAttribute(AttributeName.BEST_POSSIBLE_STATE.name(), bestPossibleStateOutput);
+
+    cache.getAssignableInstanceManager().recordAvailableThreadsPerType(clusterStatusMonitor);
   }
 
   private BestPossibleStateOutput compute(ClusterEvent event, Map<String, Resource> resourceMap,
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/task/TaskSchedulingStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/task/TaskSchedulingStage.java
@@ -90,7 +90,10 @@ public class TaskSchedulingStage extends AbstractBaseStage {
         compute(event, resourceMap, currentStateOutput);
     event.addAttribute(AttributeName.BEST_POSSIBLE_STATE.name(), bestPossibleStateOutput);
 
-    cache.getAssignableInstanceManager().recordAvailableThreadsPerType(clusterStatusMonitor);
+    if (clusterStatusMonitor != null) {
+      clusterStatusMonitor.updateAvailableThreadsPerJob(cache.getAssignableInstanceManager()
+          .getGlobalCapacityMap());
+    }
   }
 
   private BestPossibleStateOutput compute(ClusterEvent event, Map<String, Resource> resourceMap,
--- a/helix-core/src/main/java/org/apache/helix/task/AssignableInstanceManager.java
+++ b/helix-core/src/main/java/org/apache/helix/task/AssignableInstanceManager.java
@@ -37,6 +37,8 @@ import org.apache.helix.model.LiveInstance;
 import org.apache.helix.model.Message;
 import org.apache.helix.model.Partition;
 import org.apache.helix.model.Resource;
+import org.apache.helix.monitoring.mbeans.ClusterStatusMonitor;
+import org.apache.helix.monitoring.mbeans.JobMonitor;
 import org.apache.helix.task.assigner.AssignableInstance;
 import org.apache.helix.task.assigner.TaskAssignResult;
 import org.slf4j.Logger;
@@ -606,6 +608,17 @@ public class AssignableInstanceManager {
     }
     if (instanceNode.size() > 0) {
       LOG.info("Current quota capacity: {}", instanceNode.toString());
+    }
+  }
+
+  /**
+   * For each JobType, record their total available threads across all instances
+   * @param clusterStatusMonitor
+   */
+  public void recordAvailableThreadsPerType(ClusterStatusMonitor clusterStatusMonitor) {
+    for (String jobType : _globalThreadBasedQuotaMap.keySet()) {
+      JobMonitor jobMonitor = clusterStatusMonitor.getJobMonitor(jobType);
+      jobMonitor.updateAvailableThreadGauge((long) _globalThreadBasedQuotaMap.get(jobType));
     }
   }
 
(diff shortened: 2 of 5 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/task/TaskSchedulingStage.java
```diff
@@ -81,13 +81,16 @@ public void process(ClusterEvent event) throws Exception {
...
     buildQuotaBasedWorkflowPQsAndInitDispatchers(cache,
-        (HelixManager) event.getAttribute(AttributeName.helixmanager.name()),
-        (ClusterStatusMonitor) event.getAttribute(AttributeName.clusterStatusMonitor.name()));
+        (HelixManager) event.getAttribute(AttributeName.helixmanager.name()), clusterStatusMonitor);
 
     final BestPossibleStateOutput bestPossibleStateOutput =
         compute(event, resourceMap, currentStateOutput);
     event.addAttribute(AttributeName.BEST_POSSIBLE_STATE.name(), bestPossibleStateOutput);
+
+    cache.getAssignableInstanceManager().recordAvailableThreadsPerType(clusterStatusMonitor);
```
junkaixue, line 93: Have we done this kind of update before? I remember we usually have the method in clusterStatusMonitor to have the metrics logic in that place. If we do like this, then we are distributing monitoring logic in different places. Could be a potential issue.

helix-core/src/main/java/org/apache/helix/task/AssignableInstanceManager.java
```diff
@@ -609,6 +611,19 @@ public void logQuotaProfileJSON(boolean onlyDisplayIfFull) {
...
 
+  /**
+   * For each JobType, record their total available threads across all instances
+   * @param clusterStatusMonitor
+   */
+  public void recordAvailableThreadsPerType(ClusterStatusMonitor clusterStatusMonitor) {
+    if (clusterStatusMonitor != null) {
+      for (String jobType : _globalThreadBasedQuotaMap.keySet()) {
+        JobMonitor jobMonitor = clusterStatusMonitor.getJobMonitor(jobType);
+        jobMonitor.updateAvailableThreadGauge((long) _globalThreadBasedQuotaMap.get(jobType));
```
junkaixue, line 622: Question: Was this job type consistent with workflow type? I am not quite sure whether there could be some diverge of monitoring at job level against workflow level.

helix-core/src/main/java/org/apache/helix/controller/stages/task/TaskSchedulingStage.java
NealSun96: The reason why I did it this way is because `AssignableInstanceManager` already keeps a map of available threads per job type across the cluster, so if we have another place keeping track of the threads, it seems to be duplication. 
I want to draw parallel to how `AbstractTaskDispatcher` uses cluster status monitor: we set the monitor as its local variable, then the dispatcher uses the monitor. It's pretty similar to this use case in my opinion.

helix-core/src/main/java/org/apache/helix/task/AssignableInstanceManager.java
NealSun96: They are the same thing: if workflow has a type, it applies to all jobs, otherwise job can have their own types. Since each job can have a different type, would `jobType` make more sense? If we refer to it as `workflowType` and use WorkflowMonitor, then each workflow can have different types of jobs and affects different WorkflowMonitors, which is confusing.

junkaixue: I would suggest to use workflow type. We had the idea to hold different workflow type for different jobs, like a queue can handle some group of job types. Also make the workflow type could link it as the source of truth since the job type now is derived from workflow type. Logically, it could make more sense to let workflow level reporting.

helix-core/src/main/java/org/apache/helix/controller/stages/task/TaskSchedulingStage.java
junkaixue: Where to keep the data should be fine. But it depends on the where the logic is. 
You can do the reserve way to update it. For example, you can pass in the cache AssignableInstanceManager as an argument for clusterStatusMonitor.updateQuota() (you can name it differently). But keep the logic of how you update the data inside the clusterStatusMonitor. You dont have to copy the data to clusterStatusMonitor.

helix-core/src/main/java/org/apache/helix/controller/stages/task/TaskSchedulingStage.java
```diff
@@ -81,13 +81,16 @@ public void process(ClusterEvent event) throws Exception {
     // Reset current INIT/RUNNING tasks on participants for throttling
     cache.resetActiveTaskCount(currentStateOutput);
 
+    ClusterStatusMonitor clusterStatusMonitor =
+        event.getAttribute(AttributeName.clusterStatusMonitor.name());
```
narendly, line 85: Should we just do a null check here and remove null checks in other places (e.g. JobMonitor)?

helix-core/src/main/java/org/apache/helix/task/AssignableInstanceManager.java
```diff
@@ -609,6 +611,19 @@ public void logQuotaProfileJSON(boolean onlyDisplayIfFull) {
     }
   }
 
+  /**
+   * For each JobType, record their total available threads across all instances
+   * @param clusterStatusMonitor
+   */
+  public void recordAvailableThreadsPerType(ClusterStatusMonitor clusterStatusMonitor) {
+    if (clusterStatusMonitor != null) {
```
narendly, line 619: This kind of null check should ideally be done before `ClusterStatusMonitor` is passed in to `JobMonitor` to make the code more concise to read. See comment above :)

helix-core/src/main/java/org/apache/helix/task/AssignableInstanceManager.java
narendly: If I recall the API behavior correclty, if workflowType is defined, I believe all jobs inherit their parent workflow's type. Even if the user sets the job type (in addition to setting the workflow type), the parent workflow's type will override the job type.

If workflowType is not set, then users can set a different type for each individual jobType. meaning each job can have its own type.

Since it's possible for jobs to have different types when the parent workflow's type is not defined, I don't think we can use the workflow type here?

helix-core/src/main/java/org/apache/helix/controller/stages/task/TaskSchedulingStage.java
narendly: For the sake of cutting down on redundant computation, is it possible at all to move this "record" (basically a few loops that count) logic to where the global quota calculation is done so we don't loop through things twice?

NealSun96: @junkaixue I see, let me try that. 
@narendly I considered that earlier: the global quota is updated in several methods; it's not just one computation at a central place. Plus, based on what @junkaixue said, we want to "record" inside `ClusterStatusMonitor`, instead of doing it in `AssiganbleInstanceManager`. Lastly, it's 1 loop through each job type, which I don't think is too computation heavy (it doesn't loop through all instances).

helix-core/src/main/java/org/apache/helix/task/AssignableInstanceManager.java
NealSun96: @junkaixue I don't really understand your suggestion. Are you saying to not use `_globalThreadBasedQuotaMap`? Or are you saying to use `workflowMonitor` instead of `jobMonitor`?

helix-core/src/main/java/org/apache/helix/controller/stages/task/TaskSchedulingStage.java
NealSun96: Since @junkaixue also commented about doing the work in `ClusterStatusMonitor` instead, we should definitely do the null check here. Thanks for the suggestion.

helix-core/src/main/java/org/apache/helix/task/AssignableInstanceManager.java
junkaixue: @NealSun96 No need for this. I think @narendly is right. There would be meaningless to have workflow level aggregation of metrics since assignment is happened at job level.

junkaixue (approved):
LGTM

NealSun96:
Final commit message:
## Add TF Available Threads Metrics ##
Add metrics about Task Framework available threads in the cluster per job type.

PR #1842, opened 2021-08-18
Fix TestClusterAggregateMetrics

### LINKED ISSUE #1806: [Failed CI Test] testAggregateMetrics(org.apache.helix.monitoring.mbeans.TestClusterAggregateMetrics)
This issue is created for tracking unstable test: testAggregateMetrics(org.apache.helix.monitoring.mbeans.TestClusterAggregateMetrics)

NealSun96:
### Issues

Fixes #1806

### Description

Fix flaky TestClusterAggregateMetrics by wrapping maintenance mode around batch enabling of instances, otherwise the events may be processed separately, resulting in an intermediate best possible state that does not represent the final state. 

### Tests

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/monitoring/mbeans/TestClusterAggregateMetrics.java
+++ b/helix-core/src/test/java/org/apache/helix/monitoring/mbeans/TestClusterAggregateMetrics.java
@@ -151,17 +151,22 @@ public class TestClusterAggregateMetrics extends ZkTestBase {
             .build();
 
     // Everything should be up and running initially with 5 total partitions
-    updateMetrics();
-    Assert.assertEquals(_beanValueMap.get(PARTITION_COUNT), 5L);
-    Assert.assertEquals(_beanValueMap.get(ERROR_PARTITION_COUNT), 0L);
-    Assert.assertEquals(_beanValueMap.get(WITHOUT_TOPSTATE_COUNT), 0L);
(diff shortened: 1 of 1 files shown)


junkaixue (approved):

alirezazamani (approved):

NealSun96:
Final commit message:
## Fix TestClusterAggregateMetrics ##

jiajunwang:
A better way, or IMHO the correct way, is to wait until the cluster state converged. The main reason we might get unexpected results is that the controller is still rebalancing the cluster while the assert check happens. Given that saying, the current change will reduce the error by reducing pipeline running. But it is not guaranteed since only one pipeline can still run slowly.

Could you please change the test logic to wait until the cluster converges then start validation? Note the verifier should support validate instance state in the call. So you can specify the expected disabled nodes in the same check to ensure verifier does not return prematurely.

jiajunwang (changes requested):
As I commented, I think the current fix does not completely fix the problem. Please take a look and let me know if you think it makes sense : )

NealSun96:
@jiajunwang On line 178 we have `Assert.assertTrue(verifier.verifyByPolling());` That checks for cluster convergence like you said. 

The previous test causes the verifier to mistakenly verify an intermediate state instead of the final state; adding the maintenance mode should mitigate that, and let the verifier do its job.

jiajunwang (approved):
LGTM!

Issue #1849, opened 2021-08-24
Use activate/deactivate as keywords in partitionAssignment API.


xyuanlu:
**Is your feature request related to a problem? Please describe.**
Current partitionAssignment has ambiguous keyword. `Add/RemoveInstance` may indicate online/offline state or enabled/disabled state. 

**Describe the solution you'd like**
Use activate/deactivateInstances for keyword in partitionAssignment.

PR #1850, opened 2021-08-24
Use activate/DeactivateInstances as keyword in PartitionAssignment API

xyuanlu:
### Issues

#1849 

### Description

This PR changes the keyword add/remove instances to activate/deactivateInstances in partitionAssignment. 

### Tests

TestResourceAssignmentOptimizerAccessor

```
[INFO] Results:
[INFO] 
[INFO] Tests run: 179, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  05:34 min
[INFO] Finished at: 2021-08-23T19:10:18-07:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
@@ -22,6 +22,8 @@ package org.apache.helix.rest.server.resources.helix;
 import java.security.InvalidParameterException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -186,12 +188,7 @@ public class ResourceAssignmentOptimizerAccessor extends AbstractHelixResource {
 
     // One instance can only exist in one of the list in InstanceChange.
     // Validate the intersection is empty.
-    for (String instance : inputFields.activatedInstances) {
-      if (inputFields.deactivatedInstances.contains(instance)) {
-        LOG.error(
-            "Instance " + instance + " in both ActivateInstances and DeactivateInstances list.");
-      }
-    }
+    validateNoIntxnInstanceChange(inputFields);
 
     // Add instances to current liveInstances
     ClusterState clusterState = new ClusterState();
@@ -204,24 +201,31 @@ public class ResourceAssignmentOptimizerAccessor extends AbstractHelixResource {
     liveInstancesSet.addAll(
         inputFields.activatedInstances.stream().filter(s -> !liveInstancesSet.contains(s))
             .collect(Collectors.toList()));
+    liveInstancesSet.removeAll(inputFields.deactivatedInstances);
+
+    Map<String, InstanceConfig> instanceConfigMap =
+        dataAccessor.getChildValuesMap(dataAccessor.keyBuilder().instanceConfigs(), true);
 
     // Read instance and cluster config.
     // Throw exception if there is no instanceConfig for activatedInstances instance.
-    for (String instance : liveInstancesSet) {
-      InstanceConfig config = cfgAccessor.getInstanceConfig(clusterId, instance);
-      // override HELIX_ENABLED field.
-      if (inputFields.activatedInstances.contains(instance)) {
-        config.setInstanceEnabled(true);
+    for (String instance : inputFields.activatedInstances) {
+      if (instanceConfigMap.containsKey(instance)) {
+        instanceConfigMap.get(instance).setInstanceEnabled(true);
+      } else {
+        throw new InvalidParameterException(
+            "instance: " + instance + "does not have instanceConfig");
       }
-      if (inputFields.deactivatedInstances.contains(instance)) {
-        config.setInstanceEnabled(false);
-      }
-      clusterState.instanceConfigs.add(config);
     }
-    liveInstancesSet.removeAll(inputFields.deactivatedInstances);
+
+    for (String instance : inputFields.deactivatedInstances) {
+      if (instanceConfigMap.containsKey(instance)) {
+        instanceConfigMap.get(instance).setInstanceEnabled(false);
+      }
+    }
 
     clusterState.clusterConfig = cfgAccessor.getClusterConfig(clusterId);
     clusterState.liveInstances = new ArrayList<>(liveInstancesSet);
+    clusterState.instanceConfigs.addAll(instanceConfigMap.values());
     return clusterState;
   }
 
@@ -289,6 +293,32 @@ public class ResourceAssignmentOptimizerAccessor extends AbstractHelixResource {
       return currentStateFormatResult;
     }
     return idealStateFormatResult;
+  }
+
+  /*
+  * Return if there are instance exists in more than one lists in InstanceChangeMap
+  */
+  private void validateNoIntxnInstanceChange(InputFields inputFields) {
+    Set<String> tempSet = new HashSet<>();
+    List<Collection<String>> inputs = new ArrayList<>();
+    inputs.add(inputFields.activatedInstances);
+    inputs.add(inputFields.deactivatedInstances);
+    inputs.sort(Comparator.comparingInt(Collection::size));
+
+    for (int i = 0; i < inputs.size() - 1; ++i) {
+      for (String s : inputs.get(i)) {
+        if (!tempSet.add(s)) {
+          throw new InvalidParameterException("Invalid input: instance [" + s
+              + "] exist in more than one field in InstanceChange.");
+        }
+      }
+    }
+    for (String s : inputs.get(inputs.size() - 1)) {
+      if (tempSet.contains(s)) {
+        throw new InvalidParameterException(
+            "Invalid input: instance [" + s + "] exist in more than one field in InstanceChange.");
+      }
+    }
   }
 
   private void computeWagedAssignmentResult(List<IdealState> wagedResourceIdealState,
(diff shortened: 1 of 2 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -190,42 +184,44 @@ private InputFields readInput(String content)
   private ClusterState readClusterStateAndValidateInput(String clusterId, InputFields inputFields)
       throws InvalidParameterException {
 
+    // One instance can only exist in one of the list in InstanceChange.
+    // Validate the intersection is empty.
+    for (String instance : inputFields.activatedInstances) {
```
jiajunwang, line 189: To be more generic, consider we may have more lists soon, let's use a temporary hashset to check if an instance is mentioned more than once in all the lists.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -190,42 +184,44 @@ private InputFields readInput(String content)
...
+      if (inputFields.activatedInstances.contains(instance)) {
+        config.setInstanceEnabled(true);
+      }
+      if (inputFields.deactivatedInstances.contains(instance)) {
+        config.setInstanceEnabled(false);
+      }
       clusterState.instanceConfigs.add(config);
     }
-    clusterState.clusterConfig = cfgAccessor.getClusterConfig(clusterId);
+    liveInstancesSet.removeAll(inputFields.deactivatedInstances);
```
jiajunwang, line 221: nit, move to line 207 so all live instance operations happen at the same place?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -190,42 +184,44 @@ private InputFields readInput(String content)
...
+        new HashSet<>(dataAccessor.getChildNames(dataAccessor.keyBuilder().liveInstances()));
+    liveInstancesSet.addAll(
+        inputFields.activatedInstances.stream().filter(s -> !liveInstancesSet.contains(s))
+            .collect(Collectors.toList()));
 
     // Read instance and cluster config.
-    // It will throw exception if there is no instanceConfig for newly added instance.
-    for (String instance : clusterState.instances) {
+    // Throw exception if there is no instanceConfig for activatedInstances instance.
+    for (String instance : liveInstancesSet) {
```
jiajunwang, line 210: 1. if an instance is not alive, but it has an instance config, do we still add it to the input config list?
2. It seems that you will throw exception for the deactivated instances too if the corresponding config does not exist.

xyuanlu: Please correct me if I am wrong. When the code reach here, the `liveInstancesSet` contains current live instances and the instances in `activatedInstances` list? 
For instances in deactivated, these instances can only be in this list if they are currently under liveInstance, so they have corresponding config.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
xyuanlu: If we move this line to 207, then instances in deactivatedInstances list will be override to be in `out of cluster` state rather than `offline & disabled` state..?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
xyuanlu: I had a temporary hashset to check duplication when swapInstance is in the scope. Since we only have 2 keyword, won't having a separate set be a bit overkill?

jiajunwang: I'm pretty sure we will have more lists soon. Keeping a temp hashset is not a big overhead. Let's just add it to avoid refactoring.
Or if you want to do shortcut now and refactor later, then we should use stream() to simplify code.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
jiajunwang: It is not a correct design to check instance config according to live instance. After you fix this, then the concern you described will go away.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
jiajunwang: Regarding the 2nd point, let's try to stick with the design. For deactivate, if the node does not exist at all, it is totally fine.

xyuanlu: In this implementation, it does not check if node exists for deactivated only check for activatedInstances.  
In the new commit, it reads all instance config (not only for live instances), and won't check if node exists for deactivated either. Only change the enabled state to disabled for to-be-deactivated instances.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
xyuanlu: TFTR, updated

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
xyuanlu: TFTR.
Using steam or using a temp hashset will need some extra space. But I agree it won't cause big overhead.
Updated.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -190,42 +186,46 @@ private InputFields readInput(String content)
...
-          throw new InvalidParameterException("Invalid input: instance [" + instanceToRemove
-              + "] in RemoveInstances does not exist in cluster.");
-        }
-      }
-      if (!inputFields.instancesToRemove.isEmpty()) {
-        clusterState.instances.removeIf(inputFields.instancesToRemove::contains);
+    Set<String> liveInstancesSet =
+        new HashSet<>(dataAccessor.getChildNames(dataAccessor.keyBuilder().liveInstances()));
+    liveInstancesSet.addAll(
+        inputFields.activatedInstances.stream().filter(s -> !liveInstancesSet.contains(s))
```
jiajunwang, line 202: We don't need the filter since liveInstancesSet will skip duplicate items, right?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
```diff
@@ -316,6 +295,32 @@ private AssignmentResult updateAssignmentFormat(InputFields inputFields,
     return idealStateFormatResult;
   }
 
+  /*
+  * Return if there are instance exists in more than one lists in InstanceChangeMap
+  */
+  private void validateNoIntxnInstanceChange(InputFields inputFields) {
+    Set<String> tempSet = new HashSet<>();
+    List<Collection<String>> inputs = new ArrayList<>();
```
jiajunwang, line 303: Why we need this list? Adding and checking with the tempSet only should be enough?

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
xyuanlu: Yes. I was about to change this.

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/ResourceAssignmentOptimizerAccessor.java
xyuanlu: Should be enough. 
What I have in mind is that, when we have multiple list to check for intersection, I do not want to add all the list to the hashSet. The largest one could be read only. It would safe some space....

jiajunwang: I see. My preference is simpler code. But it's not a strong preference. Your call.

jiajunwang: But on my second thought, the sorting will be expensive comparing with the HashSet addition, no?

jiajunwang (approved):

xyuanlu:
Thanks for the review! 
This API is ready to be merged.  Approved by @jiajunwang 
Commit message:
"Use activate/DeactivateInstances as keyword in PartitionAssignment API"

PR #1870, opened 2021-09-13
Improve Purge Offline Instances API

### LINKED ISSUE #1869: PurgeOfflineInstances should clean up race condition leftovers
As a cleanup API, `purgeOfflineInstances` has limitations: it cannot handle the case when participantHistory is missing. This is because `purgeOfflineInstances` uses the same underlying code as `dropInstance`. That means after one round of purging, race conditions could write unwanted data (such as messages) back to ZK, and they will never be cleaned up because participantHistory is no longer present. 

`purgeOfflineInstances` should handle the unwanted data caused by race conditions. 

**Additional context**

NealSun96:
### Issues

Fixes #1869

### Description

The `purgeOfflineInstances` API is used to clean up the instance metadata in ZK that is left by offline instances. The API is only used during special use cases where a large amount of instance metadata is left uncleaned; instance metadata should usually be cleaned up manually. 

As a cleanup API, `purgeOfflineInstances` has limitations: it cannot handle the case when participantHistory is missing. This is because `purgeOfflineInstances` uses the same underlying code as `dropInstance`. That means after one round of purging, race conditions could write unwanted data (such as messages) back to ZK, and they will never be cleaned up because participantHistory is no longer present. 

This PR improves the API such that it will also purge any incomplete instance data, such as instance path without InstanceConfig or ParticipantHistory. 

### Tests

Modified TestZkHelixAdmin.testPurgeOfflineInstances

```
[INFO] Tests run: 1283, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,709.742 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1283, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] Loading execution data file /home/nesun/helix/helix-core/target/jacoco.exec
[INFO] Analyzed bundle 'Apache Helix :: Core' with 909 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  01:35 h
[INFO] Finished at: 2021-09-10T19:14:54-07:00
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
@@ -288,12 +288,6 @@ public class ZKHelixAdmin implements HelixAdmin {
   private void purgeInstance(String clusterName, String instanceName) {
     logger.info("Purge instance {} from cluster {}.", instanceName, clusterName);
 
-    String liveInstancePath = PropertyPathBuilder.liveInstance(clusterName, instanceName);
-    if (_zkClient.exists(liveInstancePath)) {
-      throw new HelixException(
-          "Node " + instanceName + " is still alive for cluster " + clusterName + ", can't purge.");
-    }
-
     String instanceConfigPath = PropertyPathBuilder.instanceConfig(clusterName, instanceName);
     _zkClient.delete(instanceConfigPath);
     String instancePath = PropertyPathBuilder.instance(clusterName, instanceName);
@@ -2209,23 +2203,25 @@ public class ZKHelixAdmin implements HelixAdmin {
     List<String> instancePathNames = accessor.getChildNames(keyBuilder.instances());
     List<String> liveNodes = accessor.getChildNames(keyBuilder.liveInstances());
 
-    Set<String> timeoutOfflineInstanceNames = new HashSet<>();
-    timeoutOfflineInstanceNames.addAll(instanceConfigNames);
-    timeoutOfflineInstanceNames.addAll(instancePathNames);
-    liveNodes.forEach(timeoutOfflineInstanceNames::remove);
-    for (String instanceName : timeoutOfflineInstanceNames) {
+    Set<String> offlineInstanceNames = new HashSet<>(instancePathNames);
+    liveNodes.forEach(offlineInstanceNames::remove);
+    long finalOfflineDuration = offlineDuration;
+    offlineInstanceNames.removeIf(instanceName -> {
       ParticipantHistory participantHistory =
           accessor.getProperty(keyBuilder.participantHistory(instanceName));
-
-      // If participant history is null, a race condition happened and should be cleaned up
-      // Otherwise, if the participant has not been offline for more than the duration, no clean up
-      if (participantHistory != null
-          && (participantHistory.getLastOfflineTime() == ParticipantHistory.ONLINE
-          || System.currentTimeMillis() - participantHistory.getLastOfflineTime()
-          < offlineDuration)) {
-        timeoutOfflineInstanceNames.remove(instanceName);
+      if (participantHistory == null && instanceConfigNames.contains(instanceName)) {
+        // this is likely caused by a new instance joining and should not be purged
+        return true;
       }
-    }
-    return timeoutOfflineInstanceNames;
+      // If participant history is null without config, a race condition happened and should be
+      // cleaned up.
+      // Otherwise, if the participant has not been offline for more than the duration, no clean up
+      return (participantHistory != null && (
+          participantHistory.getLastOfflineTime() == ParticipantHistory.ONLINE
+              || System.currentTimeMillis() - participantHistory.getLastOfflineTime()
+              < finalOfflineDuration));
+    });
+
+    return offlineInstanceNames;
   }
 }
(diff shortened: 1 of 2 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -284,6 +285,21 @@ public void purgeOfflineInstances(String clusterName, long offlineDuration) {
     }
   }
 
+  private void purgeInstance(String clusterName, String instanceName) {
+    logger.info("Purge instance {} from cluster {}.", instanceName, clusterName);
+
+    String liveInstancePath = PropertyPathBuilder.liveInstance(clusterName, instanceName);
```
jiajunwang, line 291: Why do we need to check liveInstancePath here? For this scenario, you already checked live instance in findTimeoutOfflineInstances, right? So this function should do pure removing.
Moreover, with this assumption, you can just refactor the code to:
dropInstance() {
//check;
purgeInstance();
}
purgeInstance() {
// remove all ndoes.
}

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -2176,36 +2192,40 @@ public ZKHelixAdmin build() {
...
       ParticipantHistory participantHistory =
           accessor.getProperty(keyBuilder.participantHistory(instanceName));
-      long lastOfflineTime = participantHistory.getLastOfflineTime();
-      if (lastOfflineTime == ClusterConfig.OFFLINE_DURATION_FOR_PURGE_NOT_SET
-          || System.currentTimeMillis() - lastOfflineTime < offlineDuration) {
-        toRemoveInstances.add(instanceName);
+
+      // If participant history is null, a race condition happened and should be cleaned up
+      // Otherwise, if the participant has not been offline for more than the duration, no clean up
+      if (participantHistory != null
```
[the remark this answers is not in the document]
NealSun96: You're correct that the newly created nodes may be impacted. This is a conundrum: the whole point of this change is to clean up leftover ZNodes caused by race conditions (messages sent to deleted ZNodes) because otherwise those leftovers can never be purged, however, fundamentally these ZNodes cannot be distinguished from the partial creation of new instances. 

The reason why I did it this way is because the pervious implementation warns to not use this endpoint while instances are joining. The assumption is the same for this implementation, then we also don't need to worry about new instances (although it has changed from "while no old instances are rejoining" to "while no instances are joining, old or new"). 

If we really want to keep the behavior as before, we can check for two cases:
1. Existing InstanceConfig but no History -> partial creation of new instances, don't purge;
2. Non-existing InstanceConfig and no History -> race condition, purge.
But do you think that's necessary, given the previously mentioned assumption of using this endpoint?

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
NealSun96: I adopted a similar style with the previous code, but I agree with the refactoring suggestion.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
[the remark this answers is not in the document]
jiajunwang: I guess as long as we keep the assumption that no new node will join the cluster during purging, the logic should be fine. However, it is less ideal since there is no mechanism to enforce it. For now, I think additional checks, which introduce a lot of complexity, might not be enough to cover all the use cases. Let's just keep it simple.
But I would suggest adding a warning log if an instance that is being deleted might be a new node.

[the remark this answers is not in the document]
NealSun96: It didn't sit well with me so I changed the logic - now it should exclude new instances. The logic is not complex in the end.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
```diff
@@ -231,8 +231,11 @@ public void dropInstance(String clusterName, InstanceConfig instanceConfig) {
     // delete config path
```
jiajunwang, line 231: nit, these duplicates with purgeInstance(), can we merge them?

I know it will cause two "String instancePath = PropertyPathBuilder.instance(clusterName, instanceName);" in 2 methods. But IMO, it cleans up the code further.

No strong preference here.

jiajunwang (approved):

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixAdmin.java
NealSun96: I'd rather not since it's not a strict duplication: `dropInstance` is based on `InstanceConfig`, `purgeInstance` is based on instance name. We can make `dropInstance` call `purgeInstance` but that's a lot of twist and turns.

NealSun96:
Final commit message:
## Improve Purge Offline Instances API ##
This PR improves the API such that it will also purge any incomplete instance data, such as instance path without InstanceConfig or ParticipantHistory.

PR #1875, opened 2021-09-16
Fix adding a task to a job after deleting old tasks

### LINKED ISSUE #1874: Adding new task would be successful if there is gap in pIds
### Describe the bug
If some of the tasks from an IN_PROGRESS job have been removed before, new tasks cannot be added to the job because newly added tasks would have same pId to some of the existing tasks.

### To Reproduce
Assume a job has n tasks (pID = [0,n-1]). Remove one of the tasks with pId != n-1. After that new tasks cannot be added.

### Expected behavior
If there is a gap in the pIds, the new task should be added as the missing pId

alirezazamani:
### Issues

Fixes #1874 

### Description

In this PR, the issue of dynamically adding a task to a job which some of its tasks has been deleted before is being addressed.

### Tests

TestAddDeleteTask.testDeleteMiddleTaskAndAdd

Helix-core:
```
[INFO] Tests run: 1284, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,522.236 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1284, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] Loading execution data file /home/anajari/my_repos/helix/helix-core/target/jacoco.exec
[INFO] Analyzed bundle 'Apache Helix :: Core' with 909 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  01:32 h
[INFO] Finished at: 2021-09-16T08:47:28-07:00
```
Helix-rest:
```
[INFO] Tests run: 179, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 126.955 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 179, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] Loading execution data file /home/anajari/my_repos/helix/helix-rest/target/jacoco.exec
[INFO] BUILD SUCCESS
[INFO] Total time:  02:14 min
[INFO] Finished at: 2021-09-16T09:38:30-07:00
```

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/TaskAssignmentCalculator.java
+++ b/helix-core/src/main/java/org/apache/helix/task/TaskAssignmentCalculator.java
@@ -20,6 +20,7 @@ package org.apache.helix.task;
  */
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
@@ -120,5 +121,40 @@ public abstract class TaskAssignmentCalculator {
       }
     }
     return deletedPartitions;
+  }
+
+  /**
+   * Get all the partitions/tasks that belong to the non-targeted job.
+   * @param jobCfg the task configuration
+   * @param jobCtx the task context
+   * @return set of partition numbers
+   */
+  public Set<Integer> getAllTaskPartitionsDefault(JobConfig jobCfg, JobContext jobCtx) {
+    Map<String, TaskConfig> taskMap = jobCfg.getTaskConfigMap();
+    Map<String, Integer> taskIdMap = jobCtx.getTaskIdPartitionMap();
+    // Check if a gap exists in the context due to previouslys
+    // removed tasks. If yes, the missing pIDs should be considered
+    // for newly added tasks
+    Set<Integer> existingPartitions = jobCtx.getPartitionSet();
+    Set<Integer> missingPartitions = new HashSet<>();
+    if (existingPartitions.size() != 0) {
+      for (int pId = 0; pId < Collections.max(existingPartitions); pId++) {
+        if (!existingPartitions.contains(pId)) {
+          missingPartitions.add(pId);
+        }
+      }
+    }
+    for (TaskConfig taskCfg : taskMap.values()) {
+      String taskId = taskCfg.getId();
+      if (!taskIdMap.containsKey(taskId)) {
+        int nextPartition = jobCtx.getPartitionSet().size();
+        if (missingPartitions.size()!=0) {
+          nextPartition = missingPartitions.iterator().next();
+          missingPartitions.remove(nextPartition);
+        }
+        jobCtx.setTaskIdForPartition(nextPartition, taskId);
+      }
+    }
+    return jobCtx.getPartitionSet();
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
+++ b/helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
@@ -20,6 +20,7 @@ package org.apache.helix.task;
  */
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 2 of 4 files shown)


helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
```diff
@@ -65,10 +66,26 @@ public ThreadCountBasedTaskAssignmentCalculator(TaskAssigner taskAssigner,
...
-      int nextPartition = jobCtx.getPartitionSet().size();
       if (!taskIdMap.containsKey(taskId)) {
+        int nextPartition = jobCtx.getPartitionSet().size();
+        if (missingPartitions.size()!=0) {
+          nextPartition = missingPartitions.iterator().next();
+          missingPartitions.remove(nextPartition);
+        }
         jobCtx.setTaskIdForPartition(nextPartition, taskId);
       }
     }
```
NealSun96, line 91: Is this function exactly the same as the previous one? Is there a way we can combine them and not repeat?

helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
```diff
@@ -65,10 +66,26 @@ public ThreadCountBasedTaskAssignmentCalculator(TaskAssigner taskAssigner,
...
+        }
+      }
+    }
     for (TaskConfig taskCfg : taskMap.values()) {
       String taskId = taskCfg.getId();
-      int nextPartition = jobCtx.getPartitionSet().size();
       if (!taskIdMap.containsKey(taskId)) {
+        int nextPartition = jobCtx.getPartitionSet().size();
+        if (missingPartitions.size()!=0) {
+          nextPartition = missingPartitions.iterator().next();
```
NealSun96, line 86: Is it better to instead start with "the largest partitionID in jobCtx", and simply +1 every iteration? The fill-the-gaps approach seems unnecessarily complicated, if I'm not missing anything.

NealSun96 (changes requested):

helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
alirezazamani: I thought about it actually. It was already duplicated. I think we keep a generic one for backward compatibility.

helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
alirezazamani: So the issue here is:
Let's assume we have these tasks [0,1,2,3,4]
Now the user deletes task 3 and we will have [0,1,2,4]
Now if the new task is being added, the size is 4 and the new pId will be 4 instead of 3. Hence the task would be be on context as we have a collision.
That is why we need to find the missing tasks.

junkaixue: I have concerns for you are using missing partitions. Let's say user removed 3 as you mentioned, the job is not fully cancelled and still in running stage. If they add another task and you reuse missing partition 3 as the job id. Will that maps to existing non-cancelled task and make the system in a wired state?

alirezazamani: That would not be the case. The design of the delete is to remove the task from context after it has been dropped completely and the current state is gone. It means after the cancellation we will remove context. In this case, it is safe to add a new tass with ID=3.

helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
```diff
@@ -65,10 +66,26 @@ public ThreadCountBasedTaskAssignmentCalculator(TaskAssigner taskAssigner,
       Map<String, IdealState> idealStateMap) {
     Map<String, TaskConfig> taskMap = jobCfg.getTaskConfigMap();
     Map<String, Integer> taskIdMap = jobCtx.getTaskIdPartitionMap();
+    // Check if a gap exists in the context due to previouslys
```
junkaixue, line 69: The function is exactly same as what we have in GenericTaskTaskAssignmentCalculator. Can we have two function wrapped in TaskAssignmentCalculator. So these two calculator just call it?

junkaixue (approved):
Overall, lgtm. Please address the previous comment.

helix-core/src/main/java/org/apache/helix/task/ThreadCountBasedTaskAssignmentCalculator.java
alirezazamani: Done.

alirezazamani:

Final commit message:
Fix adding a task to a job after deleting old tasks

In this commit, the issue of dynamically adding a task to a job
in which some of its tasks have been deleted before is being
addressed.

helix-core/src/main/java/org/apache/helix/task/TaskAssignmentCalculator.java
```diff
@@ -121,4 +122,39 @@ public static String getQuotaType(WorkflowConfig workflowConfig, JobConfig jobCo
...
+   * Get all the partitions/tasks that belong to the non-targeted job.
+   * @param jobCfg the task configuration
+   * @param jobCtx the task context
+   * @return set of partition numbers
+   */
+  public Set<Integer> getAllTaskPartitionsDefault(JobConfig jobCfg, JobContext jobCtx) {
+    Map<String, TaskConfig> taskMap = jobCfg.getTaskConfigMap();
+    Map<String, Integer> taskIdMap = jobCtx.getTaskIdPartitionMap();
+    // Check if a gap exists in the context due to previouslys
+    // removed tasks. If yes, the missing pIDs should be considered
```
jiajunwang, line 136: This might cause problems if we want to debug for what happened to a removed task from the log. Ideally, we should avoid any overlap even with the removed task Id. However, I also understand that even we pick up the current max Id, there is no guarantee the next number has not been used before.
Let's keep the design.

One question, though, can we just use the task Id to replace task Partition Id completely? That will resolve all the problems, right?

PR #1910, opened 2021-11-21
upgrade maven-site-plugin: fix site.xml head/footer

hboutemy:
I only did a few versions, not all :)


junkaixue:
Thank you @hboutemy ! Will try to fix others!

PR #1935, opened 2022-01-14
Implement java API and utils for virtual topology group

qqu0127:
### Description

Implement the java endpoint for adding virtual topology group for a cluster. The entry point is `VirtualTopologyGroupService#addVirtualTopologyGroup`.
This provides a virtualization layer on top of physical fault zones, so that we have more control over the topology on cloud environment where change or override may not be supported. This virtualization layer doesn't provide more fault tolerance, but allows more flexibility in zone topology setup.

This API is only expected to be invoked in cloud environment where `cloudConfig.isCloudEnabled()` is true. It's not designed for and not supported for on-prem environment.

#### What the java API does:
`void addVirtualTopologyGroup(String clusterName, Map<String, String> customFields)`

- **Validation**

Basic validation on input params, state check (isCloudEnabled, isTopologyAwareEnabled)

- **Compute virtual group assignment**

From current cluster topology, compute the virtual group assignment based on given params.

- **Ensure maintenance mode**

Ensure the cluster is in maintenance mode before next steps, either service enters it or user specifies to control it themselves outside the API lifecycle.

- **Update configs**

Override cluster topology and fault zone type configs.
Update all instance configs by adding `virtualZone` field in the domain.
 
Some previous PRs for reference:
https://github.com/apache/helix/pull/1945
https://github.com/apache/helix/pull/1948 

### Tests

`TestVirtualTopologyGroupService`

[INFO] Tests run: 198, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 227.077 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 198, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] Loading execution data file /Users/qqu/workspace/qqu-helix/helix-rest/target/jacoco.exec
[INFO] Analyzed bundle 'Apache Helix :: Restful Interface' with 87 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  03:51 min
[INFO] Finished at: 2022-02-03T13:39:34-08:00
### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
@@ -49,6 +49,9 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Service for virtual topology group.
+ * It's a virtualization layer on top of physical fault domain and topology in cloud environments.
+ * The service computes the mapping from virtual group to instances based on the current cluster topology and update the
+ * information to cluster and all instances in the cluster.
  */
 public class VirtualTopologyGroupService {
   private static final Logger LOG = LoggerFactory.getLogger(VirtualTopologyGroupService.class);
--- a/helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
@@ -1,0 +1,194 @@
+package org.apache.helix.rest.server.service;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.helix.AccessOption;
+import org.apache.helix.ConfigAccessor;
+import org.apache.helix.HelixAdmin;
+import org.apache.helix.HelixDataAccessor;
+import org.apache.helix.HelixException;
+import org.apache.helix.PropertyPathBuilder;
+import org.apache.helix.cloud.constants.VirtualTopologyGroupConstants;
+import org.apache.helix.cloud.topology.FifoVirtualGroupAssignmentAlgorithm;
+import org.apache.helix.cloud.topology.VirtualGroupAssignmentAlgorithm;
+import org.apache.helix.model.CloudConfig;
+import org.apache.helix.model.ClusterConfig;
+import org.apache.helix.model.ClusterTopologyConfig;
+import org.apache.helix.model.InstanceConfig;
+import org.apache.helix.rest.server.json.cluster.ClusterTopology;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.apache.helix.zookeeper.zkclient.DataUpdater;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Service for virtual topology group.
+ */
+public class VirtualTopologyGroupService {
+  private static final Logger LOG = LoggerFactory.getLogger(VirtualTopologyGroupService.class);
+
+  private final HelixAdmin _helixAdmin;
+  private final ClusterService _clusterService;
+  private final ConfigAccessor _configAccessor;
+  private final HelixDataAccessor _dataAccessor;
+  private final VirtualGroupAssignmentAlgorithm _assignmentAlgorithm;
+
+  public VirtualTopologyGroupService(HelixAdmin helixAdmin, ClusterService clusterService,
+      ConfigAccessor configAccessor, HelixDataAccessor dataAccessor) {
+    _helixAdmin = helixAdmin;
+    _clusterService = clusterService;
+    _configAccessor = configAccessor;
+    _dataAccessor = dataAccessor;
+    _assignmentAlgorithm = FifoVirtualGroupAssignmentAlgorithm.getInstance();
+  }
+
+  /**
+   * Add virtual topology group for a cluster.
+   * This includes calculating the virtual group assignment for all instances in the cluster then update instance config
+   * and cluster config. We override {@link ClusterConfig.ClusterConfigProperty#TOPOLOGY} and
+   * {@link ClusterConfig.ClusterConfigProperty#FAULT_ZONE_TYPE} for cluster config, and add new field to
+   * {@link InstanceConfig.InstanceConfigProperty#DOMAIN} that contains virtual topology group information.
+   * This is only supported for cloud environments. Cluster is expected to be in maintenance mode during config change.
+   * @param clusterName the cluster name.
+   * @param customFields custom fields, {@link VirtualTopologyGroupConstants#GROUP_NAME}
+   *                     and {@link VirtualTopologyGroupConstants#GROUP_NUMBER} are required,
+   *                     {@link VirtualTopologyGroupConstants#AUTO_MAINTENANCE_MODE_DISABLED} is optional.
+   *                     -- if set ture, the cluster will NOT automatically enter/exit maintenance mode during this API call;
+   *                     -- if set false or not set, the cluster will automatically enter maintenance mode and exit after
+   *                     the call succeeds. It won't proceed if the cluster is already in maintenance mode.
+   *                     Either case, the cluster must be in maintenance mode before config change.
+   */
+  public void addVirtualTopologyGroup(String clusterName, Map<String, String> customFields) {
+    // validation
+    CloudConfig cloudConfig = _configAccessor.getCloudConfig(clusterName);
+    if (cloudConfig == null || !cloudConfig.isCloudEnabled()) {
+      throw new HelixException(
+          "Cloud is not enabled, addVirtualTopologyGroup is not allowed to run in non-cloud environment.");
+    }
+    ClusterConfig clusterConfig = _configAccessor.getClusterConfig(clusterName);
+    Preconditions.checkState(clusterConfig.isTopologyAwareEnabled(),
+        "Topology-aware rebalance is not enabled in cluster " + clusterName);
+    String groupName = customFields.get(VirtualTopologyGroupConstants.GROUP_NAME);
+    String groupNumberStr = customFields.get(VirtualTopologyGroupConstants.GROUP_NUMBER);
+    Preconditions.checkArgument(!StringUtils.isEmpty(groupName), "virtualTopologyGroupName cannot be empty!");
+    Preconditions.checkArgument(!StringUtils.isEmpty(groupNumberStr), "virtualTopologyGroupNumber cannot be empty!");
+    int numGroups = 0;
+    try {
+      numGroups = Integer.parseInt(groupNumberStr);
+      Preconditions.checkArgument(numGroups > 0, "Number of virtual groups should be positive.");
+    } catch (NumberFormatException ex) {
+      throw new IllegalArgumentException("virtualTopologyGroupNumber " + groupNumberStr + " is not an integer.", ex);
+    }
+    LOG.info("Computing virtual topology group for cluster {} with param {}", clusterName, customFields);
+
+    // compute group assignment
+    ClusterTopology clusterTopology = _clusterService.getClusterTopology(clusterName);
+    Preconditions.checkArgument(numGroups <= clusterTopology.getAllInstances().size(),
+        "Number of virtual groups cannot be greater than the number of instances.");
+    Map<String, Set<String>> assignment =
+        _assignmentAlgorithm.computeAssignment(numGroups, groupName, clusterTopology.toZoneMapping());
+
+    boolean autoMaintenanceModeDisabled = Boolean.parseBoolean(
+        customFields.getOrDefault(VirtualTopologyGroupConstants.AUTO_MAINTENANCE_MODE_DISABLED, "false"));
+    // if auto mode is NOT disabled, let service enter maintenance mode and exit after the API succeeds.
+    if (!autoMaintenanceModeDisabled) {
+      Preconditions.checkState(!_helixAdmin.isInMaintenanceMode(clusterName),
+          "This operation is not allowed if cluster is already in maintenance mode before the API call. "
+              + "Please set autoMaintenanceModeDisabled=true if this is intended.");
+      _helixAdmin.manuallyEnableMaintenanceMode(clusterName, true,
+          "Enable maintenanceMode for virtual topology group change.", customFields);
+    }
+    Preconditions.checkState(_helixAdmin.isInMaintenanceMode(clusterName),
+        "Cluster is not in maintenance mode. This is required for virtual topology group setting. "
+            + "Please set autoMaintenanceModeDisabled=false (default) to let the cluster enter maintenance mode automatically, "
+            + "or use autoMaintenanceModeDisabled=true and control cluster maintenance mode in client side.");
+
+    updateConfigs(clusterName, clusterConfig, assignment);
+    if (!autoMaintenanceModeDisabled) {
+      _helixAdmin.manuallyEnableMaintenanceMode(clusterName, false,
+          "Disable maintenanceMode after virtual topology group change.", customFields);
+    }
+  }
+
+  private void updateConfigs(String clusterName, ClusterConfig clusterConfig, Map<String, Set<String>> assignment) {
+    List<String> zkPaths = new ArrayList<>();
+    List<DataUpdater<ZNRecord>> updaters = new ArrayList<>();
+    createInstanceConfigUpdater(clusterName, assignment).forEach((zkPath, updater) -> {
+      zkPaths.add(zkPath);
+      updaters.add(updater);
+    });
+    // update instance config
+    boolean[] results = _dataAccessor.updateChildren(zkPaths, updaters, AccessOption.EPHEMERAL);
+    for (int i = 0; i < results.length; i++) {
+      if (!results[i]) {
+        throw new HelixException("Failed to update instance config for path " + zkPaths.get(i));
+      }
+    }
+    // update cluster config
+    String virtualTopologyString = computeVirtualTopologyString(clusterConfig);
+    clusterConfig.setTopology(virtualTopologyString);
+    clusterConfig.setFaultZoneType(VirtualTopologyGroupConstants.VIRTUAL_FAULT_ZONE_TYPE);
+    _configAccessor.updateClusterConfig(clusterName, clusterConfig);
+    LOG.info("Successfully update instance and cluster config for {}", clusterName);
+  }
+
+  @VisibleForTesting
+  static String computeVirtualTopologyString(ClusterConfig clusterConfig) {
+    ClusterTopologyConfig clusterTopologyConfig = ClusterTopologyConfig.createFromClusterConfig(clusterConfig);
+    String endNodeType = clusterTopologyConfig.getEndNodeType();
+    String[] splits = new String[] {"", VirtualTopologyGroupConstants.VIRTUAL_FAULT_ZONE_TYPE, endNodeType};
+    return String.join(VirtualTopologyGroupConstants.PATH_NAME_SPLITTER, splits);
+  }
+
+  /**
+   * Create updater for instance config for async update.
+   * @param clusterName cluster name of the instances.
+   * @param assignment virtual group assignment.
+   * @return a map from instance zkPath to its {@link DataUpdater} to update.
+   */
+  @VisibleForTesting
+  static Map<String, DataUpdater<ZNRecord>> createInstanceConfigUpdater(
+      String clusterName, Map<String, Set<String>> assignment) {
+    Map<String, DataUpdater<ZNRecord>> updaters = new HashMap<>();
+    for (Map.Entry<String, Set<String>> entry : assignment.entrySet()) {
+      String virtualGroup = entry.getKey();
+      for (String instanceName : entry.getValue()) {
+        String path = PropertyPathBuilder.instanceConfig(clusterName, instanceName);
+        updaters.put(path, currentData -> {
+          InstanceConfig instanceConfig = new InstanceConfig(currentData);
+          Map<String, String> domainMap = instanceConfig.getDomainAsMap();
+          domainMap.put(VirtualTopologyGroupConstants.VIRTUAL_FAULT_ZONE_TYPE, virtualGroup);
+          instanceConfig.setDomain(domainMap);
+          return instanceConfig.getRecord();
+        });
+      }
+    }
+    return updaters;
+  }
+}
--- a/helix-rest/src/test/java/org/apache/helix/rest/server/service/TestVirtualTopologyGroupService.java
+++ b/helix-rest/src/test/java/org/apache/helix/rest/server/service/TestVirtualTopologyGroupService.java
@@ -1,0 +1,192 @@
+package org.apache.helix.rest.server.service;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.helix.ConfigAccessor;
+import org.apache.helix.HelixAdmin;
+import org.apache.helix.HelixDataAccessor;
+import org.apache.helix.HelixException;
+import org.apache.helix.cloud.azure.AzureConstants;
+import org.apache.helix.cloud.constants.CloudProvider;
+import org.apache.helix.model.CloudConfig;
+import org.apache.helix.model.ClusterConfig;
+import org.apache.helix.model.HelixConfigScope;
+import org.apache.helix.model.InstanceConfig;
+import org.apache.helix.model.builder.HelixConfigScopeBuilder;
+import org.apache.helix.rest.server.json.cluster.ClusterTopology;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.apache.helix.zookeeper.zkclient.DataUpdater;
+import org.testng.Assert;
+import org.testng.annotations.BeforeTest;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import static org.apache.helix.cloud.constants.VirtualTopologyGroupConstants.*;
+import static org.mockito.Mockito.*;
+
+
+public class TestVirtualTopologyGroupService {
+  private static final String TEST_CLUSTER = "Test_Cluster";
+  private static final String TEST_CLUSTER0 = "TestCluster_0";
+  private static final String TEST_CLUSTER1 = "TestCluster_1";
+
+  private final ConfigAccessor _configAccessor = mock(ConfigAccessor.class);
+  private final HelixDataAccessor _dataAccessor = mock(HelixDataAccessor.class);
+  private InstanceConfig _instanceConfig0;
+  private InstanceConfig _instanceConfig1;
+  private InstanceConfig _instanceConfig2;
+  private Map<String, DataUpdater<ZNRecord>> _updaterMap;
+  private HelixAdmin _helixAdmin;
+  private VirtualTopologyGroupService _service;
+
+  @BeforeTest
+  public void prepare() {
+    Map<String, Set<String>> assignment = new HashMap<>();
+    _instanceConfig0 = new InstanceConfig("instance_0");
+    _instanceConfig0.setDomain("helixZoneId=zone0");
+    _instanceConfig1 = new InstanceConfig("instance_1");
+    _instanceConfig1.setDomain("helixZoneId=zone0");
+    _instanceConfig2 = new InstanceConfig("instance_2");
+    _instanceConfig2.setDomain("helixZoneId=zone1");
+
+    assignment.put("virtual_group_0", ImmutableSet.of("instance_0", "instance_1"));
+    assignment.put("virtual_group_1", ImmutableSet.of("instance_2"));
+    _updaterMap = VirtualTopologyGroupService.createInstanceConfigUpdater(TEST_CLUSTER, assignment);
+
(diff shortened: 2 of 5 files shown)


helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,163 @@
...
+
+  private void updateConfigs(String clusterName, ClusterConfig clusterConfig, Map<String, Set<String>> assignment) {
+    List<String> zkPaths = new ArrayList<>();
+    List<DataUpdater<ZNRecord>> updaters = new ArrayList<>();
+    createInstanceConfigUpdater(clusterName, assignment).forEach((zkPath, updater) -> {
+      zkPaths.add(zkPath);
+      updaters.add(updater);
+    });
+    boolean[] results = _dataAccessor.updateChildren(zkPaths, updaters, AccessOption.EPHEMERAL);
+    for (int i = 0; i < results.length; i++) {
```
xyuanlu, line 124: nit: how about 
if ( results.stream().anyMatch(res-> !res) ) {
   throw new HelixException("Failed to update instance config for path " + zkPaths.get(i));
}

helix-core/src/main/java/org/apache/helix/common/VirtualTopologyGroupConstants.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import static org.apache.helix.cloud.azure.AzureConstants.*;
```
narendly, line 22: Note on style: generally we avoid static wildcard imports

helix-core/src/main/java/org/apache/helix/common/VirtualTopologyGroupConstants.java
```diff
@@ -0,0 +1,31 @@
...
+
+
+public class VirtualTopologyGroupConstants {
+  public static final String GROUP_NAME = "virtualTopologyGroupName";
+  public static final String GROUP_NUMBER = "virtualTopologyGroupNumber";
+  public static final String VIRTUAL_FAULT_ZONE_TYPE = "virtual_zone";
```
narendly, line 28: Is there a reason we're not being consistent with casing? Two are camelCase and one is underscored?

helix-core/src/main/java/org/apache/helix/controller/VirtualTopologyGroupStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.helix.util.HelixUtil;
+
+public enum VirtualTopologyGroupStrategy {
+
+  /**
+   * A default assignment strategy that is deterministic and stable whereas:
```
narendly, line 33: nit: "whereas" -> "where"

helix-core/src/main/java/org/apache/helix/controller/VirtualTopologyGroupStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+
+import com.google.common.collect.ImmutableMap;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.helix.util.HelixUtil;
+
+public enum VirtualTopologyGroupStrategy {
```
narendly, line 30: Usually Java Enums are used to define values where Java Interfaces are used to define behavior (or the default behavior and make it extendible). Is there a reason we're making this an Enum here?

What I would suggest is to create a separate package (I'm not sure this belongs to `org.apache.helix.controller` - maybe something like `*.cloud.groupingStrategy` or `*.topology.groupingStrategy` might be more appropriate?) and define an interface with a default implementation?

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
qqu0127: The reason I chose the current way is to print the failed instance address, it's the first failed one though. Let me update it in the next revision. Thanks.

helix-core/src/main/java/org/apache/helix/common/VirtualTopologyGroupConstants.java
qqu0127: Hmm, good catch. I directly followed the design, let me change it to all camelCase.

helix-core/src/main/java/org/apache/helix/controller/VirtualTopologyGroupStrategy.java
qqu0127: Thanks for the comment, it's really good question! I hesitated a lot in choosing enum over interface, here are my considerations:
1. Enum defines constants that could go with predefined behavior (enum methods), and strategy falls in the category. Plus enum is naturally static final singleton, and it's compile-time constant, method is also resolved at compile time, while interface is not.
2. Afaik, we don't have dependency injection to decouple interface and impl, instead we directly construct impl class. I would try to avoid creating interface if there is only one known impl.
3. The key method is static. With interface we could use a static or default method, but it's weird to instantiate a class for static method. 
4. Writing logic in a util class also works, while enum provides a layer of encapsulation and is type safe.

This is open discussion, using interface here won't be too bad after all. Let me know how you feel about it. Thanks again.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
qqu0127: To update, I just realized we can't stream on `boolean[]`, it has to be `Boolean[]` It seems too much effort and potential performance drop to make this work. If you don't have strong opinion on this, I'll stick with the original. Thanks!

xyuanlu: Thanks for the info. Let's keep it this way then.

helix-core/src/main/java/org/apache/helix/controller/VirtualTopologyGroupStrategy.java
narendly: Could `VirtualTopologyGroupStrategy` have multiple implementations and do we expect to add different implementations and make this configurable? If so, I think using an abstract base class could also work.

Or is the goal to provide a static method? if so i think a util class would fit better here. Since this is only used in `VirtualTopologyGroupService` and only fixed to a `DEFAULT` implementation. Is this class only going to be used in Helix REST? If that's the case, I'd place this in that module.

I'm confused because I'm not sure what the intent of this class is - it also has an abstract method and the default enum is a behavior (verb) than named variables (nouns) that enums generally are.

qqu0127: The Enum `VirtualTopologyGroupStrategy` may have other enum values, e.g. `STRATEGY1`, `STRATEGY2` (in parallel to `DEFAULT` , they can have their own method implementation. And it can be configurable which one to use by parsing string.
To clarify, the reason I created this class is to have a way to "implement" static method as in interface. So each enu value defines a strategy, while each strategy comes with an assignment algorithm (abstract method). Indeed, a util class fits here, but I chose to use Enum for better encapsulation. FYI, https://blogs.oracle.com/javamagazine/post/how-to-make-the-most-of-java-enums. 
That said, I'm afraid this is creating more confusion than I expected, I'll change it to a util class if you don't have objection. We can revisit this when we want to add other implementations. Thanks.

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualTopologyGroupStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+   * @param numGroups number of the virtual groups
+   * @param virtualGroupName virtual group name
+   * @param zoneMapping current zone mapping from zoneId to instanceIds
+   * @return the assignment as mapping from virtual group ID to instanceIds
+   */
+  public abstract Map<String, Set<String>> computeAssignment(
+      int numGroups, String virtualGroupName, Map<String, Set<String>> zoneMapping);
+
+  private static String computeVirtualGroupId(int groupIndex, String virtualGroupName) {
+    return virtualGroupName + "_" + groupIndex;
```
jiajunwang, line 78: nit, avoid hard code?

qqu0127: What's your suggestion for constructing the virtual groupId? I can go with a string template for naming, but don't think that's much difference than this.

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualTopologyGroupStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+   * 1. assignment is guaranteed consistent for same inputs.
+   * 2. number of instance shuffles is reduced.
+   */
+  DEFAULT {
+    @Override
+    public Map<String, Set<String>> computeAssignment(
+        int numGroups, String virtualGroupName, Map<String, Set<String>> zoneMapping) {
+      List<String> sortedInstances = HelixUtil.sortAndFlattenZoneMapping(zoneMapping);
+      Map<String, Set<String>> assignment = new HashMap<>();
+      int instancesPerGroup = sortedInstances.size() / numGroups;
```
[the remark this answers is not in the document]
qqu0127: How would that work? If we have 24 instances over 5 HDF, to assign 7 virtual groups. Notice `ceil(24 / 7) == 4` and `24 / 4 == 6`, 24 instances are effectively only assigned to 6 groups.

helix-core/src/main/java/org/apache/helix/common/VirtualTopologyGroupConstants.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import org.apache.helix.cloud.azure.AzureConstants;
+
+
+public class VirtualTopologyGroupConstants {
+  public static final String GROUP_NAME = "virtualTopologyGroupName";
+  public static final String GROUP_NUMBER = "virtualTopologyGroupNumber";
+  public static final String VIRTUAL_FAULT_ZONE_TYPE = "virtualZone";
+  public static final String VIRTUAL_TOPOLOGY = new StringBuilder(AzureConstants.AZURE_TOPOLOGY)
```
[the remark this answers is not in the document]
qqu0127: It's only used in Azure, but not always, and never elsewhere.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,163 @@
...
+  private final HelixDataAccessor _dataAccessor;
+  private final VirtualTopologyGroupStrategy _strategy;
+
+  public VirtualTopologyGroupService(HelixAdmin helixAdmin, ClusterService clusterService,
+      ConfigAccessor configAccessor, HelixDataAccessor dataAccessor) {
+    _helixAdmin = helixAdmin;
+    _clusterService = clusterService;
+    _configAccessor = configAccessor;
+    _dataAccessor = dataAccessor;
+    _strategy = VirtualTopologyGroupStrategy.DEFAULT;
```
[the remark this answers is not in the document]
qqu0127: It's possible to have multiple strategies in the future and to be configurable, but right now there is only one. There is also discussion with Hunter on this, see https://github.com/apache/helix/pull/1935#discussion_r786372019

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,163 @@
...
+        "virtualTopologyGroupNumber cannot be empty!");
+    Preconditions.checkState(!StringUtils.isEmpty(groupName), "virtualTopologyGroupName cannot be empty!");
+    int numGroups = Integer.parseInt(groupNumberStr);
+
+    // compute group assignment
+    ClusterTopology clusterTopology = _clusterService.getClusterTopology(clusterName);
+    Map<String, Set<String>> assignment =
+        _strategy.computeAssignment(numGroups, groupName, clusterTopology.toZoneMapping());
+
+    if (enterMaintenanceMode) {
```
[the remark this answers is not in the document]
qqu0127: >  then your disabling maintenance mode in line 111 will remove user's previous setup

How so? If `enterMaintenanceMode` is false, we rely on user setup and won't change the maintenance mode.
Are you saying for the case when `enterMaintenanceMode` is true, we shouldn't proceed if maintenance mode is already on?

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -52,6 +52,7 @@
     FAULT_ZONE_TYPE, // the type in which isolation should be applied on when Helix places the
     // replicas from same partition.
     TOPOLOGY_AWARE_ENABLED, // whether topology aware rebalance is enabled.
+    VIRTUAL_GROUP_ENABLED, // whether virtual topology group is enabled
```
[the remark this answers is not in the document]
qqu0127: The idea of setting this is to avoid unexpected API call to a wrong cluster. We'd like to have this cluster config to be safer.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,163 @@
...
+   * and cluster config. Cluster is expected to enter maintenanceMode during config update, this is either enabled/disabled
+   * in place this method or handled by client side code.
+   * @param clusterName the cluster name.
+   * @param customFields custom fields, {@link VirtualTopologyGroupConstants#GROUP_NAME}
+   *                     and {@link VirtualTopologyGroupConstants#GROUP_NUMBER} are required
+   * @param enterMaintenanceMode if enabled, the cluster will enter maintenance mode during the setup and exit once it
+   *                             completes. Otherwise, it's expected the maintenanceMode is controlled by client side.
+   */
+  public void addVirtualTopologyGroup(String clusterName, Map<String, String> customFields, boolean enterMaintenanceMode) {
+    // only support if CLOUD_ENABLED AND VIRTUAL_GROUP_ENABLED
```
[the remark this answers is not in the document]
qqu0127: Same as above.
I don't have strong opinion on this though, we can discuss more during the meeting.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,163 @@
...
+    CloudConfig cloudConfig = _configAccessor.getCloudConfig(clusterName);
+    if (cloudConfig == null || !cloudConfig.isCloudEnabled()) {
+      throw new HelixException(
+          "Cloud is not enabled, addVirtualTopologyGroup is not allowed to run in non-cloud environment.");
```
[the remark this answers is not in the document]
qqu0127: In my understanding this feature is specifically designed for Azure, where hard fault zone is out of our control. For on-prem, we have a way to adjust FZ count. @junkaixue may have more information on why this is designed for only azure?

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualTopologyGroupStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+
+import com.google.common.collect.ImmutableMap;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.helix.util.HelixUtil;
+
+public enum VirtualTopologyGroupStrategy {
```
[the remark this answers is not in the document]
qqu0127: I agree, if there is other strategy out there might be confusing. Do you have a naming proposal? Algorithm?
"strategy" still sounds fit the most to me in this use case...

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,163 @@
...
+    if (!clusterConfig.isVirtualGroupEnabled()) {
+      throw new HelixException("Virtual Group is disabled in cluster " + clusterName);
+    }
+    // validation
+    String groupName = customFields.get(VirtualTopologyGroupConstants.GROUP_NAME);
+    String groupNumberStr = Preconditions.checkNotNull(
+        customFields.get(VirtualTopologyGroupConstants.GROUP_NUMBER),
+        "virtualTopologyGroupNumber cannot be empty!");
+    Preconditions.checkState(!StringUtils.isEmpty(groupName), "virtualTopologyGroupName cannot be empty!");
+    int numGroups = Integer.parseInt(groupNumberStr);
```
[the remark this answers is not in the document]
qqu0127: We have null check on `groupNumberStr` already. 
Will simply verify it's a positive integer, I think it's fine to just let it fail at parsing (handling that `NumberFormatException` seems redundant)

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualTopologyGroupStrategy.java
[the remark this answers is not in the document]
jiajunwang: Not exactly, the reason you get the conclusion of 6 groups is that you sort the instance and compute groupIndex with "instanceInd / instancesPerGroup". My original idea is that we can directly use mod to get the index.
However, according to what we have discussed yesterday, it might be problematic to assign instances of one physical zone to multiple virtual groups. It may effectively downgrade fault tolerance. So I assume you will have to change the design here. Let's discuss further when your new design is ready : )

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualTopologyGroupStrategy.java
jiajunwang: I can see you did 

> private static final String DOMAIN_FIELD_SPLITTER = ",";
> private static final String DOMAIN_VALUE_JOINER = "=";

for the InstanceConfig. Then why hardcode here? It's better to keep the same style.

helix-core/src/main/java/org/apache/helix/common/VirtualTopologyGroupConstants.java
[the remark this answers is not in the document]
jiajunwang: I don't believe the high-level design of VirtualTopologyGroup is for Azure only. It should be determined by the caller when they call the Rest API.
As we discussed yesterday, whenever a cluster is "cloud-enabled", it is supporting virtual topology. And for Helix, "cloud-enabled" does not mean a specific cloud platform or even public cloud. As long as the cluster nodes can provide the necessary information, it is supporting necessary cloud feature from Helix perspective.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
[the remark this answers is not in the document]
jiajunwang: As we discussed offline yesterday, it is a must as long as we still want to keep the physical topology somewhere and may want to provide a way to roll back.

But logic-wise, it introduced more dependencies. For example, what if I enable virtual group without enabling topology awareness, does it even makes sense? What about when cloud features are not enabled? I can foresee many questions from our users.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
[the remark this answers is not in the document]
jiajunwang: As we discussed offline, the cloud here does not necessarily mean a real cloud platform. It solely means that Helix gets topology information from the participant under this scenario. Meaning the topology information is automatically provided instead of manually populated. So this restriction makes sense here.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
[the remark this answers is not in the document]
jiajunwang: I was thinking about negative numbers or zero. But if the group number is larger than the node count, can we still assign it?
It's also OK if you let the compute class throw exceptions. But please ensure the exception is understandable for our customers. IndexOutOfBoundsException is not acceptable.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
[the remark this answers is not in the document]
jiajunwang: > we shouldn't proceed if maintenance mode is already on?

Yes, or you don't exit maintenance mode at the end.

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualTopologyGroupStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * A default assignment strategy that is deterministic and stable where:
+   * 1. assignment is guaranteed consistent for same inputs.
+   * 2. number of instance shuffles is reduced.
```
zhangmeng916, line 35: What do we compare and conclude the "instance shuffles is reduced"? Could you please illustrate? Is this the best strategy for deterministic choices?

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,163 @@
...
+   */
+  public void addVirtualTopologyGroup(String clusterName, Map<String, String> customFields, boolean enterMaintenanceMode) {
+    // only support if CLOUD_ENABLED AND VIRTUAL_GROUP_ENABLED
+    CloudConfig cloudConfig = _configAccessor.getCloudConfig(clusterName);
+    if (cloudConfig == null || !cloudConfig.isCloudEnabled()) {
+      throw new HelixException(
+          "Cloud is not enabled, addVirtualTopologyGroup is not allowed to run in non-cloud environment.");
+    }
+    ClusterConfig clusterConfig = _configAccessor.getClusterConfig(clusterName);
+    if (!clusterConfig.isVirtualGroupEnabled()) {
```
zhangmeng916, line 85: Actually I don't see how adding this config could improve safety. As long as cloud config is on, we allow virtual grouping, at least this is the agreement for now. Whether the virtual grouping is used correctly is purely decided by users' knowledge. Asking users to turn on another config does not really help ensure the correctness.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,163 @@
...
+    List<String> zkPaths = new ArrayList<>();
+    List<DataUpdater<ZNRecord>> updaters = new ArrayList<>();
+    createInstanceConfigUpdater(clusterName, assignment).forEach((zkPath, updater) -> {
+      zkPaths.add(zkPath);
+      updaters.add(updater);
+    });
+    boolean[] results = _dataAccessor.updateChildren(zkPaths, updaters, AccessOption.EPHEMERAL);
+    for (int i = 0; i < results.length; i++) {
+      if (!results[i]) {
+        throw new HelixException("Failed to update instance config for path " + zkPaths.get(i));
```
zhangmeng916, line 126: So this fails, we have configs partially updated. Will this impact controller's pipeline logic before we update them again? Please provide some insights.

helix-core/src/main/java/org/apache/helix/common/VirtualTopologyGroupConstants.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import org.apache.helix.cloud.azure.AzureConstants;
```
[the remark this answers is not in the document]
zhangmeng916: +1. Why virtual topology should be bound with Azure? It should not have anything dependent on a specific type of cloud environment. Before using virtual grouping, you only need cloud config is on, but you don't need the topology to be the Azure one.

helix-rest/src/test/java/org/apache/helix/rest/server/service/TestVirtualTopologyGroupService.java
```diff
@@ -0,0 +1,172 @@
...
+  public void testClusterVirtualGroupConfigSetup() {
+    ClusterConfig clusterConfig2 = new ClusterConfig(TEST_CLUSTER2);
+    clusterConfig2.setVirtualGroupEnabled(false);
+    when(_configAccessor.getClusterConfig(TEST_CLUSTER2)).thenReturn(clusterConfig2);
+    _service.addVirtualTopologyGroup(
+        TEST_CLUSTER2, ImmutableMap.of(GROUP_NAME, "test-group", GROUP_NUMBER, "2"), true);
+  }
+
+  @Test
+  public void testService() {
```
zhangmeng916, line 125: Please have more concrete naming for tests. This is difficult to tell what is being tested, especially if it's in the log.

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualTopologyGroupStrategy.java
qqu0127: Thanks for the comment, the assignment algorithm will be revised and validated with more clear metrics and benchmarks. There will be a design review as well.

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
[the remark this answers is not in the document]
qqu0127: As discussed offline, there is a design change and we are discarding `VIRTUAL_GROUP_ENABLED`. Rollback is not a concern because there is always a fixed topology config on cloud environment.

helix-core/src/main/java/org/apache/helix/common/VirtualTopologyGroupConstants.java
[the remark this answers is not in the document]
qqu0127: Updated, this is not for a specific cloud platform, but should be universal.

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualTopologyGroupStrategy.java
qqu0127: Discussed offline, we implemented a different algorithm. Resolved this in separate PR -- https://github.com/apache/helix/pull/1948

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
qqu0127: Updated, we are discarding this config.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
qqu0127: We ensure the cluster is in maintenance mode during config change. So if it fails, we expect human intervention, meanwhile the cluster stays in maintenance mode to prevent any partition movement.

helix-rest/src/test/java/org/apache/helix/rest/server/service/TestVirtualTopologyGroupService.java
```diff
@@ -0,0 +1,183 @@
...
+  @Test
+  public void testVirtualTopologyGroupService() {
+    _service.addVirtualTopologyGroup(TEST_CLUSTER0, ImmutableMap.of(
+        GROUP_NAME, "test-group", GROUP_NUMBER, "2", ENTER_MAINTENANCE_MODE, "false"));
+    verify(_dataAccessor, times(1)).updateChildren(anyList(), anyList(), anyInt());
+    verify(_configAccessor, times(1)).updateClusterConfig(anyString(), any());
+  }
+
+  @Test(expectedExceptions = IllegalStateException.class,
+      expectedExceptionsMessageRegExp = "This operation is not allowed while cluster is in maintenance mode.*")
```
zhangmeng916, line 122: Is the exception correct? Should be "the operation is not allowed while cluster is not in maintenance mode"?

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,183 @@
...
+
+    // compute group assignment
+    ClusterTopology clusterTopology = _clusterService.getClusterTopology(clusterName);
+    Preconditions.checkArgument(numGroups <= clusterTopology.getAllInstances().size(),
+        "Number of virtual groups cannot be greater than the number of instances.");
+    Map<String, Set<String>> assignment =
+        _assignmentAlgorithm.computeAssignment(numGroups, groupName, clusterTopology.toZoneMapping());
+
+    boolean enterMaintenanceMode = Boolean.parseBoolean(
+        customFields.getOrDefault(VirtualTopologyGroupConstants.ENTER_MAINTENANCE_MODE, "true"));
```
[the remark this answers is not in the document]
qqu0127: No, we ensure the cluster is in maintenance mode before config change in either case.
When this is enabled, the service will enter and exit maintenance mode during the API call for safe operation. 
We also expect users may want more control and don’t want the service enter/exit maintenance mode automatically. So when it’s disabled, the service won’t change the maintenance mode, but will still check it before config change.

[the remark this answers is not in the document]
zhangmeng916: In this case, this config should be named as "AUTO_ENTER_MAINTENANCE_MODE". But it is still a bit weird that this config has a default value as "true". Normally if a boolean config is not set, the default value is false. You can think about whether you want to change it to "DISABLE_AUTO_ENTER_MAINTENANCE_MODE". Users need to manually set it to true if they would like to control it themselves.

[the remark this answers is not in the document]
qqu0127: Good point, changed to its negation with `AUTO_MAINTENANCE_MODE_DISABLED`. Thanks

helix-rest/src/test/java/org/apache/helix/rest/server/service/TestVirtualTopologyGroupService.java
qqu0127: Same as above comment, it is expected. The cluster shouldn't be already in maintenance mode if in auto mode. Updated the config name and exception message to make it more clear.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,194 @@
...
+    return String.join(VirtualTopologyGroupConstants.PATH_NAME_SPLITTER, splits);
+  }
+
+  /**
+   * Create updater for instance config for async update.
+   * @param clusterName cluster name of the instances.
+   * @param assignment virtual group assignment.
+   * @return a map from instance zkPath to its {@link DataUpdater} to update.
+   */
+  @VisibleForTesting
```
zhangmeng916, line 175: This annotation means the visibility of a type or member has been relaxed to make the code testable. Do you have such relaxation?

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
```diff
@@ -0,0 +1,194 @@
...
+import org.apache.helix.model.InstanceConfig;
+import org.apache.helix.rest.server.json.cluster.ClusterTopology;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.apache.helix.zookeeper.zkclient.DataUpdater;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Service for virtual topology group.
```
zhangmeng916, line 51: This doc does not convey anything other than the class name. If you do not have anywhere else document what is virtual group, you can consider put it here, concisely.

zhangmeng916 (approved):
I suggest you to keep commit history in the future, so that people can know all the context, especially when they read the comments.

helix-rest/src/main/java/org/apache/helix/rest/server/service/VirtualTopologyGroupService.java
qqu0127: Sure, updated.

qqu0127:
This PR is ready to merge, approved by @zhangmeng916 

Final commit message:

Implement java API and utils for virtual topology group
We introduce the java API for computing virtual topology group for cloud environments.

PR #1945, opened 2022-01-26
Refactor and move ClusterTopologyConfig from nested class to a separate one.

qqu0127:
### Issues

https://github.com/apache/helix/issues/1946

### Description

A refactoring for virtual group project.
We found a need to reference class ClusterTopologyConfig outside of the scope of controller or rebalancer. Thus we refactor it out to a separate class and do some cleanup. This should be backward compatible.

We move ClusterTopologyConfig from nested to a standalone class in helix/model and to be used by virtual topology group logic. Some code refactor and cleanup are included.

### Tests

Add new UT class TestClusterTopologyConfig

Test helix-core:

[INFO] Tests run: 1287, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,462.237 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1287, Failures: 0, Errors: 0, Skipped: 0

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
@@ -30,6 +30,7 @@ import java.util.Set;
 
 import org.apache.helix.HelixException;
 import org.apache.helix.model.ClusterConfig;
+import org.apache.helix.model.ClusterTopologyConfig;
 import org.apache.helix.model.InstanceConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -53,15 +54,7 @@ public class Topology {
   private final List<String> _allInstances;
   private final List<String> _liveInstances;
   private final Map<String, InstanceConfig> _instanceConfigMap;
-  private final ClusterConfig _clusterConfig;
-  private static final String DEFAULT_DOMAIN_PREFIX = "Helix_default_";
-
-  static class ClusterTopologyConfig {
-    String endNodeType;
-    String faultZoneType;
-    LinkedHashMap<String, String> topologyKeyDefaultValue = new LinkedHashMap<>();
-  }
-  private ClusterTopologyConfig _clusterTopologyConfig;
+  private final ClusterTopologyConfig _clusterTopologyConfig;
 
   public Topology(final List<String> allNodes, final List<String> liveNodes,
       final Map<String, InstanceConfig> instanceConfigMap, ClusterConfig clusterConfig) {
@@ -78,18 +71,16 @@ public class Topology {
       throw new HelixException(String.format("Config for instances %s is not found!",
           _allInstances.removeAll(_instanceConfigMap.keySet())));
     }
-    _clusterConfig = clusterConfig;
-    _clusterTopologyConfig = getClusterTopologySetting(clusterConfig);
-
-    _root = createClusterTree();
+    _clusterTopologyConfig = ClusterTopologyConfig.createFromClusterConfig(clusterConfig);
+    _root = createClusterTree(clusterConfig);
   }
 
   public String getEndNodeType() {
-    return _clusterTopologyConfig.endNodeType;
+    return _clusterTopologyConfig.getEndNodeType();
   }
 
   public String getFaultZoneType() {
-    return _clusterTopologyConfig.faultZoneType;
+    return _clusterTopologyConfig.getFaultZoneType();
   }
 
   public Node getRootNode() {
@@ -158,7 +149,7 @@ public class Topology {
     return newRoot;
   }
 
-  private Node createClusterTree() {
+  private Node createClusterTree(ClusterConfig clusterConfig) {
     // root
     Node root = new Node();
     root.setName("root");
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/model/ClusterTopologyConfig.java
```diff
@@ -0,0 +1,92 @@
...
+    _topologyKeyDefaultValue = topologyKeyDefaultValue;
+  }
+
+  /**
+   * Populate faultZone, endNodetype and and a LinkedHashMap containing pathKeys default values for
+   * clusterConfig.Topology. The LinkedHashMap will be empty if clusterConfig.Topology is unset.
+   *
+   * @return an instance of {@link ClusterTopologyConfig}
+   */
+  static ClusterTopologyConfig fromClusterConfig(ClusterConfig clusterConfig) {
```
zhangmeng916, line 49: The naming of the function does not comply with the standard. Suggest to rename as "createFromClusterConfig" or similar.

helix-core/src/main/java/org/apache/helix/model/ClusterTopologyConfig.java
```diff
@@ -0,0 +1,92 @@
...
+
+    String topologyDef = clusterConfig.getTopology();
+    if (topologyDef != null) {
+      for (String topologyKey : topologyDef.trim().split(TOPOLOGY_SPLITTER)) {
+        if (!topologyKey.isEmpty()) {
+          topologyKeyDefaultValue.put(topologyKey, DEFAULT_DOMAIN_PREFIX + topologyKey);
+          endNodeType = topologyKey;
+        }
+      }
+      Preconditions.checkState(!topologyKeyDefaultValue.isEmpty(),
```
zhangmeng916, line 67: This check will throws an illegal statement exception, while the original one throws illegal argument exception. Is this what you planned to change? I feel illegal argument is more proper here.

helix-core/src/test/java/org/apache/helix/model/TestClusterTopologyConfig.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class TestClusterTopologyConfig {
+
+  @Test
+  public void testClusterNonTopologyAware() {
+    ClusterConfig testConfig = new ClusterConfig("testId");
+    testConfig.setTopologyAwareEnabled(false);
+    ClusterTopologyConfig clusterTopologyConfig = ClusterTopologyConfig.fromClusterConfig(testConfig);
+    Assert.assertEquals(clusterTopologyConfig.getEndNodeType(), Topology.Types.INSTANCE.name());
+    Assert.assertEquals(clusterTopologyConfig.getFaultZoneType(), Topology.Types.INSTANCE.name());
```
zhangmeng916, line 37: Is this correct? Why the fault zone type is INSTANCE name? shall we change the default value to be meaningful?

zhangmeng916 (commented):
Please make sure you have the correct issue number linked in your PR description.

helix-core/src/main/java/org/apache/helix/model/ClusterTopologyConfig.java
qqu0127: Thanks, will update

helix-core/src/main/java/org/apache/helix/model/ClusterTopologyConfig.java
qqu0127: Good catch, let me revert to IllegalArgumentException

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
```diff
@@ -1033,6 +1033,14 @@ public void setAbnormalStateResolverMap(Map<String, String> resolverMap) {
     return idealStateRuleMap;
   }
 
+  /**
+   * Create an instance of {@link ClusterTopologyConfig} based on cluster config.
+   * @return an instance of {@link ClusterTopologyConfig}.
+   */
+  public ClusterTopologyConfig getClusterTopologyConfig() {
+    return ClusterTopologyConfig.fromClusterConfig(this);
```
NealSun96, line 1041: Seems a bit odd to me... Is there a reason why ClusterConfig has a method to construct a ClusterTopologyConfig out of itself? If ClusterTopologyConfig needs to be constructed, the caller should just directly call `ClusterTopologyConfig.fromClusterConfig(clusterConfig)`, instead of `clusterConfig.getClusterTopologyConfig()`; this method doesn't seem that necessary to me.

NealSun96 (changes requested):

helix-core/src/test/java/org/apache/helix/model/TestClusterTopologyConfig.java
qqu0127: This is what I read https://github.com/apache/helix/pull/1945/files#diff-3bfc7aaa9b4578cd05434a4b4564dd61aa561f340ba7c08d1395d3f50a881fc2L238-L243, so I just keep the behavior. 
Since I'm not really sure on the impact of changing this default value, I prefer keeping it the same. 
@xyuanlu any input? I saw you introduced [this](https://github.com/apache/helix/commit/0869e729530d83ba55228b2d44a79ae44175c081).

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
qqu0127: Good question, indeed it's not the most beautiful way... I choose this approach to avoid over expose this class, right now `ClusterTopologyConfig.fromClusterConfig` is package private.
In my view, this class is like a view of `ClusterConfig`, a container associated with ClusterConfig, and how we construct this is impl details and shouldn't be exposed outside package. I didn't write the logic directly in ClusterConfig because this class is already too long.
This is open to discussion, let me know how you feel about it. Thanks!

NealSun96: I believe ClusterTopologyConfig is moved out of the class to make it independent. It makes more sense to me that ClusterTopologyConfig class is called to create itself.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/topology/Topology.java
```diff
@@ -54,14 +55,7 @@
   private final List<String> _liveInstances;
   private final Map<String, InstanceConfig> _instanceConfigMap;
   private final ClusterConfig _clusterConfig;
```
jiajunwang, line 57: Can put all the related info to ClusterTopologyConfig and remove the reference of ClusterConfig here?

qqu0127: Good point, updated. thx

helix-core/src/main/java/org/apache/helix/model/ClusterConfig.java
qqu0127: Fair enough, updated. Thanks.

NealSun96:
@qqu0127 Looks good to me except the issue @zhangmeng916 raised. Please also check the boxes in the PR description before final checkin (btw, I think the first paragraph you have should be merged into the Description section).

helix-core/src/test/java/org/apache/helix/model/TestClusterTopologyConfig.java
NealSun96: @qqu0127 I agree that @xyuanlu should take a look. Could anyone of you check if `testCreateClusterTopologyWithDefaultTopology` is correctly testing this case, by the way? On the first glance that should have caught this case? Thanks.

qqu0127: TestTopology.testCreateClusterTopologyWithDefaultTopology PASS with the patch.

xyuanlu: The default fault zone type and end node type is INSTANCE when topologyAware is not enabled.  When it is enabled, the default fault zone type is ZONE and end node type is INSTANCE. 
This behavior is added in 2017 by change https://github.com/apache/helix/commit/cb024650c25c6ddee7348d37fb67fa1a2e93bff1#

IMHO, the current behavior make sense. When no topology enabled, the structure should be flat, meaning zone and end node should be the same.

NealSun96 (approved):
LGTM

qqu0127:
Ready to merge, approved by @NealSun96 
Final commit message:
Fix apache#1946 Refactor and move ClusterTopologyConfig

Issue #1947, opened 2022-01-27
Virtual topology group -- implement grouping assignment logic


qqu0127:
**Is your feature request related to a problem? Please describe.**
As part of virtual topology group support, we'd want to implement the virtual group mapping logic maps fixed physical zones to a greater number of "virtual" zones for each instance in the cluster. This creates a layer of virtualization to be used on cloud environment.

**Describe the solution you'd like**
The logic and algorithm to assign virtual mapping based on cluster topology and number of virtual groups.

**Additional context**

PR #1948, opened 2022-01-27
Introduce VirtualTopologyGroup and its assignment logic.

qqu0127:
### Issues

Resolve https://github.com/apache/helix/issues/1947

### Description

As part of Virtual topology group project, we introduce its assignment logic and test in this PR.

This PR introduces a virtual grouping scheme with unit test and benchmark. This is for computing a virtualization layer on top of physical zones/fault domains on cloud environment. 
The assignment is stateless and deterministic. 
The following PR will include integration, config change and service logic.

### Tests

new unit test TestVirtualTopologyGroupAssignment

There is no behavior change for existing components.

### Documentation (Optional)
The wiki will be added during the final merge to master branch.

### Commits

### Code Quality


NealSun96:
@qqu0127 I would like a clarification please: the final choice of assignment logic is one of the three choices, is that correct? If so, the benchmarking should be done in offline code and only the best one should be selected and checked in. Let me know if I have a misunderstanding.

qqu0127:
> @qqu0127 I would like a clarification please: the final choice of assignment logic is one of the three choices, is that correct? If so, the benchmarking should be done in offline code and only the best one should be selected and checked in. Let me know if I have a misunderstanding.

@NealSun96 
Eventually, yes, I expect only one is used for non-test. I open this and include all choices in order to share the code and as a reference to help better understand the design. 
We have two options here, one is to merge this PR as it is with all 3, and cleanup in a following PR. This is targeting a project branch after all. 
Second option is to update this PR after review with a new commit. I prefer the first one though.

NealSun96:
@qqu0127 
Since it's a feature branch, I will weigh your opinions more than my own. Although it seems odd to me that experiment code is checked in, I suppose the clean up can be done during the final merge. 
Note that since this PR is not targeting master, github CI wouldn't kick off an automatic build & test process. If you wish to verify your changes, you need to do so locally.

qqu0127:
Cleanup the unused code, this is ready for review. Thanks.

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualGroupAssignmentAlgorithm.java
```diff
@@ -0,0 +1,38 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.Map;
+import java.util.Set;
+
+
+public interface VirtualGroupAssignmentAlgorithm {
```
zhangmeng916, line 26: I suggest to change the class name to be more self explanatory. Currently it's not easy to see the relation of `VirtualTopologyGroupScheme` and `VirtualGroupAssignmentAlgorithm`.

helix-core/src/test/java/org/apache/helix/cloud/virtualTopologyGroup/TestVirtualTopologyGroupAssignment.java
```diff
@@ -0,0 +1,160 @@
...
+      Map<String, Set<String>> baseMapping,
+      Map<String, Set<String>> testMapping,
+      int numGroups,
+      VirtualGroupAssignmentAlgorithm algorithm) {
+    Map<String, Set<String>> baseAssignment = algorithm.computeAssignment(numGroups, GROUP_NAME, baseMapping);
+    Map<String, Set<String>> testAssignment = algorithm.computeAssignment(numGroups, GROUP_NAME, testMapping);
+    AssignmentEvaluation base = new AssignmentEvaluation(baseAssignment, baseMapping);
+    AssignmentEvaluation test = new AssignmentEvaluation(testAssignment, testMapping);
+    System.out.println("Diff for " + algorithm + " : " + base.compareAssignments(test));
+  }
```
zhangmeng916, line 107: Why do we need to include this comparison in our test suite? Since your algorithm is deterministic, the result won't change no matter what. And I don't see the point of keeping running it for each commit. What is the value of knowing the diff between base and test case? The test does not really validate/assert anything.

helix-core/src/test/java/org/apache/helix/cloud/virtualTopologyGroup/AssignmentEvaluation.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  public void print() {
+    for (Map.Entry<String, Set<String>> entry : _coveredZonesByVirtualGroup.entrySet()) {
+      System.out.println("virtual group: " + entry.getKey() + ", " + "covered physical zones: " + entry.getValue());
+    }
+    for (Map.Entry<String, Set<String>> entry : _affectedVirtualGroupByZone.entrySet()) {
+      System.out.println("physical zone " + entry.getKey() + " impacts " + entry.getValue());
+    }
+    System.out.println(generateStats().toString());
```
zhangmeng916, line 67: Are all these prints useful? If it won't help for any debugging purpose, but only gets used in one time testing. I suggest to remove them.

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualTopologyGroupScheme.java
```diff
@@ -0,0 +1,73 @@
...
+public enum VirtualTopologyGroupScheme implements VirtualGroupAssignmentAlgorithm {
+
+  /**
+   * A strategy that densely assign virtual groups with input instance list, it doesn't move to the next one until
+   * the current one is filled.
+   * Given that instances.size = instancesPerGroup * numGroups + residuals,
+   * we break [residuals] into the first few groups, as a result each virtual group will have
+   * either [instancesPerGroup] or [instancesPerGroup + 1] instances.
+   */
+  FIFO {
```
zhangmeng916, line 42: I'm a bit confused about the code structure here. You have an enum with one value only, do you plan to add more in the future? Then will the function `computeVirtualGroupId` work for all future scheme? I think you should have one class for one type of assigning scheme, which implements the interface.

desaikomal:
This is my first review, so please correct me if i am wrong.
1. Since this is a new feature, can we update the description with link to feature document
2. Can you fill out API/backward compatibility related sections too?

zhangmeng916 (commented):
Overall, I think there is unnecessary code included in this commit. I suggest you revisit and see which can be removed. More specifically, those code that is used for evaluation/one time testing does not need to be checked in. We do not need users to understand or see our whole thinking process and experiment.

helix-core/src/test/java/org/apache/helix/cloud/virtualTopologyGroup/TestVirtualTopologyGroupAssignment.java
qqu0127: To clarify, it's for computing the stability of the assignment under cluster change (add or remove instances). It's part of the benchmark, so it's not validating anything, but compute and print statistics for reference.
I do see some value for the benchmark utils, but if it's not the coding convention, I'm totally ok to remove.

zhangmeng916: So what's the expectation for people that run the test suite? Do they need to check the detail in the log to make sure the test runs correctly? Then it does not make much sense. If you can make it pass or fail with a standard, then you can keep it.

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualTopologyGroupScheme.java
qqu0127: Thanks for the comment.
My original idea is to use Enum to achieve singleton pattern while implementing an interface. Future impl can be added as enum value.
However, seems like it's causing more confusion than benefits, let me create standalone impl class instead.

desaikomal (commented):
don't we have naming convention wrt Interface and implementation Java Classes?

I agree that currently you have enum with one value and so in future, you will have more, then you will extend same interface differently??  re-using methods will not be possible with existing code.

qqu0127:
> don't we have naming convention wrt Interface and implementation Java Classes?
I think I'm following the convention, what name specifically are you suggesting?

> 
> I agree that currently you have enum with one value and so in future, you will have more, then you will extend same interface differently?? re-using methods will not be possible with existing code.
If we are following the enum approach, then each impl is a enum value with different method implementation. I don't see any blocker for code reusing even across impls, despite minor refactor. @desaikomal

helix-core/src/main/java/org/apache/helix/cloud/topology/VirtualGroupAssignmentAlgorithm.java
zhangmeng916: Add to @desaikomal 's comment regarding this. The convention is more like `baseProvider` and `dataProvider`, or `verifier` and `zkVerifier`.

zhangmeng916 (approved):
Thanks a lot for the cleanup. This looks much better. Please be sure to address all other reviewers' comments. Also, the PR name should be changed to reflect the commit.

qqu0127:
This PR is ready to merge, approved by @zhangmeng916 

Final commit message:

Introduce VirtualTopologyGroup and its assignment logic
We introduce VirtualTopologyGroup as a virtualization layer on top of physical fault zone, this commit includes its core component and assignment logic.

PR #1951, opened 2022-01-31
Cleanup messages intended for instances that are no longer in cluster

### LINKED ISSUE #1802: Messages intended for instances that are no longer in the cluster are not cleaned up
### Describe the bug
In MessageGenerationPhase, `private void logAndAddToCleanUp()` seems to be handling messages whose intended target hosts are no longer in the cluster. This means that some messages are never pruned from the messages cache even though the instances are no longer in the cluster. This might cause a memory leak and eat up heap space.

### To Reproduce
Observed in a production environment.

### Expected behavior
If the target instance of a message does not exist in the cluster, the message must be removed from the cache.

desaikomal:
Cleanup messages intended for instances that are no longer in the cluster
### Issues

Resolves #1802 

### Description

In MessageGenerationPhase.java, - process() method populates the list of live instances from cache.

But while generateMessage() method has the sessionIdMap information, it still goes through partition/resource/instance map without checking if instance is still part of the cluster or not. 

It is possible that cache has stale entry but that logic needs to be worked separately. But while generating message, we should check if the instance is still there. 

So this is a simple change. We need to still look further if cache is getting invalidated properly.

To make sure that the cache properly is handled/refreshed under instance being replaced or deletion - have filled another bug: https://github.com/apache/helix/issues/1956

### Tests

[INFO] Results:
[INFO]
[INFO] Tests run: 1287, Failures: 0, Errors: 0, Skipped: 0
[INFO]

- This is a simple bug fix and no APIs have been modified.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
@@ -193,7 +193,7 @@ public class MessageGenerationPhase extends AbstractBaseStage {
           }
         }
 
-        if (pendingMessage != null && shouldCleanUpPendingMessage(pendingMessage, currentState,
+        if (shouldCleanUpPendingMessage(pendingMessage, sessionIdMap, instanceName, currentState,
             currentStateOutput.getEndTime(resourceName, partition, instanceName))) {
           logAndAddToCleanUp(messagesToCleanUp, pendingMessage, instanceName, resourceName,
               partition, currentState, PENDING_MESSAGE);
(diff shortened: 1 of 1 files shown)


jiajunwang:
@desaikomal could you please give us some examples of what can fail without this change?

desaikomal:
> 

Essentially, you are looking at the cache to check if instance is present. If a new instance has been added but cache hasn't been updated, we will see those log messages being dropped too.

In this phase while logging messages, cache is considered source-of-truth but it can be out-of-sync wrt real cluster state.

helix-core/src/main/java/org/apache/helix/controller/stages/MessageGenerationPhase.java
```diff
@@ -194,7 +194,8 @@ private void generateMessage(final Resource resource, final BaseControllerDataPr
         }
 
         if (pendingMessage != null && shouldCleanUpPendingMessage(pendingMessage, currentState,
-            currentStateOutput.getEndTime(resourceName, partition, instanceName))) {
+            currentStateOutput.getEndTime(resourceName, partition, instanceName)) && sessionIdMap
+            .containsKey(instanceName)) {
```
junkaixue, line 198: I would suggest to move this check in side function of shouldCleanUpPendingMessage. This could make the all the check logic inside shouldCleanUpPendingMessage(.

junkaixue (approved):
Overall looks good to me. Have you created another issue to track the improvement? If yes, you can add a comment here to link to the issue for future improvement.

PR #1967, opened 2022-02-23
Add Authorization Components to helix-rest

### LINKED ISSUE #1966: Add authorization capability to helix-rest
Helix rest needs to provide possibilities to implement authorization. 

Add interfaces and objects that allow helix-rest endpoints to be protected by authorization. 

**Additional context**

NealSun96:
### Issues

Fixes #1966 

### Description

As the first step of protecting Helix REST endpoints, we're adding authorization components, such as filters and validator interfaces to helix-rest. Users will be able to implement validators and register them to HelixRestServer, which would use these validators to reject unauthorized accesses. 

### Tests

TestAuthValidator

### Documentation (Optional)

### Commits

### Code Quality


helix-rest/src/main/java/org/apache/helix/rest/server/authValidator/DefaultAuthValidator.java
```diff
@@ -0,0 +1,10 @@
+package org.apache.helix.rest.server.authValidator;
+
+import javax.ws.rs.container.ContainerRequestContext;
+
+
+public class DefaultAuthValidator implements AuthValidator {
+  public boolean validate(ContainerRequestContext request) {
+    return true;
+  }
+}
```
qqu0127, line 10: Do we have a plan to add "real" validators? 
I might be missing context, but It feels a little weird to have a standalone default class, which turns out to be a noop. A default method impl may be better depends on your use case.

helix-rest/src/main/java/org/apache/helix/rest/server/HelixRestServer.java
```diff
@@ -94,8 +101,21 @@ public HelixRestServer(List<HelixRestNamespace> namespaces, int port, String url
...
+  public HelixRestServer(List<HelixRestNamespace> namespaces, int port, String urlPrefix,
+      List<AuditLogger> auditLoggers, AuthValidator clusterAuthValidator,
+      AuthValidator namespaceAuthValidator) {
+    init(namespaces, port, urlPrefix, auditLoggers, clusterAuthValidator, namespaceAuthValidator);
+  }
+
   private void init(List<HelixRestNamespace> namespaces, int port, String urlPrefix,
       List<AuditLogger> auditLoggers) {
+    init(namespaces, port, urlPrefix, auditLoggers, new DefaultAuthValidator(),
+        new DefaultAuthValidator());
```
qqu0127, line 113: Curious, does it require different instances?

helix-rest/src/main/java/org/apache/helix/rest/server/authValidator/DefaultAuthValidator.java
NealSun96: This is the default validator for open-source users that don't want to implement validators - the default behavior is no security.

helix-rest/src/main/java/org/apache/helix/rest/server/HelixRestServer.java
NealSun96: Behavior-wise, not really; meaning-wise, they should be separate (1 for cluster, 1 for namespace). I can change it if you insist. :)

helix-rest/src/main/java/org/apache/helix/rest/server/filters/ClusterAuth.java
```diff
@@ -0,0 +1,14 @@
+package org.apache.helix.rest.server.filters;
+
```
narendly, line 2: nit: add license?

helix-rest/src/main/java/org/apache/helix/rest/server/authValidator/AuthValidator.java
```diff
@@ -0,0 +1,8 @@
+package org.apache.helix.rest.server.authValidator;
+
```
narendly, line 2: nit: add license

helix-rest/src/main/java/org/apache/helix/rest/server/authValidator/DefaultAuthValidator.java
```diff
@@ -0,0 +1,10 @@
+package org.apache.helix.rest.server.authValidator;
+
```
narendly, line 2: nit: add license

helix-rest/src/test/java/org/apache/helix/rest/server/TestAuthValidator.java
```diff
@@ -0,0 +1,143 @@
...
+    server.start();
+    _httpClient = HttpClients.createDefault();
+
+    request = buildRequest("/clusters/" + testClusterName, HttpConstants.RestVerbs.GET, "");
+    sendRequestAndValidate(request, Response.Status.OK.getStatusCode());
+    request = buildRequest("/clusters", HttpConstants.RestVerbs.GET, "");
+    sendRequestAndValidate(request, Response.Status.FORBIDDEN.getStatusCode());
+
+    server.shutdown();
+    _httpClient.close();
```
narendly, line 116: why do we need to keep closing the `_httpClient`?

helix-rest/src/main/java/org/apache/helix/rest/server/HelixRestServer.java
```diff
@@ -175,6 +197,8 @@ protected ResourceConfig getResourceConfig(HelixRestNamespace namespace, Servlet
       cfg.register(new CORSFilter());
     }
     cfg.register(new AuditLogFilter(_auditLoggers));
+    cfg.register(new ClusterAuthFilter(_clusterAuthValidator));
+    cfg.register(new NamespaceAuthFilter(_namespaceAuthValidator));
```
narendly, line 201: Is there a strict hierarchy/precedence relationship between cluster and namespace or are they independent?

helix-rest/src/main/java/org/apache/helix/rest/server/authValidator/DefaultAuthValidator.java
qqu0127: I see. In this case, the term "default" could be misleading, let's rename `DefaultAuthValidator` to e.g. `NoopAuthValidator`.
Alternatively, we can even remove this class entirely, and use default method with anonymous class like `new AuthValidator() { }`

helix-rest/src/main/java/org/apache/helix/rest/server/HelixRestServer.java
qqu0127: Not necessary, I don't have strong opinion, let's keep it as is.
I was thinking of enforcing singleton, but realized it's a more implementation specific thing, and probably should leave it to users.

helix-rest/src/test/java/org/apache/helix/rest/server/TestAuthValidator.java
NealSun96: Apparently, if I don't close it here, the 2nd call below this line will hang indefinitely. No idea why and I didn't have time to dig deep. I wanted to ask you too: do you recall something like this happening before during MSDS testing?

helix-rest/src/main/java/org/apache/helix/rest/server/HelixRestServer.java
NealSun96: They are independent: someone could have access to a cluster but not the namespace, vice versa.

narendly: That gets me to wonder why we're not using both annotations for each endpoint? 

Suppose Alice has access to the cluster but not the namespace. Do we need to fail Alice's requests because in Helix, there is a hierarchy between namespaces and clusters - a namespace can have one or more clusters, and clusters must belong to a namespace (unless namespace isn't enabled)?

NealSun96: If Alice has access to the particular cluster but not the namespace, then all requests related to that cluster will succeed; but Alice is not allowed to create clusters or delete clusters. These are 2 different types of privilege. If Alice has access to the namespace, it means Alice can create/delete clusters, but Alice cannot necessarily access a cluster created by someone else (ClusterAuth). 

In the approach you proposed, there needs to be a way to differentiate endpoints based on ClusterAuth and endpoints based on NamespaceAuth, meaning that there must be a list somewhere, hardcoded or configured. I prefer having different annotations to mark which endpoint needs ClusterAuth/NamespaceAuth, than using a list to identify them; the annotations is a cleaner approach to me.

qqu0127:
LGTM. Please resolve other comments, thanks.

helix-rest/src/main/java/org/apache/helix/rest/server/HelixRestServer.java
narendly: Thanks for explaining, it makes sense. I think the current approach should work.

helix-rest/src/test/java/org/apache/helix/rest/server/TestAuthValidator.java
narendly: I don't remember running into this issue, but I think for testing, this shouldn't be a big deal.

narendly (approved):

NealSun96:
Merging to a feature branch in case of future changes as we continue developing Helix ACL.

PR #1976, opened 2022-03-09
Add event handler and event listener interface

### LINKED ISSUE #1965: Cloud event handling mechanism is needed
When instances run on cloud, there could be events (maintenance, out of service, etc.) that happen rarely but could affect the instances. A mechanism to handle such events is needed to minimize the impact.

A listener and callback based event handling framework can help with the situation. If a listener is registered, and when an event is incoming, it will trigger the callback to do some action (disable the instance for example).

**Additional context**

mgao0:
### Issues

Fixes #1965

### Description

This commit creates a skeleton for event handling framework and adds the following classes:
1. CloudEventListener interface
2. CloudEventHandler class
3. CloudEventHandlerFactory class

### Tests

Tests to be added with next PR since this one is mainly interfaces and does not contain much logic.

### Documentation (Optional)

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/cloud/event/CloudEventHandler.java
```diff
@@ -0,0 +1,92 @@
...
+    _postEventHandlerCallback = null;
+  }
+
+  public void onPause(Object eventInfo) {
+    if (_preEventHandlerCallback != null) {
+      _preEventHandlerCallback.onPause(eventInfo);
+    }
+    for (CloudEventListener listener : _eventListeners) {
+      listener.onPause(eventInfo);
+    }
```
junkaixue, line 75: How you take the order of listeners? Are we allowing them to do it by order or parallel is fine. Let's define the behavior. If it can be parallel, we can just Java8 parallel process.

helix-core/src/main/java/org/apache/helix/cloud/event/CloudEventHandler.java
```diff
@@ -0,0 +1,92 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CloudEventHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(CloudEventHandler.class.getName());
+  @VisibleForTesting
+  protected List<CloudEventListener> _eventListeners = new ArrayList<>();
+  private PreEventHandlerCallback _preEventHandlerCallback = null;
+  private PostEventHandlerCallback _postEventHandlerCallback = null;
```
junkaixue, line 34: Except us, nobody may hook this, right? It is to fine-grained to have explicit call pre/post.

We can make internal property of CallbackHanlder to differentiate them by a flag or enum type inside EventHanlderCallback class. It will reduce functions or declarations.

mgao0: Sorry for the confusion. This is for user to hook, they can implement them and register them to the event handler.
I think you are talking about the pre- and post- **callback** inside each listener. 
These two are different - conceptually they will be the first and the last **event listener** to be triggered in an event.
If it's not that clear maybe we can rename these two interfaces?

helix-core/src/main/java/org/apache/helix/cloud/event/CloudEventHandler.java
mgao0: They'll be triggered in parallel. I'll modify the code.

junkaixue: Then we should define the rule in some places. Maybe user manual or some where.

helix-core/src/main/java/org/apache/helix/cloud/event/CloudEventHandler.java
junkaixue: Rename and simplify it with single class with different internal properties could be better. We dont have to get two separate class that invoke in the code.

helix-core/src/main/java/org/apache/helix/cloud/event/CloudEventHandler.java
```diff
@@ -0,0 +1,125 @@
...
+   * 1. PreEventHandlerCallback
+   * 2. Unordered CloudEventListener list
+   * 3. PostEventHandlerCallback
+   * @param eventInfo the object contains any information about the incoming event
+   */
+  public void onPause(Object eventInfo) {
+    if (_preEventHandlerCallback != null) {
+      _preEventHandlerCallback.onPause(eventInfo);
+    }
+    _unorderedEventListenerList.forEach(listener -> listener.onPause(eventInfo));
```
junkaixue, line 103: I remember parallel execution is using object.stream()?

helix-core/src/main/java/org/apache/helix/cloud/event/CloudEventListener.java
```diff
@@ -0,0 +1,38 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * This class is a basic unit for callbacks to handle events.
+ * The listeners can be registered to {@link CloudEventHandler}.
+ */
+public interface CloudEventListener {
```
junkaixue, line 26: Maybe we add set/get event type? So we can know whether it is pre/post by fetch that enum value?
Then the function in cloudevenhandler can reduced to two for register/unregister.

helix-core/src/main/java/org/apache/helix/cloud/event/CloudEventHandler.java
qqu0127: It should be `parallelStream()` or `stream().parallel()`, however parallelism is not guaranteed.

mgao0: Both stream() and forEach() are sequential, so I used parallelStream() instead. But I think sequential is actually fine, what we want to do here is not to guarantee a parallel processing, we just want to emphasize that the list is not ordered. What do you think? @junkaixue

helix-core/src/main/java/org/apache/helix/cloud/event/CloudEventListener.java
mgao0: Updated with getListenerType(). Please take another look

helix-core/src/main/java/org/apache/helix/cloud/event/CloudEventHandler.java
junkaixue: If it is not ordered, let's use parallelStream() since it will not be an efficient design for non order running in sequence.

junkaixue (commented):
Overall lgtm. Make sure @xyuanlu aligning on the APIs

junkaixue (approved):

xyuanlu:
LGTM

mgao0:
This PR is ready to be merged, approved by @junkaixue.

PR #1986, opened 2022-03-16
Read cloud config from zk and propagate to HelixManagerProperty in ZkHelixManager constructor

xyuanlu:
### Issues

- [X] My PR addresses the following Helix issues and references them in the PR description:

#1985 

### Description

- [X] Here are some details about my PR, including screenshots of any UI changes:

Helix user notice that when ZooScalability is enabled, property configured helix cloud participant could not auto register in cluster. This is because when ZooScalability is enabled, Helix need user to pass a cloudConfig object when initiating ZkHelixManager. When user did not set cloudConfig, it will cause nullptr error when reading cloudConfig in ParticipantManager.autoJoin().
This change reads cloudConfig from Zookeeper in ZkHelixManager constructor and propagate to HelixManagerProperty in ZkHelixManager constructor.

### Tests

- [x] The following tests are written for this issue:

NA

- The following is the result of the "mvn test" command on the appropriate module:

(If CI test fails due to known issue, please specify the issue and test PR locally. Then copy & paste the result of "mvn test" to here.)

### Changes that Break Backward Compatibility (Optional)

- My PR contains changes that break backward compatibility or previous assumptions for certain methods or API. They include:

(Consider including all behavior changes for public methods or API. Also include these changes in merge description so that other developers are aware of these changes. This allows them to make relevant code changes in feature branches accounting for the new method/API behavior.)

### Documentation (Optional)

- In case of new functionality, my PR adds documentation in the following wiki page:

(Link the GitHub wiki you added)

### Commits

- My commits all reference appropriate Apache Helix GitHub issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Code Quality

- My diff has been formatted using helix-style.xml 
(helix-style-intellij.xml if IntelliJ IDE is used)

PR #2005, opened 2022-03-31
Populate helix cloud property using cloud config

### LINKED ISSUE #1985: Helix HelixManagerProperty with null cloudConfig causing null ptr error when participant auto join.
### Describe the bug
Helix user notice that when ZooScalability is enabled, property configured helix cloud participant could not auto register in cluster. This is because when ZooScalability is enabled, Helix need user to pass a cloudConfig object when initiating ZkHelixManager. Helix does not read cloudConfig from ZK.

### To Reproduce
This is constantly reproducible. Cloud participant will fail to join cluster with correct cloud config. 

### Expected behavior
When participant has correct cloudConfig, it should be able to join cluster when no cloudConfig object is passed to ZkHelixManager constructor.

mgao0:
### Issues

Fixes #1985

### Description

This is a fix based on a previous [PR](https://github.com/apache/helix/pull/1986). The previous PR retrieves cloud config from zk and replace entire HelixCloudProperty object, which may cause some fields that user pass in in HelixCloudProperty that are not  included in cloud config get missing. This commit changes the logic to only populate fields in HelixCloudProperty with values that are present in cloud config, and leave other fields unchanged.

### Tests

TestMultiZkHelixJavaApis.testZKHelixManagerCloudConfig

[info] ./helix-core/target/surefire-reports/TestSuite.txt: Tests run: 1296, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5,863.791 s - in TestSuite

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
@@ -69,42 +69,56 @@ public class HelixCloudProperty {
   // Other customized properties that may be used.
   private Properties _customizedCloudProperties = new Properties();
 
+  public static final HelixCloudProperty EMPTY_HELIX_CLOUD_PROPERTY = new HelixCloudProperty();
+
+  private HelixCloudProperty() {
+    setCloudEnabled(false);
+  }
+
   /**
    * Initialize Helix Cloud Property based on the provider
    * @param
    */
   public HelixCloudProperty(CloudConfig cloudConfig) {
+    this();
+    populateFieldsWithCloudConfig(cloudConfig);
+  }
+
+  public void populateFieldsWithCloudConfig(CloudConfig cloudConfig) {
+    if (cloudConfig == null) {
+      return;
+    }
     setCloudEnabled(cloudConfig.isCloudEnabled());
     if (cloudConfig.isCloudEnabled()) {
       setCloudId(cloudConfig.getCloudID());
       setCloudProvider(cloudConfig.getCloudProvider());
       switch (CloudProvider.valueOf(cloudConfig.getCloudProvider())) {
-      case AZURE:
-        Properties azureProperties = new Properties();
-        try {
-          InputStream stream = Thread.currentThread().getContextClassLoader()
-              .getResourceAsStream(AZURE_CLOUD_PROPERTY_FILE);
-          azureProperties.load(stream);
-        } catch (IOException e) {
-          String errMsg =
-              "failed to open Helix Azure cloud properties file: " + AZURE_CLOUD_PROPERTY_FILE;
-          throw new IllegalArgumentException(errMsg, e);
-        }
-        LOG.info("Successfully loaded Helix Azure cloud properties: {}", azureProperties);
-        setCloudInfoSources(
-            Collections.singletonList(azureProperties.getProperty(CLOUD_INFO_SOURCE)));
-        setCloudInfoProcessorName(azureProperties.getProperty(CLOUD_INFO_PROCESSOR_NAME));
-        setCloudMaxRetry(Integer.valueOf(azureProperties.getProperty(CLOUD_MAX_RETRY)));
-        setCloudConnectionTimeout(Long.valueOf(azureProperties.getProperty(CONNECTION_TIMEOUT_MS)));
-        setCloudRequestTimeout(Long.valueOf(azureProperties.getProperty(REQUEST_TIMEOUT_MS)));
-        break;
-      case CUSTOMIZED:
-        setCloudInfoSources(cloudConfig.getCloudInfoSources());
-        setCloudInfoProcessorName(cloudConfig.getCloudInfoProcessorName());
-        break;
-      default:
-        throw new HelixException(
-            String.format("Unsupported cloud provider: %s", cloudConfig.getCloudProvider()));
+        case AZURE:
+          Properties azureProperties = new Properties();
+          try {
+            InputStream stream = Thread.currentThread().getContextClassLoader()
+                .getResourceAsStream(AZURE_CLOUD_PROPERTY_FILE);
+            azureProperties.load(stream);
+          } catch (IOException e) {
+            String errMsg =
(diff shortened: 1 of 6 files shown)


helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -66,11 +66,8 @@ private HelixManagerProperty(String version, long healthReportLatency,
   }
 
   public HelixCloudProperty getHelixCloudProperty() {
-    return _helixCloudProperty;
-  }
-
-  public void setHelixCloudProperty(CloudConfig cloudConfig) {
-    _helixCloudProperty = new HelixCloudProperty(cloudConfig);
+    return _helixCloudProperty == null ? HelixCloudProperty.EMPTY_HELIX_CLOUD_PROPERTY
+        : _helixCloudProperty;
```
qqu0127, line 70: Can we pre-populate the default value so that `_helixCloudProperty` is always not null?

junkaixue:
Have a test for this change of code?

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
mgao0: Thanks for the suggestion, but pre-populate default value won't help here because this variable is assigned in `HelixManagerProperty` constructor, not in a setter. So if null is passed into the constructor, even with pre-populated value, it will still be assigned to null.

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
```diff
@@ -69,16 +69,28 @@
...
    * @param
    */
   public HelixCloudProperty(CloudConfig cloudConfig) {
+    this();
+    populateFieldsWithCloudConfig(cloudConfig);
+  }
+
+  public void populateFieldsWithCloudConfig(CloudConfig cloudConfig) {
+    if (cloudConfig == null || !cloudConfig.isCloudEnabled()) {
+      reset();
```
xyuanlu, line 87: Can we create a default CloudConfig using `buildEmptyCloudConfig()` and populate the fields in cloudConfig to cloudProperty? 
It would be simpler I guess.

mgao0: It will still skip the assignments to the variables because `cloudConfig.isCloudEnabled` returns false. But this reminds me, maybe I should change the assignment logic in this method. Let me try and update the PR if it works.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
qqu0127: Thanks for the info.
`it will still be assigned to null.` This is technically right, but is also the confusing part. Does a null value make sense or even be accessible at all? The public getter returns either the non-null assigned value, or an empty new instance. If possible, let's try to do the null check at assignment time. Also please let me know if I misunderstood anything.

mgao0: Thanks for explaining. I was trying to prevent a potential NPE so I return a new instance instead of null, but practically the old code doesn't have functionality concerns.
Do a null check at assignment time is technically correct but is not necessary because:
1. There are multiple places (two constructors and a setter function) that this value is set, so this can lead to duplicated code.
2. There are Helix managers that will not be related to cloud at all, so this field is not useful. If we assign default value to it, it can cause memory waste.
I modified my change to do lazy instantiation on this field, but kept it in the getter. Hope you find this reasonable. Thanks.

helix-core/src/main/java/org/apache/helix/HelixCloudProperty.java
mgao0: Updated! @xyuanlu

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
qqu0127: That makes sense, thanks.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
```diff
@@ -94,6 +94,10 @@ public CloudConfig(String cluster, boolean enabled, CloudProvider cloudProvider,
     }
   }
 
+  public static CloudConfig buildEmptyCloudConfig() {
```
xyuanlu, line 97: Do you think is it better to change this to a default constructor?

xyuanlu:
Thanks for addressing the comment. 
Generally looks good to me.

helix-core/src/main/java/org/apache/helix/model/CloudConfig.java
mgao0: Good point. Updated.

junkaixue:
@xyuanlu do you mean approval for this PR? If yes, then just review with approval  option.

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -113,10 +109,10 @@ public static CloudConfig getCloudConfig(String zkAddress, String clusterName) {
       // The try-catch logic is for backward compatibility reason only. Even if the cluster is not set
       // up yet, constructing a new ZKHelixManager should not throw an exception
       try {
-        cloudConfig = configAccessor.getCloudConfig(clusterName) == null ? buildEmptyCloudConfig()
+        cloudConfig = configAccessor.getCloudConfig(clusterName) == null ? new CloudConfig()
             : configAccessor.getCloudConfig(clusterName);
```
qqu0127, line 113: This queries cloudConfig twice, and not atomic... Can we assign it with `configAccessor.getCloudConfig(clusterName)` first and then do null check? 
(sorry I missed this one in the first pass)

mgao0: Updated.

qqu0127 (approved):
LGTM, thanks for addressing the comments

mgao0:
This PR is ready to be merged, approved by @qqu0127 .

PR #2011, opened 2022-04-01
Fix TestDropResourceMetricsReset

### LINKED ISSUE #1980: [Failed CI Test] testDropWithNoCurrentState(org.apache.helix.monitoring.mbeans.TestDropResourceMetricsReset)
This issue is created for tracking unstable test: testDropWithNoCurrentState(org.apache.helix.monitoring.mbeans.TestDropResourceMetricsReset)

qqu0127:
### Issues

Fixes #1980 

#### Overview

Fix failing test TestDropResourceMetricsReset

A recent ZK version bump reveals a seemingly long-hidden code issue regarding resource metrics. It's possible the test never works in the expected way. 

#### What
Is it a faulty test or a code issue we should fix?
-- More ideas welcome! 
As of now we believe it's a code issue and need a fix, please also see https://github.com/apache/helix/pull/2011#issuecomment-1087807581

#### Why
The test failure could be reproduced stably using **either ZK version** by adding Thread.sleep() between `participant.syncStop()` and `setupTool.dropResourceFromCluster`. So this indicates the test was wrong (and succeeded in a wrong way) for a while, there are race condition and zk version change introduce unrelated delay that finally reveal the issue.

mBean is only unregistered during ExternalViewComputeStage in https://github.com/apache/helix/blob/master/helix-core/src/main/java/org/apache/helix/controller/stages/ExternalViewComputeStage.java#L99.
 It's not always happening after the resource is dropped as there is no `CurrentStateChange` event anymore, only `IdealStateChange`. 

#### How

Fixing the issue itself requires more changes and careful design. We'd want to first put a temp workaround for the test case by manually triggering a cluster event.

### Tests

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -552,7 +552,7 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
       autoExitMaintenancePipeline.addStage(new MaintenanceRecoveryStage());
 
       registry.register(ClusterEventType.IdealStateChange, dataRefresh, dataPreprocess,
-          rebalancePipeline);
+          externalViewPipeline, rebalancePipeline);
       registry.register(ClusterEventType.CurrentStateChange, dataRefresh, dataPreprocess,
           externalViewPipeline, rebalancePipeline);
       registry.register(ClusterEventType.InstanceConfigChange, dataRefresh, dataPreprocess,
--- a/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
@@ -552,7 +552,7 @@ public class GenericHelixController implements IdealStateChangeListener, LiveIns
       autoExitMaintenancePipeline.addStage(new MaintenanceRecoveryStage());
 
       registry.register(ClusterEventType.IdealStateChange, dataRefresh, dataPreprocess,
-          externalViewPipeline, rebalancePipeline);
+          rebalancePipeline);
       registry.register(ClusterEventType.CurrentStateChange, dataRefresh, dataPreprocess,
           externalViewPipeline, rebalancePipeline);
       registry.register(ClusterEventType.InstanceConfigChange, dataRefresh, dataPreprocess,
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/ExternalViewComputeStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/ExternalViewComputeStage.java
@@ -96,7 +96,7 @@ public class ExternalViewComputeStage extends AbstractAsyncBaseStage {
 
     // Keep MBeans for existing resources and unregister MBeans for dropped resources
     if (clusterStatusMonitor != null) {
-      clusterStatusMonitor.retainResourceMonitor(monitoringResources);
+      //clusterStatusMonitor.retainResourceMonitor(monitoringResources);
     }
 
     List<String> externalViewsToRemove = new ArrayList<>();
@@ -179,7 +179,7 @@ public class ExternalViewComputeStage extends AbstractAsyncBaseStage {
           .equalsIgnoreCase(DefaultSchedulerMessageHandlerFactory.SCHEDULER_TASK_QUEUE)) {
         clusterStatusMonitor
             .setResourcePendingMessages(resourceName ,totalPendingMessageCount);
-        monitoringResources.add(resourceName);
+        //monitoringResources.add(resourceName);
       }
     }
 
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/ExternalViewComputeStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/ExternalViewComputeStage.java
@@ -96,7 +96,7 @@ public class ExternalViewComputeStage extends AbstractAsyncBaseStage {
 
     // Keep MBeans for existing resources and unregister MBeans for dropped resources
     if (clusterStatusMonitor != null) {
-      //clusterStatusMonitor.retainResourceMonitor(monitoringResources);
+      clusterStatusMonitor.retainResourceMonitor(monitoringResources);
     }
 
     List<String> externalViewsToRemove = new ArrayList<>();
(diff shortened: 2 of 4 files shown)


helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
```diff
@@ -552,7 +552,7 @@ private static PipelineRegistry createDefaultRegistry(String pipelineName) {
       autoExitMaintenancePipeline.addStage(new MaintenanceRecoveryStage());
 
       registry.register(ClusterEventType.IdealStateChange, dataRefresh, dataPreprocess,
-          rebalancePipeline);
+          externalViewPipeline, rebalancePipeline);
```
desaikomal, line 555: It is non-deterministic test which we all agree as test is not failing consistently in CI. Just curious, changing the code is the right behavior or should mock<> implementation of controller needs to change??

desaikomal (commented):
Thanks a lot Quincy for taking a look at the failure (Appreciate a lot)

helix-core/src/main/java/org/apache/helix/controller/GenericHelixController.java
qqu0127: Right now I doubt the non-deterministic part, like, what if we have `setupTool.dropResourceFromCluster` long enough after `participant.syncStop()`, will the test still pass occasionally? I documented something in description regarding ExternalViewComputeStage.
This is a good topic for discussion, we need to review and define the expected behavior. If the test is describing the right behavior, then it's a bug in code we need to fix.

qqu0127:
The attempted fix caused another test failure in `TestClusterMaintenanceMode.testMaintenanceModeAddNewResource`, where it checks the external view of resource added when cluster is in maintenance mode.

helix-core/src/main/java/org/apache/helix/controller/stages/ExternalViewComputeStage.java
```diff
@@ -96,7 +96,7 @@ public void execute(final ClusterEvent event) throws Exception {
 
     // Keep MBeans for existing resources and unregister MBeans for dropped resources
     if (clusterStatusMonitor != null) {
-      clusterStatusMonitor.retainResourceMonitor(monitoringResources);
+      //clusterStatusMonitor.retainResourceMonitor(monitoringResources);
```
narendly, line 99: First off, great work on debugging this and identifying the root cause!

Have a few points to add:

- First, with a guiding question: conceptually, where would be the "correct" place for metric-cleanup logic to be? 
- We want the metric cleanup logic to be in an async stage in order to not block the critical controller pipeline. As such, adding the metric-cleanup logic in ResourceValidation may not be appropriate (this is not an async stage). If you believe that this logic shouldn't belong in ExternalViewComputeStage, then can we think about a different (or a new) async stage where this logic can be appropriately placed? With that said though, if you place it in another async stage, would we still face the same issue of this logic not triggering in tests?
- If you think ExternalViewComputeStage is an appropriate place for this cleanup logic, can we simply add a CurrentState change event to artificially trigger the stage just for the sake of making the test case work? I think this is also a valid way because in real-life scenarios we can rely on the ExtViewComputeStage to eventually get called asynchronously. We can think of this as making the test environment model the real environment more closely.

qqu0127:
Adding more ideas and discussion from offline regarding the fundamental problem statement:

- @qqu0127 : I’d like to first check with you, do you agree if it’s a code issue we need to fix? (compared to disabling / removing the test). I’m debating myself with another argument like -- it is there for a long time, it seems not hurting anything and almost a feature.
- @narendly : Short answer is that I do think we need to have coverage of this ... We have had issues for Helix where metrics were never properly cleaned up and caused JVM heap OOM issues

helix-core/src/main/java/org/apache/helix/controller/stages/ExternalViewComputeStage.java
qqu0127: Thanks for the comment!

> where would be the "correct" place for metric-cleanup logic to be?

With limited context and knowledge, I think both `ResourceValidationStage` and `ExternalViewComputeStage` make sense. MetricsMonitors belong to, or at least associated with resources. On the other hand, generally speaking, it's also a "view". 
it's also applicable to be in a standalone stage instead of mixing up with other logic.

> async stage...

Yeah, this is a good point I wasn't aware of. So essentially, we'd want to run it in async way, and make sure all related cluster events trigger such computation. 

> add a CurrentState change event to artificially trigger the stage

This is definitely an option for the fix. But could it introduce more overhead and side effect? And what the difference from adding `externalViewPipeline` to `IdealStateChange` registry? All we need is refresh the resource monitors, but we will be running a few other pipelines.

qqu0127:
The failing test `testBatchMessageThreadPoolSize` is a known issue in https://github.com/apache/helix/issues/1989

helix-core/src/test/java/org/apache/helix/monitoring/mbeans/TestDropResourceMetricsReset.java
```diff
@@ -148,6 +148,10 @@ public void testDropWithNoCurrentState() throws Exception {
     // Drop the resource
     setupTool.dropResourceFromCluster(clusterName, RESOURCE_NAME);
 
+    // TEMP WORKAROUND, trigger a CurrentStateChange to remove the resource monitor
+    // https://github.com/apache/helix/issues/1980
+    participant.syncStart();
```
narendly, line 153: could we give a little more context? (e.g. adding a liveinstance has an effect of creating a CurrentStageChange event, **which in turn triggers ExternalViewStage that includes the metric cleanup logic**)

also, should we stop this participant when we clean up (line 159 and beyond)? otherwise, it will leave a dangling thread.

qqu0127: Good point, updated. Thanks.

junkaixue (approved):

narendly (approved):
Thanks for a thorough root cause analysis! Hope you got a good look at some Helix internals through this :)

qqu0127:
This PR is ready to merge, approved by @junkaixue @narendly 
Commit message:

Fix TestDropResourceMetricsReset
Add a temp workaround by manually triggering a CurrentStageChange event for ExternalViewStage computation so that resource monitor can be cleaned up in unit test.

PR #2013, opened 2022-04-03
fix repoducible builds issue

hboutemy:
### Issues

this is the only change required to get a fully Reproducible Builds


junkaixue (approved):

junkaixue:
Thanks for help!!

Issue #2034, opened 2022-04-13
MaintenanceManagementService should allow impl of OperationInterface to return null


xyuanlu:
**Is your feature request related to a problem? Please describe.**
Maintenance Management service fails silently when user provided OperationInterface returns null in take/freeInstance API. 

**Describe the solution you'd like**
We should allow user provided OperationInterface to return null.

PR #2035, opened 2022-04-13
MaintenanceManagementService  improvement - Allow implementation of OperationInterface to return null

xyuanlu:
### Issues

#2034 

### Description

Maintenance Management service fails silently when user provided OperationInterface returns null in take/freeInstance API.
We should allow user provided OperationInterface to return null.

### Tests

Modified TestOperationImpl.java.
TestInstanceAccessor will cover null case.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-rest/src/main/java/org/apache/helix/rest/clusterMaintenanceService/MaintenanceManagementInstanceInfo.java
+++ b/helix-rest/src/main/java/org/apache/helix/rest/clusterMaintenanceService/MaintenanceManagementInstanceInfo.java
@@ -30,54 +30,54 @@ public class MaintenanceManagementInstanceInfo {
     FAILURE
   }
 
-  private String operationResult;
-  private OperationalStatus status;
-  private List<String> messages;
+  private String _operationResult;
+  private OperationalStatus _status;
+  private List<String> _messages;
 
   public MaintenanceManagementInstanceInfo(OperationalStatus status) {
-    this.status = status;
-    this.messages = new ArrayList<>();
-    this.operationResult = "";
+    this._status = status;
+    this._messages = new ArrayList<>();
+    this._operationResult = "";
   }
 
   public MaintenanceManagementInstanceInfo(OperationalStatus status, List<String> messages) {
-    this.status = status;
-    this.messages = messages;
-    this.operationResult = "";
+    this._status = status;
+    this._messages = messages;
+    this._operationResult = "";
   }
 
   public MaintenanceManagementInstanceInfo(OperationalStatus status, String newOperationResult) {
-    this.status = status;
-    this.operationResult = newOperationResult;
-    this.messages = new ArrayList<>();
+    this._status = status;
+    this._operationResult = newOperationResult;
+    this._messages = new ArrayList<>();
   }
 
   public List<String> getMessages() {
-    return messages;
+    return _messages;
   }
 
   public String getOperationResult() {
-    return operationResult;
+    return _operationResult;
   }
 
   public boolean hasOperationResult() {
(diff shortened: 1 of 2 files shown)


helix-rest/src/main/java/org/apache/helix/rest/clusterMaintenanceService/MaintenanceManagementInstanceInfo.java
```diff
@@ -84,11 +85,12 @@ public void mergeResult(MaintenanceManagementInstanceInfo info) {
...
   public void mergeResult(MaintenanceManagementInstanceInfo info, boolean nonBlockingFailure) {
-    messages.addAll(info.getMessages());
-    status =
-        (info.isSuccessful() || nonBlockingFailure) && isSuccessful() ? OperationalStatus.SUCCESS
-            : OperationalStatus.FAILURE;
-    if (info.hasOperationResult()) {
+    if (info != null) {
+      messages.addAll(info.getMessages());
+    }
+    status = ((info == null || info.isSuccessful() || nonBlockingFailure) && isSuccessful())
```
qqu0127, line 91: Is `status` a class attribute? Let's follow the naming convention like `_status`, thanks.

qqu0127:
The impl looks good with minor comments, but here are some general questions on the design.
Could you share more context on why the service silently fail? 
Also I'm wondering if a null value makes sense in the first place. Will something like a default or empty value be better?

helix-rest/src/main/java/org/apache/helix/rest/clusterMaintenanceService/MaintenanceManagementInstanceInfo.java
```diff
@@ -84,11 +85,12 @@ public void mergeResult(MaintenanceManagementInstanceInfo info) {
   }
 
   public void mergeResult(MaintenanceManagementInstanceInfo info, boolean nonBlockingFailure) {
-    messages.addAll(info.getMessages());
-    status =
-        (info.isSuccessful() || nonBlockingFailure) && isSuccessful() ? OperationalStatus.SUCCESS
-            : OperationalStatus.FAILURE;
-    if (info.hasOperationResult()) {
+    if (info != null) {
```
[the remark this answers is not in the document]
xyuanlu: Good point. Updated. 
I think if we get null `info`, we could keep status the same..?

helix-rest/src/main/java/org/apache/helix/rest/clusterMaintenanceService/MaintenanceManagementInstanceInfo.java
xyuanlu: TFTR. Updated.

xyuanlu:
> The impl looks good with minor comments, but here are some general questions on the design. Could you share more context on why the service silently fail? Also I'm wondering if a null value makes sense in the first place. Will something like a default or empty value be better?

Thanks for the reply.
If we get null for MaintenanceManagementInstanceInfo, then in MaintenanceManagementService.takeFreeSingleInstanceHelper() will throw npr exception causing operation failure.

This function execute user implemented interface `OperationInterface`. It  should make sense to allow other user to return null.

qqu0127 (approved):
LGTM, please address other comments and get committer approval.

desaikomal (commented):
LGTM ... simple change..

xyuanlu:
This PR is ready to be merged. Approved by @qqu0127 @desaikomal //Thanks!
Commit message:
Allow implementation of OperationInterface to return null

PR #2065, opened 2022-04-27
fix bug where generateJobList would iterate until the parallelism is reached

### LINKED ISSUE #2064: `RuntimeJobDag` construction time depends on ParallelJobs, not the number of jobs to execute
The following code is executed every time a `RuntimeJobDag` is constructed:

```java
  public void generateJobList() {
    resetJobListAndDependencyMaps();
    computeIndependentNodes();
    _readyJobList.addAll(_independentNodes);
    if (_isJobQueue && _readyJobList.size() > 0) {
      // For job queue, only get number of parallel jobs to run in the ready list.
      for (int i = 1; i < _numParallelJobs; i++) {
        if (_parentsToChildren.containsKey(_readyJobList.peekLast())) {
          _readyJobList.offer(_parentsToChildren.get(_readyJobList.peekLast()).iterator().next());
        }
      }
    }
    _hasDagChanged = false;
  }
```

This code looks suspicious because it only ever puts the first child onto the ready job list, but also iterates beyond the number of jobs to execute, for instance if `_numParallelJobs` is set to `Integer.MAX_VALUE` as it is in Apache Pinot [here](https://github.com/apache/pinot/blob/6eff17754a0b0ffdfb8670b54b5ada38961b7917/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/minion/PinotHelixTaskResourceManager.java#L116).

This shows up by making integration tests slow, dominating profiles of tests.
Screenshot 2022-04-27 at 12 15 50
Tracing `RuntimeJobDag` reveals that with `_numParallelJobs` equal to `Integer.MAX_VALUE` construction can take up to 20 seconds to produce very small ready job lists.

Screenshot 2022-04-27 at 12 17 26

Setting the parallelism to 1 [here](https://github.com/apache/pinot/pull/8603) halves integration test time, but prevents parallelism. 

`RuntimeJobDag. generateJobList ` should terminate once all the jobs have been added to the ready list, or the maximum parallelism has been reached, whichever happens first.

richardstartin:
### Issues

Fixes #2064

### Description

When there is a high level of parallelism configured and the JobDag is a job queue, building the ready job list would take a long time because the loop iterates until the level of parallelism is reached, and doesn't terminate when all the jobs have been processed

### Tests

`TestRuntimeJobDag#testBuildJobQueueWithParallelismExceedingJobCount` was added to verify that the correct job order is still produced. It is not a performance test, but before the fix it took 22 seconds to complete, afterwards it takes ~1ms.

None.

### Documentation (Optional)

This is a bug fix, no documentation required.

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
+++ b/helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
@@ -198,12 +198,14 @@ public class RuntimeJobDag extends JobDag {
     resetJobListAndDependencyMaps();
     computeIndependentNodes();
     _readyJobList.addAll(_independentNodes);
-    if (_isJobQueue && _readyJobList.size() > 0) {
+    if (_isJobQueue && !_readyJobList.isEmpty()) {
       // For job queue, only get number of parallel jobs to run in the ready list.
       for (int i = 1; i < _numParallelJobs; i++) {
-        if (_parentsToChildren.containsKey(_readyJobList.peekLast())) {
-          _readyJobList.offer(_parentsToChildren.get(_readyJobList.peekLast()).iterator().next());
+        Set<String> children = _parentsToChildren.get(_readyJobList.peekLast());
+        if (children == null) {
+          break;
         }
+        _readyJobList.offer(children.iterator().next());
       }
     }
     _hasDagChanged = false;
--- a/helix-core/src/test/java/org/apache/helix/integration/task/TestRuntimeJobDag.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/task/TestRuntimeJobDag.java
@@ -24,11 +24,25 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import org.apache.helix.task.JobDag;
 import org.apache.helix.task.RuntimeJobDag;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
 public class TestRuntimeJobDag {
+
+  @Test
+  public void testBuildJobQueueWithParallelismExceedingJobCount() {
+    JobDag jobDag = new JobDag();
+    jobDag.addNode("parent");
+    jobDag.addParentToChild("parent", "child");
+    jobDag.addParentToChild("child", "grandchild");
+    RuntimeJobDag runtimeJobDag = new RuntimeJobDag(jobDag, true, Integer.MAX_VALUE, 1);
+    Assert.assertEquals(runtimeJobDag.getNextJob(), "parent");
+    Assert.assertEquals(runtimeJobDag.getNextJob(), "child");
+    Assert.assertEquals(runtimeJobDag.getNextJob(), "grandchild");
+  }
+
   private Set<String> actualJobs;
   private Set<String> expectedJobs;
 


helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
```diff
@@ -198,11 +198,15 @@ public void generateJobList() {
     resetJobListAndDependencyMaps();
     computeIndependentNodes();
     _readyJobList.addAll(_independentNodes);
-    if (_isJobQueue && _readyJobList.size() > 0) {
+    if (_isJobQueue && !_readyJobList.isEmpty()) {
```
junkaixue, line 201: Not quite sure whether this is just a style change? If we dont change it, would it cause any problem?

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
```diff
@@ -198,11 +198,15 @@ public void generateJobList() {
...
       // For job queue, only get number of parallel jobs to run in the ready list.
       for (int i = 1; i < _numParallelJobs; i++) {
-        if (_parentsToChildren.containsKey(_readyJobList.peekLast())) {
-          _readyJobList.offer(_parentsToChildren.get(_readyJobList.peekLast()).iterator().next());
+        String parent = _readyJobList.peekLast();
+        Set<String> children = _parentsToChildren.get(parent);
+        if (children != null) {
+          _readyJobList.offer(children.iterator().next());
+        } else {
+          break;
```
junkaixue, line 209: NIT: it would be better make it as:

if (children == null) {
break;
} 

_readyJobList.offer(children.iterator().next());

junkaixue (approved):
Overall, it looks to me. Thanks for contributing!

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
richardstartin: ArrayDeque.size has a nontrivial implementation (https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/ArrayDeque.java#L649) but this is an emptiness check.

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
richardstartin: Sure I can change this if you like.

qqu0127 (approved):
LGTM, thanks for the contribution.

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
narendly: Little did I know! I think this is good.

helix-core/src/main/java/org/apache/helix/task/RuntimeJobDag.java
richardstartin: Done

junkaixue:
@richardstartin Please let us know when it is ready to merge. Helix usually has the last step to let author to confirm when it is ready to merge if you check the PR guidance.

richardstartin:
> @richardstartin Please let us know when it is ready to merge. Helix usually has the last step to let author to confirm when it is ready to merge if you check the PR guidance.

@junkaixue I am happy for this to be merged now, thanks for letting me know.

PR #2072, opened 2022-04-28
Issue #1970: allow clients to prevent HelixProperty from cloning the ZNRecord

### LINKED ISSUE #1970: `HelixProperty` cloning `ZNRecord` is expensive
Here is a profile from Apache Pinot's broker:
Screenshot 2022-03-04 at 10 36 58

Each red box duplicates work already done to produce a `ZNRecord` (at significant cost) in the adjacent blue box.

A lot of time is spent cloning `ZNRecord`s in `HelixProperty` but this is unfortunate because the ZNRecord used to construct the `HelixProperty` instances is never used again, so doesn't need to be cloned. I don't understand why it's important for an update to a `ZNRecord` instance not to be reflected in the corresponding `HelixProperty`.

richardstartin:
…ZNRecord

### Issues

Fixes #1970

### Description

In Apache Pinot, we first deserialize `ZNRecord`s and then construct `HelixProperty`s from them, and never use the `ZNRecord` again. `HelixProperty` makes a defensive copy of the `ZNRecord`, which is actually expensive enough to show up prominently in profiles of our helix threads. This change adds constructors to `HelixConstructor` which do not clone the `ZNRecord`. Backward compatibility is maintained, and all existing code will continue to clone the `ZNRecord`, but new code can be written to set the `clone` to false to avoid the expensive copy.

### Tests

No tests have been added as this change is trivial.

Backward compatibility has been maintained. Existing code continues to clone the `ZNRecord`.

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/HelixProperty.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixProperty.java
@@ -39,7 +39,7 @@ import org.slf4j.LoggerFactory;
  * A wrapper class for ZNRecord. Used as a base class for IdealState, CurrentState, etc.
  */
 public class HelixProperty {
-  private static Logger LOG = LoggerFactory.getLogger(HelixProperty.class);
+  private static final Logger LOG = LoggerFactory.getLogger(HelixProperty.class);
 
   public enum HelixPropertyAttribute {
     BUCKET_SIZE, BATCH_MESSAGE_MODE
@@ -156,25 +156,50 @@ public class HelixProperty {
    * @param id
    */
   public HelixProperty(String id) {
-    this(new ZNRecord(id), id);
+    this(new ZNRecord(id), id, false);
+  }
+
+  /**
+   * Initialize the property with an existing ZNRecord
+   * @param record a deep copy of the record is made, updates to this record will not be reflected
+   *               by the HelixProperty
+   */
+  public HelixProperty(ZNRecord record) {
+    this(record, true);
   }
 
   /**
    * Initialize the property with an existing ZNRecord
    * @param record
+   * @param deepCopyRecord set to true to make a copy of the ZNRecord, false otherwise
    */
-  public HelixProperty(ZNRecord record) {
-    this(record, record.getId());
+  public HelixProperty(ZNRecord record, boolean deepCopyRecord) {
+    this(record, record.getId(), deepCopyRecord);
+  }
+
+  /**
+   * Initialize the property with an existing ZNRecord with new record id
+   * @param record a deep copy of the record is made, updates to this record will not be reflected by the HelixProperty
+   * @param id
+   */
+  public HelixProperty(ZNRecord record, String id) {
+    this(record, id, true);
   }
 
   /**
    * Initialize the property with an existing ZNRecord with new record id
    * @param record
    * @param id
+   * @param deepCopyRecord whether to deep copy the ZNRecord, set to true if subsequent changes to
+   *                       the ZNRecord should not affect this HelixProperty and vice versa, or false
    */
-  public HelixProperty(ZNRecord record, String id) {
-    _record = (record instanceof SessionAwareZNRecord) ? new SessionAwareZNRecord(record, id)
-        : new ZNRecord(record, id);
+  public HelixProperty(ZNRecord record, String id, boolean deepCopyRecord) {
+    if (deepCopyRecord) {
+      _record = record instanceof SessionAwareZNRecord ? new SessionAwareZNRecord(record, id)
+          : new ZNRecord(record, id);
+    } else {
+      _record = record;
+    }
     _stat = new Stat(_record.getVersion(), _record.getCreationTime(), _record.getModifiedTime(),
         _record.getEphemeralOwner());
   }


helix-core/src/main/java/org/apache/helix/HelixProperty.java
```diff
@@ -39,7 +39,7 @@
  * A wrapper class for ZNRecord. Used as a base class for IdealState, CurrentState, etc.
  */
 public class HelixProperty {
```
narendly, line 41: Correct me if I'm wrong, but because we "deep copy" `HelixProperty` and its internal `ZNRecord`, throughout Helix we can treat this as such, and that is, we can assume that it's immutable. But in the case that we do not clone it and simply take after the internal `ZNRecord` reference, such assumptions no longer hold and therefore may open up avenues where issues could creep in. And this particular change doesn't allow us to know whether its internal `ZNRecord` reference is being shared across different `HelixProperty` instances or not.

Could we think of alternative ways to achieve this if you think this is necessary? For example, add a subclass of `HelixProperty` that shallow copies `ZNRecord`, so that there is no ambiguity?

richardstartin: Of course, this does not clone the `ZNRecord`, so code which assumes the `ZNRecord` has been mutated will not work properly, but these constructors are made available for advanced users of Helix, like Apache Pinot. In our code, the copy is needless and shows up as a surprisingly noticeable cost, which I would like to avoid incurring. This is why I took care to document the constructors; use at your own risk, take responsibility for it, but make it possible not to incur the cost.

Subclassing would not work because `HelixProperty` would need to be constructed to construct the subclass, which would result in a clone. It could work the other way around though, but this flag feels like a simple, easy to document, and backward compatible solution.

helix-core/src/main/java/org/apache/helix/HelixProperty.java
```diff
@@ -156,25 +156,49 @@ public String toString() {
...
+    this(record, true);
   }
 
   /**
-   * Initialize the property with an existing ZNRecord with new record id
+   * Initialize the property with an existing ZNRecord
    * @param record
+   * @param clone set to true to make a copy of the ZNRecord, false otherwise
+   */
+  public HelixProperty(ZNRecord record, boolean clone) {
```
junkaixue, line 175: NIT: Let's have a more clear naming for this. "clone" is a little bit confusing. Could be enableDeepCopy or something else.

junkaixue (approved):
Overall, looks good to me.

helix-core/src/main/java/org/apache/helix/HelixProperty.java
richardstartin: OK done, I went for `deepCopyRecord`

richardstartin:
@junkaixue I am happy for this to be merged when you are (but let's wait for a clean run of the tests after the variable rename).

junkaixue:
@richardstartin Thanks for making the change. But if we change it to deepCopyRecord, you may need to change the boolean value. Since the value "false" means not deep copying literally, the default constructor default value should be true.

richardstartin:
@junkaixue the record is only copied if `deepCopyRecord` is true. The default value is true _except_ when the `ZNRecord` originates within a constructor.

helix-core/src/main/java/org/apache/helix/HelixProperty.java
```diff
@@ -156,25 +156,50 @@ public String toString() {
    * @param id
    */
   public HelixProperty(String id) {
-    this(new ZNRecord(id), id);
+    this(new ZNRecord(id), id, false);
```
richardstartin, line 159: @junkaixue default value is false here because there is no need to make a copy of an `ZNRecord` the caller hasn't passed in.

helix-core/src/main/java/org/apache/helix/HelixProperty.java
```diff
@@ -156,25 +156,50 @@ public String toString() {
...
 
   /**
    * Initialize the property with an existing ZNRecord
-   * @param record
+   * @param record a deep copy of the record is made, updates to this record will not be reflected
+   *               by the HelixProperty
    */
   public HelixProperty(ZNRecord record) {
-    this(record, record.getId());
+    this(record, true);
```
richardstartin, line 168: @junkaixue default value true here because it came from the caller.

helix-core/src/main/java/org/apache/helix/HelixProperty.java
```diff
@@ -156,25 +156,50 @@ public String toString() {
...
+
+  /**
+   * Initialize the property with an existing ZNRecord with new record id
+   * @param record a deep copy of the record is made, updates to this record will not be reflected by the HelixProperty
    * @param id
    */
   public HelixProperty(ZNRecord record, String id) {
-    _record = (record instanceof SessionAwareZNRecord) ? new SessionAwareZNRecord(record, id)
-        : new ZNRecord(record, id);
+    this(record, id, true);
```
richardstartin, line 186: @junkaixue default value true here because it came from the caller.

helix-core/src/main/java/org/apache/helix/HelixProperty.java
narendly: Yes, I understand the motivation behind this. That still doesn't give us a way to tell if the object is thoroughly immutable (because the underlying ZNRecord has been deep-copied) - I was thinking maybe adding a final `boolean` that can tell us whether it's mutable or immutable? Do you think that would be useful for the users of `HelixProperty`?

junkaixue:
Since this is very fundament change, @richardstartin are you able to provide the test results? Github machine usually is slow and could cause unstable failure. Better to run it on local linux machine.

helix-core/src/main/java/org/apache/helix/HelixProperty.java
richardstartin: But you don't have that guarantee anyway - `HelixProperty` exposes a reference to its `_record` with `getRecord` and the `ZNRecord` itself is mutable.

narendly: In that case, I think this change is okay since we didn't have that guarantee to begin with. Thanks!

narendly (approved):

richardstartin:
@junkaixue the build has passed here, I'm happy for you to merge this whenever you are ready

Issue #1990, opened 2022-03-18
Race condition between instance drop and state transition message creation


Jackie-Jiang:
### Describe the bug
There is a potential race condition when an instance is dropped and a state transition message is created at the same time. The state transition message can recreate the `INSTANCE/<instanceId>` ZK directory, and cause instance show up in the cluster again, but without instance config.

### To Reproduce
Remove an instance from the IdealState and drop it at the same time.

### Expected behavior
The state transition message should not create the ZK directory if it does not exist.

junkaixue:
Thanks @Jackie-Jiang. Yeah, this is an known issue that ZK set will automatically create the folder if it does not exists. We may fix it with updater for history update.

PR #2073, opened 2022-04-28
Fix race condition between instance drop and participant history update

qqu0127:
### Issues

Fix https://github.com/apache/helix/issues/1990

### Description

When an instance is dropped while its history is updated, it's possible the history path is recreated and instance appear in cluster again. 
This PR address this issue by fixing the history update logic. On participant disconnect, we update the path if and only if the node path still exists (not always set). The creation logic stays the same, apparently we still want to have the history node.

### Tests

New test in TestParticipantManager

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
@@ -351,7 +351,7 @@ public class ParticipantManager {
 
     ParticipantHistory history = getHistory();
     history.reportOnline(_sessionId, _manager.getVersion());
-    persistHistory(history);
+    persistHistory(history, false);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
```diff
@@ -514,7 +515,7 @@ public void disconnect() {
     try {
       ParticipantHistory history = getHistory();
       history.reportOffline();
-      persistHistory(history);
+      persistHistory(history, true);
```
junkaixue, line 518: Why this is true? If disconnect, we are giving up the node and if instance has been dropped between zk connection disconnect and this call, we will still create that.

It would be meaning less if there was a path does not exist and update this path.

junkaixue (commented):
You are missing the one update by controller for updateOfflineInstanceHistory in BaseControllerDataProvider

helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java
qqu0127: Chatted offline, here skipOnEmptyPath=true is the expected behavior.
Updated for BaseControllerDataProvider.

junkaixue (approved):
lgtm

qqu0127:
This PR is ready to merge, thanks @junkaixue for review and approval!
Commit message:

Fix race condition between instance drop and participant history update

PR #2082, opened 2022-05-05
Add TTL mode to async create API in ZkClient

### LINKED ISSUE #2081: API methods in zookeeper-api don't support extended types
Zookeeper versions 3.5.3+ support some extended types for creating ZNodes, namely CONTAINER, PERSISTENT_WITH_TTL, and PERSISTENT_SEQUENTIAL_WITH_TTL. However, API methods that can support these types haven't been added to zookeeper-api module.

Add API methods for Zookeeper extended types to zookeeper-api.

rabashizade:
### Issues

Fixes #2081 

### Description

This PR adds methods that support creating persistent nodes with TTL asynchronously to ZkClient.

### Tests

The changes have been tested locally, but since Zookeeper server makes no guarantee on exactly when the Container and TTL nodes are deleted (see [here](https://zookeeper.apache.org/doc/r3.5.9/zookeeperProgrammers.html#TTL+Nodes)), the tests become flaky and as a result, are not added to this PR.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -85,6 +85,7 @@ import org.slf4j.LoggerFactory;
 public class ZkClient implements Watcher {
   private static final Logger LOG = LoggerFactory.getLogger(ZkClient.class);
 
+  public static final long TTL_NOT_SET = -1L;
   private static final long MAX_RECONNECT_INTERVAL_MS = 30000; // 30 seconds
 
   // If number of children exceeds this limit, getChildren() should not retry on connection loss.
@@ -1937,10 +1938,10 @@ public class ZkClient implements Watcher {
           new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
       return;
     }
-    doAsyncCreate(path, data, mode, startT, cb, parseExpectedSessionId(datat));
+    doAsyncCreate(path, data, mode, TTL_NOT_SET, startT, cb, parseExpectedSessionId(datat));
   }
 
-  private void doAsyncCreate(final String path, final byte[] data, final CreateMode mode,
+  private void doAsyncCreate(final String path, final byte[] data, final CreateMode mode, long ttl,
       final long startT, final ZkAsyncCallbacks.CreateCallbackHandler cb, final String expectedSessionId) {
     try {
       retryUntilConnected(() -> {
@@ -1949,17 +1950,31 @@ public class ZkClient implements Watcher {
                 GZipCompressionUtil.isCompressed(data)) {
               @Override
               protected void doRetry() {
-                doAsyncCreate(path, data, mode, System.currentTimeMillis(), cb, expectedSessionId);
+                doAsyncCreate(path, data, mode, ttl, System.currentTimeMillis(), cb, expectedSessionId);
               }
-            });
+            }, ttl);
         return null;
       });
     } catch (RuntimeException e) {
       // Process callback to release caller from waiting
       cb.processResult(KeeperException.Code.APIERROR.intValue(), path,
-          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null, null);
       throw e;
     }
+  }
+
+  public void asyncCreate(final String path, Object datat, final CreateMode mode, long ttl,
+      final ZkAsyncCallbacks.CreateCallbackHandler cb) {
+    final long startT = System.currentTimeMillis();
+    final byte[] data;
+    try {
+      data = (datat == null ? null : serialize(datat, path));
+    } catch (ZkMarshallingError e) {
+      cb.processResult(KeeperException.Code.MARSHALLINGERROR.intValue(), path,
+          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null, null);
+      return;
+    }
+    doAsyncCreate(path, data, mode, ttl, startT, cb, parseExpectedSessionId(datat));
   }
 
   // Async Data Accessors
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
@@ -23,6 +23,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.helix.zookeeper.zkclient.metric.ZkClientMonitor;
 import org.apache.zookeeper.AsyncCallback;
+import org.apache.zookeeper.AsyncCallback.Create2Callback;
 import org.apache.zookeeper.AsyncCallback.DataCallback;
 import org.apache.zookeeper.AsyncCallback.StatCallback;
 import org.apache.zookeeper.AsyncCallback.StringCallback;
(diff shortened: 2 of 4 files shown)


helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
```diff
@@ -95,6 +107,17 @@
    */
   boolean[] createChildren(List<String> paths, List<T> records, int options);
 
+  /**
+   * Use it when creating children under a parent node. This will use async api for better
+   * performance. If the child already exists it will return false.
```
narendly, line 112: the child (single) or children (all of them)?

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -61,6 +61,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.helix.zookeeper.zkclient.ZkClient.DEFAULT_TTL;
```
narendly, line 64: Usually we try to avoid static imports. ZkClient.DEFAULT_TTL should work?

helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
```diff
@@ -43,6 +43,18 @@
    */
   boolean create(String path, T record, int options);
 
+  /**
```
narendly, line 46: General comment:

Good work. This is a very large PR - I think instead of trying to include all changes in one single PR, it would help you and the reviewers to split this up. This will help us:
- Enable modular development
- Facilitate review process (faster turnaround)
- Better quality of reviews

One way I could think of off the top of my head is to first focus on the classes at the lowest level (e.g. ZkConnection, AsyncCallback, etc.) since we need to get those right first. Then move up a layer (ZkClient). And then another layer - (FederatedZkClient, DedicatedZkClient, etc. - these are all similar, so once we get one right, we probably get the rest right). And next, cover BaseDataAccessor and the like.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
rabashizade: Thank you for the suggestion. Since I ended up breaking up the PR into multiple smaller ones, this variable doesn't show up in this PR anymore. But I'll import it the way you suggested in the next PRs.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -111,12 +112,17 @@ protected void recordFailure(int rc, String path, ZkAsyncCallMonitorContext moni
     }
   }
 
-  public static class CreateCallbackHandler extends DefaultCallback implements StringCallback {
+  public static class CreateCallbackHandler extends DefaultCallback implements StringCallback, Create2Callback {
```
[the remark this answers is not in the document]
rabashizade: The reason is the callback type in Zookeeper which supports TTL mode is of type `Create2Callback` (see [here](https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/main/java/org/apache/zookeeper/ZooKeeper.java#L1602)). So I either had to add another `CallbackHandler` type that implements `Create2Callback`, or add `Create2Callback` to `CreateCallbackHandler` to support async create operations with TTL support. Adding a new type resulted in more duplicate code of the same methods with different argument types downstream, so I decided to just have `CreateCallbackHandler` implement `Create2Callback` instead.

helix-core/src/main/java/org/apache/helix/BaseDataAccessor.java
rabashizade: Thank you for the suggestion. I broke it up into multiple PRs. In this one, I just add TTL mode to async create API. In the next, I add it to sync API in `ZkClient` and `ZkConnection`, then to `RealmAwareZkClient` and its descendants, and finally to `BaseDataAcessors`.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -111,12 +112,17 @@ protected void recordFailure(int rc, String path, ZkAsyncCallMonitorContext moni
...
     @Override
     public void processResult(int rc, String path, Object ctx, String name) {
       callback(rc, path, ctx);
     }
 
+    @Override
+    public void processResult(int rc, String path, Object ctx, String name, Stat stat) {
```
desaikomal, line 122: what is 'stat' object used for??

desaikomal (commented):
For adding TTL, don't you need to add some tests?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
rabashizade: It's a Zookeeper data structure used for returning metadata about a ZNode. This particular method is part of the `Create2Callback` interface that has to be implemented.

rabashizade:
> For adding TTL, don't you need to add some tests?

This is a valid concern. However, as I explained in the PR description, the changes have been tested locally but since Zookeeper server makes no guarantee on exactly when the TTL nodes are deleted (see [here](https://zookeeper.apache.org/doc/r3.5.9/zookeeperProgrammers.html#TTL+Nodes)), the tests become flaky and as a result, are not added to this PR. I can add the tests, but we either need to wait a long time to be sure nodes are deleted, or we will see random test failures in the CI workflow.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1942,22 +1943,61 @@ public void asyncCreate(final String path, Object datat, final CreateMode mode,
...
+                });
+        return null;
+      });
+    } catch (RuntimeException e) {
+      // Process callback to release caller from waiting
+      cb.processResult(KeeperException.Code.APIERROR.intValue(), path,
+          new ZkAsyncCallMonitorContext(_monitor, startT, 0, false), null);
+      throw e;
+    }
+  }
```
narendly, line 1967: It seems that this is just a formatting change. Am I missing anything here?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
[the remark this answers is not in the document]
narendly: Makes sense. I was expecting to see a new `Create2CallbackHandler`, but since we're not doing much other than recording some stats in the monitor, I'm fine with this.

narendly (approved):
Great work! Thanks for breaking up the PRs - much easier to review.

Please consider addressing the comments and let me know if any of them are unclear.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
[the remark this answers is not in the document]
rabashizade: Thank you for the suggestion, it makes sense and I made the change.

rabashizade:

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -85,6 +86,7 @@
 public class ZkClient implements Watcher {
   private static final Logger LOG = LoggerFactory.getLogger(ZkClient.class);
 
+  public static final long DEFAULT_TTL = -1L;
```
narendly, line 89: Nit: Would `TTL_NOT_SET` be a more descriptive name for this? Also, I see this is a `public` constant - are you planning to use this in other classes?

rabashizade: Yes, that's a better name. I'll change it.
Also yes, it will be used in other classes such as BaseDataAccessors and ZkClient implementations. I was thinking of creating a separate class for this constant, but since it was only one variable I thought I'd just put it in ZkClient. Let me know if a separate constants class is preferred and I can add it.

Issue #2085, opened 2022-05-09
Bug fix and cleanup for bump-up.command 


qqu0127:
**Is your feature request related to a problem? Please describe.**
helix-view-cluster module is missing in bump-up.command thus causing development build failure.
Also there are lots of duplicated code, some code refactor will be needed.

**Describe the solution you'd like**
Bug fix and cleanup for bump-up.command

PR #2086, opened 2022-05-09
Refactor bump-up.command script and add helix-view-cluster

qqu0127:
### Issues

Fix https://github.com/apache/helix/issues/2085

### Description

Code cleanup and refactoring in bump-up.command, also added helix-view-cluster to it. There shouldn't be any functional change.

### Tests

Manually tested the script with internal lib.

### Documentation (Optional)

### Commits

### Code Quality


bump-up.command
```diff
@@ -19,6 +19,30 @@
 # under the License.
 #
 
+MODULES=("metrics-common" "metadata-store-directory-common" "zookeeper-api" "helix-common" "helix-core" \
+"helix-admin-webapp" "helix-front" "helix-rest" "helix-lock" "helix-view-aggregator" "helix-agent")
+
+update_pom_version() {
+  pom=$1
+  echo "bump up $pom"
+  sed -i "s/${version}/${new_version}/g" $pom
```
narendly, line 28: Thanks for refactoring! Another nagging issue we've had with this script is that if `sed` isn't able to locate the correct `version`, it just fails out, and those who aren't familiar with the way bump-up.command would get stuck. 

I think it might be helpful to add some error log based on the exit code - any thoughts?

narendly (approved):

bump-up.command
qqu0127: Hey @narendly , thanks for the comment. 
Not sure if I understand it correctly, but it seems `sed` doesn't complain if the pattern is not found. We actually need to do extra work to make it error out in such case https://unix.stackexchange.com/questions/55810/is-there-an-option-to-make-sed-fail-if-pattern-not-found

qqu0127: Added a check with `grep` instead since it's already there and is a smaller change.

junkaixue: For Mac, sed may not work.

qqu0127: Thanks @junkaixue  I believe this is existing behavior though.
I updated the command to make it work on both Mac and linux, referring to https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux

junkaixue:
@qqu0127 ready for merging?

Issue #2098, opened 2022-05-16
Helix should use user passed in connectionConfig to connect ZK


xyuanlu:
### Describe the bug
Helix is reading MultiZK config from jvm only when initiating ZkHelixManager. Helix should also read from user passed in connectionConfig.

PR #2099, opened 2022-05-16
Allow using passed in connection config when reading cloud config

xyuanlu:
### Issues

#2098

### Description

Helix is reading MultiZK config from jvm only when initiating ZkHelixManager. Helix should also read from user passed in connectionConfig.

### Tests

TestMultiZkConectionConfig

CI is failing on a constant failing test "TestClusterAccessor". The test is fixed in another PR.
Test failed: testGetClusters(org.apache.helix.rest.server.TestClusterAccessor) Time elapsed: 0.307 s <<< FAILURE!

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
@@ -54,7 +54,7 @@ public final class HelixPropertyFactory {
    * Clients may override these values.
    */
   public HelixManagerProperty getHelixManagerProperty(String zkAddress, String clusterName) {
-    CloudConfig cloudConfig = getCloudConfig(zkAddress, clusterName);
+    CloudConfig cloudConfig = getCloudConfig(zkAddress, clusterName, null);
     Properties properties = new Properties();
     try {
       InputStream stream = Thread.currentThread().getContextClassLoader()
@@ -80,7 +80,8 @@ public final class HelixPropertyFactory {
    * @param clusterName
    * @return
    */
-  public static CloudConfig getCloudConfig(String zkAddress, String clusterName) {
+  public static CloudConfig getCloudConfig(String zkAddress, String clusterName,
+      RealmAwareZkClient.RealmAwareZkConnectionConfig realmAwareZkConnectionConfig) {
     CloudConfig cloudConfig;
     RealmAwareZkClient dedicatedZkClient = null;
     try {
@@ -88,12 +89,14 @@ public final class HelixPropertyFactory {
         // If the multi ZK config is enabled or zkAddress is null, use realm-aware mode with
         // DedicatedZkClient
         try {
-          RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig =
-              new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder()
-                  .setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM)
-                  .setZkRealmShardingKey("/" + clusterName).build();
+          if (realmAwareZkConnectionConfig == null) {
+            realmAwareZkConnectionConfig =
+                new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder()
+                    .setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM)
+                    .setZkRealmShardingKey("/" + clusterName).build();
+          }
           dedicatedZkClient =
-              DedicatedZkClientFactory.getInstance().buildZkClient(connectionConfig);
+              DedicatedZkClientFactory.getInstance().buildZkClient(realmAwareZkConnectionConfig);
         } catch (IOException | InvalidRoutingDataException e) {
           throw new HelixException("Not able to connect on multi-ZK mode!", e);
         }
--- a/helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
@@ -80,6 +80,10 @@ public final class HelixPropertyFactory {
    * @param clusterName
    * @return
    */
+
+  public static CloudConfig getCloudConfig(String zkAddress, String clusterName) {
+    return getCloudConfig(zkAddress, clusterName, null);
+  }
   public static CloudConfig getCloudConfig(String zkAddress, String clusterName,
       RealmAwareZkClient.RealmAwareZkConnectionConfig realmAwareZkConnectionConfig) {
     CloudConfig cloudConfig;
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
@@ -282,7 +282,8 @@ public class ZKHelixManager implements HelixManager, IZkStateListener {
     // read cloud config from ZK and set cloudConfig in HelixManagerProperty
     _helixManagerProperty = helixManagerProperty;
     _helixManagerProperty.getHelixCloudProperty().populateFieldsWithCloudConfig(
-        HelixPropertyFactory.getCloudConfig(_zkAddress, _clusterName));
+        HelixPropertyFactory.getCloudConfig(_zkAddress, _clusterName,
+            helixManagerProperty.getZkConnectionConfig()));
 
     /**
      * use system property if available
--- a/helix-core/src/test/java/org/apache/helix/integration/manager/ClusterControllerManager.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/manager/ClusterControllerManager.java
@@ -19,6 +19,7 @@ package org.apache.helix.integration.manager;
  * under the License.
  */
 
+import org.apache.helix.HelixManagerProperty;
 import org.apache.helix.InstanceType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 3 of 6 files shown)


helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -80,20 +80,23 @@ public HelixManagerProperty getHelixManagerProperty(String zkAddress, String clu
    * @param clusterName
    * @return
    */
-  public static CloudConfig getCloudConfig(String zkAddress, String clusterName) {
+  public static CloudConfig getCloudConfig(String zkAddress, String clusterName,
+      RealmAwareZkClient.RealmAwareZkConnectionConfig realmAwareZkConnectionConfig) {
```
desaikomal, line 84: Is there a possibility that the supplied realmAwareZkConnectionConfig doesn't have everything populated as the default builder?

xyuanlu: TFTR.
It is possible, if the needed zk routing data is not provided, it will fall back to reading system config.
in RealAwareZkClient:
```
static MetadataStoreRoutingData getMetadataStoreRoutingData(
      RealmAwareZkConnectionConfig connectionConfig) throws InvalidRoutingDataException {
    String routingDataSourceEndpoint = connectionConfig.getRoutingDataSourceEndpoint();
    if (routingDataSourceEndpoint == null || routingDataSourceEndpoint.isEmpty()) {
      // If endpoint is not given explicitly, use HTTP and the endpoint set in System Properties
      return RoutingDataManager.getInstance().getMetadataStoreRoutingData();
  }

where in 
  public MetadataStoreRoutingData getMetadataStoreRoutingData() throws InvalidRoutingDataException {
    if (_defaultMsdsEndpoint == null || _defaultMsdsEndpoint.isEmpty()) {
      throw new IllegalStateException(
          "HttpRoutingDataReader was unable to find a valid MSDS endpoint String in System "
              + "Properties!");
    }
    return getMetadataStoreRoutingData(RoutingDataReaderType.HTTP, _defaultMsdsEndpoint);
  }

'''

File: helix-core/src/test/java/org/apache/helix/integration/manager/ClusterControllerManager.java [hunk 2]
```diff
@@ -37,6 +38,10 @@ public ClusterControllerManager(String zkAddr, String clusterName, String contro
     super(zkAddr, clusterName, controllerName, InstanceType.CONTROLLER);
   }
 
+  public ClusterControllerManager(String clusterName, HelixManagerProperty helixManagerProperty) {
```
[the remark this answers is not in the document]
xyuanlu: Tests in TestMultiZkConectionConfig do not have zkRouting env config set up and we are building helixManagerProperty to connect to ZK. The connection will fail when failing to read the Routing ZK.

```
    RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder connectionConfigBuilder =
        new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder();
    RealmAwareZkClient.RealmAwareZkConnectionConfig validZkConnectionConfig =
        connectionConfigBuilder
            .setRoutingDataSourceEndpoint(_msdsEndpoint + "," + ZK_PREFIX + ZK_START_PORT)
            .setRoutingDataSourceType(RoutingDataReaderType.HTTP_ZK_FALLBACK.name())
            .setZkRealmShardingKey("/" + clusterName).build();
    HelixManagerProperty.Builder propertyBuilder = new HelixManagerProperty.Builder()
```

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
qqu0127, line 84: This is a public method, could we maintain the original interface while adding a new one?

qqu0127 (changes requested):

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
xyuanlu: Good point. Updated.

desaikomal (commented):
thanks for adding tests. LGTM

xyuanlu:
This change is ready to be merged, approved by @desaikomal //Thanks
commit message:
Allow using passed in connection config when reading cloud config

desaikomal (approved):
Ready for check-in.

qqu0127 (approved):
LGTM, thanks for addressing the comments

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
```diff
@@ -80,20 +80,27 @@ public HelixManagerProperty getHelixManagerProperty(String zkAddress, String clu
...
     try {
       if (Boolean.getBoolean(SystemPropertyKeys.MULTI_ZK_ENABLED) || zkAddress == null) {
         // If the multi ZK config is enabled or zkAddress is null, use realm-aware mode with
         // DedicatedZkClient
         try {
-          RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig =
-              new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder()
-                  .setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM)
-                  .setZkRealmShardingKey("/" + clusterName).build();
+          if (realmAwareZkConnectionConfig == null) {
```
NealSun96, line 96: I'm not sure if this is the right pattern. I would do
```
public static CloudConfig getCloudConfig(String zkAddress, String clusterName) {
    realmAwareZkConnectionConfig =
                new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder()
                    .setRealmMode(RealmAwareZkClient.RealmMode.SINGLE_REALM)
                    .setZkRealmShardingKey("/" + clusterName).build();
    return getCloudConfig(zkAddress, clusterName, realmAwareZkConnectionConfig);
  }
```
Instead of a null check here.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -282,7 +282,8 @@ public ZKHelixManager(String clusterName, String instanceName, InstanceType inst
     // read cloud config from ZK and set cloudConfig in HelixManagerProperty
     _helixManagerProperty = helixManagerProperty;
     _helixManagerProperty.getHelixCloudProperty().populateFieldsWithCloudConfig(
-        HelixPropertyFactory.getCloudConfig(_zkAddress, _clusterName));
+        HelixPropertyFactory.getCloudConfig(_zkAddress, _clusterName,
+            helixManagerProperty.getZkConnectionConfig()));
```
NealSun96, line 286: Question: could you remind me why we make the change in the first place?  What's in `helixManagerProperty.getZkConnectionConfig()`?

helix-core/src/test/java/org/apache/helix/integration/manager/ClusterControllerManager.java
NealSun96, line 41: If this is for testing purpose, can we add @VisibleForTest or another similar decorator?

NealSun96 (changes requested):

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
xyuanlu: TFTR. When we add an API to allow user to pass in `RealmAwareZkConnectionConfig` and when user pass in a null obj, shouldn't we create a default one instead of using the null?

helix-core/src/test/java/org/apache/helix/integration/manager/ClusterControllerManager.java
xyuanlu: We need this note even when the class is defined test module? I don't think this class has any non-test usage,

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
xyuanlu: helixManagerProperty.getZkConnectionConfig() returns an object of `RealmAwareZkConnectionConfig`.
I our current Helix use case, the user is defining the multiZkRouting endpoint in the `RealmAwareZkConnectionConfig`. However, the original `getCloudConfig` create a zkClient only using the multiZkRouting in java env variable. When user pass in the config but not set it in env variable, we will fail to create zkClient even when user passed in correct config.

helix-core/src/test/java/org/apache/helix/integration/manager/ClusterControllerManager.java
NealSun96: Didn't notice, my bad.

helix-core/src/main/java/org/apache/helix/HelixPropertyFactory.java
NealSun96: I don't think a user should be allowed to pass in the null at all. If they don't have a config, they should use 
```
getCloudConfig(String zkAddress, String clusterName)
```
which would create a default for them. 

If they have a config, they should then call the new function, and they are responsible for having a correct config.

xyuanlu: From our offline discussion, we only need to create a default RealmAwareZkConnectionConfig when it is in multi ZK mode. Do it this way will initiate a  RealmAwareZkConnectionConfig object regardless we are in multiZk or not.
Let's keep the current approach.

xyuanlu: From our offline discussion, we only need to create a default RealmAwareZkConnectionConfig when it is in multi ZK mode. Do it this way will initiate a  RealmAwareZkConnectionConfig object regardless we are in multiZk or not.
Let's keep the current approach.

NealSun96 (approved):

PR #2090, opened 2022-05-11
Add TTL and Container modes to sync create API in ZkClient and ZkConnection

rabashizade:
### Issues

- [x] My PR addresses the following Helix issues and references them in the PR description:

Fixes #2081 

### Description

- [x] Here are some details about my PR, including screenshots of any UI changes:

This PR adds methods that support creating persistent nodes with Container and TTL modes synchronously to ZkClient and ZkConnection classes.

### Tests

- [x] The following tests are written for this issue:

In TestRawZkClient, the following tests have been added:
1. testUnimplementedTypes()
2. testCreatePersistentWithTTL()
3. testCreatePersistentSequentialWithTTL()
4. testCreateContainer()

However, these tests verify that node creation happens correctly. They don't test deletion of the TTL and Container nodes by Zookeeper server. These features have been tested locally, but since Zookeeper server makes no guarantee on exactly when the Container and TTL nodes are deleted (see [here](https://zookeeper.apache.org/doc/r3.5.9/zookeeperProgrammers.html#TTL+Nodes)), those tests become flaky and as a result, are not added to this PR.

- The following is the result of the "mvn test" command on the appropriate module:

(If CI test fails due to known issue, please specify the issue and test PR locally. Then copy & paste the result of "mvn test" to here.)

### Changes that Break Backward Compatibility (Optional)

- My PR contains changes that break backward compatibility or previous assumptions for certain methods or API. They include:

(Consider including all behavior changes for public methods or API. Also include these changes in merge description so that other developers are aware of these changes. This allows them to make relevant code changes in feature branches accounting for the new method/API behavior.)

### Documentation (Optional)

- In case of new functionality, my PR adds documentation in the following wiki page:

(Link the GitHub wiki you added)

### Commits

- My commits all reference appropriate Apache Helix GitHub issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Code Quality

- My diff has been formatted using helix-style.xml 
(helix-style-intellij.xml if IntelliJ IDE is used)

PR #2103, opened 2022-05-18
Add TTL and Container modes to RealmAwareZkClient and its implementations

### LINKED ISSUE #2081: API methods in zookeeper-api don't support extended types

rabashizade:
### Issues

Fixes #2081 

### Description

New API methods were added to `ZkClient` in #2082 and #2090 to support creation of ZNodes in TTL and Container modes. This PR builds on those API to add this support to `RealmAwareZkClient` and its implementations.

### Tests

Three new tests are added to `RealmAwareZkClientFactoryTestBase` (which is extended by `TestSharedZkClient` and `TestDedicatedZkClient`) and `TestFederatedZkClient`:
1. `testRealmAwareZkClientCreateContainer()`
2. `testRealmAwareZkClientCreateSequentialWithTTL()`
3. `testRealmAwareZkClientCreateWithTTL()`

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
@@ -175,9 +175,33 @@ public interface RealmAwareZkClient {
 
   void createPersistent(String path, Object data, List<ACL> acl);
 
+  void createPersistentWithTTL(String path, long ttl);
+
+  void createPersistentWithTTL(String path, boolean createParents, long ttl);
+
+  void createPersistentWithTTL(String path, boolean createParents, List<ACL> acl, long ttl);
+
+  void createPersistentWithTTL(String path, Object data, long ttl);
+
+  void createPersistentWithTTL(String path, Object data, List<ACL> acl, long ttl);
+
   String createPersistentSequential(String path, Object data);
 
   String createPersistentSequential(String path, Object data, List<ACL> acl);
+
+  String createPersistentSequentialWithTTL(String path, Object data, long ttl);
+
+  String createPersistentSequentialWithTTL(String path, Object data, List<ACL> acl, long ttl);
+
+  void createContainer(String path);
+
+  void createContainer(String path, boolean createParents);
+
+  void createContainer(String path, boolean createParents, List<ACL> acl);
+
+  void createContainer(String path, Object data);
+
+  void createContainer(String path, Object data, List<ACL> acl);
 
   void createEphemeral(final String path);
 
@@ -189,7 +213,12 @@ public interface RealmAwareZkClient {
 
   String create(final String path, Object data, final CreateMode mode);
 
+  String create(final String path, Object data, final CreateMode mode, long ttl);
+
   String create(final String path, Object datat, final List<ACL> acl, final CreateMode mode);
+
+  String create(final String path, Object datat, final List<ACL> acl, final CreateMode mode,
+      long ttl);
 
   void createEphemeral(final String path, final Object data);
 
@@ -244,6 +273,9 @@ public interface RealmAwareZkClient {
   Stat writeDataGetStat(final String path, Object datat, final int expectedVersion);
 
   void asyncCreate(final String path, Object datat, final CreateMode mode,
+      final ZkAsyncCallbacks.CreateCallbackHandler cb);
+
+  void asyncCreate(final String path, Object datat, final CreateMode mode, long ttl,
       final ZkAsyncCallbacks.CreateCallbackHandler cb);
 
   void asyncSetData(final String path, Object datat, final int version,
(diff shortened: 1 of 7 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
```diff
@@ -175,10 +175,34 @@ void unsubscribeStateChanges(
 
   void createPersistent(String path, Object data, List<ACL> acl);
 
+  void createPersistentWithTTL(String path, long ttl);
```
desaikomal, line 178: this is not specific to your review but a generic design question. 
ZK in open source will keep adding new features. As a result, do we need to keep adding similar APIs and support?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
```diff
@@ -217,12 +279,23 @@ public String create(String path, Object data, CreateMode mode) {
     return create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, mode);
   }
 
+  @Override
+  public String create(String path, Object data, CreateMode mode, long ttl) {
+    return create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, mode, ttl);
+  }
+
   @Override
   public String create(String path, Object datat, List<ACL> acl, CreateMode mode) {
```
desaikomal, line 288: Another question: how do you prevent newer ZK client talking to older version ZK server and calling the above APIs?

desaikomal, line 288: Another question: how do you prevent newer ZK client talking to older version ZK server and calling the above APIs?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/api/client/RealmAwareZkClient.java
rabashizade: I think this is a case-by-case situation. I'd say ideally, yes, but for this particular set of API, we have known use-cases that need these features so that's why I started adding them. In fact, they have been available in open-source Zookeeper for quite a while now, but we hadn't added them until now.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/DedicatedZkClient.java
rabashizade: This is a good question, we can't really prevent them because we have no way of knowing the ZK's version, but ZK itself (whether older version or newer version without the extended types enabled as explained [here](https://zookeeper.apache.org/doc/current/zookeeperAdmin.html#sc_configuration)) will throw a `KeeperException.UnimplementedException`.

rabashizade: Please see the answer to the previous comment.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/SharedZkClient.java
```diff
@@ -189,6 +189,36 @@ public void createPersistent(String path, Object data, List<ACL> acl) {
     _innerSharedZkClient.createPersistent(path, data, acl);
   }
 
+  @Override
+  public void createPersistentWithTTL(String path, long ttl) {
```
mgao0, line 193: This comment applies to all the code below:
Why don't we call the same method with more parameters, instead we repeat the code?
Taking this as an example, can we do `createPersistentWithTTL(path, false, ttl)` here?

rabashizade: Thanks for the suggestion, yes that makes sense. I will make that change.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/SharedZkClient.java
```diff
@@ -402,6 +478,13 @@ public Stat writeDataGetStat(String path, Object datat, int expectedVersion) {
...
 
+  @Override
+  public void asyncCreate(String path, Object datat, CreateMode mode, long ttl,
+      ZkAsyncCallbacks.CreateCallbackHandler cb) {
+    checkIfPathContainsShardingKey(path);
+    _innerSharedZkClient.asyncCreate(path, datat, mode, ttl, cb);
+  }
+
   @Override
   public void asyncCreate(String path, Object datat, CreateMode mode,
```
mgao0, line 489: Can we use ZkClient.TTL_NOT_SET to simplify this implementation?

rabashizade: Yes, made the change.

mgao0 (approved):
Thanks for addressing the comments.

rabashizade:
This PR is ready to be merged, approved by @mgao0.

Issue #2000, opened 2022-03-30
Batch enabling/disable instance with type and reason


xyuanlu:
**Is your feature request related to a problem? Please describe.**
Having a batching enable/disable would reduce number of ZK write. It is also helps user to solve some race conditions as all instance enabling/disabling happens at one ZNode. 

Helix already have a batch instance enable/disable feature but it is inactive now. We would like to make this feature active now as the dependency is resolved. Helix should also allow user to set disabled type and reason for these instances.

PR #2106, opened 2022-05-19
Clean format for disabled instances in ClusterConfig

xyuanlu:
### Issues

#2000 

### Description

We are reverting the batch mode for enable and disable instances. This pr cleans the DISABLED_INSTANCES in cluster Config.

### Tests

testControllerCleanUpClusterConfig

Rerun CI failed test
```
END TestP2PNoDuplicatedMessage at Mon Jun 06 10:54:58 PDT 2022
AfterClass: TestP2PNoDuplicatedMessage called.
************ SYSTEM Physical Memory:34359738368
************ total memory:200 free memory:43
---------- Test Class TestP2PNoDuplicatedMessage thread leakage detected! ---------------

===============================================
Default Suite
Total tests run: 2, Failures: 0, Skips: 0
===============================================

```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -28,14 +28,12 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.helix.HelixConstants;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixException;
@@ -63,7 +61,6 @@ import org.apache.helix.model.PauseSignal;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
 import org.apache.helix.task.TaskConstants;
-import org.apache.helix.util.ConfigStringUtil;
 import org.apache.helix.util.HelixUtil;
 import org.apache.helix.util.InstanceValidationUtil;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
@@ -246,10 +243,10 @@ public class BaseControllerDataProvider implements ControlContextProvider {
       refreshedType.add(HelixConstants.ChangeType.CLUSTER_CONFIG);
       // TODO: This is a temp function to clean up incompatible batched disabled instances format.
       // Remove in later version.
-      if (needReformatBatchedDIsabledInstance(_clusterConfig) && updateBatchDisableFormat(
-          accessor)) {
+      if (needCleanUpBatchedDisabledInstance(_clusterConfig.getRecord())
+          && cleanBatchDisableMapField(accessor)) {
         LogUtil.logInfo(logger, getClusterEventId(), String
-            .format("Clean ClusterConfig change for cluster %s, pipeline %s", _clusterName,
+            .format("Clean ClusterConfig mapField for cluster %s, pipeline %s", _clusterName,
                 getPipelineName()));
       }
       refreshAbnormalStateResolverMap(_clusterConfig);
@@ -260,10 +257,10 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     }
   }
 
-  // TODO: This function is used to clean up incompatible batched disabled instances format for
-  // "DISABLED_INSTANCES" introduced in 1.0.3.0. This temp change shoul dbe reverted after 1.0.5.0 \
+  // TODO: This function is used to clean up batched disabled instances for
+  // "DISABLED_INSTANCES" introduced in 1.0.3.0. This temp change should be reverted after 1.0.5.0 \
   // or later version.
-  private boolean updateBatchDisableFormat(final HelixDataAccessor accessor) {
+  private boolean cleanBatchDisableMapField(final HelixDataAccessor accessor) {
     boolean successful =
         accessor.updateProperty(accessor.keyBuilder().clusterConfig(), new DataUpdater<ZNRecord>() {
           @Override
@@ -272,12 +269,17 @@ public class BaseControllerDataProvider implements ControlContextProvider {
               throw new HelixException(
                   "Cluster: " + _clusterConfig.getClusterName() + ": cluster config is null");
             }
-            return updateBatchDisableFormatHelper(new ClusterConfig(currentData)).getRecord();
+            ZNRecord newRecord = new ZNRecord(currentData);
+            String batchDisabledInstanceMapFieldKey =
+                ClusterConfig.ClusterConfigProperty.DISABLED_INSTANCES.name();
+            if (needCleanUpBatchedDisabledInstance(currentData)) {
+              newRecord.getMapFields().remove(batchDisabledInstanceMapFieldKey);
+
+            }
+            return newRecord;
           }
         }, null);
-    if (successful) {
-      _clusterConfig = updateBatchDisableFormatHelper(_clusterConfig);
-    } else {
+    if (!successful) {
       LogUtil.logError(logger, getClusterEventId(), String
           .format("Failed to clean ClusterConfig change for cluster %s, pipeline %s", _clusterName,
               getPipelineName()));
@@ -285,48 +287,9 @@ public class BaseControllerDataProvider implements ControlContextProvider {
     return successful;
   }
 
-  private ClusterConfig updateBatchDisableFormatHelper(ClusterConfig oldConfig) {
-    Map<String, String> disabledInstances = oldConfig.getDisabledInstances();
-    Map<String, String> disabledInstancesWithInfo = oldConfig.getDisabledInstancesWithInfo();
-
-    ClusterConfig newClusterConfig = new ClusterConfig(oldConfig.getRecord());
-    Map<String, String> newDisabledInstances =
-        new TreeMap<>(newClusterConfig.getDisabledInstances());
-    Map<String, String> newDisabledInstancesWithInfo =
-        new TreeMap<>(newClusterConfig.getDisabledInstancesWithInfo());
-
-    // iterate through all k,v pairs and check for string format.
-    for (Map.Entry<String, String> instanceInfo : disabledInstances.entrySet()) {
-      String instanceName = instanceInfo.getKey();
-      if (!StringUtils.isNumeric(instanceInfo.getValue())) {
-        newDisabledInstances.put(instanceName,
-            ConfigStringUtil.parseConcatenatedConfig(instanceInfo.getValue())
-                .get(ClusterConfig.ClusterConfigProperty.HELIX_ENABLED_DISABLE_TIMESTAMP.name()));
-        newDisabledInstancesWithInfo.put(instanceName, instanceInfo.getValue());
-      } else if (!disabledInstancesWithInfo.containsKey(instanceName)) {
-        newDisabledInstancesWithInfo.put(instanceName, ConfigStringUtil.concatenateMapping(
-            Collections.singletonMap(
-                ClusterConfig.ClusterConfigProperty.HELIX_ENABLED_DISABLE_TIMESTAMP.name(),
-                instanceInfo.getValue())));
-      }
-    }
-    newClusterConfig.setDisabledInstances(newDisabledInstances);
-    newClusterConfig.setDisabledInstancesWithInfo(newDisabledInstancesWithInfo);
-    return newClusterConfig;
-  }
-
-  private boolean needReformatBatchedDIsabledInstance(ClusterConfig clusterConfig) {
-    Map<String, String> disabledInstances = clusterConfig.getDisabledInstances();
-    Map<String, String> DisabledInstancesWithInfo = clusterConfig.getDisabledInstancesWithInfo();
-
-    // interate through all k,v pairs and check for string format.
-    for (Map.Entry<String, String> instanceInfo : disabledInstances.entrySet()) {
-      if (!StringUtils.isNumeric(instanceInfo.getValue()) || !DisabledInstancesWithInfo
-          .containsKey(instanceInfo.getKey())) {
-        return true;
-      }
-    }
-    return false;
+  private boolean needCleanUpBatchedDisabledInstance(ZNRecord record) {
+    return record.getMapFields()
+        .containsKey(ClusterConfig.ClusterConfigProperty.DISABLED_INSTANCES.name());
   }
 
   private void refreshIdealState(final HelixDataAccessor accessor,
@@ -335,8 +298,8 @@ public class BaseControllerDataProvider implements ControlContextProvider {
       _idealStateCache.refresh(accessor);
       refreshedType.add(HelixConstants.ChangeType.IDEAL_STATE);
     } else {
-      LogUtil.logInfo(logger, getClusterEventId(), String
-          .format("No ideal state change for %s cluster, %s pipeline", _clusterName,
+      LogUtil.logInfo(logger, getClusterEventId(),
+          String.format("No ideal state change for %s cluster, %s pipeline", _clusterName,
               getPipelineName()));
     }
   }
--- a/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
@@ -28,12 +28,14 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.helix.HelixConstants;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixException;
@@ -61,8 +63,11 @@ import org.apache.helix.model.PauseSignal;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.helix.model.StateModelDefinition;
 import org.apache.helix.task.TaskConstants;
+import org.apache.helix.util.ConfigStringUtil;
 import org.apache.helix.util.HelixUtil;
 import org.apache.helix.util.InstanceValidationUtil;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+import org.apache.helix.zookeeper.zkclient.DataUpdater;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -74,8 +79,7 @@ import org.slf4j.LoggerFactory;
  * This class will be moved to helix-common module in the future
  */
 public class BaseControllerDataProvider implements ControlContextProvider {
-  private static final Logger logger =
-      LoggerFactory.getLogger(BaseControllerDataProvider.class);
+  private static final Logger logger = LoggerFactory.getLogger(BaseControllerDataProvider.class);
 
   // We only refresh EV and TEV the very first time the cluster data cache is initialized
   private static final List<HelixConstants.ChangeType> _noFullRefreshProperty = Arrays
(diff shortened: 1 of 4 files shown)


helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -241,13 +244,87 @@ private void refreshClusterConfig(final HelixDataAccessor accessor,
...
       refreshedType.add(HelixConstants.ChangeType.CLUSTER_CONFIG);
+      // TODO: This is a temp function to clean up incompatible batched disabled instances format.
+      // Remove in later version.
+      if (checkBatchedDisabledInstanceFormat(_clusterConfig) && updateBatchDisableFormat(
+          accessor)) {
+        // read from zkz one more time
+        LogUtil.logInfo(logger, getClusterEventId(), String
+            .format("Clean ClusterConfig change for cluster %s, pipeline %s", _clusterName,
+                getPipelineName()));
+        _clusterConfig = accessor.getProperty(accessor.keyBuilder().clusterConfig());
```
junkaixue, line 255: Let's not read this again. Helix computation is based on a snapshot of all data for ZK clusters. If you read cluster config in the middle of the pipeline, then cluster config is a newer version of snapshot instead of original one.

My suggestion is just reformat whatever we need to update in memory. But do not read cluster config back. Let next round pipe line refresh to refresh cluster config since there should be such type of config change notification.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -241,13 +244,87 @@ private void refreshClusterConfig(final HelixDataAccessor accessor,
...
+      }
       refreshAbnormalStateResolverMap(_clusterConfig);
     } else {
-      LogUtil.logInfo(logger, getClusterEventId(), String.format(
-          "No ClusterConfig change for cluster %s, pipeline %s", _clusterName, getPipelineName()));
+      LogUtil.logInfo(logger, getClusterEventId(), String
```
junkaixue, line 259: Let's put it to debug level?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -241,13 +244,87 @@ private void refreshClusterConfig(final HelixDataAccessor accessor,
...
+  // or later version.
+  private boolean updateBatchDisableFormat(final HelixDataAccessor accessor) {
+    return accessor
+        .updateProperty(accessor.keyBuilder().clusterConfig(), new DataUpdater<ZNRecord>() {
+          @Override
+          public ZNRecord update(ZNRecord currentData) {
+            if (currentData == null) {
+              throw new HelixException(
+                  "Cluster: " + _clusterConfig.getClusterName() + ": cluster config is null");
+            }
```
qqu0127, line 276: Similar to Junkai's comment above, let's try to combine the read-and-write pattern into one. If the currentData is null OR the format is good, we can just keep it. In such way we can combine the two methods into one updater.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -241,13 +244,87 @@ private void refreshClusterConfig(final HelixDataAccessor accessor,
...
+            newClusterConfig.setDisabledInstances(newDisabledInstances);
+            newClusterConfig.setDisabledInstancesWithInfo(newDisabledInstancesWithInfo);
+            return newClusterConfig.getRecord();
+          }
+        }, null);
+  }
+
+  private boolean checkBatchedDisabledInstanceFormat(ClusterConfig clusterConfig) {
+    Map<String, String> disabledInstances = clusterConfig.getDisabledInstances();
+    Map<String, String> DisabledInstancesWithInfo = clusterConfig.getDisabledInstancesWithInfo();
```
qqu0127, line 316: nit, naming convention

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -241,13 +244,87 @@ private void refreshClusterConfig(final HelixDataAccessor accessor,
...
+              }
+            }
+            newClusterConfig.setDisabledInstances(newDisabledInstances);
+            newClusterConfig.setDisabledInstancesWithInfo(newDisabledInstancesWithInfo);
+            return newClusterConfig.getRecord();
+          }
+        }, null);
+  }
+
+  private boolean checkBatchedDisabledInstanceFormat(ClusterConfig clusterConfig) {
```
qqu0127, line 314: Can we have a better method name? (if we still need this method) "check" might be too generic and it's unclear what it really means.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -556,8 +633,8 @@ public Set<String> getInstancesWithTag(String instanceTag) {
    */
   public Set<String> getDisabledInstancesForPartition(String resource, String partition) {
     Set<String> disabledInstancesForPartition = new HashSet<>(_disabledInstanceSet);
-    if (_disabledInstanceForPartitionMap.containsKey(resource)
-        && _disabledInstanceForPartitionMap.get(resource).containsKey(partition)) {
+    if (_disabledInstanceForPartitionMap.containsKey(resource) && _disabledInstanceForPartitionMap
+        .get(resource).containsKey(partition)) {
```
qqu0127, line 637: nit, `.getOrDefault(resource, EMPTY_MAP)` could be cleaner.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -241,13 +244,87 @@ private void refreshClusterConfig(final HelixDataAccessor accessor,
...
+          @Override
+          public ZNRecord update(ZNRecord currentData) {
+            if (currentData == null) {
+              throw new HelixException(
+                  "Cluster: " + _clusterConfig.getClusterName() + ": cluster config is null");
+            }
+
+            ClusterConfig clusterConfig = new ClusterConfig(currentData);
+            Map<String, String> disabledInstances = clusterConfig.getDisabledInstances();
+            Map<String, String> DisabledInstancesWithInfo =
```
qqu0127, line 280: nit, naming convention

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
xyuanlu: If we combine the 2 method into one updater, we will need to write either way. I would prefer have it separate.

qqu0127: I think the problem with two separate steps is that, they are not transactional. Another concern is, the check is done on the local snapshot, but the modification is applied on the real zk data. 
Correct me if I understand wrong -- 
check(_clusterConfig) -> meanwhile clusterConfig is changed or removed -> updateBatchDisableFormat may fail or not needed

xyuanlu: This is correct, so we are checking the field again in updater

qqu0127: Why not doing the check once in the updater, using the source of truth data?

xyuanlu: This will still issue a write request.

qqu0127: That's true, we are not reducing the write, but that way we are having less read and completing the process in one step.

xyuanlu: Thanks for the suggestion, please correct me if I miss anything. 

If we do check and then write using update, then every time when we refresh cache for cluster config, we do one read to refresh, then check and finally read+write in updater if needed.
If we do check in updater, we will do one read to refresh and one read+write in updater no matter if we need to rewrite format. (we will write the cluster config back without any changes is most cases.)

qqu0127: Thanks for the clarification. I had an impression that if updater returns the same value as old data, there will be no write. This might be a false assumption I made.
Let's keep it this way if a write is always triggered, but it's an interesting topic how we can improve property update to reduce write.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -241,11 +244,85 @@ private void refreshClusterConfig(final HelixDataAccessor accessor,
...
+            }
+
+            ClusterConfig clusterConfig = new ClusterConfig(currentData);
+            Map<String, String> disabledInstances = clusterConfig.getDisabledInstances();
+            Map<String, String> disabledInstancesWithInfo =
+                clusterConfig.getDisabledInstancesWithInfo();
+
+            ClusterConfig newClusterConfig = new ClusterConfig(currentData);
+            Map<String, String> newDisabledInstances =
+                new TreeMap<>(newClusterConfig.getDisabledInstances());
```
rabashizade, line 285: Is there any particular reason you used `TreeMap` here? This map and `newDisabledInstancesWithInfo` are only used for put and get, isn't a `HashMap` better for that purpose?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -241,11 +244,85 @@ private void refreshClusterConfig(final HelixDataAccessor accessor,
...
+            for (Map.Entry<String, String> instanceInfo : disabledInstances.entrySet()) {
+              String instanceName = instanceInfo.getKey();
+              if (!StringUtils.isNumeric(instanceInfo.getValue())) {
+                newDisabledInstances.put(instanceName,
+                    ConfigStringUtil.parseConcatenatedConfig(instanceInfo.getValue()).get(
+                        ClusterConfig.ClusterConfigProperty.HELIX_ENABLED_DISABLE_TIMESTAMP
+                            .name()));
+                newDisabledInstancesWithInfo.put(instanceName, instanceInfo.getValue());
+              } else {
+                if (!disabledInstancesWithInfo.containsKey(instanceName)) {
```
rabashizade, line 299: Nit: can these two lines be reformatted as `} else if () {`?

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
xyuanlu: TFTR.
This list will be written to ZK. When SRE or other Helix user read the disabled list, having a sorted list is more human readable.

rabashizade: I see, this makes sense then. The only thing is, how often is it written to ZK and how large is it usually? I'm just trying to see if it might potentially slow down the pipeline when instances go offline very frequently. If this isn't a concern, then I can resolve the conversation.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
xyuanlu: This is a valid concern. This change will not add a lot load to pipeline because:
1. `refreshClusterConfig` will be a no-op if there is no clusterConfig change callback. clusterConfig change callback will only be triggered in a. the ever first pipeline run or b. there is actual update to clusterConfig.
2. If the format in clusterConfig do not need rewrite (due to one of the following reason: a. The format is clean to begin with, b. the format is already rewritten), `needReformatBatchedDIsabledInstance` will return false and there will be no write back to ZK.
3. This is a temporary clean up fix. This change will be rolled back in next+ version (probably 1.0.5.0).

rabashizade: You're right, I had confused this format correction with the general writing of disabled instances. This looks good to me then. Thanks for the explanation.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
qqu0127: Second to Junkai, the immediate _clusterConfig refresh may bring in other changes in the middle of the pipeline. It will become tricky to tell when a clusterConfig cache is refreshed. 
The format change isn't a substantial one, I don't think it will cause any new issue in the pipeline (assuming it was in a good state). Could you please confirm?

xyuanlu: TFTR.
Waged will read disabled time in BestPossibleComputeState. If we keep the uncleaned format, it will cause issue for the pipeline run.
I think the question is, is it ok for us to have several failed pipeline until the change in cluster config is picked up?

rabashizade: I think what Junkai is suggesting is instead of writing corrected data to ZK in `updateBatchDisableFormat()`, and then reading it back in `_clusterConfig` here, directly update `_clusterConfig` in `updateBatchDisableFormat()` and let the data written to ZK be picked up in `_clusterConfig` the next time pipeline is triggered. This way, we won't have the risk of pipeline failure due to incorrect disabled time format, or the risk of inconsistent `_clusterConfig` between pipeline stages.

xyuanlu: Thanks for the review.
Please correct me if I am wrong, if update failed then we should not update ` _clusterConfig ` ?

rabashizade: That's a good question, if we know the format is wrong and we fail to update ZK, do we want to keep the wrong format in cache as well? My personal opinion is that the cache should be fixed regardless, but I might be missing something. Regardless, I think you can check the result of the update and based on that update `_clusterConfig`. Would that cause duplicate work?

xyuanlu: In my understanding, the baseline is that whatever in cache should be the same as ZNode. If we failed to update ZNode and the format causing pipeline fail, then the pipeline fail and we try again.  
I am open to discussion.

rabashizade: In general, I agree that cache and ZK must have the same view. However, in this case what's the risk of updating cache ahead of ZK? It's not new information, but just fixing formatting. The same fix to ZK will be attempted the next time the pipeline is running. But even if we want to avoid version mismatch, the success of ZK update can be checked and then instead of reading from ZK again, the same changes can be applied to `_clusterConfig`.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -241,11 +244,89 @@ private void refreshClusterConfig(final HelixDataAccessor accessor,
...
+          public ZNRecord update(ZNRecord currentData) {
+            if (currentData == null) {
+              throw new HelixException(
+                  "Cluster: " + _clusterConfig.getClusterName() + ": cluster config is null");
+            }
+            return updateBatchDisableFormatHelper(new ClusterConfig(currentData)).getRecord();
+          }
+        }, null);
+    if (successful) {
+      _clusterConfig = updateBatchDisableFormatHelper(_clusterConfig);
```
rabashizade, line 279: Mostly curious, is it possible to avoid calling this method again? If not, I don't have any other concerns.

xyuanlu: TFTR. It should be OK to call this twice as the helper do not issue new IO.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
xyuanlu: Thanks for suggestion. Updated in follow up commit.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
```diff
@@ -241,21 +241,64 @@ private void refreshClusterConfig(final HelixDataAccessor accessor,
...
+          @Override
+          public ZNRecord update(ZNRecord currentData) {
+            if (currentData == null) {
+              throw new HelixException(
+                  "Cluster: " + _clusterConfig.getClusterName() + ": cluster config is null");
+            }
+            ZNRecord newRecord = new ZNRecord(currentData);
+            String batchDisabledInstanceMapFieldKey =
+                ClusterConfig.ClusterConfigProperty.DISABLED_INSTANCES.name();
+            if (needCleanUpBatchedDisabledInstance(currentData)) {
```
desaikomal, line 275: you are calling needCleanUpBatchedDisabledInstance() twice.
first on line:246 and here too..  any particular reason

helix-core/src/test/java/org/apache/helix/integration/controller/TestControllerLeadershipChange.java
```diff
@@ -74,6 +82,51 @@ public void afterClass() {
   }
 
   @Test
+  public void testControllerCleanUpClusterConfig() {
```
desaikomal, line 85: there are 3 testcases: 1. we don't have DISABLED_INSTANCES at all. before the pipeline and after the pipeline config should match as no update should have happened. 2. this test scenario.   3. original DISABLE_INSTANCE related test-case verifying that it remains same.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
xyuanlu: It is possible that there is a race condition that 2 pipeline both check clusterConfig that need update. By adding another check in updater, we reduce the race window and updater also garente that only one will write back.

helix-core/src/test/java/org/apache/helix/integration/controller/TestControllerLeadershipChange.java
xyuanlu: TFTR. I will add these 2 test cases after  #2134 is merged. (otherwise it is hard to test when we still have batch mode enabled and we write to DISABLED_INSTANCES)

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
desaikomal: in my mind, first condition doesn't help you. to avoid race, you do it here, should be sufficient.

helix-core/src/test/java/org/apache/helix/integration/controller/TestControllerLeadershipChange.java
desaikomal: makes sense.. but once #2134 is checked-in, you can add.

helix-core/src/main/java/org/apache/helix/controller/dataproviders/BaseControllerDataProvider.java
xyuanlu: I think the first if helps. If that first condition did not satisfy, we will safe a read of this updater.

desaikomal: in my mind, doing it unconditionally once is sufficient. reading/update as one step (method) is easier to reason through, but i am ok with this 2-step approach

helix-core/src/test/java/org/apache/helix/integration/controller/TestControllerLeadershipChange.java
xyuanlu: Added test and rebased.

desaikomal (approved):
LGTM..

mgao0 (approved):

qqu0127 (approved):
LGTM, please run and verify the failed tests locally

xyuanlu:
This change is ready to be merged. Approved by @mgao0 @desaikomal @mgao0 

Commit message:
**Clean disabled instances field in ClusterConfig**
We reverted enabling batch mode for instance enable/disable. This change hooks readData stage when controller starts in the first place and will clean up batch enabled field in cluster config. 
This change should be reverted after next next version. (1.0.5.0 +)

PR #2118, opened 2022-05-25
Improve efficiency of ClusterStatusMonitor and PerInstanceResourceMonitor

### LINKED ISSUE #2117: ClusterStatusMonitor is a bottleneck in BestPossibleStateCalcStage.process
This profile was taken from an Apache Pinot controller:
Screenshot 2022-05-25 at 09 53 04

About 18% of the samples were in `ClusterStatusMonitor.setPerInstanceResourceStatus` (zoomed in)

Screenshot 2022-05-25 at 09 53 22

The root causes are:

* inefficient hashCode implementation in `PerInstanceResourceMonitor$BeanName`
* doing twice the work in `ClusterStatusMonitor.setPerInstanceResourceStatus` because `containsKey` is called before `get`, when the same effect could be achieved efficiently by `computeIfAbsent`

richardstartin:
### Issues

Fixes #2117

### Description

This PR makes minor changes to `ClusterStatusMonitor` and `PerInstanceResourceMonitor` to avoid the bottleneck described in #2117.

Notably:
* Use `computeIfAbsent` instead of multiple `HashMap` methods which duplicate work
* More efficient `hashCode` implementation in `PerInstanceResourceMonitor$BeanName` (avoid calling `toString`)
* More efficient `toString` implementation in `PerInstanceResourceMonitor$BeanName` (`String.format` is expensive)

Also removes some code warnings in `PerInstanceResourceMonitor` - remove unnecessary boxing, calls to `toString` etc.

### Tests

This does not have any functional impact.

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
@@ -459,10 +459,7 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
           String state = instanceStateMap.get(instance);
           PerInstanceResourceMonitor.BeanName beanName =
               new PerInstanceResourceMonitor.BeanName(_clusterName, instance, resource);
-          if (!beanMap.containsKey(beanName)) {
-            beanMap.put(beanName, new HashMap<Partition, String>());
-          }
-          beanMap.get(beanName).put(partition, state);
+          beanMap.computeIfAbsent(beanName, k -> new HashMap<>()).put(partition, state);
         }
       }
     }
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/PerInstanceResourceMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/PerInstanceResourceMonitor.java
@@ -70,8 +70,7 @@ public class PerInstanceResourceMonitor extends DynamicMBeanProvider {
 
     public ObjectName objectName() {
       try {
-        return new ObjectName(String.format("%s:%s", MonitorDomainNames.ClusterStatus.name(),
-            new BeanName(_clusterName, _instanceName, _resourceName).toString()));
+        return new ObjectName(MonitorDomainNames.ClusterStatus.name() + ":" + this);
       } catch (MalformedObjectNameException e) {
         LOG.error("Failed to create object name for cluster: {}, instance: {}, resource: {}.",
             _clusterName, _instanceName, _resourceName);
@@ -81,7 +80,7 @@ public class PerInstanceResourceMonitor extends DynamicMBeanProvider {
 
     @Override
     public boolean equals(Object obj) {
-      if (obj == null || !(obj instanceof BeanName)) {
+      if (!(obj instanceof BeanName)) {
         return false;
       }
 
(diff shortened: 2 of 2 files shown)


richardstartin:
`TestClusterAccessor` passes locally, but has failed here 3x

desaikomal (approved):
Thanks @richardstartin - code changes look good.

Do we have any performance numbers? Because Helix code has been using String.format at many places and so if we can estimate the improvements, then we can further optimize.

thanks,
komal

richardstartin:
> Thanks @richardstartin - code changes look good.
> 
> Do we have any performance numbers? Because Helix code has been using String.format at many places and so if we can estimate the improvements, then we can further optimize.
> 
> thanks, komal

Hi @desaikomal, I can't get you numbers from a Pinot cluster because I would need to back port these changes to 0.9.x to set it up; I'm hoping we can reap a performance improvement once the upgrade is done. I'm working on experience in Java performance optimisation and I can promise each one of these changes is a "sure thing." Replacing `containsKey`/`get` with `computeIfAbsent` intuitively halves the work, hashing without calling `toString` is also intuitively cheaper.

However, `String.format` would indeed be a rich seam for Helix engineers to work on, so it's worth sharing some representative numbers:

```java
@State(Scope.Benchmark)
public class BenchmarkStringFormat {

  public static final String RESOURCE_DN_KEY = "resourceName";
  static final String INSTANCE_DN_KEY = "instanceName";
  public static final String CLUSTER_DN_KEY = "cluster";

  @Param("resource1234")
  String _resourceName;
  @Param("instance1234")
  String _instanceName;
  @Param("cluster1234")
  String _clusterName;

  @Benchmark
  public String format() {
    return String.format("%s=%s,%s=%s,%s=%s", CLUSTER_DN_KEY, _clusterName, INSTANCE_DN_KEY, _instanceName,
        RESOURCE_DN_KEY, _resourceName);
  }

  @Benchmark
  public String concat() {
    return CLUSTER_DN_KEY + "=" + _clusterName + ","
        + INSTANCE_DN_KEY + "=" + _instanceName + ","
        + RESOURCE_DN_KEY + "=" + _resourceName;
  }
}
```

JDK11, my MacBook Pro, no attempt made to control noise:
```
Benchmark                     (_clusterName)  (_instanceName)  (_resourceName)  Mode  Cnt     Score    Error  Units
BenchmarkStringFormat.concat     cluster1234     instance1234     resource1234  avgt    5    35.098 ±  3.046  ns/op
BenchmarkStringFormat.format     cluster1234     instance1234     resource1234  avgt    5  1263.995 ± 69.025  ns/op
```

String concatenation, when all the arguments are known statically, is actually very efficient on recent JVMs - compiling to JDK7-JDK8 bytecode levels would generate the equivalent `StringBuilder` calls, from JDK9 onwards an `invokedynamic` instruction is generated, which leads to generation of specialised concatenation code at runtime. On the other hand, it's very difficult for JDK engineers to optimise `String.format` without breaking API compatibility. Though it has got a little faster on JDK17, removing `String.format` where concatenation or `StringBuilder` can be used would save a lot of money at scale.

desaikomal:
Thanks @richardstartin  - very helpful and great learnings. Helix code base still uses Java-8. May be we will need effort to move to Java-11. Let me ask the existing Helix committers if we can take in that effort.

junkaixue:
Thanks @richardstartin! Is this ready to merge? Please follow the: https://github.com/apache/helix/wiki/Pull-Request-Merge-Steps if you are good to go.

richardstartin:
@junkaixue This PR is ready to be merged

PR #200, opened 2018-04-30
throw new Exception to avoid ugly NPE

brettKK:
We have developed a static analysis tool [NPEDetector](https://github.com/lujiefsi/NPEDetector) to find some potential NPE. Our analysis shows that some callees may return null in corner case(e.g. node crash , IOException), some of their callers have  !=null check but some do not have. In this issue we post a patch which can add  !=null  based on existed !=null  check. For example:

ZkGrep#parseZkSnapshot:
```
  return retFiles;
} catch (Exception e) {
  LOG.error("fail to parse zkSnapshot: " + lastZkSnapshot, e);
}

return null;
```
So parseZkSnapshot will return null while IOException happens. but its caller ZkGrep#processCommandLineArgs have no null checker:
```
File[] lastZkSnapshot = parseZkSnapshot(zkDataDirs[1], byTime);

// lastZkSnapshot[1] is the parsed last snapshot by byTime
grepZkSnapshot(lastZkSnapshot[1], patterns);
```
We should terminate the process while  lastZkSnapshot == null

PR #2120, opened 2022-05-25
Fix failing ZkClient TTL test

rabashizade:
### Issues

### Description

In the recently added TTL feature to ZKClient, some tests are added which rely on `zookeeper.extendedTypesEnabled` system property to be set. This property was set in each of tests, and cleared at the end of the test. However, parallel execution of the tests resulted in the property being cleaned by one completed test while another test that relies on this property is still running. Due the test failures caused by this, the property is only set in the tests and never cleaned.

### Tests

command: mvn -pl zookeeper-api test
```
[INFO] Results:
[INFO] 
[INFO] Tests run: 66, Failures: 0, Errors: 0, Skipped: 0
```

### Documentation (Optional)

### Commits

### Code Quality


zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -89,29 +89,6 @@ public void afterClass() {
     _zkClient.close();
   }
 
-  @Test
-  void testUnimplementedTypes() {
-    // Make sure extended types are disabled
-    System.clearProperty("zookeeper.extendedTypesEnabled");
```
NealSun96, line 95: Why is this test removed? @rabashizade

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -146,7 +123,6 @@ void testCreatePersistentWithTTL() {
     // Clean up
     zkClient.deleteRecursively(parentPath);
     zkClient.close();
-    System.clearProperty("zookeeper.extendedTypesEnabled");
   }
 
   @Test
```
NealSun96, line 128: When you said parallel test execution, do you mean test classes running in parallel or test cases (methods) running in parallel?
For former, I remember we had a setting that prevents classes from running in parallel; for latter, we could set test dependency to force sequential execution.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -239,7 +239,6 @@ public void testSyncCreateWithTTL() {
     Assert.assertNotNull(getRecord);
     Assert.assertEquals(getRecord.getSimpleFields().size(), 0);
 
-    System.clearProperty("zookeeper.extendedTypesEnabled");
```
NealSun96, line 242: I only see removal on System.clearProperty but not System.setProperty. Is this property set by test cases and are not cleared?

NealSun96 (changes requested):

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
rabashizade: It relies on the `zookeeper.extendedTypesEnabled` not being set. Unsetting this property runs the risk of failing parallel tests that rely on it being set. Since this test just checks if ZK server throws the expected type of exception, I thought it's okay to remove it (I had added it myself in [this PR](https://github.com/apache/helix/pull/2090/files#diff-709c5951f5f7f2abe19a5641bdd677e3413bc46304e32eb29ba3513cc48029e1R93) for this feature) given this is really the ZK server behavior, and keeping it makes other tests unstable.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
rabashizade: Some of the newly added tests need this property. But if we clear it in any of them, we run the risk of failing some other test that expects it to be there but finds it's not. Unless there's a way to avoid parallel test execution between different classes, I don't know how to get around this. Please let me know if you have any suggestions.

NealSun96: I meant to ask when this "zookeeper.extendedTypesEnabled" property is ever set?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
NealSun96: I remember it was a setting in the test framework... Would you be able to try serializing test case executions using "dependsOn"? That alone should prevent parallel test issues.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
rabashizade: Do you mean in tests or in real world? In real world, it's never set from Helix's/ZkClient's perspective. It would only be set on the ZK server, if it's intended for the ZK to support TTL and Container modes. In tests, whenever we want to test creating a ZNode in TTL or Container mode, this needs to be set. Otherwise, ZK server throws an UnimplementedException.

mgao0 (commented):
A general question: can we just make the tests sequential? IMO has different system properties are necessary for a better test coverage.

rabashizade:
> A general question: can we just make the tests sequential? IMO has different system properties are necessary for a better test coverage.

This is a good point. I realized the tests are indeed being executed sequentially. However, different methods from different classes can be interleaved, e.g., not all tests from the class A finish before tests from class B start. It could be that some tests from class A run, then some from class B, and so on. This caused the test failure, and so I updated the property and the beginning and end of every method that needs it. Now the tests succeed.

mgao0 (approved):
Thanks for addressing the comment.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
NealSun96: Synced offline

rabashizade:
This PR is ready to be merged, approved by @narendly, @mgao0, and @NealSun96.

Final commit message:
**Fix failing ZkClient TTL test**

Some tests for the TTL feature rely on zookeeper.extendedTypesEnabled system property to be set. However, interleaving of test methods from different classes resulted in the property being cleaned by one completed test class while another class has tests that rely on this property still running. This commit sets the property at the beginning of every method that needs it, and cleans at the end of the method.

PR #2127, opened 2022-06-01
Add HelixEventHandlingUtil and use that util for cloud event - Add API

xyuanlu:
### Issues

#2133 

### Description

Previously helix event handling depends on batch enable/disable instance. We would like to move away the dependence on generic batch enable/disable instance and use a dedicated Util only for event handling use case.

### Tests

This change only adds PAI signature. Will add test in following PR for API impl.

### Documentation (Optional)

### Commits

### Code Quality


helix-core/src/main/java/org/apache/helix/cloud/event/HelixEventHandlingUtil.java
```diff
@@ -0,0 +1,23 @@
+package org.apache.helix.cloud.event;
+
+import org.apache.helix.BaseDataAccessor;
+
+
+public class HelixEventHandlingUtil {
+
+  public static boolean enableInstanceForCloudEvent(String clusterName, String instanceName, boolean isEnable, long timeStamp,
```
mgao0, line 8: 1. Open for discussion. I think our intention is to hide away these APIs from users, maybe we can restrict the access to package-only?
2. Making a note on our "return false" or "throw exception" discussion when the action is failed
3. Should we use the timestamp when the instance is actually enabled/disabled instead of when we initiate this move?

helix-core/src/main/java/org/apache/helix/cloud/event/HelixEventHandlingUtil.java
```diff
@@ -0,0 +1,23 @@
...
+      BaseDataAccessor dataAccessor) {
+    return true;
+  }
+
+  public static boolean IsInstanceDisabledForCloudEvent(String clusterName, String instanceName,
```
mgao0, line 13: I understand why you name it like that - so it's more clear that this check is "for cloud event". But it is a bit counter intuitive because we return true for disabled, and return false for enabled. I'm thinking maybe we could still name it "enableInstance", "isInstanceEnabled", etc. But stress that these methods are for use cases under the cloud event context only in the method java doc, and should not be used for other purposes. What do you think?

helix-core/src/main/java/org/apache/helix/cloud/event/HelixEventHandlingUtil.java
```diff
@@ -0,0 +1,23 @@
...
+      BaseDataAccessor dataAccessor) {
+    return true;
+  }
+
+  public static Long getInstanceDisabledTimestamp(String clusterName, String instanceName,
```
mgao0, line 18: Is this timestamp for cloud event disable or for the general disable?

helix-core/src/main/java/org/apache/helix/cloud/event/HelixEventHandlingUtil.java
```diff
@@ -0,0 +1,23 @@
+package org.apache.helix.cloud.event;
+
+import org.apache.helix.BaseDataAccessor;
+
+
+public class HelixEventHandlingUtil {
```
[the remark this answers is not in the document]
xyuanlu: TFTR. Updated

helix-core/src/main/java/org/apache/helix/cloud/event/HelixEventHandlingUtil.java
xyuanlu: Is intend to be the timestamp for cloud event. I updated the signature to take a message. It will be less ambiguous.

helix-core/src/main/java/org/apache/helix/cloud/event/HelixEventHandlingUtil.java
xyuanlu: Based on our offline discussion, let's keep the current name..

helix-core/src/main/java/org/apache/helix/cloud/event/HelixEventHandlingUtil.java
xyuanlu: Updated.

helix-core/src/main/java/org/apache/helix/cloud/event/helix/HelixEventHandlingUtil.java
```diff
@@ -0,0 +1,56 @@
...
+   * @param instanceName
+   * @param message
+   * @param isEnable
+   * @param dataAccessor
+   * @return return failure when either enable/disable failed or update cluster ZNode failed.
+   */
+   static boolean enableInstanceForCloudEvent(String clusterName, String instanceName, String message,
+      boolean isEnable, BaseDataAccessor dataAccessor) {
+    return true;
+  }
```
qqu0127, line 41: Let's add TODO for this and below methods.

helix-core/src/main/java/org/apache/helix/cloud/event/helix/HelixEventHandlingUtil.java
```diff
@@ -0,0 +1,56 @@
...
+ class HelixEventHandlingUtil {
+
+  /**
+   * Enable or disable an instance for cloud event.
+   * It will enable/disable Helix for that instance. Also add the instance cloud event info to
+   * clusterConfig Znode when enable.
+   * @param clusterName
+   * @param instanceName
+   * @param message
+   * @param isEnable
```
qqu0127, line 34: nit: enabled or isEnabled would be better

helix-core/src/main/java/org/apache/helix/cloud/event/helix/DefaultCloudEventCallbackImpl.java
```diff
@@ -90,10 +90,9 @@ public void exitMaintenanceMode(HelixManager manager, Object eventInfo) {
     // Check if there is any disabled live instance that was disabled due to cloud event,
     // if none left, exit maintenance mode
     HelixDataAccessor accessor = manager.getHelixDataAccessor();
-    if (instances.stream().noneMatch(instance ->
-        InstanceValidationUtil.getInstanceHelixDisabledType(accessor, instance)
-            .equals(InstanceConstants.InstanceDisabledType.CLOUD_EVENT.name())
-            && InstanceValidationUtil.isAlive(accessor, instance))) {
+    if (instances.stream().noneMatch(instance -> HelixEventHandlingUtil
+        .IsInstanceDisabledForCloudEvent(manager.getClusterName(), instance, accessor.getBaseDataAccessor())
+        && InstanceValidationUtil.isAlive(accessor, instance))) {
```
qqu0127, line 95: With current implementation of `IsInstanceDisabledForCloudEvent`, it's always returning true, and we won't exit maintenance mode at all. Is this an issue?

qqu0127 (commented):
Please also add more background and context to this PR

helix-core/src/main/java/org/apache/helix/cloud/event/helix/HelixEventHandlingUtil.java
xyuanlu: This function will enable or disable an instance. would `isEnable` be better to indicate the operation type? I feel like "enabled" is used to describe the state if an instance, not the operation type.

qqu0127: I'm fine either way. Feel free to resolve this.

qqu0127 (approved):
LGTM

xyuanlu:
This change is ready to be merged. Approved by @qqu0127

Final commit message:
**Add HelixEventHandlingUtil and use that util for cloud event - API Only**
Previously helix event handling depends on batch enable/disable instance. We would like to move away the dependence on generic batch enable/disable instance and use a dedicated Util only for event handling use case. This change adds API signature for util functions/

PR #2137, opened 2022-06-02
Add backward compatibility support for constructing ZKHelixManager

### LINKED ISSUE #2160: HelixManagerFactory doesn't support instantiating ZKHelixManager with HelixManagerProperty for Non-MSDS use cases
### Describe the bug
 ` public static HelixManager getZKHelixManager(String clusterName, String instanceName,
      InstanceType type, HelixManagerStateListener stateListener,
      HelixManagerProperty helixManagerProperty);` only works for ZooScalability cases.
Reason: 
If user passes in HelixManagerProperty in the call to this method, right now the code treats it as a ZooScalability use case, so it will try to connect using the ZooScalability logic.
It was fine until now because non-ZooScalability users will use another method in the HelixManagerFactory `  public static HelixManager getZKHelixManager(String clusterName, String instanceName,
      InstanceType type, String zkAddr, HelixManagerStateListener stateListener) ` to pass in zkAddr, so it will connect as the traditional way.
But the need for non-ZooScalability users to use the method with HelixManagerProperty emerges because HelixManagerProperty contains HelixCloudProperty. So backward compatibility support for  ` public static HelixManager getZKHelixManager(String clusterName, String instanceName,
      InstanceType type, HelixManagerStateListener stateListener,
      HelixManagerProperty helixManagerProperty);`  is needed, so it will support non-ZooScalability use cases as well.

### To Reproduce
Steps to reproduce the behavior.

mgao0:
### Issues

Resolves #2160 

### Description

This commit uses a deprecated class `HelixZkClient.ZkConnectionConfig` to convey the information for non-ZooScalability users in HelixManagerProperty, while ZooScalability users use `RealmAwareZkClient.RealmAwareZkConnectionConfig`. This commit adds some validation around there should be at least one, and only one of `HelixZkClient.ZkConnectionConfig`, `RealmAwareZkClient.RealmAwareZkConnectionConfig`, and zk address exists, and this one field will be used as the source of information to establish connection with ZK server.

### Tests

Added TestHelixManagerFactory.
Also ran TestMultiZkHelixJavaApis.java to make sure this commit doesn't break current logic for ZooScalability cases.

Running

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
@@ -21,6 +21,7 @@ package org.apache.helix;
 
 import java.util.Properties;
 
+import org.apache.helix.manager.zk.client.HelixZkClient;
 import org.apache.helix.model.CloudConfig;
 import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.slf4j.Logger;
@@ -35,8 +36,11 @@ public class HelixManagerProperty {
   private String _version;
   private long _healthReportLatency;
   private HelixCloudProperty _helixCloudProperty;
-  private RealmAwareZkClient.RealmAwareZkConnectionConfig _zkConnectionConfig;
+  private RealmAwareZkClient.RealmAwareZkConnectionConfig _realmAwareZkConnectionConfig;
   private RealmAwareZkClient.RealmAwareZkClientConfig _zkClientConfig;
+  // This field is added for backward compatibility, when this field is set,
+  // _realmAwareZkConnectionConfig should not be set
+  private HelixZkClient.ZkConnectionConfig _zkConnectionConfig;
 
   /**
    * ** Deprecated - HelixManagerProperty should be a general property/config object used for
@@ -56,13 +60,15 @@ public class HelixManagerProperty {
 
   private HelixManagerProperty(String version, long healthReportLatency,
       HelixCloudProperty helixCloudProperty,
-      RealmAwareZkClient.RealmAwareZkConnectionConfig zkConnectionConfig,
-      RealmAwareZkClient.RealmAwareZkClientConfig zkClientConfig) {
+      RealmAwareZkClient.RealmAwareZkConnectionConfig realmAwareZkConnectionConfig,
+      RealmAwareZkClient.RealmAwareZkClientConfig zkClientConfig,
+      HelixZkClient.ZkConnectionConfig zkConnectionConfig) {
     _version = version;
     _healthReportLatency = healthReportLatency;
     _helixCloudProperty = helixCloudProperty;
-    _zkConnectionConfig = zkConnectionConfig;
+    _realmAwareZkConnectionConfig = realmAwareZkConnectionConfig;
     _zkClientConfig = zkClientConfig;
+    _zkConnectionConfig = zkConnectionConfig;
   }
 
   public HelixCloudProperty getHelixCloudProperty() {
@@ -80,27 +86,32 @@ public class HelixManagerProperty {
     return _healthReportLatency;
   }
 
-  public RealmAwareZkClient.RealmAwareZkConnectionConfig getZkConnectionConfig() {
-    return _zkConnectionConfig;
+  public RealmAwareZkClient.RealmAwareZkConnectionConfig getRealmAwareZkConnectionConfig() {
+    return _realmAwareZkConnectionConfig;
   }
 
   public RealmAwareZkClient.RealmAwareZkClientConfig getZkClientConfig() {
     return _zkClientConfig;
   }
 
+  public HelixZkClient.ZkConnectionConfig getZkConnectionConfig() {
+    return _zkConnectionConfig;
+  }
+
   public static class Builder {
     private String _version;
     private long _healthReportLatency;
     private HelixCloudProperty _helixCloudProperty;
-    private RealmAwareZkClient.RealmAwareZkConnectionConfig _zkConnectionConfig;
+    private RealmAwareZkClient.RealmAwareZkConnectionConfig _realmAwareZkConnectionConfig;
     private RealmAwareZkClient.RealmAwareZkClientConfig _zkClientConfig;
+    private HelixZkClient.ZkConnectionConfig _zkConnectionConfig;
 
     public Builder() {
     }
 
     public HelixManagerProperty build() {
       return new HelixManagerProperty(_version, _healthReportLatency, _helixCloudProperty,
-          _zkConnectionConfig, _zkClientConfig);
+          _realmAwareZkConnectionConfig, _zkClientConfig, _zkConnectionConfig);
     }
 
     public Builder setVersion(String version) {
@@ -120,13 +131,18 @@ public class HelixManagerProperty {
 
     public Builder setRealmAWareZkConnectionConfig(
         RealmAwareZkClient.RealmAwareZkConnectionConfig zkConnectionConfig) {
-      _zkConnectionConfig = zkConnectionConfig;
+      _realmAwareZkConnectionConfig = zkConnectionConfig;
       return this;
     }
 
     public Builder setRealmAwareZkClientConfig(
         RealmAwareZkClient.RealmAwareZkClientConfig zkClientConfig) {
       _zkClientConfig = zkClientConfig;
+      return this;
+    }
+
+    public Builder setZkConnectionConfig(HelixZkClient.ZkConnectionConfig zkConnectionConfig) {
+      _zkConnectionConfig = zkConnectionConfig;
       return this;
     }
   }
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1560,30 +1560,47 @@ private RealmAwareZkClient resolveZkClient(HelixZkClientFactory zkClientFactory,
...
-          helixManagerProperty.getZkConnectionConfig();
+          helixManagerProperty.getRealmAwareZkConnectionConfig();
       if (connectionConfig.getZkRealmShardingKey() == null || connectionConfig
           .getZkRealmShardingKey().isEmpty()) {
         throw new HelixException(
             "ZKHelixManager::ZK path sharding key must be set for ZKHelixManager! ZKHelixManager "
                 + "is only available on single-realm mode.");
       }
       _realmAwareZkConnectionConfig = connectionConfig;
+    } else if (hasZkConnectionConfig) {
```
junkaixue, line 1593: Can these two be coexist? Or it is either or relationship.

helix-core/src/test/java/org/apache/helix/manager/zk/TestHelixManagerFactory.java
```diff
@@ -0,0 +1,10 @@
+package org.apache.helix.manager.zk;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TestHelixManagerFactory {
+  private static Logger LOG = LoggerFactory.getLogger(TestHelixManagerFactory.class);
+
```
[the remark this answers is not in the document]
mgao0: It's still work in progress. I will add the tests once they are completed.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
mgao0: These two and zk address should not coexist, there should be only one of them. Line 1576 - 1577 is used to enforce this relationship.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -80,27 +86,32 @@ public long getHealthReportLatency() {
...
     private RealmAwareZkClient.RealmAwareZkClientConfig _zkClientConfig;
+    private HelixZkClient.ZkConnectionConfig _zkConnectionConfig;
 
     public Builder() {
     }
 
     public HelixManagerProperty build() {
       return new HelixManagerProperty(_version, _healthReportLatency, _helixCloudProperty,
-          _zkConnectionConfig, _zkClientConfig);
+          _realmAwareZkConnectionConfig, _zkClientConfig, _zkConnectionConfig);
```
xyuanlu, line 114: Question: For this public function, shall we keep the original signature and add this as a new one?

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -35,8 +36,11 @@ public class HelixManagerProperty {
   private String _version;
   private long _healthReportLatency;
   private HelixCloudProperty _helixCloudProperty;
-  private RealmAwareZkClient.RealmAwareZkConnectionConfig _zkConnectionConfig;
+  private RealmAwareZkClient.RealmAwareZkConnectionConfig _realmAwareZkConnectionConfig;
   private RealmAwareZkClient.RealmAwareZkClientConfig _zkClientConfig;
+  // This field is added for backward compatibility, when this field is set,
+  // _realmAwareZkConnectionConfig should not be set
+  private HelixZkClient.ZkConnectionConfig _zkConnectionConfig;
```
desaikomal, line 43: Do we use Optional property indicating that it may not be set?

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -56,13 +60,15 @@ public HelixManagerProperty(Properties helixManagerProperties, CloudConfig cloud
 
   private HelixManagerProperty(String version, long healthReportLatency,
       HelixCloudProperty helixCloudProperty,
-      RealmAwareZkClient.RealmAwareZkConnectionConfig zkConnectionConfig,
-      RealmAwareZkClient.RealmAwareZkClientConfig zkClientConfig) {
+      RealmAwareZkClient.RealmAwareZkConnectionConfig realmAwareZkConnectionConfig,
+      RealmAwareZkClient.RealmAwareZkClientConfig zkClientConfig,
+      HelixZkClient.ZkConnectionConfig zkConnectionConfig) {
```
desaikomal, line 65: if i read your previous comment, you want either realmAwareZkConnection or regular ZkConnection, so can we not have 2 distinct constructors and not break any existing users?

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -80,27 +86,32 @@ public long getHealthReportLatency() {
     return _healthReportLatency;
   }
 
-  public RealmAwareZkClient.RealmAwareZkConnectionConfig getZkConnectionConfig() {
-    return _zkConnectionConfig;
+  public RealmAwareZkClient.RealmAwareZkConnectionConfig getRealmAwareZkConnectionConfig() {
```
desaikomal, line 89: this can break users as they used to get RealmAwareZkClient earlier, and will need to update code to use appropriate

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -80,27 +86,32 @@ public long getHealthReportLatency() {
...
+    return _realmAwareZkConnectionConfig;
   }
 
   public RealmAwareZkClient.RealmAwareZkClientConfig getZkClientConfig() {
     return _zkClientConfig;
   }
 
+  public HelixZkClient.ZkConnectionConfig getZkConnectionConfig() {
```
desaikomal, line 97: since this is new property, you can rename this method, so that clients don't have to change.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
desaikomal: I agree, we should have 2 distinct methods

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -238,7 +239,6 @@ public ZKHelixManager(String clusterName, String instanceName, InstanceType inst
       HelixManagerProperty helixManagerProperty) {
     validateZkConnectionSettings(zkAddress, helixManagerProperty);
 
-    _zkAddress = zkAddress;
```
desaikomal, line 241: any particular reason, zkAddress is not set here. Customer has provided as input.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -1560,30 +1560,47 @@ private RealmAwareZkClient resolveZkClient(HelixZkClientFactory zkClientFactory,
...
-        throw new HelixException(
-            "ZKHelixManager: cannot have both ZkAddress and ZkConnectionConfig set!");
-      }
+    boolean hasRealmAwareZkConnectionConfig = helixManagerProperty != null
+        && helixManagerProperty.getRealmAwareZkConnectionConfig() != null;
+    boolean hasZkConnectionConfig =
+        helixManagerProperty != null && helixManagerProperty.getZkConnectionConfig() != null;
+    if (Stream.of(zkAddress != null, hasRealmAwareZkConnectionConfig, hasZkConnectionConfig)
+        .filter(condition -> condition).count() != 1) {
+      throw new HelixException(
```
desaikomal, line 1578: i find this pattern a bit counter intuitive. You have 3 different choices. Typically, we would have 3 distinct constructors (eg. Integer(int), Integer(Integer), Integer(string)??)  

Other thing is if customer has used a particular constructor, other properties will be null or ignored.AFAIK we don't allow update of Helix manager once constructor is called.

helix-core/src/test/java/org/apache/helix/manager/zk/TestHelixManagerFactory.java
```diff
@@ -0,0 +1,70 @@
...
+    // Create zk helix manager using HelixManagerFactory API
+    HelixManager managerParticipant = HelixManagerFactory
+        .getZKHelixManager(CLUSTER_NAME, participantName, InstanceType.PARTICIPANT, null,
+            new HelixManagerProperty.Builder()
+                .setZkConnectionConfig(new HelixZkClient.ZkConnectionConfig(ZK_ADDR)).build());
+    managerParticipant.connect();
+
+    // Verify manager
+    InstanceConfig instanceConfigRead =
```
desaikomal, line 65: you will need multiple negative test scenario. ie. only zkAddress case, only realmZkClientConnectconfig and combinations throwing exception etc.

mgao0:
Closing this PR as this is not a necessary feature, and may introduce other backward incompatibilities in order to fix it. Will evaluate the situation later and reopen if necessary.

PR #2161, opened 2022-06-17
Fix TestHelixViewAggregator test

qqu0127:
### Issues

Fixes https://github.com/apache/helix/issues/2049

### Description

CI test history shows that test TestHelixViewAggregator is unstable. We suspect two causes:
1. Delayed rebalance triggered during testing. 
2. Cluster events from previous test received after monitor reset therefore overlap with the following tests.

This PR disables delayed rebalance for the source clusters in the test, and updates the event verification logic.

### Tests

Local test passed.

[INFO] Tests run: 13, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 60.877 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 13, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] --- jacoco-maven-plugin:0.8.6:report (generate-code-coverage-report) @ helix-view-aggregator ---
[INFO] Loading execution data file /home/qqu/workspace/qqu-helix/helix-view-aggregator/target/jacoco.exec
[INFO] Analyzed bundle 'Apache Helix :: View Aggregator' with 15 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  01:07 min
[INFO] Finished at: 2022-06-30T14:02:03-07:00

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-view-aggregator/src/test/java/org/apache/helix/view/integration/TestHelixViewAggregator.java
+++ b/helix-view-aggregator/src/test/java/org/apache/helix/view/integration/TestHelixViewAggregator.java
@@ -123,7 +123,8 @@ public class TestHelixViewAggregator extends ViewAggregatorIntegrationTestBase {
 
     // Wait for refresh and verify
     Thread.sleep((_viewClusterRefreshPeriodSec + 2) * 1000);
-    verifyViewClusterEventChanges(false, true, true);
+    verifyInstanceConfigChange();
+    verifyLiveInstanceChange();
     Set<String> allParticipantNames = new HashSet<>();
     for (MockParticipantManager participant : _allParticipants) {
       allParticipantNames.add(participant.getInstanceName());
@@ -142,7 +143,7 @@ public class TestHelixViewAggregator extends ViewAggregatorIntegrationTestBase {
 
     // Wait for refresh and verify
     Thread.sleep((_viewClusterRefreshPeriodSec + 2) * 1000);
-    verifyViewClusterEventChanges(true, false, false);
+    verifyExternalViewChange();
     Assert.assertEquals(
         new HashSet<>(_monitor.getPropertyNamesFromViewCluster(PropertyType.EXTERNALVIEW)),
         _allResources);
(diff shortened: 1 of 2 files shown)


helix-view-aggregator/src/test/java/org/apache/helix/view/integration/TestHelixViewAggregator.java
```diff
@@ -125,7 +123,8 @@ public void testHelixViewAggregator() throws Exception {
 
     // Wait for refresh and verify
     Thread.sleep((_viewClusterRefreshPeriodSec + 2) * 1000);
-    verifyViewClusterEventChanges(false, true, true);
+    verifyInstanceConfigChange();
```
NealSun96, line 126: This as well as all the other similar changes:

The behavior is changed here. When it was (false, true, true), it was verifying that `_monitor.getExternalViewChangeCount() <= 0`, but that's no longer a check here. Is there a reason for that?

NealSun96 (changes requested):

helix-view-aggregator/src/test/java/org/apache/helix/view/integration/TestHelixViewAggregator.java
qqu0127: You are right about the behavior change, we are narrowing down the test scope since "No cluster event change" is actually not a functional requirement. 
On the other side, the overlapping effect is also introducing noise to the overall CI tests.

qqu0127: Yes. In extreme scenario, cluster event triggered by previous test can happen after the monitor reset and received during the following test, this would cause unexpected test failure if we check `_monitor.getExternalViewChangeCount() <= 0` e.g.
And yes, there could be false negative in some cases (regardless of this patch or not), where `> 0` test pass because of events from previous test or other sources.

junkaixue:
Do we still this PR?

Issue #2175, opened 2022-07-13
Enable setting msgHandlerResetTimeout through HelixManagerFactory


qqu0127:
**Is your feature request related to a problem? Please describe.**
https://github.com/apache/helix/pull/1920 implemented the timeout logic for TaskExecutor, but client wouldn't be able to pass down the parameter if they don't have access to MessagingService.

**Describe the solution you'd like**
We should enable setting this parameter through HelixManagerFactory through HelixManagerProperty

PR #1920, opened 2021-12-09
Make theadpool shutdown timeout configurable for the HelixTaskExecutor.

jiajunwang:
### Issues

- [X] My PR addresses the following Helix issues and references them in the PR description:

Resolves #1919 

### Description

- [X] Here are some details about my PR, including screenshots of any UI changes:

Make theadpool shutdown timeout configurable for the HelixTaskExecutor.

Add TestHelixTaskExecutor.testHandlerResetTimeout() to cover the new changes.
Also refactoring the related code to reduce duplicate and confusing code.

### Tests

- [X] The following tests are written for this issue:

 TestHelixTaskExecutor.testHandlerResetTimeout()

- The following is the result of the "mvn test" command on the appropriate module:

(If CI test fails due to known issue, please specify the issue and test PR locally. Then copy & paste the result of "mvn test" to here.)

### Changes that Break Backward Compatibility (Optional)

- My PR contains changes that break backward compatibility or previous assumptions for certain methods or API. They include:

(Consider including all behavior changes for public methods or API. Also include these changes in merge description so that other developers are aware of these changes. This allows them to make relevant code changes in feature branches accounting for the new method/API behavior.)

### Documentation (Optional)

- In case of new functionality, my PR adds documentation in the following wiki page:

(Link the GitHub wiki you added)

### Commits

- My commits all reference appropriate Apache Helix GitHub issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Code Quality

- My diff has been formatted using helix-style.xml 
(helix-style-intellij.xml if IntelliJ IDE is used)

PR #2176, opened 2022-07-13
Wire in the logic for TaskExecutor reset timeout parameter

qqu0127:
### Issues

Fix https://github.com/apache/helix/issues/2175

### Description

The timeout logic introduced in https://github.com/apache/helix/pull/1920, failed to provide a client side entry point. As a followup, this PR address the issue by passing down the parameter from HelixManagerFactory down to HelixTaskExecutor so that helix users can specify.
The parameter is encapsulated in HelixManagerProperty to enable client to specify value and pass down from HelixManagerFactory

Update:

We add zkAddr to HelixManagerProperty and parameter validation to resolve param conflicts. Rollback the new constructor in HelixManagerFactory.
Validations in two places:
1. HelixManagerProperty construction, at most one of {zkAddr and ZkConnectionConfig} shall be set. They can both be empty though, a warning is logged but won't block.
2. (Existing logic) ZKHelixManager:validateZkConnectionSettings, if ZkConnectionConfig is set, zkAddress from ZKHelixManager constructor must be null, otherwise early stop.

With above two checks, we can ensure: if ZkConnectionConfig is set => the zkAddress must be null, either from HelixManagerProperty or from ZKHelixManager.
We set higher precedence for zkAddr from HelixManagerProperty, if this is empty, zkAddress passed from constructor is used.

### Tests

Modified existing test case in [TestParticipantManager.java](https://github.com/apache/helix/compare/master...qqu0127:reset-timeout?expand=1#diff-41a02eec6acda934146e47c98a53813c4af8fd82091db2aadd9f80165af04d17)

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
@@ -62,6 +62,22 @@ public final class HelixManagerFactory {
   }
 
   /**
+   * Construct a zk-based cluster manager that enforces all types (PARTICIPANT, CONTROLLER, and
+   * SPECTATOR) to have a name
+   * @param clusterName
+   * @param instanceName
+   * @param type
+   * @param zkAddr
+   * @param stateListener
+   * @param helixManagerProperty
+   * @return a HelixManager backed by Zookeeper
+   */
+  public static HelixManager getZKHelixManager(String clusterName, String instanceName,
+      InstanceType type, String zkAddr, HelixManagerStateListener stateListener, HelixManagerProperty helixManagerProperty) {
+    return new ZKHelixManager(clusterName, instanceName, type, zkAddr, stateListener, helixManagerProperty);
+  }
+
+  /**
    * Construct a ZkHelixManager using the HelixManagerProperty instance given.
    * HelixManagerProperty given must contain a valid ZkConnectionConfig.
    * @param clusterName
--- a/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
@@ -62,30 +62,16 @@ public final class HelixManagerFactory {
   }
 
   /**
-   * Construct a zk-based cluster manager that enforces all types (PARTICIPANT, CONTROLLER, and
-   * SPECTATOR) to have a name
-   * @param clusterName
-   * @param instanceName
-   * @param type
-   * @param zkAddr
-   * @param stateListener
-   * @param helixManagerProperty
-   * @return a HelixManager backed by Zookeeper
-   */
-  public static HelixManager getZKHelixManager(String clusterName, String instanceName,
-      InstanceType type, String zkAddr, HelixManagerStateListener stateListener, HelixManagerProperty helixManagerProperty) {
-    return new ZKHelixManager(clusterName, instanceName, type, zkAddr, stateListener, helixManagerProperty);
-  }
-
-  /**
    * Construct a ZkHelixManager using the HelixManagerProperty instance given.
-   * HelixManagerProperty given must contain a valid ZkConnectionConfig.
+   * HelixManagerProperty given must contain a valid ZkConnectionConfig or zkAddress, but not both.
    * @param clusterName
    * @param instanceName
    * @param type
    * @param stateListener
-   * @param helixManagerProperty must contain a valid ZkConnectionConfig
-   * @return
+   * @param helixManagerProperty must contain one and only one of the following, not both:
+   *                             1. a valid ZkConnectionConfig
+   *                             2. a valid zkAddress
+   * @return a HelixManager backed by Zookeeper
    */
   public static HelixManager getZKHelixManager(String clusterName, String instanceName,
       InstanceType type, HelixManagerStateListener stateListener,
--- a/helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
+++ b/helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
@@ -21,6 +21,7 @@ package org.apache.helix;
 
 import java.util.Properties;
 
+import org.apache.helix.messaging.handling.TaskExecutor;
 import org.apache.helix.model.CloudConfig;
 import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
 import org.slf4j.Logger;
@@ -34,6 +35,7 @@ public class HelixManagerProperty {
   private static final Logger LOG = LoggerFactory.getLogger(HelixManagerProperty.class.getName());
   private String _version;
   private long _healthReportLatency;
+  private int _msgHandlerResetTimeout = TaskExecutor.DEFAULT_MSG_HANDLER_RESET_TIMEOUT_MS;
   private HelixCloudProperty _helixCloudProperty;
   private RealmAwareZkClient.RealmAwareZkConnectionConfig _zkConnectionConfig;
   private RealmAwareZkClient.RealmAwareZkClientConfig _zkClientConfig;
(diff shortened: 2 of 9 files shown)


helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
```diff
@@ -61,6 +61,22 @@ public static HelixManager getZKHelixManager(String clusterName, String instance
     return new ZKHelixManager(clusterName, instanceName, type, zkAddr, stateListener);
   }
 
+  /**
```
desaikomal, line 64: nit:  this comment is same as all other constructors. 
What is the unique difference of this constructor needs to be documented?

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -34,6 +35,7 @@ public class HelixManagerProperty {
   private static final Logger LOG = LoggerFactory.getLogger(HelixManagerProperty.class.getName());
   private String _version;
   private long _healthReportLatency;
+  private int _msgHandlerResetTimeout = TaskExecutor.DEFAULT_MSG_HANDLER_RESET_TIMEOUT_MS;
```
desaikomal, line 38: I feel this is how Helix and Task framework code is intermingled.  msgHandlerResetTimeout - is it specific to Task? or it is generic msgHandler?

If it is generic, then having default in previous file makes sense. if it is specific to task, then variable name should be changed to indicate it is taskSpecific.
As far as I know it is generic.

helix-core/src/main/java/org/apache/helix/messaging/DefaultMessagingService.java
```diff
@@ -72,8 +78,8 @@ public DefaultMessagingService(HelixManager manager) {
         new ParticipantStatusMonitor(isParticipant, manager.getInstanceName()),
         new MessageQueueMonitor(manager.getClusterName(), manager.getInstanceName()));
     _asyncCallbackService = new AsyncCallbackService();
-    _taskExecutor.registerMessageHandlerFactory(MessageType.TASK_REPLY.name(),
-        _asyncCallbackService);
+    _taskExecutor.registerMessageHandlerFactory(_asyncCallbackService, TaskExecutor.DEFAULT_PARALLEL_TASKS, resetTimeoutMs);
```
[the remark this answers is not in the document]
qqu0127: This is essentially the same with method that takes in `resetTimeoutMs`. 
`_asyncCallbackService` is an instance of `MultiTypeMessageHandlerFactory` so we can use method `  public void registerMessageHandlerFactory(MultiTypeMessageHandlerFactory factory,
      int threadPoolSize, int resetTimeoutMs) `

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
```diff
@@ -61,6 +61,22 @@ public static HelixManager getZKHelixManager(String clusterName, String instance
...
+   * @param instanceName
+   * @param type
+   * @param zkAddr
+   * @param stateListener
+   * @param helixManagerProperty
+   * @return a HelixManager backed by Zookeeper
+   */
+  public static HelixManager getZKHelixManager(String clusterName, String instanceName,
+      InstanceType type, String zkAddr, HelixManagerStateListener stateListener, HelixManagerProperty helixManagerProperty) {
+    return new ZKHelixManager(clusterName, instanceName, type, zkAddr, stateListener, helixManagerProperty);
```
NealSun96, line 77: I have concerns over this constructor, and also storing the reset timeout in HelixManagerProperty in general:

If you check the ZkHelixManager logic, you can see that `zkAddr` and `helixManagerProperty.getZkConnectionConfig()` cannot coexist. This is why HelixManagerProperty is only provided in the constructor below this new one: the `zkAddr` in that constructor is null. Therefore, it concerns me that this constructor can lead to a lot of ZkHelixManager initialization failures because of the above mentioned conflict. I believe most users will provide both a zkAddr and a connection config.

helix-core/src/main/java/org/apache/helix/HelixManagerProperty.java
```diff
@@ -54,12 +56,13 @@ public HelixManagerProperty(Properties helixManagerProperties, CloudConfig cloud
         helixManagerProperties.getProperty(SystemPropertyKeys.PARTICIPANT_HEALTH_REPORT_LATENCY));
   }
 
-  private HelixManagerProperty(String version, long healthReportLatency,
+  private HelixManagerProperty(String version, long healthReportLatency, int msgHandlerResetTimeout,
```
NealSun96, line 59: Same as above: I have concerns over adding this timeout in HelixManagerProperty, due to the connectionConfig it carries.

helix-core/src/main/java/org/apache/helix/messaging/DefaultMessagingService.java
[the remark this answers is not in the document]
NealSun96: I can answer for @qqu0127 : the new signature he used here checks for `_asyncCallbackService.getMessageTypes()` and calls the line 75 `registerMessageHandlerFactory` for each of the types. Since `_asyncCallbackService` only has one message type, which is `TASK_REPLY`, this call is identical with the old call, except that it also adds a resetTimeoutMS.

NealSun96 (commented):
As I said, I don't think HelixManagerProperty is the best place to store the reset timeout - the problem is not with your design, but unfortunately our existing logic will cause confusions to customers; without changing the existing logic, that's not the best place.

helix-core/src/test/java/org/apache/helix/integration/manager/TestParticipantManager.java
```diff
@@ -101,8 +103,13 @@ public void simpleIntegrationTest() throws Exception {
         "MasterSlave", true); // do rebalance
 
     String instanceName = "localhost_12918";
-    HelixManager participant =
-        new ZKHelixManager(_clusterName, instanceName, InstanceType.PARTICIPANT, ZK_ADDR);
+    int resetTimeout = 300;
```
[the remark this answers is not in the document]
qqu0127: This is very good question, it's not new logic, and I don't see a maximum value from https://github.com/apache/helix/pull/1920
I assume the idea is to delegate this to user for a reasonable value, it's hard to set a global max and override. Let me know if you think otherwise. Thanks.

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
qqu0127: To add on, `public ZKHelixManager(String clusterName, String instanceName, InstanceType instanceType,
      String zkAddress, HelixManagerStateListener stateListener,
      HelixManagerProperty helixManagerProperty)` is already a public constructor in `ZkHelixManager` for a while.

NealSun96: While that constructor has been there for a while, the more common pattern for customers is to create a ZkHelixManager through HelixManagerFactory, and in the factory, the existence of zkAddr and HelixManagerProperty is so far mutually exclusive. 
With this new constructor provided, I predict there will be confusion - it's the first time zkAddr and HelixManagerProperty are both accepted, and we now must make it explicit to the customers that only 1 should be provided. I don't think adding a warning comment will be enough - do you have some ideas about "better logic handling" you mentioned above?
I believe the best way for this problem is not to enforce the RealmAwareConnectionConfig in HelixManagerProperty, but that's out of the scope of this PR. 
With what we have now, I will much prefer the reset value to be set in another function, similar to `setLiveInstanceInfoProvider`, so something like `setMessageHandlerResetTimeout`.

mgao0:
You'll need to modify more logic to enable the constructor with both zk address and helix manager property. Please take a look at this [PR](https://github.com/apache/helix/pull/2137), it's purpose was to add a constructor like that.

helix-core/src/main/java/org/apache/helix/HelixManagerFactory.java
NealSun96: The `insert zkAddress into helixManagerProperty` approach is what makes sense to me, but this is again outside the scope. I agree that `setMessageHandlerResetTimeout` will require more work to refactor. 
It's not perfect but I will leave it to your call. The bare minimum I will ask for is a note in the javadoc to highlight that zkaddress should not coexist with the connection config, in case untested code gets deployed to participants and fail.

qqu0127: Agree, implemented this one and updated

mgao0:
> > You'll need to modify more logic to enable the constructor with both zk address and helix manager property. Please take a look at this [PR](https://github.com/apache/helix/pull/2137), its purpose was to add a constructor like that.
> 
> Thanks @mgao0 for the pointer. Is this the same thing Neal mentioned? I feel if we make sure `_zkConnectionConfig` and `_zkClientConfig` are null in `HelixManagerProperty`, while a valid zkAddr is provided, we should be good to go. Could you please confirm?

I saw your latest commit that added zkAddr to HelixManagerProperty, it makes sense.

Issue #2178, opened 2022-07-22
Thread leakage from BestPossibleExternalViewVerifier.verify()


qqu0127:
### Describe the bug

We detected a resource leakage from BestPossibleExternalViewVerifier.verify(), there is a loop where ZkClient could be continuously created but never closed. This may drain JVM thread and ZK connection and severely damage system stability.

The code path for zkclient creation:
`ZkHelixClusterVerifier.handleDataChange() -> verifyState() submit to verifier threadpool -> BestPossibleExternalViewVerifier.calcBestPossState() -> DryrunWagedRebalancer -> ZkBucketDataAccessor -> createRealmAwareZkClient`
Notice that ReadOnlyAssignmentMetadataStore override `close()` with a dummy method, 
https://github.com/apache/helix/blob/49aef7c54a7f93f119f0b921068235d3dd52cd25/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/ReadOnlyWagedRebalancer.java#L94
as a result zkclient in `ZkBucketDataAccessor` is never closed, where it should as in https://github.com/apache/helix/blob/49aef7c54a7f93f119f0b921068235d3dd52cd25/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java#L253

Impact: all participants or controllers that are using BestPossibleExternalViewVerifier.verify() or verifyByCallback() or verifyByPolling() may have thread leak and large number of ZK connections.

### Expected behavior
ZkClient and other resources should be properly closed.

### Additional context
Suspect breaking change: https://github.com/apache/helix/pull/1747
Impact version: 1.0.3+

Issue #2179, opened 2022-07-22
Sharing existing zkClient instead of creating new one while creating DryrunWagedRebalancer


rahulrane50:
### Describe the bug
Currently every time a calcBestPossState is called for any verify event, while creating DryrunWagedRebalancer we just pass metadataStoreAddresses from zkClient object and create a new zkClient object in ZkBucketDataAccessor. Ideally whole zkClient object should be passed and re-used by ZkBucketDataAccessor

### To Reproduce
Steps to reproduce the behavior.

### Expected behavior
We need to investigate why is this zkClient object not shared but created newly. I believe we can atleast share zkConnection if not zkClient object.

jiajunwang:
I believe there was a leakage so we decided to create new client for easier cleanup. Please be mindful to ensure the client will be closed properly if we are going to change this logic.

PR #2180, opened 2022-07-25
Reuse zkclient in BestPossibleExternalViewVerifier and fix resource leak

qqu0127:
### Issues

Fix https://github.com/apache/helix/issues/2178 and https://github.com/apache/helix/issues/2179

### Description

A few changes for resource leak fix:
1. Reuse zkclient in verifier
2. Allow passing in zkclient in `ZkBucketDataAccessor`
3. Improve resource closure logic in a few classes with AutoCloseable
4. Change to use external ZkClient in `ServerContext` to fix unit test failure, this also prevents unnecessary zkclient creation

### Tests

Central CI passed https://github.com/apache/helix/runs/7604254487?check_suite_focus=true 

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
@@ -41,6 +41,7 @@ import org.apache.helix.SystemPropertyKeys;
 import org.apache.helix.msdcommon.exception.InvalidRoutingDataException;
 import org.apache.helix.zookeeper.api.client.HelixZkClient;
 import org.apache.helix.zookeeper.api.client.RealmAwareZkClient;
+import org.apache.helix.zookeeper.datamodel.serializer.ZNRecordJacksonSerializer;
 import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.apache.helix.zookeeper.impl.client.FederatedZkClient;
 import org.apache.helix.zookeeper.impl.factory.DedicatedZkClientFactory;
@@ -75,10 +76,11 @@ public class ZkBucketDataAccessor implements BucketDataAccessor, AutoCloseable {
 
   private final int _bucketSize;
   private final long _versionTTLms;
-  private ZkSerializer _zkSerializer;
-  private RealmAwareZkClient _zkClient;
-  private ZkBaseDataAccessor<byte[]> _zkBaseDataAccessor;
-  private Map<String, ScheduledFuture> _gcTaskFutureMap = new HashMap<>();
+  private final ZkSerializer _zkSerializer;
+  private final RealmAwareZkClient _zkClient;
+  private final ZkBaseDataAccessor<byte[]> _zkBaseDataAccessor;
+  private final Map<String, ScheduledFuture> _gcTaskFutureMap = new HashMap<>();
+  private boolean _usesExternalZkClient = false;
 
   /**
    * Constructor that allows a custom bucket size.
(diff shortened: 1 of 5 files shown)


helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
```diff
@@ -456,15 +450,17 @@ public String toString() {
...
+  // TODO: to clean up, finalize is deprecated in Java 9
   @Override
   public void finalize() {
+    super.finalize();
     close();
   }
 
-  private class DryrunWagedRebalancer extends org.apache.helix.controller.rebalancer.waged.ReadOnlyWagedRebalancer {
-    public DryrunWagedRebalancer(String metadataStoreAddress, String clusterName,
+  private static class DryrunWagedRebalancer extends ReadOnlyWagedRebalancer implements AutoCloseable {
```
rahulrane50, line 460: Good call on Autoclosable! Just for my understanding why did we make static ?

rahulrane50 (commented):
Overall great refactoring and great work! Thanks for working on this!

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
```diff
@@ -44,7 +44,7 @@
 import org.slf4j.LoggerFactory;
 
 public abstract class ZkHelixClusterVerifier
-    implements IZkChildListener, IZkDataListener, HelixClusterVerifier {
+    implements IZkChildListener, IZkDataListener, HelixClusterVerifier, AutoCloseable {
```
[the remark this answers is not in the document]
qqu0127: Yes, what you said sounds correct. 
To make sure we are on the same page, `public void close()` has been implemented in this class to close zkclient and with respect to usesExternalZkClient. Its child class, e.g. BestPossibleExternalViewVerifier, there is no need to override again since it has no other resource to close. Implementing `close()` here benefits other children classes as well.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
qqu0127: This is to make it a static nested class. Previously, it's non-static, so it has access to the enclosing class members, which it doesn't need (and actually shouldn't). Moreover, non-static class instance implicitly holds a reference to its outer instance, this is something we don't need and can get away too.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBucketDataAccessor.java
```diff
@@ -135,6 +133,7 @@ private static RealmAwareZkClient createRealmAwareZkClient(String zkAddr) {
       zkClient = DedicatedZkClientFactory.getInstance()
           .buildZkClient(new HelixZkClient.ZkConnectionConfig(zkAddr));
     }
+    zkClient.setZkSerializer(new ByteArraySerializer());
     return zkClient;
```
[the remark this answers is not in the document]
qqu0127: Previously it's using an anonymous class that is essentially the same as `ByteArraySerializer`.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
```diff
@@ -456,15 +450,17 @@ public String toString() {
        + (_resources != null ? Arrays.toString(_resources.toArray()) : "") + "])";
   }
 
+  // TODO: to clean up, finalize is deprecated in Java 9
   @Override
   public void finalize() {
+    super.finalize();
```
jiajunwang, line 456: It might depend on specific logic. But I feel that we should do close() first, and than finalize()? It makes more sense to put finalize the last step.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
```diff
@@ -475,5 +471,10 @@ protected Map<String, ResourceAssignment> computeBestPossibleAssignment(
       return getBestPossibleAssignment(getAssignmentMetadataStore(), currentStateOutput,
           resourceMap.keySet());
     }
+
+    @Override
```
jiajunwang, line 475: Do we really need this? I think it will call super.close by default even you don't have this declaration here.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
qqu0127: That makes sense, thanks.

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
qqu0127: Yeah, it should get from inheritance. I added it for better clarity. I can remove it

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/ZkHelixClusterVerifier.java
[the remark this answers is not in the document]
rahulrane50: Umm I partially agree it's just that making this base class AutoClosable is little risky. But i think it should be okay since it would "auto" close the resource only if it's implementer uses "try-with-resources" statement and this PR only modifies BestPossibleExternalViewVerifier this. One neat i can suggest here is to add clear comments about expectations from it's implementers and stating that it won't share _zkClient.

rahulrane50 (approved):
LGTM!

desaikomal (approved):
LGTM

helix-core/src/main/java/org/apache/helix/tools/ClusterVerifiers/BestPossibleExternalViewVerifier.java
jiajunwang: I would suggest less code if possible : )

qqu0127: Sounds good, it has been updated. Thanks for the review!

qqu0127:
This PR is ready to merge, approved by @desaikomal and @rahulrane50 
Commit message:

Reuse zkclient in BestPossibleExternalViewVerifier and fix resource leak
    
Reuse zkclient in verifier and improve resource closure logic to avoid resource leak.

Issue #2182, opened 2022-08-03
Reduce HelixStateMachineEngine.reset() in participant shutdown


qqu0127:
**Is your feature request related to a problem? Please describe.**
We observed that HelixStateMachineEngine.reset() is invoked 4 times during a participant shutdown and that could take a few minutes in total. 
The root cause might be that `HelixStateMachineEngine` is registered as the message handler for 2 message types. Also there are two threads (main thread and `Cleanup thread for %s-%s-%s` in `ZKHelixManager`) doing similar cleanup upon shutdown.

**Describe the solution you'd like**
We'd like to refactor the logic to optimize the participant shutdown, especially regarding TaskExecutor, to speed up and reduce method calls.

PR #2183, opened 2022-08-03
Optimize HelixTaskExecutor reset() in event of shutdown

qqu0127:
### Issues

Resolve https://github.com/apache/helix/issues/2182 

### Description

`HelixStateMachineEngine.reset()` is invoked 4 times during participant shutdown and that could take up to a few minutes. Two causes:

1. The above is triggered by `HelixTaskExecutor` twice because the same HelixStateMachineEngine is registered for two different message types.
2. There is also an additional thread doing similar cleanup, from `ZKHelixManager.cleanupCallbackHandlers`

This PR refactor the logic in HelixTaskExecutor to reduce unnecessary method call.

### Tests

New test case in helix-core/src/test/java/org/apache/helix/messaging/handling/TestHelixTaskExecutor.java

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -137,6 +137,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
   private final ParticipantStatusMonitor _monitor;
   public static final String MAX_THREADS = "maxThreads";
 
+  private volatile boolean _isResetComplete = false;
   private MessageQueueMonitor _messageQueueMonitor;
   private GenericHelixController _controller;
   private Long _lastSessionSyncTime;
@@ -238,6 +239,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
     }
 
     _isShuttingDown = false;
+    _isResetComplete = false;
 
     MsgHandlerFactoryRegistryItem newItem = new MsgHandlerFactoryRegistryItem(factory, threadpoolSize, resetTimeoutMs);
     MsgHandlerFactoryRegistryItem prevItem = _hdlrFtyRegistry.putIfAbsent(type, newItem);
@@ -678,7 +680,11 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
     }
   }
 
-  void reset() {
+  synchronized void reset() {
+    if (_isResetComplete) {
+      LOG.info("HelixTaskExecutor.reset() has completed, no need to reset again");
+      return;
+    }
     LOG.info("Reset HelixTaskExecutor");
 
     if (_messageQueueMonitor != null) {
@@ -727,6 +733,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
     _knownMessageIds.clear();
 
     _lastSessionSyncTime = null;
+    _isResetComplete = true;
   }
 
   void init() {
@@ -737,6 +744,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
     }
 
     _isShuttingDown = false;
+    _isResetComplete = false;
 
     // Re-init all existing factories
     for (final String msgType : _hdlrFtyRegistry.keySet()) {
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -28,6 +28,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -678,52 +680,56 @@ private void syncFactoryState() {
   }
 
   void reset() {
-    LOG.info("Reset HelixTaskExecutor");
+    synchronized (_resetLock) {
```
desaikomal, line 683: do you need to lock the whole method?? if so, why not add synchronized at method itself.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -827,7 +833,11 @@ public void onMessage(String instanceName, List<Message> messages,
     // and terminate all the thread pools
     // TODO: see if we should have a separate notification call for resetting
     if (changeContext.getType() == Type.FINALIZE) {
-      reset();
+      if (_isShuttingDown) {
```
desaikomal, line 836: where is the flag being set?

qqu0127: https://github.com/apache/helix/blob/132715785e90803ad8991da491f4621db1668fb8/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java#L1454

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
qqu0127: Synchronized method is not used since I don't think this has to be mutual exclusive to other methods synchronized on the instance.
Do we need this synchronization at all? Maybe not, I'm still trying staging changes hoping to better improve the reset().

desaikomal: synchronized method doesn't mean across different methods. only one call to this method. 
I agree that  you don't need synchronization. reset() method has 3 states: No one has invoked it. it is in progress and it has completed the job and nothing to reset. So if you model as appropriate status enum, you can work around where reset() is called only once in shutdown path. my 2cents

qqu0127: Thanks for the comment.
I updated a bit to use synchronization and a boolean state, as I find in_progress state is unnecessary. Hope that makes sense.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -237,6 +239,7 @@ private void registerMessageHandlerFactory(String type, MessageHandlerFactory fa
     }
 
     _isShuttingDown = false;
+    _isResetComplete = false;
```
junkaixue, line 242: NIT: this is more like an action flag instead of state flag. Personally, I prefer flag like isCleanState or something else.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -694,16 +701,19 @@ void reset() {
...
+          .map(MsgHandlerFactoryRegistryItem::factory)
+          .distinct()
+          .filter(Objects::nonNull)
+          .forEach(factory -> {
+            try {
+              factory.reset();
+            } catch (Exception ex) {
+              LOG.error("Failed to reset the factory {}.", factory.toString(), ex);
+            }
+          });
```
junkaixue, line 716: I would suggest to make parallel reset in a different PR. This requires more thinking. 1) we cannot guarantee there is not dependencies and user are not leveraging dependencies because of sequential reset. 2) if there is a problem for parallel reset, you can just revert reset PR instead of entire PR to revert dedup logic.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -734,6 +744,7 @@ void init() {
     }
 
     _isShuttingDown = false;
+    _isResetComplete = false;
```
junkaixue, line 747: Why not put in onMessage? OnMessage is the place starting STs and changing the states.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -827,7 +838,11 @@ public void onMessage(String instanceName, List<Message> messages,
     // and terminate all the thread pools
     // TODO: see if we should have a separate notification call for resetting
     if (changeContext.getType() == Type.FINALIZE) {
-      reset();
+      if (_isShuttingDown) {
```
junkaixue, line 841: Since we already use new flag to protect it. Let's not use this isShutdowning to do another filtering.

There could be race condition happening. The flag started with flag and if there is a session timeout and restart everything. Before this flag set to true in the registryFactory, you get finalize message from ZK server. You will bypass the reset and use the old data.

If we already decide to use lowest component to do dedup, let's not add extra optimization in up layer, which may have potential issues.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
qqu0127: Correct me if I'm wrong. I think the problem is with `init()`, it initializes threadpools that need to be shutdown in reset(), so reset() should still be run even without state transition.
What `reset()` does should actually be shutdown() or close().

junkaixue (approved):

qqu0127:
This PR is ready to merge, approved by @junkaixue 
Commit message:

Optimize HelixTaskExecutor reset() in event of shutdown

Some instances maybe reset() multiple times during participant shutdown.
This commit refactor logic in HelixTaskExecutor to reduce unnecessary
method call.

PR #2186, opened 2022-08-04
Fixes #2185 NullPointerException in ResourceUtil for admin-webapp

### LINKED ISSUE #2185: admin-webapp throws NullPointerException for every request
### Describe the bug
Every API call the admin-webapp throws a NullPointerException. It's due to the fact that mapper [here](https://github.com/apache/helix/blob/master/helix-admin-webapp/src/main/java/org/apache/helix/webapp/resources/ResourceUtil.java#L50) is used in the line above it. Since Java's initialization of static members follows the order in which they're declared, the mapper hasn't been instantiated yet, causing the NullPointerException [here](https://github.com/apache/helix/blob/11846ccbfe30820a4403b90e6b115c9cdb56416a/helix-admin-webapp/src/main/java/org/apache/helix/webapp/resources/ResourceUtil.java#L111)  when line 49 of that file is evaluated.

### To Reproduce
Use any API in the admin-webapp which will return a 500 (due to the NullPointerException)

### Expected behavior
The API shouldn't return a 500 and instead return what it should.

### Additional context
Looks like this has been broken since 2020, but due to the Rest App 2.0, probably went unnoticed.

wmorgan6796:
### Issues
Fixes #2185 

### Description

Fixes a NullPointerException present in ResourceUtil.java. Its due to the order in which static class members are initialized and leads to the given exception because a different static object that is created before the `mapper` object uses the mapper (on line 49).

### Tests

Tests that were previously broken should pass.

Does not break backwards compatability


helix-admin-webapp/src/main/java/org/apache/helix/webapp/resources/ResourceUtil.java
```diff
@@ -45,9 +45,9 @@
 
 public class ResourceUtil {
   private static final Logger LOG = LoggerFactory.getLogger(ResourceUtil.class);
+  private static ObjectMapper mapper = new ObjectMapper();
   private static final String EMPTY_ZNRECORD_STRING =
       objectToJson(ClusterRepresentationUtil.EMPTY_ZNRECORD);
```
qqu0127, line 50: Thanks for the contribution.
I think you can safely remove `EMPTY_ZNRECORD_STRING` and `objectToJson`, I don't think they are ever used.

wmorgan6796: I'll try it out.

wmorgan6796: Was able to remove and added a `final` to the mapper (per Intellij's recommendations)

junkaixue (approved):

PR #2189, opened 2022-08-09
WAGED pipeline change

NealSun96:
### Issues

https://github.com/apache/helix/issues/2287

### Description

NOTE: no test or metrics are included in this PR due to PR size. Once reviewed, I will merge this PR into a feature branch instead of master branch, and a subsequent PR will be created for tests and metrics. The test changes here are for test passing only. 

This PR includes a couple feature changes:
1. Introducing the emergency rebalance process. 
2. Move partial rebalance to asynchronous. 
3. Add conditions to skip rebalance overwrite. 
4. Move cluster model generation to asynchronous global rebalance thread. 
5. Changes some API to AssignmentMetadata for operation purposes. For example, the persistX functions no longer include the "check for same assignment" logic for early termination. This allows us to a) accurately track write latency and b) tie boolean return values to actual persist failures. 

### Tests

No test or metrics are included in this PR due to PR size. A subsequent PR will be submitted later. 

```
[ERROR] Tests run: 1318, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 15,419.043 s <<< FAILURE! - in TestSuite
[ERROR] testP2PStateTransitionEnabled(org.apache.helix.integration.messaging.TestP2PNoDuplicatedMessage)  Time elapsed: 31.816 s  <<< FAILURE!
java.lang.AssertionError: There are duplicated transition messages sent at same time! expected:<0> but was:<3>
	at org.apache.helix.integration.messaging.TestP2PNoDuplicatedMessage.testP2PStateTransitionEnabled(TestP2PNoDuplicatedMessage.java:183)

[ERROR] testAggregateMetrics(org.apache.helix.monitoring.mbeans.TestClusterAggregateMetrics)  Time elapsed: 1.626 s  <<< FAILURE!
	at org.apache.helix.monitoring.mbeans.TestClusterAggregateMetrics.testAggregateMetrics(TestClusterAggregateMetrics.java:186)

[INFO]
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestP2PNoDuplicatedMessage.testP2PStateTransitionEnabled:183 There are duplicated transition messages sent at same time! expected:<0> but was:<3>
[ERROR]   TestClusterAggregateMetrics.testAggregateMetrics:186 expected:<true> but was:<false>
[INFO]
[ERROR] Tests run: 1318, Failures: 2, Errors: 0, Skipped: 0
```

```
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 208.345 s - in TestSuite
[INFO]
[INFO] Results:
[INFO]
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -111,6 +111,10 @@ public class AssignmentMetadataStore {
     }
   }
 
+  /**
+   * Persist a new baseline assignment to metadata store first, then to memory
+   * @param globalBaseline
+   */
   public synchronized void persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
     // write to metadata store
     persistAssignmentToMetadataStore(globalBaseline, _baselinePath, BASELINE_KEY);
@@ -119,8 +123,12 @@ public class AssignmentMetadataStore {
     getBaseline().putAll(globalBaseline);
   }
 
-  public synchronized void persistBestPossibleAssignment(
-      Map<String, ResourceAssignment> bestPossibleAssignment) {
+  /**
+   * Persist a new best possible assignment to metadata store first, then to memory.
+   * Increment best possible version by 1 - this is a high priority in-memory write.
+   * @param bestPossibleAssignment
+   */
+  public synchronized void persistBestPossibleAssignment(Map<String, ResourceAssignment> bestPossibleAssignment) {
     // write to metadata store
     persistAssignmentToMetadataStore(bestPossibleAssignment, _bestPossiblePath, BEST_POSSIBLE_KEY);
     // write to memory
@@ -131,6 +139,7 @@ public class AssignmentMetadataStore {
 
   /**
    * Attempts to persist Best Possible Assignment in memory from an asynchronous thread.
+   * Persist only happens when the provided version is not stale - this is a low priority in-memory write.
    * @param bestPossibleAssignment - new assignment to be persisted
    * @param newVersion - attempted new version to write. This version is obtained earlier from getBestPossibleVersion()
    * @return true if the attempt succeeded, false otherwise.
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -50,6 +50,7 @@ public class AssignmentMetadataStore {
   private String _bestPossiblePath;
   protected Map<String, ResourceAssignment> _globalBaseline;
   protected Map<String, ResourceAssignment> _bestPossibleAssignment;
+  private int _bestPossibleVersion = 0;
 
   AssignmentMetadataStore(String metadataStoreAddrs, String clusterName) {
     this(new ZkBucketDataAccessor(metadataStoreAddrs), clusterName);
@@ -92,46 +93,14 @@ public class AssignmentMetadataStore {
   }
 
   /**
-   * @return true if a new baseline was persisted.
-   * @throws HelixException if the method failed to persist the baseline.
-   */
-  public synchronized boolean persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
-    return persistAssignment(globalBaseline, getBaseline(), _baselinePath, BASELINE_KEY);
-  }
-
-  /**
-   * @return true if a new best possible assignment was persisted.
-   * @throws HelixException if the method failed to persist the baseline.
-   */
-  public synchronized boolean persistBestPossibleAssignment(
-      Map<String, ResourceAssignment> bestPossibleAssignment) {
-    return persistAssignment(bestPossibleAssignment, getBestPossibleAssignment(), _bestPossiblePath,
-        BEST_POSSIBLE_KEY);
-  }
-
-  public synchronized void clearAssignmentMetadata() {
-    persistAssignment(Collections.emptyMap(), getBaseline(), _baselinePath, BASELINE_KEY);
-    persistAssignment(Collections.emptyMap(), getBestPossibleAssignment(), _bestPossiblePath,
-        BEST_POSSIBLE_KEY);
-  }
-
-  /**
    * @param newAssignment
-   * @param cachedAssignment
    * @param path the path of the assignment record
    * @param key  the key of the assignment in the record
-   * @return true if a new assignment was persisted.
+   * @throws HelixException if the method failed to persist the baseline.
    */
-  // TODO: Enhance the return value so it is more intuitive to understand when the persist fails and
-  // TODO: when it is skipped.
-  private boolean persistAssignment(Map<String, ResourceAssignment> newAssignment,
-      Map<String, ResourceAssignment> cachedAssignment, String path,
+  private void persistAssignmentToMetadataStore(Map<String, ResourceAssignment> newAssignment, String path,
       String key) {
     // TODO: Make the write async?
-    // If the assignment hasn't changed, skip writing to metadata store
-    if (compareAssignments(cachedAssignment, newAssignment)) {
-      return false;
-    }
     // Persist to ZK
     HelixProperty combinedAssignments = combineAssignments(key, newAssignment);
     try {
@@ -141,11 +110,54 @@ public class AssignmentMetadataStore {
       throw new HelixException(
           String.format("Failed to persist %s assignment to path %s", key, path), e);
     }
+  }
 
-    // Update the in-memory reference
-    cachedAssignment.clear();
-    cachedAssignment.putAll(newAssignment);
-    return true;
+  public synchronized void persistBaseline(Map<String, ResourceAssignment> globalBaseline) {
+    // write to metadata store
+    persistAssignmentToMetadataStore(globalBaseline, _baselinePath, BASELINE_KEY);
+    // write to memory
+    getBaseline().clear();
+    getBaseline().putAll(globalBaseline);
+  }
+
+  public synchronized void persistBestPossibleAssignment(
+      Map<String, ResourceAssignment> bestPossibleAssignment) {
+    // write to metadata store
+    persistAssignmentToMetadataStore(bestPossibleAssignment, _bestPossiblePath, BEST_POSSIBLE_KEY);
+    // write to memory
+    getBestPossibleAssignment().clear();
+    getBestPossibleAssignment().putAll(bestPossibleAssignment);
+    _bestPossibleVersion++;
+  }
+
+  /**
+   * Attempts to persist Best Possible Assignment in memory from an asynchronous thread.
+   * @param bestPossibleAssignment - new assignment to be persisted
+   * @param newVersion - attempted new version to write. This version is obtained earlier from getBestPossibleVersion()
+   * @return true if the attempt succeeded, false otherwise.
+   */
+  public synchronized boolean asyncPersistBestPossibleAssignmentInMemory(
+      Map<String, ResourceAssignment> bestPossibleAssignment, int newVersion) {
+    // Check if the version is stale by this point
+    if (newVersion > _bestPossibleVersion) {
+      getBestPossibleAssignment().clear();
+      getBestPossibleAssignment().putAll(bestPossibleAssignment);
+      _bestPossibleVersion = newVersion;
+      return true;
+    }
+
+    return false;
+  }
+
+  public int getBestPossibleVersion() {
+    return _bestPossibleVersion;
+  }
+
+  public synchronized void clearAssignmentMetadata() {
+    persistAssignmentToMetadataStore(Collections.emptyMap(), _baselinePath, BASELINE_KEY);
+    persistAssignmentToMetadataStore(Collections.emptyMap(), _bestPossiblePath, BEST_POSSIBLE_KEY);
+    getBaseline().clear();
+    getBestPossibleAssignment().clear();
   }
 
   protected synchronized void reset() {
@@ -206,10 +218,18 @@ public class AssignmentMetadataStore {
    * @param newAssignment
    * @return true if they are the same. False otherwise or oldAssignment is null
    */
-  protected boolean compareAssignments(Map<String, ResourceAssignment> oldAssignment,
+  private boolean compareAssignments(Map<String, ResourceAssignment> oldAssignment,
       Map<String, ResourceAssignment> newAssignment) {
     // If oldAssignment is null, that means that we haven't read from/written to
     // the metadata store yet. In that case, we return false so that we write to metadata store.
     return oldAssignment != null && oldAssignment.equals(newAssignment);
+  }
+
+  protected boolean compareBaseline(Map<String, ResourceAssignment> newBaseline) {
+    return compareAssignments(getBaseline(), newBaseline);
+  }
+
+  protected boolean compareBestPossibleAssignment(Map<String, ResourceAssignment> newBaseline) {
+    return compareAssignments(getBestPossibleAssignment(), newBaseline);
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -50,7 +50,7 @@ public class AssignmentMetadataStore {
   private String _bestPossiblePath;
   protected Map<String, ResourceAssignment> _globalBaseline;
   protected Map<String, ResourceAssignment> _bestPossibleAssignment;
-  private int _bestPossibleVersion = 0;
+  protected int _bestPossibleVersion = 0;
 
   AssignmentMetadataStore(String metadataStoreAddrs, String clusterName) {
     this(new ZkBucketDataAccessor(metadataStoreAddrs), clusterName);
@@ -98,8 +98,8 @@ public class AssignmentMetadataStore {
    * @param key  the key of the assignment in the record
    * @throws HelixException if the method failed to persist the baseline.
    */
-  private void persistAssignmentToMetadataStore(Map<String, ResourceAssignment> newAssignment, String path,
-      String key) {
+  private void persistAssignmentToMetadataStore(Map<String, ResourceAssignment> newAssignment, String path, String key)
+      throws HelixException {
     // TODO: Make the write async?
     // Persist to ZK
     HelixProperty combinedAssignments = combineAssignments(key, newAssignment);
@@ -107,8 +107,7 @@ public class AssignmentMetadataStore {
       _dataAccessor.compressedBucketWrite(path, combinedAssignments);
     } catch (IOException e) {
       // TODO: Improve failure handling
-      throw new HelixException(
-          String.format("Failed to persist %s assignment to path %s", key, path), e);
+      throw new HelixException(String.format("Failed to persist %s assignment to path %s", key, path), e);
     }
   }
 
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -233,11 +233,11 @@ public class AssignmentMetadataStore {
     return oldAssignment != null && oldAssignment.equals(newAssignment);
   }
 
-  protected boolean compareBaseline(Map<String, ResourceAssignment> newBaseline) {
+  protected boolean isBaselineChanged(Map<String, ResourceAssignment> newBaseline) {
     return compareAssignments(getBaseline(), newBaseline);
   }
 
-  protected boolean compareBestPossibleAssignment(Map<String, ResourceAssignment> newBaseline) {
+  protected boolean isBestPossibleChanged(Map<String, ResourceAssignment> newBaseline) {
     return compareAssignments(getBestPossibleAssignment(), newBaseline);
   }
 }
--- a/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
@@ -106,7 +106,6 @@ public class AssignmentMetadataStore {
     try {
       _dataAccessor.compressedBucketWrite(path, combinedAssignments);
     } catch (IOException e) {
-      // TODO: Improve failure handling
       throw new HelixException(String.format("Failed to persist %s assignment to path %s", key, path), e);
     }
   }
@@ -144,7 +143,7 @@ public class AssignmentMetadataStore {
    * @param newVersion - attempted new version to write. This version is obtained earlier from getBestPossibleVersion()
    * @return true if the attempt succeeded, false otherwise.
    */
-  public synchronized boolean asyncPersistBestPossibleAssignmentInMemory(
+  public synchronized boolean asyncUpdateBestPossibleAssignmentCache(
       Map<String, ResourceAssignment> bestPossibleAssignment, int newVersion) {
     // Check if the version is stale by this point
     if (newVersion > _bestPossibleVersion) {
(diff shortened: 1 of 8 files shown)


helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -562,6 +590,57 @@ private Map<String, ResourceAssignment> partialRebalance(
...
+
+    // Only calculate if there are illegal placements, meaning the cluster model is non null;
+    // otherwise, start partial rebalance and return the current best possible.
+    if (clusterModel == null) {
+      // Perform partial rebalance for a new best possible assignment
+      partialRebalance(clusterData, resourceMap, activeNodes, currentStateOutput, algorithm);
+      return currentBestPossibleAssignment;
+    }
+
+    _asyncPartialRebalanceResult.cancel(true);
```
qqu0127, line 626: nit: some logs might be useful for cases it's interrupted VS already done?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -562,6 +590,57 @@ private Map<String, ResourceAssignment> partialRebalance(
...
+          .generateClusterModelForEmergencyRebalance(clusterData, resourceMap, activeNodes,
+              currentBestPossibleAssignment);
+    } catch (Exception ex) {
+      throw new HelixRebalanceException("Failed to generate cluster model for emergency rebalance.",
+          HelixRebalanceException.Type.INVALID_CLUSTER_STATUS, ex);
+    }
+
+    // Only calculate if there are illegal placements, meaning the cluster model is non null;
+    // otherwise, start partial rebalance and return the current best possible.
+    if (clusterModel == null) {
```
qqu0127, line 620: Just thinking aloud, it feels weird to have a null model. Is there a better way than using null value? Say, check whether there is illegal placements here? 
If it's too much trouble, let's document this nullable in the method signature.

qqu0127 (commented):
Thanks for writing this up. Some quick comments as below.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -358,9 +361,10 @@ protected Map<String, ResourceAssignment> computeBestPossibleAssignment(
       throws HelixRebalanceException {
     // Perform global rebalance for a new baseline assignment
     globalRebalance(clusterData, resourceMap, currentStateOutput, algorithm);
-    // Perform partial rebalance for a new best possible assignment
+    // Perform emergency rebalance for a new best possible assignment
```
desaikomal, line 364: Q: looks like this is for case(1)?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -534,8 +535,34 @@ private Map<String, ResourceAssignment> partialRebalance(
       throw new HelixRebalanceException("Failed to generate cluster model for partial rebalance.",
           HelixRebalanceException.Type.INVALID_CLUSTER_STATUS, ex);
     }
-    Map<String, ResourceAssignment> newAssignment = calculateAssignment(clusterModel, algorithm);
 
+    _asyncPartialRebalanceResult = _bestPossibleCalculateExecutor.submit(() -> {
```
desaikomal, line 539: would be useful to have log message about calculating Emergency pipeline. (i know method already has log statement, but ability to trace call-path will be useful)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
desaikomal: +1

desaikomal: do we cancel the earlier partialRebalance because the data has changed?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
```diff
@@ -391,6 +406,34 @@ private static Set<AssignableReplica> findToBeAssignedReplicasByClusterChanges(
     return toBeAssignedReplicas;
   }
 
+  private static Set<AssignableReplica> findToBeAssignedReplicasIllegalPlacements(
```
desaikomal, line 409: nit: useful to have some high level comments for method.
if i understand, it is trying to find 'currentIllegalPlacements'

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
```diff
@@ -391,6 +406,34 @@ private static Set<AssignableReplica> findToBeAssignedReplicasByClusterChanges(
...
+    for (String resourceName : replicaMap.keySet()) {
+      Map<String, Map<String, Set<String>>> currentPartitionStateMap =
+          getValidStateInstanceMap(currentAssignment.get(resourceName), activeInstances);
+      for (AssignableReplica replica : replicaMap.get(resourceName)) {
+        String partitionName = replica.getPartitionName();
+        String replicaState = replica.getReplicaState();
+        Set<String> currentAllocations =
+            currentPartitionStateMap.getOrDefault(partitionName, Collections.emptyMap())
+                .getOrDefault(replicaState, Collections.emptySet());
+        List<String> currentAllocationsList = new ArrayList<>(currentAllocations);
```
desaikomal, line 425: i am sorry, but why are we creating this new List<>?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -562,6 +590,57 @@ private Map<String, ResourceAssignment> partialRebalance(
...
     LOG.info("Finish calculating the new best possible assignment.");
+
+    if (isBestPossibleChanged) {
+      LOG.info("Schedule a new rebalance after the new best possible calculation has finished.");
+      RebalanceUtil.scheduleOnDemandPipeline(clusterData.getClusterName(), 0L, false);
+    }
+  }
+
+  private Map<String, ResourceAssignment> emergencyRebalance(
+      ResourceControllerDataProvider clusterData, Map<String, Resource> resourceMap,
```
xyuanlu, line 601: May I suggest let's have an emergency rebalance counter as well? (similar to the _partialRebalanceCounter)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -562,6 +590,57 @@ private Map<String, ResourceAssignment> partialRebalance(
...
+      try {
+        _writeLatency.startMeasuringLatency();
+        _assignmentMetadataStore.persistBestPossibleAssignment(newAssignment);
+        _writeLatency.endMeasuringLatency();
+      } catch (Exception ex) {
+        throw new HelixRebalanceException("Failed to persist the new best possible assignment.",
+            HelixRebalanceException.Type.INVALID_REBALANCER_STATUS, ex);
+      }
+    } else {
+      LOG.debug("Assignment Metadata Store is null. Skip persisting the best possible assignment.");
```
xyuanlu, line 640: Q: Shouldn't this be an error or at least warning log? Why we have debug here?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
NealSun96: @qqu0127 Good point. Will add that for better visibility. 
@desaikomal Attempting to cancel here because there have been illegal placements, therefore the previous calculation can no longer be applied. This cancel operation is to cut down unnecessary work.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
NealSun96: This PR is a benchmarking prototype at this moment, so I will keep this behavior. 
When it's production ready, I will change the null behavior.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
NealSun96: Both cases start with emergency rebalance - it's an alternative to the original design, which I think simplifies things.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
NealSun96: Could you elaborate on where logs should happen? This is the partial rebalance function.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
NealSun96: The code tries to remove any element from the set, and the list is created to mimic the behavior on the partial rebalance side. However, I will find a better way to remove the element.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
NealSun96: This part is copied from the partial rebalance code - this log is unlikely to happen so a debug log makes sense to me.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -410,43 +429,22 @@ protected List<HelixRebalanceException.Type> failureTypesToPropagate() {
...
-      RebalanceAlgorithm algorithm)
-      throws HelixRebalanceException {
+  private void globalRebalance(ResourceControllerDataProvider clusterData, Map<String, Resource> resourceMap,
+      final CurrentStateOutput currentStateOutput, RebalanceAlgorithm algorithm) throws HelixRebalanceException {
     _changeDetector.updateSnapshots(clusterData);
     // Get all the changed items' information. Filter for the items that have content changed.
-    final Map<HelixConstants.ChangeType, Set<String>> clusterChanges =
-        _changeDetector.getAllChanges();
-
-    if (clusterChanges.keySet().stream()
```
NealSun96, line 422: Moving cluster model generation to the concrete and async doGlobalRebalance to improve performance

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -469,27 +467,40 @@ private void globalRebalance(ResourceControllerDataProvider clusterData,
...
+          ClusterModelProvider.generateClusterModelForBaseline(clusterData, resourceMap, clusterData.getAllInstances(),
+              clusterChanges, currentBaseline);
+    } catch (Exception ex) {
+      throw new HelixRebalanceException("Failed to generate cluster model for global rebalance.",
+          HelixRebalanceException.Type.INVALID_CLUSTER_STATUS, ex);
+    }
+
     Map<String, ResourceAssignment> newBaseline = calculateAssignment(clusterModel, algorithm);
+    boolean isBaselineChanged =
+        _assignmentMetadataStore != null && !_assignmentMetadataStore.compareBaseline(newBaseline);
```
NealSun96, line 498: I made similar changes like this - the logic has not changed: if the assignment is the same, do not persist. However, I want to compare result outside of the persistX function for 2 reasons: 
1. The async persist function (in partial rebalance) can return a boolean that indicates stale assignment for metric purpose (otherwise, I cannot differentiate between stale data and same assignment, since both return false).
2. The write latency metric is now correctly measuring writes only, instead of including data points for early returns due to same assignments.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
```diff
@@ -391,6 +418,46 @@ private static Set<AssignableReplica> findToBeAssignedReplicasByClusterChanges(
...
+      for (AssignableReplica replica : replicaMap.get(resourceName)) {
+        String partitionName = replica.getPartitionName();
+        String replicaState = replica.getReplicaState();
+        Set<String> currentAllocations =
+            stateInstanceMap.getOrDefault(partitionName, Collections.emptyMap())
+                .getOrDefault(replicaState, Collections.emptySet());
+        if (!currentAllocations.isEmpty()) {
+          String allocatedInstance = currentAllocations.iterator().next();
+          if (activeInstances.contains(allocatedInstance)) {
+            allocatedReplicas.computeIfAbsent(allocatedInstance, key -> new HashSet<>()).add(replica);
```
qqu0127, line 448: for my own understanding, do we only check the first allocatedInstance? Any chance there is more than one element in the set?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -501,21 +512,64 @@ private void calculateAndUpdateBaseline(ClusterModel clusterModel, RebalanceAlgo
...
+  private void partialRebalance(
       ResourceControllerDataProvider clusterData, Map<String, Resource> resourceMap,
       Set<String> activeNodes, final CurrentStateOutput currentStateOutput,
       RebalanceAlgorithm algorithm)
       throws HelixRebalanceException {
+    // If partial rebalance is async and the previous result is not completed yet,
+    // do not start another partial rebalance.
+    if (_asyncPartialRebalanceEnabled && _asyncPartialRebalanceResult != null
+        && !_asyncPartialRebalanceResult.isDone()) {
+      return;
```
qqu0127, line 530: For my own understanding, does this mean we have more frequent emergency rebalance? 

also nit: Let's log something here.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
```diff
@@ -501,21 +512,64 @@ private void calculateAndUpdateBaseline(ClusterModel clusterModel, RebalanceAlgo
...
-    if (isBaselineChanged && shouldSchedulePartialRebalance) {
+    if (isBaselineChanged && shouldTriggerMainPipeline) {
       LOG.info("Schedule a new rebalance after the new baseline calculation has finished.");
-      RebalanceUtil.scheduleOnDemandPipeline(clusterName, 0L, false);
+      RebalanceUtil.scheduleOnDemandPipeline(clusterData.getClusterName(), 0L, false);
     }
   }
 
-  private Map<String, ResourceAssignment> partialRebalance(
+  private void partialRebalance(
```
qqu0127, line 521: General comment on the overall structure, open to discussion -- 
Can we make partial rebalance a standalone class? This `WagedRebalancer` class is getting more and more complicated, and "partial rebalance" is more like a separate module now, it has its own computation logic and execution context.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -92,60 +93,79 @@ public synchronized Map<String, ResourceAssignment> getBestPossibleAssignment()
...
-    }
-    // Persist to ZK
-    HelixProperty combinedAssignments = combineAssignments(key, newAssignment);
-    try {
-      _dataAccessor.compressedBucketWrite(path, combinedAssignments);
-    } catch (IOException e) {
-      // TODO: Improve failure handling
-      throw new HelixException(
-          String.format("Failed to persist %s assignment to path %s", key, path), e);
+  public synchronized boolean asyncPersistBestPossibleAssignmentInMemory(
```
qqu0127, line 147: Is it applicable to use finer scoped lock? same for other methods as well.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -92,60 +93,79 @@ public synchronized Map<String, ResourceAssignment> getBestPossibleAssignment()
...
+  private void persistAssignmentToMetadataStore(Map<String, ResourceAssignment> newAssignment, String path, String key)
+      throws HelixException {
+    // TODO: Make the write async?
+    // Persist to ZK
+    HelixProperty combinedAssignments = combineAssignments(key, newAssignment);
+    try {
+      _dataAccessor.compressedBucketWrite(path, combinedAssignments);
+    } catch (IOException e) {
+      // TODO: Improve failure handling
+      throw new HelixException(String.format("Failed to persist %s assignment to path %s", key, path), e);
```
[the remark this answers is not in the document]
NealSun96: That's a copied TODO from earlier times, not too sure what it's referring to. :)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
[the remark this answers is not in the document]
NealSun96: How about `asyncUpdateBestPossibleAssignmentCache`? I really want to emphasis that this is an async function that is a low priority update.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -206,10 +226,18 @@ private Map<String, ResourceAssignment> splitAssignments(HelixProperty property)
...
    */
-  protected boolean compareAssignments(Map<String, ResourceAssignment> oldAssignment,
+  private boolean compareAssignments(Map<String, ResourceAssignment> oldAssignment,
       Map<String, ResourceAssignment> newAssignment) {
     // If oldAssignment is null, that means that we haven't read from/written to
     // the metadata store yet. In that case, we return false so that we write to metadata store.
     return oldAssignment != null && oldAssignment.equals(newAssignment);
   }
+
+  protected boolean compareBaseline(Map<String, ResourceAssignment> newBaseline) {
```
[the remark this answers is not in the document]
NealSun96: I like that.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
NealSun96: Since the loop is based on per replica, we're checking instances one at a time; there could be 2 instances in the currentAllocations set, but this set will also be visited 2 times if there're 2 FOLLOWER replica.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
NealSun96: More frequent emergency rebalance? Do you mean the sync pipeline is triggered more often? I don't fully understand the question. 

I will add more logs in a followup PR. :)

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
NealSun96: It would be strange to me that "partial rebalance" lives outside of WagedRebalancer, because it will never be meaningful outside of WAGED context. 

I could be biased here since this class was written like this in the beginning.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
NealSun96: I think so, but I don't see the need for it. Keep persisting synchronized is less error prone without much performance sacrifice (cache refresh is very fast, so it wouldn't block zookeeper persist for too long).

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
qqu0127: Yes, I'm trying to understand what it means if we enter this block. Essentially, if the sync pipeline is triggered too frequently, we'll see more attempts to start partial rebalance while there is one still in progress. Is that right?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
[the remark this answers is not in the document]
jiajunwang: I guess that means you can just remove it. Was it referring to some transactional support that on error, we should keep the data in a consistent way?

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
NealSun96: That's correct, but that also wouldn't actually spawn a thread.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/WagedRebalancer.java
qqu0127: In general, that will be more object oriented and potentially easier for mock up and test. But you are also right, so I'll let you make the call.

NealSun96: I gave a short answer here. I think the bigger problem is how we want to design this new class and how it works with the original WagedRebalancer class - that in itself is a big conversation. Right now the state of WagedRebalancer doesn't warrant that redesign, which I think you would agree too.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
qqu0127: then I'm confused, the code is using `currentAllocations.iterator().next();` each time, from your description it should be reusing the same iterator.

NealSun96: That's because `currentAllocations` is losing one element each time a corresponding replica is there, if you look a few lines down. So `currentAllocations.iterator().next()` is just the first element (whatever that means in a set) of this set.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
```diff
@@ -206,10 +226,18 @@ private Map<String, ResourceAssignment> splitAssignments(HelixProperty property)
...
     return oldAssignment != null && oldAssignment.equals(newAssignment);
   }
+
+  protected boolean isBaselineChanged(Map<String, ResourceAssignment> newBaseline) {
+    return compareAssignments(getBaseline(), newBaseline);
+  }
+
+  protected boolean isBestPossibleChanged(Map<String, ResourceAssignment> newBaseline) {
+    return compareAssignments(getBestPossibleAssignment(), newBaseline);
+  }
```
qqu0127, line 242: compareAssignments `@return true if they are the same`, so actually we should do a negation here and in its downstream?
also the method `compareAssignments` is very shallow, we could probably just remove it and use `Objects.equals(a, b)` instead.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
qqu0127: Fair enough, the current state looks ok to me. Thanks.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/model/ClusterModelProvider.java
qqu0127: I see the logic here, thanks for the elaboration. 
One thing to improve is not to remove and add element one-by-one, but in one pass for each resource. Not sure if that's applicable at all, so just nice to have. Feel free to close this.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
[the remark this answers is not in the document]
qqu0127: Sure, that sounds good. Thanks.

helix-core/src/main/java/org/apache/helix/controller/rebalancer/waged/AssignmentMetadataStore.java
NealSun96: Good catch, thank you. Also removing compareAssignments, since getBaseline() will never be null.

NealSun96:
Review has been completed - I will proceed to merge this into a feature branch. The feature branch will merge back into master after tests and metrics are added.

PR #2201, opened 2022-08-31
Removed Error Log to Avoid Spam Logging

Marcosrico:
Deleted logging error in `ZkClient` delete method to eliminate log noise.

### Issues

https://github.com/apache/helix/issues/2202#issue-1357887240
References an issue where despite a successful retry, failure to remove a node still throws an error log.

### Description

https://github.com/apache/helix/issues/2202#issue-1357887240

Removed a failed remove log in `ZkClient.delete()`. This log error shows when there is an attempt to delete a parent ZNode with non-empty children. However, it automatically calls another function that recursively deletes the children of the node and in the case this operation is successful, having a stack-trace error from before can confuse users of a problem that is in fact trivial. By deleting the log shown when the above error occurs, we eliminate these misleading logs and avoid log flooding. We can do this as the exception is already being thrown so it is not necessary to log it. 

### Tests

Added `testDeleteNodeWithChildren()` in `TestZkBaseDataAccessor.java`

The test creates a 10 instance message records. It proceeds to call remove on the parent of those children. The test checks whether remove returns true, which it does. However, the purpose of it is to simulate the initial error that brought upon this change. Going through the debugger, we see that it does go through the right path and effectively does not throw any logs (as it has been removed).

```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 57.175 s - in org.apache.helix.manager.zk.TestZkBaseDataAccessor
[INFO] Results:
```

Local test result for failed CI tests:
For `testZKHelixManager` and `afterClass`

```
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 208.496 s - in org.apache.helix.integration.multizk.TestMultiZkConectionConfig
[INFO] Results:
```

For `testZkClientCreationMultiZk`

```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 79.998 s - in org.apache.helix.task.TestTaskStateModelFactory
[INFO] Results:
```

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -726,7 +726,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
       // case, we try to delete recursively
       _zkClient.delete(path);
     } catch (ZkException e) {
-      LOG.debug("Failed to delete {} with opts {}, err: {}. Try recursive delete", path, options,
+      LOG.info("Failed to delete {} with opts {}, err: {}. Try recursive delete", path, options,
           e.getMessage());
       try {
         _zkClient.deleteRecursively(path);
--- a/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
+++ b/helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
@@ -726,7 +726,7 @@ public class ZkBaseDataAccessor<T> implements BaseDataAccessor<T> {
       // case, we try to delete recursively
       _zkClient.delete(path);
     } catch (ZkException e) {
-      LOG.info("Failed to delete {} with opts {}, err: {}. Try recursive delete", path, options,
+      LOG.debug("Failed to delete {} with opts {}, err: {}. Try recursive delete", path, options,
           e.getMessage());
       try {
         _zkClient.deleteRecursively(path);
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
@@ -434,6 +434,29 @@ public class TestZkBaseDataAccessor extends ZkUnitTestBase {
   }
 
   @Test
+  public void testDeleteLogging() {
+    String root = _rootPath;
+    _gZkClient.deleteRecursively("/" + root);
+
+    ZkBaseDataAccessor<ZNRecord> accessor = new ZkBaseDataAccessor<>(_gZkClient);
+
+    // CreateChildren
+    List<ZNRecord> records = new ArrayList<>();
+    List<String> paths = new ArrayList<>();
+    for (int i = 0; i < 10; i++) {
+      String msgId = "msg_" + i;
+      paths.add(PropertyPathBuilder.instanceMessage(root, "host_1", msgId));
+      records.add(new ZNRecord(msgId));
+    }
+    boolean[] success = accessor.createChildren(paths, records, AccessOption.PERSISTENT);
+
+    // Attempt to remove parent. Shouldn't throw an error or warning log.
+    // Should return True if recursive deletion succeeds.
+    Assert.assertTrue(accessor.remove(PropertyPathBuilder.instanceMessage(root, "host_1"), 0),
+            "Should return True despite log errors.");
+  }
+
+  @Test
   public void testSyncGet() {
     String className = TestHelper.getTestClassName();
     String methodName = TestHelper.getTestMethodName();
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
@@ -454,6 +454,14 @@ public class TestZkBaseDataAccessor extends ZkUnitTestBase {
     // Should return True if recursive deletion succeeds.
     Assert.assertTrue(accessor.remove(PropertyPathBuilder.instanceMessage(root, "host_1"), 0),
             "Should return True despite log errors.");
+
+    // Assert child message nodes were removed when calling remove on parent
+    for (int i = 0; i < 10; i++) {
+      String msgId = "msg_" + i;
+      String path = PropertyPathBuilder.instanceMessage(root, "host_1", msgId);
+      boolean pathExists = _gZkClient.exists(path);
+      Assert.assertFalse(pathExists, "Message znode should have been removed by accessor msgId=" + msgId);
+    }
   }
 
   @Test
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
@@ -434,7 +434,7 @@ public class TestZkBaseDataAccessor extends ZkUnitTestBase {
   }
 
   @Test
-  public void testDeleteLogging() {
+  public void testDeleteNodeWithChildren() {
     String root = _rootPath;
 
     ZkBaseDataAccessor<ZNRecord> accessor = new ZkBaseDataAccessor<>(_gZkClient);
--- a/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
+++ b/helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
@@ -436,7 +436,6 @@ public class TestZkBaseDataAccessor extends ZkUnitTestBase {
   @Test
   public void testDeleteLogging() {
     String root = _rootPath;
-    _gZkClient.deleteRecursively("/" + root);
 
     ZkBaseDataAccessor<ZNRecord> accessor = new ZkBaseDataAccessor<>(_gZkClient);
 
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -2073,7 +2073,7 @@ public class ZkClient implements Watcher {
       record(path, null, startT, ZkClientMonitor.AccessType.WRITE);
     } catch (Exception e) {
       recordFailure(path, ZkClientMonitor.AccessType.WRITE);
-      LOG.warn("zkclient {}, Failed to delete path {}! ", _uid, path, e);
+      LOG.info("zkclient {}, Failed to delete path {}! ", _uid, path, e);
       throw e;
     } finally {
       long endT = System.currentTimeMillis();
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -2066,14 +2066,11 @@ public class ZkClient implements Watcher {
         success = true;
       } catch (ZkNoNodeException e) {
         success = false;
-        if (LOG.isDebugEnabled()) {
-          LOG.debug("zkclient {}, Failed to delete path {}, znode does not exist!", _uid, path);
-        }
+        LOG.debug("zkclient {}, Failed to delete path {}, znode does not exist!", _uid, path);
       }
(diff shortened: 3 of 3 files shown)


qqu0127:
Hi Marcos, one quick thing, we don't post internal ticket to github, it's not accessible to outside world anyway. Please use github issue instead.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -433,6 +433,29 @@ public void testSyncRemove() {
...
+      records.add(new ZNRecord(msgId));
+    }
+    boolean[] success = accessor.createChildren(paths, records, AccessOption.PERSISTENT);
+
+    // Attempt to remove parent. Shouldn't throw an error or warning log.
+    // Should return True if recursive deletion succeeds.
+    Assert.assertTrue(accessor.remove(PropertyPathBuilder.instanceMessage(root, "host_1"), 0),
+            "Should return True despite log errors.");
+  }
+
```
xyuanlu, line 458: Thanks for making this change!! 
I would suggest validate that these ZNode has already being deleted.

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -433,6 +433,37 @@ public void testSyncRemove() {
     System.out.println("END " + testName + " at " + new Date(System.currentTimeMillis()));
   }
 
+  @Test
+  public void testDeleteLogging() {
+    String root = _rootPath;
+    _gZkClient.deleteRecursively("/" + root);
```
xyuanlu, line 439: I might have missed it last time. Why we are deleting everything under root?

xyuanlu: Also please link the issue number if the description.

Marcosrico: This is to make sure we work with a clean path. However, it might not be necessary now that you point it out.

junkaixue:
NIT: please remove original template context in the details.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -726,7 +726,7 @@ public boolean remove(String path, int options) {
       // case, we try to delete recursively
       _zkClient.delete(path);
     } catch (ZkException e) {
-      LOG.debug("Failed to delete {} with opts {}, err: {}. Try recursive delete", path, options,
+      LOG.info("Failed to delete {} with opts {}, err: {}. Try recursive delete", path, options,
```
qqu0127, line 729: Just FYI, change from debug to info is a up-level. We may see more logs like this, is this what we want? (most default log level is at INFO or plus)

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
```diff
@@ -433,6 +433,36 @@ public void testSyncRemove() {
     System.out.println("END " + testName + " at " + new Date(System.currentTimeMillis()));
   }
 
+  @Test
+  public void testDeleteLogging() {
+    String root = _rootPath;
+
+    ZkBaseDataAccessor<ZNRecord> accessor = new ZkBaseDataAccessor<>(_gZkClient);
```
qqu0127, line 440: Thanks for writing up the test. 
Logging test can be tricky, it depends on the configuration of log level in test environment, and there isn't a good way to verify the log message. 
If I understand correctly, this test is only to test logging, but not to test the delete functionality?

Marcosrico: Correct, this is just to check that the specific logging message is displayed. I believe delete functionality is tested in `TestZkBaseDataAccessor.testAsyncZkBaseDataAccessor()`

qqu0127: Thanks. In this case, I don't think we need this test at all.

Marcosrico: I might have phrased my answer wrong. The test does check whether zk catches the error of removing a parent node with children, which I believe is not tested elsewhere, though not sure if that is considered as delete functionality. The primary motive for the test is to catch this error and make sure the right logging is made. Perhaps that is not necessary given the variable nature of log configs - let me know, thanks!

qqu0127: Ok, I see your point. My concern is that, `"make sure the right logging is made"` is checked manually. It doesn't add much value to the unit test in regard to logging. Functionally, it does have value to verify no exception is thrown. So we may keep the test, and rename it to e.g. `testDeleteNodeWithChildren`, just my two cents.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
Marcosrico: Hmm you're right. What we're looking for is to diminish logging clutter - increasing the logging level would worsen that! I wanted to have the logging level consistent with the log error in the remove function but that wouldn't make sense to the overall objective of this fix. Thanks for the comment!

helix-core/src/test/java/org/apache/helix/manager/zk/TestZkBaseDataAccessor.java
Marcosrico: I agree with renaming the tests for better clarity - thanks for the feedback!

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
xyuanlu: The aim for this change is to make the log less misleading. By increasing the log level will indicate user that the previous warning log saying failed to remove znode won't be an issue and we will retry remove recursively. 
This is only in failure path and won't be causing log flooding.

xyuanlu:
For line 2069 in [ZkClient.java](https://github.com/apache/helix/pull/2201/files#diff-c20891e1adafdc5bd7c55936d81ec52e84081b76f7df3759da762f932d3f1cae), we should remove 
`if (LOG.isDebugEnabled()) {`
It is redundant.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
rahulrane50: neat : I understand the overall intention of changing this to INFO level but I would be very cautious about this change since this is very core method which is used in many places. Any small changes in logging here would flood logs in specific scenarios. Also could you help me understand why was the "debug" log noise in the first place. It shouldn't even show up in logs unless debug logging is enabled right? (I assume debug logging is disabled by default)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2073,7 +2073,7 @@ public Object call() throws Exception {
       record(path, null, startT, ZkClientMonitor.AccessType.WRITE);
     } catch (Exception e) {
       recordFailure(path, ZkClientMonitor.AccessType.WRITE);
-      LOG.warn("zkclient {}, Failed to delete path {}! ", _uid, path, e);
+      LOG.debug("zkclient {}, Failed to delete path {}! ", _uid, path, e);
```
rahulrane50, line 2076: Umm is this correct? There are 3 parameters for 2 positional brackets.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
Marcosrico: I am not too familiar with what the default logging is for the different applications of these methods. However, the ticket and issue created arose from customers (i.e. Venice) who have run into the issue of these "noisy" logs. Ideally, we would want to completely avoid calling these logs altogether (in the case that deletion succeeds further on). However, we weren't able to come up with a liable solution. Thus, if the log error is to be shown regardless, the purpose of this change as Xiaoyuan mentioned is to make it less misleading. Let me know what you think!

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
Marcosrico: I think the first two parameters are for the positional brackets. The last one since it's an exception it is thrown as a separate message. Let me know if I'm mistaken

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
jiajunwang: I'm confused here.
If set the log level to info, then orignal code will NOT print debug log, but the new change will be printing more logs. It contradicts to the description.

> A log request of level p in a logger with level q is enabled if p >= q. This rule is at the heart of log4j. It assumes that levels are ordered. For the standard levels, we have ALL < DEBUG < INFO < WARN < ERROR < FATAL < OFF.
Here, in the original code p == debug and q == info. So the log won't be printed.

Debug to Info is considered upgrade instead of downgrade.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2073,7 +2073,7 @@ public Object call() throws Exception {
       record(path, null, startT, ZkClientMonitor.AccessType.WRITE);
     } catch (Exception e) {
       recordFailure(path, ZkClientMonitor.AccessType.WRITE);
-      LOG.warn("zkclient {}, Failed to delete path {}! ", _uid, path, e);
+      LOG.info("zkclient {}, Failed to delete path {}! ", _uid, path, e);
```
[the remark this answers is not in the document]
xyuanlu: +1 on this.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
Marcosrico: You're right, I made a mistake in the description, thank you for pointing that out! I'll update the description with this feedback.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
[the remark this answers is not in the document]
Marcosrico: That is a very good point, thank you!

jiajunwang:
> For line 2069 in [ZkClient.java](https://github.com/apache/helix/pull/2201/files#diff-c20891e1adafdc5bd7c55936d81ec52e84081b76f7df3759da762f932d3f1cae), we should remove `if (LOG.isDebugEnabled()) {` It is redundant.

This is in general to avoid error message computing (if applicable), like format. Not necessarily redundent. But IMO, I think it looks ugly and does not really save much, I vote for removing them : )

xyuanlu:
LGTM. Thanks for making the change.

xyuanlu (approved):

NealSun96:
CI workflows approved.

xyuanlu:
> 

Please correct me if I am wrong. The this kind of log formatter, log4j is already doing lazily log construction. So we will only do string construction when the log level<= debug. 
https://logging.apache.org/log4j/2.0/manual/api.html#Java_8_lambda_support_for_lazy_logging

xyuanlu:
> > For line 2069 in [ZkClient.java](https://github.com/apache/helix/pull/2201/files#diff-c20891e1adafdc5bd7c55936d81ec52e84081b76f7df3759da762f932d3f1cae), we should remove `if (LOG.isDebugEnabled()) {` It is redundant.
> 
> This is in general to avoid error message computing (if applicable), like format. Not necessarily redundent. But IMO, I think it looks ugly and does not really save much, I vote for removing them : )

Please correct me if I am wrong. I think with the correct log formatter, log4j will do lazily log message construction. We should also do this kind of log improvement for Helix as well....
link:
https://logging.apache.org/log4j/2.x/manual/api.html

Marcosrico:
This PR is ready to merge, approved by @xyuanlu 
Failed tests are not stable and pass locally. See description for test results.

Commit message:

Removed Error Log to Avoid Spam Logging
Removed an unnecessary and misleading error log in ZkClient. This cleans up the logging and avoids clutter.

Issue #2207, opened 2022-09-09
Unnecessary logs before throwing exceptions


Marcosrico:
Per good coding practice, if an error exception is thrown there is no need to log. This occurs this several times looking throughout Helix. 

### Expected behavior
If an error is thrown there shouldn't, the method shouldn't log beforehand. The caller can do so if necessary.

PR #2208, opened 2022-09-09
Unnecessary log removal

Marcosrico:
Remove unnecessary logs to clean up code and logging.

### Issues

https://github.com/apache/helix/issues/2207

### Description

If a method throws an error there is no need to log. The caller can log if necessary. Removed any log calls where an exception is thrown afterwards.


helix-core/src/main/java/org/apache/helix/manager/zk/CallbackEventThreadPoolFactory.java
```diff
@@ -83,8 +83,6 @@ private static ThreadPoolExecutor getOrCreateThreadPoolHelper(int hash) {
         _managerToCallBackThreadPoolMap.put(hash, result);
         _callBackEventProcessorCountPerThreadPool.put(hash, new AtomicInteger(1));
       } catch (Exception e) {
-        logger.error(String
-            .format("Error when creating new ThreadPoolExecutor for %s.", hash), e);
```
xyuanlu, line 87: If we catch and throw, do we need to actually catch it here?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2066,14 +2066,11 @@ public Object call() throws Exception {
...
         success = false;
-        if (LOG.isDebugEnabled()) {
-          LOG.debug("zkclient {}, Failed to delete path {}, znode does not exist!", _uid, path);
-        }
+        LOG.debug("zkclient {}, Failed to delete path {}, znode does not exist!", _uid, path);
       }
       record(path, null, startT, ZkClientMonitor.AccessType.WRITE);
     } catch (Exception e) {
       recordFailure(path, ZkClientMonitor.AccessType.WRITE);
-      LOG.warn("zkclient {}, Failed to delete path {}! ", _uid, path, e);
```
qqu0127, line 2076: Rather a general comment, let's be cautious on the removal. The thrown exception may not contain local context like `_uid` and `path`, so it may make debugging harder. 
`ZkClient` is a very fundamental class, my personal take is to keep it, no strong preference though.

jiajunwang:
Thanks for the following up @Marcosrico. Let's proceed with cautious. Even log is not necessary, it might still be a good idea to inject some information to the exception for supporting debug.

I found this articial which presents a good idea,
https://www.ibm.com/support/pages/best-practice-catching-and-re-throwing-java-exceptions

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/InstancesAccessor.java
```diff
@@ -277,10 +277,7 @@ private Response batchGetStoppableInstances(String clusterId, JsonNode node, boo
           .error(String.format("Current cluster %s has issue with health checks!", clusterId), e);
       throw new HelixHealthException(e);
     } catch (Exception e) {
-      _logger.error(String.format(
-              "Failed to get parallel stoppable instances for cluster %s with a HelixException!",
-              clusterId), e);
-      throw e;
+      throw new Exception("Failed to get parallel stoppable instances for cluster " + clusterId + " with a HelixException!", e);
```
xyuanlu, line 280: probably we do not want to change the exception time as it may change caller's behavior if they have a catch for certain exception. If no variable is logged, can we just remove the catch block.

Marcosrico: Not sure I understand. Is throwing a new Exception with `e` passed in as arguments not the same time as exception `e` by itself?

xyuanlu: The caught exception is a base type, the actual object can be any derived type (for example, HelixException). The object you throw is an object of the base type. 
If there is a try-catch block in caller that catch any derived type (for example, HelixException), the base type exception won't be caught. -> we will have behavior change.

helix-core/src/main/java/org/apache/helix/manager/zk/ZkBaseDataAccessor.java
```diff
@@ -393,9 +393,6 @@ public AccessResult doSet(String path, T record, int expectVersion, int options)
           result._retCode = RetCode.ERROR;
           return result;
         }
-      } catch (ZkBadVersionException e) {
-        LOG.debug("Exception while setting path: " + path, e);
-        throw e;
```
xyuanlu, line 398: we may introduce a behavior change here. As now when we get `ZkBadVersionException`, it will go to general exception branch.

helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java
```diff
@@ -781,9 +781,6 @@ void createClient() throws Exception {
          */
         handleNewSession(ZKUtil.toHexSessionId(sessionId));
         break;
-      } catch (HelixException e) {
-        LOG.error("fail to createClient.", e);
-        throw e;
```
xyuanlu, line 786: Same here. We may have behavior change introduced here..

helix-rest/src/main/java/org/apache/helix/rest/server/resources/helix/InstancesAccessor.java
```diff
@@ -274,13 +274,8 @@ private Response batchGetStoppableInstances(String clusterId, JsonNode node, boo
       return JSONRepresentation(result);
     } catch (HelixException e) {
       _logger
-          .error(String.format("Current cluster %s has issue with health checks!", clusterId), e);
+              .error(String.format("Current cluster %s has issue with health checks!", clusterId), e);
       throw new HelixHealthException(e);
-    } catch (Exception e) {
-      _logger.error(String.format(
-              "Failed to get parallel stoppable instances for cluster %s with a HelixException!",
-              clusterId), e);
```
xyuanlu, line 282: Let's keep the log here? Logging the cluster name will be helpful when we debug.

helix-view-aggregator/src/main/java/org/apache/helix/view/statemodel/DistViewAggregatorStateModel.java
```diff
@@ -57,8 +57,6 @@ public void onBecomeLeaderFromStandby(Message message, NotificationContext conte
       _aggregator = new HelixViewAggregator(viewClusterName, _zkAddr);
       _aggregator.start();
     } catch (Exception e) {
-      logger.error("Aggregator failed to become leader from stand by for view cluster {}",
-          viewClusterName, e);
```
xyuanlu, line 61: same here.

xyuanlu (approved):
LGTM. Thanks for making the change.

qqu0127 (approved):
LGTM

PR #2214, opened 2022-09-19
Add new interface IZkChildEventListener to allow pass in WatchedEvent

qqu0127:
### Issues

### Description

Add a new interface for zk child listener and refactor to maintain backward compatibility.
The previous https://github.com/apache/helix/blob/ee61ff434ee37d4ab8456c739b2229f99d9e0e72/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/IZkChildListener.java doesn't pass in `WatchedEvent` to the method, as a result implementation can't depend on different event types.

This PR introduce a new interface with the ideal API signature. The old interface is refactored to depend on the new one, AND use a default method for backward compatibility. New use cases should be changed to use `IZkChildEventListener` while there is no breaking change for existing users.
A few other places are changed to allow passing in the event and method signature change.

### Tests

Updated `TestZkBasis` and `TestBatchMessage`

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/IZkChildListener.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/IZkChildListener.java
@@ -20,6 +20,8 @@ package org.apache.helix.zookeeper.zkclient;
  */
 
 import java.util.List;
+import org.apache.zookeeper.WatchedEvent;
+
 
 /**
  * An {@link IZkChildListener} can be registered at a {@link ZkClient} for listening on zk child changes for a given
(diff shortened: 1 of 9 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/IZkChildListener.java
```diff
@@ -29,17 +31,31 @@
  * guaranteed that events on the path are missing (see http://zookeeper.wiki.sourceforge.net/ZooKeeperWatches). An
  * implementation of this class should take that into account.
  *
+ * Deprecated: This interface is kept to maintain backward compatibility, please use {@link IZkChildEventListener}.
  */
-public interface IZkChildListener {
+@Deprecated
+public interface IZkChildListener extends IZkChildEventListener {
```
junkaixue, line 37: This is same effect as I mentioned. If you force this extends IZkChildEventListener, the impl of new API will be missing.

qqu0127: How is that? There is a default implementation of the new method. There is no breaking change for existing classes, `ZkHelixClusterVerifier` is an example

PR #2219, opened 2022-09-20
Upgrade to zookeeper 3.6.3

### LINKED ISSUE #2220: Helix needs to upgrade zookeeper dependency
The current zk version 3.5.9 is at EOL and doesn't support SSL/TLS. Helix should upgrade zookeeper dependency to a higher version.

Upgrade to 3.6.3 which is a stable version and is backward compatible with 3.5.9.

**Additional context**

mgao0:
### Issues

Resolves #2220.

### Description

This commit includes several efforts in order to bump up zookeeper version:
1. Upgrade zookeeper version to 3.6.3 in pom.xml and ivy file
2. Add snappy-java dependency to helix-core
3. Some code change in ZKLogFormatter to fix the differences in SerializeUtils class between zk version 3.5.9 and 3.6.3
4. Due to the change of zookeeper server shutdown mode ([link](https://github.com/apache/zookeeper/commit/722ba9409a44a35d287aac803813f508cff2420a)) in zookeeper version 3.6.0+, one test `testGetChildrenOnLargeNumChildren` failed after upgrade. The detailed reason is that zookeeper server shutdown failed because of a too large transaction log entry with the large number of ephemeral children nodes created in the test. This commit changes the children nodes created from ephemeral mode to persistent mode, so it won't cause large transaction log entry, and deletes those children nodes at the end of the test.
5. Change the shutdown() in ZkServer to align the shutdown new behavior with old one.

### Tests

[ERROR] Failures:
[ERROR]   TestZkConnectionLost.testLostZkConnection:176 » Helix Workflow testLostZkConne...
[ERROR]   TestP2PNoDuplicatedMessage.testP2PStateTransitionEnabled:180 expected:<true> but was:<false>
[INFO]
[ERROR] Tests run: 1318, Failures: 2, Errors: 0, Skipped: 0
[INFO]
Both are flaky tests, passed after re-run.

CI tests:
All failed tests are known flaky tests, will fix separately
Error:  Test failed: testZKHelixManager(org.apache.helix.integration.multizk.TestMultiZkConectionConfig)  Time elapsed: 900.02 s  <<< FAILURE! [https://github.com/apache/helix/issues/2192](url)
Error:  Test failed: afterClass(org.apache.helix.integration.multizk.TestMultiZkConectionConfig)  Time elapsed: 965.409 s  <<< FAILURE! [https://github.com/apache/helix/issues/2193](url)
Error:  Test failed: testZkClientCreationMultiZk(org.apache.helix.task.TestTaskStateModelFactory)  Time elapsed: 0.028 s  <<< FAILURE! [https://github.com/apache/helix/issues/2194](url)

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/helix-core-1.0.5-SNAPSHOT.ivy
+++ b/helix-core/helix-core-1.0.5-SNAPSHOT.ivy
@@ -52,7 +52,7 @@ under the License.
     <dependency org="org.apache.logging.log4j" name="log4j-slf4j-impl" rev="2.17.1" force="true" conf="compile->compile(*),master(*);runtime->runtime(*)">
         <artifact name="log4j-slf4j-impl" ext="jar"/>
     </dependency>
-    <dependency org="org.apache.zookeeper" name="zookeeper" rev="3.5.9" conf="compile->compile(default);runtime->runtime(default);default->default"/>
+    <dependency org="org.apache.zookeeper" name="zookeeper" rev="3.6.3" conf="compile->compile(default);runtime->runtime(default);default->default"/>
 		<dependency org="com.fasterxml.jackson.core" name="jackson-databind" rev="2.12.6.1" conf="compile->compile(default);runtime->runtime(default);default->default"/>
     <dependency org="com.fasterxml.jackson.core" name="jackson-core" rev="2.12.6" conf="compile->compile(default);runtime->runtime(default);default->default"/>
     <dependency org="commons-io" name="commons-io" rev="2.11.0" conf="compile->compile(default);runtime->runtime(default);default->default"/>
@@ -63,6 +63,7 @@ under the License.
     <dependency org="org.yaml" name="snakeyaml" rev="1.30" conf="compile->compile(default);runtime->runtime(default);default->default"/>
     <dependency org="commons-logging" name="commons-logging-api" rev="1.1" conf="compile->compile(*),master(*);runtime->runtime(*)"/>
     <dependency org="io.dropwizard.metrics" name="metrics-core" rev="4.1.14" conf="compile->compile(default);runtime->runtime(default);default->default"/>
+    <dependency org="org.xerial.snappy" name="snappy-java" rev="1.1.7" conf="compile->compile(default);runtime->runtime(default);default->default"/>
 	</dependencies>
 </ivy-module>
 
--- a/zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
+++ b/zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
@@ -1063,13 +1063,15 @@ public class TestRawZkClient extends ZkTestBase {
    * Tests getChildren() when there are an excessive number of children and connection loss happens,
    * the operation should terminate and exit retry loop.
    */
-  @Test(timeOut = 30 * 1000L)
+  @Test
   public void testGetChildrenOnLargeNumChildren() throws Exception {
     final String methodName = TestHelper.getTestMethodName();
     System.out.println("Start test: " + methodName);
+    String[] nodePaths = new String[110001];
     // Create 110K children to make packet length of children exceed 4 MB
     // and cause connection loss for getChildren() operation
     String path = "/" + methodName;
+    nodePaths[110000] = path;
 
     _zkClient.createPersistent(path);
 
@@ -1077,9 +1079,9 @@ public class TestRawZkClient extends ZkTestBase {
       List<Op> ops = new ArrayList<>(1000);
       for (int j = 0; j < 1000; j++) {
         String childPath = path + "/" + UUID.randomUUID().toString();
-        // Create ephemeral nodes so closing zkClient deletes them for cleanup
+        nodePaths[1000 * i + j] = childPath;
         ops.add(
-            Op.create(childPath, new byte[0], ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL));
+            Op.create(childPath, new byte[0], ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT));
       }
       // Reduce total creation time by batch creating znodes
       _zkClient.multi(ops);
@@ -1098,11 +1100,14 @@ public class TestRawZkClient extends ZkTestBase {
       _zkClient = new ZkClient(ZkTestBase.ZK_ADDR);
 
       Assert.assertTrue(TestHelper.verify(() -> {
-        try {
-          return _zkClient.delete(path);
-        } catch (ZkException e) {
-          return false;
+        for (String toDelete: nodePaths) {
+          try {
+            _zkClient.delete(toDelete);
+          } catch (ZkException e) {
+            return false;
+          }
         }
+        return true;
       }, TestHelper.WAIT_DURATION));
     }
     System.out.println("End test: " + methodName);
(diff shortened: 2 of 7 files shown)


zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -1063,23 +1063,26 @@ public void testAsyncWriteByExpectedSession() throws Exception {
...
       for (int j = 0; j < 1000; j++) {
         String childPath = path + "/" + UUID.randomUUID().toString();
+        nodePaths[1000 * i + j] = childPath;
         // Create ephemeral nodes so closing zkClient deletes them for cleanup
         ops.add(
-            Op.create(childPath, new byte[0], ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL));
+            Op.create(childPath, new byte[0], ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT));
       }
       // Reduce total creation time by batch creating znodes
       _zkClient.multi(ops);
```
Marcosrico, line 1088: Forgive me for I might not have a good understanding on this but what is `ops` and what is the purpose of calling `multi` on it? I believe `multi()` isn't supported at the moment, just wondering if I am missing something. Thanks!

mgao0: Thanks for your comment!
`ops` is an iterable of `op` (which is a single operation https://zookeeper.apache.org/doc/r3.5.5/api/org/apache/zookeeper/Op.html). Transaction support is a zk server side feature so we already have it.
` multi() isn't supported at the moment` this means we don't have the multi() API in Helix ZKClient.
Please let me know if it makes sense.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -1063,23 +1063,25 @@ public void testAsyncWriteByExpectedSession() throws Exception {
    * Tests getChildren() when there are an excessive number of children and connection loss happens,
    * the operation should terminate and exit retry loop.
    */
-  @Test(timeOut = 30 * 1000L)
+  @Test
   public void testGetChildrenOnLargeNumChildren() throws Exception {
     final String methodName = TestHelper.getTestMethodName();
     System.out.println("Start test: " + methodName);
+    String[] nodePaths = new String[110001];
```
qqu0127, line 1070: I assume it's 110 * 1000 + 1 ? 
nit: Let's use constants and multiplication for these numbers. Same as below in the for-loop. Thanks!

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
```diff
@@ -1063,23 +1063,29 @@ public void testAsyncWriteByExpectedSession() throws Exception {
...
-      for (int j = 0; j < 1000; j++) {
+    for (int i = 0; i < numOps; i++) {
+      List<Op> ops = new ArrayList<>(numOpInOps);
+      for (int j = 0; j < numOpInOps; j++) {
         String childPath = path + "/" + UUID.randomUUID().toString();
-        // Create ephemeral nodes so closing zkClient deletes them for cleanup
+        nodePaths[numOpInOps * i + j] = childPath;
         ops.add(
-            Op.create(childPath, new byte[0], ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL));
+            Op.create(childPath, new byte[0], ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT));
```
rahulrane50, line 1088: I understand for this test creating persistent node is also fine but let's discuss if this a problem if someone is trying to create large number of ephemeral nodes? Couple of questions : 
1. Why was it working previously? Was it because jutemax buffer was 4M and now it's 1M? because if that's the case then we should ideally keep jutemaxbuffer for this test same as what we kept in prod and then reason out failure/test scenario.
2. I'm still not understanding why would a large transaction log has anything to do with data consistency. (Issue you mentioned in description). Can you please explain?

helix-core/helix-core-1.0.5-SNAPSHOT.ivy
```diff
@@ -63,6 +63,7 @@ under the License.
     <dependency org="org.yaml" name="snakeyaml" rev="1.30" conf="compile->compile(default);runtime->runtime(default);default->default"/>
     <dependency org="commons-logging" name="commons-logging-api" rev="1.1" conf="compile->compile(*),master(*);runtime->runtime(*)"/>
     <dependency org="io.dropwizard.metrics" name="metrics-core" rev="4.1.14" conf="compile->compile(default);runtime->runtime(default);default->default"/>
+    <dependency org="org.xerial.snappy" name="snappy-java" rev="1.1.7" conf="compile->compile(default);runtime->runtime(default);default->default"/>
```
rahulrane50, line 66: just curious, do we need this in snapshot as well? I thought just adding this in pom.xml is good enough?
Adding it to snapshot would affect backward compatibility or not?

rahulrane50 (commented):
Good find @mgao0 and thanks for fixing this!

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
mgao0: Thanks for the questions. Bottom line is that we should limit the number of ephemeral nodes that someone can create in one session. You were working with ephemeral nodes throttling before so you might also have some context.
1. It worked before because the before bump up, it doesn't need to read transaction log. Please see my 2nd point. It is not necessarily about jute max buffer. But if jute max buffer is large enough (>9MB) this test will pass without changing the test because it won't fail when reading transaction log.
2. I updated the description for a more specific commit. You can see that in 3.6.0 it added a zookeeper server shutdown mode which is non-fullyshutdown. And it will require a fastForwardDataBase in order to do this kind of shutdown. The error occurred when reading the transaction log during fastForwardDataBase.

helix-core/helix-core-1.0.5-SNAPSHOT.ivy
mgao0: My understanding is to keep pom file and ivy file in sync, pom is for maven, and ivy is for gradle. But they should be same. I'll look more into this.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
rahulrane50: Ah i got your point now, thanks a lot for explaining!

rahulrane50 (approved):
LGTM!

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestRawZkClient.java
Marcosrico: Got it, makes sense. Thanks for the clarification!

qqu0127 (approved):
LGTM, thank you for the fix!

mgao0:
This PR is ready to be merged, approved by [rahulrane50](https://github.com/rahulrane50) and [qqu0127](https://github.com/qqu0127). Thanks!

Commit message:
Bump up zookeeper dependency version to 3.6.3

abhilash1in (approved):
LGTM

PR #2223, opened 2022-09-26
Add default method for TTL support

qqu0127:
Implement default methods for two TTL related methods in BaseDataAccessor to maintain compatibility. 
Also fix the snappy java dependency issue by removing the "test" scope.

### DIFF
--- a/helix-core/pom.xml
+++ b/helix-core/pom.xml
@@ -150,7 +150,6 @@
       <groupId>org.xerial.snappy</groupId>
       <artifactId>snappy-java</artifactId>
       <version>1.1.7</version>
-      <scope>test</scope>
     </dependency>
   </dependencies>
   <build>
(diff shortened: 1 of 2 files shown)


rahulrane50:
LGTM

desaikomal (approved):
Thanks @qqu0127  for quick turn-around.
Looks good to me. 
May be in future, we should add a test-class which extends BaseAccessor and just to make sure we don't cause regression with BaseAccessor.

qqu0127:
> Thanks @qqu0127 for quick turn-around. Looks good to me. May be in future, we should add a test-class which extends BaseAccessor and just to make sure we don't cause regression with BaseAccessor.

Thanks @desaikomal , this is true. But I'm afraid it's still more on developer's own awareness. Normally people will fix all compile error caused by the interface change. That's end of story for the repo itself, but not for external user.

qqu0127:
This is ready to merge, approved by @desaikomal and @rahulrane50 

Commit message: 
Implement default methods for two TTL related methods in BaseDataAccessor to maintain compatibility.

PR #2238, opened 2022-10-17
2235/identity token

### LINKED ISSUE #2235: helix front: support identity token at login
In helix-front: support a configurable identity token at login.

micahstubbs:
### Description

This PR adds support for a configurable identity token source to the helix-front ExpressJS server. 

Fix #2235

### Tests

New unit or integration tests: No new tests. 

<details>

### Code Style

Formatted using [Prettier](https://prettier.io)

### Commits

- My commits all reference appropriate Apache Helix GitHub issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"

- [ ] My PR contains changes that break backward compatibility or previous assumptions for certain methods or API. They include: -->

### Documentation (Optional)

- [ ] In case of new functionality, my PR adds documentation in the following wiki page: -->

</details>

### DIFF
--- a/helix-front/server/config.example.ts
+++ b/helix-front/server/config.example.ts
@@ -5,9 +5,7 @@ export const HELIX_ENDPOINTS = {
     },
   ],
 };
-
 export const SESSION_STORE = undefined;
-
 export const SSL = {
   port: 0,
   keyfile: '',
@@ -15,10 +13,13 @@ export const SSL = {
   passfile: '',
   cafiles: [],
 };
-
 export const LDAP = {
   uri: 'ldap://example.com',
   base: 'DC=example,DC=com',
   principalSuffix: '@example.com',
   adminGroup: 'admin',
 };
+
+export const IDENTITY_TOKEN_SOURCE = undefined; // 'www.example.com';
+export const CUSTOM_IDENTITY_TOKEN_REQUEST_BODY = {};
+export const TOKEN_RESPONSE_KEY = 'token';
--- a/helix-front/server/config.example.ts
+++ b/helix-front/server/config.example.ts
@@ -20,6 +20,25 @@ export const LDAP = {
   adminGroup: 'admin',
 };
 
-export const IDENTITY_TOKEN_SOURCE = undefined; // 'www.example.com';
-export const CUSTOM_IDENTITY_TOKEN_REQUEST_BODY = {};
-export const TOKEN_RESPONSE_KEY = 'token';
+/**
+ * The url of your Identity Token API.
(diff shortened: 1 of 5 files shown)


helix-front/server/config.example.ts
```diff
@@ -5,20 +5,21 @@ export const HELIX_ENDPOINTS = {
...
-
 export const LDAP = {
   uri: 'ldap://example.com',
   base: 'DC=example,DC=com',
   principalSuffix: '@example.com',
   adminGroup: 'admin',
 };
+
+export const IDENTITY_TOKEN_SOURCE = undefined; // 'www.example.com';
+export const CUSTOM_IDENTITY_TOKEN_REQUEST_BODY = {};
```
cliffordfajardo, line 24: Are there specific attributes that are required to be present for  `CUSTOM_IDENTITY_TOKEN_REQUEST_BODY`


Does it make sense to add a `type` or `interface for `CUSTOM_IDENTITY_TOKEN_REQUEST_BODY`
Example
```
export const CUSTOM_IDENTITY_TOKEN_REQUEST_BODY: {some_inline_type} = {};
```

cliffordfajardo (commented):
LGTM
Just verify it works end to end

helix-front/server/config.example.ts
micahstubbs: Good question.  I don't think so, since by definition it is custom, so it totally depends on the Identity Source / Auth Provider that you decide to use helix-front with.  It would probably make sense to explicitly type this as `any` to communicate that better. Thanks for the suggestion!

helix-front/server/controllers/d.ts
```diff
@@ -5,6 +5,7 @@ export interface HelixUserRequest extends Request {
 }
 
 interface HelixSession {
+  identityToken: any;
```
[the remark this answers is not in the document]
micahstubbs: I would generally agree, however since this token format is user-customizable, this really is `any` from helix-front's point of view.  helix-front doesn't know what Identity Source it will be used with, or what type that Identity Source's identityToken will be.

helix-front/server/config.example.ts
micahstubbs: Added comments to clarify this.

micahstubbs:
This PR is ready to be merged, approved by @somecodemonkey @cliffordfajardo       
Final commit message:
## Add support for optional identity token source in helix-front (#2235)
During helix-front login, optionally request an identity token
from a configurable identity token API.

PR #2239, opened 2022-10-18
[WIP] Rearranging Tests to Accommodate ZkMulti

Marcosrico:
To test implementation of FederatedZkClient Multi, tests must be written. Mainly rearranged current test base for cleaner code and added specific test for FederatedZkClient multi support.

### Issues

N.A.

### Tests

Solely rearranged tests under `/multizk` and combined similar code to clean up the tests. Wrote separate test to test future implementation of FederatedZkClient ZkMulti in `testMultiInMultiZk`. Currently FederatedZkClient ZkMulti is not supported. 

Current test failure is known issue: https://github.com/apache/helix/issues/2286

### Code Quality


zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -479,8 +479,18 @@ public long getCreationTime(String path) {
 
   @Override
   public List<OpResult> multi(Iterable<Op> ops) {
-    throwUnsupportedOperationException();
-    return null;
+    if (ops == null) {
+      throw new NullPointerException("ops must not be null.");
+    }
+    String op_path = ops.iterator().next().getPath();
```
xyuanlu, line 485: I think we can call `getZkRealm(op_path)` here and store the realm to same some map lookup.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -479,8 +479,18 @@ public long getCreationTime(String path) {
```
xyuanlu, line 479: Please add API description here and explain where are supported in federatedZK client and in witch cases we throw exception.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -479,8 +479,18 @@ public long getCreationTime(String path) {
...
-    throwUnsupportedOperationException();
-    return null;
+    if (ops == null) {
+      throw new NullPointerException("ops must not be null.");
+    }
+    String op_path = ops.iterator().next().getPath();
+    // Check whether all ops are connected to the same ZK server. If any differ, multi can't be run.
+    for (Op op : ops) {
+      if (!getZkRealm(op_path).equals(getZkRealm(op.getPath()))){
+        throwUnsupportedOperationException();
```
qqu0127, line 489: We should better throw IllegalArgumentException

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/RealmAwareZkClientFactoryTestBase.java
```diff
@@ -44,6 +51,7 @@ public abstract class RealmAwareZkClientFactoryTestBase extends RealmAwareZkClie
   protected RealmAwareZkClientFactory _realmAwareZkClientFactory;
   protected RealmAwareZkClient _realmAwareZkClient;
   private static final ZNRecord DUMMY_RECORD = new ZNRecord("DummyRecord");
+  protected String PARENT_PATH;
```
qqu0127, line 54: Can this be a static final constant? Let's try to put it in its subclasses, not base class.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
qqu0127, line 485: Two things:
Let's unify the variable naming with camelcase. 
Can we embed the zk path check in the for-loop? There is a small overhead of instantiating this extra iterator.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
```diff
@@ -662,4 +654,124 @@ public void testClose() {
       Assert.assertEquals(ex.getMessage(), "FederatedZkClient is closed!");
     }
   }
+
+  /**
+   * Test that zk multi works for op.create.
+   */
```
xyuanlu, line 660: I think since this test class already extends RealmAwareZkClientTestBase, the following tests should already be included. 
We could add test cases that are specific to federatedZKClient implementation of multi. For example, try to create multiple ZNode that do not belongs to one realm and expect a failure.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestFederatedZkClient.java
```diff
@@ -662,4 +654,124 @@ public void testClose() {
...
+
+    //Execute transactional support on operations and verify they were run
+    try {
+      List<OpResult> opResults = _realmAwareZkClient.multi(ops);
+      Assert.assertTrue(opResults.get(0) instanceof OpResult.CreateResult);
+      Assert.assertTrue(opResults.get(1) instanceof OpResult.CreateResult);
+      cleanup();
+    } catch (UnsupportedOperationException e) {
+      Assert.fail("Zk Multi Not Supported!");
+    }
```
qqu0127, line 681: Is this catch necessary?

helix-core/src/test/java/org/apache/helix/integration/multizk/MultiZkTestBase.java
```diff
@@ -0,0 +1,150 @@
+package org.apache.helix.integration.multizk;
```
Marcosrico, line 1: Copied Before and After class from pre-existing tests in MultiZkConnectionConfig to avoid code repetition for the other test cases.

helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiInMultiZk.java
```diff
@@ -0,0 +1,74 @@
...
+                System.out.println(elm);
+            }
+        }
+        System.out.println("end start");
+    }
+
+    @AfterClass
+    public void afterClass() throws Exception {
+        super.afterClass();
+    }
```
qqu0127, line 47: is this necessary?

helix-core/src/test/java/org/apache/helix/integration/multizk/MultiZkTestBase.java
qqu0127: please add apache license, thanks

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -477,10 +477,31 @@ public long getCreationTime(String path) {
...
+      throw new NullPointerException("ops must not be null.");
+    }
+    String opPath = null;
+    String opPathRealm = null;
+    for (Op op : ops) {
+      if (opPath == null) {
+        opPath = op.getPath();
+        opPathRealm = getZkRealm(op.getPath());
+      } else {
+        if (!opPathRealm.equals(getZkRealm(op.getPath()))){
```
qqu0127, line 499: else if can be combined

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -477,10 +477,31 @@ public long getCreationTime(String path) {
...
+          throw new IllegalArgumentException();
```
qqu0127, line 500: Some exception message will be great, to indicate not all operations are using the same realm.

Issue #2237, opened 2022-10-17
 Pluggable metadata storage for Helix - MetaClient library


xyuanlu:
**Is your feature request related to a problem? Please describe.**
Helix is a large and powerful complex system containing many sub-modules. Currently, Helix logic in all these submodules is heavily dependent on Zookeeper. 
With the growth of number of user data, existing ZooKeeper cluster may facing scalability issue. It is hard to migrate any Helix service to another metadata service with the current convoluted ZkClient structure.

**Describe the solution you'd like**
We will provide a MetaClient library. It contains a generic metadata access interfaces as well as widely used functionality oriented classes build on top of the it. The generic metadata access client interfaces can be implemented into multiple flavors that take advantage of different underlying metadata stores.

PR #2249, opened 2022-10-20
Add listener interfaces and config class

xyuanlu:
### Issues

#2237

### Description

This change adds factory and listener classes 

### Tests

NA

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/AsyncCallback.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/AsyncCallback.java
@@ -22,18 +22,16 @@ package org.apache.helix.metaclient.api;
 import java.util.List;
 
 /**
- * An asynchronous callback is deferred to invoke after an async CRUD operation finishes and returns.
+ * An asynchronous callback is deferred to invoke after an async CRUD operation finish and return.
  * The corresponding callback is registered when async CRUD API is invoked. Implementation processes
  * the result of each CRUD call. It should check return code and perform accordingly.
- *
- *
  */
 // TODO: define return code. failure code should map to MetaClient exceptions.
 public interface AsyncCallback {
   //This callback is used when stat object is returned from the operation.
   interface StatCallback extends AsyncCallback {
     /**
-     *
+     * Process the result of asynchronous calls that returns a stat object.
      * @param returnCode  The return code of the call.
      * @param key the key that passed to asynchronous calls.
      * @param context context object that passed to asynchronous calls.
@@ -45,7 +43,7 @@ public interface AsyncCallback {
   //This callback is used when data is returned from the operation.
   interface DataCallback extends AsyncCallback {
     /**
-     *
+     * Process the result of asynchronous calls that returns entry data.
      * @param returnCode  The return code of the call.
      * @param key The key that passed to asynchronous calls.
      * @param context context object that passed to asynchronous calls.
@@ -58,7 +56,7 @@ public interface AsyncCallback {
   //This callback is used when nothing is returned from the operation.
   interface VoidCallback extends AsyncCallback {
     /**
-     *
+     * Process the result of asynchronous calls that has no return value.
      * @param returnCode  The return code of the call.
      * @param key he key that passed to asynchronous calls.
      * @param context context object that passed to asynchronous calls.
@@ -69,7 +67,7 @@ public interface AsyncCallback {
   //This callback is used to process the list if OpResults from a single transactional call.
   interface TransactionCallback extends AsyncCallback {
     /**
-     *
+     * Process the result of asynchronous transactional calls.
      * @param returnCode  The return code of the transaction call.
      * @param keys List of keys passed to the async transactional call.
      * @param context context object that passed to asynchronous calls.
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/AsyncCallback.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/AsyncCallback.java
@@ -22,28 +22,60 @@ package org.apache.helix.metaclient.api;
 import java.util.List;
 
 /**
- * An asynchronous callback is deferred to invoke after an async CRUD operation returns.
- * The corresponding callback is registered when async CRUD API is invoked.
+ * An asynchronous callback is deferred to invoke after an async CRUD operation finishes and returns.
+ * The corresponding callback is registered when async CRUD API is invoked. Implementation processes
+ * the result of each CRUD call. It should check return code and perform accordingly.
+ *
+ *
  */
+// TODO: define return code. failure code should map to MetaClient exceptions.
 public interface AsyncCallback {
   //This callback is used when stat object is returned from the operation.
   interface StatCallback extends AsyncCallback {
+    /**
+     *
+     * @param returnCode  The return code of the call.
+     * @param key the key that passed to asynchronous calls.
+     * @param context context object that passed to asynchronous calls.
+     * @param stat the stats of the entry of the given key, returned from the async call.
+     */
     void processResult(int returnCode, String key, Object context, MetaClientInterface.Stat stat);
   }
 
   //This callback is used when data is returned from the operation.
   interface DataCallback extends AsyncCallback {
+    /**
+     *
+     * @param returnCode  The return code of the call.
+     * @param key The key that passed to asynchronous calls.
+     * @param context context object that passed to asynchronous calls.
+     * @param data returned entry data from the call.
+     * @param stat the stats of the entry of the given key.
+     */
     void processResult(int returnCode, String key, Object context, byte[] data, MetaClientInterface.Stat stat);
   }
 
   //This callback is used when nothing is returned from the operation.
   interface VoidCallback extends AsyncCallback {
+    /**
+     *
+     * @param returnCode  The return code of the call.
+     * @param key he key that passed to asynchronous calls.
+     * @param context context object that passed to asynchronous calls.
+     */
     void processResult(int returnCode, String key, Object context);
   }
 
   //This callback is used to process the list if OpResults from a single transactional call.
   interface TransactionCallback extends AsyncCallback {
-    void processResult(int returnCode, String key, Object context, List<OpResult> opResults);
+    /**
+     *
+     * @param returnCode  The return code of the transaction call.
+     * @param keys List of keys passed to the async transactional call.
+     * @param context context object that passed to asynchronous calls.
+     * @param opResults The list of transactional results.
+     */
+    void processResult(int returnCode, List<String> keys, Object context, List<OpResult> opResults);
   }
 
 }
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/ChildChangeListener.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/ChildChangeListener.java
@@ -19,15 +19,21 @@ package org.apache.helix.metaclient.api;
  * under the License.
  */
 
-import java.util.List;
-
-
-public class DirectSubEntrySubscribeResult {
-  private final List<String> _subEntries;
-  private final boolean _isInstalled;
-
-  public DirectSubEntrySubscribeResult(List<String> subEntries, boolean isInstalled) {
-    _subEntries = subEntries;
-    _isInstalled = isInstalled;
+/**
+ * Listener interface for children change events on a particular key. It includes new child
+ * creation, child deletion, child data change.
+ * TODO: add type for persist listener is removed
+ * For hierarchy key spaces like zookeeper, it refers to an entry's entire subtree.
+ * For flat key spaces, it refers to keys that matches `prefix*`.
+ */
+public interface ChildChangeListener {
+  enum ChangeType {
+    ENTRY_CREATED,     // Any child entry created
+    ENTRY_DELETED,     // Any child entry deleted
+    ENTRY_DATA_CHANGE  // Any child entry has value change
   }
+  /**
+   * Called when any child of the current key has changed.
+   */
+  void handleChildChange(String changedPath, ChangeType changeType) throws Exception;
 }
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
@@ -20,12 +20,8 @@ package org.apache.helix.metaclient.api;
  */
 
 public interface ConnectStateChangeListener {
-
-
-
-
   /**
-   * Called when the connection state has changed. Implementation could hook
+   * Called when the connection state has changed. I
    * @param prevState previous state before state change event.
    * @param currentState client state after state change event. If it is a one time listsner, it is
    *                     possible that the metaclient state changes again
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
@@ -20,10 +20,29 @@ package org.apache.helix.metaclient.api;
  */
 
 public interface ConnectStateChangeListener {
+
+  enum ConnectState {
+    // Client is connected to server
+    CONNECTED,
+
+    // Authentication failed.
+    AUTH_FAILED,
+
+    // Server has expired this connection.
+    EXPIRED,
+
+    // When client failed to connect server.
+    INIT_FAILED,
+
+    // When client explicitly call disconnect
+    CLOSED_BY_CLIENT
+  }
+
+
   /**
    * Called when the connection state has changed.
    */
-  void handleConnectStateChanged(MetaClientInterface.ConnectState currentState) throws Exception;
+  void handleConnectStateChanged(ConnectState currentState) throws Exception;
   /**
    * Called when new connection is established
    */
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
@@ -21,31 +21,20 @@ package org.apache.helix.metaclient.api;
 
 public interface ConnectStateChangeListener {
 
-  enum ConnectState {
-    // Client is connected to server
-    CONNECTED,
 
-    // Authentication failed.
-    AUTH_FAILED,
-
-    // Server has expired this connection.
-    EXPIRED,
-
-    // When client failed to connect server.
-    INIT_FAILED,
-
-    // When client explicitly call disconnect.
-    CLOSED_BY_CLIENT
-  }
 
 
   /**
-   * Called when the connection state has changed.
+   * Called when the connection state has changed. Implementation could hook
+   * @param prevState previous state before state change event.
+   * @param currentState client state after state change event. If it is a one time listsner, it is
+   *                     possible that the metaclient state changes again
    */
-  void handleConnectStateChanged(ConnectState currentState) throws Exception;
+  void handleConnectStateChanged(MetaClientInterface.ConnectState prevState, MetaClientInterface.ConnectState currentState) throws Exception;
 
   /**
-   * Called when new connection failed to established
+   * Called when new connection failed to established.
+   * @param error error returned from metaclient or metadata service.
    */
   void handleConnectionEstablishmentError(final Throwable error) throws Exception;
 
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
@@ -34,7 +34,7 @@ public interface ConnectStateChangeListener {
     // When client failed to connect server.
     INIT_FAILED,
 
-    // When client explicitly call disconnect
+    // When client explicitly call disconnect.
     CLOSED_BY_CLIENT
   }
 
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
@@ -1,0 +1,36 @@
+package org.apache.helix.metaclient.api;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+public interface ConnectStateChangeListener {
+  /**
+   * Called when the connection state has changed.
+   */
+  void handleConnectStateChanged(MetaClientInterface.ConnectState currentState) throws Exception;
+  /**
+   * Called when new connection is established
+   */
+  void handleNewConnection(final String sessionId) throws Exception;
+  /**
+   * Called when new connection failed to established
+   */
+  void handleConnectionEstablishmentError(final Throwable error) throws Exception;
+
+}
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
@@ -43,13 +43,12 @@ public interface ConnectStateChangeListener {
    * Called when the connection state has changed.
    */
   void handleConnectStateChanged(ConnectState currentState) throws Exception;
-  /**
-   * Called when new connection is established
-   */
-  void handleNewConnection(final String sessionId) throws Exception;
+
   /**
    * Called when new connection failed to established
    */
   void handleConnectionEstablishmentError(final Throwable error) throws Exception;
+
+  // TODO: Consider add a callback for new connection when we add support for session ID.
 
 }
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/DataChangeListener.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/DataChangeListener.java
@@ -24,9 +24,9 @@ package org.apache.helix.metaclient.api;
  */
 public interface DataChangeListener {
   enum ChangeType {
-    NODE_CREATED,     // Node created of the specific path that the listener register to
-    NODE_DELETED,     // Node deleted of the specific path that the listener register to
-    VALUE_UPDATE      // Node value updated
+    ENTRY_CREATED,     // Node created of the specific path that the listener register to
+    ENTRY_DELETED,     // Node deleted of the specific path that the listener register to
+    ENTRY_UPDATE      // Node value updated
   }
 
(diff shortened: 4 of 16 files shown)


meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -0,0 +1,91 @@
...
+  // private boolean _resetWatchWhenReConnect; // re-register previous existing watcher when reconnect
+  // private RetryProtocol _retryProtocol;
+
+  private MetaClientConfig(Builder builder) {
+    _connectionAddress = builder._connectionAddress;
+    _connectionTimeout = builder._connectionTimeout;
+    _enableAuth = builder._enableAuth;
+    _storeType = builder._storeType;
+
+  }
```
qqu0127, line 44: Let's input every single variables, but not depends on builder

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -0,0 +1,91 @@
...
+    private long _connectionTimeout;
+    private boolean _enableAuth;
+    //private RetryProtocol _retryProtocol;
+    private StoreType _storeType;
+
+
+
+    public MetaClientConfig build() {
+      validate();
+      return new MetaClientConfig(this);
```
qqu0127, line 58: Exposing "this" isn't a good practice, let's try to avoid

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -0,0 +1,91 @@
+package org.apache.helix.metaclient.factories;
+
+class MetaClientConfig {
+
+  public enum RetryProtocol {
+    NO_RETRY, EXP_BACK_OFF, CONST_RETRY_INTERVAL
+  }
```
qqu0127, line 7: RetryProtocol should better be a standalone class, we may want to add their custom logic and make it self-contained.

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public enum StoreType {
+    ZOOKEEPER, ETCD
+  }
+
+  private String _connectionAddress;
+  private long _connectionTimeout;
+  private boolean _enableAuth;
+  private StoreType _storeType;
```
qqu0127, line 16: with builder pattern, all these can be defined as final

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
xyuanlu: TFTR. Wouldn't input every single variables be the opposite of builder pattern?

qqu0127: For builder pattern, my own understanding is that it separates the construction of a complex object from its representation so that the same builder can create different representations with little twist. 
The constructor is still private, and people are only supposed to call builder.build(). The difference is the constructor will not depend on the builder class. Just my two cents.

desaikomal: I was reading online for builder pattern and what Qunicy says seems to be the norm.

qqu0127: Put it another way, the builder class needs to know how to build the config class, that dependency is required, but the config class doesn't need to depend on the builder. In the new way, the dependency is one-way, not circular.

xyuanlu: Updated.

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
xyuanlu: I removed the `RetryProtocol` enum. We could add back when we introduce retry config.

meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
```diff
@@ -0,0 +1,36 @@
...
+   */
+  void handleConnectStateChanged(MetaClientInterface.ConnectState currentState) throws Exception;
+  /**
+   * Called when new connection is established
+   */
+  void handleNewConnection(final String sessionId) throws Exception;
+  /**
+   * Called when new connection failed to established
+   */
+  void handleConnectionEstablishmentError(final Throwable error) throws Exception;
```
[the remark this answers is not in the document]
xyuanlu: ZkMetaClient will use a realAwareZkClient. The ZK stateChangeListsner throws a generic exception. If we catch and throw another kind of exception in metaClient, then it may break existing user behavior.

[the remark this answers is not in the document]
qqu0127: Ok, that makes sense. Thanks.

qqu0127 (approved):
LGTM

xyuanlu:
This PR is ready to be merged. Approved by @qqu0127 
Final commit message:
Add listener interfaces and config class for MetaClientInterface

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -0,0 +1,115 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+class MetaClientConfig {
+
+  public enum StoreType {
+    ZOOKEEPER, ETCD
```
junkaixue, line 25: Better need a customized mode?

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectChildrenChangeListener.java
```diff
@@ -0,0 +1,29 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Listener interface for direct children change event on a particular key. It includes new subentry creation or subentry deletion. The callback won't differentiate these types.
+ * For hierarchy key spaces like zookeeper, it refers to an entry's direct children nodes.
+ * For flat key spaces, it refers to keys that matches `prefix*separator`.
+ */
+public interface DirectChildrenChangeListener {
```
junkaixue, line 27: Why we need this? Children change listener can be recursively subscribe if it is hierarchy model

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
```diff
@@ -20,27 +20,36 @@
...
   enum EntryMode {
     //The node will be removed automatically when the session associated with the creation
     // of the node expires.
-    EPHEMERAL,
-    //The node will not be automatically deleted upon client's disconnect.
+    EPHEMERAL, //The node will not be automatically deleted upon client's disconnect.
     PERSISTENT
   }
 
+  enum ConnectState {
```
junkaixue, line 35: Have we considered session concept before? If not , how we simulate session?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -0,0 +1,231 @@
...
+  @Override
+  public boolean subscribeDataChange(String key, DataChangeListener listener,
+      DataChangeListener.ChangeType eventType, boolean skipWatchingNonExistNode,
+      boolean persistListener) {
+    return false;
+  }
+
+  @Override
+  public DirectEntrySubscribeResult subscribeDirectEntryChange(String key,
+      DirectChildrenChangeListener listener, boolean skipWatchingNonExistNode,
```
NealSun96, line 156: Why is the naming here not consistent? Shouldn't it all be `DirectChildrenChange`?

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  private final String _connectionAddress;
+  private final long _connectionTimeout;
+  private final boolean _enableAuth;
+  private final StoreType _storeType;
```
NealSun96, line 31: ZkClients has many more variables than this (operation timeout, serializer). How do they fit in this picture?

meta-client/src/main/java/org/apache/helix/metaclient/api/PersistSubEntryChangeListener.java
```diff
@@ -0,0 +1,40 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Listener interface for children change events on a particular key. It includes new subentry
+ * creation, subentry deletion, subentry data change and when the listener is removed.
+ * This listener type can only be registered as a persist listener.
+ * For hierarchy key spaces like zookeeper, it refers to an entry's entire subtree.
```
NealSun96, line 26: Question: the current `subscribeChildChanges` is only direct children if my understanding is right. Does that mean this listener will not be implemented as a part of ZkMetaClient?

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
```diff
@@ -20,27 +20,36 @@
...
 
 
 public interface MetaClientInterface<T> {
 
   enum EntryMode {
     //The node will be removed automatically when the session associated with the creation
     // of the node expires.
-    EPHEMERAL,
-    //The node will not be automatically deleted upon client's disconnect.
+    EPHEMERAL, //The node will not be automatically deleted upon client's disconnect.
```
NealSun96, line 31: nit: is this intended?

NealSun96 (changes requested):

meta-client/src/main/java/org/apache/helix/metaclient/api/PersistSubEntryChangeListener.java
xyuanlu: TFTR. Current Zookeeper support recursive watcher. ZkMetaClient uses ZkClient underlying and we may add recursive children listener if needed.

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
xyuanlu: Right. Those will be added in following commits when adding ZkMetaClient config. 
Some variables will be added here if it is generals enough.

meta-client/src/main/java/org/apache/helix/metaclient/api/DataChangeListener.java
```diff
@@ -24,9 +24,9 @@
  */
 public interface DataChangeListener {
   enum ChangeType {
-    NODE_CREATED,     // Node created of the specific path that the listener register to
-    NODE_DELETED,     // Node deleted of the specific path that the listener register to
-    VALUE_UPDATE      // Node value updated
+    ENTRY_CREATED,     // Node created of the specific path that the listener register to
+    ENTRY_DELETED,     // Node deleted of the specific path that the listener register to
+    ENTRY_UPDATE      // Node value updated
```
qqu0127, line 29: nit, might need to update the comments accordingly, node -> entry and children -> sub entry

meta-client/src/main/java/org/apache/helix/metaclient/api/DataChangeListener.java
```diff
@@ -0,0 +1,33 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Listener interface for events on a particular key, including entry creating, deleting and value change.
+ */
```
qqu0127, line 24: nit, comment styling, for class I believe it should start with /** instead of /*. Please follow the java coding standard

meta-client/src/main/java/org/apache/helix/metaclient/api/SubEntryChangeListener.java
```diff
@@ -0,0 +1,40 @@
...
+ * under the License.
+ */
+
+/*
+ * Listener interface for children change events on a particular key. It includes new subentry
+ * creation, subentry deletion, subentry data change and when the listener is removed.
+ * This listener type can only be registered as a persist listener.
+ * For hierarchy key spaces like zookeeper, it refers to an entry's entire subtree.
+ * For flat key spaces, it refers to keys that matches `prefix*`.
+ */
```
qqu0127, line 28: same

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectSubEntryChangeListener.java
```diff
@@ -0,0 +1,29 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Listener interface for direct children change event on a particular key. It includes new subentry creation or subentry deletion. The callback won't differentiate these types.
+ * For hierarchy key spaces like zookeeper, it refers to an entry's direct children nodes.
+ * For flat key spaces, it refers to keys that matches `prefix*separator`.
+ */
```
qqu0127, line 26: comment styling and wording

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectSubEntryChangeListener.java
```diff
@@ -0,0 +1,30 @@
...
+ * under the License.
+ */
+
+/**
+ * Listener interface for direct subentry change event on a particular key. It includes new
+ * subentry creation or subentry deletion. The callback won't differentiate these types.
+ * For hierarchy key spaces like zookeeper, it refers to an entry's direct subentry nodes.
+ * For flat key spaces, it refers to keys that matches `prefix*separator`.
+ */
+public interface DirectSubEntryChangeListener {
```
junkaixue, line 28: Subentry? Do you mean children？

meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
```diff
@@ -0,0 +1,55 @@
...
+
+
+  /**
+   * Called when the connection state has changed.
+   */
+  void handleConnectStateChanged(ConnectState currentState) throws Exception;
+  /**
+   * Called when new connection is established
+   */
+  void handleNewConnection(final String sessionId) throws Exception;
```
junkaixue, line 49: Are we allowing session ID passin? This is confusing...

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
```diff
@@ -20,139 +20,382 @@
...
-  void asyncCreate(final String key, T data, int version, long ttl,
+
+  /**
+   * The asynchronous version of create.
+   * @param key key to identify the entry
+   * @param data value of the entry
+   * @param mode EntryMode identifying if the entry will be deleted upon client disconnect
+   * @param cb An user defined VoidCallback implementation that will be invoked when async create return.
+   */
+  void asyncCreate(final String key, final T data, final EntryMode mode,
```
junkaixue, line 171: All these async APIs, shall we return some results? Or we rely on callbacks? How about exceptions happen?

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
```diff
@@ -20,139 +20,382 @@
...
+   */
+  void asyncDelete(final String key, AsyncCallback.VoidCallback cb);
 
-  void asyncTransaction(final String keys, AsyncCallback.TransactionCallback cb);
+  /**
+   * The asynchronous version of transaction operations.
+   * @param ops A list of operations
+   * @param cb An user defined TransactionCallback implementation that will be invoked when transaction operations return.
+   */
+  void asyncTransaction(final Iterable<Op> ops, AsyncCallback.TransactionCallback cb);
```
junkaixue, line 228: Especially for transaction, if any rollback operations required, do you expect the callback is the one user providing rollback operations? Or this callback is just designed for user return.

meta-client/src/main/java/org/apache/helix/metaclient/api/SubEntryChangeListener.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Listener interface for children change events on a particular key. It includes new subentry
+ * creation, subentry deletion, subentry data change and when the listener is removed.
+ * This listener type can only be registered as a persist listener.
+ * For hierarchy key spaces like zookeeper, it refers to an entry's entire subtree.
+ * For flat key spaces, it refers to keys that matches `prefix*`.
+ */
+public interface SubEntryChangeListener {
```
junkaixue, line 29: Same as commented before. I would suggest to use children instead of subentry. Because if this is not hierarchy data structure, then there is no sub entry concept. But if it is hierarchy structure, then children makes more sense as the world.

Even for prefix match, subentry is confusing word as "sub"

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
xyuanlu: We rely on callbacks. The return code in callback will include exception type. We remain the same behavior as ZK.
And will define the exceptions in following commits.

meta-client/src/main/java/org/apache/helix/metaclient/api/SubEntryChangeListener.java
xyuanlu: For non hierarchy data structure, doesn't subentry means prefix matched entries?

meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
xyuanlu: The original idea is to pass user the new session id when new connection is established. Since we are not supporting reading raw session ID in the first version, it might be better to remove this callback and only keep the `handleConnectStateChanged` as it includes the following two.

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
xyuanlu: Had an offline discussion with Junkai. TransactionCallback will be executed when the transactional operation finished, and the callback implementation should read the execution result return code and perform corresponding action. 
Will add more detail in comment section.

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectChildChangeListener.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Listener interface for direct child change event on a particular key. It includes new
+ * child creation or child deletion. The callback won't differentiate these types.
+ * For hierarchy key spaces like zookeeper, it refers to an entry's direct child nodes.
+ * For flat key spaces, it refers to keys that matches `prefix*separator`.
+ */
+public interface DirectChildChangeListener {
+   /**
+    * Called when there is a direct child entry creation or deleted.
+    * @param key The parent key where child change listener is subscribed.
```
junkaixue, line 31: Would this be full key as hierarchy or just one level key?

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectChildSubscribeResult.java
```diff
@@ -0,0 +1,41 @@
...
+    _children = children;
+    _isInstalled = isInstalled;
+  }
+
+  public List<String> getDirectChildren() {
+    return _children;
+  }
+
+  public boolean isInstalled() {
+    return _isInstalled;
```
junkaixue, line 39: What does it mean installed? Please have comments for that.

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
```diff
@@ -20,139 +20,414 @@
...
+   * @param key key to identify the entry
+   * @param cb An user defined VoidCallback implementation that will be invoked when async delete return.
+   */
+  void asyncDelete(final String key, AsyncCallback.VoidCallback cb);
 
-  void asyncTransaction(final String keys, AsyncCallback.TransactionCallback cb);
+  /**
+   * The asynchronous version of transaction operations.
+   * @param ops A list of operations
+   * @param cb An user defined TransactionCallback implementation that will be invoked when transaction operations return.
```
junkaixue, line 251: Please update the comment for behavior.

junkaixue (commented):
Overall, looks good to me. Please address some minor issues,

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectChildChangeListener.java
xyuanlu: TFTR. It is the parent key where the listener is registered.

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectChildSubscribeResult.java
```diff
@@ -0,0 +1,45 @@
...
+  public DirectChildSubscribeResult(List<String> children, boolean isRegistered) {
+    _children = children;
+    _isRegistered = isRegistered;
+  }
+
+  public List<String> getDirectChildren() {
+    return _children;
+  }
+
+  public boolean isInstalled() {
```
junkaixue, line 42: API change.

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectChildSubscribeResult.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+
+
+public class DirectChildSubscribeResult {
+  // A list of direct children names at the time when change is subscribed.
+  // It includes only one level child name, does not include further sub children names.
+  private final List<String> _children;
+
+  // true means the listener is registered successfully.
+  private final boolean _isRegistered;
```
junkaixue, line 31: I am not sure why we would like to have. registered here? Do you mean the result registered? What could be scenario failed to registered?

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectChildSubscribeResult.java
xyuanlu: TFTR. Updated.

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectChildSubscribeResult.java
xyuanlu: Per offline discussion, let's keep the behavior align with our current ZkClient APIs for now.

junkaixue (approved):
LGTM. Good job!

xyuanlu:
This PR is ready to be merged. Approved by @junkaixue // Thanks for the review!! 

Commit message: 
Add listener interfaces and factory class for metaclient interface

meta-client/src/main/java/org/apache/helix/metaclient/api/ChildChangeListener.java
```diff
@@ -0,0 +1,39 @@
...
+ * creation, child deletion, child data change.
+ * TODO: add type for persist listener is removed
+ * For hierarchy key spaces like zookeeper, it refers to an entry's entire subtree.
+ * For flat key spaces, it refers to keys that matches `prefix*`.
+ */
+public interface ChildChangeListener {
+  enum ChangeType {
+    ENTRY_CREATED,     // Any child entry created
+    ENTRY_DELETED,     // Any child entry deleted
+    ENTRY_DATA_CHANGE  // Any child entry has value change
```
rahulrane50, line 33: Neat. API doc also talks about "PERSIST_LISTENER_REMOVED", are we going to add this in future? also API documentation is missing with comments about what does it mean.

meta-client/src/main/java/org/apache/helix/metaclient/api/ChildChangeListener.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+/**
+ * Listener interface for children change events on a particular key. It includes new child
+ * creation, child deletion, child data change.
+ * TODO: add type for persist listener is removed
+ * For hierarchy key spaces like zookeeper, it refers to an entry's entire subtree.
+ * For flat key spaces, it refers to keys that matches `prefix*`.
+ */
+public interface ChildChangeListener {
```
rahulrane50, line 29: neat. API doc says "DirectChildChangeListener", either of them should be corrected.

meta-client/src/main/java/org/apache/helix/metaclient/api/ConnectStateChangeListener.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+public interface ConnectStateChangeListener {
+  /**
+   * Called when the connection state has changed. I
+   * @param prevState previous state before state change event.
+   * @param currentState client state after state change event. If it is a one time listsner, it is
+   *                     possible that the metaclient state changes again
+   */
+  void handleConnectStateChanged(MetaClientInterface.ConnectState prevState, MetaClientInterface.ConnectState currentState) throws Exception;
```
rahulrane50, line 29: Curious, why ConnectState is not defined here like other listener implementations?

rahulrane50: Also we don't have any "handleNewSession" handler so are we going to use same ("handleConnectStateChanged") handler when new connection/session is established? if that's the case then what would be prevState in that case because i don't see any "NOT_CONNECTED" state in ConnectState enum

meta-client/src/main/java/org/apache/helix/metaclient/api/DirectChildChangeListener.java
```diff
@@ -0,0 +1,37 @@
...
+ * under the License.
+ */
+
+/**
+ * Listener interface for direct child change event on a particular key. It includes new
+ * child creation or child deletion. The callback won't differentiate these types.
+ * For hierarchy key spaces like zookeeper, it refers to an entry's direct child nodes.
+ * For flat key spaces, it refers to keys that matches `prefix*separator`.
+ */
+public interface DirectChildChangeListener {
```
rahulrane50, line 28: Same as above comment. I can't figure out from comments about difference between ChildChangeListener and DirectChildChangeListener. Also API doc needs to be updated with both listeners i guess.

rahulrane50 (commented):
Otherwise LGTM! Great work on this big piece of item @xyuanlu!

meta-client/src/main/java/org/apache/helix/metaclient/api/AsyncCallback.java
```diff
@@ -22,28 +22,58 @@
...
 
 /**
- * An asynchronous callback is deferred to invoke after an async CRUD operation returns.
- * The corresponding callback is registered when async CRUD API is invoked.
+ * An asynchronous callback is deferred to invoke after an async CRUD operation finish and return.
+ * The corresponding callback is registered when async CRUD API is invoked. Implementation processes
+ * the result of each CRUD call. It should check return code and perform accordingly.
  */
+// TODO: define return code. failure code should map to MetaClient exceptions.
 public interface AsyncCallback {
```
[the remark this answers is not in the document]
xyuanlu: defaultCallback and CancellableAsyncCallback are Helix introduced wrapper.  We do not need this for generic ZK user

PR #2255, opened 2022-10-24
(WIP) Remove ThreadLeakageChecker and test improve

qqu0127:
### Issues

### Description

### Tests

### Documentation (Optional)

### Commits

### Code Quality


zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestSharedZkClient.java
```diff
@@ -40,7 +40,7 @@ public void beforeClass() throws IOException, InvalidRoutingDataException {
     _realmAwareZkClientFactory = SharedZkClientFactory.getInstance();
   }
 
-  @Test
+  @Test(dependsOnMethods = "testRealmAwareZkClientCreation")
```
NealSun96, line 43: Is there any point on keeping this test case at a different class? If it depends on _realmAwareZkClient that has to be created, then maybe this should just go to RealmAwareZkClientFactoryTestBase since that's what this is testing.

helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkConectionConfig.java
```diff
@@ -225,18 +220,6 @@ public void afterClass() throws Exception {
...
-    try {
-      status = ThreadLeakageChecker.afterClassCheck(testClassName);
-    } catch (Exception e) {
-      LOG.error("ThreadLeakageChecker exception:", e);
-    }
-    // todo: We should fail test here once we achieved 0 leakage and remove the following System print
-    if (!status) {
-      System.out.println(
-          "---------- Test Class " + testClassName + " thread leakage detected! ---------------");
-    }
```
NealSun96, line 239: About thread leakage: back in 2020 there was a huge effort to crack down thread leak, and I believe those code is a part of that effort. Now that we're removing it, how do we ensure that there's no thread leakage? What's the strategy here?

NealSun96 (changes requested):

helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkConectionConfig.java
micahstubbs: Thanks for sharing the history @NealSun96!

qqu0127: Thanks for the comments!
With all due respect, I feel either this tool doesn't work well, or we have serious thread leak problems. Right now one CI ([example](https://pipelines.actions.githubusercontent.com/serviceHosts/7ea409d0-72c7-46ee-906c-148950b1480d/_apis/pipelines/1/runs/10241/signedlogcontent/2?urlExpires=2022-10-25T18%3A34%3A12.7529931Z&urlSigningMethod=HMACV1&urlSignature=RsB0PvBcFCUHv%2F5G7nRtcsRCGOTjEYuk%2FL1Rlvi5Uzw%3D)), this has 295 instances of `thread leakage detected!`. And there is no action taken.
Also I don't see much value of printing this to System.out without failing the test, right now it's just informative, and to be honest, I simply ignore them.
My take is, we can probably keep the class, but remove it from runtime execution until we have fully resolved the current issues (either the code or the tooling).
Hope that makes sense.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/TestSharedZkClient.java
qqu0127: `RealmAwareZkClientFactoryTestBase` is extended in both `TestSharedZkClient` and `TestDedicatedZkClient`. Putting this method in the base class will cause duplicated test and logic change in its other subclass. So I think it's better to keep it this way. Thanks.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/ZkTestBase.java
```diff
@@ -124,7 +123,7 @@ private void setupZooKeepers() {
    * @param zkAddress
    * @return
    */
-  protected ZkServer startZkServer(final String zkAddress) {
+  protected synchronized static ZkServer startZkServer(final String zkAddress) {
```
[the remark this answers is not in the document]
qqu0127: Good question. This method is public to its subclass, and I do want to prevent creating with the same ZK address at the same time.

NealSun96 (approved):
LGTM. However, please track thread leak issues in another way; if we simply delete them, the effort in the past will be lost. We need to keep track of it and eventually resolve it like you said.

qqu0127:
This PR is ready to merge, approved by @NealSun96 

Commit message:
Fix potential race condition issue in unit test and other improvement.

PR #2291, opened 2022-11-22
[WIP] Add Meta client factory and ZkMetaClient constructor

xyuanlu:
### Issues

### Description

### Tests

### Documentation (Optional)

### Commits

### Code Quality


meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientConstants.java
```diff
@@ -0,0 +1,16 @@
+package org.apache.helix.metaclient.constants;
+
+public interface MetaClientConstants {
```
qqu0127, line 3: Add apache license, same for a few other classes

qqu0127, line 3: Also why do we make this class an interface?

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -56,56 +69,81 @@ public StoreType getStoreType() {
...
   }
 
-  public static class Builder {
-    private String _connectionAddress;
-
-    private long _connectionTimeout;
-    private boolean _enableAuth;
-    //private RetryProtocol _retryProtocol;
-    private StoreType _storeType;
+  public static class MetaClientConfigBuilder<B extends MetaClientConfigBuilder<B>> {
```
qqu0127, line 81: Could you please clarify why this generics is necessary? Thanks.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -31,15 +31,21 @@
...
+import org.apache.helix.zookeeper.impl.client.ZkClient;
+import org.apache.helix.zookeeper.zkclient.ZkConnection;
 
 
 public class ZkMetaClient implements MetaClientInterface {
 
-  private RealmAwareZkClient _zkClient;
-
-  public ZkMetaClient() {
+  private ZkClient _zkClient;
```
qqu0127, line 41: can this be final?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/factory/ZkMetaClientConfig.java
```diff
@@ -0,0 +1,126 @@
...
+  protected PathBasedZkSerializer _zkSerializer;
+
+  // Monitoring related fields. MBean names are crated using following variables in format of
+  // MonitorPrefix_monitorType_monitorKey_monitorInstanceName, where _monitorInstanceName is optional
+  // TODO: right now all zkClient mBean object has prefix `HelixZkClient` had coded. We should change
+  // it to a configurable name.
+  protected String _monitorType;
+  protected String _monitorKey;
+  protected String _monitorInstanceName = null;
+  protected boolean _monitorRootPathOnly = true;
```
qqu0127, line 21: Usually with builder pattern, we can make member variables final. Thanks.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/factory/ZkMetaClientFactory.java
```diff
@@ -0,0 +1,18 @@
...
+
+
+public class ZkMetaClientFactory extends MetaClientFactory {
+  @Override
+  public MetaClientInterface getMetaClient(MetaClientConfig config) {
+    if (config.getStoreType() == MetaClientConfig.StoreType.ZOOKEEPER
+        && config instanceof ZkMetaClientConfig) {
+      return new ZkMetaClient((ZkMetaClientConfig) config);
+    }
+    return null;
```
qqu0127, line 16: I'm thinking if we should just throw `IllegalArgumentException` instead of return null. What's your thought on this?

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientConstants.java
```diff
@@ -0,0 +1,16 @@
...
+
+  // Stop retrying when we reach timeout
+  int DEFAULT_OPERATION_RETRY_TIMEOUT = Integer.MAX_VALUE;
+
+  // maxMsToWaitUntilConnected
+  int DEFAULT_CONNECTION_INIT_TIMEOUT = 60 * 1000;
+
+  // When a client becomes partitioned from the metadata service for more than session timeout,
+  // new session will be established.
+  int DEFAULT_SESSION_TIMEOUT = 30 * 1000;
```
qqu0127, line 13: Are these all in millisecond? 
nit: maybe add comments or make it clear on the variable names?

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientConstants.java
```diff
@@ -0,0 +1,16 @@
+package org.apache.helix.metaclient.constants;
+
+public interface MetaClientConstants {
+
+  // Stop retrying when we reach timeout
+  int DEFAULT_OPERATION_RETRY_TIMEOUT = Integer.MAX_VALUE;
```
qqu0127, line 6: I think this value is inherited. But could we use a smaller timeout value for this? Open to discussion on this.

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
xyuanlu: If we are not using generics with builder, calling setter function defined in base class will returns a bace class object instead.

```
ZkMetaClientConfigBuilder.setAuthEnabled(true).setZkSerializer(serializer)
```
will throw exception. 

https://stackoverflow.com/questions/17164375/subclassing-a-java-builder-class

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientConstants.java
xyuanlu: I followed the current Helix constants class and uses interface class. It seems to be an anti pattern behavior after some research. 
Changed to a final class with private constructor.

xyuanlu: TFTR. Updated

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientConstants.java
xyuanlu: If we change to a smaller number, will it change user behavior when user witch to metaclient?

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
qqu0127: Thanks, that makes sense. This seems to be a very good use of generics. 
Can we make this class standalone? It's a `public static` anyway, that way might be more clear.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -31,15 +31,21 @@
...
-
-  public ZkMetaClient() {
+  private final ZkClient _zkClient;
 
+  public ZkMetaClient(ZkMetaClientConfig config) {
+    _zkClient =  new ZkClient(new ZkConnection(config.getConnectionAddress(),
+        (int) config.getSessionTimeoutInMillis()),
+        (int) config.getConnectionInitTimeoutInMillis(), -1 /*operationRetryTimeout*/,
+        config.getZkSerializer(), config.getMonitorType(), config.getMonitorKey(),
+        config.getMonitorInstanceName(), config.getMonitorRootPathOnly());
```
qqu0127, line 48: Curious what's our preference here for instantiation? With constructor or builder class?

xyuanlu: The preferred method for Client instantiation is through Factory class, however, user could use constructor as well.  The config can only be created using Builder. 
(Same as Current FederatedZkClient.)

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientConstants.java
qqu0127: Chatted offline, there is a similar change in current behavior https://github.com/apache/helix/pull/2293 
A shorter default timeout makes more sense and we should keep in sync

qqu0127 (approved):

xyuanlu:
This change is ready to be merged. Approved by @qqu0127 

Final commit message:
Add Meta client factory and ZkMetaClient constructor

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -19,23 +19,32 @@
  * under the License.
  */
 
-class MetaClientConfig {
+import org.apache.helix.metaclient.constants.MetaClientConstants;
+
+public class MetaClientConfig {
 
   public enum StoreType {
     ZOOKEEPER, ETCD, CUSTOMIZED
```
rahulrane50, line 27: Interesting to see "etcd" here. Should we mention this here yet or just say customized? (could there be licensing issue? )

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -19,23 +19,32 @@
...
-  private final long _connectionTimeout;
+
+  // Wait for init timeout time until connection is initiated
+  private final long _connectionInitTimeoutInMillis;
+
+  // When a client becomes partitioned from the metadata service for more than session timeout,
+  // new session will be established when reconnect.
+  private final long _sessionTimeoutInMillis;
+
   private final boolean _enableAuth;
```
rahulrane50, line 39: Should we add this now? I think we should add whole security APIs and config stuff later a single PR because just _enableAuth won't be enough but whole "AuthConfig" objection should be passed to builder class. This AuthConfig will have lot of different configs in it.

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -19,23 +19,32 @@
...
+
+public class MetaClientConfig {
 
   public enum StoreType {
     ZOOKEEPER, ETCD, CUSTOMIZED
   }
 
   private final String _connectionAddress;
-  private final long _connectionTimeout;
+
```
rahulrane50, line 31: Neat: would it be a good idea to add all "connection/session" related configs in separate class called ConnectionConfig, that way we would just need to pass ConnectionConfig in builder class and in future we can add more params related to connection/session without changing much code.

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -56,56 +69,82 @@ public StoreType getStoreType() {
...
+    public MetaClientConfigBuilder() {
       // set default values
       setAuthEnabled(false);
-      setConnectionTimeout(-1);
+      setConnectionInitTimeoutInMillis(MetaClientConstants.DEFAULT_CONNECTION_INIT_TIMEOUT_MS);
+      setSessionTimeoutInMillis(MetaClientConstants.DEFAULT_SESSION_TIMEOUT_MS);
     }
 
-    public Builder setConnectionAddress(String connectionAddress) {
+    public B setConnectionAddress(String connectionAddress) {
```
rahulrane50, line 106: Curious what is this "B"? I'm assuming it's template class but isn't it usually referred as "T"? Same question for line 81 as well

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/factory/ZkMetaClientConfig.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.helix.zookeeper.zkclient.serialize.PathBasedZkSerializer;
+import org.apache.helix.zookeeper.zkclient.serialize.SerializableSerializer;
+import org.apache.helix.zookeeper.zkclient.serialize.ZkSerializer;
+
+
+public class ZkMetaClientConfig extends MetaClientConfig {
+
+  protected final PathBasedZkSerializer _zkSerializer;
+
+  // Monitoring related fields. MBean names are crated using following variables in format of
```
rahulrane50, line 33: neat few typos in comments :)

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
```diff
@@ -56,56 +69,82 @@ public StoreType getStoreType() {
...
+      return self();
+    }
+
+    @SuppressWarnings("unchecked")
+    final B self() {
+      return (B) this;
     }
 
-    private void validate() {
+    protected void validate() {
```
rahulrane50, line 147: We should add more validations here right? Are you planning to do it in future? I think i can think of is having separate "validate" for each major config classes like "ConnectionConfig", "AuthConfig" etc and make use of them here.

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
xyuanlu: If i understand currectly, etcd should not have any license issue. It is ok to mention it here but we do not plan to have any implementation in near future.

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
xyuanlu: B as in Builder

meta-client/src/main/java/org/apache/helix/metaclient/factories/MetaClientConfig.java
xyuanlu: Yes we plan to add it later in future.

PR #2327, opened 2022-12-22
Meta client - implement CRUD for zkMetaClient

xyuanlu:
### Issues

#2237 

### Description

This change adds basic CRUD operation implementation for ZkMetaClient (excluding create)

### Tests

TestZkMetaClient

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
@@ -22,6 +22,9 @@ package org.apache.helix.metaclient.api;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.helix.metaclient.constants.MetaClientInterruptException;
+import org.apache.helix.metaclient.constants.MetaClientTimeoutException;
+
 
 public interface MetaClientInterface<T> {
 
@@ -349,7 +352,12 @@ public interface MetaClientInterface<T> {
   /**
    * Maintains a connection with underlying metadata service based on config params. Connection
    * created by this method will be used to perform CRUD operations on metadata service.
-   * @Throws MetaClientException when failed to connect.
+   * @throws MetaClientInterruptException
+   *          if the connection timed out due to thread interruption
+   * @throws MetaClientTimeoutException
+   *          if the connection timed out
+   * @throws IllegalStateException
+   *         if already connected or the connection is already closed explicitly
    */
   void connect();
 
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
@@ -36,7 +36,11 @@ public interface MetaClientInterface<T> {
 
     // The node will not be automatically deleted when the last sub-entry of the node is deleted.
     // The node is an ephemeral node.
-    CONTAINER
+    CONTAINER,
+
+    // If the entry is not modified within the TTL and has no children it will become a candidate
+    // to be deleted by the server at some point in the future.
+    TTL
   }
 
   enum ConnectState {
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
@@ -71,6 +71,11 @@ public interface MetaClientInterface<T> {
     public int getVersion() {
       return _version;
     }
+
+    public Stat (EntryMode mode, int version) {
+      _version = version;
+      _entryMode = mode;
+    }
   }
 
   //synced CRUD API
@@ -143,7 +148,7 @@ public interface MetaClientInterface<T> {
    * @eturn Return a list of children keys. Return direct child name only for hierarchical key
    *        space, return the whole sub key for non-hierarchical key space.
    */
-  List<String> getDirestChildrenKeys(final String key);
+  List<String> getDirectChildrenKeys(final String key);
 
   /**
    * Return the number of children for the given keys.
@@ -152,7 +157,7 @@ public interface MetaClientInterface<T> {
    *            For metadata storage that has non-hierarchical key space (e.g. etcd), the key would
    *            be a prefix key.
    */
-  int countDirestChildren(final String key);
+  int countDirectChildren(final String key);
 
   /**
    * Remove the entry associated with the given key.
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
@@ -34,12 +34,20 @@ public interface MetaClientInterface<T> {
     // An ephemeral node cannot have sub entry.
     PERSISTENT,
 
-    // The node will not be automatically deleted when the last sub-entry of the node is deleted.
+    // For metadata storage that has hierarchical key space (e.g. ZK), the node will be
+    // automatically deleted at some point in the future if the last child of the node is deleted.
+    // For metadata storage that has non-hierarchical key space (e.g. etcd), the node will be
+    // automatically deleted at some point in the future if the last entry that has the prefix
+    // is deleted.
     // The node is an ephemeral node.
     CONTAINER,
 
-    // If the entry is not modified within the TTL and has no children it will become a candidate
-    // to be deleted by the server at some point in the future.
+    // For metadata storage that has hierarchical key space (e.g. ZK) If the entry is not modified
+    // within the TTL and has no children it will become a candidate to be deleted by the server
+    // at some point in the future.
+    // For metadata storage that has non-hierarchical key space (e.g. etcd) If the entry is not modified
+    // within the TTL, it will become a candidate to be deleted by the server at some point in the
+    // future.
     TTL
   }
 
@@ -341,9 +349,9 @@ public interface MetaClientInterface<T> {
   /**
    * Maintains a connection with underlying metadata service based on config params. Connection
    * created by this method will be used to perform CRUD operations on metadata service.
-   * @return True if connection is successfully established.
+   * @Throws MetaClientException when failed to connect.
    */
-  boolean connect();
+  void connect();
 
   /**
    * Disconnect from server explicitly.
--- a/meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientBadVersionException.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientBadVersionException.java
@@ -1,0 +1,20 @@
+package org.apache.helix.metaclient.constants;
+
+public class MetaClientBadVersionException extends MetaClientException {
+  public MetaClientBadVersionException() {
+    super();
+  }
+
+  public MetaClientBadVersionException(String message, Throwable cause) {
+    super(message, cause);
+  }
+
+  public MetaClientBadVersionException(String message) {
+    super(message);
+  }
+
+  public MetaClientBadVersionException(Throwable cause) {
+    super(cause);
+  }
+
+}
--- a/meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientBadVersionException.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientBadVersionException.java
@@ -1,6 +1,6 @@
 package org.apache.helix.metaclient.constants;
 
-public class MetaClientBadVersionException extends MetaClientException {
+public final class MetaClientBadVersionException extends MetaClientException {
   public MetaClientBadVersionException() {
     super();
   }
--- a/meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientInterruptException.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientInterruptException.java
@@ -1,0 +1,19 @@
+package org.apache.helix.metaclient.constants;
+
+public final class MetaClientInterruptException extends MetaClientException {
+  public MetaClientInterruptException() {
+    super();
+  }
+
+  public MetaClientInterruptException(String message, Throwable cause) {
+    super(message, cause);
+  }
+
+  public MetaClientInterruptException(String message) {
+    super(message);
+  }
+
+  public MetaClientInterruptException(Throwable cause) {
+    super(cause);
+  }
+}
--- a/meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientInterruptException.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientInterruptException.java
@@ -1,5 +1,24 @@
 package org.apache.helix.metaclient.constants;
 
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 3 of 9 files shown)


meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientBadVersionException.java
```diff
@@ -0,0 +1,20 @@
+package org.apache.helix.metaclient.constants;
+
+public class MetaClientBadVersionException extends MetaClientException {
+  public MetaClientBadVersionException() {
+    super();
+  }
```
qqu0127, line 6: Same for others, do we really need such empty constructor?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -49,55 +61,97 @@ public ZkMetaClient(ZkMetaClientConfig config) {
...
+      return _zkClient.delete(key);
+    } catch (ZkException e) {
+      throw new MetaClientException(e);
+    }
   }
 
   @Override
   public boolean recursiveDelete(String key) {
-    return false;
+    _zkClient.deleteRecursively(key);
```
qqu0127, line 147: Just curious, can we have a return value from `deleteRecursively` ?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -155,12 +209,15 @@ public void asyncTransaction(Iterable iterable, AsyncCallback.TransactionCallbac
 
   @Override
   public boolean connect() {
-    return false;
+    // TODO: This is a tempp impl for test only. no proper event handling and error handling.
+    _zkClient.connect(Integer.MAX_VALUE, _zkClient);
+    return true;
```
qqu0127, line 214: For my own understanding, does this work with the current test?
( I remember during my own test, it fails if the zkclient has already connected, which is the case in native zkclient as it's connected during initialization.)

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestZkMetaClient.java
```diff
@@ -41,8 +53,125 @@ public void prepare() {
   }
 
   @Test
-  public void dummyTest() {
-    Assert.assertEquals(1+1, 2);
+  public void testGet() {
+    final String key = "/TestZkMetaClient_testGet";
+    ZkMetaClient<String> zkMetaClient = createZkMetaClient();
+    String value;
+    zkMetaClient.create(key, ENTRY_STRING_VALUE);
```
qqu0127, line 60: Following the connect() comment above, after the full implementation, metaclient should call connect() before any operation, right?

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestZkMetaClient.java
```diff
@@ -41,8 +53,125 @@ public void prepare() {
...
+    String dataValue = zkMetaClient.get(key);
+    Assert.assertEquals(dataValue, ENTRY_STRING_VALUE);
+
+    value = zkMetaClient.get(key + "/a/b/c");
+    Assert.assertNull(value);
+
+    zkMetaClient.delete(key);
+
+    value = zkMetaClient.get(key);
+    Assert.assertNull(value);
```
qqu0127, line 70: Should we close metaclient to avoid leakage in test? Maybe we can wrap the test in a try-with-resource block.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
xyuanlu: zkClient.deleteRecursively() has void as return value.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
xyuanlu: Thanks for bring this up. this connect is not called in test as of now. I plan to change it after your PR that includes constructor change is added.

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestZkMetaClient.java
xyuanlu: Good point. we should close it in afterClass.

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientBadVersionException.java
xyuanlu: Why not? `RuntimeException` has an empty constructor. We should provide one for user who may implement a metaclient wants to use it.

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestZkMetaClient.java
xyuanlu: yes. connect() is needed before the first operation.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -49,55 +61,97 @@ public ZkMetaClient(ZkMetaClientConfig config) {
...
+    return _zkClient.countChildren(key);
   }
 
   @Override
   public boolean delete(String key) {
-    return false;
+    try {
+      return _zkClient.delete(key);
+    } catch (ZkException e) {
+      throw new MetaClientException(e);
```
Marcosrico, line 141: Given that the `MetaClientException` is caught in the `translateZkExceptionToMetaclientException` method, what is the reason for not using the latter all the time? I assume readability is better if the exception is known to be a general one but perhaps then we would want to remain consistent with the other errors and catch/throw each individual type. Just a thought

xyuanlu: Good catch. I was planning to use `translateZkExceptionToMetaclientException` and it is a typo. :D

desaikomal (commented):
Can you please update the main template of the change?
It doesn't describe in detail what is being implemented.

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
```diff
@@ -36,7 +36,11 @@ enum EntryMode {
 
     // The node will not be automatically deleted when the last sub-entry of the node is deleted.
     // The node is an ephemeral node.
-    CONTAINER
+    CONTAINER,
+
+    // If the entry is not modified within the TTL and has no children it will become a candidate
+    // to be deleted by the server at some point in the future.
```
desaikomal, line 42: terminology is still very ZK like.  Pure KV doesn't have concept of "child" node. I understand that for ZK, we need to maintain as Helix is using it.

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
```diff
@@ -143,7 +152,7 @@ public int getVersion() {
    * @eturn Return a list of children keys. Return direct child name only for hierarchical key
    *        space, return the whole sub key for non-hierarchical key space.
    */
-  List<String> getDirestChildrenKeys(final String key);
+  List<String> getDirectChildrenKeys(final String key);
```
desaikomal, line 155: this is very confusing. for non-hierarchical, shouldn't it return NotImplemented ?

desaikomal (commented):
KV store which are not like ZK(directory structure), we should throw Not implemented for all child-related apis. It is confusing as 'getChildren()' and 'get()' will return same result and user will get confused.

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
xyuanlu: TFTR. We want to keep the possibility that user may want to mimic hierarchical key space using a non-hierarchical KV store. (e.g. Helix explored the possibility of using etcd). In this case, user will pass a prefix as key the `getDirectChildrenKeys` returns all keys matches the prefix.

desaikomal: Got it. So interface will still say, getChildren() but Etcd will implement using prefix search.. in that case, name is very confusing for real metadata client user. For example, kv store, will they use 'getchildren() or should use getprefixsearch() or equivalent.

xyuanlu: They will use `getchildren()` as there is no `getprefixsearch ()` in the interface.  We could add that for specific impl in the future.

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientInterruptException.java
```diff
@@ -0,0 +1,19 @@
+package org.apache.helix.metaclient.constants;
+
+public final class MetaClientInterruptException extends MetaClientException {
```
qqu0127, line 3: add apache license

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientNoNodeException.java
```diff
@@ -0,0 +1,20 @@
+package org.apache.helix.metaclient.constants;
+
+public final class MetaClientNoNodeException extends MetaClientException {
```
qqu0127, line 3: license

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientTimeoutException.java
```diff
@@ -0,0 +1,20 @@
+package org.apache.helix.metaclient.constants;
+
+public final class MetaClientTimeoutException extends MetaClientException {
```
qqu0127, line 3: license

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
```diff
@@ -332,9 +349,9 @@ void asyncCreate(final String key, final T data, final EntryMode mode,
   /**
    * Maintains a connection with underlying metadata service based on config params. Connection
    * created by this method will be used to perform CRUD operations on metadata service.
-   * @return True if connection is successfully established.
+   * @Throws MetaClientException when failed to connect.
    */
-  boolean connect();
+  void connect();
```
qqu0127, line 354: shall we mark exception type at the method?

qqu0127: also nit, some behavior clarification on repeated connect() will be good.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -301,4 +363,34 @@ public int hashCode() {
       return _listener.hashCode();
     }
   }
+  private MetaClientException translateZkExceptionToMetaclientException(ZkException e) {
```
qqu0127, line 366: nit: empty line
Also this method can be static?

qqu0127 (commented):
Overall looks good, commented on some nits

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
xyuanlu: TFTR. Added a `todo`. Will add the connect/disconnect part in following PR since it needs a bit more error handling and it is orthogonal to this CRUD change.

qqu0127 (approved):
LGTM, please make sure no CI regression before merge.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -30,76 +30,133 @@
...
+      _zkClient.writeData(key, data, version);
+    } catch (ZkException e) {
+      throw translateZkExceptionToMetaclientException(e);
+    }
   }
 
   @Override
-  public Object update(String key, DataUpdater updater) {
-    return null;
+  public T update( String key, DataUpdater<T> updater) {
```
rahulrane50, line 90: shouldn't this T extend an Object class?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -30,76 +30,133 @@
...
+      T newData = updater.update(oldData);
+      set(key, newData, stat.getVersion());
+      return newData;
+    } catch (ZkException e) {
+      throw translateZkExceptionToMetaclientException(e);
+    }
   }
 
   @Override
   public Stat exists(String key) {
```
rahulrane50, line 104: Umm now I see implementation of "exists" i think more about is it "check if this key exists" API or it's more like "get this key if exists". Implementation suggests later case then i think name could be different.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
rahulrane50: This comment at all other places as well.

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
```diff
@@ -332,9 +352,14 @@ void asyncCreate(final String key, final T data, final EntryMode mode,
   /**
    * Maintains a connection with underlying metadata service based on config params. Connection
    * created by this method will be used to perform CRUD operations on metadata service.
-   * @return True if connection is successfully established.
-   */
-  boolean connect();
```
rahulrane50, line 337: Any reason we changed return type from boolean to void?

rahulrane50 (commented):
@xyuanlu can you share some context on why do we have separate exception files per exception? (I understand we used to have it back in helix but do you think we can combine it and reduce some redundant code?)

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientBadVersionException.java
```diff
@@ -0,0 +1,20 @@
+package org.apache.helix.metaclient.constants;
+
+public final class MetaClientBadVersionException extends MetaClientException {
```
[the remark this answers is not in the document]
qqu0127: Yes, it should be included in an IDE setup.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
xyuanlu: There is no fundamental difference between these two in our use case. Since `Object` is a base class for all class, we should just use `T` instead of the other. (will change all other existing `T extends Object` later.. )

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
xyuanlu: We keep the API and behavior of current ZK. I feel it more easy for user to adopt if we keep the same behavior.

rahulrane50 (approved):
LGTM! Thanks @xyuanlu great work!

meta-client/src/main/java/org/apache/helix/metaclient/api/MetaClientInterface.java
xyuanlu: Based on offline discussion, let's keep the naming for now and may add more API when needed in future.

Issue #2240, opened 2022-10-18
Adding Transactional Support to Helix Zk Client


Marcosrico:
Implementing zk multi for the various helix zk clients to add transactional support to helix zk clients.

PR #2328, opened 2022-12-23
Multi Implementation and Test Cases

Marcosrico:
### Issues

https://github.com/apache/helix/issues/2240

### Description

Implemented the logic for zk.multi to enable zk transactional support.

### Tests

Wrote test cases to test transactional support in all clients (dedicated, shared, and federated).

### Code Quality

### DIFF
--- a/helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiInMultiZk.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiInMultiZk.java
@@ -63,6 +63,7 @@ public class TestMultiInMultiZk extends MultiZkTestBase {
                 System.out.println(elm);
             }
         }
+        System.out.println("end start");
     }
 
     /**
@@ -87,9 +88,10 @@ public class TestMultiInMultiZk extends MultiZkTestBase {
         try {
             //Execute transactional support on operations and verify they were run
             _zkClient.multi(ops);
-            Assert.fail("Should have thrown an exception. Multi not supported");
-        } catch (UnsupportedOperationException ex) {
-            Assert.assertTrue(ex.getMessage().startsWith("Session-aware operation is not supported by FederatedZkClient."));
+            Assert.fail("Should have thrown an exception. Cannot run multi on ops of different servers.");
+        } catch (IllegalArgumentException e) {
+            boolean pathExists = _zkClient.exists("/" + CLUSTER_LIST.get(0) + "/test");
+            Assert.assertFalse(pathExists, "Path should not have been created.");
         }
         System.out.println("END " + _className + "_" + methodName + " at " + new Date(System.currentTimeMillis()));
     }
--- a/helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiInMultiZk.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiInMultiZk.java
@@ -63,7 +63,6 @@ public class TestMultiInMultiZk extends MultiZkTestBase {
                 System.out.println(elm);
             }
         }
-        System.out.println("end start");
     }
 
     /**
--- a/helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkHelixJavaApis.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkHelixJavaApis.java
@@ -78,7 +78,7 @@ import org.testng.annotations.Test;
  * cluster-Zk realm routing information.
  * This test verifies that all Helix Java APIs work as expected.
  */
-public abstract class TestMultiZkHelixJavaApis extends TestMultiZkConnectionConfig {
+public class TestMultiZkHelixJavaApis extends TestMultiZkConnectionConfig {
   // For testing different MSDS endpoint configs.
   private static final String CLUSTER_ONE = CLUSTER_LIST.get(0);
   private static final String CLUSTER_FOUR = "CLUSTER_4";
--- a/helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkHelixJavaApis.java
+++ b/helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkHelixJavaApis.java
@@ -32,7 +32,13 @@ import java.util.Set;
 import java.util.Date;
 
 import com.google.common.collect.ImmutableMap;
-import org.apache.helix.*;
+import org.apache.helix.AccessOption;
+import org.apache.helix.BaseDataAccessor;
+import org.apache.helix.ConfigAccessor;
+import org.apache.helix.HelixAdmin;
+import org.apache.helix.HelixException;
+import org.apache.helix.InstanceType;
+import org.apache.helix.TestHelper;
 import org.apache.helix.api.config.RebalanceConfig;
 import org.apache.helix.controller.rebalancer.DelayedAutoRebalancer;
 import org.apache.helix.controller.rebalancer.strategy.CrushEdRebalanceStrategy;
@@ -72,7 +78,7 @@ import org.testng.annotations.Test;
  * cluster-Zk realm routing information.
  * This test verifies that all Helix Java APIs work as expected.
  */
-public class TestMultiZkHelixJavaApis extends TestMultiZkConnectionConfig {
+public abstract class TestMultiZkHelixJavaApis extends TestMultiZkConnectionConfig {
   // For testing different MSDS endpoint configs.
   private static final String CLUSTER_ONE = CLUSTER_LIST.get(0);
   private static final String CLUSTER_FOUR = "CLUSTER_4";
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
@@ -477,10 +477,31 @@ public class FederatedZkClient implements RealmAwareZkClient {
     return getZkClient(path).getCreationTime(path);
   }
 
+  /**
+   * Executes ZkMulti on operations that are connected to the same Zk server.
+   * Will throw exception if any operation's server connection is different.
+   * @param ops
+   * @return
+   * @throws IllegalArgumentException
+   */
   @Override
   public List<OpResult> multi(Iterable<Op> ops) {
-    throwUnsupportedOperationException();
-    return null;
+    if (ops == null) {
+      throw new NullPointerException("ops must not be null.");
+    }
+    String opPath = null;
+    String opPathRealm = null;
+    for (Op op : ops) {
+      if (opPath == null) {
+        opPath = op.getPath();
+        opPathRealm = getZkRealm(op.getPath());
+      } else {
+        if (!opPathRealm.equals(getZkRealm(op.getPath()))){
+          throw new IllegalArgumentException("Cannot execute multi on ops of different realms!");
+        }
+      }
+    }
+    return getZkClient(opPath).multi(ops);
   }
 
   @Override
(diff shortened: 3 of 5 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -477,10 +477,31 @@ public long getCreationTime(String path) {
...
-    return null;
+    if (ops == null) {
+      throw new NullPointerException("ops must not be null.");
+    }
+    String opPath = null;
+    String opPathRealm = null;
+    for (Op op : ops) {
+      if (opPath == null) {
+        opPath = op.getPath();
+        opPathRealm = getZkRealm(op.getPath());
```
rahulrane50, line 497: Neat: we already got opPath on line 496, can we re-use it here?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -477,10 +477,31 @@ public long getCreationTime(String path) {
...
   @Override
   public List<OpResult> multi(Iterable<Op> ops) {
-    throwUnsupportedOperationException();
-    return null;
+    if (ops == null) {
+      throw new NullPointerException("ops must not be null.");
+    }
+    String opPath = null;
+    String opPathRealm = null;
+    for (Op op : ops) {
```
rahulrane50, line 494: From my understanding of code, in this for loop you want to filter ops who have same ZkRealm. Just wondering if there could be case where first few ops in given ops list won't have any realm but later ones have it. In that case, here even those first few ops will be sent for "multi" operation right? Should we use any lambda filter method which filters out all ops with same ZkRealm.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
```diff
@@ -477,10 +477,31 @@ public long getCreationTime(String path) {
...
+      if (opPath == null) {
+        opPath = op.getPath();
+        opPathRealm = getZkRealm(op.getPath());
+      } else {
+        if (!opPathRealm.equals(getZkRealm(op.getPath()))){
+          throw new IllegalArgumentException("Cannot execute multi on ops of different realms!");
+        }
+      }
+    }
+    return getZkClient(opPath).multi(ops);
```
rahulrane50, line 504: also this getZkClient might fail if none of the ops in provided list have valid opPath right?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/RealmAwareZkClientTestBase.java
```diff
@@ -67,4 +76,127 @@ public void afterClass() {
...
+    Assert.assertTrue(opResults.get(1) instanceof OpResult.CreateResult);
+
+    cleanup();
+  }
+
+  /**
+   * Multi should be an all or nothing transaction. Creating correct
+   * paths and a singular bad one should all fail.
+   */
+  @Test(dependsOnMethods = "testMultiCreate")
```
rahulrane50, line 125: Curious, why does it depends on "testMultiCreate"?

rahulrane50 (commented):
Overall great works!

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
Marcosrico: I think an op must have a ZkRealm in order to be valid. Trying to get the path of an op with no ZkRealm would throw an error (in `TrieRoutingData.getMetadataStoreRealm()`) and so the multi operation won't succeed.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
Marcosrico: getZkRealm() method checks whether the input path is a valid opPath so it would fail before entering this call. Perhaps we should write a catch / throw clause in the case that occurs?

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/RealmAwareZkClientTestBase.java
Marcosrico: It doesn't necessarily have to depend on "testMultiCreate" but my intention was to make sure each test completes in the same order to avoid any race conditions where we try and create an op with a path that already exists (and hasn't been cleaned up yet). I think this achieves it?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
mgao0, line 494: I'll suggest to implement this logic like this 
`realms = ops.stream()
     .map(op -> getZkRealm(op.getPath()))
     .distinct()
     .collect(toList());`
`if (realms.size() == 1) {
....
} else {
....
}`
This might be a cleaner solution from the code aspect, but your implementation maybe able to fail early if there are different realms. Just a thought, you can think about the trade-offs.

helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiZkHelixJavaApis.java
```diff
@@ -72,7 +78,7 @@
  * cluster-Zk realm routing information.
  * This test verifies that all Helix Java APIs work as expected.
  */
-public class TestMultiZkHelixJavaApis extends TestMultiZkConnectionConfig {
+public abstract class TestMultiZkHelixJavaApis extends TestMultiZkConnectionConfig {
```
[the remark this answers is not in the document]
Marcosrico: No reason, I don't remember changing it. Thanks for the catch!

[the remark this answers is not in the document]
qqu0127: has this been resolved yet?

helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiInMultiZk.java
```diff
@@ -63,6 +63,7 @@ public void beforeClass() throws Exception {
                 System.out.println(elm);
             }
         }
+        System.out.println("end start");
```
qqu0127, line 66: Can we remove this?

qqu0127 (commented):
Overall looks good, please make sure other comments are resolved.

helix-core/src/test/java/org/apache/helix/integration/multizk/TestMultiInMultiZk.java
Marcosrico: Yup!

zookeeper-api/src/main/java/org/apache/helix/zookeeper/impl/client/FederatedZkClient.java
Marcosrico: Thanks for the suggestion! The solution looks cleaner and while my past solution might fail early, the cost of getting the zkrealm is minimum. I modified your implementation slightly, let me know if it looks good!

rahulrane50 (approved):

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/RealmAwareZkClientTestBase.java
```diff
@@ -67,4 +76,127 @@ public void afterClass() {
...
+            new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder();
+    RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig =
+            builder.setZkRealmShardingKey(ZK_SHARDING_KEY_PREFIX).build();
+    try {
+      _realmAwareZkClient = new FederatedZkClient(connectionConfig,
+              new RealmAwareZkClient.RealmAwareZkClientConfig());
+    } catch (IllegalArgumentException e) {
+      Assert.fail("Invalid Sharding Key.");
+    } catch (Exception e) {
+      Assert.fail("Should not see any other types of Exceptions: " + e);
```
xyuanlu, line 95: If this test class is only for setting up test env, I would suggest a better name.

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/RealmAwareZkClientTestBase.java
```diff
@@ -67,4 +76,127 @@ public void afterClass() {
...
+    RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder builder =
+            new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder();
+    RealmAwareZkClient.RealmAwareZkConnectionConfig connectionConfig =
+            builder.setZkRealmShardingKey(ZK_SHARDING_KEY_PREFIX).build();
+    try {
+      _realmAwareZkClient = new FederatedZkClient(connectionConfig,
+              new RealmAwareZkClient.RealmAwareZkClientConfig());
+    } catch (IllegalArgumentException e) {
+      Assert.fail("Invalid Sharding Key.");
+    } catch (Exception e) {
```
xyuanlu, line 94: can we integrate these 2 catches?

Marcosrico: how would you integrate both? Each catch is to specify a different error origin

qqu0127: Just to share a different thought, if the exception is abnormal, maybe we don't need to catch and check at all. Just let it throw.

xyuanlu: If you want to integrate these two, then you can just catch `Exception ` since `IllegalArgumentException` is a subclass of general `Exception ` ( https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html)
Throw the unexpected exception also work as the stack trace will contain enough info. 
In my opinion both work.

Marcosrico: Sound good, I think think I'll have the exceptions be thrown without catching them then. Thank you for the input!

zookeeper-api/src/test/java/org/apache/helix/zookeeper/impl/client/RealmAwareZkClientTestBase.java
```diff
@@ -67,4 +76,132 @@ public void afterClass() {
...
+                    ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT),
+            Op.delete(PARENT_PATH + test_name, -1));
+
+    List<OpResult> opResults = _realmAwareZkClient.multi(ops);
+    Assert.assertTrue(opResults.get(0) instanceof OpResult.CreateResult);
+    Assert.assertTrue(opResults.get(1) instanceof OpResult.CreateResult);
+    Assert.assertTrue(opResults.get(2) instanceof OpResult.DeleteResult);
+
+    Assert.assertTrue(_realmAwareZkClient.exists(PARENT_PATH), "Path has not been created.");
+    Assert.assertFalse(_realmAwareZkClient.exists(PARENT_PATH + test_name), "Path should have been removed.");
```
xyuanlu, line 166: I guess the message should be the opposite?

Marcosrico: I think the message only shows when the condition isn't true. For example, if PARENT_PATH does not exist, then it would error with "Path has not been created"

xyuanlu (approved):

Marcosrico:
This PR is ready to merge, approved by @xyuanlu and @rahulrane50 
Commit message:
Federated Zk Client Multi implementation for Zk transactional support.

PR #2338, opened 2023-01-10
Implement a thread-safe listener container for zkclient

qqu0127:
### Issues

N.A.

### Description

Implement a thread-safe listener container for zkclient listener management.
This is introduced to better support metaclient implementation, where:
1. New functionalities added with the introduction of "one-time" listener in metaclient (now in zkclient, all listeners are "persistent")
2. Better encapsulation and code refactoring

This PR added the main class `ListenerContainer` that contains the core functionalities.
The next PR will be have zkclient adopt this new component, then implement the new funcs for one-time listener.

### Tests

`TestListenerContainer`

[INFO] Tests run: 78, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1,115.267 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 78, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] --- jacoco-maven-plugin:0.8.6:report (generate-code-coverage-report) @ zookeeper-api ---
[INFO] Loading execution data file /Users/qqu/workspace/qqu-helix/zookeeper-api/target/jacoco.exec
[INFO] Analyzed bundle 'Apache Helix :: ZooKeeper API' with 116 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  18:50 min
[INFO] Finished at: 2023-01-10T14:48:02-05:00

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
@@ -1,0 +1,186 @@
+package org.apache.helix.zookeeper.zkclient;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
+
+
+/**
+ * A thread-safe container wrapper class for Listeners registered to ZkClient.
+ * It stores 2 types of listener separately, one-time listener and persistent listener. The former ones are removed
+ * right after its consumption, while the latter need to be explicitly removed. 
+ * @param <T> the type of listener
+ */
+public class ListenerContainer<T> {
+  private final ReentrantLock _lock = new ReentrantLock(true);
+  private final Map<String, Set<T>> _persistentListener = new ConcurrentHashMap<>();
+  private final Map<String, Set<T>> _onetimeListener = new ConcurrentHashMap<>();
+
+  /**
+   * Add listener to the container with specified key.
+   * @param key the key to register to
+   * @param listener the listener to register
+   * @param persistListener true if the listener is persistent
+   */
+  public void addListener(String key, T listener, boolean persistListener) {
+    addListener(key, listener, persistListener ? _persistentListener : _onetimeListener);
+  }
+
+  /**
+   * Remove listener from the container.
+   * This operation removes both one-time and persistent listener from the specified key.
+   * @param key the key to remove
+   * @param listener the listener to remove
+   */
+  public void removeListener(String key, T listener) {
+    _lock.lock();
+    try {
+      removeFromListenerMap(key, listener, _persistentListener);
+      removeFromListenerMap(key, listener, _onetimeListener);
+    } finally {
+      _lock.unlock();
+    }
+  }
+
+  /**
+   * Remove listener from the container.
+   * @param key the key to remove
+   * @param listener the listener to remove
+   * @param persistListener true if remove the persistent listener, otherwise remove the one-time listener
+   */
+  public void removeListener(String key, T listener, boolean persistListener) {
+    _lock.lock();
+    try {
+      if (persistListener) {
+        removeFromListenerMap(key, listener, _persistentListener);
+      } else {
+        removeFromListenerMap(key, listener, _onetimeListener);
+      }
+    } finally {
+      _lock.unlock();
+    }
+  }
+
+  /**
+   * Remove listeners from the container.
+   * @param key the key to remove
+   * @param listeners the listeners to remove
+   * @param persistListener true if remove the persistent listener, otherwise remove the one-time listener
+   */
+  public void removeListeners(String key, Collection<T> listeners, boolean persistListener) {
+    _lock.lock();
+    try {
+      if (persistListener) {
+        removeFromListenerMap(key, listeners, _persistentListener);
+      } else {
+        removeFromListenerMap(key, listeners, _onetimeListener);
+      }
+    } finally {
+      _lock.unlock();
+    }
+  }
+
+  /**
+   * Consume the listeners registered to the given key. The given consumer is applied to all one-time and persistent
+   * listeners with the given key.
+   * The one-time listeners are removed after this operation, as a result, this operation is NOT idempotent.
+   * @param key the key of the listeners
+   * @param consumer the consuming action for the listeners
+   */
+  public void consumeListeners(String key, Consumer<? super T> consumer) {
+    Set<T> onetimeListeners;
+    Set<T> persistentListeners;
+    _lock.lock();
+    try {
+      // remove one-time listeners
+      onetimeListeners = _onetimeListener.remove(key);
+    } finally {
+      _lock.unlock();
+    }
+    if (onetimeListeners == null) {
+      onetimeListeners = Collections.emptySet();
+    }
+    persistentListeners = _persistentListener.getOrDefault(key, Collections.emptySet());
+    if (persistentListeners.isEmpty() && onetimeListeners.isEmpty()) {
+      return;
+    }
+    Set<T> listeners = new HashSet<>(onetimeListeners);
+    listeners.addAll(persistentListeners);
+    listeners.forEach(consumer);
+  }
+
+  public boolean isEmpty(String key) {
+    return !_persistentListener.containsKey(key) && !_onetimeListener.containsKey(key);
+  }
+
+  private void addListener(String key, T listener, Map<String, Set<T>> listenerMap) {
+    _lock.lock();
+    try {
+      Set<T> entryListeners = listenerMap.computeIfAbsent(key, k -> new CopyOnWriteArraySet<>());
+      entryListeners.add(listener);
+    } finally {
+      _lock.unlock();
+    }
+  }
+
+  private static <T> void removeFromListenerMap(String key, T listener, Map<String, Set<T>> listenerMap) {
+    Set<T> listeners = listenerMap.get(key);
+    if (listeners == null) {
+      return;
+    }
+    listeners.remove(listener);
+    if (listeners.isEmpty()) {
+      listenerMap.remove(key);
+    }
+  }
+
+  private static <T> void removeFromListenerMap(String key, Collection<T> listeners, Map<String, Set<T>> listenerMap) {
+    if (listeners == null || listeners.isEmpty()) {
+      return;
+    }
+    Set<T> current = listenerMap.get(key);
+    if (current == null || current.isEmpty()) {
+      return;
+    }
+    current.removeAll(listeners);
+    if (current.isEmpty()) {
+      listenerMap.remove(key);
+    }
+  }
+
+  @VisibleForTesting
+  Map<String, Set<T>> getPersistentListener() {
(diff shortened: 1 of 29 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
```diff
@@ -0,0 +1,186 @@
...
+    try {
+      // remove one-time listeners
+      onetimeListeners = _onetimeListener.remove(key);
+    } finally {
+      _lock.unlock();
+    }
+    if (onetimeListeners == null) {
+      onetimeListeners = Collections.emptySet();
+    }
+    persistentListeners = _persistentListener.getOrDefault(key, Collections.emptySet());
```
rahulrane50, line 129: just curious, why do we need lock for onetimelisteners and not here? Also we can just use getOrDefault for onetimelisteners as well right?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
```diff
@@ -0,0 +1,186 @@
...
+      } else {
+        removeFromListenerMap(key, listeners, _onetimeListener);
+      }
+    } finally {
+      _lock.unlock();
+    }
+  }
+
+  /**
+   * Consume the listeners registered to the given key. The given consumer is applied to all one-time and persistent
```
rahulrane50, line 110: Umm this are great APIs but just few questions : 
1. Does it make sense or would there be scenarios where both persistent as well as one time listeners will be added for same key?
2. If above case can be true then it would be interesting to know the behaviour of having multiple listeners for same key. My guess is that should be okay but just calling this out.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
```diff
@@ -0,0 +1,186 @@
...
+
+
+/**
+ * A thread-safe container wrapper class for Listeners registered to ZkClient.
+ * It stores 2 types of listener separately, one-time listener and persistent listener. The former ones are removed
+ * right after its consumption, while the latter need to be explicitly removed. 
+ * @param <T> the type of listener
+ */
+public class ListenerContainer<T> {
+  private final ReentrantLock _lock = new ReentrantLock(true);
```
rahulrane50, line 41: I see that this lock is used only once in below APIs. just commented there and also would like to know if this is really needed?

rahulrane50 (commented):
Very clean APIs and great documentation with tests! @qqu0127

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
qqu0127: For one-time listener, it's a write operation which removes the entry, we'll need lock. For persistent listener, it's pure read, we are able to go lock-free with the specified thread-safe primitives.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
```diff
@@ -0,0 +1,186 @@
...
+   * @param key the key of the listeners
+   * @param consumer the consuming action for the listeners
+   */
+  public void consumeListeners(String key, Consumer<? super T> consumer) {
+    Set<T> onetimeListeners;
+    Set<T> persistentListeners;
+    _lock.lock();
+    try {
+      // remove one-time listeners
+      onetimeListeners = _onetimeListener.remove(key);
```
Marcosrico, line 122: Please correct me if I'm wrong but I believe the remove method in ConcurrentHashMap are atomic in itself so not sure if it is necessary to lock beforehand.

Also should we check if key is null? Given underlying data structure of a ConcurrentHashMap is a HashTable, attempting to remove a null key would throw a NullPointerException which might not be the desired effect. Just a thought

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
```diff
@@ -0,0 +1,186 @@
...
+    addListener(key, listener, persistListener ? _persistentListener : _onetimeListener);
+  }
+
+  /**
+   * Remove listener from the container.
+   * This operation removes both one-time and persistent listener from the specified key.
+   * @param key the key to remove
+   * @param listener the listener to remove
+   */
+  public void removeListener(String key, T listener) {
```
desaikomal, line 61: If we remove one-time listener, lot of the code will become simpler.
Can you first please let me know who needs one-time listener?

thanks,

desaikomal (commented):
Do we need this one-time Listener?
Is Helix or Curator has use for this listener?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
qqu0127: Good question, I think there is a subtle race condition between `consume` and `add`. When thread A adds a one-time listener T, while thread B consume it as in line 122, if the timing is unfortunate, `onetimeListener` may be getting the snapshot before A completes, which doesn't contain T. However, because the entry together with the COWS instance is removed, the new listener T is lost.
(The unit test fails if we remove the lock)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
+
+
+/**
+ * A thread-safe container wrapper class for Listeners registered to ZkClient.
+ * It stores 2 types of listener separately, one-time listener and persistent listener. The former ones are removed
```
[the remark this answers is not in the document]
qqu0127: No one is using it at the moment. I get the idea we can skip the certain impl under time pressure.
This is implemented to be compatible with the API. We chatted with Junkai offline, I think we can just keep the code and not use this class. No existing class is using it, while I think it's still valuable in the future.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
```diff
@@ -0,0 +1,186 @@
...
+ * right after its consumption, while the latter need to be explicitly removed. 
+ * @param <T> the type of listener
+ */
+public class ListenerContainer<T> {
+  private final ReentrantLock _lock = new ReentrantLock(true);
+  private final Map<String, Set<T>> _persistentListener = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
qqu0127: ConcurrentHashMap has instance level thread-safety, but it doesn't guarantee thread-safety in its value set. The lock is required, you can see https://github.com/apache/helix/pull/2338#discussion_r1068357047

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
desaikomal: for deterministic test behavior, i think, lock will help. but we code for correctness of the system.

If you are going to use lock explicitly, then you can use regular HashMap. We are paying the cost twice as ConcurrentHashMap implementation has lock per segment:
https://github.com/frohoff/jdk8u-jdk/blob/master/src/share/classes/java/util/concurrent/ConcurrentHashMap.java#L1370

qqu0127: Thanks @desaikomal for the pointer. Short answer on this, we need the lock for correctness, and it's the current impl.
The use of CopyOnWriteArraySet as entry in ConcurrentHashMap requires a lock to make the overall structure thread-safe. Otherwise there is race condition between consume and add, and listener might be lost. it's also the same for current implementation, please see https://github.com/apache/helix/blob/master/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java#L290

On the other hand, the use of HashMap instead of ConcurrentHashMap could work, but it requires we use a read lock on iteration read. In our use case, reads vastly out number writes, so a lock-free read is preferable.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
```diff
@@ -0,0 +1,186 @@
...
+  /**
+   * Remove listener from the container.
+   * This operation removes both one-time and persistent listener from the specified key.
+   * @param key the key to remove
+   * @param listener the listener to remove
+   */
+  public void removeListener(String key, T listener) {
+    _lock.lock();
+    try {
+      removeFromListenerMap(key, listener, _persistentListener);
```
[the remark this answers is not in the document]
qqu0127: Thanks for the review @jiajunwang !
For your first point, this class is generics, it can support any type of listener already.
Second, sync block might be better compared to the whole method, that should be equivalent. For locking it self, we need synchronization on most writes operation even if there are concurrent hashmap. Please see other comment for more details.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+  }
+
+  /**
+   * Remove listener from the container.
+   * @param key the key to remove
+   * @param listener the listener to remove
+   * @param persistListener true if remove the persistent listener, otherwise remove the one-time listener
+   */
+  public void removeListener(String key, T listener, boolean persistListener) {
```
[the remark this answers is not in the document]
jiajunwang: What I meant is that the method should look like this,
"public void removeListener(String key, T listener)"

Because caller already specify the "persistListener" type when adding the listener. The main problem of this design is, if caller specify a different type when request to remove, the listener won't be removed and the caller won't know.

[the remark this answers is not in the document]
qqu0127: The method you described is already there, it removes from both entries. This one is implemented considering some clients might want to remove only one type of listener when both types are registered.
But I think what you said makes sense. There is very little a container can do other than hoping the clients be good citizens.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
rahulrane50: Oops i missed that part!

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
qqu0127: For 1, yes, it's possible, we don't block that
2, for the same key, there is a set union operation, so it's only triggered once.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
```diff
@@ -0,0 +1,161 @@
...
+   * @param consumer the consuming action for the listeners
+   */
+  public void consumeListeners(String key, Consumer<? super T> consumer) {
+    Set<T> onetimeListeners;
+    Set<T> persistentListeners;
+    synchronized (this) {
+      // remove one-time listeners
+      onetimeListeners = _onetimeListener.remove(key);
+    }
+    if (onetimeListeners == null) {
```
mgao0, line 106: I feel that from line 106 to line 114 can be simplified, such that only if it's non-null then wen add it to the result set, and there's no need of instantiating empty set when there's no listener registered.

mgao0 (commented):
One general comment is that using T for listener types might be too wildcard, and `listener(consumer)` this signature to consume the listener is a bit unintuitive.  Is it possible that we define a generic interface for listener, and define some APIs such as `fireEvent` or `consume`.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ListenerContainer.java
qqu0127: Thanks for the review. 
To my understanding, the reference to `Collections.emptySet()` won't instantiate any new instance, so there is no overhead. I think this is also good for readability and unify the control logic with non-null values. 
Let me know if you have other thoughts on this.

qqu0127:
@mgao0 Thanks for the comments. `using T for listener types might be too wildcard` this is a great point. However I don't see an applicable way to abstract a unified interface out of all possible listener consumption. At best we could apply marker interfaces, but that involves global changes to existing ones, so I didn't follow that.

desaikomal:
@qqu0127 - is this PR still relevant - if not, can  we please close it? thanks

PR #2343, opened 2023-01-18
ZkMetaclient - implementation of TransactionOp

Marcosrico:
### Issues

#2237 

### Description

1. Implementation of the transactionOp() operation method in zkMetaClient. This method enables transactional support by converting metaclient operations into zk ops and running the zk.multi() method. The results are converted back from zk opResults to metaclient opResults.
2. Created a separate util class zkMetaClientUtil for Util functions and the logic behind the implementation of transactionOp().

### Tests

`testMultiOps` and `testMultiFail` in `TestZkMetaClient`.

```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 75.077 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 9, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] --- jacoco-maven-plugin:0.8.6:report (generate-code-coverage-report) @ meta-client ---
[INFO] Loading execution data file /Users/mapeng/Metaclient/helix/meta-client/target/jacoco.exec
[INFO] Analyzed bundle 'Apache Helix :: Meta Client' with 36 classes
[INFO] BUILD SUCCESS
[INFO] Total time:  01:18 min
[INFO] Finished at: 2023-01-20T10:45:57-05:00

```

### Code Quality


meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -414,4 +414,130 @@ private static EntryMode convertZkEntryMode(long ephemeralOwner) {
...
+    return zkOpResultToMetaClient(zkResult);
+  }
+
+  /**
+   * Helper function for transactionOp. Converts MetaClient Op's into Zk Ops to execute
+   * zk transactional support.
+   * @param ops
+   * @return
+   */
+  public List<org.apache.zookeeper.Op> metaClientOpToZk(Iterable<Op> ops) throws KeeperException {
```
desaikomal, line 435: you changed exception from KeeperException to IllegalArgument.

so is this still required?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -414,4 +414,130 @@ private static EntryMode convertZkEntryMode(long ephemeralOwner) {
...
+    return zkOps;
+  }
+
+  /**
+   * Helper function for transactionOP. Converts the result from calling zk transactional support into
+   * metaclient OpResults.
+   * @param zkResult
+   * @return
+   */
+  public List<OpResult> zkOpResultToMetaClient(List<org.apache.zookeeper.OpResult> zkResult) throws KeeperException.BadArgumentsException {
```
desaikomal, line 471: ditto.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -414,4 +414,130 @@ private static EntryMode convertZkEntryMode(long ephemeralOwner) {
...
+   * @param ops
+   * @return
+   */
+  public List<org.apache.zookeeper.Op> metaClientOpToZk(Iterable<Op> ops) throws KeeperException {
+    List<org.apache.zookeeper.Op> zkOps = new ArrayList<>();
+    org.apache.zookeeper.Op temp;
+    for (Op op : ops) {
+      switch (op.getType()) {
+        case CREATE:
+          int zkFlag = zkFlagFromEntryMode(((Op.Create) op).getEntryMode());
```
desaikomal, line 441: this seems like double time evaluation. You want to pass CreateMode type. Can you not have a simple enum which you can pass directly? You are anyway, hardcoding to Op.Create, so why not pass that enum directly. We don't want to create code complexity if passing something directly saves CPU cycles.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
xyuanlu: Please define metaclient exception under org.apache.helix.metaclient.constants.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -414,4 +414,130 @@ private static EntryMode convertZkEntryMode(long ephemeralOwner) {
...
+   */
+  public List<OpResult> zkOpResultToMetaClient(List<org.apache.zookeeper.OpResult> zkResult) throws KeeperException.BadArgumentsException {
+    List<OpResult> metaClientOpResult = new ArrayList<>();
+    OpResult temp;
+    EntryMode zkMode;
+    for (org.apache.zookeeper.OpResult opResult : zkResult) {
+      Stat metaClientStat;
+      switch (opResult.getType()) {
+        // CreateResult
+        case 1:
```
xyuanlu, line 479: Please try to avoid magic number.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
Marcosrico: Nope not required, thanks for the catch!

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
Marcosrico: Sounds good, will create a specific MetaClientKeeperException for these cases (where KeeperExceptions are thrown)

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
Marcosrico: Sounds good, I'll use the class name instead of opResult type.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -414,4 +415,128 @@ private static EntryMode convertZkEntryMode(long ephemeralOwner) {
...
+   */
+  public List<OpResult> zkOpResultToMetaClient(List<org.apache.zookeeper.OpResult> zkResult) {
+    List<OpResult> metaClientOpResult = new ArrayList<>();
+    OpResult temp;
+    String resultClass;
+    for (org.apache.zookeeper.OpResult opResult : zkResult) {
+      Stat metaClientStat;
+      resultClass = opResult.getClass().getSimpleName();
+      switch (resultClass) {
+        case "CreateResult":
```
qqu0127, line 484: Comparing on classname is too risky and low on performance (it's still magic number to some extent). If you must do it this way, please check `instance of`.

qqu0127: I see the difficulties here in binding two classes without knowing the type in compile time. One potential way to improve similar code, is to try using a static map to store the binding and utilize lambda functional.
e.g.
```
// mapping from class object => construction function of OpResult
Map<Class<? extends org.apache.zookeeper.OpResult>, Function<org.apache.zookeeper.OpResult, OpResult>> map = new HashMap<>();
map.put(org.apache.zookeeper.OpResult.CheckResult.class, op -> new OpResult.CheckResult()); 
for (org.apache.zookeeper.OpResult opResult : zkResult) {
  map.get(opResult.getClass()).apply(opResult);
}

```

qqu0127 (commented):
Some "translation" code between ZK and metaclient classes need some major changes. Let me know if you have any question.

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientKeeperException.java
```diff
@@ -0,0 +1,39 @@
+package org.apache.helix.metaclient.constants;
```
desaikomal, line 1: you should have a separate directory for exception. 

eg:
ls <HELIX-home>/metadata-store-directory-common/src/main/java/org/apache/helix/msdcommon/exception
InvalidRoutingDataException.java

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -50,11 +51,15 @@
...
 import org.apache.zookeeper.server.EphemeralType;
+import org.apache.zookeeper.ZooDefs;
+import org.apache.zookeeper.data.ACL;
 
 
 public class ZkMetaClient<T> implements MetaClientInterface<T>, AutoCloseable {
 
   private final ZkClient _zkClient;
+  //Default ACL value until metaClient Op has ACL of its own.
+  private final List<ACL> DEFAULT_ACL = ZooDefs.Ids.OPEN_ACL_UNSAFE;
```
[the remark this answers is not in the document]
Marcosrico: Correct, I'll change it. Thanks!

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -414,4 +415,128 @@ private static EntryMode convertZkEntryMode(long ephemeralOwner) {
...
+    return zkOps;
+  }
+
+  /**
+   * Helper function for transactionOP. Converts the result from calling zk transactional support into
+   * metaclient OpResults.
+   * @param zkResult
+   * @return
+   */
+  public List<OpResult> zkOpResultToMetaClient(List<org.apache.zookeeper.OpResult> zkResult) {
```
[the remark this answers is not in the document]
Marcosrico: Are you saying a separate class in the same directory with these two helper functions?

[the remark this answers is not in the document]
qqu0127: Yes, with a util class, you can also create a util package under impl/zk

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientKeeperException.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+public final class MetaClientKeeperException extends MetaClientException {
```
xyuanlu, line 22: I think we probably want more specific exception than just keeper exception. 
Helix defined  ZkException translate KeeperException to ZkException, and metaclient translates ZkException to metaclient defined exception. 
you can do 
```
   try {
     //........ 
    } catch (KeeperException e) {
      throw translateZkExceptionToMetaclientException(ZkException.create(e));
    }
```

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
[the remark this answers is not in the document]
Marcosrico: `ZooDefs.Ids.OPEN_ACL_UNSAFE` is an arraylist. To make it immutable I can call `Collections.unmodifiableList()` though that seems a bit convoluted. Would making it static and final not be sufficient?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -414,4 +414,130 @@ private static EntryMode convertZkEntryMode(long ephemeralOwner) {
...
+  private int zkFlagFromEntryMode (EntryMode entryMode) {
+    String mode = entryMode.name();
+    if (mode.equals(EntryMode.PERSISTENT.name())) {
+      return 0;
+    }
+    if (mode.equals(EntryMode.EPHEMERAL.name())) {
+      return 1;
+    }
+    return -1;
+  }
```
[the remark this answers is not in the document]
Marcosrico: Okay I'll modify the helper to translate entryMode directly to CreateMode. Thanks for the feedback!

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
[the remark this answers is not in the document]
Marcosrico: Realizing that it would still be modifiable in this case. Calling Collections.unmodifiableList(DEFAULT_ACL) is the best course of action

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientKeeperException.java
Marcosrico: In this case, it wouldn't be necessary to create a separate MetaClientKeeperException then? A ZkException then throws a generic MetaClient exception

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,156 @@
...
+
+    List<OpResult> metaClientOpResult = new ArrayList<>();
+    for (org.apache.zookeeper.OpResult opResult : zkResult) {
+      metaClientOpResult.add(map.get(opResult.getClass()).apply(opResult));
+    }
+
+    return metaClientOpResult;
+  }
+
+  public static MetaClientInterface.EntryMode convertZkEntryMode(long ephemeralOwner) {
```
Marcosrico, line 126: This method is refactored from ZkMetaClient

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,156 @@
...
+        return MetaClientInterface.EntryMode.EPHEMERAL;
+      // TODO: TTL is not supported now.
+      //case TTL:
+      //  return EntryMode.TTL;
+      default:
+        throw new IllegalArgumentException(zkEphemeralType + " is not supported.");
+    }
+  }
+
+  public static MetaClientException translateZkExceptionToMetaclientException(ZkException e) {
```
Marcosrico, line 143: Same as above

meta-client/src/main/java/org/apache/helix/metaclient/constants/MetaClientKeeperException.java
Marcosrico: As discussed asynchronously, I will create a separate PR with these changes after these are merged

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  /**
+   * Helper function for transactionOp. Converts MetaClient Op's into Zk Ops to execute
+   * zk transactional support.
+   * @param ops
+   * @return
+   */
+  public static List<Op> metaClientOpToZk(Iterable<org.apache.helix.metaclient.api.Op> ops) {
+    Map<org.apache.helix.metaclient.api.Op.Type, Function<org.apache.helix.metaclient.api.Op, Op>> opMap = new HashMap<>();
```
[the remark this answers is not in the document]
qqu0127: For this particular case, since the key is enum, please use EnumMap for better performance.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,156 @@
...
+        ((org.apache.zookeeper.OpResult.GetChildrenResult) opResult).getChildren()));
+
+    map.put(org.apache.zookeeper.OpResult.CheckResult.class, opResult -> new OpResult.CheckResult());
+
+    map.put(org.apache.zookeeper.OpResult.ErrorResult.class, opResult -> new OpResult.ErrorResult(
+        ((org.apache.zookeeper.OpResult.ErrorResult) opResult).getErr()));
+
+    List<OpResult> metaClientOpResult = new ArrayList<>();
+    for (org.apache.zookeeper.OpResult opResult : zkResult) {
+      metaClientOpResult.add(map.get(opResult.getClass()).apply(opResult));
```
qqu0127, line 120: let's do a null check after map.get() to be safer

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  private static CreateMode convertMetaClientMode(MetaClientInterface.EntryMode entryMode) throws KeeperException {
+    String mode = entryMode.name();
+    if (mode.equals(MetaClientInterface.EntryMode.PERSISTENT.name())) {
+      return CreateMode.fromFlag(0);
+    }
+    if (mode.equals(MetaClientInterface.EntryMode.EPHEMERAL.name())) {
+      return CreateMode.fromFlag(1);
+    }
```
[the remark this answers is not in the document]
Marcosrico: Thanks for the review, will fix the comparison.

CreateMode.fromFlag() throws KeeperException which is why this method throws the exception. I catch it in metaClientOpToZk that calls this method.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,156 @@
...
+    return CreateMode.fromFlag(-1);
```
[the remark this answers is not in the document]
Marcosrico: Yes that's correct and sounds good

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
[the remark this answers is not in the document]
Marcosrico: Sounds good thank you for the review. I updated the methods with a lazy and thread safe implementation of the maps, please let me know if it looks good, thanks.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -391,34 +380,13 @@ public int hashCode() {
...
-      case NORMAL:
-        return EntryMode.EPHEMERAL;
-      // TODO: TTL is not supported now.
-      //case TTL:
-      //  return EntryMode.TTL;
-      default:
-        throw new IllegalArgumentException(zkEphemeralType + " is not supported.");
-    }
+  @Override
+  public List<OpResult> transactionOP(Iterable<Op> iterable) {
```
rahulrane50, line 384: neat: iterable looks bad variable name. Do you want to change it something like iterableOps or just ops or anything which tells that these are ops list.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,201 @@
...
+  private ZkMetaClientUtil(){
+  }
+
+  /**
+   * Helper function for transactionOp. Converts MetaClient Op's into Zk Ops to execute
+   * zk transactional support.
+   * @param ops
+   * @return
+   */
+  public static List<Op> metaClientOpToZk(Iterable<org.apache.helix.metaclient.api.Op> ops) {
```
rahulrane50, line 54: neat: If it's gonna return list of ops then should the name of the method be metaClientOpsToZkOps? or anything which suggest plural.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,201 @@
...
+    throw new IllegalArgumentException(entryMode.name() + " is not a supported EntryMode.");
+  }
+
+  /**
+   * Helper function for transactionOP. Converts the result from calling zk transactional support into
+   * metaclient OpResults.
+   * @param zkResult
+   * @return
+   */
+  public static List<OpResult> zkOpResultToMetaClient(List<org.apache.zookeeper.OpResult> zkResult) {
```
rahulrane50, line 114: neat: similar comment about plural.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,156 @@
...
+import org.apache.zookeeper.data.ACL;
+import org.apache.zookeeper.server.EphemeralType;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.ArrayList;
+import java.util.function.Function;
+
```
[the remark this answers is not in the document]
rahulrane50: Okey i am seeing this pattern a lot now let me file a issue to automate this.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  /**
+   * Helper function for transactionOp. Converts MetaClient Op's into Zk Ops to execute
+   * zk transactional support.
+   * @param ops
+   * @return
+   */
+  public static List<Op> metaClientOpToZk(Iterable<org.apache.helix.metaclient.api.Op> ops) {
+    getOpMap().put(org.apache.helix.metaclient.api.Op.Type.CREATE, op -> {
```
rahulrane50, line 55: Can we avoid having full name of org.apache.helix.metaclient.api.Op.Type here? i see that at multiple places down below.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.Collections;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.function.Function;
+
+public class ZkMetaClientUtil {
+  //Default ACL value until metaClient Op has ACL of its own.
```
rahulrane50, line 42: neat: let's add a TODO here as well. We might need to think about ACLs being set by helix.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,201 @@
...
+      if (function != null) {
+        zkOps.add(function.apply(op));
+      } else {
+        throw new IllegalArgumentException("Op type " + op.getType().name() + "is not supported.");
+      }
+    }
+    return zkOps;
+  }
+
+  private static final class OpmapHolder {
```
rahulrane50, line 86: neat: OpMapHolder.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,201 @@
...
+    static final Map<org.apache.helix.metaclient.api.Op.Type, Function<org.apache.helix.metaclient.api.Op, Op>> OPMAP =
+        new EnumMap<>(org.apache.helix.metaclient.api.Op.Type.class);
+  }
+
+  private static Map<org.apache.helix.metaclient.api.Op.Type, Function<org.apache.helix.metaclient.api.Op, Op>> getOpMap() {
+    return OpmapHolder.OPMAP;
+  }
+
+  private static CreateMode convertMetaClientMode(MetaClientInterface.EntryMode entryMode) throws KeeperException {
+    if (entryMode == MetaClientInterface.EntryMode.PERSISTENT) {
```
rahulrane50, line 96: just wondering if making it to switch case would make it cleaner?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  /**
+   * Helper function for transactionOP. Converts the result from calling zk transactional support into
+   * metaclient OpResults.
+   * @param zkResult
+   * @return
+   */
+  public static List<OpResult> zkOpResultToMetaClient(List<org.apache.zookeeper.OpResult> zkResult) {
+    getOpResultMap().put(org.apache.zookeeper.OpResult.CreateResult.class, opResult -> {
```
qqu0127, line 115: This doesn't look like lazy loading. Each time the method invoked, we still go through all map.put() process.

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestZkMetaClient.java
```diff
@@ -39,8 +39,11 @@
 import org.apache.commons.lang3.NotImplementedException;
 import org.apache.helix.metaclient.api.DataChangeListener;
 import org.apache.helix.metaclient.impl.zk.factory.ZkMetaClientConfig;
+import org.apache.helix.metaclient.api.Op;
+import org.apache.helix.metaclient.api.OpResult;
 import org.apache.helix.zookeeper.zkclient.IDefaultNameSpace;
 import org.apache.helix.zookeeper.zkclient.ZkServer;
+import org.apache.zookeeper.*;
```
rahulrane50, line 46: this would be huge import! Do we want everything from this package?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,201 @@
...
+        new EnumMap<>(org.apache.helix.metaclient.api.Op.Type.class);
+  }
+
+  private static Map<org.apache.helix.metaclient.api.Op.Type, Function<org.apache.helix.metaclient.api.Op, Op>> getOpMap() {
+    return OpmapHolder.OPMAP;
+  }
+
+  private static CreateMode convertMetaClientMode(MetaClientInterface.EntryMode entryMode) throws KeeperException {
+    if (entryMode == MetaClientInterface.EntryMode.PERSISTENT) {
+      return CreateMode.fromFlag(0);
```
rahulrane50, line 97: Should we use something verbose or informative than using hard coded values like 0, 1 or 4?

rahulrane50 (commented):
Few minor neats but otherwise functionality wise looks good to me!

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
Marcosrico: Good catch. The metaclient interface has it as ops as well so it only makes sense to be consistent with the naming too.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
Marcosrico: Correct, that would make more sense!

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
Marcosrico: Unfortunately because both package names (metaclient and zookeeper) have the same name (op and opResult) they must be specified with their full name.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
Marcosrico: Sounds good!

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
qqu0127: +1 to Rahul's comment, I think we can just clearly type to return CreateMode.PERSISTENT or CreateMode.EPHEMERAL etc.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
Marcosrico: Yup it definitely would, thanks for the suggestion!

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestZkMetaClient.java
Marcosrico: Absolutely not! Must have auto corrected after many imports, nice catch!

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
Marcosrico: Yes that makes more sense as the flag parameter provides no value. Thanks for the suggestions!

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,213 @@
...
+      }
+    }
+
+    return metaClientOpResult;
+  }
+
+  private static final class OpResultMapHolder {
+    static final Map<Class<? extends org.apache.zookeeper.OpResult>, Function<org.apache.zookeeper.OpResult, OpResult>> OPRESULTMAP =
+        new HashMap<>();
+  }
```
qqu0127, line 141: Thanks for the work Marcos. 
I see the current lazy loading functionally works, but it's kind of redundant with this holder. You might miss the point of using a on-demand holder. I believe we can move the init() method inline. Please refer to JVM classloading and on-demand holder idiom.

Marcosrico: I removed the unnecessary null check before each "op translate" as initializing the map guarantees non null value of opMap and opResultMap. Let me know if that addresses the comment, thanks!

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+    return zkOps;
+  }
+
+  private static final class OpMapHolder {
+    static final Map<org.apache.helix.metaclient.api.Op.Type, Function<org.apache.helix.metaclient.api.Op, Op>> OPMAP =
+        new EnumMap<>(org.apache.helix.metaclient.api.Op.Type.class);
+  }
+
+  private static Map<org.apache.helix.metaclient.api.Op.Type, Function<org.apache.helix.metaclient.api.Op, Op>> getOpMap() {
```
mgao0, line 74: Curious, why do we need this method? Can we directly refer to OPMAP?

mgao0: Thanks for the reply. Now I understand that you are trying to do lazy initialization here. But let's say you use `OpMapHolder.OPMAP` instead of `getOpMap()`, it should work for the lazy instantiation as well? I'm also not sure if it's necessary to create a nested class in order to do lazy instantiation if the map itself is not something that's too costly.

Marcosrico: Very true, calling OPMAP directly would still work for lazy instantiation. However, the get method provides a layer of abstraction in case the implementation of OPMAP changes, which in my opinion outweighs the minor overhead cost of calling the get method.

And you bring up a good point, I was outweighing the costs of implementing lazy instantiation to begin with and having the map be eagerly instantiated in the beginning of ZkMetaClientUtil class. Considering that this class is a util class for anything zkmetaclient related, it can be used in the future for other actions not related to transactional support. While there isn't much else going on, by providing lazy instantiation we can minimize overhead in the long run if ZkMetaClientUtil has increased use cases and/or each map gets larger.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/util/ZkMetaClientUtil.java
qqu0127: Has this been addressed? Why initializeOpResultMap is not invoked for OPRESULTMAP?

Marcosrico: Correct me if I am wrong, but I believe what you're saying is to have the map be populated when the static variable `OPRESULTMAP` is declared? I have attempted several methods and none seemed feasible:
1. Create a constructor that invokes `initializeOpResultMap` method when `OpResultMapHolder` object is created. However, given the nature of it being a util class, making it an object is counterproductive.
2. Populating the map directly when declaring the `OPRESULTMAP` static variable instead of having a separate method. This can only be done with java JDK 9+ (current library is at 8) so decided against it.

With the current implementation, `initializeOpResultMap` is called just once when needed in the `zkOpResultToMetaClientOpResults` method. Inside, `getOpResultMap` is called first which initializes the `OPRESULTMAP` for the first time, effectively achieving lazy loading. Though ideally when initializing `OPRESULTMAP` for the first time, we would like it to automatically be populated with its values, as mentioned above, I couldn't find a reasonable way to do so. Please suggest a solution that I may have missed.

qqu0127: This is incorrect. In your impl, `initializeOpResultMap` is called every time `zkOpResultToMetaClientOpResults` is called, which trace back to `ZkMetaClient.transactionOP`. 
I don't get why point 2 is not applicable. To achieve lazy loading, you can combine the "initialize" and "value population" in your static method `initializeOpResultMap` by changing the signature and content to build and return the map, so you can call it directly at declaration.

Marcosrico: Thank you for your correction, I understand what you mean. I was implementing option 2 wrong which was why the error occurred. Thanks for the feedback and I have implemented those changes, let me know if it's good!

qqu0127 (approved):
LGTM, please address other comments too

Marcosrico:
This PR is ready to merge, approved by @qqu0127 
Commit message:
ZkMetaClient - Implementation of TransactionOp

PR #2344, opened 2023-01-19
Added new metric to report real time missing top state for partition

### LINKED ISSUE #2345: Add new metric to report missing top state for partition
Issue :
Currently, if waged rebalancer is used then there could be a situation where multiple leader replicas are residing on same node. If that node goes down for maintainence or for any other reason then there is no current metric which reports about missing top state until any rebalancer event is triggered.

Ask :
Ask here is to add a metric which satisfies following conditions :

The metric should be at resource level and should be reported irrespective of top state replica is recovered or not for any partition of that resource.
This metric should be started as soon as "any" partition for a resource has top state missing and it should be "reset" only after sliding window reset time i.e., 1 minute.

**Additional context**

rahulrane50:
### Issues

Fixes #2345 : Adds a new metric to report real time missing top state for any partition for each resource in a cluster.

### Description
Issue :
Currently, if waged rebalancer is used then there could be a situation where multiple leader replicas are residing on same node. If that node goes down for maintainence or for any other reason then there is no current metric which reports about missing top state until any rebalancer event is triggered. 

Ask :
Ask here is to add a metric which satisfies following conditions :
1. The metric should be at resource level and should be reported irrespective of top state replica is recovered or not for any partition of that resource.
2. This metric should be started as soon as "any" partition for a resource has top state missing and it should be "reset" only when that resource has no partition with missing top state.

Solution :
The solution proposed in this PR starts a async thread as soon as first partition of any resource has top state missing and it continuously reports missing top state duration (atleast once in sliding window interval) for partitions at resource level as long as that resource has "at least one" partition with missing top state. If there are no resources with any partitions with missing top state then this async thread will sleep.
The guage reported by this thread is histogram which won't give exact values but will approximate about duration increasing and will be reset when all top state replicas for that resource are recovered. 

### Tests

- [X ] The following tests are written for this issue:

CI pipeline and added new tests for verifying metrics.

```
mvn test -Dtest=TestTopStateHandoffMetrics
[WARNING] Tests run: 12, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 18.723 s - in org.apache.helix.monitoring.mbeans.TestTopStateHandoffMetrics
[INFO]
[INFO] Results:
[INFO]
[WARNING] Tests run: 12, Failures: 0, Errors: 0, Skipped: 1

```

CI is failing with helix-rest tests. Verified locally that tests are running
Failing test results : 
```
2023-01-19T10:12:27.2107893Z [info] ./metadata-store-directory-common/target/surefire-reports/TestSuite.txt: Tests run: 31, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.633 s - in TestSuite
2023-01-19T10:12:27.2108983Z [info] ./helix-core/target/surefire-reports/TestSuite.txt: Tests run: 1322, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 5,906.49 s - in TestSuite
2023-01-19T10:12:27.2110115Z [info] ./helix-view-aggregator/target/surefire-reports/TestSuite.txt: Tests run: 15, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 100.176 s - in TestSuite
2023-01-19T10:12:27.2110941Z [info] ./recipes/distributed-lock-manager/target/surefire-reports/TestSuite.txt: Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.492 s - in TestSuite
2023-01-19T10:12:27.2111791Z [info] ./recipes/rabbitmq-consumer-group/target/surefire-reports/TestSuite.txt: Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.692 s - in TestSuite
2023-01-19T10:12:27.2112591Z [info] ./recipes/task-execution/target/surefire-reports/TestSuite.txt: Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.534 s - in TestSuite
2023-01-19T10:12:27.2113368Z [info] ./recipes/service-discovery/target/surefire-reports/TestSuite.txt: Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.63 s - in TestSuite
2023-01-19T10:12:27.2114188Z [info] ./recipes/rsync-replicated-file-system/target/surefire-reports/TestSuite.txt: Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.582 s - in TestSuite
2023-01-19T10:12:27.2114980Z [info] ./helix-lock/target/surefire-reports/TestSuite.txt: Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 55.692 s - in TestSuite
2023-01-19T10:12:27.2115716Z [info] ./metrics-common/target/surefire-reports/TestSuite.txt: Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.372 s - in TestSuite
2023-01-19T10:12:27.2116463Z [info] ./zookeeper-api/target/surefire-reports/TestSuite.txt: Tests run: 82, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 152.431 s - in TestSuite
2023-01-19T10:12:27.2117226Z [info] ./helix-rest/target/surefire-reports/TestSuite.txt: Tests run: 209, Failures: 1, Errors: 0, Skipped: 18, Time elapsed: 151.752 s <<< FAILURE! - in TestSuite
2023-01-19T10:12:27.2144940Z ##[error] Test failed: testActivateSuperCluster(org.apache.helix.rest.server.TestClusterAccessor)  Time elapsed: 8.741 s  <<< FAILURE!
2023-01-19T10:12:27.2146629Z [info] ./helix-common/target/surefire-reports/TestSuite.txt: Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.264 s - in TestSuite

```

Verified locally :
```
[INFO] Tests run: 39, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 208.189 s - in org.apache.helix.rest.server.TestClusterAccessor
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 39, Failures: 0, Errors: 0, Skipped: 0
[INFO]
```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
@@ -85,6 +85,9 @@ public class TopStateHandoffReportStage extends AbstractBaseStage {
     if (cache.getClusterConfig() != null) {
       durationThreshold = cache.getClusterConfig().getMissTopStateDurationThreshold();
     }
+    if (clusterStatusMonitor != null) {
+      clusterStatusMonitor.updateMissingTopStateDurationThreshold(durationThreshold);
+    }
 
     // Remove any resource records that no longer exists
     missingTopStateMap.keySet().retainAll(resourceMap.keySet());
@@ -206,7 +209,7 @@ public class TopStateHandoffReportStage extends AbstractBaseStage {
 
     if (missingTopStateMap.containsKey(resourceName) && missingTopStateMap.get(resourceName)
         .containsKey(partition.getPartitionName())) {
-      // We previously recorded a top state missing, and it's not coming back
+      // We previously recorded a top state missing, and it has not come back
       reportTopStateComesBack(cache, currentStateOutput.getCurrentStateMap(resourceName, partition),
           resourceName, partition, clusterStatusMonitor, durationThreshold,
           stateModelDef.getTopState());
@@ -319,12 +322,17 @@ public class TopStateHandoffReportStage extends AbstractBaseStage {
     String partitionName = partition.getPartitionName();
     MissingTopStateRecord record = missingTopStateMap.get(resourceName).get(partitionName);
     long startTime = record.getStartTimeStamp();
-    if (startTime > 0 && System.currentTimeMillis() - startTime > durationThreshold && !record
+    if (startTime > 0 && !record
         .isFailed()) {
-      record.setFailed();
-      missingTopStateMap.get(resourceName).put(partitionName, record);
       if (clusterStatusMonitor != null) {
-        clusterStatusMonitor.updateMissingTopStateDurationStats(resourceName, 0L, 0L, false, false);
+        clusterStatusMonitor.updateMissingTopStateResourceMap(resourceName, partitionName, true, startTime);
+      }
+      if (System.currentTimeMillis() - startTime > durationThreshold) {
+        record.setFailed();
+        missingTopStateMap.get(resourceName).put(partitionName, record);
+        if (clusterStatusMonitor != null) {
+          clusterStatusMonitor.updateMissingTopStateDurationStats(resourceName, 0L, 0L, false, false);
+        }
       }
     }
   }
@@ -488,6 +496,10 @@ public class TopStateHandoffReportStage extends AbstractBaseStage {
       }
     }
     removeFromStatsMap(missingTopStateMap, resourceName, partition);
+    if (clusterStatusMonitor != null) {
+      // Update recovered top state for partition in resource map.
+      clusterStatusMonitor.updateMissingTopStateResourceMap(resourceName, partition.getPartitionName(), false, 0L);
+    }
   }
 
   private void removeFromStatsMap(
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
@@ -322,8 +322,7 @@ public class TopStateHandoffReportStage extends AbstractBaseStage {
     String partitionName = partition.getPartitionName();
     MissingTopStateRecord record = missingTopStateMap.get(resourceName).get(partitionName);
     long startTime = record.getStartTimeStamp();
-    if (startTime > 0 && !record
-        .isFailed()) {
+    if (startTime > 0 && !record.isFailed()) {
       if (clusterStatusMonitor != null) {
         clusterStatusMonitor.updateMissingTopStateResourceMap(resourceName, partitionName, true, startTime);
       }
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
@@ -83,25 +83,30 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
               ResourceMonitor resourceMonitor = getOrCreateResourceMonitor(resourcePartitionEntry.getKey());
               // If all partitions of resource has top state recovered then reset the counter
               if (resourcePartitionEntry.getValue().isEmpty()) {
-                resourceMonitor.resetOneOrManyPartitionsMissingTopStateRealTimeGuage();
+                resourceMonitor.resetMissingTopStateDurationGuage();
                 resourcePartitionIt.remove();
               } else {
                 for (Long missingTopStateStartTime : resourcePartitionEntry.getValue().values()) {
                   if (_missingTopStateDurationThreshold < Long.MAX_VALUE && System.currentTimeMillis() - missingTopStateStartTime > _missingTopStateDurationThreshold) {
-                    resourceMonitor.updateOneOrManyPartitionsMissingTopStateRealTimeGuage();
+                    resourceMonitor.updateMissingTopStateDurationGuage(System.currentTimeMillis() - missingTopStateStartTime);
                   }
                 }
 
               }
             }
-            // TODO: Check if this SLEEP_TIME is correct? Thread should keep on increasing the counter continuously until top
-            //  state is recovered but it can sleep for reasonable amount of time in between.
-            sleep(100);
           }
         }
       } catch (InterruptedException e) {
         LOG.error("AsyncMissingTopStateMonitor has been interrupted.", e);
       }
+    }
+
+    public void reset() {
+      for (String resource : _missingTopStateResourceMap.keySet()) {
+        ResourceMonitor resourceMonitor = getOrCreateResourceMonitor(resource);
+        resourceMonitor.resetMissingTopStateDurationGuage();
+      }
+      _missingTopStateResourceMap.clear();
     }
   };
   private static final Logger LOG = LoggerFactory.getLogger(ClusterStatusMonitor.class);
@@ -167,12 +172,6 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
   public ClusterStatusMonitor(String clusterName) {
     _clusterName = clusterName;
     _beanServer = ManagementFactory.getPlatformMBeanServer();
-    /**
-     * Start a async thread for each cluster which keeps monitoring missing top states of any resource for a cluster.
-     * This thread will keep on iterating over resources and report missingTopStateDuration for them until there are at
-     * least one resource with missing top state for a cluster.
-     */
-    _asyncMissingTopStateMonitor.start();
   }
 
   public ObjectName getObjectName(String name) throws MalformedObjectNameException {
@@ -710,6 +709,12 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
     LOG.info("Active ClusterStatusMonitor");
     try {
       register(this, getObjectName(clusterBeanName()));
+      /**
+       * Start a async thread for each cluster which keeps monitoring missing top states of any resource for a cluster.
+       * This thread will keep on iterating over resources and report missingTopStateDuration for them until there are at
+       * least one resource with missing top state for a cluster.
+       */
+      _asyncMissingTopStateMonitor.start();
     } catch (Exception e) {
       LOG.error("Fail to register ClusterStatusMonitor", e);
     }
@@ -738,8 +743,9 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
       _rebalanceFailureCount.set(0L);
       _continuousResourceRebalanceFailureCount.set(0L);
       _continuousTaskRebalanceFailureCount.set(0L);
+      // No need to wait until this async thread finishes because interrupting it means stop reporting metric.
       _asyncMissingTopStateMonitor.interrupt();
-      _missingTopStateResourceMap.clear();
+      _asyncMissingTopStateMonitor.reset();
     } catch (Exception e) {
       LOG.error("Fail to reset ClusterStatusMonitor, cluster: " + _clusterName, e);
     }
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
@@ -744,8 +744,8 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
       _continuousResourceRebalanceFailureCount.set(0L);
       _continuousTaskRebalanceFailureCount.set(0L);
       // No need to wait until this async thread finishes because interrupting it means stop reporting metric.
-      _asyncMissingTopStateMonitor.interrupt();
       _asyncMissingTopStateMonitor.reset();
+      _asyncMissingTopStateMonitor.interrupt();
     } catch (Exception e) {
       LOG.error("Fail to reset ClusterStatusMonitor, cluster: " + _clusterName, e);
     }
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
@@ -99,7 +99,7 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
           }
         }
       } catch (InterruptedException e) {
-        LOG.error("AsyncMissingTopStateMonitor has been interrupted {}", e);
+        LOG.error("AsyncMissingTopStateMonitor has been interrupted.", e);
       }
     }
   };
@@ -739,6 +739,8 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
       _rebalanceFailureCount.set(0L);
       _continuousResourceRebalanceFailureCount.set(0L);
       _continuousTaskRebalanceFailureCount.set(0L);
+      _asyncMissingTopStateMonitor.interrupt();
+      _missingTopStateResourceMap.clear();
     } catch (Exception e) {
       LOG.error("Fail to reset ClusterStatusMonitor, cluster: " + _clusterName, e);
     }
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
@@ -94,6 +94,7 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
 
               }
             }
+            sleep(50); // Instead of providing stream of durtion values to histogram thread can sleep in between to save some CPU cycles.
           }
         }
       } catch (InterruptedException e) {
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
@@ -57,7 +57,7 @@ import org.slf4j.LoggerFactory;
 public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
   private class AsyncMissingTopStateMonitor extends Thread {
     private final ConcurrentHashMap<String, Map<String, Long>> _missingTopStateResourceMap;
-    private long _missingTopStateDurationThreshold = Long.MAX_VALUE;;
+    private long _missingTopStateDurationThreshold = Long.MAX_VALUE;
 
     public AsyncMissingTopStateMonitor(ConcurrentHashMap<String, Map<String, Long>> missingTopStateResourceMap) {
       _missingTopStateResourceMap = missingTopStateResourceMap;
@@ -68,6 +68,18 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
     }
 
     @Override
+    /**
+     * This async metric reporting thread does following things :
+     *    1. Async thread is started when ClusterStatusMonitor is activated and stopped when ClusterStatusMonitor is reset.
+     *    2. Sleeps until any resource has at least one partition with missing top state.
+     *    3. Once waken up by ClusterStatusMonitor, scans over all the resources which are added
+     *       in _missingTopStateResourceMap and for partition of each resource, reports a duration
+     *       (CurrentTime - TopStateMissingStartTime).
+     *       NOTE : Now since, reporting this metric for every partition in each iteration won't make sense
+     *       because monitoring system might not be reading this metrics every ms. Hence we report metric for partition
+     *       only if it has not been reported in sliding window interval (1 min is current sliding window interval) for that
+     *       partition.
+     */
     public void run() {
       try {
         synchronized (this) {
@@ -75,26 +87,34 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
             while (_missingTopStateResourceMap.size() == 0) {
               this.wait();
             }
+            Map<String, Long> partitionToLastReportTimeMap = new HashMap<String, Long>();
             for (Iterator<Map.Entry<String, Map<String, Long>>> resourcePartitionIt =
                 _missingTopStateResourceMap.entrySet().iterator(); resourcePartitionIt.hasNext(); ) {
               Map.Entry<String, Map<String, Long>> resourcePartitionEntry = resourcePartitionIt.next();
-              // Iterate over all partitions and if any partition has missing top state greater than threshold then report
-              // it.
               ResourceMonitor resourceMonitor = getOrCreateResourceMonitor(resourcePartitionEntry.getKey());
+              long resetInterval = resourceMonitor.getSlidingWindowResetIntervalInMs();
               // If all partitions of resource has top state recovered then reset the counter
               if (resourcePartitionEntry.getValue().isEmpty()) {
                 resourceMonitor.resetMissingTopStateDurationGuage();
                 resourcePartitionIt.remove();
               } else {
+                // Iterate over all partitions and if any partition has missing top state greater than threshold then report
+                // it.
                 for (Long missingTopStateStartTime : resourcePartitionEntry.getValue().values()) {
-                  if (_missingTopStateDurationThreshold < Long.MAX_VALUE && System.currentTimeMillis() - missingTopStateStartTime > _missingTopStateDurationThreshold) {
+                  // If metric for this partition has already reported within last sliding window interval then it can be skipped.
+                  if (System.currentTimeMillis() -
+                      partitionToLastReportTimeMap.getOrDefault(resourcePartitionEntry.getKey(), 0L) < resetInterval) {
+                    continue;
+                  }
+                  if (_missingTopStateDurationThreshold < Long.MAX_VALUE &&
+                      System.currentTimeMillis() - missingTopStateStartTime > _missingTopStateDurationThreshold) {
                     resourceMonitor.updateMissingTopStateDurationGuage(System.currentTimeMillis() - missingTopStateStartTime);
+                    // Reset last reported time for the partition
+                    partitionToLastReportTimeMap.put(resourcePartitionEntry.getKey(), System.currentTimeMillis());
                   }
                 }
-
               }
             }
-            sleep(50); // Instead of providing stream of durtion values to histogram thread can sleep in between to save some CPU cycles.
           }
         }
       } catch (InterruptedException e) {
@@ -168,7 +188,8 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
    * Missing top state resource map: resourceName-><PartitionName->startTimeOfMissingTopState>
    */
   private final ConcurrentHashMap<String, Map<String, Long>> _missingTopStateResourceMap = new ConcurrentHashMap<>();
-  private final AsyncMissingTopStateMonitor _asyncMissingTopStateMonitor = new AsyncMissingTopStateMonitor(_missingTopStateResourceMap);
+  private final AsyncMissingTopStateMonitor _asyncMissingTopStateMonitor =
+      new AsyncMissingTopStateMonitor(_missingTopStateResourceMap);
 
   public ClusterStatusMonitor(String clusterName) {
     _clusterName = clusterName;
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
@@ -55,6 +55,54 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
+  private class AsyncMissingTopStateMonitor extends Thread {
+    private final Map<String, Map<String, Long>> _missingTopStateResourceMap;
+    private long _missingTopStateDurationThreshold = Long.MAX_VALUE;;
+
+    public AsyncMissingTopStateMonitor(Map<String, Map<String, Long>> missingTopStateResourceMap) {
+      _missingTopStateResourceMap = missingTopStateResourceMap;
+    }
+
+    public void setMissingTopStateDurationThreshold(long missingTopStateDurationThreshold) {
+      _missingTopStateDurationThreshold = missingTopStateDurationThreshold;
+    }
+
+    @Override
+    public void run() {
+      try {
+        synchronized (this) {
+          while (true) {
+            while (_missingTopStateResourceMap.size() == 0) {
+              this.wait();
+            }
+            for (Iterator<Map.Entry<String, Map<String, Long>>> resourcePartitionIt =
+                _missingTopStateResourceMap.entrySet().iterator(); resourcePartitionIt.hasNext(); ) {
+              Map.Entry<String, Map<String, Long>> resourcePartitionEntry = resourcePartitionIt.next();
+              // Iterate over all partitions and if any partition has missing top state greater than threshold then report
+              // it.
+              ResourceMonitor resourceMonitor = getOrCreateResourceMonitor(resourcePartitionEntry.getKey());
+              // If all partitions of resource has top state recovered then reset the counter
+              if (resourcePartitionEntry.getValue().isEmpty()) {
+                resourceMonitor.resetOneOrManyPartitionsMissingTopStateRealTimeGuage();
+                resourcePartitionIt.remove();
+              } else {
+                for (Long missingTopStateStartTime : resourcePartitionEntry.getValue().values()) {
+                  if (_missingTopStateDurationThreshold < Long.MAX_VALUE && System.currentTimeMillis() - missingTopStateStartTime > _missingTopStateDurationThreshold) {
+                    resourceMonitor.updateOneOrManyPartitionsMissingTopStateRealTimeGuage();
+                  }
+                }
+              }
+            }
+            // TODO: Check if this SLEEP_TIME is correct? Thread should keep on increasing the counter continuously until top
+            //  state is recovered but it can sleep for reasonable amount of time in between.
+            sleep(100);
+          }
+        }
+      } catch (InterruptedException e) {
+        LOG.error("AsyncMissingTopStateMonitor has been interrupted {}", e);
+      }
+    }
+  };
   private static final Logger LOG = LoggerFactory.getLogger(ClusterStatusMonitor.class);
 
   static final String MESSAGE_QUEUE_STATUS_KEY = "MessageQueueStatus";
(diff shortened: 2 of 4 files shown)


desaikomal:
Just a small nit: since we have an ask, you can create an Issue with all the details and then in Issues section, can say, 'Fixes #<issue>',

helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
```diff
@@ -319,12 +322,17 @@ private void reportTopStateHandoffFailIfNecessary(ResourceControllerDataProvider
     String partitionName = partition.getPartitionName();
     MissingTopStateRecord record = missingTopStateMap.get(resourceName).get(partitionName);
     long startTime = record.getStartTimeStamp();
-    if (startTime > 0 && System.currentTimeMillis() - startTime > durationThreshold && !record
+    if (startTime > 0 && !record
         .isFailed()) {
```
xyuanlu, line 326: nit: probably keep them in same line

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
```diff
@@ -335,12 +340,21 @@ private void resetResourceStateGauges() {
     _numOfPartitionsInExternalView.updateValue(0L);
     _numLessMinActiveReplicaPartitions.updateValue(0L);
     _numLessReplicaPartitions.updateValue(0L);
+    _oneOrManyPartitionsMissingTopStateRealTimeGuage.updateValue(0L);
```
junkaixue, line 343: This is very long name. Also it already represents how many partitions missing as aggregation. Better be: _numPartitionsMissingTopStateRealTime

helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
junkaixue: +1. If this is refactoring. Better do it in a different PR without logic change. Would be easy for reviewing.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,54 @@
...
+      _missingTopStateDurationThreshold = missingTopStateDurationThreshold;
+    }
+
+    @Override
+    public void run() {
+      try {
+        synchronized (this) {
+          while (true) {
+            while (_missingTopStateResourceMap.size() == 0) {
+              this.wait();
```
junkaixue, line 76: this wait here and here is the wake up process? Would you like to use conditional variable instead?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,54 @@
...
+                for (Long missingTopStateStartTime : resourcePartitionEntry.getValue().values()) {
+                  if (_missingTopStateDurationThreshold < Long.MAX_VALUE && System.currentTimeMillis() - missingTopStateStartTime > _missingTopStateDurationThreshold) {
+                    resourceMonitor.updateOneOrManyPartitionsMissingTopStateRealTimeGuage();
+                  }
+                }
+              }
+            }
+            // TODO: Check if this SLEEP_TIME is correct? Thread should keep on increasing the counter continuously until top
+            //  state is recovered but it can sleep for reasonable amount of time in between.
+            sleep(100);
```
junkaixue, line 98: As I mentioned, instead of using sleep, maybe it would be better to use conditional variable.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,54 @@
...
+              // Iterate over all partitions and if any partition has missing top state greater than threshold then report
+              // it.
+              ResourceMonitor resourceMonitor = getOrCreateResourceMonitor(resourcePartitionEntry.getKey());
+              // If all partitions of resource has top state recovered then reset the counter
+              if (resourcePartitionEntry.getValue().isEmpty()) {
+                resourceMonitor.resetOneOrManyPartitionsMissingTopStateRealTimeGuage();
+                resourcePartitionIt.remove();
+              } else {
+                for (Long missingTopStateStartTime : resourcePartitionEntry.getValue().values()) {
+                  if (_missingTopStateDurationThreshold < Long.MAX_VALUE && System.currentTimeMillis() - missingTopStateStartTime > _missingTopStateDurationThreshold) {
```
junkaixue, line 90: You can use Java parallel compute lamda feature.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,54 @@
 import org.slf4j.LoggerFactory;
 
 public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
+  private class AsyncMissingTopStateMonitor extends Thread {
+    private final Map<String, Map<String, Long>> _missingTopStateResourceMap;
```
qqu0127, line 59: Does the map have to be a concurrent map? Any risk of concurrent modification?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,54 @@
...
+                _missingTopStateResourceMap.entrySet().iterator(); resourcePartitionIt.hasNext(); ) {
+              Map.Entry<String, Map<String, Long>> resourcePartitionEntry = resourcePartitionIt.next();
+              // Iterate over all partitions and if any partition has missing top state greater than threshold then report
+              // it.
+              ResourceMonitor resourceMonitor = getOrCreateResourceMonitor(resourcePartitionEntry.getKey());
+              // If all partitions of resource has top state recovered then reset the counter
+              if (resourcePartitionEntry.getValue().isEmpty()) {
+                resourceMonitor.resetOneOrManyPartitionsMissingTopStateRealTimeGuage();
+                resourcePartitionIt.remove();
+              } else {
```
qqu0127, line 88: For me I'm not a fan of this iterate and modify pattern. It's not wrong, but any chance we can make it a pure function?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -109,9 +157,21 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
 
   private final Map<String, JobMonitor> _perTypeJobMonitorMap = new ConcurrentHashMap<>();
 
+  /**
+   * Missing top state resource map: resourceName-><PartitionName->startTimeOfMissingTopState>
+   */
+  private final Map<String, Map<String, Long>> _missingTopStateResourceMap = new ConcurrentHashMap<>();
+  private final AsyncMissingTopStateMonitor _asyncMissingTopStateMonitor = new AsyncMissingTopStateMonitor(_missingTopStateResourceMap);
```
qqu0127, line 164: Now I see it's indeed a ConcurrentHashMap. 
One nit, let's make this type explicit in the signature.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -583,6 +643,33 @@ public void updateMissingTopStateDurationStats(String resourceName, long totalDu
...
+    if (isTopStateMissing) {
+      // Wake up asyncMissingTopStateMonitor thread on first resource being added to map
+      if (_missingTopStateResourceMap.isEmpty()) {
+        synchronized (_asyncMissingTopStateMonitor) {
+          _asyncMissingTopStateMonitor.notify();
+        }
+      }
+      if (!_missingTopStateResourceMap.containsKey(resourceName)) {
+        _missingTopStateResourceMap.put(resourceName, new HashMap<String, Long>());
+      }
```
qqu0127, line 661: This can be reduced to computeIfAbsent ?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -583,6 +643,33 @@ public void updateMissingTopStateDurationStats(String resourceName, long totalDu
...
+      _missingTopStateResourceMap.get(resourceName).put(partitionName, startTime);
+    } else { // top state recovered
+      // remove partitions from resourceMap whose top state has been recovered, this will put
+      // asyncMissingTopStateMonitor thread to sleep when no resources left to monitor.
+      Map<String, Long> entry = _missingTopStateResourceMap.get(resourceName);
+      if (entry != null) {
+        entry.remove(partitionName);
+      }
```
qqu0127, line 669: Related to the above comment on iterator. Can we do all the GC-like map cleanup here?

helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
rahulrane50: Sure let me fix indentation. But this is not refactoring and needed for this fix.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
rahulrane50: TFTR @junkaixue! I agree with long name but i just wanted it to be verbose. Actually counter value doesn't indicate number of partitions missing top state (not even at aggregated level) because this number is continuously increasing once started. Any positive value of this counter just indicates that there could be at least one partition (could be more) with missing top states.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -109,9 +157,21 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
 
   private final Map<String, JobMonitor> _perTypeJobMonitorMap = new ConcurrentHashMap<>();
 
+  /**
+   * Missing top state resource map: resourceName-><PartitionName->startTimeOfMissingTopState>
+   */
+  private final Map<String, Map<String, Long>> _missingTopStateResourceMap = new ConcurrentHashMap<>();
```
mgao0, line 163: Can this line and next line can be part of ClusterStatusMonitor constructor? Just define the type of variable here, but do the instantiation in the constructor.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: Just replied on last comment. So this sleep() is not related to main conditional sleeping. This sleep is added because this for loop will continuously increment counters on every iteration which can overflow counter if topstate missing situation continues for long time and also it's not really needed to increment counter for every single ms/ns. As the TODO says async thread should continuously increment counter but with some reasonable(may be avg topstate missing duration/4 ms or something like this) sleep in between.

mgao0 (commented):
Thanks for the PR. I have several general question: 
1. Is there a reason why we need a long-running thread for AsyncMissingTopStateMonitor?  Can we couple the update of ResourceMonitor with the update of updateMissingTopStateResourceMap in ClusterStatusMonitor, instead of checking the map periodically?
2. Can we make this metric as optional, one that can be turned on and turned off by config?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
desaikomal: Rahul, i am also not following why you need sleep. 
I would keep the 'startTime' and always report 'currentTime - startTime as part of the reporting. so everytime this method gets called by our underlying system, you do that. why explicitly sleep?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: That's a good suggestion, but even if we parallelize this loop but still all those threads would be updating same guage value so it would be sequential only, right?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: That's good suggestion, thanks!

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -109,9 +157,21 @@ public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
...
+
   public ClusterStatusMonitor(String clusterName) {
     _clusterName = clusterName;
     _beanServer = ManagementFactory.getPlatformMBeanServer();
+    /**
+     * Start a async thread for each cluster which keeps monitoring missing top states of any resource for a cluster.
+     * This thread will keep on iterating over resources and report missingTopStateDuration for them until there are at
+     * least one resource with missing top state for a cluster.
+     */
+    _asyncMissingTopStateMonitor.start();
```
xyuanlu, line 174: Do we need to join this thread eventually?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: Sure umm only other option i can think of is to curate the list of keys which needs to be deleted and then at the end of for loop iterate over them and delete it. But i'm not sure if that would be any better in performance or code readability wise. Any suggestions?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
xyuanlu: +1 on conditional variable. Then no sleep is needed here.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: That's a good point! Reason of defining (allocating memory) in main process is that main process (ClusterStatusMonitor) will be the one who is updating this map and async thread will be just reading from it.

rahulrane50:
TFTR @mgao0 please find answers inline!
> Thanks for the PR. I have several general question:
> 
> 1. Is there a reason why we need a long-running thread for AsyncMissingTopStateMonitor?  Can we couple the update of ResourceMonitor with the update of updateMissingTopStateResourceMap in ClusterStatusMonitor, instead of checking the map periodically?
--> The reason of decoupling metric reporting with existing main process of clusterstatusmonitor is to report metric irrespective of any event is being handled or not. Hence we couple of this with clusterstatusmonitor then thread will report metric only when resourcemap is updated ie., any event happens. About long-running thread, it sleeps whenever all resources have all partitions with top state recovered. To save some resources i had added a sleep in thread.
> 2. Can we make this metric as optional, one that can be turned on and turned off by config?
--> That's a good point! To be honest i'm not sure. In my mind this is very small features hence guarding it behind flag may not be that useful. In terms resource usage or performance it should not that much since it's single thread which should be running only when any resources have any partitions with missing top state. I'm open for suggestions though :)

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: let me know if above comment makes sense

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: TFTR @qqu0127! Fixed it.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: TFTR @desaikomal @xyuanlu let me know if my comment makes sense!

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: Good question! I think it has to be stopped/interrupted for sure by ClusterStatusMonitor when it's cleaning up all state but main process (ClusterStatusMonitor) do not have to wait until it's finished because it's async long running thread and metrics reporting is never ending job so it's lifecycle can be tied with main process. Let me know if that makes sense.

xyuanlu: Sorry I did not follow. When ClusterStatusMonitor terminates, we need to join this thread dont we?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -652,6 +738,8 @@ public void reset() {
       _rebalanceFailureCount.set(0L);
       _continuousResourceRebalanceFailureCount.set(0L);
       _continuousTaskRebalanceFailureCount.set(0L);
+      _asyncMissingTopStateMonitor.interrupt();
+      _missingTopStateResourceMap.clear();
```
xyuanlu, line 742: do we want to reset the metrics to 0 as well?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,55 @@
...
+
+              }
+            }
+            // TODO: Check if this SLEEP_TIME is correct? Thread should keep on increasing the counter continuously until top
+            //  state is recovered but it can sleep for reasonable amount of time in between.
+            sleep(100);
+          }
+        }
+      } catch (InterruptedException e) {
+        LOG.error("AsyncMissingTopStateMonitor has been interrupted.", e);
```
xyuanlu, line 103: Question about the interrupt handling here and in reset(). 
When the thread receives an interrupt in reset() and we log error and returns..? Then no monitor thread is created and loop to continue monitoring...?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
xyuanlu: Having a sleep() with some random number is not as preferable as using conditional variable...

xyuanlu:
A general question here. To me it seems the new metrics is using some non 0 random number indicating at least one partition had no top state. (and 0 means all partition is good). I feel like we could have a more logistic meaningful number.

xyuanlu:
--> The reason of decoupling metric reporting with existing main process of clusterstatusmonitor is to report metric irrespective of any event is being handled or not.

Please correct me if I am wrong. In your implementation, I think update happens at "TopStateHandoffReportStage" witch is trigged when event (update on current stage etc.) We didn't quite decouple these two...?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
qqu0127: One potential way in my mind is to separate the metrics update logic and entry remove. We can call `map.entrySet().removeIf()` after the forloop. But essentially it's iterating again. 
I don't have strong opinion on this, so you feel free to make the call.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: That's a very good point and catch. Handled this correctly now in active() and reset() flow.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: yup handled it now. Thanks for catching this.

rahulrane50:
@xyuanlu and @desaikomal I have updated PR with duration now instead up just randomly incrementing counter. I hope it makes sense now.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: I checked the flow and async thread should be tied with ClusterStatusMonitor thread. That means currently, all beans are registered in active() method and unregistered or reset in reset() method. So whenever Helix controller will activate cluster status monitor it should start this async thread and whenever helix controller stops/resets cluster status monitor it should stop this async thread.

xyuanlu: Does this mean after `reset()` we should kill this metrics reporting thread?

rahulrane50: Yup. that's what is done here. In reset() we unregister all things, clear map and then stop/interrupt thread. Async thread will be started again in active() call.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,60 @@
...
+            for (Iterator<Map.Entry<String, Map<String, Long>>> resourcePartitionIt =
+                _missingTopStateResourceMap.entrySet().iterator(); resourcePartitionIt.hasNext(); ) {
+              Map.Entry<String, Map<String, Long>> resourcePartitionEntry = resourcePartitionIt.next();
+              // Iterate over all partitions and if any partition has missing top state greater than threshold then report
+              // it.
+              ResourceMonitor resourceMonitor = getOrCreateResourceMonitor(resourcePartitionEntry.getKey());
+              // If all partitions of resource has top state recovered then reset the counter
+              if (resourcePartitionEntry.getValue().isEmpty()) {
+                resourceMonitor.resetMissingTopStateDurationGuage();
+                resourcePartitionIt.remove();
```
desaikomal, line 87: when you remove the entry from the map on which you are already iterating, is it ok?
in C++, i know that we can't have iterator and remove from iterator in same loop as "iterator" will need to reassign.
Please check.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -583,6 +643,31 @@ public void updateMissingTopStateDurationStats(String resourceName, long totalDu
     }
   }
 
+  public void updateMissingTopStateDurationThreshold(long missingTopStateDurationThreshold) {
+    _asyncMissingTopStateMonitor.setMissingTopStateDurationThreshold(missingTopStateDurationThreshold);
```
desaikomal, line 647: now this is where i have question: since your _asyncMissingTopStateMonitor has synchronized and is in tight-loop, will this method go through? May be when the method is in "wait" state, it releases the lock and so this method can go through

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,60 @@
...
+          }
+        }
+      } catch (InterruptedException e) {
+        LOG.error("AsyncMissingTopStateMonitor has been interrupted.", e);
+      }
+    }
+
+    public void reset() {
+      for (String resource : _missingTopStateResourceMap.keySet()) {
+        ResourceMonitor resourceMonitor = getOrCreateResourceMonitor(resource);
```
desaikomal, line 106: can you not just use getResourceMonitor() as you don't want to create resource monitor if you are in reset path

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
```diff
@@ -139,6 +142,10 @@ public ResourceMonitor(String clusterName, String resourceName, ObjectName objec
     _partitionTopStateNonGracefulHandoffDurationGauge =
         new HistogramDynamicMetric("PartitionTopStateNonGracefulHandoffGauge", new Histogram(
             new SlidingTimeWindowArrayReservoir(getResetIntervalInMs(), TimeUnit.MILLISECONDS)));
+    _missingTopStateDurationGuage =
+        new HistogramDynamicMetric("MissingTopStateDurationGauge", new Histogram(
+            new SlidingTimeWindowArrayReservoir(getResetIntervalInMs(), TimeUnit.MILLISECONDS)));
```
desaikomal, line 147: can this be your sleep duration?

helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
```diff
@@ -85,6 +85,9 @@ private void updateTopStateStatus(ResourceControllerDataProvider cache,
     if (cache.getClusterConfig() != null) {
       durationThreshold = cache.getClusterConfig().getMissTopStateDurationThreshold();
     }
+    if (clusterStatusMonitor != null) {
+      clusterStatusMonitor.updateMissingTopStateDurationThreshold(durationThreshold);
```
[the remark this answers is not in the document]
rahulrane50: That's correct if threshold is not set then it's no-op thread. But in this case most of the metrics are not reported so i'm assuming that this config is used commonly.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: This is the only way i could find for removing entries while iterative over map in java. It works fine

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: From comments above method getResourceMonitor() it says it's only for tests. I think it's because it's not synchronized like getOrCreateResourceMonitor().  Anyways it doesn't create resource monitor everytime.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,60 @@
...
+            }
+          }
+        }
+      } catch (InterruptedException e) {
+        LOG.error("AsyncMissingTopStateMonitor has been interrupted.", e);
+      }
+    }
+
+    public void reset() {
+      for (String resource : _missingTopStateResourceMap.keySet()) {
```
xyuanlu, line 105: May I ask why we need a new reset() method here? It is only being called in ClusterStatusMonitor reset() and that reset() is basically a close() function for ClusterStatusMonitor?

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,60 @@
...
+    }
+
+    public void setMissingTopStateDurationThreshold(long missingTopStateDurationThreshold) {
+      _missingTopStateDurationThreshold = missingTopStateDurationThreshold;
+    }
+
+    @Override
+    public void run() {
+      try {
+        synchronized (this) {
```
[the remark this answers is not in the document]
rahulrane50: So other methods are not synchronized hence other thread (ClusterStatusMonitor) thread can still call these methods without need of acquiring thread objects's lock. (https://users.cs.fiu.edu/~weiss/cop4338_spr06/lectures/Threads/threads.pdf See slide 19). IMHO it's okay because setMissingTopStateDurationThreshold is not updating any shared resource. And reset method is like reseting the resource map which is again okay.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
rahulrane50: Answered in above comments. It's just not about in wait() state but whenever it's timesliced and this thread get it's timeslice it can call any non-synchronized methods of _asyncMissingTopStateMonitor.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
rahulrane50: That's an excellent suggestion. After digging deeper, so the way Histrogram metrics works is based on reservoir sampling meaning even if stream of samples are provided to this guage but it will pick only k random samples (or based on sliding window provided here) and discard other samples provided. But still i just add sleep for very less duration to save some CPU cycles. So to summarize, as long as metric reporting is concerned, no matter how many samples are provided to this guage only handlful of samples will be emitted by helix controller. 

Code pointer for reference : https://www.tabnine.com/code/java/classes/com.codahale.metrics.SlidingTimeWindowArrayReservoir

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
junkaixue: Yes. But it will save our sequential computational time as this update is synchronized in Helix regular pipelines.

helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
```diff
@@ -319,12 +322,16 @@ private void reportTopStateHandoffFailIfNecessary(ResourceControllerDataProvider
...
+    if (startTime > 0 && !record.isFailed()) {
       if (clusterStatusMonitor != null) {
-        clusterStatusMonitor.updateMissingTopStateDurationStats(resourceName, 0L, 0L, false, false);
+        clusterStatusMonitor.updateMissingTopStateResourceMap(resourceName, partitionName, true, startTime);
+      }
+      if (System.currentTimeMillis() - startTime > durationThreshold) {
+        record.setFailed();
+        missingTopStateMap.get(resourceName).put(partitionName, record);
+        if (clusterStatusMonitor != null) {
+          clusterStatusMonitor.updateMissingTopStateDurationStats(resourceName, 0L, 0L, false, false);
```
junkaixue, line 333: Maybe we can deprecate this piece of logic in the future. Because anyway we already report the top state missing there is a one.

I am not sure whether would be useful to keep the duration one for end-to-end.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,61 @@
...
+              } else {
+                for (Long missingTopStateStartTime : resourcePartitionEntry.getValue().values()) {
+                  if (_missingTopStateDurationThreshold < Long.MAX_VALUE && System.currentTimeMillis() - missingTopStateStartTime > _missingTopStateDurationThreshold) {
+                    resourceMonitor.updateMissingTopStateDurationGuage(System.currentTimeMillis() - missingTopStateStartTime);
+                  }
+                }
+
+              }
+            }
+            sleep(50); // Instead of providing stream of durtion values to histogram thread can sleep in between to save some CPU cycles.
```
junkaixue, line 97: I am still concerning this number as hard coding here... It would not be a good thing. Could you explain why adding this number helps?

rahulrane50: Had a offline discussion with Junkai. After giving it a thought i came with below solution please let me know if this looks okay. Summary :
1. Original sleep here was to save few computational cycles in this tight for loop but it's difficult to justify correct sleep duration. 
2. In new solution, thread will report a metric per partition only if it's not reported within last sliding window reset interval. Now this has few benefits as : one if sleeping stops thread to report metrics for all resources and all partitions. But that's may be wrong because what if during that sleep time resource has new partitions with missing top state. Hence ideally if thread has reported at least one duration for that partition then it can skip that partition until it's sliding window has started new window. 
@desaikomal hence i didn't add sleep here for sliding window reset time but used that value to determine of duration should be reported for that partition or not.

helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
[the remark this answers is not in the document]
xyuanlu: I am not sure if we can assume that.  Normally we have default value and user can set -1 if they want to disable this.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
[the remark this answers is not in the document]
xyuanlu: Then why we want synchronized...?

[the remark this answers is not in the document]
rahulrane50: @xyuanlu because it's needed for wait() and notify() mechanism.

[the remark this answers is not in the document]
xyuanlu: You just need to lock the part where we check if condition is met. Not the whole while(true) block.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ClusterStatusMonitor.java
```diff
@@ -55,6 +55,81 @@
 import org.slf4j.LoggerFactory;
 
 public class ClusterStatusMonitor implements ClusterStatusMonitorMBean {
+  private class AsyncMissingTopStateMonitor extends Thread {
```
junkaixue, line 58: This will be meaningful if our pipeline speed is very fast. Otherwise, it will not help as the cache updated state is refreshed per pipeline. I dont believe we need build this thread but just rely on pipeline call as we did before. 

But we need to remove the constraint of doing only final reporting.

We can discuss about it tomorrow f2f.

rahulrane50:
After internal syncup, it makes sense to not have this real-time metric. Helix will emit a metric as a part of current default pipeline (which is triggered on any helix event), to report number of partitions with missing top state beyond set threshold value. I will close this PR and create a new one to address this.

PR #2359, opened 2023-02-01
Fix ZkClient retry logic for customized callback and test

xyuanlu:
### Issues

#2363 
When async retry is canceled in ZkClient, user injected callback that implements handle() in ZkAsyncCallback.DefaultCallback() will never be executed.
If there is no retry is needed, user logic in handle() is executed in either operation failed or succeeded cases.
These 2 behavior does not align.

### Description

To solve the previous issue, this change add execution of handle() when retry is canceled.

### Tests

testAsyncRetryCustomizedCallback

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
@@ -54,6 +54,14 @@ public class ZkAsyncCallbacks {
       callback(rc, path, ctx);
     }
 
+    public Stat getStat() {
+      return _stat;
+    }
+
+    public byte[] getData() {
+      return _data;
+    }
+
     @Override
     public void handle() {
       // TODO Auto-generated method stub
@@ -61,7 +69,7 @@ public class ZkAsyncCallbacks {
 
     @Override
     protected void recordFailure(int rc, String path, ZkAsyncCallMonitorContext monitor) {
-      if(rc != Code.NONODE.intValue()) {
+      if (rc != Code.NONODE.intValue()) {
         monitor.recordFailure(path);
       }
     }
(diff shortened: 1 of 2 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -259,10 +276,15 @@ public int getRc() {
...
 
+    public String getPath() {
+      return _path;
+    }
+
     @Override
     public void notifyCallers() {
       LOG.warn("The callback {} has been cancelled.", this);
       markOperationDone();
+      handle();
```
junkaixue, line 287: This confuses me. If we already mark operations are done, then why we need to call handle. Shall we put handle before operation done mark?

Also in previous function logic, mark operation done is in finally statement. It is lastly called after handle(). This logic is changed order.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
```diff
@@ -198,12 +211,14 @@ public void callback(int rc, String path, Object ctx) {
       }
 
       _rc = rc;
+      _path = path;
```
junkaixue, line 214: Why we need to update this path?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
xyuanlu: Thanks for the suggestion. Putting handle() before markOperationDone() make sense.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/callback/ZkAsyncCallbacks.java
xyuanlu: I added the _path car so that it can be acquired by caller when implementing async for metaclient.

junkaixue (approved):
lgtm. Address the other comments pls.

PR #2371, opened 2023-02-07
Change async API

xyuanlu:
### Issues

#2237

### Description

This refines async API.
1. Allow caller to pass a context to callback.
2. Change callback type of asyncSet

### Tests

NA
Test will be added when implementation is added.

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/meta-client/src/main/java/org/apache/helix/metaclient/api/AsyncCallback.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/api/AsyncCallback.java
@@ -20,6 +20,8 @@ package org.apache.helix.metaclient.api;
  */
 
 import java.util.List;
+import javax.annotation.Nullable;
+
 
 /**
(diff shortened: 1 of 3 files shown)


junkaixue (commented):
Code change is fine to me. But why we need to allow user to pass context? If we already give them callback, why they still need the input

xyuanlu:
The context is passed to user callback. If user need an object while executing callback, we need to provide them the option. 
As a reference, ZK async API provides the same signature. Helix pass a retry event handling object in this context for retry when failure.

junkaixue:
Why the context cannot be passed in callback as an argument?

xyuanlu:
Removed context field as our offline discussion. Caller should maintain and manager object they owned if it is needed in callback function. User object should not be passed to client if client does not use this info.

xyuanlu:
This PR is ready for merge. Commit message:
Update async API for metaclient.

PR #2377, opened 2023-02-14
MetaClient Container Node Implementation

Marcosrico:
### Issues

https://github.com/apache/helix/issues/2237

### Description

Implementation of Container Node in ZkMetaClient.

### Tests

Unit test `createContainer()`.
Note: Cannot test end to end behavior (i.e. that the container node disappears after it's children are deleted) as the current setup of the zk server doesn't permit it. This is also a issue in HelixZkClient testbase. However, it has been tested and confirmed that the actual behavior does work (in native zkclient) and is merely a unit testing bug. Will track and fix this effort in the future but current implementation will use existing test base framework.

### Code Quality

### DIFF
--- a/meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestZkMetaClient.java
+++ b/meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestZkMetaClient.java
@@ -57,9 +57,8 @@ public class TestZkMetaClient {
   private static final String ZK_ADDR = "localhost:2183";
   private static final int DEFAULT_TIMEOUT_MS = 1000;
   private static final String ENTRY_STRING_VALUE = "test-value";
-  protected static final String TRANSACTION_TEST_KEY_PREFIX = "/sharding-key-0";
-  protected static String PARENT_PATH = TRANSACTION_TEST_KEY_PREFIX + "/RealmAwareZkClient";
-  protected static final String TEST_INVALID_PATH = TRANSACTION_TEST_KEY_PREFIX + "_invalid" + "/a/b/c";
+  protected static String PARENT_PATH = "/ZkClient";
+  protected static final String TEST_INVALID_PATH = "_invalid" + "/a/b/c";
 
   private final Object _syncObject = new Object();
 
@@ -342,7 +341,6 @@ public class TestZkMetaClient {
 
     try(ZkMetaClient<String> zkMetaClient = createZkMetaClient()) {
       zkMetaClient.connect();
-      zkMetaClient.create(TRANSACTION_TEST_KEY_PREFIX, ENTRY_STRING_VALUE);
 
       //Create Nodes
       List<Op> ops = Arrays.asList(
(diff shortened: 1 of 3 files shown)


meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestZkMetaClient.java
```diff
@@ -67,8 +68,16 @@ public class TestZkMetaClient {
 
   private ZkServer _zkServer;
 
+  /**
+   * Creates local Zk Server
+   * Note: Cannot test container / TTL node end to end behavior as
+   * the zk server setup doesn't allow for that. However, the actual
+   * behavior has been verified to work on native ZK Client.
```
xyuanlu, line 75: Thanks for the change. Could you please give a bit more detail explaining why the set up dues not support container / TTL node and what are needed if we want to support that in unit test?

mgao0: +1 
Does setting this `System.setProperty("zookeeper.extendedTypesEnabled", "true");` not help?

Marcosrico: To enable proper end-to-end behavior between TTL and Container nodes in unit testing, it is necessary to invoke ContainerManager.java when setting up the Zk server. ContainerManager object is responsible for managing the creation / deletion of znodes (lifecycle of the nodes).

Since nodes with TTL and Container modes are designed to be deleted after a certain interval, a zk server with a ContainerManager is required for running these nodes. This can be achieved by using classes such as LeaderZooKeeperServer.java or ZooKeeperServerMain.java. However, the current implementation of zk server in metaclient and helix test base utilizes a simplified version of ZooKeeperServer, which lacks a ContainerManager implementation. 

It is necessary to create both a ContainerManager and `System.setProperty("zookeeper.extendedTypesEnabled", "true");` in order for TTL and Container nodes to function properly.

For future work, the setup of the zk server needs to be changed to use the appropriate classes mentioned above to enable the proper functioning of TTL and Container nodes in unit testing. I'll update the comments.

qqu0127: Thanks for the info Marcos! 
Is this a Zookeeper side issue for the inconsistency? Do we need ZK version upgrade to have ContainerManager? Just want to understand if we have external dependency in order to test this.

qqu0127 (approved):
Thanks. Code change LGTM, one question for my own understanding

qqu0127 (approved):
Thanks. Code change LGTM, one question for my own understanding

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestZkMetaClient.java
Marcosrico: Of course! And no to both. I believe the `ZooKeeperServer `class exists in Zk just to create servers easier but with less functionalities. And we do not need to upgrade any ZK version as the classes mentioned above are available in the current version of zk that we use.

Just for additional information, I was able to successfully create a zk server using `ZooKeeperServerMain.java,` but replacing the existing framework and modifying it so it's thread safe would require more bandwidth than available. And it is not a major issue so the priority is low.

Marcosrico:
This PR is ready to merge, approved by @qqu0127
Commit message:
Implementation of Container Nodes in ZkMetaClient

PR #2380, opened 2023-02-17
Refactor updating and searching for threadpool for state transition messages

### LINKED ISSUE #2379: Simplify logic in HelixTaskExecutor to allow easier plug-in of customized threadpool APIs
Current there are two APIs where Helix allows users to provide threadpool for executing certain types of state transition message, which is `public ExecutorService getExecutorService(String resourceName);` and `public ExecutorService getExecutorService(String resourceName, String fromState, String toState);`. Each of these two APIs is implemented internally by leveraging a threadpoolkey (example: STATE_TRANSITION.resourceName, or STATE_TRANSITIOIN.resourceName.fromState.toState), so that we categorize the messages appropriately. In participant side logic (HelixTaskExecutor.class), we used one set for each API to determine if a threadpoolkey has been assigned a threadpool before. And there are if-else clauses to determine which threadpool to be used from the results of calling these two APIs. It is not very scalable, meaning a lot of code needs to be implemented in order to add another parameter to finer grind categorize messages and assign customized threadpool. We need a more generic solution.

For generating threadpoolkey, and determine which threadpool to be used, we should make it so that each time we just need to do minimal code change, and no need to duplicate similar logic.

**Additional context**

mgao0:
### Issues

Resolves #2379

### Description

This commit adds a MessageInfo sub-class in Message class, which can help managing the parameters we use to categorize messages, and to generate a corresponding key as a message identifier. 

### Tests

TestHelixTaskExecutor.testUpdateAndFindMessageThreadpool

[ERROR] Failures: 
[ERROR]   TestNoThrottleDisabledPartitions.testDisablingTopStateReplicaByDisablingInstance:98 expected:<false> but was:<true>
[ERROR]   TestNoThrottleDisabledPartitions.testNoThrottleOnDisabledInstance:231->setupEnvironment:317->setupCluster:436 » ZkClient
[ERROR]   TestP2PNoDuplicatedMessage.testP2PStateTransitionEnabled:180 expected:<true> but was:<false>
[ERROR]   TestWagedRebalanceFaultZone.testAddZone:270->validate:318->validateZoneAndTagIsolation:342 expected:<3> but was:<2>
[ERROR]   TestRecurringJobQueue.testDeletingRecurrentQueueWithHistory:298 expected:<true> but was:<false>
[INFO] 
[ERROR] Tests run: 1323, Failures: 5, Errors: 0, Skipped: 0
[INFO] 

Run each test locally, and it will pass

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -161,8 +161,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
   final Set<String> _knownMessageIds;
 
   /* Resources whose configuration for dedicate thread pool has been checked.*/
-  final Set<String> _resourcesThreadpoolChecked;
-  final Set<String> _transitionTypeThreadpoolChecked;
+  final Set<String> _threadpoolChecked;
 
   // timer for schedule timeout tasks
   final Timer _timer;
@@ -191,8 +190,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
     _batchMessageExecutorService = Executors.newCachedThreadPool();
     _monitor.createExecutorMonitor("BatchMessageExecutor", _batchMessageExecutorService);
 
-    _resourcesThreadpoolChecked = Collections.newSetFromMap(new ConcurrentHashMap<>());
-    _transitionTypeThreadpoolChecked = Collections.newSetFromMap(new ConcurrentHashMap<>());
+    _threadpoolChecked = Collections.newSetFromMap(new ConcurrentHashMap<>());
 
     _lock = new Object();
     _statusUpdateUtil = new StatusUpdateUtil();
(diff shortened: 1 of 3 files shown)


helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -191,8 +190,7 @@ public HelixTaskExecutor(ParticipantStatusMonitor participantStatusMonitor,
     _batchMessageExecutorService = Executors.newCachedThreadPool();
     _monitor.createExecutorMonitor("BatchMessageExecutor", _batchMessageExecutorService);
 
-    _resourcesThreadpoolChecked = Collections.newSetFromMap(new ConcurrentHashMap<>());
-    _transitionTypeThreadpoolChecked = Collections.newSetFromMap(new ConcurrentHashMap<>());
+    _threadpoolChecked = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
junkaixue, line 193: How you can combine them? Say we just need the "resource" based thread pool, are you doing prefixing searching?


This is already changing the logic.

junkaixue:
If this changes logic, let's remove "refactor" word. Refactor means the logic is not changed.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
mgao0: Previously we use different sets to do search for different property-based threadpool. Right now I'm combining them into one set, but making the key generation structural, so we maintain the capability of searching threadpool in different levels.

junkaixue: I read through the code. It breaks the logic... Let's have some chat about it.

junkaixue: Synced offline. It is not logic change but code style could be hard to read. Let's refactor it with minimum changes.

helix-core/src/main/java/org/apache/helix/model/Message.java
```diff
@@ -966,4 +966,46 @@ public boolean isValid() {
...
+   * This class is for categorizing state transition messages based on certain properties, and generating
+   */
+  public static class MessageInfo {
+    public String _msgType;
+    public String _resourceName;
+    public String _fromState;
+    public String _toState;
+
+    public enum MessageIdentifierBase {
+      PER_RESOURCE, PER_STATE_TRANSITION_TYPE
```
junkaixue, line 980: NIT: formatter problem?

helix-core/src/main/java/org/apache/helix/model/Message.java
```diff
@@ -966,4 +966,46 @@ public boolean isValid() {
...
+      _toState = message.getToState();
+    }
+
+    public String getMessageIdentifier(MessageIdentifierBase basis) {
+      String delimiter = ".";
+      if (_msgType == null) {
+        return null;
+      }
+      String identifier = _msgType;
+      if (basis.ordinal() >= MessageIdentifierBase.PER_RESOURCE.ordinal()) {
```
junkaixue, line 996: How was the ordinal work? The number larger means the position higher? What if this non-existing type?

junkaixue (commented):
I dont have major concern. But would like to understand more on how it works.

helix-core/src/main/java/org/apache/helix/model/Message.java
mgao0: The original code is to concatenate strings for each level resource -> state transition type... This is for the same purpose.
But as we discussed earlier, some levels can be parallel, not necessarily hierarchical. For example we can have resourceName_stateTransitionType, resourceName_rebalanceType, where stateTransitionType and rebalanceType are parallel. I modified the code of comparing the ordinal, please take another look. Thanks.

helix-core/src/main/java/org/apache/helix/model/Message.java
```diff
@@ -966,4 +966,45 @@ public boolean isValid() {
...
+      if (_resourceName == null) {
+        return null;
+      }
+      identifier = String.join(delimiter, identifier, _resourceName);
+      if (basis == MessageIdentifierBase.PER_STATE_TRANSITION_TYPE) {
+        if (_fromState == null || _toState == null) {
+          return null;
+        }
+        identifier = String.join(delimiter, identifier, _fromState, _toState);
+      }
```
junkaixue, line 1006: Let's restructure it as switch case for enum. So if we have more other types, it can help.

junkaixue (approved):

mgao0:
This PR is ready to be merged, approved by @junkaixue , thanks!

Final commit message: Refactor updating and searching for threadpool for state transition messages

PR #2381, opened 2023-02-23
Adding a new metric to report number of partitions with missing top state beyond threshold

### LINKED ISSUE #2345: Add new metric to report missing top state for partition

rahulrane50:
### Issues

Fixes #2345 

### Description

This PR emits a new metric which will report number of partitions with missing top state beyond threshold value. It will update this value as soon as top state goes missing for a partition. This metric will be reset only after default reset time i.e., 1 min. 
Existing two metrics - FailedTopStateHandoffCounter and PartitionTopStateNonGracefulHandoffGauge are very much similar and they both try to say that top state is missing beyond certain threshold. So we would deprecate both this metrics and encourage to use this new guage.

### Tests

Current test testTopStateFailedHandoff takes care of all scenarios. Updated this test with detailed scenario.
```
mvn test -Dtest=TestTopStateHandoffMetrics
[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 18.308 s - in org.apache.helix.monitoring.mbeans.TestTopStateHandoffMetrics
[INFO]
[INFO] Results:
[INFO]
[INFO]
[INFO]
```

None

### Documentation (Optional)

NA

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
@@ -206,7 +206,9 @@ public class TopStateHandoffReportStage extends AbstractBaseStage {
 
     if (missingTopStateMap.containsKey(resourceName) && missingTopStateMap.get(resourceName)
         .containsKey(partition.getPartitionName())) {
-      // We previously recorded a top state missing, and it's not coming back
+      // We previously recorded a top state missing, and it's coming back.
+      // Note : Decrement missingTopStatePartitionsBeyondGuage in this code path because this guage will be incremented
+      //        only if we were able to record it in the first place.
       reportTopStateComesBack(cache, currentStateOutput.getCurrentStateMap(resourceName, partition),
           resourceName, partition, clusterStatusMonitor, durationThreshold,
           stateModelDef.getTopState());
@@ -304,7 +306,8 @@ public class TopStateHandoffReportStage extends AbstractBaseStage {
 
   /**
    * Check if the given partition of the given resource has a missing top state duration larger
-   * than the threshold, if so, report a top state transition failure
+   * than the threshold, if so, report a top state transition failure as well increment a guage which reports number
+   * of partitions with missing top state.
    *
    * @param cache cluster data cache
    * @param resourceName resource name
@@ -471,6 +474,7 @@ public class TopStateHandoffReportStage extends AbstractBaseStage {
       }
     }
 
+    // TODO: why are we checking if handoff happened within threshold here? it should be reported either way.
     if (handOffStartTime > 0 && handOffEndTime - handOffStartTime <= threshold) {
       long duration = handOffEndTime - handOffStartTime;
       long helixLatency = duration - fromTopStateUserLatency - toTopStateUserLatency;
--- a/helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
+++ b/helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
@@ -474,7 +474,6 @@ public class TopStateHandoffReportStage extends AbstractBaseStage {
       }
     }
 
-    // TODO: why are we checking if handoff happened within threshold here? it should be reported either way.
     if (handOffStartTime > 0 && handOffEndTime - handOffStartTime <= threshold) {
       long duration = handOffEndTime - handOffStartTime;
       long helixLatency = duration - fromTopStateUserLatency - toTopStateUserLatency;
--- a/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
+++ b/helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
@@ -71,11 +71,16 @@ public class ResourceMonitor extends DynamicMBeanProvider {
   // Counters
   private SimpleDynamicMetric<Long> _successfulTopStateHandoffDurationCounter;
   private SimpleDynamicMetric<Long> _successTopStateHandoffCounter;
-  // A new Gauage for reporting number of partitions with missing top state has been added. The main reason of deprecating
-  // this counter because it doesn't give clear picture on how many partitions are missing with top state at any point
-  // of time. New gauage will help with that. Please find more info at https://github.com/apache/helix/pull/2381
+
+  // A new Gauage _missingTopStatePartitionsBeyondThresholdGauge for reporting number of partitions with missing top
+  // state has been added. Reason of deprecating this two metrics is because they are similar and doesn't tell about
+  // how many partitions are missing beyond threshold. The average duration of different partitions hands-off would not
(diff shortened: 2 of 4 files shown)


helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
```diff
@@ -471,6 +474,7 @@ private void reportTopStateComesBack(ResourceControllerDataProvider cache, Map<S
       }
     }
 
+    // TODO: why are we checking if handoff happened within threshold here? it should be reported either way.
```
rahulrane50, line 477: @junkaixue i have a question here. Why are we updating topstate handoff duration only if it's within threshold. This means that if threshold is set to "x" ms and any partition's top state has recovered beyond "x" ms then we would not report it. Basically same behaviour as reportSingleTopStateHandoff. (If my understanding is correct then i can remove this, i have also changed test to verify this.)

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
```diff
@@ -125,6 +128,7 @@ public ResourceMonitor(String clusterName, String resourceName, ObjectName objec
     _numLessMinActiveReplicaPartitions =
         new SimpleDynamicMetric("MissingMinActiveReplicaPartitionGauge", 0L);
     _numNonTopStatePartitions = new SimpleDynamicMetric("MissingTopStatePartitionGauge", 0L);
+    _missingTopStatePartitionsBeyondThresholdGuage = new SimpleDynamicMetric("MissingTopStatePartitionsBeyondThresholdGauge", 0L);
```
qqu0127, line 131: Typo? guage -> gauge

qqu0127 (approved):
Overall LGTM, one nit, thanks

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
rahulrane50: Good catch! TFTR!

rahulrane50:
Approved by @qqu0127 , Ready to merge!
Commit message : Adding a new metric to report number of partitions with missing top state beyond threshold

helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
rahulrane50: Filed #2383. Let's move this discussion there.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
```diff
@@ -70,6 +71,7 @@ public enum RebalanceStatus {
   // Counters
   private SimpleDynamicMetric<Long> _successfulTopStateHandoffDurationCounter;
   private SimpleDynamicMetric<Long> _successTopStateHandoffCounter;
+  @Deprecated
```
desaikomal, line 74: do you want to add comment as to we are deprecating this and replacing it with another one, so that others have context on why we marked them for deprecation

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
```diff
@@ -54,6 +54,7 @@ public enum RebalanceStatus {
 
   // Gauges
   private SimpleDynamicMetric<Long> _numOfPartitions;
+  private SimpleDynamicMetric<Long> _missingTopStatePartitionsBeyondThresholdGauge;
```
desaikomal, line 57: comment or link to PR, so that we know why we added this

desaikomal (approved):
Looks good to me. it took us time to get here but we learned and reached.

desaikomal (approved):
LGTM

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
```diff
@@ -78,6 +83,13 @@ public enum RebalanceStatus {
   // Histograms
   private HistogramDynamicMetric _partitionTopStateHandoffDurationGauge;
   private HistogramDynamicMetric _partitionTopStateHandoffHelixLatencyGauge;
+  // A new Gauage for reporting number of partitions with missing top state has been added. The main reason of deprecating
+  // this gauge because it's similar to other failedTopStateCounter and averaging out duration of all those handsoff
+  // won't really help in debugging the issue. Hence please use new gauage _missingTopStatePartitionsBeyondThresholdGauge
+  // to find number of partitions with missing top state beyond threshold and to find duration use
+  // _maxSinglePartitionTopStateHandoffDuration. New gauage will help with that. Please find more info at
+  // https://github.com/apache/helix/pull/2381
+  @Deprecated
```
xyuanlu, line 92: No need to add duplicate comments.

rahulrane50: It's not duplicate, the reason to deprecate  _partitionTopStateNonGracefulHandoffDurationGauge is slightly different than other one.

xyuanlu:
Thanks for making the change. Code change looks good to me. I would suggest resolve open question and then let's put conclusion in comment, and merge the change. I prefer not to have open comment in final code.

helix-core/src/main/java/org/apache/helix/monitoring/mbeans/ResourceMonitor.java
xyuanlu: Thanks for the reply. Please consider using more succinct wording.

helix-core/src/main/java/org/apache/helix/controller/stages/TopStateHandoffReportStage.java
xyuanlu, line 477: If handoff happened within timeout, Helix treat it as regular handoff. 
If handoff happened longer than timeout, Helix treat it as handoff failure. I think there may be different metrics counting this. Please read it though and see how the new metrics fit into picture. 
I would prefer not leave question as comments in code.

rahulrane50: Sure let's discuss this over above filed issue. I removed this TODO because this is different thread of discussion and I don't want to hold this PR until we conclude on above topic. Thanks for the suggestion!

xyuanlu (approved):
Thanks for addressing comments. 
Please rerun failed test and update description.

PR #2390, opened 2023-03-02
Add API for users to provide customized threadpool for different categories of state transition messages

### LINKED ISSUE #2389: Need an API to provide customized threadpool based on different rebalance type
In some cases on the participant side, we have been observing that executing load rebalance messages takes long time which eats up the resource for executing recovery rebalance messages (which is more important). Therefore, some customers request to have an API where they can supply different threadpools for executing load rebalance messages and recovery rebalance messages.

Add an API on top of the two we current have right now.

**Additional context**

mgao0:
### Issues

Resolves #2389

### Description

Currently we have MessageInfo for categorizing messages, this commit adds one more property RebalanceType to MessageInfo so that user can provide threadpool based on this property. 

### Tests

Added a section in TestHelixTaskExecutor. testUpdateAndFindMessageThreadpool().

Running

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -297,9 +297,11 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
           stateModelFactory.getExecutorService(msgInfo);
       if (customizedExecutorService != null) {
         String msgInfoBasedKey = msgInfo.getMessageIdentifier(customizedExecutorService.getBase());
-        _msgInfoBasedThreadpoolChecked.add(msgInfoBasedKey);
-        _executorMap.put(msgInfoBasedKey, customizedExecutorService.getExecutorService());
-        return;
+        if (msgInfoBasedKey != null) {
+          _msgInfoBasedThreadpoolChecked.add(msgInfoBasedKey);
+          _executorMap.put(msgInfoBasedKey, customizedExecutorService.getExecutorService());
+          return;
+        }
       }
     }
 
--- a/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
+++ b/helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
@@ -163,6 +163,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
   /* Resources whose configuration for dedicate thread pool has been checked.*/
   final Set<String> _resourcesThreadpoolChecked;
   final Set<String> _transitionTypeThreadpoolChecked;
+  final Set<String> _msgInfoBasedThreadpoolChecked;
 
   // timer for schedule timeout tasks
   final Timer _timer;
@@ -193,6 +194,7 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
 
     _resourcesThreadpoolChecked = Collections.newSetFromMap(new ConcurrentHashMap<>());
     _transitionTypeThreadpoolChecked = Collections.newSetFromMap(new ConcurrentHashMap<>());
+    _msgInfoBasedThreadpoolChecked = Collections.newSetFromMap(new ConcurrentHashMap<>());
 
     _lock = new Object();
     _statusUpdateUtil = new StatusUpdateUtil();
@@ -290,6 +292,17 @@ public class HelixTaskExecutor implements MessageListener, TaskExecutor {
         manager.getStateMachineEngine().getStateModelFactory(stateModelName, factoryName);
 
     Message.MessageInfo msgInfo = new Message.MessageInfo(message);
+    if (stateModelFactory != null) {
+      StateModelFactory.CustomizedExecutorService customizedExecutorService =
+          stateModelFactory.getExecutorService(msgInfo);
+      if (customizedExecutorService != null) {
+        String msgInfoBasedKey = msgInfo.getMessageIdentifier(customizedExecutorService.getBase());
+        _msgInfoBasedThreadpoolChecked.add(msgInfoBasedKey);
+        _executorMap.put(msgInfoBasedKey, customizedExecutorService.getExecutorService());
+        return;
+      }
+    }
+
     String perStateTransitionTypeKey =
         msgInfo.getMessageIdentifier(Message.MessageInfo.MessageIdentifierBase.PER_STATE_TRANSITION_TYPE);
     if (perStateTransitionTypeKey != null && stateModelFactory != null && !_transitionTypeThreadpoolChecked.contains(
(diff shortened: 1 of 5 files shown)


helix-core/src/main/java/org/apache/helix/model/Message.java
```diff
@@ -660,6 +666,18 @@ public void setBucketSize(int bucketSize) {
     }
   }
 
+  public void setSTRebalanceType(STRebalanceType stRebalanceType) {
+    setAttribute(Attributes.ST_REBALANCE_TYPE, stRebalanceType.name());
+  }
+
+  public STRebalanceType getSTRebalanceType() {
+    String rebalanceTypeStr = getAttribute(Attributes.ST_REBALANCE_TYPE);
+    if (rebalanceTypeStr == null) {
```
xyuanlu, line 675: nit: can be inline with 678

helix-core/src/main/java/org/apache/helix/model/Message.java
```diff
@@ -969,23 +987,32 @@ public boolean isValid() {
...
+    /**
+     * This class is used for defining what properties are used to categorize messages.
+     * Note that the bases should be arranged in a lower level -> higher level order,
+     * because we do backwards searching when looking for customized threadpool.
+     */
     public enum MessageIdentifierBase {
-      PER_RESOURCE,
-      PER_STATE_TRANSITION_TYPE
+      PER_RESOURCE, // Most basic level (L0), just include message type and resource name
+      PER_REBALANCE_TYPE, // One level above PER_RESOURCE (L1), concatenate rebalance type (load / recovery) after PER_RESOURCE identifier string
```
xyuanlu, line 1006: nit: maybe the enum order align with categorize order?

rahulrane50 (commented):
I don't have lot of context but overall LGTM!

helix-core/src/main/java/org/apache/helix/participant/statemachine/StateModelFactory.java
```diff
@@ -260,4 +261,42 @@ public ExecutorService getExecutorService(String resourceName) {
   public ExecutorService getExecutorService(String resourceName, String fromState, String toState) {
     return null;
   }
+
+  public static class CustomizedExecutorService {
+    private Message.MessageInfo.MessageIdentifierBase _base;
+    private ExecutorService _executorService;
```
qqu0127, line 267: These two can be final

helix-core/src/main/java/org/apache/helix/model/Message.java
```diff
@@ -115,6 +116,11 @@ public enum MessageState {
     UNPROCESSABLE // get exception when create handler
   }
 
+  public enum STRebalanceType {
+    LOAD_REBALANCE,
+    RECOVERY_REBALANCE
+  }
+
```
qqu0127, line 123: Not an issue, just a thought. I think there is existing `RebalanceType` in a throttling related class. For widely used and fundamental enum like this, maybe it deserves a separate class and reduce duplication

mgao0: Yes, we have thought about reusing the RebalanceType in StateTransitionThrottleConfig class, but reusing it may cause confusion as this use case is not related to throttling. Separating it out may be an option, but the enum in StateTransitionThrottleConfig is widely used, and it will need a lot of code changes, overall it's not worthy to do it, so it is also not considered right now.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -290,6 +292,17 @@ private void updateStateTransitionMessageThreadPool(Message message, HelixManage
         manager.getStateMachineEngine().getStateModelFactory(stateModelName, factoryName);
 
     Message.MessageInfo msgInfo = new Message.MessageInfo(message);
+    if (stateModelFactory != null) {
+      StateModelFactory.CustomizedExecutorService customizedExecutorService =
+          stateModelFactory.getExecutorService(msgInfo);
+      if (customizedExecutorService != null) {
+        String msgInfoBasedKey = msgInfo.getMessageIdentifier(customizedExecutorService.getBase());
+        _msgInfoBasedThreadpoolChecked.add(msgInfoBasedKey);
```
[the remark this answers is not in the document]
mgao0: Good catch, updated!

helix-core/src/main/java/org/apache/helix/model/Message.java
mgao0: I put PER_REBALANCE_TYPE before PER_STATE_TRANSITION_TYPE is because currently most users are using  PER_STATE_TRANSITION_TYPE, so when we do reverse order traversal to look up for the threadpool, we search PER_STATE_TRANSITION_TYPE first, so it saves some checks. Since these two are in the same level, it should be okay.
But this is a good suggestion, I found that the enum and the order in getMessageIdentifier() is not aligned, so I updated it, please take a look, thanks.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -361,17 +374,13 @@ ExecutorService findExecutorServiceForMsg(Message message) {
...
-          executorService = _executorMap.get(perStateTransitionTypeKey);
-        } else if (_executorMap.containsKey(perResourceTypeKey)) {
-          LOG.info("Find per-resource thread pool with key: " + perResourceTypeKey);
-          executorService = _executorMap.get(perResourceTypeKey);
+        for (int i = Message.MessageInfo.MessageIdentifierBase.values().length - 1; i >= 0; i--) {
+          String msgIdentifer = msgInfo.getMessageIdentifier(Message.MessageInfo.MessageIdentifierBase.values()[i]);
+          if (msgIdentifer != null && _executorMap.containsKey(msgIdentifer)) {
+            LOG.info(String.format("Find customized threadpool for %s", msgIdentifer));
+            executorService = _executorMap.get(msgIdentifer);
+            break;
```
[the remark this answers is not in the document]
mgao0: In the Enum we declare types generally from coarser to finer, so we search in reverse order. If a finer type has a threadpool assigned, then we use the finer one.

helix-core/src/main/java/org/apache/helix/model/Message.java
qqu0127: Agree, feel free to close this one. Thanks

qqu0127 (approved):
Overall LGTM, please address other comments

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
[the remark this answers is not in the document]
mgao0: Oh I see, I think declare from lower level to higher level is more intuitive, so I will keep it this way.

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
```diff
@@ -361,17 +376,13 @@ ExecutorService findExecutorServiceForMsg(Message message) {
...
-        String perStateTransitionTypeKey =
-            msgInfo.getMessageIdentifier(Message.MessageInfo.MessageIdentifierBase.PER_STATE_TRANSITION_TYPE);
-        if (perStateTransitionTypeKey != null && _executorMap.containsKey(perStateTransitionTypeKey)) {
-          LOG.info(String.format("Find per state transition type thread pool for resource %s from %s to %s",
-              message.getResourceName(), message.getFromState(), message.getToState()));
-          executorService = _executorMap.get(perStateTransitionTypeKey);
-        } else if (_executorMap.containsKey(perResourceTypeKey)) {
-          LOG.info("Find per-resource thread pool with key: " + perResourceTypeKey);
-          executorService = _executorMap.get(perResourceTypeKey);
+        for (int i = Message.MessageInfo.MessageIdentifierBase.values().length - 1; i >= 0; i--) {
```
xyuanlu, line 379: nit: 
`for (Message.MessageInfo.MessageIdentifierBase value : Message.MessageInfo.MessageIdentifierBase.values())`
and this `value` can be used in line 380 as well

helix-core/src/main/java/org/apache/helix/messaging/handling/HelixTaskExecutor.java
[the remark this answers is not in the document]
xyuanlu: +1 on this. And we could use:
`for (Message.MessageInfo.MessageIdentifierBase value : Message.MessageInfo.MessageIdentifierBase.values())`
for simpler code and this value can be used in line 380 as well.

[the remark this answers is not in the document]
mgao0: Discussed with @xyuanlu offline. Changing enum order does simplify code, but users will not see this block of code because this is our internal logic. What users see is the enum declaration when they use the API, so making the enum easy to understand for users is more important. Therefore, keeping this order.
The original comments I put besides each enum declaration are a bit confusing in terms of their hierarchy, updated them.

xyuanlu (approved):
LGTM. Thanks for addressing comments.

mgao0:
This PR is ready to be merged, approved by @qqu0127 @xyuanlu . Thanks.

Final commit message: Add API for users to provide customized threadpool for different categories of state transition messages

PR #2392, opened 2023-03-03
check connnect only called once

xyuanlu:
### Issues

#2393

### Description

This change adds an extra check in ZkClient.connect so that it will only be invoked once.
Also avoid close called in between of `isclosed` check and connection establishment in connect().

### Tests

NA

```
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 337.691 s - in org.apache.helix.rest.server.TestResourceAssignmentOptimizerAccessor
[INFO] 
[INFO] Results:
[INFO] 
[INFO] 

```

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -33,6 +33,7 @@ import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.management.JMException;
+
 import org.apache.helix.zookeeper.api.client.ChildrenSubscribeResult;
 import org.apache.helix.zookeeper.constant.ZkSystemPropertyKeys;
 import org.apache.helix.zookeeper.datamodel.SessionAwareZNRecord;
@@ -2508,11 +2509,17 @@ public class ZkClient implements Watcher {
    */
   public void connect(final long maxMsToWaitUntilConnected, Watcher watcher)
       throws ZkInterruptedException, ZkTimeoutException, IllegalStateException {
+    acquireEventLock();
+
     if (isClosed()) {
       throw new IllegalStateException("ZkClient already closed!");
     }
+    if (_currentState != null) {
+      throw new IllegalStateException(
+          "ZkClient is not in init state. connect() has already been called.");
+    }
+
     boolean started = false;
-    acquireEventLock();
     try {
       setShutdownTrigger(false);
 
--- a/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
+++ b/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
@@ -2509,18 +2509,18 @@ public class ZkClient implements Watcher {
    */
   public void connect(final long maxMsToWaitUntilConnected, Watcher watcher)
       throws ZkInterruptedException, ZkTimeoutException, IllegalStateException {
-    acquireEventLock();
-
-    if (isClosed()) {
-      throw new IllegalStateException("ZkClient already closed!");
-    }
-    if (_currentState != null) {
-      throw new IllegalStateException(
-          "ZkClient is not in init state. connect() has already been called.");
-    }
-
     boolean started = false;
+
     try {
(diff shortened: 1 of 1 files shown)


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2508,11 +2509,17 @@ public void connect(final long maxMsToWaitUntilConnected)
...
+    acquireEventLock();
+
     if (isClosed()) {
       throw new IllegalStateException("ZkClient already closed!");
     }
+    if (_currentState != null) {
+      throw new IllegalStateException(
+          "ZkClient is not in init state. connect() has already been called.");
+    }
+
```
qqu0127, line 2521: Why do we need the lock at the beginning?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2508,11 +2509,17 @@ public void connect(final long maxMsToWaitUntilConnected)
...
   public void connect(final long maxMsToWaitUntilConnected, Watcher watcher)
       throws ZkInterruptedException, ZkTimeoutException, IllegalStateException {
+    acquireEventLock();
+
     if (isClosed()) {
       throw new IllegalStateException("ZkClient already closed!");
     }
+    if (_currentState != null) {
+      throw new IllegalStateException(
+          "ZkClient is not in init state. connect() has already been called.");
```
qqu0127, line 2519: If I understand correctly, we should be releasing the lock

rahulrane50: +1

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
xyuanlu: `_currentState` and `_isClosed` are set in critical section in `close()`. 
We need to check these 2 state and create connection in critical section as well.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2508,12 +2509,18 @@ public void connect(final long maxMsToWaitUntilConnected)
...
-    if (isClosed()) {
-      throw new IllegalStateException("ZkClient already closed!");
-    }
     boolean started = false;
-    acquireEventLock();
+
     try {
+      acquireEventLock();
+
+      if (isClosed()) {
```
rahulrane50, line 2517: @xyuanlu can you please help me in understanding how this gonna change current gap? Here is my understanding : 
1. Current isClosed() method also tries to take a zk event lock. But after your change we would try to acquire same zk event lock "interruptibly". I don't think there is much difference between these two right? 
2. I didn't understand why did we resolve old comment about not releasing lock in case of _currentState being null.

xyuanlu: Thanks for the comment.
1. Having state check and operation in the same critical section is a common practice. Taking the following flow char as example. In the code logic before this change, we do acquire lock before check if state is closed and release lock after that. The actual connection and close operation is also protected by this event lock. However, the `close()` operation can happen in between of 2 release lock and acquire lock. So generally we put when doing state check and operation in the same critical section.
Screenshot 2023-03-08 at 1 37 55 PM
2. lock.release() in the final block will be executed when try block finishes smoothly or throw exceptions. The lock is released there. 

I always find the java doc for ReentrantLock is the best place to refer to. It has the most accurate explanation and solid examples. :D (https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html#lockInterruptibly())

rahulrane50: Oh ok now I understand about this change! Thanks a lot @xyuanlu for explaining it patiently. Can we change the description for PR to this detailed explanation. I was wondering how are we adding extra check previously.

rahulrane50 (approved):
LGTM!

xyuanlu:
This change is ready to be merged. Approved by @rahulrane50

PR #2409, opened 2023-03-14
Implement timeout for auto reconnect

xyuanlu:
### Issues

### Description

### Tests

### Documentation (Optional)

### Commits

### Code Quality


meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -47,33 +52,41 @@
...
   private final long _initConnectionTimeout;
   private final long _reconnectTimeout;
 
+  private final ScheduledExecutorService _zkClientReconnectMonitor;
+  private ScheduledFuture<?> _reconnectMonitorFuture;
+  private ReentrantLock _zkClientConnectionMutex = new ReentrantLock();
+
   public ZkMetaClient(ZkMetaClientConfig config) {
-    _initConnectionTimeout = config.getConnectionInitTimeoutInMillis();
+      _initConnectionTimeout = config.getConnectionInitTimeoutInMillis();
```
desaikomal, line 79: nit: does this have extra space?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -47,33 +52,41 @@
...
     _reconnectTimeout = config.getMetaClientReconnectPolicy().getAutoReconnectTimeout();
     // TODO: Right new ZkClient reconnect using exp backoff with fixed max backoff interval. We should
-    // 1. Allow user to config max backoff interval (next PR)
-    // 2. Allow user to config reconnect policy (future PR)
+    // 1. Allow user to config max and init backoff interval
```
desaikomal, line 82: too many knobs are not always good. unless we have strong requirements from API compatibility

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -266,18 +279,25 @@ public void asyncSet(String key, T data, int version, AsyncCallback.StatCallback
 
   @Override
   public void connect() {
-    // TODO: throws IllegalStateException when already connected
     try {
+      _zkClientConnectionMutex.lock();
       _zkClient.connect(_initConnectionTimeout, _zkClient);
+      // register this client as state change listener to react to ZkClient EXPIRED event.
+      // When ZkClient has expired connection to ZK, it sill auto reconnect until ZkClient
```
desaikomal, line 286: nit: sill -> still

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -47,33 +52,41 @@
...
 
-public class ZkMetaClient<T> implements MetaClientInterface<T>, AutoCloseable {
+
+public class ZkMetaClient<T> implements MetaClientInterface<T>, AutoCloseable, IZkStateListener {
   private static final Logger LOG = LoggerFactory.getLogger(ZkMetaClient.class);
   private final ZkClient _zkClient;
   private final long _initConnectionTimeout;
   private final long _reconnectTimeout;
 
+  private final ScheduledExecutorService _zkClientReconnectMonitor;
```
desaikomal, line 74: nit: suggestion - if adding comments is allowed, can you add comment on what is the purpose of this Executor service??  not sure, what is coding guidelines

meta-client/src/main/java/org/apache/helix/metaclient/policy/ExponentialBackoffReconnectPolicy.java
```diff
@@ -34,20 +35,42 @@ public class ExponentialBackoffReconnectPolicy implements MetaClientReconnectPol
...
+  }
+
+  public ExponentialBackoffReconnectPolicy(long autoReconnectTimeout) {
+    _initBackoffInterval = DEFAULT_INIT_EXP_BACKOFF_RETRY_INTERVAL_MS;
+    _maxBackOffInterval = DEFAULT_MAX_EXP_BACKOFF_RETRY_INTERVAL_MS;
+    _autoReconnectTimeout = autoReconnectTimeout;
   }
 
-  public ExponentialBackoffReconnectPolicy(long maxBackOffInterval, long initBackoffInterval) {
+  // TODO: Allow user to pass maxBackOffInterval and initBackoffInterval.
```
desaikomal, line 62: if we don't use it, then pleae remove it.

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestConnectStateChangeListenerAndRetry.java
```diff
@@ -0,0 +1,150 @@
...
+  @AfterSuite
+  public void cleanUp() {
+
+  }
+
+  @Test
+  public void testConnectState() {
+    try (ZkMetaClient<String> zkMetaClient = createZkMetaClientReconnectTest()) {
+      zkMetaClient.connect();
+      zkMetaClient.connect();
```
desaikomal, line 81: should it say, already connected ???

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestConnectStateChangeListenerAndRetry.java
```diff
@@ -0,0 +1,150 @@
...
+
+  // test mock zkclient event
+  @Test(dependsOnMethods = "testConnectState")
+  public void testReConnectSucceed() throws InterruptedException {
+    try (ZkMetaClient<String> zkMetaClient = createZkMetaClientReconnectTest()) {
+      zkMetaClient.connect();
+      simulateZkStateReconnected(zkMetaClient.getZkClient());
+      Thread.sleep(AUTO_RECONNECT_WAIT_TIME_EXD);
+      // When ZK reconnect happens within timeout window, zkMetaClient should ba able to perform CRUD.
+      Assert.assertTrue(zkMetaClient.getZkClient().getConnection().getZookeeperState().isConnected());
```
desaikomal, line 96: can you test the CRUD operation as you are saying should work?

desaikomal (commented):
Overall, i understood the code :-)
Want to make sure that we have all test scenarios covered.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -266,18 +279,25 @@ public void asyncSet(String key, T data, int version, AsyncCallback.StatCallback
...
   public void connect() {
-    // TODO: throws IllegalStateException when already connected
     try {
+      _zkClientConnectionMutex.lock();
       _zkClient.connect(_initConnectionTimeout, _zkClient);
+      // register this client as state change listener to react to ZkClient EXPIRED event.
+      // When ZkClient has expired connection to ZK, it sill auto reconnect until ZkClient
+      // is closed or connection re-established.
+      // We will need to close ZkClient when user set retry connection timeout.
+      _zkClient.subscribeStateChanges(this);
```
mgao0, line 289: Curious: Does this state change event going into the event queue with other type of events? Let's say if zk client connection is expired, other events (such as child change or data change) processing will fail because meta client cannot read zk now, will this block the queue so that zk state change event never gets to be processed? I saw that there is a method setAsyncExecPoolSize(), wondering is this how we prevent this kind of situation?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -394,4 +414,69 @@ public byte[] serialize(T data, String path) {
   public T deserialize(byte[] bytes, String path) {
     return _zkClient.deserialize(bytes, path);
   }
+
+  /**
+   * A clean up method called when connect state change or MetaClient is closing.
+   * @param cancel If we want to cancel the reconnect monitor thread.
+   * @param close If we want to close ZkClient.
+   */
+  private void cleanUpAndClose(boolean cancel, boolean close) {
```
mgao0, line 423: Nit: It is already well explained in your comment, but it would be better if the method signature is something like `private void cleanUpAndClose(boolean cancelMonitor, boolean closeZKClient)`.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -394,4 +414,69 @@ public byte[] serialize(T data, String path) {
...
+    _zkClientConnectionMutex.lock();
+
+    if (close && !_zkClient.isClosed()) {
+      _zkClient.close();
+      LOG.info("ZkClient is closed");
+    }
+
+    if (cancel && _reconnectMonitorFuture != null) {
+      _reconnectMonitorFuture.cancel(true);
```
mgao0, line 432: Cancel the future before closing zk client?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -47,33 +52,41 @@
...
+
+public class ZkMetaClient<T> implements MetaClientInterface<T>, AutoCloseable, IZkStateListener {
   private static final Logger LOG = LoggerFactory.getLogger(ZkMetaClient.class);
   private final ZkClient _zkClient;
   private final long _initConnectionTimeout;
   private final long _reconnectTimeout;
 
+  private final ScheduledExecutorService _zkClientReconnectMonitor;
+  private ScheduledFuture<?> _reconnectMonitorFuture;
+  private ReentrantLock _zkClientConnectionMutex = new ReentrantLock();
```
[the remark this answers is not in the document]
xyuanlu: `getEventLock()._stateChangedCondition` will be used to notify all pending CRUD retries there is a state change happening and they can retry CRUD operation.
`_zkClientConnectionMutex ` is only used to guard the `_reconnectMonitorFuture `. I probably should change to a more accurate name to avoid confusion.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -394,4 +414,69 @@ public byte[] serialize(T data, String path) {
...
+      if (_reconnectMonitorFuture == null || _reconnectMonitorFuture.isCancelled()
+          || _reconnectMonitorFuture.isDone()) {
+        _reconnectMonitorFuture = _zkClientReconnectMonitor.schedule(() -> {
+          if (!_zkClient.getConnection().getZookeeperState().isConnected()) {
+            cleanUpAndClose(false, true);
+          }
+        }, _reconnectTimeout, TimeUnit.MILLISECONDS);
+        LOG.info("ZkClient is Disconnected, schedule a reconnect monitor after {}",
+            _reconnectTimeout);
+      }
```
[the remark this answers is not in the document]
xyuanlu: As discussed offline, `_reconnectMonitorFuture ` is read and modified in _reconnectMonitorFuture  in both `handleStateChanged` and  `cleanUpAndClose`. All read and assignment in both places need to be in critical section.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
xyuanlu: TFTR. Good suggestion. Will update.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
xyuanlu: TFTR. This is a good point.
As our offline discussed, since ZkClient has a single thread (_eventThread) handling all events, it is possible that one previous event handling callback blocking all following event being send out. This is a generic problem with single event thread. To coup with this, we
1. Mentioned explicitly in document that user should not register blocking callbacks to client. (ZooKeeper document also suggested this)
2. In Helix use case, all callbacks are just simply forward.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -266,18 +279,25 @@ public void asyncSet(String key, T data, int version, AsyncCallback.StatCallback
...
+      _zkClientConnectionMutex.unlock();
     }
   }
 
   @Override
   public void disconnect() {
-    // TODO: This is a temp impl for test only. no proper interrupt handling and error handling.
-    _zkClient.close();
+    cleanUpAndClose(true, true);
+    _zkClientReconnectMonitor.shutdownNow();
```
rahulrane50, line 300: Don't we want to take lock here?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -394,4 +414,69 @@ public byte[] serialize(T data, String path) {
...
+        }, _reconnectTimeout, TimeUnit.MILLISECONDS);
+        LOG.info("ZkClient is Disconnected, schedule a reconnect monitor after {}",
+            _reconnectTimeout);
+      }
+      _zkClientConnectionMutex.unlock();
+    } else if (state == Watcher.Event.KeeperState.SyncConnected
+        || state == Watcher.Event.KeeperState.ConnectedReadOnly) {
+      cleanUpAndClose(true, false);
+      LOG.info("ZkClient is SyncConnected, reconnect monitor thread is canceled (if any)");
+    }
```
rahulrane50, line 461: I have usually seen this at other places as well but just wondering should we add any error logs or some logging for "else" block here?

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -394,4 +414,69 @@ public byte[] serialize(T data, String path) {
...
+  // Cancel the monitor thread when connected.
+  @Override
+  public void handleStateChanged(Watcher.Event.KeeperState state) throws Exception {
+    if (state == Watcher.Event.KeeperState.Disconnected) {
+      // Expired. start a new event monitoring retry
+      _zkClientConnectionMutex.lockInterruptibly();
+      if (_reconnectMonitorFuture == null || _reconnectMonitorFuture.isCancelled()
+          || _reconnectMonitorFuture.isDone()) {
+        _reconnectMonitorFuture = _zkClientReconnectMonitor.schedule(() -> {
+          if (!_zkClient.getConnection().getZookeeperState().isConnected()) {
```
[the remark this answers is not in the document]
xyuanlu: _zkClient will never be null because it is a `final` variable that is initialized in constructor.
_connection will never be null because it is a `final` variable that is initialized in constructor.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
xyuanlu: `_zkClientReconnectMonitor` is not assigned to any new value after init. I don't think synchronization is needed.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -47,33 +52,41 @@
...
-    // 1. Allow user to config max backoff interval (next PR)
-    // 2. Allow user to config reconnect policy (future PR)
+    // 1. Allow user to config max and init backoff interval
+    // 2. Allow user to config reconnect policy
     _zkClient = new ZkClient(
         new ZkConnection(config.getConnectionAddress(), (int) config.getSessionTimeoutInMillis()),
         (int) _initConnectionTimeout, _reconnectTimeout /*use reconnect timeout for retry timeout*/,
         config.getZkSerializer(), config.getMonitorType(), config.getMonitorKey(),
         config.getMonitorInstanceName(), config.getMonitorRootPathOnly(), false);
+    _zkClientReconnectMonitor = Executors.newScheduledThreadPool(1);
```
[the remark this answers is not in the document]
xyuanlu: Good suggestion. Will update.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -47,33 +52,41 @@
...
+import org.apache.zookeeper.Watcher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static org.apache.helix.metaclient.impl.zk.util.ZkMetaClientUtil.convertZkEntryModeToMetaClientEntryMode;
 import static org.apache.helix.metaclient.impl.zk.util.ZkMetaClientUtil.translateZkExceptionToMetaclientException;
 
-public class ZkMetaClient<T> implements MetaClientInterface<T>, AutoCloseable {
+
+public class ZkMetaClient<T> implements MetaClientInterface<T>, AutoCloseable, IZkStateListener {
```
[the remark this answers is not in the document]
xyuanlu: TFTR.
`ZkMetaClient` as a user of `ZkClient`, it implements `Zkclient.IZkStateListener`, a listener type that `ZkClient provided`. Just as how `ZkClient` as a user of native ZooKeeper client implements the listener type that ZooKeeper provide . So it won't introduce circular dependencies.

Also introducing another object to listen on state change will add complexity and performance overhead. Please refer to the previous comment.

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestConnectStateChangeListenerAndRetry.java
```diff
@@ -0,0 +1,150 @@
...
+
+  @BeforeSuite
+  public void prepare() {
+    // start local zookeeper server
+    _zkServer = ZkMetaClientTestBase.startZkServer(ZK_ADDR);
+  }
+
+  @AfterSuite
+  public void cleanUp() {
+
```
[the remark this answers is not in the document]
xyuanlu: We on purposely override the clean up to not close ZK server since it is closed in test `testConnectStateChangeListener` to test state change handling when we lost connection to ZK (line 127).

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
xyuanlu: ZkClient has log for other state change event. We are not doing anything related to reconnect monitor in other state.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -394,4 +414,69 @@ public byte[] serialize(T data, String path) {
...
+    _zkClientConnectionMutex.unlock();
+  }
+
+  // Schedule a monitor to track ZkClient auto reconnect when Disconnected
+  // Cancel the monitor thread when connected.
+  @Override
+  public void handleStateChanged(Watcher.Event.KeeperState state) throws Exception {
+    if (state == Watcher.Event.KeeperState.Disconnected) {
+      // Expired. start a new event monitoring retry
+      _zkClientConnectionMutex.lockInterruptibly();
```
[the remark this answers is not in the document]
xyuanlu: I would prefer to use one mutex as introducing 2 locks and acquire then one after another will have higher risk of dead lock.

desaikomal (commented):
I want to say this is the most interesting code-review which i have seen in a while. I like the healthy discussion :-) continue

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
xyuanlu: TFTR. Will update.

xyuanlu:
> multiple metaclient/zkmetaclient objects and they all can use same monitor thread

Thanks for the review. I would prefer each ZkMetaClient has one monitor thread because

1. Each zkmetaclient objects need to monitor reconnect process and have its own counter with unique start time. It is hard to coordinate using one thread.
2. We need to maintain the lifecycle of this monitor thread. Having multiple zkmetaclient objects and they all can use same monitor thread would be hard to manage and cause thread leak.
3. We need this monitor because ZkClient does retry infinity. Other underlying metadata service may not have same reconnect protocol. It only applies to ZooKeeper.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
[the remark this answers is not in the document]
xyuanlu: Thanks for the explanation. Make sense I will update. :D

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -47,33 +53,47 @@
...
 import static org.apache.helix.metaclient.impl.zk.util.ZkMetaClientUtil.translateZkExceptionToMetaclientException;
 
+
 public class ZkMetaClient<T> implements MetaClientInterface<T>, AutoCloseable {
   private static final Logger LOG = LoggerFactory.getLogger(ZkMetaClient.class);
   private final ZkClient _zkClient;
   private final long _initConnectionTimeout;
   private final long _reconnectTimeout;
 
+  // After ZkClient become disconnect from ZK server, it keeps retrying connection until connection
```
rahulrane50, line 75: nit: "After ZKClient gets disconnected from ZK server"

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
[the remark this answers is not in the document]
rahulrane50: Yes if this mutex is only to protect this monitor thread then it should renamed to indicate so.

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -394,4 +421,75 @@ public byte[] serialize(T data, String path) {
...
+      if (cancel && _reconnectMonitorFuture != null) {
+        _reconnectMonitorFuture.cancel(true);
+        LOG.info("ZkClient reconnect monitor thread is canceled");
+      }
+    } finally {
+      _zkClientConnectionMutex.unlock();
+    }
+  }
+
+  private class ReconnectStateChangeListener implements IZkStateListener {
```
rahulrane50, line 447: Thanks for the change, this looks much cleaner now.

rahulrane50 (approved):
Thanks for making changes and sharing your thoughts on single/multiple monitor thread per metaclient object. I have follow-up comments on that :
1. Well we can create pool of threads and each metaclient object can schedule future task whenever needed with this unique start time. It would be wise use of thread pool as well.
2. management of lifycyles is still there but just needed to be done by one layer up. 
3. This is very good point. And i agree with this.
I think overall I don't have major concerns about this design but i would love to hear other's opinion as well. Please resolve all comments before merge. Thanks for working on this!

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestConnectStateChangeListenerAndRetry.java
```diff
@@ -0,0 +1,150 @@
...
+import static org.apache.helix.metaclient.constants.MetaClientConstants.DEFAULT_INIT_EXP_BACKOFF_RETRY_INTERVAL_MS;
+import static org.apache.helix.metaclient.constants.MetaClientConstants.DEFAULT_MAX_EXP_BACKOFF_RETRY_INTERVAL_MS;
+
+
+public class TestConnectStateChangeListenerAndRetry  {
+  protected static final String ZK_ADDR = "localhost:2181";
+  protected static ZkServer _zkServer;
+
+  private static final long AUTO_RECONNECT_TIMEOUT_MS_FOR_TEST = 3 * 1000;
+  private static final long AUTO_RECONNECT_WAIT_TIME_WITHIN = 1 * 1000;
```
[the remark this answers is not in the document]
xyuanlu: Sorry I missed this comment last time.
This `AUTO_RECONNECT_WAIT_TIME_WITHIN` and `AUTO_RECONNECT_WAIT_TIME_EXD` is indeed a relative number comparing to the `AUTO_RECONNECT_TIMEOUT_MS_FOR_TEST`, witch we used to initiate ZkMetaClient.

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestConnectStateChangeListenerAndRetry.java
xyuanlu: Updated.

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestConnectStateChangeListenerAndRetry.java
```diff
@@ -0,0 +1,180 @@
...
+          null);
+      zkClient.process(event);
+  }
+
+  @BeforeSuite
+  public void prepare() {
+    System.out.println("START TestConnectStateChangeListenerAndRetry at " + new Date(System.currentTimeMillis()));
+    // start local zookeeper server
+    _zkServer = ZkMetaClientTestBase.startZkServer(ZK_ADDR);
+  }
```
qqu0127, line 77: We should be careful about beforesuite, especially with static variable. 
Actually I'm curious whether this one or the beforesuite in `ZkMetaClientTestBase` got triggered, or both?

qqu0127 (approved):
Overall LGTM, one non-blocking question for my own understanding.
This is a great PR, good job.

desaikomal (approved):
i have looked through the main source, it looks good.. thanks

xyuanlu:
This change os ready to me merged. Approved by @desaikomal @rahulrane50  // Thanks!
Final commit message:
Implement timeout for auto reconnect
ZkClient does auto reconnect when it get's disconnected from Zookeeper. There is no timeout when reconnect can not be established. In ZkMetaclient, we have a separate thread to monitor reconnect status and close ZkClient when connection can not be reestablished when timed out.

PR #2410, opened 2023-03-15
use reconnect timeout for crud retry timeout

xyuanlu:
### Issues

#2237

### Description

Had an offline discussion with other committers. 
Since operation retry waits for connect state change event, it will retry when reconnect succeeded and retry will be canceled when reconnect errored or timed out. There is no need to have 2 configurable timeout for CRUD operation retry and reconnect retry.
This change remove crud operation retry timeout.

### Tests

NA

### Documentation (Optional)

### Commits

### Code Quality

### DIFF
--- a/meta-client/src/main/java/org/apache/helix/metaclient/policy/MetaClientReconnectPolicy.java
+++ b/meta-client/src/main/java/org/apache/helix/metaclient/policy/MetaClientReconnectPolicy.java
@@ -37,7 +37,6 @@ public interface MetaClientReconnectPolicy {
 
   RetryPolicyName getPolicyName();
 
-  // TODO: add reconnect timeout
   default  long getAutoReconnectTimeout() {
     return DEFAULT_AUTO_RECONNECT_TIMEOUT_MS;
   }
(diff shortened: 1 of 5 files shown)


meta-client/src/main/java/org/apache/helix/metaclient/policy/MetaClientReconnectPolicy.java
```diff
@@ -35,4 +38,7 @@ enum RetryPolicyName {
   RetryPolicyName getPolicyName();
 
   // TODO: add reconnect timeout
```
rahulrane50, line 40: nit: you can remove above TODO now.

rahulrane50:
The change LGTM but I have one basic question, so previously we used to have two knobs for "reconnect" and "retry" now we are combining both knobs. Now what if user doesn't want to retry the op if reconnect happens after suppose 5 mins? (previously user can set operationRetryTimeout lesser than 5 mins and op won't be retried right?)

rahulrane50 (commented):
The change LGTM but I have one basic question, so previously we used to have two knobs for "reconnect" and "retry" now we are combining both knobs. Now what if user doesn't want to retry the op if reconnect happens after suppose 5 mins? (previously user can set operationRetryTimeout lesser than 5 mins and op won't be retried right?)

xyuanlu:
> The change LGTM but I have one basic question, so previously we used to have two knobs for "reconnect" and "retry" now we are combining both knobs. Now what if user doesn't want to retry the op if reconnect happens after suppose 5 mins? (previously user can set operationRetryTimeout lesser than 5 mins and op won't be retried right?)

Thanks for the comments. OperationRetry depends on auto reconnect. As of now there is no use case of having 2 separate config. So I would prefer let keep the config "simple and stupid". Easier for user to use. :D

PR #2421, opened 2023-03-27
MetaClient LockClient Interface

Marcosrico:
### Issues

#2237 

### Description

Added LockClient interface with key methods. Added a LockInfo object class that will be implemented in a following PR. For now it is just a placeholder.

### Tests

### Code Quality


meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockClientInterface.java
```diff
@@ -0,0 +1,62 @@
...
+   * @param mode EntryMode identifying if the entry will be deleted upon client disconnect
+   * @return True if the lock is acquired. False if failed to acquire (catches exception).
+   */
+  boolean acquireLock(String key, LockInfo info, MetaClientInterface.EntryMode mode);
+
+  /**
+   * Renews lock for a TTL Node.
+   * Will fail if key is an invalid path or isn't of type TTL.
+   * @param key key to identify the entry
+   * @return True if the lock was renews. False if failed to renew (catches exception).
```
mgao0, line 39: nit: did you mean "renewed"?

Marcosrico: Yup, typo. Thanks!

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockClientInterface.java
```diff
@@ -0,0 +1,62 @@
...
+   * Obtains the metadata of a lock (the LockInfo).
+   * @param key key to identify the entry
+   * @return LockInfo object of the node at key. If key doesn't exist, raise exception.
+   */
+  LockInfo retrieveLock(String key);
+
+  /**
+   * Closes the underlying client.
+   */
+  void close();
```
qqu0127, line 61: iMHO, this doesn't need to be included. We can have it by implementing `AutoCloseable`

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockClientInterface.java
```diff
@@ -0,0 +1,62 @@
...
+   */
+  boolean renewTTLLock(String key);
+
+  /**
+   * Releases the lock by deleted the node at entry key.
+   * Will fail if key is an invalid path.
+   * @param key key to identify the entry
+   * @return True if the lock was released. False if failed to release (catches exception).
+   */
+  boolean releaseLock(String key);
```
qqu0127, line 49: Is this method idempotent? What will happen if user release a lock multiple times?

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockClientInterface.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Obtains the metadata of a lock (the LockInfo).
+   * @param key key to identify the entry
+   * @return LockInfo object of the node at key. If key doesn't exist, raise exception.
```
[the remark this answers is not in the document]
Marcosrico: Sounds good. I think returning null makes most sense

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockClientInterface.java
```diff
@@ -0,0 +1,62 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import org.apache.helix.metaclient.api.MetaClientInterface;
+
+public interface LockClientInterface {
+  /**
+   * Acquires a lock by creating a node at entry key and lockinfo info.
```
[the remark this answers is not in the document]
Marcosrico: Got it, I'll generalize the descriptions. Thanks!

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockClientInterface.java
Marcosrico: Sounds good.

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockClientInterface.java
xyuanlu: +1 on qqu's comment.
Probably we want more detail regarding what does "invalid path" mean here. 
When releasing an lock doe second time, do we want to differentiate already released lock or non-existing lock? It is hard to do when underlying datastore does remove lock entry when released.

Marcosrico: @qqu0127 This I think depends on the implementation of release lock, though the expected behavior of releasing a lock on one that either doesn't exist or already has been released should error. As per the design, releaslock releases the lock using transactional support to avoid race conditions.

@xyuanlu I think this is related to @mgao0 question above on creating a separate class to surface exceptions/errors. Depending on the implementation, that will or will not differentiate between those two errors (though I think it is a good idea to too). With a separate error/result wrapper class we can capture this error and return it, regardless of the underlying datastore.

qqu0127: @Marcosrico thanks for the reply.
Yes, there could be different implementations. But the idea is, we are now defining the interface, thus now is the time we define the behavior and semantics.
Back to the `releaseLock` behavior, idempotent API will have some benefits for simplifying logic especially in case of retry. It could be that, as long as the lock has been released, it returns true or success. 
IMO, it's just one way to define the behavior, and there are tradeoff. But the bottomline is we should document the expected behavior clearly (or clearly say it's completely up to the impl). We can discuss more on this 😄

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockClientInterface.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+import org.apache.helix.metaclient.api.MetaClientInterface;
+
+public interface LockClientInterface {
+  /**
+   * Acquires a lock at key.
+   * @param key key to identify the entry
+   * @param info Metadata of the lock
+   * @param mode EntryMode identifying if the entry will be deleted upon client disconnect
```
mgao0, line 29: Maybe we can mention what are the entry modes, might be easier to understand.

mgao0 (approved):
Thanks for addressing the comments! Left one minor comment, otherwise LGTM

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockClientInterface.java
Marcosrico: Sounds good, addressed!

PR #2429, opened 2023-04-03
MetaClient LockInfo Object Implementation

Marcosrico:
### Issues

https://github.com/apache/helix/issues/2237

### Description

Implementing the LockInfo object for MetaClient LockClient that stores relevant metadata of a lock. This is taken directly from Lock-in but made simpler.

### Tests

`LockInfoTest.java`

### Code Quality


meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
```diff
@@ -0,0 +1,193 @@
...
+   * @param lockId value of LOCK_ID attribute
+   * @param ownerId value of OWNER_ID attribute
+   * @param clientId value of CLIENT_ID attribute
+   * @param clientData value of CLIENT_DATA attribute
+   * @param grantTime the time the lock was granted
+   * @param lastRenewalTime the last time the lock was renewed
+   * @param timeout value of TIMEOUT attribute
+   */
+  public LockInfo(String lockId, String ownerId, String clientId,
+                  String clientData, long grantTime, long lastRenewalTime, Duration timeout) {
```
mgao0, line 97: In the original design we want to take clientData from client as a generic type, then we can serialize it and store it. I'm okay with changing the design to allow user to provide a serialized string of clientData, so they are responsible of serializing / deserializing the data. But please do an analysis on more known use cases, and see if will cause a huge overhead on using lock client.

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
```diff
@@ -0,0 +1,193 @@
...
+   * @param ownerId value of OWNER_ID attribute
+   * @param clientId value of CLIENT_ID attribute
+   * @param clientData value of CLIENT_DATA attribute
+   * @param grantTime the time the lock was granted
+   * @param lastRenewalTime the last time the lock was renewed
+   * @param timeout value of TIMEOUT attribute
+   */
+  private void setLockInfoFields(String lockId, String ownerId, String clientId, String clientData, long grantTime, long lastRenewalTime,
+                                 Duration timeout) {
+    _record.setSimpleField(LockInfoAttribute.LOCK_ID.name(), lockId == null ? DEFAULT_LOCK_ID_TEXT : lockId);
```
mgao0, line 115: Let's convert these lines into individual setters of each field, I think we may need them in lock client internal logic. Our goal is to completely wrap the ZNRecord so the external classes do not need to know that it is internally implemented using a ZNRecord.

meta-client/src/test/java/org/apache/helix/metaclient/recipes/lock/LockInfoTest.java
```diff
@@ -0,0 +1,50 @@
...
+ * under the License.
+ */
+
+
+import java.time.Duration;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class LockInfoTest {
```
mgao0, line 28: Let's also test convert a ZNRecord to LockInfo

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
Marcosrico: Thank you for that. After an offline discussion with @xyuanlu , we came to the conclusion that passing in the serialized version is easier to implement. Since the user can have their data as different objects, each can serialize their object however they wish. Furthermore, considering the znRecord does not have an explicit parameter for generic objects but rather we would put it under payload, it makes more sense to use their built in `getSimpleField()`. Let me know if this makes sense and answers your concerns, thanks!

mgao0: Sure sounds good. Please make sure this is clear to the users through Javadoc.

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
Marcosrico: I am not sure I understand what you mean, do you mean having a separate method to set each of these values?

mgao0: Yes, I meant that we can have a setter method for each of the values, right now you already have getters.

meta-client/src/test/java/org/apache/helix/metaclient/recipes/lock/LockInfoTest.java
Marcosrico: Not sure if i understand this. If lockinfo is a wrapper around znrecord, how would one convert a znrecord to lockinfo?

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
Marcosrico: Got it, thanks!

meta-client/src/test/java/org/apache/helix/metaclient/recipes/lock/LockInfoTest.java
mgao0: I meant this constructor `  public LockInfo(ZNRecord znRecord) `

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
```diff
@@ -0,0 +1,250 @@
...
+  private final ZNRecord _record;
+
+  /**
+   * The keys to lock information
+   */
+  public enum LockInfoAttribute {
+    LOCK_ID,
+    OWNER_ID,
+    CLIENT_ID,
+    //CLIENT_DATA is a string that represents the serialized object passed in from the user.
```
xyuanlu, line 50: nit, inline comment.

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
```diff
@@ -0,0 +1,250 @@
...
+    setLockInfoFields(DEFAULT_LOCK_ID_TEXT, DEFAULT_OWNER_ID_TEXT, DEFAULT_CLIENT_ID_TEXT, DEFAULT_CLIENT_DATA, DEFAULT_GRANTED_AT_LONG,
+        DEFAULT_LAST_RENEWED_AT_LONG, DEFAULT_TIMEOUT_DURATION);
+  }
+
+  /**
+   * Initialize a LockInfo with a ZNRecord, set all info fields to default data
+   * @param znRecord The ZNRecord contains lock node data that used to initialize the LockInfo
+   */
+  public LockInfo(ZNRecord znRecord) {
+    this();
```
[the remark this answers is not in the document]
Marcosrico: I don't quite understand. Do you mean change just the name of the variable instead of znrecord to something more generic or create a wrapper around znrecord?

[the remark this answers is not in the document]
mgao0: @xyuanlu I remember ZNRecord class is already used in some other MetaClient classes. @Marcosrico Can you confirm? If so, we can do a refactor later with a separate commit.

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
```diff
@@ -19,6 +19,231 @@
...
+   */
+  public void setLockId(String lockId) {
+    _record.setSimpleField(LockInfoAttribute.LOCK_ID.name(), lockId == null ? DEFAULT_LOCK_ID_TEXT : lockId);
+  }
+
+  /**
+   * Get the value for OWNER_ID attribute of the lock
+   * @param ownerId String representing the owner id
+   */
+  public void setOwnerId(String ownerId) {
```
xyuanlu, line 137: I would suggest adding more detail and some explanation on these fields.

Marcosrico: Hmm while I agree I am not sure what other information to put. The id of the owner is a unique string that represents the owner of each lock (similar to the rest of them). If you could give me an pointer to what additional information I can add that would be greatly appreciated, thank you!

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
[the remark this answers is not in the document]
Marcosrico: Thanks for the clarification, fixed it. Please take a look, thanks!

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
xyuanlu: There are 3 `ID` fields in this info. We probably want to explain what's the difference, what info is suitable for each of these fields, maybe with an example.

meta-client/src/main/java/org/apache/helix/metaclient/datamodel/DataRecord.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
+
+/**
+ * The DataRecord object is a wrapper around ZNRecord.
+ * TODO: Create an interface to decouple DataRecord and have a pluggable record store.
+ */
+public class DataRecord extends ZNRecord {
+  public DataRecord(String ZNODE_ID) {
```
xyuanlu, line 29: nit:
generic naming and lower case.

meta-client/src/main/java/org/apache/helix/metaclient/recipes/lock/LockInfo.java
Marcosrico: Okay that makes sense, I added more details though was hesitant on actual examples. Please let me know if this is detailed enough, thanks!

mgao0 (approved):
LGTM. Please address other comments before merge

PR #2432, opened 2023-04-05
persist watcher draft

xyuanlu:
### Issues

### Description

### Tests

### Documentation (Optional)

### Commits

### Code Quality


zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2921,10 +2988,47 @@ private void addChildListener(String path, IZkChildListener listener) {
...
+      _persistListenerMutex.lockInterruptibly();
+      if (listener instanceof IZkChildListener) {
+        addChildListener(path, (IZkChildListener) listener);
+      } else if (listener instanceof IZkDataListener) {
+        addDataListener(path, (IZkDataListener) listener);
+      }
+    } catch (InterruptedException ex) {
+      throw new ZkInterruptedException(ex);
+    } finally {
+      _persistListenerMutex.unlock();
```
junkaixue, line 3002: Would it be possible to make these pieces of code generic?

_mutx.lock();
passedin method.invoke(xxxx)

catch(e){}
finally{
_mutex.unlock();
}

xyuanlu: TFTR. Good point. Will update.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -384,6 +405,21 @@ public void unsubscribeStateChanges(
...
+        _childListener.forEach((k, v) -> paths.add(k));
+        _dataListener.forEach((k, v) -> paths.add(k));
+        paths.forEach(p -> {
+          try {
+            getConnection().removeWatches(p, this, WatcherType.Any);
+          } catch (InterruptedException | KeeperException e) {
+            LOG.info("Failed to remove persistent watcher for {} ", p, e);
+          }
+        });
+      };
```
qqu0127, line 420: This looks like BiFunciton out of box, https://docs.oracle.com/javase/8/docs/api/java/util/function/BiFunction.html, can we use it by any chance?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2927,4 +2990,54 @@ private void removeChildListener(String path, IZkChildListener listener) {
...
+  }
+
+  private void addPersistListener(String path, Object listener) {
+    ManipulateListener addListeners = (String, Object) -> {
+      if (listener instanceof IZkChildListener) {
+        addChildListener(path, (IZkChildListener) listener);
+      } else if (listener instanceof IZkDataListener) {
+        addDataListener(path, (IZkDataListener) listener);
+      }
+    };
```
qqu0127, line 3005: Hmmm, I'm so confused by the syntax.
On one hand you are defining the lambda expression with two inputs, on the other hand, the ManipulateListener defined here doesn't actually use the input. What's the point of using such bi-function?
(Plus, why use capital "String" as input name? It also confuses with the class name.)

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2927,4 +2990,54 @@ private void removeChildListener(String path, IZkChildListener listener) {
...
+
+    executeWithInPersistListenerMutex(removeListeners, path, listener);
+  }
+
+  private void executeWithInPersistListenerMutex(ManipulateListener runnable, String path,
+      Object listener) {
+    try {
+      _persistListenerMutex.lockInterruptibly();
+      runnable.run(path, listener);
```
qqu0127, line 3034: Similar concern as brought up above. Correct me if I'm wrong, I'm afraid the input (path, listener) here isn't actually being used.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2927,4 +2990,54 @@ private void removeChildListener(String path, IZkChildListener listener) {
...
+      } catch (KeeperException.NoWatcherException e) {
+        LOG.warn("Persist watcher is already removed");
+      }
+    };
+
+    executeWithInPersistListenerMutex(removeListeners, path, listener);
+  }
+
+  private void executeWithInPersistListenerMutex(ManipulateListener runnable, String path,
+      Object listener) {
```
qqu0127, line 3031: Not an issue, but thought on improving the raw Object class type: 
what if we create an empty interface and let the two listeners interface extend that as marker interface? It only looks slightly better.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -384,6 +405,21 @@ public void unsubscribeStateChanges(
...
+            getConnection().removeWatches(p, this, WatcherType.Any);
+          } catch (InterruptedException | KeeperException e) {
+            LOG.info("Failed to remove persistent watcher for {} ", p, e);
+          }
+        });
+      };
+      executeWithInPersistListenerMutex(removeAllListeners, null, null);
+    } else {
     synchronized (_childListener) {
       _childListener.clear();
```
qqu0127, line 424: nit: you can just return after the first "if", so no need to do "else". Same for below.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
xyuanlu: TFTR. Missed that when update signature. Updated.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -384,6 +405,21 @@ public void unsubscribeStateChanges(
   }
 
   public void unsubscribeAll() {
+    if (_usePersistWatcher) {
```
desaikomal, line 408: I am getting a bit confused. if we have recursive watcher and persist watcher, this code will change, correct?

xyuanlu: persist watcher only refers to non recursive data/child watcher. 
Same as data/child watcher, recursive watcher will has its own registry, we will need to clean that registry as well here, but it won't change current impl.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -2927,4 +2990,54 @@ private void removeChildListener(String path, IZkChildListener listener) {
       listeners.remove(listener);
     }
   }
+
+  interface ManipulateListener<T> {
+    void run() throws KeeperException, InterruptedException;
+  }
```
qqu0127, line 2996: Do we still need this? Can we simply use Runnable?

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1313,7 +1354,7 @@ private void fireAllEvents(WatchedEvent event) {
    * are deleted before the last page is fetched. The upstream caller should be able to handle this.
    */
   public List<String> getChildren(String path) {
-    return getChildren(path, hasListeners(path));
+    return getChildren(path, (hasListeners(path) && !_usePersistWatcher));
```
qqu0127, line 1357: check `_usePersistWatcher` first, this should save some computation!

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
```diff
@@ -1785,7 +1826,7 @@ public void run() throws Exception {
                 }
                 try {
                   // TODO: the data is redundantly read multiple times when multiple listeners exist
-                  data = readData(path, null, true);
+                  data = readData(path, null, false); ///true);
```
qqu0127, line 1829: cleanup the unused change?

qqu0127 (commented):
Overall, please add more comments on the behavior, especially code around `_usePersistWatcher`. Thanks

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
xyuanlu: Good suggestion! Updated

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
xyuanlu: Updated

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
xyuanlu: Thanks for the suggestion. I think Runnable.run don't have exception signature? We need to handle these exceptions explicitly.

qqu0127: Ok, that's a valid point. 
We can still use Callable with null return, to remove this wrapper. It's up to you.
(If you decide to stay with this interface, let's remove the generic, it's not used anyway)

xyuanlu: Yea already updated. :D

qqu0127 (approved):
Overall LGTM, small comments, please also have others review

meta-client/src/test/java/org/apache/helix/metaclient/impl/zk/TestUtil.java
```diff
@@ -0,0 +1,97 @@
...
+      throws NoSuchFieldException {
+    try {
+      return clazz.getDeclaredField(fieldName);
+    } catch (NoSuchFieldException e) {
+      Class superClass = clazz.getSuperclass();
+      if (superClass == null) {
+        throw e;
+      } else {
+        return getField(superClass, fieldName);
+      }
```
junkaixue, line 47: No need for else clause.

It could be:
if (superClass == null) {
        throw e;
 }
return getField(superClass, fieldName);

meta-client/src/main/java/org/apache/helix/metaclient/impl/zk/ZkMetaClient.java
```diff
@@ -315,7 +315,7 @@ public ConnectState getClientConnectionState() {
   @Override
   public boolean subscribeDataChange(String key, DataChangeListener listener, boolean skipWatchingNonExistNode) {
     _zkClient.subscribeDataChanges(key, new DataListenerAdapter(listener));
-    return false;
+    return true;
```
desaikomal, line 318: default "true" return type means that this method should be void.

xyuanlu: Good point. Will do interface change in following PR.

zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java
xyuanlu: Thanks for the suggestion. I will add a todo here and may change in following PR as this change is already large.

desaimom (approved):
LGTM.  Thanks for working on this.

